PR #5574, opened 2019-07-16
Option type info

mrambacher:
This change is another predecessor to the Configurable change.  It is an attempt to make the OptionTypeInfo structure richer as required by the Configurable code.

- The is_mutable boolean was replaced by flags, to allow more functionality to be added;
- The OptionTypeInfo can now support optional functions to get/set/compare options
- The Parse/Serialize functions now take an OptionTypeInfo as a parameter (instead of just the Type).


riversand963:
Should this be closed? See my comments on #5753

mrambacher:
This has been replaced by #6198 and #6422

PR #5584, opened 2019-07-17
Move the uncompression dictionary object out of the block cache

ltamasi:
Summary:
RocksDB has historically stored uncompression dictionary objects in the block
cache as opposed to storing just the block contents. This neccesitated
evicting the object upon table close. With the new code, only the raw blocks
are stored in the cache, eliminating the need for eviction.

In addition, the patch makes the following improvements:

1) Compression dictionary blocks are now prefetched/pinned similarly to
index/filter blocks.
2) A copy operation got eliminated when the uncompression dictionary is
retrieved.
3) Errors related to retrieving the uncompression dictionary are propagated as
opposed to silently ignored.

Note: the patch temporarily breaks the compression dictionary evicition stats.
They will be fixed in a separate phase.

Test Plan:
make asan_check


util/compression.h
```diff
@@ -216,36 +217,53 @@ struct CompressionDict {
 
 // Holds dictionary and related data, like ZSTD's digested uncompression
 // dictionary.
-struct UncompressionDict {
+struct UncompressionDict : public Cleanable {
+  // Block containing the data for the compression dictionary. It is non-empty
+  // only if the constructor that takes a string parameter is used.
+  std::string dict_;
```
riversand963: Trying to understand the change here.
If dict_ is empty, it means the data of the dictionary is in block cache. It also implicitly indicates that UncompressionDict does not own it.
If dict_ is not empty, it means UncompressionDict owns the data, and it means the data is not managed in the block cache?
Am I missing anything?

ltamasi: There are two cases:

1) The string constructor is used (as in BlockBasedTableBuilder): dict_ contains the data, slice_ points to dict_, and UncompressionDict is self-contained.
2) The Slice constructor is used (as on the read code paths): in this case dict_ is empty, slice_ is whatever was passed in, and UncompressionDict is not necessarily self-contained. However, since UncompressionDict is now a Cleanable, on the read paths we can and do transfer the ownership of the actual compression dictionary bytes from the CachableEntry to the UncompressionDict itself, once again creating a self-contained UncompressionDict object.

I will extend the comments to make this clearer.

ltamasi: P.S. In the case of the block cache, by "management" I mean releasing the cache handle when the UncompressionDict is destroyed.

table/block_based/uncompression_dict_reader.h
```diff
@@ -0,0 +1,66 @@
...
+struct BlockCacheLookupContext;
+class FilePrefetchBuffer;
+class GetContext;
+struct ReadOptions;
+struct UncompressionDict;
+
+// Provides access to the uncompression dictionary regardless of whether
+// it is owned by the reader or stored in the cache, or whether it is pinned
+// in the cache or not.
+class UncompressionDictReader {
```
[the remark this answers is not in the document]
ltamasi: Fixed, thanks.

util/compression.h
ltamasi: Added comments to the UncompressionDict constructors to explain the semantics.

table/block_based/uncompression_dict_reader.h
```diff
@@ -0,0 +1,65 @@
...
+      const ReadOptions& read_options, GetContext* get_context,
+      BlockCacheLookupContext* lookup_context,
+      CachableEntry<BlockContents>* uncompression_dict_block);
+
+  Status GetOrReadUncompressionDictionaryBlock(
+      FilePrefetchBuffer* prefetch_buffer, bool no_io, GetContext* get_context,
+      BlockCacheLookupContext* lookup_context,
+      CachableEntry<BlockContents>* uncompression_dict_block) const;
+
+ private:
```
riversand963: This 'private' can be removed.

PR #5589, opened 2019-07-18
Fix asan_check failures

anand1976:
Test plan:
TEST_TMPDIR=/dev/shm/rocksdb COMPILE_WITH_ASAN=1 OPT=-g make J=64 -j64 asan_check


miasantreble:
do you mind also fixing the following error under LITE?
```
db/import_column_family_test.cc:558:14: error: unused parameter ‘argc’ [-Werror=unused-parameter]
              ^
db/import_column_family_test.cc:558:27: error: unused parameter ‘argv’ [-Werror=unused-parameter]
                           ^
```

anand1976:
The LITE mode build fix is already committed in #5588

PR #5596, opened 2019-07-19
Block cache analyzer: Compute correlation of features and human readable trace file.

HaoyuHuang:
- Compute correlation between a few features and predictions, e.g., number of accesses since the last access vs number of accesses till the next access on a block. 
- Output human readable trace file so python can consume it. 

Test plan: make clean && USE_CLANG=1 make check -j32


tools/block_cache_trace_analyzer.cc
```diff
@@ -279,6 +304,34 @@ double percent(uint64_t numerator, uint64_t denomenator) {
...
+  return adjusted_time_stats;
+}
+
+int get_free_memory() {
+  FILE* fpipe = nullptr;
+  std::string command = "free -g | grep \"Mem\" | grep -o '[^ ]*$'";
+  char c = 0;
+  std::string output;
+  if (0 == (fpipe = (FILE*)popen(command.c_str(), "r"))) {
+    fprintf(stderr, "popen() failed.");
```
riversand963: Print a newline.

tools/block_cache_trace_analyzer.cc
```diff
@@ -314,13 +372,293 @@ void BlockCacheTraceAnalyzer::WriteMissRatioCurves() const {
...
+      (*label_predictions)[label].num_accesses_till_next_access.push_back(
+          num_accesses_since_last_access);
+      (*label_predictions)[label].elapsed_time_till_next_access.push_back(
+          elapsed_time_since_last_access);
+    }
+  }
+}
+
+void BlockCacheTraceAnalyzer::WriteMissRatioTimeline(uint64_t time_unit) const {
+  if (!cache_simulator_) {
```
riversand963: If you are not going to print some extra logging, I think you can combine these two early-return cases.

tools/block_cache_trace_analyzer.cc
```diff
@@ -7,11 +7,16 @@
 #ifdef GFLAGS
 #include "tools/block_cache_trace_analyzer.h"
 
+#include <stdio.h>
```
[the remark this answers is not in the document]
HaoyuHuang: Thanks. I changed to use cstdio, cstdlib.

tools/block_cache_trace_analyzer.cc
```diff
@@ -279,6 +304,34 @@ double percent(uint64_t numerator, uint64_t denomenator) {
...
+  }
+  return adjusted_time_stats;
+}
+
+int get_free_memory() {
+  FILE* fpipe = nullptr;
+  std::string command = "free -g | grep \"Mem\" | grep -o '[^ ]*$'";
+  char c = 0;
+  std::string output;
+  if (0 == (fpipe = (FILE*)popen(command.c_str(), "r"))) {
```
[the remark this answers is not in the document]
HaoyuHuang: As discussed offline, I removed this function and will add a GetFreeMemory function in env.

tools/block_cache_trace_analyzer.cc
```diff
@@ -279,6 +304,34 @@ double percent(uint64_t numerator, uint64_t denomenator) {
...
+  for (auto const& time : time_stats) {
+    adjusted_time_stats[static_cast<uint64_t>(time.first / time_unit)] +=
+        time.second;
+  }
+  return adjusted_time_stats;
+}
+
+int get_free_memory() {
+  FILE* fpipe = nullptr;
+  std::string command = "free -g | grep \"Mem\" | grep -o '[^ ]*$'";
```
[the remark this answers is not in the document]
HaoyuHuang: As discussed offline, I removed this function and will add a GetFreeMemory function in env. I will use sysinfo for Linux platforms.

tools/block_cache_trace_analyzer.cc
HaoyuHuang: Thanks. I removed this function.

tools/block_cache_trace_analyzer.cc
HaoyuHuang: Sure. I combined these two.

PR #5608, opened 2019-07-22
Parallelize file_reader_writer_test in order to reduce timeouts

anand1976:
Test Plan:
make check
buck test @mode/dev-tsan internal_repo_rocksdb/repo:file_reader_writer_test -- --run-disabled

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -480,7 +480,6 @@ TESTS = \
 	fault_injection_test \
 	filelock_test \
 	filename_test \
-	file_reader_writer_test \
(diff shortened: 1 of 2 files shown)


anand1976:
Good point! I wonder if the test uses sync_repo.sh. If it does, then buckify_rocksdb.py will update the TARGETS file. Let me check.

anand1976:
In any case, it doesn't hurt to update the TARGETS file, so let me just do that.

PR #5610, opened 2019-07-23
Block cache simulator: Add pysim to simulate caches using reinforcement learning.

HaoyuHuang:
This PR implements cache eviction using reinforcement learning. It includes two implementations: 
1. An implementation of Thompson Sampling for the Bernoulli Bandit [1].    
2. An implementation of LinUCB with disjoint linear models [2].

The idea is that a cache uses multiple eviction policies, e.g., MRU, LRU, and LFU. The cache learns which eviction policy is the best and uses it upon a cache miss. 
Thompson Sampling is contextless and does not include any features. 
LinUCB includes features such as level, block type, caller, column family id to decide which eviction policy to use. 

[1] Daniel J. Russo, Benjamin Van Roy, Abbas Kazerouni, Ian Osband, and Zheng Wen. 2018. A Tutorial on Thompson Sampling. Found. Trends Mach. Learn. 11, 1 (July 2018), 1-96. DOI: https://doi.org/10.1561/2200000070
[2] Lihong Li, Wei Chu, John Langford, and Robert E. Schapire. 2010. A contextual-bandit approach to personalized news article recommendation. In Proceedings of the 19th international conference on World wide web (WWW '10). ACM, New York, NY, USA, 661-670. DOI=http://dx.doi.org/10.1145/1772690.1772758


.gitignore
```diff
@@ -34,6 +34,7 @@ manifest_dump
 sst_dump
 blob_dump
 block_cache_trace_analyzer
+tools/*pyc
```
riversand963: `*.pyc`?

tools/__init__.py
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env python3
```
riversand963: Not a python expert.
Are you going to put all python scripts under `tools` in the same package? Is it better to move block cache trace analyzer and other related scripts in the same sub directory under `tools`?

tools/block_cache_pysim.py
```diff
@@ -0,0 +1,857 @@
...
+kSampleSize = 16  # The sample size used when performing eviction.
+kMicrosInSecond = 1000000
+kSecondsInMinute = 60
+kSecondsInHour = 3600
+
+
+class TraceRecord:
+    """A trace record represents a block access."""
+
+    def __init__(
```
riversand963: Add a comment or linter rule. It's better to enforce parity between this and its C++ counterparts. If it's too hard, then I am fine with a comment linking these two, reminding to modify the other if one is being modified.

tools/block_cache_pysim.sh
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env bash
```
riversand963: Curious why not use python script, but you don't have to change it. :)

.gitignore
HaoyuHuang: Thanks. I changed to *.pyc.

tools/__init__.py
HaoyuHuang: I created block_cache_trace directory. I moved all scripts and block cache analyzer.cc into the sub directory.

tools/block_cache_pysim.sh
HaoyuHuang: I'm not sure how effective is the python multi-threading so I used the bash script to start multiple processes to speed up simulation.

tools/block_cache_pysim.py
HaoyuHuang: As discussed offline, I added a comment inline.

PR #5627, opened 2019-07-25
WriteUnPrepared: savepoint support

lth:
Add savepoint support when the current transaction has flushed unprepared batches.

Rolling back to savepoint is similar to rolling back a transaction. It requires the set of keys that have changed since the savepoint, re-reading the keys at the snapshot at that savepoint, and the restoring the old keys by writing out another unprepared batch.

For this strategy to work though, we must be capable of reading keys at a savepoint. This does not work if keys were written out using the same sequence number before and after a savepoint. Therefore, when we flush out unprepared batches, we must split the batch by savepoint if any savepoints exist.

eg. If we have the following:
```
Put(A)
Put(B)
Put(C)
SetSavePoint()
Put(D)
Put(E)
SetSavePoint()
Put(F)
```

Then we will write out 3 separate unprepared batches:
```
Put(A) 1
Put(B) 1
Put(C) 1
Put(D) 2
Put(E) 2
Put(F) 3
```

This is so that when we rollback to eg. the first savepoint, we can just read keys at snapshot_seq = 1.

### DIFF
--- a/db/write_batch.cc
+++ b/db/write_batch.cc
@@ -519,6 +519,210 @@ Status WriteBatch::Iterate(Handler* handler) const {
                                      rep_.size());
 }
 
+Status WriteBatchInternal::Iterate(const WriteBatch* wb,
+                                   WriteBatch::Handler* handler, size_t begin,
+                                   size_t end) {
+  if (begin > wb->rep_.size() || end > wb->rep_.size() || end < begin) {
+    return Status::Corruption("Invalid start/end bounds for Iterate");
+  }
+  assert(begin <= end);
+  Slice input(wb->rep_.data() + begin, static_cast<size_t>(end - begin));
+  bool whole_batch =
+      (begin == WriteBatchInternal::kHeader) && (end == wb->rep_.size());
+
+  Slice key, value, blob, xid;
+  // Sometimes a sub-batch starts with a Noop. We want to exclude such Noops as
+  // the batch boundary symbols otherwise we would mis-count the number of
+  // batches. We do that by checking whether the accumulated batch is empty
+  // before seeing the next Noop.
+  bool empty_batch = true;
+  int found = 0;
+  Status s;
+  char tag = 0;
+  uint32_t column_family = 0;  // default
+  bool last_was_try_again = false;
+  bool handler_continue = true;
+  while (((s.ok() && !input.empty()) || UNLIKELY(s.IsTryAgain()))) {
+    handler_continue = handler->Continue();
+    if (!handler_continue) {
+      break;
+    }
+
+    if (LIKELY(!s.IsTryAgain())) {
+      last_was_try_again = false;
+      tag = 0;
+      column_family = 0;  // default
+
+      s = ReadRecordFromWriteBatch(&input, &tag, &column_family, &key, &value,
+                                   &blob, &xid);
+      if (!s.ok()) {
+        return s;
+      }
+    } else {
+      assert(s.IsTryAgain());
+      assert(!last_was_try_again);  // to detect infinite loop bugs
+      if (UNLIKELY(last_was_try_again)) {
+        return Status::Corruption(
+            "two consecutive TryAgain in WriteBatch handler; this is either a "
+            "software bug or data corruption.");
+      }
+      last_was_try_again = true;
+      s = Status::OK();
+    }
+
+    switch (tag) {
+      case kTypeColumnFamilyValue:
+      case kTypeValue:
+        assert(wb->content_flags_.load(std::memory_order_relaxed) &
+               (ContentFlags::DEFERRED | ContentFlags::HAS_PUT));
+        s = handler->PutCF(column_family, key, value);
+        if (LIKELY(s.ok())) {
+          empty_batch = false;
+          found++;
+        }
+        break;
+      case kTypeColumnFamilyDeletion:
+      case kTypeDeletion:
+        assert(wb->content_flags_.load(std::memory_order_relaxed) &
+               (ContentFlags::DEFERRED | ContentFlags::HAS_DELETE));
+        s = handler->DeleteCF(column_family, key);
+        if (LIKELY(s.ok())) {
+          empty_batch = false;
+          found++;
+        }
+        break;
+      case kTypeColumnFamilySingleDeletion:
+      case kTypeSingleDeletion:
+        assert(wb->content_flags_.load(std::memory_order_relaxed) &
+               (ContentFlags::DEFERRED | ContentFlags::HAS_SINGLE_DELETE));
+        s = handler->SingleDeleteCF(column_family, key);
+        if (LIKELY(s.ok())) {
+          empty_batch = false;
+          found++;
+        }
+        break;
+      case kTypeColumnFamilyRangeDeletion:
+      case kTypeRangeDeletion:
+        assert(wb->content_flags_.load(std::memory_order_relaxed) &
+               (ContentFlags::DEFERRED | ContentFlags::HAS_DELETE_RANGE));
+        s = handler->DeleteRangeCF(column_family, key, value);
+        if (LIKELY(s.ok())) {
+          empty_batch = false;
+          found++;
+        }
+        break;
+      case kTypeColumnFamilyMerge:
+      case kTypeMerge:
+        assert(wb->content_flags_.load(std::memory_order_relaxed) &
+               (ContentFlags::DEFERRED | ContentFlags::HAS_MERGE));
+        s = handler->MergeCF(column_family, key, value);
+        if (LIKELY(s.ok())) {
+          empty_batch = false;
+          found++;
+        }
+        break;
+      case kTypeColumnFamilyBlobIndex:
+      case kTypeBlobIndex:
+        assert(wb->content_flags_.load(std::memory_order_relaxed) &
+               (ContentFlags::DEFERRED | ContentFlags::HAS_BLOB_INDEX));
+        s = handler->PutBlobIndexCF(column_family, key, value);
+        if (LIKELY(s.ok())) {
+          found++;
+        }
+        break;
+      case kTypeLogData:
+        handler->LogData(blob);
+        // A batch might have nothing but LogData. It is still a batch.
+        empty_batch = false;
+        break;
+      case kTypeBeginPrepareXID:
+        assert(wb->content_flags_.load(std::memory_order_relaxed) &
+               (ContentFlags::DEFERRED | ContentFlags::HAS_BEGIN_PREPARE));
+        handler->MarkBeginPrepare();
+        empty_batch = false;
+        if (!handler->WriteAfterCommit()) {
+          s = Status::NotSupported(
+              "WriteCommitted txn tag when write_after_commit_ is disabled (in "
+              "WritePrepared/WriteUnprepared mode). If it is not due to "
+              "corruption, the WAL must be emptied before changing the "
+              "WritePolicy.");
+        }
+        if (handler->WriteBeforePrepare()) {
+          s = Status::NotSupported(
+              "WriteCommitted txn tag when write_before_prepare_ is enabled "
+              "(in WriteUnprepared mode). If it is not due to corruption, the "
+              "WAL must be emptied before changing the WritePolicy.");
+        }
+        break;
+      case kTypeBeginPersistedPrepareXID:
+        assert(wb->content_flags_.load(std::memory_order_relaxed) &
+               (ContentFlags::DEFERRED | ContentFlags::HAS_BEGIN_PREPARE));
+        handler->MarkBeginPrepare();
+        empty_batch = false;
+        if (handler->WriteAfterCommit()) {
+          s = Status::NotSupported(
+              "WritePrepared/WriteUnprepared txn tag when write_after_commit_ "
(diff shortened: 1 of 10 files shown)


utilities/transactions/write_unprepared_txn.cc
```diff
@@ -519,6 +641,104 @@ void WriteUnpreparedTxn::Clear() {
...
+  return s;
+}
+
+Status WriteUnpreparedTxn::PopSavePoint() {
+  Status s = PessimisticTransaction::PopSavePoint();
+  if (!s.IsNotFound()) {
+    return s;
+  }
+
+  if (wup_save_points_ != nullptr && !wup_save_points_->empty()) {
```
maysamyabandeh: I am confused! Dont we have a WriteUnprepared::SavePoint per TransactionBase::SavePoint?

utilities/transactions/write_unprepared_txn.h
```diff
@@ -199,6 +208,31 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
...
+    // old values.
+    //
+    // Since all unprep_seqs_ sets further down the stack must be subsets, this
+    // can potentially deduplicated by just storing set difference.
+    std::map<SequenceNumber, size_t> unprep_seqs_;
+    TransactionBaseImpl::SavePoint save_point_;
+
+    SavePoint(const std::map<SequenceNumber, size_t>& seqs,
+              TransactionBaseImpl::SavePoint&& save)
+        : unprep_seqs_(seqs), save_point_(std::move(save)){};
```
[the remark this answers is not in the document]
lth: I got rid of this. `TransactionBaseImpl::SavePoint` will now always live in on `TransactionBaseImpl::save_points_`

utilities/transactions/write_unprepared_txn.h
```diff
@@ -199,6 +208,31 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
...
+
+  // This holds savepoints that were set on write batches that have been already
+  // flushed.
+  std::unique_ptr<autovector<WriteUnpreparedTxn::SavePoint>> wup_save_points_;
+
+  // Record the size of the write batch at this savepoint. This information is
+  // actually also recorded in WriteBatch::save_points_, but it is not easily
+  // accessible. This information is used in FlushWriteBatchToDB so that data in
+  // different savepoints get flushed in different batches.
+  std::unique_ptr<autovector<size_t>> save_point_boundaries_;
```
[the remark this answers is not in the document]
lth: I added some more comments to verify, but basically `WriteUnpreparedTxn::SavePoint` only tracks savepoints for flushed unprepared batches. This only tracks savepoints for the current in-memory write batch.

utilities/transactions/write_unprepared_txn.h
```diff
@@ -199,6 +208,31 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
...
   // value when calling RollbackToSavepoint.
   SequenceNumber largest_validated_seq_;
+
+  struct SavePoint {
+    // Record of unprep_seqs_ at this savepoint. The set of unprep_seq is
+    // used during RollbackToSavepoint to determine visibility when restoring
+    // old values.
+    //
+    // Since all unprep_seqs_ sets further down the stack must be subsets, this
+    // can potentially deduplicated by just storing set difference.
```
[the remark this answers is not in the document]
lth: There isn't necessarily a TODO, since I don't know if spending extra time computing set difference is worth the memory savings (especially if you don't expect unprep_seqs to be that large).

If it's too confusing, I can just get rid of the comment.

utilities/transactions/write_unprepared_txn.cc
lth: I added more comments in the header to describe the relationship between all the data structures holding savepoint info.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -236,6 +238,20 @@ Status WriteUnpreparedTxn::MaybeFlushWriteBatchToDB() {
 }
 
 Status WriteUnpreparedTxn::FlushWriteBatchToDB(bool prepared) {
+  // If the current write batch contains savepoints, then some special handling
+  // is required so that RollbackToSavepoint can work.
+  //
+  // RollbackToSavepoint is not supported after Prepare() is called, so only do
```
maysamyabandeh: Can we assert that if prepared is true, save_point_boundaries_ is empty?

db/write_batch.cc
```diff
@@ -2010,4 +1820,207 @@ size_t WriteBatchInternal::AppendedByteSize(size_t leftByteSize,
   }
 }
 
+Status WriteBatchInternal::Iterate(const WriteBatch* wb,
```
maysamyabandeh: If we put this function right below WriteBatch::Iterate observing the changes would be easier as "git diff" would only highly the changes to the body of the function.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -292,6 +308,106 @@ Status WriteUnpreparedTxn::FlushWriteBatchToDB(bool prepared) {
...
+    Status MarkCommit(const Slice&) override {
+      return Status::InvalidArgument();
+    }
+
+    Status MarkRollback(const Slice&) override {
+      return Status::InvalidArgument();
+    }
+  };
+
+  WriteBatchWithIndex wb(wpt_db_->DefaultColumnFamily()->GetComparator(), 0,
```
maysamyabandeh: Why default?

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -292,6 +308,106 @@ Status WriteUnpreparedTxn::FlushWriteBatchToDB(bool prepared) {
...
+                         true, 0);
+  // Swap with write_batch_ so that wb contains the complete write batch. The
+  // actual write batch that will be flushed to DB will be built in
+  // write_batch_, and will be read by FlushWriteBatchToDBInternal.
+  std::swap(wb, write_batch_);
+
+  size_t prev_boundary = WriteBatchInternal::kHeader;
+  const bool kPrepared = true;
+  for (size_t i = 0; i < save_point_boundaries_->size(); i++) {
+    WriteBatchInternal::InsertNoop(write_batch_.GetWriteBatch());
```
maysamyabandeh: Why insert noop?

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -292,6 +308,106 @@ Status WriteUnpreparedTxn::FlushWriteBatchToDB(bool prepared) {
...
+      wup_save_points_.reset(new autovector<WriteUnpreparedTxn::SavePoint>());
+    }
+    wup_save_points_->emplace_back(unprep_seqs_);
+
+    prev_boundary = curr_boundary;
+    write_batch_.Clear();
+  }
+
+  save_point_boundaries_->clear();
+  WriteBatchInternal::InsertNoop(write_batch_.GetWriteBatch());
```
maysamyabandeh: why insert noop?

utilities/transactions/write_unprepared_txn.h
```diff
@@ -199,6 +208,49 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
...
+  // already recorded in WriteBatch::save_points_, but because it is not easily
+  // accessible, it is duplicated here. If the contents of
+  // WriteBatch::save_points_ were exposed, and can be traversed in LIFO order,
+  // then we can get rid of save_point_boundaries_.
+  //
+  // Based on this information, here are some invariants:
+  // size(save_point_boundaries_) == size(write_batch_.save_points_)
+  // size(wup_save_points_) + size(save_point_boundaries_) = size(save_points_)
+  //
+  std::unique_ptr<autovector<WriteUnpreparedTxn::SavePoint>> wup_save_points_;
```
maysamyabandeh: I suggest renaming them to this:
s/wup_save_points_/wup_persisted_save_points_
s/save_point_boundaries_/wup_inmem_save_pionts_

utilities/transactions/write_unprepared_txn.h
```diff
@@ -199,6 +208,49 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
...
+  // which keys have been modified in every savepoint.
+  //
+  // WriteUnpreparedTxn::wup_save_points_ holds information about savepoints set
+  // on unprepared batches that have already flushed. It just holds the
+  // unprep_seqs_ at that savepoint, so that the rollback process can determine
+  // which keys were visible at that point in time.
+  //
+  // WriteUnpreparecTxn::save_point_boundaries_ holds information about
+  // savepoints on the current in-memory write_batch_. It simply records the
+  // size of the write batch at every savepoint. This information is actually
```
maysamyabandeh: The line after "This information" could be replaced with a simple TODO:
// TODO(lth): remove the redundancy between save_point_boundaries_ and write_batch_.save_points_

utilities/transactions/write_unprepared_txn.h
```diff
@@ -199,6 +208,49 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
...
+  // 1. TransactionBaseImpl::save_points_
+  // 2. WriteUnpreparedTxn::wup_save_points_
+  // 3. WriteUnpreparecTxn::save_point_boundaries_
+  //
+  // TransactionBaseImpl::save_points_ holds information about all write
+  // batches, including the current in-memory write_batch_, or unprepared
+  // batches that have been written out. Its responsibility is just to track
+  // which keys have been modified in every savepoint.
+  //
+  // WriteUnpreparedTxn::wup_save_points_ holds information about savepoints set
```
maysamyabandeh: If we do the renaming, the this and next paragraph would be redundant as the name already clarifies what each data structure do.

include/rocksdb/write_batch.h
```diff
@@ -272,6 +272,7 @@ class WriteBatch : public WriteBatchBase {
 
    protected:
     friend class WriteBatch;
+    friend class WriteBatchInternal;
```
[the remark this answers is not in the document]
maysamyabandeh: Good point. Let's keep the friend relation. When it comes to public API the least changes the better.

utilities/transactions/write_unprepared_txn.cc
maysamyabandeh: Makes sense. I guess it is copied from what we did here: 
https://github.com/facebook/rocksdb/blob/399f477818578c0d3e4614f6f148e8d7859121a2/utilities/transactions/transaction_base.cc#L33
But the rationale might not be clear to future readers. We should either put an inline comment here or make an inline function in TransactionBase::InitWriteBatch and perform InsertNoop inside that function.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -524,6 +648,119 @@ void WriteUnpreparedTxn::Clear() {
...
+         (save_points_ ? save_points_->size() : 0));
+  if (save_point_boundaries_ != nullptr && save_point_boundaries_->size() > 0) {
+    Status s = PessimisticTransaction::PopSavePoint();
+    assert(!s.IsNotFound());
+    save_point_boundaries_->pop_back();
+    return s;
+  }
+
+  if (wup_save_points_ != nullptr && !wup_save_points_->empty()) {
+    write_batch_.SetSavePoint();
```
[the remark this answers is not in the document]
maysamyabandeh: Since it is not intuitive, a short inline comment would help.

utilities/transactions/write_unprepared_txn.cc
maysamyabandeh: Apparently it is a backup comparator in case cf id is not valid. Does not sound safe to me as it would swallow a potential bug with cf id. But anyway here is not the place to fix it.
Perhaps a short inline comment explaining why these default values are passes to constructor would help.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -680,27 +685,31 @@ Status WriteUnpreparedTxn::RollbackToSavePoint() {
...
     const auto cfid = cfkey.first;
     const auto& keys = cfkey.second;
+
+    roptions.snapshot = top.snapshot_->snapshot();
+    SequenceNumber min_uncommitted =
+        static_cast_with_check<const SnapshotImpl, const Snapshot>(
+            roptions.snapshot)
+            ->min_uncommitted_;
+    SequenceNumber snap_seq = roptions.snapshot->GetSequenceNumber();
+    WriteUnpreparedTxnReadCallback callback(wupt_db_, snap_seq, min_uncommitted,
```
maysamyabandeh: Why is this moved to inside the for loop?

utilities/transactions/write_unprepared_txn_db.cc
```diff
@@ -280,7 +280,7 @@ Status WriteUnpreparedTxnDB::Initialize(
     }
 
     wupt->write_batch_.Clear();
-    WriteBatchInternal::InsertNoop(wupt->write_batch_.GetWriteBatch());
+    wupt->InitWriteBatch();
```
maysamyabandeh: why not passing clear=true?

utilities/transactions/write_unprepared_txn.cc
lth: I thought I had to recalculate this value, but turns out not to be the case. I'll change this.

PR #5628, opened 2019-07-25
Fix PopSavePoint to merge info into the previous savepoint

### LINKED ISSUE #5618: PopSavePoint discards lock information
### Expected behavior
If you have:
```
txn->SetSavePoint();
txn->Put("A")
txn->SetSavePoint();
txn->Put("B")
txn->PopSavePoint()
txn->RollbackToSavePoint()
```

Then the write batch is rolled back to the first savepoint, meaning that both keys A and B are rolled back. I also expect that keys A and B will both be unlocked as well.

### Actual behavior
Only A is unlocked, and B remains locked. The problem is that every savepoint tracks keys locked since its last savepoint, and that information is used for unlocking if RollbackToSavePoint is called. When PopSavePoint is called, instead of simply popping off the stack and discarding that information, the key set should be merged with the key set of the previous savepoint instead, so that all keys get unlocked.

Add this to `utilities/transactions/transaction_test.cc`
```
TEST_P(TransactionTest, SavepointTest4) {
  WriteOptions write_options;
  ReadOptions read_options;
  TransactionOptions txn_options;
  Status s;

  txn_options.lock_timeout = 1;  // 1 ms
  Transaction* txn1 = db->BeginTransaction(write_options, txn_options);
  ASSERT_TRUE(txn1);

  txn1->SetSavePoint(); // 1
  s = txn1->Put("A", "");
  ASSERT_OK(s);

  txn1->SetSavePoint(); // 2
  s = txn1->Put("B", "");
  ASSERT_OK(s);

  s = txn1->PopSavePoint(); // Remove 2
  ASSERT_OK(s);

  ASSERT_OK(txn1->RollbackToSavePoint());  // Rollback to 1

  // Verify that everything was rolled back.
  std::string value;
  s = txn1->Get(read_options, "A", &value);
  ASSERT_TRUE(s.IsNotFound());

  s = txn1->Get(read_options, "B", &value);
  ASSERT_TRUE(s.IsNotFound());

  // Nothing should be locked
  Transaction* txn2 = db->BeginTransaction(write_options, txn_options);
  ASSERT_TRUE(txn2);

  s = txn2->Put("A", "");
  ASSERT_OK(s);

  s = txn2->Put("B", "");
  ASSERT_OK(s); // This fails

  delete txn2;
  delete txn1;
}
```

lth:
Transaction::RollbackToSavePoint undos the modification made since the SavePoint beginning, and also unlocks the corresponding keys, which are tracked in the last SavePoint. Currently ::PopSavePoint simply discard these tracked keys, leaving them locked in the lock manager. This breaks a subsequent ::RollbackToSavePoint behavior as it loses track of such keys, and thus cannot unlock them. The patch fixes ::PopSavePoint by passing on the track key information to the previous SavePoint.
Fixes #5618

### DIFF
--- a/utilities/transactions/transaction_base.cc
+++ b/utilities/transactions/transaction_base.cc
@@ -227,7 +227,7 @@ Status TransactionBaseImpl::PopSavePoint() {
           } else {
             TransactionKeyMapInfo& prev_info_val = prev_info->second;
 
-            assert(prev_info_val.seq < info.seq);
+            assert(prev_info_val.seq <= info.seq);
             prev_info_val.num_reads += info.num_reads;
             prev_info_val.num_writes += info.num_writes;
             prev_info_val.exclusive |= info.exclusive;
--- a/utilities/transactions/transaction_base.cc
+++ b/utilities/transactions/transaction_base.cc
@@ -196,47 +196,42 @@ Status TransactionBaseImpl::PopSavePoint() {
   // to inherit tracked_keys in B so that if we rollback to savepoint A, we
   // remember to unlock keys in B. If there is no other savepoint below, then we
   // can safely discard savepoint info.
-  if (save_points_->size() > 1) {
+  if (save_points_->size() == 1) {
+    save_points_->pop();
+  } else {
     TransactionBaseImpl::SavePoint top;
     std::swap(top, save_points_->top());
     save_points_->pop();
 
-    const TransactionKeyMap& curr_key_map = top.new_keys_;
-    TransactionKeyMap& prev_key_map = save_points_->top().new_keys_;
+    const TransactionKeyMap& curr_cf_key_map = top.new_keys_;
+    TransactionKeyMap& prev_cf_key_map = save_points_->top().new_keys_;
 
-    for (const auto& key_map_iter : curr_key_map) {
-      uint32_t column_family_id = key_map_iter.first;
-      auto& keys = key_map_iter.second;
+    for (const auto& curr_cf_key_iter : curr_cf_key_map) {
+      uint32_t column_family_id = curr_cf_key_iter.first;
+      const std::unordered_map<std::string, TransactionKeyMapInfo>& curr_keys =
+          curr_cf_key_iter.second;
 
       // If cfid was not previously tracked, just copy everything over.
-      auto prev_tracked_keys = prev_key_map.find(column_family_id);
-      if (prev_tracked_keys == prev_key_map.end()) {
-        prev_key_map.emplace(key_map_iter);
+      auto prev_keys_iter = prev_cf_key_map.find(column_family_id);
+      if (prev_keys_iter == prev_cf_key_map.end()) {
+        prev_cf_key_map.emplace(curr_cf_key_iter);
       } else {
-        auto& prev_tracked_keys_map = prev_tracked_keys->second;
-
-        for (const auto& key_iter : keys) {
+        std::unordered_map<std::string, TransactionKeyMapInfo>& prev_keys =
+            prev_keys_iter->second;
+        for (const auto& key_iter : curr_keys) {
           const std::string& key = key_iter.first;
           const TransactionKeyMapInfo& info = key_iter.second;
-
           // If key was not previously tracked, just copy the whole struct over.
           // Otherwise, some merging needs to occur.
-          auto prev_info = prev_tracked_keys_map.find(key);
-          if (prev_info == prev_tracked_keys_map.end()) {
-            prev_tracked_keys_map.emplace(key_iter);
+          auto prev_info = prev_keys.find(key);
+          if (prev_info == prev_keys.end()) {
+            prev_keys.emplace(key_iter);
           } else {
-            TransactionKeyMapInfo& prev_info_val = prev_info->second;
-
-            assert(prev_info_val.seq <= info.seq);
-            prev_info_val.num_reads += info.num_reads;
-            prev_info_val.num_writes += info.num_writes;
-            prev_info_val.exclusive |= info.exclusive;
+            prev_info->second.Merge(info);
           }
         }
       }
     }
-  } else {
-    save_points_->pop();
   }
 
   return write_batch_.PopSavePoint();
--- a/utilities/transactions/transaction_base.cc
+++ b/utilities/transactions/transaction_base.cc
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
   }
 
   assert(!save_points_->empty());
-  save_points_->pop();
+  // If there is another savepoint A below the current savepoint B, then A needs
+  // to inherit tracked_keys in B so that if we rollback to savepoint A, we
+  // remember to unlock keys in B. If there is no other savepoint below, then we
+  // can safely discard savepoint info.
+  if (save_points_->size() > 1) {
+    TransactionBaseImpl::SavePoint top;
+    std::swap(top, save_points_->top());
+    save_points_->pop();
+
+    const TransactionKeyMap& curr_key_map = top.new_keys_;
+    TransactionKeyMap& prev_key_map = save_points_->top().new_keys_;
+
+    for (const auto& key_map_iter : curr_key_map) {
+      uint32_t column_family_id = key_map_iter.first;
+      auto& keys = key_map_iter.second;
+
+      // If cfid was not previously tracked, just copy everything over.
+      auto prev_tracked_keys = prev_key_map.find(column_family_id);
+      if (prev_tracked_keys == prev_key_map.end()) {
+        prev_key_map.emplace(key_map_iter);
+      } else {
+        auto& prev_tracked_keys_map = prev_tracked_keys->second;
+
+        for (const auto& key_iter : keys) {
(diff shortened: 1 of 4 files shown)


utilities/transactions/transaction_test.cc
```diff
@@ -4030,6 +4030,51 @@ TEST_P(TransactionTest, SavepointTest3) {
...
+  ASSERT_OK(txn1->RollbackToSavePoint());  // Rollback to 1
+
+  // Verify that everything was rolled back.
+  std::string value;
+  s = txn1->Get(read_options, "A", &value);
+  ASSERT_TRUE(s.IsNotFound());
+
+  s = txn1->Get(read_options, "B", &value);
+  ASSERT_TRUE(s.IsNotFound());
+
```
maysamyabandeh: Would the test be more interesting if we commit txn1 and check the content of A and B via txn2 before overwriting them?

utilities/transactions/transaction_base.cc
```diff
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
...
+            assert(prev_info_val.seq <= info.seq);
+            prev_info_val.num_reads += info.num_reads;
+            prev_info_val.num_writes += info.num_writes;
+            prev_info_val.exclusive |= info.exclusive;
+          }
+        }
+      }
+    }
+  } else {
+    save_points_->pop();
```
maysamyabandeh: I find the code easier to follow if the smaller case is on top, i.e., `if (save_points_->size() == 1) {`

utilities/transactions/transaction_test.cc
```diff
@@ -4030,6 +4030,51 @@ TEST_P(TransactionTest, SavepointTest3) {
...
+  // Nothing should be locked
```
maysamyabandeh: Although the keys are not locked, they are still in txn track_keys_. When deleting the transaction, it also unlocks all the tracked keys, which will get us to this line: 
https://github.com/facebook/rocksdb/blob/9625a2bc2b56c92487922c192f4b903083e63c2c/utilities/transactions/transaction_lock_mgr.cc#L620
I am concerned this assert might be invalid if the keys are unlocked via a RollbackToSavePoint.

utilities/transactions/transaction_base.cc
```diff
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
...
+    TransactionBaseImpl::SavePoint top;
+    std::swap(top, save_points_->top());
+    save_points_->pop();
+
+    const TransactionKeyMap& curr_key_map = top.new_keys_;
+    TransactionKeyMap& prev_key_map = save_points_->top().new_keys_;
+
+    for (const auto& key_map_iter : curr_key_map) {
+      uint32_t column_family_id = key_map_iter.first;
+      auto& keys = key_map_iter.second;
```
maysamyabandeh: Since "first" and "second" are not descriptive, it would help readability to replace auto with the explicit type.

utilities/transactions/transaction_base.cc
```diff
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
...
+  // If there is another savepoint A below the current savepoint B, then A needs
+  // to inherit tracked_keys in B so that if we rollback to savepoint A, we
+  // remember to unlock keys in B. If there is no other savepoint below, then we
+  // can safely discard savepoint info.
+  if (save_points_->size() > 1) {
+    TransactionBaseImpl::SavePoint top;
+    std::swap(top, save_points_->top());
+    save_points_->pop();
+
+    const TransactionKeyMap& curr_key_map = top.new_keys_;
```
maysamyabandeh: The names are very similar and thus confusing to me. How about this?
s/curr_key_map/curr_cf_key_map
s/prev_key_map/prev_cf_key_map
s/key_map_iter/curr_cf_key_iter
s/keys/curr_keys
s/prev_tracked_keys/prev_keys   # we know from context that we are talking only about tracked keys
s/prev_tracked_keys_map/prev_keys

utilities/transactions/transaction_base.cc
```diff
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
...
+    for (const auto& key_map_iter : curr_key_map) {
+      uint32_t column_family_id = key_map_iter.first;
+      auto& keys = key_map_iter.second;
+
+      // If cfid was not previously tracked, just copy everything over.
+      auto prev_tracked_keys = prev_key_map.find(column_family_id);
+      if (prev_tracked_keys == prev_key_map.end()) {
+        prev_key_map.emplace(key_map_iter);
+      } else {
+        auto& prev_tracked_keys_map = prev_tracked_keys->second;
```
maysamyabandeh: ditto about auto

utilities/transactions/transaction_base.cc
```diff
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
...
+
+    const TransactionKeyMap& curr_key_map = top.new_keys_;
+    TransactionKeyMap& prev_key_map = save_points_->top().new_keys_;
+
+    for (const auto& key_map_iter : curr_key_map) {
+      uint32_t column_family_id = key_map_iter.first;
+      auto& keys = key_map_iter.second;
+
+      // If cfid was not previously tracked, just copy everything over.
+      auto prev_tracked_keys = prev_key_map.find(column_family_id);
```
maysamyabandeh: I think it would make it clear if auto is replaced with the explicit type here.

utilities/transactions/transaction_base.cc
```diff
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
...
+          const TransactionKeyMapInfo& info = key_iter.second;
+
+          // If key was not previously tracked, just copy the whole struct over.
+          // Otherwise, some merging needs to occur.
+          auto prev_info = prev_tracked_keys_map.find(key);
+          if (prev_info == prev_tracked_keys_map.end()) {
+            prev_tracked_keys_map.emplace(key_iter);
+          } else {
+            TransactionKeyMapInfo& prev_info_val = prev_info->second;
+
```
maysamyabandeh: I am not sure if the empty line here is helpful.

utilities/transactions/transaction_base.cc
```diff
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
...
+            assert(prev_info_val.seq <= info.seq);
```
maysamyabandeh: How about a merge function in TransactionKeyMapInfo struct?

utilities/transactions/transaction_base.cc
```diff
@@ -192,7 +192,53 @@ Status TransactionBaseImpl::PopSavePoint() {
...
+      auto prev_tracked_keys = prev_key_map.find(column_family_id);
+      if (prev_tracked_keys == prev_key_map.end()) {
+        prev_key_map.emplace(key_map_iter);
+      } else {
+        auto& prev_tracked_keys_map = prev_tracked_keys->second;
+
+        for (const auto& key_iter : keys) {
+          const std::string& key = key_iter.first;
+          const TransactionKeyMapInfo& info = key_iter.second;
+
```
maysamyabandeh: I am not sure if the empty line here is helpful.

utilities/transactions/transaction_test.cc
lth: If we determine that the keys were first tracked (ie. locked) in the current savepoint, we actually remove them from `tracked_keys_` so that they don't get unlocked twice:

https://github.com/facebook/rocksdb/blob/9625a2bc2b56c92487922c192f4b903083e63c2c/utilities/transactions/transaction_base.cc#L172

I guess this is kind of interesting for wup if it uses tracked_keys_ now.. It means that on rollback, we won't necessarily have to rollback the entire key set touched, if RollbackToSavepoint has already done the work for us...

utilities/transactions/transaction_test.cc
lth: I don't think we want to commit txn1 because that would release its locks, but we want to test that RollbackToSavepoint releases its locks. I guess we could have txn2 check that A/B are empty here.

utilities/transactions/transaction_test.cc
```diff
@@ -4030,6 +4030,51 @@ TEST_P(TransactionTest, SavepointTest3) {
   ASSERT_TRUE(s.IsNotFound());
 }
 
+TEST_P(TransactionTest, SavepointTest4) {
```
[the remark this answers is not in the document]
lth: Actually I noticed the other tests use the name Savepoint, so I'll just follow the others for now.

utilities/transactions/transaction_test.cc
maysamyabandeh: You are right. I take it back.

utilities/transactions/transaction_base.cc
lth: Since this is a std::pair, I think I'll leave it as auto since the type is quite long. Hopefully the explicit types elsewhere will help the reader.

PR #5630, opened 2019-07-26
added new functions to c-api

abrosimov:
Few functions from options added to C-api

### DIFF
--- a/include/rocksdb/c.h
+++ b/include/rocksdb/c.h
@@ -1796,9 +1796,12 @@ extern ROCKSDB_LIBRARY_API uint64_t
 rocksdb_approximate_memory_usage_get_cache_total(
     rocksdb_memory_usage_t* memory_usage);
 
-extern ROCKSDB_LIBRARY_API void rocksdb_options_set_dump_malloc_stats(rocksdb_options_t*, int);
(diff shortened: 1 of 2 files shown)


include/rocksdb/c.h
```diff
@@ -1765,6 +1771,10 @@ extern ROCKSDB_LIBRARY_API uint64_t
 rocksdb_approximate_memory_usage_get_cache_total(
     rocksdb_memory_usage_t* memory_usage);
 
+extern ROCKSDB_LIBRARY_API void rocksdb_options_set_dump_malloc_stats(rocksdb_options_t*, int);
```
ltamasi: Could you change the type of the second parameter to `unsigned char`? That's what we generally use as a stand-in for `bool` in the C API. Note that this applies to the next method as well.

ltamasi:
Thanks @abrosimov ! Could you please rebase your patch and run the code formatter (`make format`)?

include/rocksdb/c.h
abrosimov: done

abrosimov:
@ltamasi I did it, but there is some problems with CI (travic-ci tests took more than 50 minutes).
I think that these problems doesn't related to changes I've made

PR #5633, opened 2019-07-26
WriteUnPrepared: Add new variable write_batch_flush_threshold

lth:
Instead of reusing `TransactionOptions::max_write_batch_size` for determining when to flush a write batch for write unprepared, add a new variable called `write_batch_flush_threshold` for this use case instead.

Also add `TransactionDBOptions::default_write_batch_flush_threshold` which sets the default value if `TransactionOptions::write_batch_flush_threshold` is unspecified.

### DIFF
--- a/include/rocksdb/utilities/transaction_db.h
+++ b/include/rocksdb/utilities/transaction_db.h
@@ -101,6 +101,11 @@ struct TransactionDBOptions {
   // ordering rather than concurrency control.
   bool skip_concurrency_control = false;
 
+  // This option is only valid for write unprepared. If a write batch exceeds
+  // this threshold, then the transaction will implicitly flush the currently
+  // pending writes into the database. A value of 0 or less means no limit.
+  ssize_t default_write_batch_flush_threshold = 0;
+
  private:
   // 128 entries
   size_t wp_snapshot_cache_bits = static_cast<size_t>(7);
@@ -162,6 +167,11 @@ struct TransactionOptions {
   // back/commit before new transactions start.
   // Default: false
   bool skip_concurrency_control = false;
+
+  // See TransactionDBOptions::default_write_batch_flush_threshold for
+  // description. If a negative value is specified, then the default value from
+  // TransactionDBOptions is used.
+  ssize_t write_batch_flush_threshold = -1;
 };
 
 // The per-write optimizations that do not involve transactions. TransactionDB
--- a/utilities/transactions/write_unprepared_txn.cc
+++ b/utilities/transactions/write_unprepared_txn.cc
@@ -35,13 +35,12 @@ WriteUnpreparedTxn::WriteUnpreparedTxn(WriteUnpreparedTxnDB* txn_db,
       wupt_db_(txn_db),
       recovered_txn_(false),
       largest_validated_seq_(0) {
-  max_write_batch_size_ = txn_options.max_write_batch_size;
-  // We set max bytes to zero so that we don't get a memory limit error.
-  // Instead of trying to keep write batch strictly under the size limit, we
-  // just flush to DB when the limit is exceeded in write unprepared, to avoid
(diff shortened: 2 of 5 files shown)


include/rocksdb/utilities/transaction_db.h
```diff
@@ -101,6 +101,11 @@ struct TransactionDBOptions {
   // ordering rather than concurrency control.
   bool skip_concurrency_control = false;
 
+  // This option is only valid for write unprepared. If a write batch exceeds
+  // this threshold, then the transaction will implicitly flush the currently
+  // pending writes into the database. A value of 0 or less means no limit.
+  ssize_t default_write_batch_flush_threshold = 0;
```
maysamyabandeh: ssize_t is used rarely in the code base. What adv it brings over size_t in this case?

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -71,8 +70,13 @@ WriteUnpreparedTxn::~WriteUnpreparedTxn() {
 
 void WriteUnpreparedTxn::Initialize(const TransactionOptions& txn_options) {
   PessimisticTransaction::Initialize(txn_options);
-  max_write_batch_size_ = txn_options.max_write_batch_size;
-  write_batch_.SetMaxBytes(0);
```
maysamyabandeh: Why is this line deleted?

utilities/transactions/write_unprepared_txn.h
```diff
@@ -164,10 +164,10 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
   Status HandleWrite(std::function<Status()> do_write);
 
   // For write unprepared, we check on every writebatch append to see if
-  // max_write_batch_size_ has been exceeded, and then call
+  // write_batch_flush_threshold has been exceeded, and then call
   // FlushWriteBatchToDB if so. This logic is encapsulated in
   // MaybeFlushWriteBatchToDB.
-  size_t max_write_batch_size_;
+  ssize_t write_batch_flush_threshold_;
```
maysamyabandeh: Why does this have to be ssize_t?

utilities/transactions/write_unprepared_txn.h
```diff
@@ -164,10 +164,10 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
   Status HandleWrite(std::function<Status()> do_write);
 
   // For write unprepared, we check on every writebatch append to see if
-  // max_write_batch_size_ has been exceeded, and then call
+  // write_batch_flush_threshold has been exceeded, and then call
```
maysamyabandeh: Do we need underscore after write_batch_flush_threshold?

include/rocksdb/utilities/transaction_db.h
lth: This is because `TransactionOptions::write_batch_flush_threshold` is signed. A value of -1 there means "unspecified".

If `default_write_batch_flush_threshold` were unsigned, then we would have to be more careful when assigning `default_write_batch_flush_threshold` onto `default_write_batch_flush_threshold`, because the max of size_t is larger than ssize_t.

I guess I could add validation logic in `WriteUnpreparedTxnDB::Initialize` to check that `default_write_batch_flush_threshold` is less than max of ssize_t.

utilities/transactions/write_unprepared_txn.cc
lth: We're setting the max write batch size to infinite here. However, the goal of this change is that we want write_batch_ to have whatever limit is specified in `txn_options.max_write_batch_size`, to have consistent behaviour with other write policies. This happens in `PessimisticTransaction::Initialize`.

utilities/transactions/write_unprepared_txn.h
lth: Same reason as above.

PR #5635, opened 2019-07-26
Reduce the number of random iterations in compact_on_deletion_collector_test

ltamasi:
Summary:
This test frequently times out under TSAN; reducing the number of random
iterations to make it complete faster.

Test Plan:
buck test @mode/dev-tsan internal_repo_rocksdb/repo:compact_on_deletion_collector_test


riversand963:
Is it possible to increase the test time-out instead of reducing number of iterations? If so, I prefer to increase time-out.

ltamasi:
No, unfortunately the timeout is not configurable.

PR #5022, opened 2019-02-27
Refactor trimming logic for immutable memtables

miasantreble:
MyRocks currently sets `max_write_buffer_number_to_maintain` in order to maintain enough history for transaction conflict checking. The effectiveness of this approach depends on the size of memtables. When memtables are small, it may not keep enough history; when memtables are large, this may consume too much memory. 
We are proposing a new way to configure memtable list history: by limiting the memory usage of immutable memtables. The new option is `max_write_buffer_size_to_maintain` and it will take precedence over the old `max_write_buffer_number_to_maintain` if they are both set to non-zero values. The new option accounts for the total memory usage of flushed immutable memtables and mutable memtable. When the total usage exceeds the limit, RocksDB may start dropping immutable memtables (which is also called trimming history), starting from the oldest one. 
The semantics of the old option actually works both as an upper bound and lower bound. History trimming will start if number of immutable memtables exceeds the limit, but it will never go below (limit-1) due to history trimming. 
In order the mimic the behavior with the new option, history trimming will stop if dropping the next immutable memtable causes the total memory usage go below the size limit. For example, assuming the size limit is set to 64MB, and there are 3 immutable memtables with sizes of 20, 30, 30. Although the total memory usage is 80MB > 64MB, dropping the oldest memtable will reduce the memory usage to 60MB < 64MB, so in this case no memtable will be dropped.

PR #5636, opened 2019-07-26
Fix race condition in FlushScheduler

miasantreble:
In normal write pipeline, the memtable write schedule flushes. In the next write group, the leader, which is not concurrent with any memtable write, reads the scheduled flush requests. The concurrency mechanism in flush scheduler says that the scheduling call can be concurrent with each other but cannot be concurrent with the leader that is reading from flush scheduler.
In Pipeline writes this assumption is violated since the memtable writes o the last write group could be run in parallel with the leader of the current group that is reading from the flush scheduler. This creates a race condition.
This PR converts FlushScheduler into a lock-free stack to eliminate the race condition. Also enable enable_pipelined_write in a few tests. 
Test plan: COMPILE_WITH_TSAN=1 make -j32 check


db/flush_scheduler.cc
```diff
@@ -34,36 +34,37 @@ void FlushScheduler::ScheduleFlush(ColumnFamilyData* cfd) {
...
 ColumnFamilyData* FlushScheduler::TakeNextColumnFamily() {
-  while (true) {
-    if (head_.load(std::memory_order_relaxed) == nullptr) {
+  Node* node;
+  do {
+    if ((node = head_.load(std::memory_order_acquire)) == nullptr) {
       return nullptr;
     }
+  // dequeue the head
+  } while (!head_.compare_exchange_strong(
```
shixiao: This implementation is susceptible to the ABA problem.

Although, the original implementation suggests that there's some other synchronization mechanism at play here, which I don't have full context on; perhaps it's even OK to use a lock instead.

miasantreble: Yes you are right, the concurrency situation here is that push (ScheduleFlush) will be called by different threads concurrently, pop (TakeNextColumnFamily) will called sequentially because it's under protection of DB mutex. In other words, there will be N pushes and 1 pop at the same time. Given that, do you think this implementation is thread safe?

yiwu-arbug: One ABA problem I can see is: there are 3 threads T1, T2 and T3:
1. T1 call `ScheduleFlush` and go to sleep after creating new node (line 25)
2. T2 call `TakeNextColumnFamily` and delete the current `head_`.
3. T3 call `ScheduleFlush` to insert a new head, which happen to reuse the same address as the deleted `head_`.
4. T1 resume. The CAS operation success by mistake, since the new and old head node has same address.

yiwu-arbug: I'm wondering what's the performance impact if `FlushSchedule` use mutex instead of lock-free stack. Flush operation should be infrequent enough?

miasantreble: Thanks @yiwu-arbug, @shixiao for input. After discussing with @siying , we feel it's probably best to keep `FlushScheduler` the way it is right now given that PR https://github.com/facebook/rocksdb/pull/5716 eliminated the race condition in pipelined writes. As for `TrimHistoryScheduler` in PR https://github.com/facebook/rocksdb/pull/5022, since it's not on critical path, we can use a slower implementation which uses mutex to simplify the logic.

yiwu-arbug: In that case, shall we leave comment in `FlushScheduler` stating its thread unsafeness?

miasantreble: It's already mentioned in the .h file: https://github.com/facebook/rocksdb/blob/master/db/flush_scheduler.h#L23-L24

yiwu-arbug: ok

PR #5638, opened 2019-07-26
WriteUnPrepared: Use int64_t instead of ssize_t

lth:
The ssize_t type was introduced in https://github.com/facebook/rocksdb/pull/5633, but it seems like it's a POSIX specific type.

I just need a signed type to represent number of bytes, so use int64_t instead. It seems like we have a typedef from SSIZE_T for Windows, but it doesn't seem like we ever include "port/port.h" in our public header files.


miasantreble:
LGTM, please wait for AppVeyor/Travis pass before landing.

PR #5649, opened 2019-07-30
Fix IngestExternalFile overlapping check

jeffrey-xiao:
Previously, the end key of a range deletion tombstone was considered exclusive for the purposes of deletion, but considered inclusive when checking if two SSTables overlap. For example, an SSTable with a range deletion tombstone [a, b) would be considered overlapping with an SSTable with a range deletion tombstone [b, c). This commit fixes this check.

### DIFF
--- a/db/external_sst_file_basic_test.cc
+++ b/db/external_sst_file_basic_test.cc
@@ -835,6 +835,48 @@ TEST_P(ExternalSSTFileBasicTest, IngestionWithRangeDeletions) {
   ASSERT_EQ(2, NumTableFilesAtLevel(options.num_levels - 1));
 }
 
+TEST_F(ExternalSSTFileBasicTest, AdjacentRangeDeletionTombstones) {
+  Options options = CurrentOptions();
+  SstFileWriter sst_file_writer(EnvOptions(), options);
+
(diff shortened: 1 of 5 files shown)


db/external_sst_file_basic_test.cc
```diff
@@ -835,6 +835,46 @@ TEST_P(ExternalSSTFileBasicTest, IngestionWithRangeDeletions) {
   ASSERT_EQ(2, NumTableFilesAtLevel(options.num_levels - 1));
 }
 
+TEST_F(ExternalSSTFileBasicTest, AdjacentRangeDeletionTombstone) {
```
nvanbenschoten: s/AdjacentRangeDeletionTombstone/AdjacentRangeDeletionTombstones/

nvanbenschoten: Also, add a comment somewhere in here about what this is testing and why.

ajkr:
@anand1976 would you be interested in reviewing? Right now multi-file ingestion is unnecessarily rejecting certain inputs as "overlapping", while in fact they are non-overlapping once the exclusiveness of range tombstone end keys is considered.

PR #5658, opened 2019-07-31
WriteUnPrepared: increase test coverage in transaction_test

lth:
The changes transaction_test to set `txn_db_options.default_write_batch_flush_threshold = 1` in order to give better test coverage for WriteUnprepared.

As part of the change, some tests had to be updated.

### DIFF
--- a/utilities/transactions/transaction_test.cc
+++ b/utilities/transactions/transaction_test.cc
@@ -3128,17 +3128,12 @@ TEST_P(TransactionTest, LostUpdate) {
 }
 
 TEST_P(TransactionTest, UntrackedWrites) {
-  // In WriteUnprepared, untracked writes will break snapshot validation logic.
-  // Snapshot validation will only check the largest sequence number of a key to
-  // see if it was committed or not. However, an untracked unprepared write will
-  // hide smaller committed sequence numbers.
-  //
-  // TODO(lth): To fix this, snapshot validation will have to validate all
-  // values larger than snap_seq. If the performance cost is too high, then for
-  // WriteUnprepard, return Status::NotSupported for the untracked calls.
   if (txn_db_options.write_policy == WRITE_UNPREPARED) {
+    // TODO(lth): For WriteUnprepared, validate that untracked writes are
+    // not supported.
     return;
   }
+
   WriteOptions write_options;
   ReadOptions read_options;
   std::string value;
--- a/utilities/transactions/transaction_test.cc
+++ b/utilities/transactions/transaction_test.cc
@@ -2606,7 +2606,6 @@ TEST_P(TransactionTest, ColumnFamiliesTest) {
 
   ASSERT_OK(ReOpenNoDelete(column_families, &handles));
   assert(db != nullptr);
-  ASSERT_OK(s);
 
   Transaction* txn = db->BeginTransaction(write_options);
   ASSERT_TRUE(txn);
@@ -2770,7 +2769,6 @@ TEST_P(TransactionTest, MultiGetBatchedTest) {
   options.merge_operator = MergeOperators::CreateStringAppendOperator();
   ASSERT_OK(ReOpenNoDelete(column_families, &handles));
   assert(db != nullptr);
-  ASSERT_OK(s);
 
   // Write some data to the db
   WriteBatch batch;
@@ -3133,8 +3131,11 @@ TEST_P(TransactionTest, UntrackedWrites) {
   // In WriteUnprepared, untracked writes will break snapshot validation logic.
   // Snapshot validation will only check the largest sequence number of a key to
   // see if it was committed or not. However, an untracked unprepared write will
-  // hide smaller committed sequence numbers. To fix this, snapshot validation
-  // will have to validate all values larger than snap_seq.
+  // hide smaller committed sequence numbers.
+  //
+  // TODO(lth): To fix this, snapshot validation will have to validate all
+  // values larger than snap_seq. If the performance cost is too high, then for
+  // WriteUnprepard, return Status::NotSupported for the untracked calls.
   if (txn_db_options.write_policy == WRITE_UNPREPARED) {
     return;
   }
@@ -3382,7 +3383,6 @@ TEST_P(TransactionTest, LockLimitTest) {
 
   // Open DB with a lock limit of 3
   txn_db_options.max_num_locks = 3;
-  s = TransactionDB::Open(options, txn_db_options, dbname, &db);
   ASSERT_OK(ReOpen());
   assert(db != nullptr);
   ASSERT_OK(s);
@@ -5292,6 +5292,8 @@ TEST_P(TransactionTest, MemoryLimitTest) {
   TransactionOptions txn_options;
   // Header (12 bytes) + NOOP (1 byte) + 2 * 8 bytes for data.
   txn_options.max_write_batch_size = 29;
+  // Set threshold to unlimited so that the write batch does not get flushed,
+  // and can hit the memory limit.
   txn_options.write_batch_flush_threshold = 0;
   std::string value;
   Status s;
--- a/utilities/transactions/transaction_test.cc
+++ b/utilities/transactions/transaction_test.cc
@@ -1727,7 +1727,7 @@ TEST_P(TransactionTest, TwoPhaseLogRollingTest) {
   // our log should be in the heap
   ASSERT_EQ(db_impl->TEST_FindMinLogContainingOutstandingPrep(),
             txn1->GetLogNumber());
-  ASSERT_EQ(db_impl->TEST_LogfileNumber(), txn1->GetLogNumber());
+  ASSERT_EQ(db_impl->TEST_LogfileNumber(), txn1->GetLastLogNumber());
 
   // flush default cf to crate new log
   s = db->Put(wopts, "foo", "bar");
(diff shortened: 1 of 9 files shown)


utilities/transactions/write_unprepared_txn.cc
```diff
@@ -52,10 +53,15 @@ WriteUnpreparedTxn::~WriteUnpreparedTxn() {
     // We should rollback regardless of GetState, but some unit tests that
     // test crash recovery run the destructor assuming that rollback does not
     // happen, so that rollback during recovery can be exercised.
-    if (GetState() == STARTED) {
-      auto s __attribute__((__unused__)) = RollbackInternal();
-      // TODO(lth): Better error handling.
+    if (GetState() == STARTED || GetState() == LOCKS_STOLEN) {
```
maysamyabandeh: Why LOCKS_STOLEN is added here?

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -227,6 +233,7 @@ Status WriteUnpreparedTxn::MaybeFlushWriteBatchToDB() {
   const bool kPrepared = true;
   Status s;
   if (write_batch_flush_threshold_ > 0 &&
+      WriteBatchInternal::Count(write_batch_.GetWriteBatch()) > 0 &&
```
maysamyabandeh: There is also a WriteBatch::Count function.

utilities/transactions/transaction_test.cc
```diff
@@ -3132,6 +3130,14 @@ TEST_P(TransactionTest, LostUpdate) {
 }
 
 TEST_P(TransactionTest, UntrackedWrites) {
+  // In WriteUnprepared, untracked writes will break snapshot validation logic.
+  // Snapshot validation will only check the largest sequence number of a key to
+  // see if it was committed or not. However, an untracked unprepared write will
+  // hide smaller committed sequence numbers. To fix this, snapshot validation
+  // will have to validate all values larger than snap_seq.
```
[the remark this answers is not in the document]
lth: I'll add a TODO here. Alternatively, we could just not support the untracked versions for wup, and return `Status::NotSupported` if we don't want to pay the perf regression.

utilities/transactions/transaction_test.cc
```diff
@@ -3377,6 +3383,7 @@ TEST_P(TransactionTest, LockLimitTest) {
   // Open DB with a lock limit of 3
   txn_db_options.max_num_locks = 3;
   s = TransactionDB::Open(options, txn_db_options, dbname, &db);
+  ASSERT_OK(ReOpen());
```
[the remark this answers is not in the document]
lth: I meant to remove the Open call.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -52,10 +53,15 @@ WriteUnpreparedTxn::~WriteUnpreparedTxn() {
...
     // test crash recovery run the destructor assuming that rollback does not
     // happen, so that rollback during recovery can be exercised.
-    if (GetState() == STARTED) {
-      auto s __attribute__((__unused__)) = RollbackInternal();
-      // TODO(lth): Better error handling.
+    if (GetState() == STARTED || GetState() == LOCKS_STOLEN) {
+      auto s = RollbackInternal();
       assert(s.ok());
+      if (!s.ok()) {
+        ROCKS_LOG_ERROR(
```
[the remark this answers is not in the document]
lth: Yes those are good ideas.

utilities/transactions/write_unprepared_txn.cc
lth: In one of the tests that tests transaction expiration, we were calling `delete` on transactions that have expired. In this case, we do want rollback behaviour, since we're implicitly assuming that expired transactions have rolled back.

utilities/transactions/write_unprepared_txn_db.h
```diff
@@ -34,6 +34,11 @@ class WriteUnpreparedTxnDB : public WritePreparedTxnDB {
 
  private:
   Status RollbackRecoveredTransaction(const DBImpl::RecoveredTransaction* rtxn);
+
+#ifndef NDEBUG
+ public:
```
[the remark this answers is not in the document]
lth: Actually, I just noticed that we have we have private settings in `TransactionDBOptions` (like `wp_snapshot_cache_bits`). Maybe this belongs better there.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -237,7 +244,16 @@ Status WriteUnpreparedTxn::MaybeFlushWriteBatchToDB() {
 
 Status WriteUnpreparedTxn::FlushWriteBatchToDB(bool prepared) {
   if (name_.empty()) {
-    return Status::InvalidArgument("Cannot write to DB without SetName.");
+    assert(!prepared);
+#ifndef NDEBUG
+    // To avoid changing all tests to call SetName, just autogenerate one.
+    if (wupt_db_->autogenerate_name_) {
+      SetName(std::string("xid") + ToString(db_->GetEnv()->NowMicros()));
```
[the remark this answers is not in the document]
lth: I'll use an autoinc number.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -263,15 +263,13 @@ Status WriteUnpreparedTxn::FlushWriteBatchToDB(bool prepared) {
 }
 
 Status WriteUnpreparedTxn::FlushWriteBatchToDBInternal(bool prepared) {
+  static std::atomic_ullong autogen_id{0};
   if (name_.empty()) {
     assert(!prepared);
-#ifndef NDEBUG
```
maysamyabandeh: Since this will be used only in testing, it might be a bad idea to put back NDEBUG here. And also for autogen_id.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -65,6 +65,9 @@ WriteUnpreparedTxn::~WriteUnpreparedTxn() {
             wupt_db_->info_log_,
```
maysamyabandeh: Should it be FATAL error?

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -238,10 +241,7 @@ Status WriteUnpreparedTxn::RebuildFromWriteBatch(WriteBatch* wb) {
 Status WriteUnpreparedTxn::MaybeFlushWriteBatchToDB() {
   const bool kPrepared = true;
   Status s;
-  if (write_batch_flush_threshold_ > 0 &&
-      WriteBatchInternal::Count(write_batch_.GetWriteBatch()) > 0 &&
-      write_batch_.GetDataSize() >
-          static_cast<size_t>(write_batch_flush_threshold_)) {
+  if (write_batch_flush_threshold_ > 0 && write_batch_.GetWriteBatch()->Count() > 0 && write_batch_.GetDataSize() > static_cast<size_t>(write_batch_flush_threshold_)) {
```
maysamyabandeh: Is "make format" called on this line?

utilities/transactions/transaction_test.cc
```diff
@@ -3132,6 +3128,17 @@ TEST_P(TransactionTest, LostUpdate) {
 }
 
 TEST_P(TransactionTest, UntrackedWrites) {
+  // In WriteUnprepared, untracked writes will break snapshot validation logic.
+  // Snapshot validation will only check the largest sequence number of a key to
+  // see if it was committed or not. However, an untracked unprepared write will
+  // hide smaller committed sequence numbers.
+  //
+  // TODO(lth): To fix this, snapshot validation will have to validate all
```
maysamyabandeh: I do not remember if we have documented this TODO somewhere else or not but if WrteUnprepared is imposing a new limitation it should be documented in the main code base, rather than in the test files. If we have such place in the main code base, the TODO might be better off there.

lth: I don't think there's a convenient place for me to put this, so I'll just implement my TODO.

lth: Put it in the header file

utilities/transactions/write_unprepared_txn.h
```diff
@@ -145,6 +145,15 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
...
 
+  // In WriteUnprepared, untracked writes will break snapshot validation logic.
+  // Snapshot validation will only check the largest sequence number of a key to
+  // see if it was committed or not. However, an untracked unprepared write will
+  // hide smaller committed sequence numbers.
+  //
+  // TODO(lth): Investigate whether it is worth having snapshot validation
+  // validate all values larger than snap_seq. Otherwise, we should return
+  // Status::NotSupported for untracked writes.
+
```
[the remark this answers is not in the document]
lth: It doesn't really go with the next line (`RebuildFromWriteBatch`). I just wanted a comment in the .h file.

utilities/transactions/write_unprepared_transaction_test.cc
```diff
@@ -335,7 +335,7 @@ TEST_P(WriteUnpreparedTransactionTest, RecoveryTest) {
           for (int i = 0; i < num_batches; i++) {
             ASSERT_OK(txn->Put("k" + ToString(i), "value" + ToString(i)));
             if (txn_options.write_batch_flush_threshold == 1) {
-              ASSERT_EQ(wup_txn->GetUnpreparedSequenceNumbers().size(), i + 1);
```
maysamyabandeh: This change needs some comments. Otherwise when the test fails, it would be difficult to argue about the correct value expected from GetUnpreparedSequenceNumbers.

PR #5662, opened 2019-08-01
cmake: cmake related cleanups

tchaikov:
- cmake: use the builtin FindBzip2.cmake from CMake 
- cmake: require CMake v3.5.1
- cmake: add imported target for 3rd party libraries
- cmake: extract ReadVersion.cmake out and refactor it


maysamyabandeh:
@hliu18 @burtonli Would appreciate it if you could verify the suggested changes.

hliu18:
Looks fine to me.

CMakeLists.txt
```diff
@@ -32,11 +32,10 @@
...
 # 4. make -j
 
-cmake_minimum_required(VERSION 2.8.12)
-project(rocksdb)
-enable_language(CXX)
-enable_language(C)
-enable_language(ASM)
+cmake_minimum_required(VERSION 3.5.1)
+project(rocksdb
+  VERSION 6.2.0
```
yiwu-arbug: is it possible to get version from version.h?

hliu18: it's better to not duplicate in two places. Is "VERSION 6.2.0" needed here?

cmake/modules/Findlz4.cmake
```diff
@@ -1,21 +1,29 @@
 # - Find Lz4
 # Find the lz4 compression library and includes
 #
-# LZ4_INCLUDE_DIR - where to find lz4.h, etc.
-# LZ4_LIBRARIES - List of libraries when using lz4.
-# LZ4_FOUND - True if lz4 found.
+# lz4_INCLUDE_DIR - where to find lz4.h, etc.
```
burtonli: BZIP2_INCLUDE_DIR**S** ?

CMakeLists.txt
```diff
@@ -94,17 +93,16 @@ else()
 
   option(WITH_BZ2 "build with bzip2" OFF)
   if(WITH_BZ2)
-    find_package(bzip2 REQUIRED)
+    find_package(BZip2 REQUIRED)
     add_definitions(-DBZIP2)
-    include_directories(${BZIP2_INCLUDE_DIR})
+    include_directories(${BZIP2_INCLUDE_DIRS})
```
burtonli: should we also update Findbzip2.cmake accordingly?

tchaikov: `Findbzip2.cmake` is removed in the same commit.  please refer the commit message for more details.

CMakeLists.txt
tchaikov: agreed. i extracted `ReadVersion.cmake` out, and use it instead.

cmake/modules/Findlz4.cmake
tchaikov: thanks. fixed.

tchaikov:
changelog

- s/lz4_INCLUDE_DIR/lz4_INCLUDE_DIRS/
- add imported target for more `Find*.cmake`. and use it for `VERSION` in `project()` command
- extract `ReadVersion.cmake` out

maysamyabandeh:
@tchaikov could you please update the PR description with a summary of the changes.

maysamyabandeh:
@burtonli @yiwu-arbug  Thanks for the review. In the PR good for landing?

tchaikov:
@maysamyabandeh sure. updated.

burtonli:
LGTM

yiwu-arbug:
LGTM

CMakeLists.txt
```diff
@@ -93,43 +93,38 @@ else()
   if(WITH_SNAPPY)
     find_package(snappy REQUIRED)
     add_definitions(-DSNAPPY)
-    include_directories(${SNAPPY_INCLUDE_DIR})
```
merryChris: @tchaikov Why do you remove this line? Seems that cannot find it in include path. Errors like 
`/root/cs/suzanwen/repos/C++/storage/rocksdb_xxx/rocksdb/util/compression.h:31:20: fatal error: snappy.h: No such file or directory
 #include <snappy.h>
                    ^
compilation terminated.`

tchaikov: because, IIUC, snappy::snappy should be able to populate the "INSTALL_INTERFACE" property to the library/executable linked against it, so cmake will determine the correct build property of consumer of snappy::snappy and library publicly linked against this library.

https://github.com/facebook/rocksdb/blob/f65ec09ef86c8dffd20aa63a12126a23d6d3c5b4/cmake/modules/Findsnappy.cmake#L28

merryChris: Yeah, that's right for the most targets, such as `ROCKSDB_STATIC_LIB`.  But not for `TESTUTILLIB`. Plz have a glance at the following snippet.
https://github.com/facebook/rocksdb/blob/f65ec09ef86c8dffd20aa63a12126a23d6d3c5b4/CMakeLists.txt#L1072

`snappy.h` was included in `db/db_test_util.cc`(just as the source file in target `TESTUTILLIB`). Porbably we need make it works fine with `target_link_libraries(${TESTUTILLIB} ${LIBS})`.

merryChris: @tchaikov https://github.com/facebook/rocksdb/pull/6043 pr as it is, plz make a review.

wbrown-lg:
As noted in an inline comment, this breaks the `rocksdb` build if we do a `CMake` `add_subdirectory` on `rocksdb`.

wbrown-lg:
As noted in an inline comment, this breaks the `rocksdb` build if we do a `CMake` `add_subdirectory` on `rocksdb`.

tchaikov:
@wbrown-lg thanks for pointing out this. Will fix it shortly.

tchaikov:
@wbrown-lg this was fixed by adcf920f407f0d7b78b71348017e13632fdb95e9 . could you use a commit which includes that change instead?

PR #5665, opened 2019-08-02
Fix MultiGet() bug when whole_key_filtering is disabled

anand1976:
The batched MultiGet() implementation was not correctly handling bloom filter lookups when whole_key_filtering is disabled. It was incorrectly skipping keys not in the prefix_extractor domain, and not calling transform for keys in domain. This PR fixes both problems by moving the domain check and transformation to the FilterBlockReader.

Tests:
Unit test (confirmed failed before the fix)
make check

### DIFF
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -1287,10 +1287,21 @@ TEST_F(DBBasicTest, MultiGetBatchedMultiLevel) {
   }
 }
 
-TEST_F(DBBasicTest, MultiGetPrefixExtractor) {
+// Test class for batched MultiGet with prefix extractor
+// Param bool - If true, use partitioned filters
+//              If false, use full filter block
+class MultiGetPrefixExtractorTest
+    : public DBBasicTest,
+      public ::testing::WithParamInterface<bool> {};
+
+TEST_P(MultiGetPrefixExtractorTest, Batched) {
   Options options = CurrentOptions();
   options.prefix_extractor.reset(NewFixedPrefixTransform(2));
   BlockBasedTableOptions bbto;
+  if (GetParam()) {
+    bbto.index_type = BlockBasedTableOptions::IndexType::kTwoLevelIndexSearch;
+    bbto.partition_filters = true;
+  }
   bbto.filter_policy.reset(NewBloomFilterPolicy(10, false));
   bbto.whole_key_filtering = false;
   bbto.cache_index_and_filter_blocks = false;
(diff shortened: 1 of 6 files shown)


db/db_basic_test.cc
```diff
@@ -1287,6 +1287,45 @@ TEST_F(DBBasicTest, MultiGetBatchedMultiLevel) {
...
+
+  db_->MultiGet(ReadOptions(), dbfull()->DefaultColumnFamily(), keys.size(),
+                keys.data(), values.data(), s.data(), true);
+
+  ASSERT_EQ(values.size(), keys.size());
+  ASSERT_EQ(std::string(values[0].data(), values[0].size()), "v0");
+  ASSERT_EQ(std::string(values[1].data(), values[1].size()), "v1");
+  ASSERT_EQ(std::string(values[2].data(), values[2].size()), "v2");
+  ASSERT_EQ(std::string(values[3].data(), values[3].size()), "v3");
+  ASSERT_EQ(std::string(values[4].data(), values[4].size()), "v4");
```
miasantreble: would it be useful to add verification using stats or perf_context like bloom filter hits or misses?  so that we can know for sure that "It was incorrectly skipping keys not in the prefix_extractor domain, and not calling transform for keys in domain." is no longer the case?

anand1976: I structured the KVs such that if whole key filter lookup is used, the lookup will fail. But yes, it would be useful to add stats verification as it'll make it more robust to collisions etc.

anand1976: On second thoughts, it may not be that useful, since it won't tell us whether the bloom filter hit was due to prefix lookup or not.

table/block_based/full_filter_block.cc
```diff
@@ -254,15 +255,28 @@ void FullFilterBlockReader::MayMatch(
   // expensive compared to autovector
   Slice* keys[MultiGetContext::MAX_BATCH_SIZE];
   bool may_match[MultiGetContext::MAX_BATCH_SIZE] = {false};
```
ltamasi: Shouldn't these may_match flags default to true? I.e. if a key is not in the prefix extractor's domain, we should treat it as potentially present, no? (Edit: this probably does not make a practical difference though since the code only seems to use the first num_keys elements of this array from this point on, and those get set by FullFilterBitsReader. We might still consider changing it for the sake of consistency/clarity.)

table/block_based/filter_block.h
```diff
@@ -137,7 +137,8 @@ class FilterBlockReader {
       const Slice ukey = iter->ukey;
       const Slice ikey = iter->ikey;
       GetContext* const get_context = iter->get_context;
-      if (!KeyMayMatch(prefix_extractor->Transform(ukey), prefix_extractor,
+      if (prefix_extractor->InDomain(ukey) &&
+          !KeyMayMatch(prefix_extractor->Transform(ukey), prefix_extractor,
```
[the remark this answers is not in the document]
anand1976: Yes, the call to ```KeyMayMatch``` should be ```PrefixMayMatch``` instead. I'll change it. As a practical matter, it may not make much difference. Also, the ```InDomain``` check is still required as ```PrefixMayMatch``` implementation does not check it.

table/block_based/full_filter_block.cc
anand1976: Yes, makes sense to be conservative.

db/db_basic_test.cc
anand1976: Turns out it was useful after all, as I wasn't updating the perf_context stats !

db/db_basic_test.cc
```diff
@@ -1287,6 +1287,48 @@ TEST_F(DBBasicTest, MultiGetBatchedMultiLevel) {
   }
 }
 
+TEST_F(DBBasicTest, MultiGetPrefixExtractor) {
```
ltamasi: P.S. As it stands right now, we're only testing FullFilterBlockReader here; we could extend this so it also tests PartitionedFilterBlockReader (which I believe uses the base class implementation of PrefixesMayMatch).

PR #5672, opened 2019-08-05
Fix make target 'all' and 'check'

riversand963:
If a test is one of parallel tests, then it should also be one of the 'tests'.
Otherwise, `make all` won't build the binaries. For examle,
```
$COMPILE_WITH_ASAN=1 make -j32 all
```
Then if you do
```
$make check
```
The second command will invoke the compilation and building for db_bloom_test
and file_reader_writer_test **without** the `COMPILE_WITH_ASAN=1`, causing the
command to fail.

Test plan (on devserver):
```
$make -j32 all
```
Verify all binaries are built so that `make check` won't have to compile any
thing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -445,6 +445,7 @@ TESTS = \
 	db_iter_test \
 	db_iter_stress_test \
 	db_log_iter_test \
+	db_bloom_filter_test \
 	db_compaction_filter_test \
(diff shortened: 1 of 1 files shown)


riversand963:
cc @anand1976 and @ltamasi regarding #5608 and #5632

PR #5679, opened 2019-08-07
Block cache analyzer: Support reading from human readable trace file.

HaoyuHuang:
This PR adds support in block cache trace analyzer to read from human readable trace file. This is needed when a user does not have access to the binary trace file. 

Test plan:
USE_CLANG=1 make check -j32


trace_replay/block_cache_tracer.cc
```diff
@@ -300,6 +304,139 @@ Status BlockCacheTraceReader::ReadAccess(BlockCacheTraceRecord* record) {
...
+  }
+  std::stringstream ss(line);
+  std::vector<std::string> record_strs;
+  while (ss.good()) {
+    std::string substr;
+    getline(ss, substr, ',');
+    record_strs.push_back(substr);
+  }
+  if (record_strs.size() != 21) {
+    return Status::Incomplete("Records length is wrong.");
```
riversand963: Length or format?

PR #5694, opened 2019-08-13
arm64 crc optimise

HouBingjian:
here I do two optimise:
1，modify CMakeList to pass "+crypto" flag if  armv8-a+crc support
2，prefetch data for following block to avoid cache miss when doing crc caculate

I do performance test at huawei kunpeng-920 server(arm-v8, 64core@2.6GHz)
./db_bench  --benchmarks=crc32c --block_size=500000000
before optimise :   678515.000 micros/op 1 ops/sec;  702.8 MB/s (500000000 per op)
after    optimise :   206497.000 micros/op 4 ops/sec; 2309.1 MB/s (500000000 per op)


HouBingjian:
@facebook-github-bot     can assign to some reviewer ?

PR #5696, opened 2019-08-13
use c++17's try_emplace if available

jsteemann:
This avoids rehashing the key in TrackKey() in case the key is not already
in the map of tracked keys, which will happen at least once per key used in a
transaction.

Additionally fix two typos.

### DIFF
--- a/utilities/transactions/transaction_base.cc
+++ b/utilities/transactions/transaction_base.cc
@@ -632,6 +632,17 @@ void TransactionBaseImpl::TrackKey(TransactionKeyMap* key_map, uint32_t cfh_id,
                                    const std::string& key, SequenceNumber seq,
                                    bool read_only, bool exclusive) {
   auto& cf_key_map = (*key_map)[cfh_id];
+#ifdef __cpp_lib_unordered_map_try_emplace
+  // use c++17's try_emplace if available, to avoid rehashing the key
+  // in case it is not already in the map
+  auto result = cf_key_map.try_emplace(key, TransactionKeyMapInfo(seq));
+  auto iter = result.first;
+  if (!result.second && 
+      seq < iter->second.seq) {
+    // Now tracking this key with an earlier sequence number
+    iter->second.seq = seq;
+  }
+#else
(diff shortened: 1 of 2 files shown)


utilities/transactions/transaction_base.cc
```diff
@@ -632,6 +632,17 @@ void TransactionBaseImpl::TrackKey(TransactionKeyMap* key_map, uint32_t cfh_id,
                                    const std::string& key, SequenceNumber seq,
                                    bool read_only, bool exclusive) {
   auto& cf_key_map = (*key_map)[cfh_id];
+#ifdef __cpp_lib_unordered_map_try_emplace
+  // use c++17's try_emplace if available, to avoid rehashing the key
+  // in case it is not already in the map
+  auto result = cf_key_map.try_emplace(key, TransactionKeyMapInfo(seq));
```
lth: Actually, can't we just use regular `emplace` in this case?

From https://en.cppreference.com/w/cpp/container/map/try_emplace:

> Unlike insert or emplace, these functions do not move from rvalue arguments if the insertion does not happen...

In this case though, key is not an rvalue reference, and the value is just a temporary, so we don't care if it got moved or not, so we're okay if that's left in an unspecified state.

jsteemann: IMHO emplace will always perform a dynamic memory allocation, regardless of whether the key is already in the map or not.
try_emplace however only allocates memory if the key isnt yet contained in the map.
So if you expect lots of duplicates, emplace may actually perform worse than a combination of find and insert/emplace.
try_emplace seems to be a good general solution, in case there are many duplicates and also in case there arent.

lth: Makes sense.

I think you can avoid constructing `TransactionKeyMapInfo` in the failure case if you just pass in seq.

jsteemann:
@lth: oh yes, you are right. I think one can avoid the construction of `TransactionKeyMapInfo` in case it is not required. I just pushed that small change into the PR as well. Thanks for pointing this out, I overlooked it!

PR #5697, opened 2019-08-13
WriteUnPrepared: Relax restriction on iterators and writes with no snapshot

lth:
Currently, if a write is done without a snapshot, then `largest_validated_seq_` is set to `kMaxSequenceNumber`. This is too aggressive, because an iterator with a snapshot created after this write should be valid.

Set `largest_validated_seq_` to `GetLastPublishedSequence` instead. The variable means that no keys in the current tracked key set has changed by other transactions since `largest_validated_seq_`.

Also, do some extra cleanup in Clear() for safety.

### DIFF
--- a/utilities/transactions/write_unprepared_txn.cc
+++ b/utilities/transactions/write_unprepared_txn.cc
@@ -106,7 +106,10 @@ Status WriteUnpreparedTxn::HandleWrite(std::function<Status()> do_write) {
       largest_validated_seq_ =
           std::max(largest_validated_seq_, snapshot_->GetSequenceNumber());
     } else {
-      largest_validated_seq_ = kMaxSequenceNumber;
+      // TODO(lth): We should use the same number as tracked_at_seq in TryLock,
+      // because what is actually being tracked is the sequence number at which
+      // this key was locked at.
+      largest_validated_seq_ = db_impl_->GetLastPublishedSequence();
     }
   }
   return s;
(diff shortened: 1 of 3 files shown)


utilities/transactions/write_unprepared_txn.cc
```diff
@@ -680,6 +683,11 @@ void WriteUnpreparedTxn::Clear() {
   if (!recovered_txn_) {
     txn_db_impl_->UnLock(this, &GetTrackedKeys());
   }
+  unprep_seqs_.clear();
```
maysamyabandeh: Can we mention this change in the PR description?

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -106,7 +106,10 @@ Status WriteUnpreparedTxn::HandleWrite(std::function<Status()> do_write) {
       largest_validated_seq_ =
           std::max(largest_validated_seq_, snapshot_->GetSequenceNumber());
     } else {
-      largest_validated_seq_ = kMaxSequenceNumber;
+      // TODO(lth): We should use the same number as tracked_at_seq in TryLock,
+      // because what is actually being tracked is the sequence number at which
+      // this key was locked at.
+      largest_validated_seq_ = db_impl_->GetLastPublishedSequence();
```
maysamyabandeh: Shall we also update the variable inline comment in https://github.com/facebook/rocksdb/blob/04a849b7b408e05d55557a16d1d677b80c543b30/utilities/transactions/write_unprepared_txn.h#L244

PR #5698, opened 2019-08-13
Fix data races in BlobDB

ltamasi:
Summary:
Some accesses to blob_files_ and open_ttl_files_ in BlobDBImpl, as well
as to expiration_range_ in BlobFile were not properly synchronized.
The patch fixes this and also makes sure the invariant that obsolete_files_
is a subset of blob_files_ holds even when an attempt to delete an obsolete
blob file fails.

Test Plan:
COMPILE_WITH_TSAN=1 make blob_db_test
gtest-parallel --repeat=1000 ./blob_db_test

The test fails with TSAN errors ~20 times out of 1000 in the test case ShutdownWait
without the patch but completes successfully 1000 out of 1000 times with the fix.

### DIFF
--- a/utilities/blob_db/blob_db_impl.cc
+++ b/utilities/blob_db/blob_db_impl.cc
@@ -724,6 +724,7 @@ Status BlobDBImpl::PutBlobValue(const WriteOptions& /*options*/,
     }
     if (s.ok()) {
       if (expiration != kNoExpiration) {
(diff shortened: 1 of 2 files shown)


ltamasi:
Actually it seems like BlobFile's mutex_ is meant to be locked by the client (i.e. BlobDBImpl, which is a friend); will rework that part.

PR #5703, opened 2019-08-14
WriteUnPrepared: Fix bug in savepoints

lth:
Fix a bug in write unprepared savepoints. When flushing the write batch according to savepoint boundaries, we were forgetting to flush the last write batch after the last savepoint, meaning that some data was not written to DB.

Also, add a small optimization where we avoid flushing empty batches.

### DIFF
--- a/utilities/transactions/write_unprepared_txn.cc
+++ b/utilities/transactions/write_unprepared_txn.cc
@@ -407,10 +407,13 @@ Status WriteUnpreparedTxn::FlushWriteBatchWithSavePointToDB() {
 
   size_t prev_boundary = WriteBatchInternal::kHeader;
   const bool kPrepared = true;
-  for (size_t i = 0; i < unflushed_save_points_->size(); i++) {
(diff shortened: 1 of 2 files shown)


utilities/transactions/write_unprepared_txn.cc
```diff
@@ -729,7 +736,6 @@ Status WriteUnpreparedTxn::RollbackToSavePointInternal() {
   assert(flushed_save_points_->size() > 0);
   WriteUnpreparedTxn::SavePoint& top = flushed_save_points_->back();
 
-  assert(top.unprep_seqs_.size() > 0);
```
maysamyabandeh: Why is this assert removed?

lth: This is removed because now we may skip flushing a write batch, if that write batch is empty, but we still want to save the savepoint.

If we set savepoint right after begin transaction, then when we flush the write batch, the first savepoint would not have any unprep_seqs because no flush was done.

PR #5706, opened 2019-08-14
Support row cache with batched MultiGet

anand1976:
This PR adds support for row cache in ```rocksdb::TableCache::MultiGet```. 

Test plan:
1. Unit tests in db_basic_test
2. db_bench results with batch size of 2 (```Get``` is faster than ```MultiGet``` for single key) -
Get -
readrandom   :       3.935 micros/op 254116 ops/sec;   28.1 MB/s (22870998 of 22870999 found)
MultiGet -
multireadrandom :       3.743 micros/op 267190 ops/sec; (24047998 of 24047998 found)

Command used -
TEST_TMPDIR=/dev/shm/multiget numactl -C 10  ./db_bench -use_existing_db=true -use_existing_keys=false -benchmarks="readtorowcache,[read|multiread]random" -write_buffer_size=16777216 -target_file_size_base=4194304 -max_bytes_for_level_base=16777216 -num=12000000 -reads=12000000 -duration=90 -threads=1 -compression_type=none -cache_size=4194304000 -row_cache_size=4194304000 -batch_size=2 -disable_auto_compactions=true -bloom_bits=10 -cache_index_and_filter_blocks=true -pin_l0_filter_and_index_blocks_in_cache=true -multiread_batched=true -multiread_stride=131072

### DIFF
--- a/db/table_cache.cc
+++ b/db/table_cache.cc
@@ -426,7 +426,91 @@ Status TableCache::MultiGet(const ReadOptions& options,
   Status s;
   TableReader* t = fd.table_reader;
   Cache::Handle* handle = nullptr;
-  if (s.ok()) {
+  MultiGetRange table_range(*mget_range, mget_range->begin(), mget_range->end());
+#ifndef ROCKSDB_LITE
+  autovector<std::string, MultiGetContext::MAX_BATCH_SIZE> row_cache_entries;
+  IterKey row_cache_key;
+  size_t row_cache_key_prefix = 0;
+  KeyContext& first_key = *table_range.begin();
+  bool lookup_row_cache = ioptions_.row_cache &&
+          !first_key.get_context->NeedToReadSequence();
+
+  // Check row cache if enabled. Since row cache does not currently store
+  // sequence numbers, we cannot use it if we need to fetch the sequence.
+  if (lookup_row_cache) {
+    uint64_t fd_number = fd.GetNumber();
+    // We use the user key as cache key instead of the internal key,
+    // otherwise the whole cache would be invalidated every time the
+    // sequence key increases. However, to support caching snapshot
+    // reads, we append the sequence number (incremented by 1 to
+    // distinguish from 0) only in this case.
+    // We only need to look at the sequence number of the first key in the
+    // batch since all keys in the MultiGet batch would be read from the
+    // same snapshot.
+    uint64_t seq_no = 0;
+    if (options.snapshot != nullptr &&
+        (first_key.get_context->has_callback() ||
+         static_cast_with_check<const SnapshotImpl, const Snapshot>(
+             options.snapshot)
+                 ->GetSequenceNumber() <= fd.largest_seqno)) {
+      // We should consider to use options.snapshot->GetSequenceNumber()
+      // instead of GetInternalKeySeqno(k), which will make the code
+      // easier to understand.
+      seq_no = 1 + GetInternalKeySeqno(first_key.ikey);
+    }
+
+    // Prepare the common row cache key prefix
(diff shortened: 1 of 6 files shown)


db/table_cache.cc
```diff
@@ -404,7 +404,91 @@ Status TableCache::MultiGet(const ReadOptions& options,
...
   Cache::Handle* handle = nullptr;
-  if (s.ok()) {
+  MultiGetRange table_range(*mget_range, mget_range->begin(), mget_range->end());
+#ifndef ROCKSDB_LITE
+  autovector<std::string, MultiGetContext::MAX_BATCH_SIZE> row_cache_entries;
+  IterKey row_cache_key;
+  size_t row_cache_key_prefix = 0;
+  KeyContext& first_key = *table_range.begin();
+  bool lookup_row_cache = ioptions_.row_cache &&
+          !first_key.get_context->NeedToReadSequence();
```
ltamasi: Is NeedToReadSequence() guaranteed to return the same result for all GetContexts? Same question for has_callback() below.

db/table_cache.cc
```diff
@@ -404,7 +404,91 @@ Status TableCache::MultiGet(const ReadOptions& options,
   Status s;
   TableReader* t = fd.table_reader;
   Cache::Handle* handle = nullptr;
-  if (s.ok()) {
+  MultiGetRange table_range(*mget_range, mget_range->begin(), mget_range->end());
+#ifndef ROCKSDB_LITE
+  autovector<std::string, MultiGetContext::MAX_BATCH_SIZE> row_cache_entries;
```
ltamasi: We could consider factoring out this logic into separate helper method(s) to eliminate the code duplication between Get() and MultiGet().

db/table_cache.cc
```diff
@@ -444,6 +528,31 @@ Status TableCache::MultiGet(const ReadOptions& options,
     }
   }
 
+#ifndef ROCKSDB_LITE
+  if (lookup_row_cache) {
+    size_t row_idx = 0;
```
ltamasi: I think this could also be factored out into a helper method and reused in Get/MultiGet.

db/version_set.cc
```diff
@@ -1812,7 +1812,7 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
         user_comparator(), merge_operator_, info_log_, db_statistics_,
         iter->s->ok() ? GetContext::kNotFound : GetContext::kMerge, iter->ukey,
         iter->value, nullptr, &(iter->merge_context), true,
-        &iter->max_covering_tombstone_seq, this->env_, &iter->seq,
+        &iter->max_covering_tombstone_seq, this->env_, nullptr,
```
ltamasi: Does this mean we no longer support retrieving the sequence numbers for MultiGet?

db/table_cache.cc
anand1976: Yes. At present, getting internal sequence number is not supported in MultiGet(), so NeedToReadSequence() will always return false. We may have to revisit in future when we support ```MultiGetForUpdate()``` in TransactionDB. The has_callback() is also guaranteed to return the same value for all keys, since a single callback is passed to ```DBImpl::MultiGet()```.

db/table_cache.cc
anand1976: Let me see how this can be refactored. Need to be careful not to add overhead to Get() due to bookkeeping overhead of working with a batch.

db/version_set.cc
anand1976: It was never actually supported. As mentioned in a previous comment, we may have to revisit when we support MultiGetForUpdate().

PR #5707, opened 2019-08-15
Add a db_bench benchmark to warm up the row cache

anand1976:


tools/db_bench_tool.cc
```diff
@@ -4605,6 +4614,65 @@ class Benchmark {
     }
   }
 
+  void ReadToRowCache(ThreadState* thread) {
+    int64_t read = 0;
+    int64_t found = 0;
+    int64_t bytes = 0;
+    int64_t key_rand = 0;
```
ltamasi: It's not really random, is it? :) Otherwise LGTM, thanks!

PR #5726, opened 2019-08-21
Refactor MultiGet names in BlockBasedTable

maysamyabandeh:
To improve code readability, since RetrieveBlock already calls MaybeReadBlockAndLoadToCache, we avoid name similarity of the functions that call RetrieveBlock with MaybeReadBlockAndLoadToCache. The patch thus renames MaybeLoadBlocksToCache to RetrieveMultipleBlocks and deletes GetDataBlockFromCache, which contains only two lines.

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -2254,7 +2254,7 @@ Status BlockBasedTable::MaybeReadBlockAndLoadToCache(
 //         found in cache
 // handles - A vector of block handles. Some of them me be NULL handles
 // scratch - An optional contiguous buffer to read compressed blocks into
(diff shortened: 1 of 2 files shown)


table/block_based/block_based_table_reader.cc
```diff
@@ -2275,7 +2254,7 @@ Status BlockBasedTable::MaybeReadBlockAndLoadToCache(
 //         found in cache
 // handles - A vector of block handles. Some of them me be NULL handles
 // scratch - An optional contiguous buffer to read compressed blocks into
-void BlockBasedTable::MaybeLoadBlocksToCache(
+void BlockBasedTable::RetrieveMultipleBlock(
```
anand1976: Minor nit - call it RetrieveMultipleBlocks().

PR #5752, opened 2019-08-28
Use the Configurable class to Configure ColumnFamilyOptions

mrambacher:
This  PR uses the Configurable class as a basis for configuring the ColumnFamilyOptions, replacing the old way of doing so.


riversand963:
I guess this PR can also be closed due to #5753?

PR #5497, opened 2019-06-21
Configurable DBOptions

mrambacher:
This PR uses the new Configurable class to convert the DBOptions (and MutableDBOptions) to/fro Strings.

PR #5362, opened 2019-05-25
Make the TableFactory into a Configurable object

mrambacher:
This change demonstrates how to use the Configurable class by changing the TableFactory to be an instance of this class.  This change allows many of the methods that were written to configure a TableFactory to be removed, as the Configurable class itself implements this functionality.

Additionally, the CuckooTableFactory can now be configured via properties (it still cannot be created from an INI file -- stay tuned for that change!)

PR #5361, opened 2019-05-25
Create a Configurable base class for use in get/set values from properties and INI files

mrambacher:
This change adds a new Configurable class.  A Configurable object can:
- Be populated by ConfigureFromString or ConfigureFromMap
- Have individual fields updated, via ConfigureOption
- Convert itself into a string suitable for storing in initialization files (GetOptionsString)
- Dump its properties to a Logger
- See if it matches another Configurable (Matches)
- Check/Change its settings if they are not proper (SanitizeOptions)
- Verify that it settings are valid (VerifyOptions)
- Return a pointer to the underlying options object (GetOptions)

The Configurable class uses an OptionsTypeMap (map<string, OptionTypeInfo) in order to implement this functionality.  

The code for this class is based on the code for configuring a TableFactory instance.  A follow-on PR will show this code being used by the TableFactory class.

Any class can be a Configurable simply by inheritance.  By registering an OptionsTypeMap with the base implementation, the object will automatically gain the Configurable functionality.  If no map is registered, the class will behave as it does today (the methods will do nothing).

PR #5753, opened 2019-08-28
Bring the Configurable options together

mrambacher:
This PR merges the functionality of making the ColumnFamilyOptions, TableFactory, and DBOptions into Configurable into a single PR, resolving any merge conflicts


riversand963:
Hi @mrambacher , starting to look at the configurable PRs. I noticed that https://github.com/facebook/rocksdb/pull/5753 seems to be a combination of https://github.com/facebook/rocksdb/pull/5362, https://github.com/facebook/rocksdb/pull/5497, https://github.com/facebook/rocksdb/pull/5752/ and https://github.com/facebook/rocksdb/pull/5574/files
Should we be looking at just #5753 and close the others?

mrambacher:
#5574 is the base PR, which adds more functionality to the OptionTypeInfo class.  #5361  (Configurable class) uses #5574 to implement a base configurable class.  

#5362, #5497, and #5752 use the Configurable class to configure their respective pieces whereas #5753 merges those respective pieces into one chunk.    

I would recommend reviewing #5574 then #5361.  If you are comfortable with the classes, #5753 is what you really want and not the other three PRs (as they are subsumed eventually by this one anway).

mrambacher:
Just to clarify, those three PRs (#5574, #5361, and #5753) build on each other.  You could start with just the last one and that would hit all of the functionality.  They were broken up into the littler pieces in an attempt to make it easier to review, but it is your choice how you want to tackle them.

riversand963:
Thanks @mrambacher for the clarification. I will start with #5753. In case I need more context, I can use the preceding PRs as a reference to help understand.

riversand963:
I am looking at your PR, and there is one thing I am not following. In Configurable::SetOptions, we already know the option name, why can't we use the name to do a point lookup in options_? Currently we iterate over all entries options_ until we find a match in the option map associated with that entry. Could you point me to an example to show how this works? Thanks

riversand963:
Chatted offline. Thanks @mrambacher for clarifying some aspects of his design.

With recursion put aside, each `Configurable` object has a two-level mapping. The first level maps named parts of the object to a corresponding memory address plus another map. This second map is the second layer mapping from named fields to metadata, e.g. `OptionTypeInfo`, of the field's type. An object has one entry in the `options_` map (first level mapping) iff. `RegisterOptionsMap()` is called once. It is possible for an object to have more entries in `options_` iff. `RegisterOptionsMap()` multiple times. This can happen in multiple cases. One case is object B *has* object A, but the user does not specify a unified `option_type_map` for B. Another case is that object B is a subclass of A with some extra configurable members, and the user does not specify a unified `options_type_map` for B.
I also notice that the mapping can be recursive, i.e. an entry in `options_type_map` can correspond to another configurable object, thus having its own levels of mappings.
Whether a programmer defines a single, unified `options_type_map` for a `Configurable` object, or multiple separate maps, is a choice up to the programmer.

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,493 @@
...
+ *
+ * The Configurable class also provides hooks to validate the option values
+ * associated with it
+ *   - The SanitizeOptions method can be used to check and change any options
+ * required for validity
+ *   - The ValidateOptions method can be used to check (but not change the
+ * values) of the options. These two methods must be overridden by derived
+ * classes to implement class-specific validation.
+ */
+class Configurable {
```
riversand963: According to https://google.github.io/styleguide/cppguide.html#Declaration_Order, can we re-organize the layout of the members of `Configurable`?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,493 @@
...
+  void Dump(Logger* log, const std::string& indent,
+            uint32_t mode = OptionStringMode::kOptionNone) const {
+    DumpOptions(log, indent, mode);
+  }
+  void Dump(Logger* log, uint32_t mode = OptionStringMode::kOptionNone) const {
+    Dump(log, "              ", mode);
+  }
+
+  /**
+   *  Returns the raw pointer of the named options that is used by this
```
riversand963: Indentation.

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,493 @@
...
+
+namespace rocksdb {
+struct DBOptions;
+struct ColumnFamilyOptions;
+class Logger;
+struct OptionTypeInfo;
+typedef std::unordered_map<std::string, OptionTypeInfo> OptionTypeMap;
+
+/**
+ * Configurable is an is an base class used by the rocksdb that describes a
```
riversand963: `is an is an` -> `is a`.

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,493 @@
...
+   * Returns an OK status if configuration was successful.
+   * Parameters:
+   *   opt_map                 The name/value pair map of the options to set
+   *   opt_str                 String of name/value pairs of the options to set
+   *   input_strings_escaped   True if the strings are escaped (old-style?)
+   *   ignore_unusued_options  If true and there are any unused options,
+   *                           they will be ignored and OK will be returned
+   *   unused_opts             The set of any unused option names from the map
+   */
+  Status ConfigureFromMap(const DBOptions&,
```
riversand963: I believe we have been trying to avoid default arguments for the purpose of less ambiguity, especially if there are multiple overloads. For legacy code, we may have to live with them, but for newly-added code, it is better if we can avoid them.

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,493 @@
...
+  // Returns printable options for dump
+  // If this method returns non-empty string, the result is used for dumping
+  // the options. If an empty string is returned, the registered options are
+  // used.
+  virtual std::string GetPrintableOptions() const { return ""; }
+  // Returns true if this configurable represents a mutable object
+  // Mutable classes should override this method.
+  virtual bool IsMutable() const { return false; }
+  /**
+   *  Given a prefixed name (e.g. rocksdb.my.type.opt), returns the short name
```
riversand963: Thanks for the example.

include/rocksdb/utilities/options_type.h
```diff
@@ -0,0 +1,224 @@
...
+        pfunc(_pfunc),
+        sfunc(_sfunc),
+        efunc(_efunc) {}
+
+  int offset;
+  OptionType type;
+  OptionVerificationType verification;
+  OptionTypeFlags flags;  // This is a bitmask of OptionTypeFlag values
+  int mutable_offset;
+  ParserFunc pfunc;
```
riversand963: Nit: maybe rename to `parser_func`, `string_func` and `equals_func`?

include/rocksdb/utilities/options_type.h
```diff
@@ -0,0 +1,224 @@
...
+  }
+  bool IsUniquePtr() const {
+    return (flags & OptionTypeFlags::kUnique) == OptionTypeFlags::kUnique;
+  }
+  bool IsRawPtr() const {
+    return (flags & OptionTypeFlags::kPointer) == OptionTypeFlags::kPointer;
+  }
+
+  bool IsEnum() const {
+    return type == OptionType::kEnum ||
```
riversand963: Is it possible that `type != OptionType::kEnum && ((flags & OptionTypeFlags::kEnum) == OptionTypeFlags::kEnum)`? Would you mind giving an example?

include/rocksdb/utilities/options_type.h
```diff
@@ -0,0 +1,224 @@
...
+           ((flags & OptionTypeFlags::kEnum) == OptionTypeFlags::kEnum);
+  }
+
+  bool IsConfigurable() const {
+    return ((type == OptionType::kConfigurable) ||
```
riversand963: Ditto, sorry I don't follow the case in which `type != OptionType::kConfigurable && ((flags & OptionTypeFlags::kConfigurable) == OptionTypeFlags::kConfigurable)`.

include/rocksdb/utilities/options_type.h
```diff
@@ -0,0 +1,224 @@
...
+            (flags & OptionTypeFlags::kConfigurable) ==
+                OptionTypeFlags::kConfigurable);
+  }
+
+  template <typename T>
+  const T *GetPointer(const char *addr) const {
```
riversand963: This function and its non-const counterpart seem to be returning a raw pointer to caller, whether the `addr` is nullptr, a pointer to `shared_ptr`, `unique_ptr` or raw pointer. Would `GetRawPointer()` be a better name? I do not have a strong opinion here though.

include/rocksdb/utilities/options_type.h
mrambacher: Yes (this applies to both kEnum and kConfig).  The type is a specific type in the OptionType enumeration.  Some enums or configs may have an entry here.  If so, that means those types should/may have special-case code for dealing with them in the ParseOptionHelper (and related methods) for at least one of the three (parse, string, equals) methods.

For example, there may be code to deal with the specific parse arguments for a value in ParseOptionHelper but the other methods (string, equals) are generic.  In this case, the type might be kSpecific but the OptionTypeFlags would be kEnum/kConfig.

This becomes even more common in the next PR round where I introduce a kCustom type.  The kCustom typically needs a special method for parse but can use the generic method for string and equals.

include/rocksdb/utilities/options_type.h
mrambacher: I can rename it to GetRawPointer.  Would "AsRawPointer" be a better naming choice, since it is converting "addr" into the raw pointer bits?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,495 @@
...
+#endif  // ROCKSDB_LITE
+  /**
+   * To disambiguate options from different Configurable objects,configurations
+   * may specify a prefix. It is strongly recommended that derived classes
+   * override this method to some unique representative value.
+   */
+  virtual const std::string& GetOptionsPrefix() const { return kDefaultPrefix; }
+
+ public:
+  Configurable(bool is_mutable = false) : is_mutable_(is_mutable) {}
```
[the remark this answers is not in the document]
mrambacher: Are you asking for all methods to no longer have default arguments, or only the constructor?  There are a lot of methods that have default arguments (like input_strings_escaped, ignore_unknown_options and delimiter, mode).  I can get rid of the defaults, but it seems a little silly, as the function with fewer parameters will just turn around and call the one with more.

I did eliminate the default argument to the constructor and understand the issue you raised there.

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,485 @@
+
+
```
riversand963: Copyright header is missing.

include/rocksdb/utilities/options_type.h
riversand963: SGTM

options/configurable.cc
```diff
@@ -0,0 +1,971 @@
...
+Status Configurable::SetOption(const DBOptions& db_opts,
+                               const std::string& name,
+                               const std::string& value,
+                               bool input_strings_escaped, bool* found_option) {
+  // By the time this method is called, the name is the short name (no prefix)
+  // and the value is unescaped.
+  Status status;
+  *found_option = false;
+  // Look for the name in all of the registered option maps until it is found
+  for (auto iter : options_) {
```
riversand963: `const auto& iter : options_` to avoid copy?

options/configurable.cc
```diff
@@ -0,0 +1,971 @@
...
+    auto opt_iter = FindOption(name, opt_map);
+    if (opt_iter != opt_map.end()) {
+      const auto& opt_info = opt_iter->second;
+      *found_option = true;
+      if (name == opt_iter->first) {
+        status = ParseOption(opt_info, db_opts, opt_ptr, name, value,
+                             input_strings_escaped);
+      } else {
+        status = ParseOption(opt_info, db_opts, opt_ptr,
+                             name.substr(opt_iter->first.size() + 1), value,
```
riversand963: If `name` is "test.simple.int", and `opt_iter->first` is "int", then `name.substr(opt_iter->first.size() + 1)` does not seem right.

options/configurable.cc
```diff
@@ -0,0 +1,971 @@
...
+  return status;
+}
+
+Status Configurable::ParseOption(const OptionTypeInfo& opt_info,
+                                 const DBOptions& db_opts, void* opt_ptr,
+                                 const std::string& opt_name,
+                                 const std::string& opt_value,
+                                 bool input_strings_escaped) {
+  Status status;
+  if (opt_info.verification != OptionVerificationType::kDeprecated) {
```
riversand963: Looks like we can do `if (opt_info.verification == OptionVerificationType::kDeprecated)` and return early so as to save some indentation in the remaining part of this function.

include/rocksdb/configurable.h
mrambacher: Do you want all default arguments removed, or only from the constructors?  I removed the ones from the constructors, but there are a lot of places that take other defaults (like ignore_unknown_options, input_strings_escaped, delimiter).  I can make new signatures without default arguments, but in this case I am not sure it simplifies anything to add them.

options/configurable.cc
mrambacher: I do not think that can happen.  If name is "test.simple.int", then opt_iter->first would either be "test.simple.int" (match completely) or "test" (matches to the first ".").  This case is handling the latter, and would pass "simple.int" to the next layer to be matched.

Also note that any options prefix is stripped off before this point.  So if the class had an option prefix of "test.simple" and the names passed in originally are "test.simple.int", test.simple.struct.double", and "test.bool", then the options processed here would be "int", "struct.double", and "test.bool" as the prefix would have already been removed.

include/rocksdb/configurable.h
riversand963: For existing interface, we can keep default arguments. For newly-added functions, I think we should avoid them.

include/rocksdb/configurable.h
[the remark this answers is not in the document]
riversand963: hmm, not sure why my embedded reply did not appear here.
For newly added methods, we should avoid default arguments. For existing code, we can let it go if removal of default arguments cause bulk change.

options/configurable.cc
riversand963: According to `simple_options_info` in configurable_test.cc, this can happen. Specifically in the example in configurable.cc, the opt_iter->first is "int", "bool", etc. and I believe the name can be "test.simple.int" or "test.simple.bool" if long form is used.

riversand963: Please disregard my previous comment about whether it can happen or not: we both agree it can happen.

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,491 @@
...
+   *   input_strings_escaped   True if the strings are escaped (old-style?)
+   *   ignore_unusued_options  If true and there are any unused options,
+   *                           they will be ignored and OK will be returned
+   *   unused_opts             The set of any unused option names from the map
+   */
+  Status ConfigureFromMap(const DBOptions&,
+                          const std::unordered_map<std::string, std::string>&,
+                          bool input_strings_escaped = false,
+                          bool ignore_unused_options = false);
+  Status ConfigureFromMap(const DBOptions&,
```
riversand963: I think we should avoid default arguments for both `ConfigureFromMap` and `ConfigureFromString`.
Compilers may perform implicit conversion from pointer to boolean types (https://en.cppreference.com/w/cpp/language/implicit_conversion) rather than report error as we expect.
An example is as follows.
When you accidentally call `ConfigureFromMap(..., unused)` in which `unused` is a pointer to `std::set`, the compiler will cast `unused` to a boolean and call `ConfigureFromMap(const DBOptions&, const std::unordered_map<std::string, std::string>&, bool, bool)` instead of complaining about type mismatch.

options/configurable.cc
```diff
@@ -0,0 +1,972 @@
...
+  Status s = DoConfigureFromMap(db_opts, opts_map, input_strings_escaped,
+                                ignore_unused_options, &unused_opts);
+  return s;
+}
+
+Status Configurable::ConfigureFromMap(
+    const DBOptions& db_opts,
+    const std::unordered_map<std::string, std::string>& opts_map,
+    bool input_strings_escaped, std::unordered_set<std::string>* unused) {
+  return DoConfigureFromMap(db_opts, opts_map, input_strings_escaped, true,
```
riversand963: Nit: we can add a piece of comment near constants to indicate what they mean. For example `/*ignore_unused_options=*/ true`.

options/configurable.cc
```diff
@@ -0,0 +1,975 @@
...
+      } else {
+        status = ParseOption(opt_info, db_opts, opt_ptr,
+                             name.substr(opt_iter->first.size() + 1), value,
+                             input_strings_escaped);
+      }
+      if (status.IsNotFound()) {
+        *found_option = false;
+      }
+      return status;
+    } else {
```
riversand963: Can you elaborate what this case means, i.e. `FindOption(...) == opt_map.end()`? Why in this case, we try to configure **every** field if possible by calling `AsRawPointer` until `status` is not `NotFound`?
Instead of `!status.IsNotFound()`, should it be `s.ok()`?
My understanding is: when we configure an object, and the name cannot be found, should we just flag it as an error and return to application?

mrambacher: I can add a Wiki once this is finished on how to use it.

As for the FindOption() == end() case, this code handles the case where you have option prefixes and embedded options (the ConfigureFromPropsTest shows that).  If you have an object ("test.outer") with an embedded configuration ("test.inner") and properties like:
"test.outer.int = 1; test.inner.bool = true"
 The code that removes the prefixes cannot tell that "test.inner" is a prefix for the inner case.  So we have to traverse all of the configurable embedded options to figure this out.

If the properties are written as:
int=x
inner.bool=true
We will use the FindOption() != end() case.  If the properties all have prefixes, we may use this case.

riversand963: Is the string corresponding to an object similar to our denotation of "namespace". If the string corresponding to an object is "test.out", and the object has an embedded member "inner", I expect the string corresponding to "inner" to be "test.out.inner".  What is your rule of mapping string to objects?

mrambacher: Each configuration class has a "prefix" associated with it.  This prefix is very similar to a namespace.  The default prefix is "rocksdb.".  Based on the "mode" parameter to GetOptionsString, options can be written to strings with or without their prefixes.  If written with a prefix, the prefix is appended to every option.

Additionally, options can be written embedded or detached.  Detached only applies to nested options.  When a nested option is encountered and the mode is embedded, the option is written out as a single string with brackets, e.g. "nested={int=42; string=hello}".  When detached, the options are written one per line:
nested.int=42
nested.string=hello
 
Having said that, there is a bug in how the code is read/written currently that I will need to solve.  If you have two nested configurations of the same type, the option strings cannot be differentiated.  One place where this would be a problem is in the BlockBasedTableFactory where there is a block_cache and block_cache_compressed that would not be differentiated.  Note that this issue only applies to the detached mode and not embedded.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Prerequisites for Windows:
+ Prerequisites for Windows:
```
zhichao-cao, line 1: Not need change here right?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+// Configurable is a base class used by the rocksdb that describes a
+// standard way of configuring objects.  A Configurable object can:
+//   -> Populate itself given:
+//        - One or more "name/value" pair strings
+//        - A string repesenting the set of name=value properties
+//        - A map of name/value properties.
+//   -> Convert itself into its string representation
+//   -> Dump itself to a Logger
+//   -> Compare itself to another Configurable object to see if the two objects
+// are equivalent
```
zhichao-cao, line 36: Maybe better to explain what is "two objects are equivalent".

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+    return reinterpret_cast<const T*>(GetOptionsPtr(name));
+  }
+  template <typename T>
+  T* GetOptions(const std::string& name) {
+    return reinterpret_cast<T*>(const_cast<void*>(GetOptionsPtr(name)));
+  }
+
+  // Configures the options for this class based on the input parameters.
+  // On successful completion, the object is updated with the settings from
+  // the opt_map.  If this method fails, an attempt is made to revert the
```
zhichao-cao, line 81: So only when it returns Status::OK(), the object will be updated. In all other cases, the object will be reverted to the original state?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+  // @param config_options Controls how the name/value is processed.
+  // @param name The name of the option to update
+  // @param value The value to set for the named option
+  // @return OK If the named field was successfully updated to value.
+  // @return NotFound If the name is not valid for this object.
+  // @return NotSupported  If the name is valid but the object does
+  //       not know how to convert the value.  This can happen if, for example,
+  //       there is some nested Configurable that cannot be created.
+  // @return InvalidArgument If the value cannot be successfully  parsed.
+  Status ConfigureOption(const ConfigOptions& config_options,
```
zhichao-cao, line 125: Would be better to name it ConfigureOneOption? What do you think?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+#ifndef ROCKSDB_LITE
+  // Returns the serialized options for this object.
+  // This method is similar to GetOptionString with no errors.
+  // @param config_options Controls how serialization happens.
+  // @param prefix A string to prepend to every option.
+  // @return The serialized representation of the options for this object
+  std::string ToString(const ConfigOptions& config_options) const {
+    return ToString(config_options, "");
+  }
+  std::string ToString(const ConfigOptions& config_options,
```
zhichao-cao, line 172: What is the prefix used for? Can you give an example?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+  // Returns the value of the option associated with the input name
+  // This method is the functional inverse of ConfigureOption
+  // @param config_options Controls how the value is returned
+  // @param name The name of the option to return a value for.
+  // @param value The returned value associated with the named option.
+  // @return OK If the named field was successfully updated to value.
+  // @return NotFound If the name is not valid for this object.
+  // @param InvalidArgument If the name is valid for this object but
+  //      its value cannot be serialized.
+  virtual Status GetOption(const ConfigOptions& config_options,
```
zhichao-cao, line 192: `GetOneOption` would be better?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+#endif  // ROCKSDB_LITE
+
+  // Checks to see if this Configurable is equivalent to other.
+  // This method assumes that the two objects are of the same class.
+  // @param config_options Controls how the options are compared.
+  // @param other The other object to compare to.
+  // @param mismatch If the objects do not match, this parameter contains
+  //      the name of the option that triggered the match failure.
+  // @param True if the objects match, false otherwise.
+  virtual bool AreEqual(const ConfigOptions& config_options,
```
zhichao-cao, line 203: If all the options of object A is the same as object B, but object B has new options (e.g., code updates), the two objects are not equal. Is it possible that we will have backward compatible issues?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+  virtual Status ValidateOptions(const DBOptions& db_opts,
+                                 const ColumnFamilyOptions& cf_opts) const;
+
+ protected:
+  // True once the object is prepared.  Once the object is prepared, only
+  // mutable options can be configured.
+  bool prepared_;
+  // If this class is a wrapper (has-a), this method should be
+  // over-written to return the inner configurable (like an EnvWrapper).
+  virtual Configurable* Inner() const { return nullptr; }
```
zhichao-cao, line 252: If the wrapper is for example 3 level instead of 2 levels (env->EnvWrapper), it will return the inner most object  pointer or just one level inside?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+                               const OptionTypeInfo& opt_info,
+                               const std::string& name,
+                               const void* const this_ptr,
+                               const void* const that_ptr,
+                               std::string* bad_name) const;
+#endif
+#ifndef ROCKSDB_LITE
+  // Internal method to serialize options (ToString)
+  // Classes may override this value to change its behavior.
+  virtual std::string AsString(const ConfigOptions& config_options,
```
zhichao-cao, line 300: Maybe directly call it `SerializeOptions`?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+
+  // Internal method to serialize a set of options for this object.
+  // Classes may override this value to change its behavior.
+  // @param config_options Controls how the options are being serialized
+  // @param prefix String to prepend to every option being serialized
+  // @param type_map The map of options for this name
+  // @param opt_ptr Pointer to the object being configured for this option set.
+  // @param result The serailized result
+  // @return OK If all of the options in type_map were successfully serialized
+  // @return InvalidArgument If an option could not be serialized.
```
zhichao-cao, line 311: Where is this function?

include/rocksdb/convenience.h
```diff
@@ -46,9 +46,15 @@ struct ConfigOptions {
   // When true, any unused options will be ignored and OK will be returned
   bool ignore_unknown_options = false;
 
+  // When true, any unsupported options will be ignored and OK will be returned
+  bool ignore_unsupported_options = true;
```
zhichao-cao, line 50: What' the difference between `ignore_unknown_options` and `ignore_unsupported_options`? Are the "unsupported" options a subset of "known options"？

include/rocksdb/utilities/object_registry.h
```diff
@@ -169,7 +169,7 @@ class ObjectRegistry {
 
   // Creates a new static T using the input factory functions.
   // Returns OK if a new static T was successfully created
-  // Returns NotFound if the type/target could not be created
+  // Returns NotSupporte if the type/target could not be created
```
zhichao-cao, line 172: Nit: NotSupporte->NotSupported

include/rocksdb/utilities/options_type.h
```diff
@@ -75,7 +75,12 @@ enum class OptionTypeFlags : uint32_t {
   kCompareExact = ConfigOptions::kSanityLevelExactMatch,
 
   kMutable = 0x0100,        // Option is mutable
+  kPointer = 0x0200,        // The option is stored as a pointer
```
zhichao-cao, line 78: Can we make it clear: kPointer->kRawPointer (if I understand correctly. Or it is any pointer type including shared and unique?); kShared->kSharedPointer; kUnique->kUniquePointer?

options/cf_options.cc
```diff
@@ -451,184 +549,224 @@ std::unordered_map<std::string, OptionTypeInfo>
           }}},
         {"table_factory",
          {offset_of(&ColumnFamilyOptions::table_factory),
-          OptionType::kTableFactory, OptionVerificationType::kByName,
-          OptionTypeFlags::kCompareLoose, 0}},
+          OptionType::kConfigurable, OptionVerificationType::kByName,
+          (OptionTypeFlags::kShared | OptionTypeFlags::kCompareLoose |
```
zhichao-cao, line 553: We may need to give the guidelines for developer if they are adding new entry like table_tactoy or other factory, how to decide the OptionTypeFlags. For example, tt might be hard to understand why it is "OptionTypeFlags::kShared | OptionTypeFlags::kCompareLoose | OptionTypeFlags::kDontPrepare".

include/rocksdb/configurable.h
mrambacher: An attempt is made to revert it back to the original state, but that may always be possible (at least currently).  For example, if you had a BlockBasedTable with a block cache, the cache may not be reverted to the original one.

include/rocksdb/configurable.h
mrambacher: I think the fact that it is singular (ConfigureOption) and not plural should be good enough to show that it is configuring a single item.  And in reality it could be configuring multiple items if the "name" refers to something which itself has options (like a Struct).

include/rocksdb/configurable.h
mrambacher: ConfigureOption and GetOption are the inverse of one another.  We can rename them both if you want, but I think the current names are fine myself.

include/rocksdb/configurable.h
mrambacher: At the moment, the prefix is probably not used.  However, there was talk of being able to do a couple different things that would require the prefix:
- Be able to list the embedded options as one per line rather than inside of brackets
- Be able to specify a prefix to print for every option (e.g. rocksdb.XXX or rocksdb.table.block.XXX)
Both of these will require a prefix.

include/rocksdb/configurable.h
mrambacher: I do not think there is a backward compatibility issue, though perhaps this method should be "AreEquivalent" as opposed to "AreEqual".
- If every option in A matches the value in B, then B is equivalent to A.  
- If every option in B also matches the value in A, then A and B would be equal.

I am not sure where there is an issue of backwards compatibility that did not already exist in the "old" way of doing things.  If there is a test you would like me to run to prove/verify that, please let me know what I need to do.

include/rocksdb/configurable.h
mrambacher: Inner should not recurse.  It should only return one-level deep.  If the caller wants to know the deepest level, they would have to implement the recursion themselves.

This method is useful in other cases as well that will come up later.  For example, if passed an Env object and I need to find the Default or Posix one, I could loop over the Inner() ones until I found the one that matched the pattern I need.

include/rocksdb/configurable.h
mrambacher: I am not sure what you are asking for here.

include/rocksdb/convenience.h
mrambacher: Unsupported options are the ones who, when being configured, return Status::NotSupported.  These are potentially valid options for the object under certain conditions but fail in this case.  For example, the CuckooTable might be a valid "table_factory" but was not enabled or compiled in.  In this case, requesting to configure a table_factory=CuckooTable would return NotSupported.  Unsupported implies that, the named option is a valid option but for some reason the value is not valid under these circumstances.

Unknown options are a catch-all for a lot of errors, and are already used as such.  An Unknown option might be one were the name is not valid (e.g. "no_such_option=value") or it might be one who has an illegal value to int (my_int=fred).

include/rocksdb/configurable.h
zhichao-cao: Oh, I see. Thanks for the explanation.

include/rocksdb/configurable.h
zhichao-cao: Maybe better to point out it is not recursive in the comments.

include/rocksdb/configurable.h
zhichao-cao: I mean naming it to `SerializeOptions` instead of `AsString`

include/rocksdb/configurable.h
zhichao-cao: Maybe better add this example to the comment.

include/rocksdb/configurable.h
zhichao-cao: I may prefer to change change both, make it explicit.

include/rocksdb/configurable.h
zhichao-cao: Yes, `AreEquivalent` is more clear.

options/configurable.cc
```diff
@@ -0,0 +1,606 @@
...
+        if (s.IsNotFound()) {
+          ++it;
+        } else if (s.IsNotSupported()) {
+          notsup = s;
+          unsupported.insert(it->first);
+          ++it;  // Skip it for now
+        } else {
+          found++;
+          it = options->erase(it);
+          if (!s.ok()) {
```
zhichao-cao, line 339: In this case, `result` stores the last non-ok status?

options/configurable.cc
```diff
@@ -0,0 +1,606 @@
...
+}
+
+Status ConfigurableHelper::ConfigureOption(
+    const ConfigOptions& config_options, Configurable& configurable,
+    const OptionTypeInfo& opt_info, const std::string& opt_name,
+    const std::string& name, const std::string& value, void* opt_ptr) {
+  if (opt_name == name) {
+    return configurable.ParseOption(config_options, opt_info, opt_name, value,
+                                    opt_ptr);
+  } else if (opt_info.IsStruct()) {
```
zhichao-cao, line 391: We can combine the condition as `if (opt_info.IsStruct() || opt_info.IsConfigurable())`

include/rocksdb/configurable.h
zhichao-cao: Sure. keeping the current names is fine.

options/configurable.cc
mrambacher: Yes, added a comment on their declaration

options/configurable.cc
mrambacher: Done

options/cf_options.cc
mrambacher: Added more comments

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+  // @param opt_map Options map that controls how this option is configured.
+  void RegisterOptions(
+      const std::string& name, void* opt_ptr,
+      const std::unordered_map<std::string, OptionTypeInfo>* opt_map);
+
+  //  Given a name (e.g. rocksdb.my.type.opt), returns the short name (opt)
+  virtual std::string GetOptionName(const std::string& long_name) const;
+
+ private:
+  std::vector<RegisteredOptions> options_;
```
zhichao-cao, line 331: Need a comments here to explain what is stored in `options_` and how it is supposed to be used.

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,333 @@
...
+  virtual bool AreEquivalent(const ConfigOptions& config_options,
+                             const Configurable* other,
+                             std::string* name) const;
+
+  // Returns a pretty-printed, human-readable version of the options.
+  // This method is typically used to dump the options to a log file.
+  // Classes should override this method
+  virtual std::string GetPrintableOptions() const { return ""; }
+
+  // Initializes the options and makes sures the settings are valid/consistent
```
zhichao-cao, line 215: Should be more clear to state like: `Initializes the options parameters/controls for comparing objects and converting to/from strings, and .....`

options/configurable.cc
```diff
@@ -0,0 +1,604 @@
...
+  }
+  return s;
+}
+
+#ifndef ROCKSDB_LITE
+/**
+ * Updates the object with the named-value property values, returning OK on
+ * succcess. Any properties that were found are removed from the options list;
+ * upon return only options that were not found in this opt_map remain.
+ * Returns OK if the properties were successfully updated, N
```
zhichao-cao, line 304: What is the `N` in the comment here?

options/configurable_test.cc
```diff
@@ -0,0 +1,791 @@
...
+  ASSERT_EQ(opts_str, "false");
+}
+
+TEST_F(ConfigurableTest, NestedUniqueConfigTest) {
+  std::unique_ptr<Configurable> simple;
+  simple.reset(
+      SimpleConfigurable::Create("Outer", TestConfigMode::kAllOptMode));
+  const auto outer = simple->GetOptions<TestOptions>("Outer");
+  const auto unique =
+      simple->GetOptions<std::unique_ptr<Configurable>>("OuterUnique");
```
zhichao-cao, line 415: `OuterUnique` is not registered in the simple (outer is registered when calling Create), should unique be `nullptr`? I'm not clear here.

mrambacher: In the SimpleConfigurable constructor around line 65 it registers the "OuterUnique" name.  It does this by taking the name of the object ("Outer") and concatenating it with "Unique".

Note that how the tests are registered and manipulated is way more complex than what the normal code is.  I tried to write the tests to use as few test classes to cover as many cases as possible and it created more complex tests than I would have liked.

zhichao-cao: I see. Thanks for the explanation!

zhichao-cao:
A clear statement in HISTORY.md is needed since it makes the bottle layer implementation and adds the new interfaces.

options/configurable_test.cc
```diff
@@ -0,0 +1,791 @@
...
+  ASSERT_NE(outer, nullptr);
+  ASSERT_NE(unique, nullptr);
+  ASSERT_OK(
```
zhichao-cao, line 418: Just to clarify, unique is the pointer to the object in `SimpleConfigurable`, pointed by `unique_`. It is a configurable created with name"Unique" + name_ (UniqueOuter). When we use nested options (unique={int=42;string=nested}), it configures the inner object pointed by unique_. So, `unique->get()->GetOptions<TestOptions>("UniqueOuter") `will get the object registered in `unique_` configured with "42" and "nexted".

HISTORY.md
```diff
@@ -4,7 +4,7 @@
 * Encryption file classes now exposed for inheritance in env_encryption.h
 * File I/O listener is extended to cover more I/O operations. Now class `EventListener` in listener.h contains new callback functions: `OnFileFlushFinish()`, `OnFileSyncFinish()`, `OnFileRangeSyncFinish()`, `OnFileTruncateFinish()`, and ``OnFileCloseFinish()``.
 * `FileOperationInfo` now reports `duration` measured by `std::chrono::steady_clock` and `start_ts` measured by `std::chrono::system_clock` instead of start and finish timestamps measured by `system_clock`. Note that `system_clock` is called before `steady_clock` in program order at operation starts.
-
+* Methods to configure serialize, and compare -- such as TableFactory -- are exposed directly through the Configurable base class (from which these objects inherity).  This change will allow for better and more thorough configuration management and retrieval in the future
```
zhichao-cao, line 7: I suggest to move it to ### New Features instead of ###Public API Change. It is a new way of using Configurable to configure, serialize, and control the configure the DB and some classes. I suggest to write more details (but in brief) including the ColumnFamilyOptions, TableFactory, and DBOptions changes and the new way of using them when developing. Also summarize the new APIs.

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,335 @@
...
+  }
+  std::string ToString(const ConfigOptions& config_options,
+                       const std::string& prefix) const;
+
+  // Returns the list of option names associated with this configurable
+  // @param config_options Controls how the names are returned
+  // @param result The set of option names for this object. Note that
+  //      options that are deprecated or aliases are not returned.
+  // @return OK on success.
+  Status GetOptionNames(const ConfigOptions& config_options,
```
zhichao-cao, line 183: Just double check, any class that inherited from Configurable should not/does not need to override any functions above?

include/rocksdb/configurable.h
```diff
@@ -0,0 +1,335 @@
...
+  // returning a sub-option)
+  //
+  // The default implemntation looks at the registered options.  If the
+  // input name matches that of a registered option, the pointer registered
+  // with that name is returned.
+  // e.g,, RegisterOptions("X", &my_ptr, ...); GetOptionsPtr("X") returns
+  // "my_ptr"
+  virtual const void* GetOptionsPtr(const std::string& name) const;
+
+  virtual Status ParseStringOptions(const ConfigOptions& config_options,
```
zhichao-cao, line 283: Need some comments for `ParseStringOptions`, `ConfigureOptions` and `ParseOption` here as protected members

options/cf_options.cc
```diff
@@ -506,204 +549,233 @@ std::unordered_map<std::string, OptionTypeInfo>
...
+           OptionTypeFlags::kDontPrepare),
```
zhichao-cao, line 554: what's the reason that `table_factory` should not call `PrepareOptions`?

include/rocksdb/configurable.h
mrambacher: Any class that is Configurable and registers their options via the RegisterOptions method, will get the Configurable functionality for free.

In the instances where I have had to override methods in the implementation (like ParseOption), it is because of an existing nuance in how that class handled errors typically.  I have implemented many new classes as Configurable (in a future PR) and have not seen the need to override methods.

options/cf_options.cc
mrambacher: It is because of how the table factory representation is stored in the Options file.  Inside of the options file, the "Column Family" will contain something like "table_factory = BlockBased" with just the class of the TableFactory being used.  A later section of the Options file will have the configuration for the TableFactory for that Column Family Option".  Because the table factory ends up being configured (and prepared) later, it should not be prepared as part of the column family but as part of the setup of the table factory (at least when configured via the Options file).

OTOH, the "block_based_table_factory" and "plain_table_factory" options are used for testing and setup (not persisting the options of the table factory).  Therefore they can be configured earlier.

options/configurable_test.cc
mrambacher: I believe that is true, yes.  Note that this odd nature/complication is just part of the test suite and not the configurable stuff itself.  I do not envision classes having such a strange hierarchy ever in their implementations, but adding a bunch of one-off classes to do the testing was creating headaches.

HISTORY.md
```diff
@@ -4,13 +4,13 @@
...
 * When a DB sets `stats_dump_period_sec > 0`, either as the initial value for DB open or as a dynamic option change, the first stats dump is staggered in the following X seconds, where X is an integer in `[0, stats_dump_period_sec)`. Subsequent stats dumps are still spaced `stats_dump_period_sec` seconds apart.
 * When the paranoid_file_checks option is true, a hash is generated of all keys and values are generated when the SST file is written, and then the values are read back in to validate the file.  A corruption is signaled if the two hashes do not match.
+* The settings of the DBOptions and ColumnFamilyOptions are now managed by Configurable objects (see New Features).  The same convenience methods to configure these options still exist but the backend implementation has been unified under a common implementation.
```
ajkr, line 13: I think for "Behavior Changes" section we should be more specific about exactly what change the user will observe. If the APIs behave the same from the user's perspective, perhaps this fits better in "New Features".

mrambacher: I will move this wherever you deem appropriate.  The APIs to get and set the DB and CF Options have not changed and there is no real change from the user perspective (except possibly an occasional different status).  The change for DB/CF Options is not really a "New Feature" either, as it still behaves the same under a "New Implementation".

What would you like done?

ajkr: Maybe start a "General Improvements" section for mentioning things that don't fall under a category. It looks like there's a precedent for this in 6.3. BTW, this will need to be rebased as the sections it modifies are part of the already in-progress 6.12 release.

PR #5759, opened 2019-08-30
Option to make write group size configurable

ronak-07:
The max batch size that we can write to the WAL is controlled by a static manner. So if the leader write is less than 128 KB we will have the batch size as leader write size + 128 KB else the limit will be 1 MB. Both of them are statically defined.


ronak-07:
@siying Can you please take a look?

ronak-07:
@riversand963 Could you please take a look?

db/write_thread.h
```diff
@@ -360,6 +360,14 @@ class WriteThread {
   // Enable pipelined write to WAL and memtable.
   const bool enable_pipelined_write_;
 
+  // The lower limit of maximum number of bytes that are written in a single
+  // batch of WAL or memtable write.
+  const uint64_t lower_limit_write_batch_group_size_bytes;
```
siying: Consider a better description than "lower_limit". In my understanding, if a write batch is larger than this value, they are not batched? Than maybe something like "max_write_size_to_batch", or something like that?

db/write_thread.cc
```diff
@@ -406,9 +410,9 @@ size_t WriteThread::EnterAsBatchGroupLeader(Writer* leader,
   // Allow the group to grow up to a maximum size, but if the
   // original write is small, limit the growth so we do not slow
   // down the small write too much.
-  size_t max_size = 1 << 20;
-  if (size <= (128 << 10)) {
-    max_size = size + (128 << 10);
+  size_t max_size = upper_limit_write_batch_group_size_bytes;
+  if (size <= lower_limit_write_batch_group_size_bytes) {
```
siying: We always try to minimize options. Do we need both of upper and lower for now?
Can we say users only take upper limit, and the lower limit is going to be 1/8 of that?

db/write_thread.h
```diff
@@ -360,6 +360,14 @@ class WriteThread {
...
+
+  // The upper limit of maximum number of bytes that are written in a single
+  // batch of WAL or memtable write.
+  const uint64_t upper_limit_write_batch_group_size_bytes;
```
siying: Also think about a better name. Is `max_write_batch_group_size` better, for example?

db/write_thread.h
ronak-07: Updated the name to 'max_write_batch_group_size_bytes'

db/write_thread.cc
ronak-07: Good suggestion. Updated options to allow configuration for just the max limit. The lower limit will be 1/8

db/write_thread.h
ronak-07: Updated the name to 'max_write_batch_group_size_bytes'.
Appended bytes in the name to specify unit of size.

ronak-07:
Thank you Siying for the review. Addressed your comments.

siying:
appveyor seems to show some errors:
```
       "C:\projects\rocksdb\build\rocksdb-shared.vcxproj" (default target) (176) ->
         c:\projects\rocksdb\db\write_thread.cc(413): warning C4244: '=': conversion from 'double' to 'size_t', possible loss of data [C:\projects\rocksdb\build\rocksdb-shared.vcxproj]
         c:\projects\rocksdb\db\write_thread.cc(492): warning C4244: '=': conversion from 'double' to 'size_t', possible loss of data [C:\projects\rocksdb\build\rocksdb-shared.vcxproj]
```

db/write_thread.cc
```diff
@@ -485,9 +487,9 @@ void WriteThread::EnterAsMemTableWriter(Writer* leader,
   // Allow the group to grow up to a maximum size, but if the
   // original write is small, limit the growth so we do not slow
   // down the small write too much.
-  size_t max_size = 1 << 20;
-  if (size <= (128 << 10)) {
-    max_size = size + (128 << 10);
+  size_t max_size = max_write_batch_group_size_bytes;
+  if (size <= 0.125 * max_write_batch_group_size_bytes) {
```
siying: Do "/8" instead to avoid the type conversion trouble.

db/write_thread.cc
```diff
@@ -485,9 +487,9 @@ void WriteThread::EnterAsMemTableWriter(Writer* leader,
...
+    max_size = size + 0.125 * max_write_batch_group_size_bytes;
```
siying: Maybe define a variable with value 0.125 * max_write_batch_group_size_bytes, rather than repeat it twice.

db/write_thread.cc
ronak-07: Done in the latest patch.

db/write_thread.cc
ronak-07: Done in latest patch.

db/write_thread.cc
```diff
@@ -410,7 +410,7 @@ size_t WriteThread::EnterAsBatchGroupLeader(Writer* leader,
   // down the small write too much.
   size_t max_size = max_write_batch_group_size_bytes;
   if (size <= 0.125 * max_write_batch_group_size_bytes) {
-    max_size = size + 0.125 * max_write_batch_group_size_bytes;
+    max_size = size + (size_t)(0.125 * max_write_batch_group_size_bytes);
```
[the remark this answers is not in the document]
ronak-07: Yes, I fixed with /8 suggestion in latest patch.

Actually when I checked RocksDb code, at some places they had used C style casts only. Though I should have used C++ style casting.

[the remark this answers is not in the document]
siying: Yes I noticed that too. We need to fix them eventually.

siying:
@ronak-07 I'm landing it because I don't want one more round-trip, but I remember one thing: we need to add the feature description to HISTORY.md. Can you add it as a separate pull request?

ronak-07:
@siying Sure. Would add it separately.
Thanks for the comments and review.

PR #5772, opened 2019-09-06
add start_level in CompactRangeOptions

hicqu:
We are trying to use `CompactionFilter` to improvement TiKV's GC. We want to do transaction garbage collection with RocksDB's compaction. We want to call `CompactRange` manually if there is too much garbage in some regions. However currently `CompactRange` must flush memtables, and then compact level 0, which is too expensive. This PR add a `start_level` in `CompactRangeOptions`. By default it's `-1`, which means the compaction starts from memtable, which is the same behavior with the current implementation.

And we also add a `output_level` field in `CompactionFilterContext`.


yiwu-arbug:
@siying any comment about this one? Thanks.

hicqu:
PTAL @ltamasi thanks!

PR #5775, opened 2019-09-06
Remove some code duplication in bloom filter code

anand1976:
Consolidate multiple definitions of ```GetTotalBitsForLocality()``` into one inline function. The duplication was also causing a failure in the Sandcastle unity test.

Test plan:
make check
make asan_check
make unity_test


anand1976:
Looks like #5762 has addressed this. Closing the PR.

PR #5783, opened 2019-09-09
Fix block allocation bug in new DynamicBloom

pdillinger:
Summary: Bug found by valgrind. New DynamicBloom wasn't allocating in
block sizes. New assertion added that probes starting in final word
would be in bounds.

Test Plan: ROCKSDB_VALGRIND_RUN=1 DISABLE_JEMALLOC=1 valgrind --leak-check=full ./dynamic_bloom_test


pdillinger:
> LGTM. Some comments about kNumDoubleProbes would help understand the calculations better.

It's documented in the .h file.

PR #5790, opened 2019-09-10
Fix signed/unsigned mismatch in assertion in db/plain_table_db_test.cc

pdillinger:
Summary: Caused a windows build error


pdillinger:
https://github.com/facebook/rocksdb/pull/5788

PR #5796, opened 2019-09-12
fast look up purge_queue

yiwu-arbug:
purge_queue_ maybe contains thousands sst files, for example manual compact a range. If full scan is triggered at the same time and the total sst files number is large, RocksDB will be blocked at https://github.com/facebook/rocksdb/blob/master/db/db_impl_files.cc#L150 for several seconds. In our environment we have 140,000 sst files and the manual compaction delete about 1000 sst files, it blocked about 2 minutes.

Commandeering #5290.


db/db_impl/db_impl.cc
```diff
@@ -1309,32 +1309,28 @@ void DBImpl::SchedulePurge() {
...
+  while (!logs_to_free_queue_.empty()) {
+    assert(!logs_to_free_queue_.empty());
+    log::Writer* log_writer = *(logs_to_free_queue_.begin());
+    logs_to_free_queue_.pop_front();
+    mutex_.Unlock();
+    delete log_writer;
+    mutex_.Lock();
+  }
+  for (auto file : purge_files_) {
+    auto purge_file = file.second;
```
riversand963: Perhaps it is clearer to specify the types explicitly. I know we used `auto` in the past, but would you mind adding the types back? https://google.github.io/styleguide/cppguide.html I am looking at the "Type deduction" section's decision.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3039,34 +3039,25 @@ void DBImpl::InstallSuperVersionAndScheduleWork(
...
 // full_scan=true can issue I/O requests to obtain list of files in
 // directories, e.g. env_->getChildren while holding db mutex.
-// In the future, if we want to reduce the cost of search, we may try to keep
-// the vector sorted.
 bool DBImpl::ShouldPurge(uint64_t file_number) const {
-  for (auto fn : files_grabbed_for_purge_) {
-    if (file_number == fn) {
-      return false;
-    }
+  if (files_grabbed_for_purge_.find(file_number) !=
```
riversand963: Can we combine these into one statement? Something like the following.
```
return files_grabbed_for_purge_.find(file_number) == files_grabbed_for_purge_.end() &&
            purge_files_.find(file_number) == purge_files_.end();
```

db/db_impl/db_impl_files.cc
```diff
@@ -495,13 +495,15 @@ void DBImpl::PurgeObsoleteFiles(JobContext& state, bool schedule_only) {
     // After purging obsolete files, remove them from files_grabbed_for_purge_.
     // Use a temporary vector to perform bulk deletion via swap.
     InstrumentedMutexLock guard_lock(&mutex_);
-    std::vector<uint64_t> tmp;
+    std::vector<uint64_t> to_be_removed;
```
riversand963: How about using `autovector`?

db/obsolete_files_test.cc
```diff
@@ -171,8 +171,8 @@ TEST_F(ObsoleteFilesTest, RaceForObsoleteFileDeletion) {
       });
   SyncPoint::GetInstance()->SetCallBack(
       "DBImpl::CloseHelper:PendingPurgeFinished", [&](void* arg) {
-        std::vector<uint64_t>* files_grabbed_for_purge_ptr =
-            reinterpret_cast<std::vector<uint64_t>*>(arg);
+        std::unordered_set<uint64_t>* files_grabbed_for_purge_ptr =
```
riversand963: Here, the type declaration can be `auto` since `reinterpret_cast` already specifies the type.

riversand963:
Friendly ping.

yiwu-arbug:
Sorry for delay. Updated.

PR #5811, opened 2019-09-16
Temporarily disable partitioned index/filter in stress test

ltamasi:
Summary:
PR #4020 enabled partitioned indexes/filters in stress tests; however,
this causes assertion failures in BatchedOpsStressTest. This patch
disables them until we can root cause the failures.

Test Plan:
Ran the script and made sure it only uses the binary search index.


riversand963:
There seems to be a bug, because #4020 was a long time ago.

PR #5818, opened 2019-09-17
MultiGet batching in memtable

vjnadimpalli:
RocksDB has a MultiGet() API that implements batched key lookup for higher performance (https://github.com/facebook/rocksdb/blob/master/include/rocksdb/db.h#L468). Currently, batching is implemented in BlockBasedTableReader::MultiGet() for SST file lookups. One of the ways it improves performance is by pipelining bloom filter lookups (by prefetching required cachelines for all the keys in the batch, and then doing the probe) and thus hiding the cache miss latency. The same concept can be extended to the memtable as well. This PR involves implementing a pipelined bloom filter lookup in DynamicBloom, and implementing MemTable::MultiGet() that can leverage it.

Test plan:
Existing tests

Performance Test:
Ran the below command which fills up the memtable and makes sure there are no flushes and then call multiget. Ran it on master and on the new change and see an average of 3.5% improvement across 10 runs. 

Screen Shot 2019-10-03 at 2 00 28 PM

TEST_TMPDIR=. numactl -C 10 ./db_bench -benchmarks="fillseq,readrandom,multireadrandom" -num=6000000 -compression_type="none" -level_compaction_dynamic_level_bytes -write_buffer_size=2000000000 -target_file_size_base=2000000000 -max_bytes_for_level_base=16777216 -reads=90000 -threads=1 -compression_type=none -cache_size=4194304000 -batch_size=32 -disable_auto_compactions=true -bloom_bits=10 -cache_index_and_filter_blocks=true -pin_l0_filter_and_index_blocks_in_cache=true -multiread_batched=true -multiread_stride=16 -statistics -memtable_whole_key_filtering=true -memtable_bloom_size_ratio=10

### DIFF
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -1362,6 +1362,7 @@ class MultiGetPrefixExtractorTest : public DBBasicTest,
 TEST_P(MultiGetPrefixExtractorTest, Batched) {
   Options options = CurrentOptions();
   options.prefix_extractor.reset(NewFixedPrefixTransform(2));
+  options.memtable_prefix_bloom_size_ratio = 10;
   BlockBasedTableOptions bbto;
   if (GetParam()) {
     bbto.index_type = BlockBasedTableOptions::IndexType::kTwoLevelIndexSearch;
@@ -1379,9 +1380,17 @@ TEST_P(MultiGetPrefixExtractorTest, Batched) {
   ASSERT_OK(Put("kk2", "v2"));
   ASSERT_OK(Put("kk3", "v3"));
   ASSERT_OK(Put("kk4", "v4"));
+  std::vector<std::string> keys({"k", "kk1", "kk2", "kk3", "kk4"});
+  std::vector<std::string> inmem_values;
+  inmem_values = MultiGet(keys, nullptr);
+  ASSERT_EQ(inmem_values[0], "v0");
+  ASSERT_EQ(inmem_values[1], "v1");
+  ASSERT_EQ(inmem_values[2], "v2");
+  ASSERT_EQ(inmem_values[3], "v3");
+  ASSERT_EQ(inmem_values[4], "v4");
+
   ASSERT_OK(Flush());
 
-  std::vector<std::string> keys({"k", "kk1", "kk2", "kk3", "kk4"});
   std::vector<std::string> values;
   SetPerfLevel(kEnableCount);
   get_perf_context()->Reset();
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -1374,25 +1374,30 @@ TEST_P(MultiGetPrefixExtractorTest, Batched) {
   options.table_factory.reset(NewBlockBasedTableFactory(bbto));
   Reopen(options);
 
+  SetPerfLevel(kEnableCount);
+  get_perf_context()->Reset();
+
   // First key is not in the prefix_extractor domain
   ASSERT_OK(Put("k", "v0"));
   ASSERT_OK(Put("kk1", "v1"));
   ASSERT_OK(Put("kk2", "v2"));
   ASSERT_OK(Put("kk3", "v3"));
   ASSERT_OK(Put("kk4", "v4"));
-  std::vector<std::string> keys({"k", "kk1", "kk2", "kk3", "kk4"});
+  std::vector<std::string> mem_keys(
+      {"k", "kk1", "kk2", "kk3", "kk4", "rofl", "lmho"});
   std::vector<std::string> inmem_values;
-  inmem_values = MultiGet(keys, nullptr);
+  inmem_values = MultiGet(mem_keys, nullptr);
   ASSERT_EQ(inmem_values[0], "v0");
   ASSERT_EQ(inmem_values[1], "v1");
   ASSERT_EQ(inmem_values[2], "v2");
   ASSERT_EQ(inmem_values[3], "v3");
   ASSERT_EQ(inmem_values[4], "v4");
-
+  ASSERT_EQ(get_perf_context()->bloom_memtable_miss_count, 2);
+  ASSERT_EQ(get_perf_context()->bloom_memtable_hit_count, 5);
   ASSERT_OK(Flush());
 
+  std::vector<std::string> keys({"k", "kk1", "kk2", "kk3", "kk4"});
   std::vector<std::string> values;
-  SetPerfLevel(kEnableCount);
   get_perf_context()->Reset();
   values = MultiGet(keys, nullptr);
   ASSERT_EQ(values[0], "v0");
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -1280,6 +1280,8 @@ TEST_F(DBBasicTest, MultiGetBatchedSimpleSorted) {
 TEST_F(DBBasicTest, MultiGetBatchedMultiLevel) {
   Options options = CurrentOptions();
   options.disable_auto_compactions = true;
+  options.memtable_whole_key_filtering = true;
+  options.memtable_prefix_bloom_size_ratio = 10;
   Reopen(options);
   int num_keys = 0;
 
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -1280,8 +1280,6 @@ TEST_F(DBBasicTest, MultiGetBatchedSimpleSorted) {
 TEST_F(DBBasicTest, MultiGetBatchedMultiLevel) {
   Options options = CurrentOptions();
   options.disable_auto_compactions = true;
-  options.memtable_whole_key_filtering = true;
-  options.memtable_prefix_bloom_size_ratio = 10;
   Reopen(options);
   int num_keys = 0;
 
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -1959,41 +1959,58 @@ void DBImpl::MultiGetImpl(
     bool lookup_current = false;
 
     keys_left -= batch_size;
+//    for (auto mget_iter = range.begin(); mget_iter != range.end();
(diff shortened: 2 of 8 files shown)


db/memtable.cc
```diff
@@ -769,7 +770,8 @@ bool MemTable::Get(const LookupKey& key, std::string* value, Status* s,
                    MergeContext* merge_context,
                    SequenceNumber* max_covering_tombstone_seq,
                    SequenceNumber* seq, const ReadOptions& read_opts,
-                   ReadCallback* callback, bool* is_blob_index, bool do_merge) {
+                   ReadCallback* callback, bool* is_blob_index, bool do_merge,
+                   bool skip_filter_check) {
```
anand1976: Rather than add another parameter with default value to skip bloom filter, I'd suggest putting the code to call table_->Get in a separate function that can be called by both Get and MultiGet.

db/memtable.cc
```diff
@@ -842,6 +846,66 @@ bool MemTable::Get(const LookupKey& key, std::string* value, Status* s,
...
+  for (auto iter = range->begin(); iter != range->end(); ++iter) {
+    Slice user_key = iter->lkey->user_key();
+    size_t ts_sz =
+        GetInternalKeyComparator().user_comparator()->timestamp_size();
+    bool may_contain = true;
+    if (bloom_filter_) {
+      // when both memtable_whole_key_filtering and prefix_extractor_ are set,
+      // only do whole key filtering for Get() to save CPU
+      if (moptions_.memtable_whole_key_filtering) {
+        may_contain = bloom_filter_->MayContain(
```
anand1976: We'll still end up probing the bloom filter one by one for each key, which doesn't allow the cache access to be optimized. The range can be pushed down to the bloom filter implementation. Then all the relevant cachelines can be prefetched in one shot, followed by the probes here - https://github.com/facebook/rocksdb/blob/master/util/dynamic_bloom.h#L157. We do something similar in the full filter block - https://github.com/facebook/rocksdb/blob/master/table/block_based/full_filter_block.cc#L191-L218

db/memtable.cc
```diff
@@ -842,6 +846,66 @@ bool MemTable::Get(const LookupKey& key, std::string* value, Status* s,
...
+            StripTimestampFromUserKey(user_key, ts_sz));
+      } else {
+        assert(prefix_extractor_);
+        may_contain =
+            !prefix_extractor_->InDomain(user_key) ||
+            bloom_filter_->MayContain(prefix_extractor_->Transform(user_key));
+      }
+
+      if (may_contain) {
+        keys.set(idx);
```
anand1976: If ```DynamicBloom::MayContain()``` is passed a MultiGetRange, it can just call ```SkipKey(iter)``` on that range to skip a key that got a negative result in the bloom filter lookup. Then all you have to do later is just iterate through that range to find keys that may be present in the memtable. There'd be no need for the ```keys``` bitset.

db/memtable.cc
```diff
@@ -842,6 +846,66 @@ bool MemTable::Get(const LookupKey& key, std::string* value, Status* s,
...
+bool MemTable::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
+                        ReadCallback* callback, bool* is_blob) {
+  // The sequence number is updated synchronously in version_set.h
+  if (IsEmpty()) {
+    // Avoiding recording stats for speed.
+    return false;
+  }
+  std::bitset<MultiGetContext::MAX_BATCH_SIZE> keys;
+  int idx = 0;
+  for (auto iter = range->begin(); iter != range->end(); ++iter) {
```
anand1976: Create a new MultiGetRange from the input range parameter. Then use the new range in this function, as well as pass it down to ```bloom_filter_->MayContain()```. It will allow you to simplify the rest of the function, as explained in the following comments.

Using a new range here also ensures that the original range remains intact, and can be used for subsequent stages of ```DB::MultiGet```.

db/db_impl/db_impl.cc
```diff
@@ -1959,41 +1959,58 @@ void DBImpl::MultiGetImpl(
...
-        } else if (super_version->imm->Get(
-                       *(mget_iter->lkey), value->GetSelf(), &s, &merge_context,
-                       &mget_iter->max_covering_tombstone_seq, read_options,
-                       callback, is_blob_index)) {
-          done = true;
-          value->PinSelf();
-          RecordTick(stats_, MEMTABLE_HIT);
+    bool skip_memtable = (read_options.read_tier == kPersistedTier && has_unpersisted_data_.load(std::memory_order_relaxed));
+    if (!skip_memtable) {
+        bool found_all_keys = super_version->mem->MultiGet(read_options, &range, callback, is_blob_index);
```
anand1976: ```Memtable::MultiGet``` doesn't need to return found_all_keys. You can simply call ```range.empty()``` to check if any keys remain to be found.

db/memtable.cc
```diff
@@ -842,6 +826,105 @@ bool MemTable::Get(const LookupKey& key, std::string* value, Status* s,
   return found_final_value;
 }
 
+void MemTable::GetFromTable(const LookupKey& key, std::string* value, Status* s,
```
anand1976: Put all input arguments before output arguments as per convention. I think key, max_covering_tombstone_seq and callback are the input arguments. Maybe pass max_covering_tombstone_seq by value so its clear that its an input argument.

db/memtable.cc
```diff
@@ -842,6 +826,105 @@ bool MemTable::Get(const LookupKey& key, std::string* value, Status* s,
...
+    int num_keys = 0;
+    for (auto iter = temp_range.begin(); iter != temp_range.end(); ++iter) {
+      if (!prefix_extractor_) {
+        keys[num_keys++] = &iter->ukey;
+      } else if (prefix_extractor_->InDomain(iter->ukey)) {
+        prefixes.emplace_back(prefix_extractor_->Transform(iter->ukey));
+        keys[num_keys++] = &prefixes.back();
+      } else {
+        temp_range.SkipKey(iter);
+        RecordTick(moptions_.statistics, MEMTABLE_MISS);
```
anand1976: Update the MEMTABLE_MISS stat in DBImpl::MultiGetImpl if it has to look in the SST files for a key. It may be a miss here, but may be found in the immutable memtables.

db/memtable.cc
```diff
@@ -842,6 +826,105 @@ bool MemTable::Get(const LookupKey& key, std::string* value, Status* s,
...
+      }
+    }
+    bloom_filter_->MayContain(num_keys, &keys[0], &may_match[0]);
+    int idx = 0;
+    for (auto iter = temp_range.begin(); iter != temp_range.end(); ++iter) {
+      if (!may_match[idx]) {
+        temp_range.SkipKey(iter);
+        PERF_COUNTER_ADD(bloom_memtable_miss_count, 1);
+        RecordTick(moptions_.statistics, MEMTABLE_MISS);
```
anand1976: Don't update MEMTABLE_MISS here.

db/memtable.cc
```diff
@@ -842,6 +826,105 @@ bool MemTable::Get(const LookupKey& key, std::string* value, Status* s,
...
+      } else {
+        PERF_COUNTER_ADD(bloom_memtable_hit_count, 1);
+      }
+      idx++;
+    }
+  }
+  for (auto iter = temp_range.begin(); iter != temp_range.end(); ++iter) {
```
anand1976: The temp_range will skip keys that are not in the prefix extractor domain. But we don't want that. For such keys, we only skip bloom filter check but do the actual memtable check.

db/db_impl/db_impl.cc
```diff
@@ -1961,41 +1961,24 @@ void DBImpl::MultiGetImpl(
...
+      mget_iter->merge_context.Clear();
+      *mget_iter->s = Status::OK();
+    }
 
-      bool skip_memtable =
-          (read_options.read_tier == kPersistedTier &&
-           has_unpersisted_data_.load(std::memory_order_relaxed));
-      bool done = false;
-      if (!skip_memtable) {
-        if (super_version->mem->Get(*(mget_iter->lkey), value->GetSelf(), &s,
```
anand1976: I'm wondering if, for the sake of consistency, we should also change Memtable::Get() to take a PinnableSlice for value instead of std::string, i.e move this operation of value->GetSelf() into Memtable::Get().

vjnadimpalli: Sounds good..maybe do it in a separate PR.

db/db_basic_test.cc
```diff
@@ -1380,9 +1381,17 @@ TEST_P(MultiGetPrefixExtractorTest, Batched) {
   ASSERT_OK(Put("kk2", "v2"));
   ASSERT_OK(Put("kk3", "v3"));
   ASSERT_OK(Put("kk4", "v4"));
+  std::vector<std::string> keys({"k", "kk1", "kk2", "kk3", "kk4"});
```
anand1976: Add some non-existent keys here to test bloom filter -ve results.

db/db_basic_test.cc
```diff
@@ -1380,9 +1381,17 @@ TEST_P(MultiGetPrefixExtractorTest, Batched) {
...
+  std::vector<std::string> inmem_values;
+  inmem_values = MultiGet(keys, nullptr);
+  ASSERT_EQ(inmem_values[0], "v0");
+  ASSERT_EQ(inmem_values[1], "v1");
+  ASSERT_EQ(inmem_values[2], "v2");
+  ASSERT_EQ(inmem_values[3], "v3");
+  ASSERT_EQ(inmem_values[4], "v4");
+
```
anand1976: It would be good to also verify the stats.

anand1976:
@vjnadimpalli One more minor comment - in the PR description, would be good to avoid links to FB internal posts. Instead, explicitly mention the relevant information.

PR #5822, opened 2019-09-18
Refactor deletefile_test.cc

riversand963:
Make DeleteFileTest inherit DBTestBase to avoid code duplication.

Test Plan (on devserver)
```
$make deletefile_test
$./deletefile_test
```


riversand963:
Here is a similar one to #5820. Could you take a look as well? Thanks! @vjnadimpalli !

PR #5823, opened 2019-09-18
Correctly handle Windows FS that does not support hard links

darinpp:
Ingesting a file from a file system that doesn't
support hard links or from a network location, returns
an error that wasn't handled correctly and the result was
IO error instead of Not Supported status.

Fixes cockroachdb/cockroach#38789


maysamyabandeh:
@hliu18 @burtonli  Can you please review the patch?

rcane:
Can this PR be merged please since I just ran into this bug?

PR #5825, opened 2019-09-18
filter_bench - a prelim tool for SST filter benchmarking

pdillinger:
Summary: Example: using the tool before and after PR #5784 shows that
the refactoring, presumed performance-neutral, actually sped up SST
filters by about 3% to 8% (repeatable result):

Before:
-  Dry run ns/op: 22.4725
-  Single filter ns/op: 51.1078
-  Random filter ns/op: 120.133

After:
+  Dry run ns/op: 22.2301
+  Single filter run ns/op: 47.4313
+  Random filter ns/op: 115.9

Only tests filters for the block-based table (full filters and
partitioned filters - same implementation; not block-based filters),
which seems to be the recommended format/implementation.

### DIFF
--- a/util/filter_bench.cc
+++ b/util/filter_bench.cc
@@ -2,10 +2,6 @@
 //  This source code is licensed under both the GPLv2 (found in the
 //  COPYING file in the root directory) and Apache 2.0 License
 //  (found in the LICENSE.Apache file in the root directory).
-//
-// Copyright (c) 2011 The LevelDB Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE file. See the AUTHORS file for names of contributors.
 
 #ifndef GFLAGS
 #include <cstdio>
@@ -56,12 +52,16 @@ DEFINE_bool(quick, false, "Run more limited set of tests, fewer queries");
 
 DEFINE_bool(allow_bad_fp_rate, false, "Continue even if FP rate is bad");
 
+DEFINE_bool(legend, false,
+            "Print more information about interpreting results instead of "
+            "running tests");
+
 void _always_assert_fail(int line, const char *file, const char *expr) {
   fprintf(stderr, "%s: %d: Assertion %s failed\n", file, line, expr);
   abort();
 }
 
-#define always_assert(cond) \
+#define ALWAYS_ASSERT(cond) \
   ((cond) ? (void)0 : ::_always_assert_fail(__LINE__, __FILE__, #cond))
 
 using rocksdb::BlockContents;
@@ -85,7 +85,7 @@ struct KeyMaker {
   Slice slice_;
   uint32_t *vals_;
 
-  Slice &Get(uint32_t filter_num, uint32_t val_num) {
+  Slice Get(uint32_t filter_num, uint32_t val_num) {
     vals_[0] = filter_num + val_num;
     vals_[1] = val_num;
     return slice_;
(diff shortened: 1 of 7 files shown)


pdillinger:
(fixed formatting with force-push)

pdillinger:
(minor fix with unique_ptr and rebase to latest master)

util/filter_bench.cc
```diff
@@ -0,0 +1,403 @@
...
+  void Go();
+
+  void RandomQueryTest(bool inside, bool dry_run, TestMode mode);
+};
+
+void FilterBench::Go() {
+  std::unique_ptr<rocksdb::FilterBitsBuilder> builder(
+      policy->GetFilterBitsBuilder());
+
+  uint32_t varianceMask = 1;
```
siying: We follow Google C++ Style and the naming convention for variables is like `variance_mask`: https://google.github.io/styleguide/cppguide.html#Variable_Names

util/filter_bench.cc
```diff
@@ -0,0 +1,403 @@
...
+DEFINE_bool(quick, false, "Run more limited set of tests, fewer queries");
+
+DEFINE_bool(allow_bad_fp_rate, false, "Continue even if FP rate is bad");
+
+void _always_assert_fail(int line, const char *file, const char *expr) {
+  fprintf(stderr, "%s: %d: Assertion %s failed\n", file, line, expr);
+  abort();
+}
+
+#define always_assert(cond) \
```
siying: I don't find a rule but macro seems to be all capital in this code base, like `ALWAYS_ASSERT`

util/filter_bench.cc
```diff
@@ -0,0 +1,403 @@
...
+void FilterBench::RandomQueryTest(bool inside, bool dry_run, TestMode mode) {
+  for (auto &info : infos) {
+    info.outside_queries_ = 0;
+    info.false_positives_ = 0;
+  }
+
+  uint32_t dryRunHash = 0;
+  uint64_t maxQueries = static_cast<uint64_t>(FLAGS_m_queries) * 1000000;
+  size_t numPrimaryFilters = infos.size();
+  uint32_t primaryFilterThreshold = 0xffffffff;
```
siying: Can you add some comments to the function to explain how it works? Also the variables. What does "primary filter" mean is not clear when I read up to here.

util/filter_bench.cc
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  rocksdb::StopWatchNano timer(rocksdb::Env::Default(), true);
+
+  for (uint64_t q = 0; q < maxQueries; q += batchSize) {
+    uint32_t filterIndex;
+    if (random() <= primaryFilterThreshold) {
+      filterIndex = fastrange32(numPrimaryFilters, random());
+    } else {
+      // secondary
```
siying: Need to explain in comment somehere what is primary and what is secondary.

util/filter_bench.cc
```diff
@@ -0,0 +1,403 @@
...
+      std::cout << "    Best possible bits/key: "
+                << -std::log((double)fp / q) / std::log(2.0) << std::endl;
+    }
+  }
+}
+
+int main(int argc, char **argv) {
+  rocksdb::port::InstallStackTraceHandler();
+  SetUsageMessage(std::string("\nUSAGE:\n") + std::string(argv[0]) +
+                  " [OPTIONS]...");
```
siying: It might be a good idea to mention more usage here. How to use the benchmark tool is not very intuitive to me.

Anyway, we need to document somewhere the main usage.

pdillinger: You can't currently pick between different implementations, so you literally just run it. I use -quick most often (I'll add that to USAGE) but no other options seem critical to mention. Thanks to gflags, -help tells you about all the options. I'll add some info about interpreting results.

util/filter_bench.cc
pdillinger: "assert" excepted ;)

util/filter_bench.cc
pdillinger: And I think it's too niche for going on the wiki, for now. Perhaps it would make sense when/if we expose new user options to pick between SST filter implementations.

util/filter_bench.cc
```diff
@@ -0,0 +1,403 @@
...
+  uint64_t false_positives_ = 0;
+};
+
+enum TestMode {
+  TM_SingleFilter,
+  TM_BatchPrepared,
+  TM_BatchUnprepared,
+  TM_FiftyOneFilter,
+  TM_EightyTwentyFilter,
+  TM_RandomFilter,
```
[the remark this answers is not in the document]
pdillinger: I changed to kSingleFilter per convention

util/filter_bench.cc
```diff
@@ -0,0 +1,403 @@
...
+
+  std::cout << "Building..." << std::endl;
+
+  size_t totalMemoryUsed = 0;
+  size_t totalKeysAdded = 0;
+
+  rocksdb::StopWatchNano timer(rocksdb::Env::Default(), true);
+
+  while (totalMemoryUsed < 1024 * 1024 * FLAGS_working_mem_size_mb) {
+    uint32_t filterId = random();
```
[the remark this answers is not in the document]
pdillinger: I want 32-bit (making Random64 inconvenient, possibly slower), and I don't trust "a very simple random number generator" anywhere near something evaluating FP rates. ;)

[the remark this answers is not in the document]
pdillinger: Declined, unless you insist.

pdillinger:
Appveyor appears to be success but not propagated here.

util/filter_bench.cc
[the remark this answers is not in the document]
siying: I am a little bit concerned about portability of random() function. Do you know it is supposed to be widely available everywhere? If that is the case, it's OK then.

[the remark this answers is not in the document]
pdillinger: > $ git grep mt19937
db/forward_iterator_bench.cc:    std::mt19937 rng{std::random_device()()};
db/forward_iterator_bench.cc:  std::mt19937 rng{std::random_device()()};
tools/write_stress.cc:    std::mt19937 rng(static_cast<unsigned int>(FLAGS_seed));
tools/write_stress.cc:    auto random_string = [](std::mt19937& r, int len) {
tools/write_stress.cc:    std::mt19937 rng(static_cast<unsigned int>(FLAGS_seed));
util/heap_test.cc:  std::mt19937 rng(static_cast<unsigned int>(RNG_SEED));
util/random.h:// A simple 64bit random number generator based on std::mt19937_64
util/random.h:  std::mt19937_64 generator_;

PR #5829, opened 2019-09-19
Code comment for Version Edit

vjnadimpalli:
Added comment for Version Edit. 
Test plan:
Run existing tests

### DIFF
--- a/db/version_edit.h
+++ b/db/version_edit.h
@@ -191,11 +191,10 @@ struct LevelFilesBrief {
   }
 };
 
-// A certain state of RocksDB at any given time is referred to as a version (aka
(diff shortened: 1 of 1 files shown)


db/version_edit.h
```diff
@@ -191,6 +191,10 @@ struct LevelFilesBrief {
   }
 };
 
+// The state of a DB at any given time is referred to as a Version.
```
riversand963: Version does not include memtables.

vjnadimpalli: Right thats indirectly already documented at the Version class. Here is the comment for Version '// A column family's version consists of the SST files owned by the column
// family at a certain point in time.'

PR #5832, opened 2019-09-19
Add class comment for Block

ltamasi:


table/block_based/block.h
```diff
@@ -135,6 +135,18 @@ class BlockReadAmpBitmap {
   uint32_t rnd_;
 };
 
+// This Block class is not for any old block: it is designed to hold only
```
maysamyabandeh: I am confused by "any old block".

Issue #4277, opened 2018-08-15
Raspberry pi 3 build error.


jing5877:
### Expected behavior
builds sucessfully

### Actual behavior
Generating target/apidocs/deprecated-list.html...
Building index for all classes...
Generating target/apidocs/allclasses-frame.html...
Generating target/apidocs/allclasses-noframe.html...
Generating target/apidocs/index.html...
Generating target/apidocs/overview-summary.html...
Generating target/apidocs/help-doc.html...
22 warnings
make[1]: Leaving directory '/home/pi/rocksdb-master/java'
In file included from /usr/include/c++/6/cassert:44:0,
                 from ./include/rocksdb/compaction_filter.h:12,
                 from ./java/./rocksjni/compaction_filter_factory_jnicallback.h:15,
                 from ./java/./rocksjni/portal.h:31,
                 from java/rocksjni/loggerjnicallback.cc:14:
java/rocksjni/loggerjnicallback.cc: In member function ‘virtual void rocksdb::LoggerJniCallback::Logv(rocksdb::InfoLogLevel, const char*, va_list)’:
java/rocksjni/loggerjnicallback.cc:134:15: error: invalid operands of types ‘va_list {aka __va_list}’ and ‘std::nullptr_t’ to binary ‘operator!=’
     assert(ap != nullptr);
            ~~~^~~~
Makefile:1855: recipe for target 'rocksdbjava' failed
make: *** [rocksdbjava] Error 1

### Steps to reproduce the behavior
I was build on a raspberry pi 3 (Raspbian GNU/Linux 9.4) ，use command make rocksdbjava.

yiwu-arbug:
@jing5877 Thank you for reporting! Are you interested in sending a PR for a fix?

jneedham83:
Just encountered on legacy (573) dbg build on aarch64

Yikun:
I also encountered this when I set DEBUG_LEVEL=1 (default) in my aarch64 enviroment with gcc 7.4.0.

Yikun:
I think we could remove this invalid assert directly to fix this problem, see pr: https://github.com/facebook/rocksdb/pull/5836

PR #5836, opened 2019-09-21
Remove invalid comparison of va_list and nullptr

Yikun:
The comparison of va_list and nullptr is always False under any arch, and will raise invalid operands of types error in aarch64 env (`error: invalid operands of types ‘va_list {aka __va_list}’ and ‘std::nullptr_t’ to binary ‘operator!=’`).

This patch removes this invalid assert.

Closes: https://github.com/facebook/rocksdb/issues/4277


Yikun:
cc @adamretter

This pr make the `make jtest` passed in arm64 environment.

Issue #5746, opened 2019-08-27
VS BUILD ERROR C2131


ananclub:
block_cache_analyzer\block_cache_trace_analyzer.cc	1212

ananclub:
vs 2017&2019 release has same error

not support dynamic array.

4>block_cache_trace_analyzer.cc
5>block_cache_trace_analyzer.cc
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(653,57): warning C4244:  “参数”: 从“uint64_t”转换到“unsigned int”，可能丢失数据
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(1212,24): error C2131:  表达式的计算结果不是常数
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(1212,24): message :  因读取超过生命周期的变量而失败
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(1212,24): message :  请参见“reuse_vector_size”的用法
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(1212,43): error C2131:  表达式的计算结果不是常数
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(1212,43): message :  因读取超过生命周期的变量而失败
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(1212,43): message :  请参见“reuse_vector_size”的用法
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(1216,37): error C3863:  不可指定数组类型“uint64_t [reuse_vector_size][reuse_vector_size]”
4>F:\project\rocksdb\tools\block_cache_analyzer\block_cache_trace_analyzer.cc(1260,1): error C2036:  “uint64_t [reuse_vector_size][reuse_vector_size]”: 未知的大小

riversand963:
Have you given it a shot to fixing these compilation errors according to the messages from the compiler? With out information about version number/commit hash, it is difficult for other people to match the warnings/errors to the exact line numbers if they cannot reproduce it.
I haven't been able to test this, but if you (or anybody else with access to VS build) can eliminate these errors, a PR will be welcome.

solicomo:
I think this is a dup to #5730.
[...]

PR #5838, opened 2019-09-22
Fix msvc build errors in block_cache_trace_analyzer.cc (#5746)

ccaughie:
Use a dynamically allocated array for reuse_table, since msvc does not
support stack-allocated arrays whose size is unknown at compile time.
Add a couple of explicit casts to avoid numeric conversion warnings.

N.B. I have not been able to test this change properly since I can't get the tests to run successfully using "make check" on Linux. Please let me know if there is a better way to do this.

### DIFF
--- a/tools/block_cache_analyzer/block_cache_trace_analyzer.cc
+++ b/tools/block_cache_analyzer/block_cache_trace_analyzer.cc
@@ -1211,8 +1211,8 @@ void BlockCacheTraceAnalyzer::WriteBlockReuseTimeline(
   TraverseBlocks(block_callback);
 
   // A cell is the number of blocks accessed in a reuse window.
-  const auto reuse_table =
-      std::make_unique<uint64_t[]>(reuse_vector_size * reuse_vector_size);
+  const std::unique_ptr<uint64_t[]> reuse_table(
+      new uint64_t[reuse_vector_size * reuse_vector_size]);
 
   for (uint64_t start_time = 0; start_time < reuse_vector_size; start_time++) {
     // Initialize the reuse_table.
--- a/tools/block_cache_analyzer/block_cache_trace_analyzer.cc
+++ b/tools/block_cache_analyzer/block_cache_trace_analyzer.cc
@@ -14,6 +14,7 @@
 #include <fstream>
 #include <iomanip>
 #include <iostream>
+#include <memory>
 #include <random>
 #include <sstream>
 
--- a/tools/block_cache_analyzer/block_cache_trace_analyzer.cc
+++ b/tools/block_cache_analyzer/block_cache_trace_analyzer.cc
@@ -650,7 +650,8 @@ void BlockCacheTraceAnalyzer::WriteCorrelationFeaturesToFile(
     const std::map<std::string, Features>& label_features,
     const std::map<std::string, Predictions>& label_predictions,
     uint32_t max_number_of_values) const {
-  std::default_random_engine rand_engine(env_->NowMicros());
+  std::default_random_engine rand_engine(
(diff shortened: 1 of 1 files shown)


tools/block_cache_analyzer/block_cache_trace_analyzer.cc
```diff
@@ -14,6 +14,7 @@
 #include <fstream>
 #include <iomanip>
 #include <iostream>
+#include <memory>
```
riversand963: `<memory>` is already included by including other headers.

tools/block_cache_analyzer/block_cache_trace_analyzer.cc
```diff
@@ -1209,11 +1211,14 @@ void BlockCacheTraceAnalyzer::WriteBlockReuseTimeline(
   TraverseBlocks(block_callback);
 
   // A cell is the number of blocks accessed in a reuse window.
-  uint64_t reuse_table[reuse_vector_size][reuse_vector_size];
+  const std::unique_ptr<uint64_t[]> reuse_table(
```
riversand963: Curious, MSVC forbids two-dimensional arrays now? Can @hliu18 or @burtonli verify this behavior?

hliu18: MSVC supports two-dimensional arrays, but does not support variable length arrays, i.e. dimension sizes need to be known at compile time. Variable length array is not part of recent versions of C++ standard.

riversand963: Will the following compile and run on MSVC?
```
#include <cstdio>

  const size_t sz = argc;
  int test_arr[sz];
  fprintf(stdout, "size is %d\n", sizeof(test_arr));
  return 0;
}
```
If not, this change is necessary.
It probably won't, according to your previous comment about compile-time.

hliu18: It does not compile: "error C2131: expression did not evaluate to a constant"
The compiler complains about [sz].

riversand963: Thanks @hliu18 for verifying this.

tools/block_cache_analyzer/block_cache_trace_analyzer.cc
ccaughie: You're right, I originally thought that was why the xcode build was failing. But it turns out xcode clang doesn't like std::make_unique at all, at least with the compiler options that are being used here.

riversand963: Does this mean we can use `#if defined()` to enclose this include?

ccaughie: I would just remove it altogether. If I had looked more closely I would have realized that std::unique_ptr is already used in several other places in this file, so including <memory> is clearly not necessary for this new use of it.
Alternatively you could keep it in; some people would argue that each file should explicitly include the headers for what it uses so that it is not dependent on other header files including them first. All of those headers will have include guards so the impact on compile time should be negligible.

tools/block_cache_analyzer/block_cache_trace_analyzer.cc
hliu18: I see this PR was imported but not checked in due to conflicts. Can someone help move this PR forward? I was going to fix a couple of build break on MSVC and noticed this.

riversand963:
Fixed by https://github.com/facebook/rocksdb/pull/6081

PR #5841, opened 2019-09-23
Optimize inserting into memtable with multiple thread.

Little-Wallace:
## Design

* Although RocksDB could use `allow_concurrent_memtable_write` to allow write-thread to insert data into memtable together, which belong to one group. However, if one thread call `Write` while there is still one leader inserting into memtable , it must wait for new leader to pull it into another write-group.
 * `unordered_write` has almost solved this problem. But there is flaw, that is, no longer provide a consistent snapshot read.
 * So I have one new interface `MultiThreadWrite`,  which allow customer to write multiple `WriteBatch` at once.  Every thread could help leader to insert into memtable when they wait for next leader. Of course, it provides **snapshot read**.
* I also add a `ThreadPool` to help write thread for some special considerations. For example, application with RocksDB may have multiple write-thread, and they are not balanced. And this application must write data into RocksDB in order, because it is a system requiring linear consistency, such as `Raft` or `Paxos`.
* use `enable_multi_thread_write` to enable this feature.

## Bench

There is the script of benchmark.
```
mkdir -p data
rm -rf data/*

./db_bench --db=./data --disable_wal=true --enable_pipelined_write=true --use_multi_thread_write=true --memtable_write_pool_size=0 --key_size=20 --value_size=100 --write_buffer_size=4096000000  --benchmarks=fillrandom --batch_size=512 --num=1000000 --threads=6
rm -rf data/*
./db_bench --db=./data --disable_wal=true --enable_pipelined_write=true --key_size=20 --value_size=100 --write_buffer_size=4096000000  --benchmarks=fillrandom --batch_size=512 --num=1000000 --threads=6
rm -rf data/*
./db_bench --db=./data --disable_wal=true --enable_pipelined_write=false --unordered_write=true --key_size=20 --value_size=100 --write_buffer_size=4096000000  --benchmarks=fillrandom --batch_size=512 --num=1000000 --threads=6
```

option\thread | 6 |   12  |  24
-|-|-|-
multi_thread_write | 214 MB/s | 361 MB/s | 524 MB/s |
pipelined_write | 131 MB/s | 191 MB/s | 285 MB/s |
unordered_write | 220MB/s | 383 MB/s | 545 MB/s |

## Conclusion

* Obliviously, multi_thread_write could have similar performance compare to unordered_write, and both of them are much better than pipelined_write.


maysamyabandeh:
This is a very interesting direction. Thanks for looking into this. I am still not sure what is the gist of the design and where the efficiency is coming from. But in the meanwhile I gave the patch a try with the same benchmark that was provided in unordered_write PR. The results that I get are quite different. Have you tried these benchmarks? Do you where the discrepancy is coming from?
```
conf/wal                    WAL          no-WAL
base                          275           445  MB/s
unordered_write              439           1135
pull                          70              71
```
```
TEST_TMPDIR=/dev/shm/ ./db_bench_pull --benchmarks=fillrandom --threads=32 --num=10000000 -max_write_buffer_number=16 --max_background_jobs=64 --batch_size=8 --writes=3000000 -level0_file_num_compaction_trigger=99999 --level0_slowdown_writes_trigger=99999 --level0_stop_writes_trigger=99999  -disable_auto_compactions  --unordered_write=0 --enable_pipelined_write=true --use_multi_thread_write=true --memtable_write_pool_size=0 --disable_wal=1
```

yiwu-arbug:
@Little-Wallace may have more insight, but my understanding is the patch is useful in scenario when there's only few threads concurrent writing (e.g. 6 threads) but with large write batch. Due to the way write group is formed, for the default write impl, a write group may consist of fewer writer than the number of user write threads, so the concurrency is lower than expected; while with the patch, we can simply set a larger write thread pool size to achieve higher memtable write concurrency.

For our use case we have relatively large write batch and we need atomic write for the whole batch, so it's difficult to split the batch into several and concurrent write from multiple user threads.

maysamyabandeh:
Thanks @yiwu-arbug So I understand that the patch is to optimize for large batch_size 512.
Have you tried running the benchmark with --transaction_db which provide immutable snapshot. When I do I get results similar to what the patch does.
```
24t+nowal+unordered_write          743 MB/s
24t+nowal+unordered_write+txn      676 MB/s
24t+nowal+patch                    681 MB/s
```
```
TEST_TMPDIR=/dev/shm/ ./db_bench_pull --disable_wal=true --enable_pipelined_write=false --unordered_write=true --key_size=20 --value_size=100 --write_buffer_size=4096000000  --benchmarks=fillrandom --batch_size=512 --num=1000000 --threads=24

TEST_TMPDIR=/dev/shm/ ./db_bench_pull --disable_wal=true --enable_pipelined_write=false --unordered_write=true --key_size=20 --value_size=100 --write_buffer_size=4096000000  --benchmarks=fillrandom --batch_size=512 --num=1000000 --threads=24 --transaction_db

TEST_TMPDIR=/dev/shm/ ./db_bench_pull --disable_wal=true --enable_pipelined_write=true --use_multi_thread_write=true --memtable_write_pool_size=0 --key_size=20 --value_size=100 --write_buffer_size=4096000000  --benchmarks=fillrandom --batch_size=512 --num=1000000 --threads=24
```

Little-Wallace:
Thanks @maysamyabandeh for your attention.
Emmm， I means that this pr would perform better than normal pipelined-write. But it is still slightly slower than unordered-write. 
What I have optimized is just normal write, and it could provide consistant read. In our project , unordered-write would make our development more difficult, so we use pipelined-write until now.  I thought there may be some developer who want more efficiency pipelined-write.

maysamyabandeh:
Did you try the benchmark with --transaction_db which does provide immutable snapshot?

Little-Wallace:
Thanks your suggestion, I have tried it. 
I agree that the write performance of transaction_db is pretty good.  It is amazing for me.
But there is still other reason which make us not able to use transaction_db this feature, such as it cost more time on `RandomSeek`, and usage of the interface of transaction_db may be a little different from normal DB.  
In the future, we  would try transaction_db in our project,  but we also hope improve write performance with normal DB.

maysamyabandeh:
Thanks for trying it out. If you look at db_bench_tool.cc, when you pass --transaction_db it opens it with TransactionDB (with certain configs) but then assigns the pointer to DB*, meaning that in the rest of the benchmark only DB:: interfaces are used. This is actually number 1 question in the FAQ of write prepared wiki page: https://github.com/facebook/rocksdb/wiki/unordered_write#faq
If you do see a problem with `RandomSeek` with `TransactionDB` please do bring it up and we shall investigate.

yiwu-arbug:
@maysamyabandeh Do you know how does `TransactionDB` (after cast into `DB`) compare to vanilla `DBImpl` in general? Do you have users using `TransactionDB` in this way? Thanks.

yiwu-arbug:
@maysamyabandeh btw, does `TransactionDB` support writing without WAL? My understanding is it heavy rely on WAL, but want to double check with you.

Little-Wallace:
@maysamyabandeh  when `BenchTool` open one DB with transaction_db, and use `DB::NewIterator` rather than `Transaction::GetIterator` to seek one key, how it could provide consistency read?

maysamyabandeh:
@yiwu-arbug TransactionDB should be exactly the same as DB when used with the same API. WAL is mandatory when ::Prepare is called which is a TransactionDB specific API. When used with the standard DB API, WAL could be disabled similarly to when using DBImpl.

maysamyabandeh:
@Little-Wallace One need to define consistent reads to be more clear. When calling TransactionDB::NewIterator it provides exactly the same guarantees as DBImpl::NewIterator. So whatever your definition of consistency is, it will be compromised when switched to TransactionDB.

PR #5847, opened 2019-09-24
Add custom env support to ldb

riversand963:
Summary: add custom env support to ldb.

Test Plan (on devserver):
```
$make all && make check
```


riversand963:
Plan to merge to #5845

Issue #5878, opened 2019-10-03
RocksDB crash during ingest external file


koldat:
There is an issue that under some conditions ingest file tries to erase uninitialized iterator:
```
  std::list<uint64_t>::iterator pending_output_elem;
  size_t total = 0;
  for (const auto& arg : args) {
    total += arg.external_files.size();
  }
  uint64_t next_file_number = 0;
  Status status = ReserveFileNumbersBeforeIngestion(
      static_cast<ColumnFamilyHandleImpl*>(args[0].column_family)->cfd(), total,
      &pending_output_elem, &next_file_number);
  if (!status.ok()) {
    !!!HERE WE CAN END UP WITH NOT INITIALIZED ITERATOR!!!
    InstrumentedMutexLock l(&mutex_);
    ReleaseFileNumberFromPendingOutputs(pending_output_elem);
    return status;
  }
```

Why it happens? On some NFS databases we are hitting some OS race condition where ingest move file succeeds, but compaction job that is started immediately does not see that file yet. It looks like this:

```
2019/09/19-08:07:13.785002 7fc921d56700 [/external_sst_file_ingestion_job.cc:282] [AddFile] External SST file ./data/nass_archivedb/1.spool_1568879400.0.ingest.tmp was ingested in L1 with path ./data/nass_archivedb/000084.sst (global_seqno=182) {"table_properties": {"data_size": 48158985, "index_size": 730761, "top_level_index_size": 8630, "index_partitions": 178, "filter_size": 0, "raw_key_size": 100522972, "raw_average_key_size": 44, "raw_value_size": 431207811, "raw_average_value_size": 188, "num_data_blocks": 14856, "num_entries": 2284613, "filter_policy_name": ""}}
2019/09/19-08:07:13.785177 7fc93d803700 [/compaction_job.cc:1645] [default] [JOB 380] Compacting 1@1 + 1@2 files to L2, score 1.09
2019/09/19-08:07:13.785189 7fc93d803700 [/compaction_job.cc:1649] [default] Compaction start summary: Base version 137 Base level 1, inputs: [84(46MB)], [83(45MB)]
2019/09/19-08:07:13.785219 7fc93d803700 EVENT_LOG_v1 {"time_micros": 1568880433785200, "job": 380, "event": "compaction_started", "compaction_reason": "LevelMaxLevelSize", "files_L1": [84], "files_L2": [83], "score": 1.0863, "input_data_size": 95582278}
[...]

PR #5879, opened 2019-10-03
Fix crash when background task fails

koldat:
Fixing crash. Full story in issue: #5878


koldat:
I have updated HISTORY.md.

PR #5884, opened 2019-10-04
[RocksJava] Add missing table options

jay-zhuang:
Add TableOption IndexShorteningMode and kBinarySearchWithFirstKey.

Test Plan: make jtest and tested with Rocksandra


maysamyabandeh:
@adamretter Does it look good to you?

PR #5889, opened 2019-10-06
Remove deprecated RocksDBCommonHelper and cont_integration.sh

katherinez:
As titled. RocksDBCommonHelper contains references to legacy APIs.


anand1976:
@katherinez What is the replacement for these files? I don't think anyone in the team is very familiar with this code, so just checking.

PR #5890, opened 2019-10-07
Adding centos7 build script

l1x:
Just bumping version to latest for rocksdb and zstd


update_fork.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+git fetch upstream
+git checkout master
+git push origin master
```
riversand963: I believe this file is not necessary. You can easily use alias to achieve the same goal.

l1x: True, removing.

PR #5896, opened 2019-10-08
Fix small issues

jsteemann:
The individual commits in this PR should be self-explanatory.
All small and _very_ low-priority changes.

### DIFF
--- a/include/rocksdb/status.h
+++ b/include/rocksdb/status.h
@@ -451,12 +451,11 @@ class Status {
   std::string ToString() const;
 
  protected:
-  // A nullptr state_ (which is always the case for OK) means the message
-  // is empty, else state_ points to message.
-
   Code code_;
   SubCode subcode_;
   Severity sev_;
+  // A nullptr state_ (which is at least the case for OK) means the extra
+  // message is empty.
   const char* state_;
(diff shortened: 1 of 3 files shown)


include/rocksdb/status.h
```diff
@@ -307,14 +307,11 @@ class Status {
...
  private:
-  // A nullptr state_ (which is always the case for OK) means the message
-  // is empty.
-  // of the following form:
-  //    state_[0..3] == length of message
-  //    state_[4..]  == message
   Code code_;
   SubCode subcode_;
   Severity sev_;
+  // A nullptr state_ (which is always the case for OK) means the message
```
yiwu-arbug: Seems empty message is also used for NotFound status, e.g. https://github.com/facebook/rocksdb/blob/master/db/version_set.cc#L1800

jsteemann:
```suggestion
  // A nullptr state_ (which is at least the case for OK) means the extra message
```

jsteemann:
@yiwu-arbug : looks like the Ci tests succeeded except one of the builds on TravisCI, which hit the build timeout of 50 minutes. 
Is there any chance to get this PR merged? Thanks!

yiwu-arbug:
@jsteemann sorry, I'm not on the team any more. Maybe someone else could help.

PR #5900, opened 2019-10-10
Fix concurrent full purge and WAL recycling

ajkr:
We were removing the file from `log_recycle_files_` before renaming it
with `ReuseWritableFile()`. Since `ReuseWritableFile()` occurs outside
the DB mutex, it was possible for a concurrent full purge to sneak in
and delete the file before it could be renamed. Consequently, `SwitchMemtable()`
would fail and the DB would enter read-only mode.

The fix is to hold the old file number in `log_recycle_files_` until
after the file has been renamed. Full purge uses that list to decide
which files to keep, so it can no longer delete a file pending recycling.

Test Plan: new unit test

### DIFF
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -1383,6 +1383,7 @@ class DBImpl : public DB {
   Status ThrottleLowPriWritesIfNeeded(const WriteOptions& write_options,
                                       WriteBatch* my_batch);
 
+  // REQUIRES: mutex locked and in write thread.
   Status ScheduleFlushes(WriteContext* context);
 
   void MaybeFlushStatsCF(autovector<ColumnFamilyData*>* cfds);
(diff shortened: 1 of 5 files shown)


ajkr:
Quick question - are there known cases where full purge is still necessary? Would it be a bad idea to turn it off (in our use case)?

ajkr:
Feel free to disregard my question. I still think this is important to prevent rare and difficult to debug crashes for users of WAL recycling.

ajkr:
I still think this fix is important.

ajkr:
Rotate reviewers - I recall @riversand963 knows about file purging.

riversand963:
Oops, just noticed this. Let me take a look today.

ajkr:
> LGTM. Thanks @ajkr for fixing this.
> With this change, calling `SwitchMemtable()` not in the single write thread can lead to two threads trying to rename the same WAL. This is not possible within RocksDB, but there is a `TEST_SwitchMemtable()` function which people may call for testing purposes and may forget this. Maybe I am being paranoid, do you think we should add a note or assertion somewhere to guard against this case?

I wonder if we could alternatively make `TEST_SwitchMemtable()` conform to the requirements of calling `SwitchMemtable()`, in particular, the former should only call the latter as a write group leader. I need to review a bit but think calling `write_thread_.EnterUnbatched()` before the memtable switch might work.

ajkr:
Thanks for the review, @riversand963. I tried making sure all callers of `SwitchMemtable()` (including test callers) are in the serialized write thread. Tests pass, so ready for another look.

PR #5904, opened 2019-10-11
Revert "Enable partitioned index/filter in stress tests (#5895)"

riversand963:
This reverts commit 2f4e288143f1c0c9f5a9e829d1b20ce62aca4224.
We revert it since a failure in our stress tests is suspected to be started after this commit.


riversand963:
Reverting #5895 and see if the test failures go away.

PR #5905, opened 2019-10-11
OnTableFileCreationCompleted use "(nil)" for empty file during flush

riversand963:
Compaction can call OnTableFileCreationCompleted(). If file is empty, "(nil)"
is used as the file name.
Do the same for flush.

Test plan (dev server):
```
make all
make check
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -23,7 +23,7 @@
 * Added DisableManualCompaction/EnableManualCompaction to stop and resume manual compaction.
 * Add TryCatchUpWithPrimary() to StackableDB in non-LITE mode.
 * Add a new Env::LoadEnv() overloaded function to return a shared_ptr to Env.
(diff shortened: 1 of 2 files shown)


HISTORY.md
```diff
@@ -23,6 +23,7 @@
 * Added DisableManualCompaction/EnableManualCompaction to stop and resume manual compaction.
 * Add TryCatchUpWithPrimary() to StackableDB in non-LITE mode.
 * Add a new Env::LoadEnv() overloaded function to return a shared_ptr to Env.
+* Flush sets file name to "(nil)" for OnTableFileCreationCompleted() if the flush-generated file is empty.
```
al13n321: I think it's not as important that the file was empty as that it's deleted. Like, after OnTableFileCreationCompleted() you would expect the file with the given name to exist and be part of the DB (at least for some time), but in this case there's no file.

So, maybe "if the flush produced no file (e.g. if the memtable was empty)"?

al13n321:
Lgtm.

PR #5924, opened 2019-10-15
Fix clang analyzer error

riversand963:
Summary:
Without this PR, clang analyzer complains.
```
$USE_CLANG=1 make analyze
db/compaction/compaction_job_test.cc:161:20: warning: The left operand of '==' is a garbage value
      if (key.type == kTypeBlobIndex) {
                ~~~~~~~~ ^
                1 warning generated.
```

Test Plan (on devserver)
```
$USE_CLANG=1 make analyze
```

### DIFF
--- a/db/compaction/compaction_job_test.cc
+++ b/db/compaction/compaction_job_test.cc
@@ -140,7 +140,7 @@ class CompactionJobTest : public testing::Test {
       std::string skey;
       std::string value;
       std::tie(skey, value) = kv;
-      ParseInternalKey(skey, &key);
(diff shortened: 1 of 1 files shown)


riversand963:
Thanks for taking a look @ltamasi. Seems the internal test failure is a legit one. Would you mind taking a look? Thanks

ltamasi:
Seems to me the only thing that's failing is the `SimpleCorrupted` test case, which intentionally corrupts some keys. So we should probably check that the key can be successfully parsed instead of `assert`ing it.

PR #5936, opened 2019-10-17
Store the filter bits reader alongside the filter block contents

ltamasi:
Summary:
Amongst other things, PR #5504 refactored the filter block readers so that
only the filter block contents are stored in the block cache (as opposed to the
earlier design where the cache stored the filter block reader itself, leading to
potentially dangling pointers and concurrency bugs). However, this change
introduced a performance hit since with the new code, the metadata fields are
re-parsed upon every access. This patch reunites the block contents with the
filter bits reader to eliminate this overhead; since this is still a self-contained
pure data object, it is safe to store it in the cache. (Note: this is similar to how
the zstd digest is handled.)

Test Plan:
make asan_check

filter_bench results for the old code:

```
$ ./filter_bench -quick
WARNING: Assertions are enabled; benchmarks unnecessarily slow
Building...
Build avg ns/key: 26.7153
Number of filters: 16669
Total memory (MB): 200.009
Bits/key actual: 10.0647
----------------------------
Inside queries...
  Dry run (46b) ns/op: 33.4258
  Single filter ns/op: 42.5974
  Random filter ns/op: 217.861
----------------------------
Outside queries...
  Dry run (25d) ns/op: 32.4217
  Single filter ns/op: 50.9855
  Random filter ns/op: 219.167
    Average FP rate %: 1.13993
----------------------------
Done. (For more info, run with -legend or -help.)

$ ./filter_bench -quick -use_full_block_reader
WARNING: Assertions are enabled; benchmarks unnecessarily slow
Building...
Build avg ns/key: 26.5172
Number of filters: 16669
Total memory (MB): 200.009
Bits/key actual: 10.0647
----------------------------
Inside queries...
  Dry run (46b) ns/op: 32.3556
  Single filter ns/op: 83.2239
  Random filter ns/op: 370.676
----------------------------
Outside queries...
  Dry run (25d) ns/op: 32.2265
  Single filter ns/op: 93.5651
  Random filter ns/op: 408.393
    Average FP rate %: 1.13993
----------------------------
Done. (For more info, run with -legend or -help.)
```

With the new code (see the reduction in overhead when the full filter block reader is used):

```
$ ./filter_bench -quick
WARNING: Assertions are enabled; benchmarks unnecessarily slow
Building...
Build avg ns/key: 25.4285
Number of filters: 16669
Total memory (MB): 200.009
Bits/key actual: 10.0647
----------------------------
Inside queries...
  Dry run (46b) ns/op: 31.0594
  Single filter ns/op: 43.8974
  Random filter ns/op: 226.075
----------------------------
Outside queries...
  Dry run (25d) ns/op: 31.0295
  Single filter ns/op: 50.3824
  Random filter ns/op: 226.805
    Average FP rate %: 1.13993
----------------------------
Done. (For more info, run with -legend or -help.)

$ ./filter_bench -quick -use_full_block_reader
WARNING: Assertions are enabled; benchmarks unnecessarily slow
Building...
Build avg ns/key: 26.5308
Number of filters: 16669
Total memory (MB): 200.009
Bits/key actual: 10.0647
----------------------------
Inside queries...
  Dry run (46b) ns/op: 33.2968
  Single filter ns/op: 58.6163
  Random filter ns/op: 291.434
----------------------------
Outside queries...
  Dry run (25d) ns/op: 32.1839
  Single filter ns/op: 66.9039
  Random filter ns/op: 292.828
    Average FP rate %: 1.13993
----------------------------
Done. (For more info, run with -legend or -help.)
```


pdillinger:
The important bit of the performance results are going from 40ns overhead associated with best-case full block reader query (Inside Single filter, 83ns - 43ns), to 15ns (59ns-44ns).

table/block_based/full_filter_data.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "table/format.h"
+
+namespace rocksdb {
+
+class FilterBitsReader;
+class FilterPolicy;
+
+// The sharable/cachable part of the full filter.
+class FullFilterData {
```
pdillinger: I think this should be called FullFilterCacheData or something, since I don't see it having a use outside of CachableEntry.

table/block_based/full_filter_block.cc
```diff
@@ -220,7 +213,7 @@ void FullFilterBlockReader::PrefixesMayMatch(
 void FullFilterBlockReader::MayMatch(
     MultiGetRange* range, bool no_io, const SliceTransform* prefix_extractor,
     BlockCacheLookupContext* lookup_context) const {
-  CachableEntry<BlockContents> filter_block;
+  CachableEntry<FullFilterData> filter_block;
```
pdillinger: filter_block -> filter_cache_data?

table/block_based/full_filter_block.cc
```diff
@@ -158,7 +158,7 @@ bool FullFilterBlockReader::PrefixMayMatch(
 bool FullFilterBlockReader::MayMatch(
     const Slice& entry, bool no_io, GetContext* get_context,
     BlockCacheLookupContext* lookup_context) const {
-  CachableEntry<BlockContents> filter_block;
+  CachableEntry<FullFilterData> filter_block;
```
pdillinger: filter_block -> filter_cache_data?

table/block_based/block_based_table_reader.cc
```diff
@@ -93,12 +94,30 @@ class BlocklikeTraits<BlockContents> {
...
+};
+
 template <>
 class BlocklikeTraits<Block> {
  public:
   static Block* Create(BlockContents&& contents, SequenceNumber global_seqno,
                        size_t read_amp_bytes_per_bit, Statistics* statistics,
-                       bool /* using_zstd */) {
+                       bool /* using_zstd */,
+                       const FilterPolicy* /* filter_policy */) {
```
pdillinger: Extra param is unfortunate, but ...

Actually, I'm not a C++ guru, but is this a place where we could use parameter packs? https://en.cppreference.com/w/cpp/language/parameter_pack

That would also save us passing in null when not needed.

table/block_based/full_filter_data.h
ltamasi: `CachableEntry` is a resource handle that refers to an object that might or might *not* be in the cache (not the greatest name, although technically not a misnomer either; see `cachable_entry.h` for more details). Bottom line is that it's quite possible that the filter data it refers to is *not* in the cache; so I would rather keep the name.

table/block_based/full_filter_block.cc
ltamasi: Considering the above, how about `filter_data` ?

table/block_based/full_filter_data.h
pdillinger: Logically, it doesn't encapsulate any more data than the reader. It encapsulates additional implementation details related to table blocks--for block cache management purposes.

How about the name ParsedFullFilterBlock? It still holds an association with being a table block but is in a ready-to-use state.

ltamasi: I like `ParsedFullFilterBlock` for a variety of reasons, one being that we can keep calling the variables `block`.

table/block_based/block_based_table_reader.cc
ltamasi: Yeah, I think we could make this cleaner by perfect forwarding the ctor arguments. It would need to be done all the way down through the call stack though. I'll take a stab at it.

pdillinger: Recommended, but optional

table/block_based/parsed_full_filter_block.h
```diff
@@ -0,0 +1,40 @@
...
+ public:
+  ParsedFullFilterBlock(const FilterPolicy* filter_policy,
+                        BlockContents&& contents);
+  ~ParsedFullFilterBlock();
+
+  FilterBitsReader* filter_bits_reader() const {
+    return filter_bits_reader_.get();
+  }
+
+  // TODO: consider memory usage of FullFilterBitsReader
```
pdillinger: Potentially non-trivial for partitioned filters. Optional

table/block_based/full_filter_block.cc
```diff
@@ -261,7 +246,10 @@ void FullFilterBlockReader::MayMatch(
       filter_range.SkipKey(iter);
     }
   }
-  filter_bits_reader->MayMatch(num_keys, &keys[0], &may_match[0]);
+
+  assert(filter_block.GetValue()->filter_bits_reader());
```
pdillinger: Consider stuffing `filter_block.GetValue()->filter_bits_reader()` in a variable. And removing this assert, because it's redundant with an if in the same function just 24 lines away. And we immediately deref anyway.

table/block_based/full_filter_block.cc
```diff
@@ -168,16 +169,8 @@ bool FullFilterBlockReader::MayMatch(
...
-    assert(table());
-    assert(table()->get_rep());
-
-    std::unique_ptr<FilterBitsReader> filter_bits_reader(
-        table()->get_rep()->filter_policy->GetFilterBitsReader(
-            filter_block.GetValue()->data));
-    assert(filter_bits_reader != nullptr);
-
-    if (filter_bits_reader->MayMatch(entry)) {
+  if (filter_block.GetValue()->filter_bits_reader()) {
```
pdillinger: Consider stuffing `filter_block.GetValue()->filter_bits_reader()` in a local variable

table/block_based/block_based_table_reader.cc
ltamasi: As it turns out, that's a bit more involved change; I'll take care of it in a separate patch.

PR #5955, opened 2019-10-22
Fix the potential memory leak in trace_replay

zhichao-cao:
In the previous PR #5934 , in the while loop, if/else if is used without ending with else to free the object referenced by ra, it might cause potential memory leak (warning during compiling). Fix it by adding "else" at the end to handle other trace entry types and replacing raw pointer with unique_ptr to avoid potential memory leak

Test plan: pass make asan check, pass the USE_CLANG=1 TEST_TMPDIR=/dev/shm/rocksdb OPT=-g make -j64 analyze.

### DIFF
--- a/trace_replay/trace_replay.cc
+++ b/trace_replay/trace_replay.cc
@@ -315,11 +315,10 @@ Status Replayer::MultiThreadReplay(uint32_t threads_num) {
   ReadOptions roptions;
   uint64_t ops = 0;
   while (s.ok()) {
-    ReplayerWorkerArg* ra = new ReplayerWorkerArg;
+    std::unique_ptr<ReplayerWorkerArg> ra(new ReplayerWorkerArg);
     ra->db = db_;
     s = ReadTrace(&(ra->trace_entry));
     if (!s.ok()) {
-      delete ra;
       break;
     }
     ra->woptions = woptions;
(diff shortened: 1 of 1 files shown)


trace_replay/trace_replay.cc
```diff
@@ -341,7 +341,8 @@ Status Replayer::MultiThreadReplay(uint32_t threads_num) {
       thread_pool.Schedule(&Replayer::BGWorkIterSeekForPrev, ra, nullptr,
                            nullptr);
       ops++;
-    } else if (ra->trace_entry.type == kTraceEnd) {
+    } else {
+      // It is the case that ra->trace_entry.type == kTraceEnd
```
riversand963: Can you put a `assert(ra->trace_entry.type == kTraceEnd)` here?

zhichao-cao: Sure.

riversand963: Or maybe keep original `else if ()`, but add an extra `else` in which assert(false).
Another thing I suggest is using unique_ptr to guard `ra`. You can avoid the `delete ra` if break out of the loop. You can pass the pointer to `Schedule` by calling the `unique_ptr::release()` funciton.

zhichao-cao: You mean using unique_ptr and release the ownership when pass to Schedule, such that I can remove all the delete?

riversand963: Yes.

trace_replay/trace_replay.cc
```diff
@@ -329,26 +328,28 @@ Status Replayer::MultiThreadReplay(uint32_t threads_num) {
...
-      // It is the case that ra->trace_entry.type == kTraceEnd
-      assert(ra->trace_entry.type == kTraceEnd);
-
+    } else if (ra->trace_entry.type == kTraceEnd) {
       // Do nothing for now, free the ReplayerWorkerArg.
       // TODO: Add some validations later.
-      delete ra;
       break;
+    } else {
+      // Other trace entry types that are not replayed.
```
riversand963: Wait, I thought we should never reach here?

zhichao-cao: No....I have tested in one real trace, it reach there. There might be some other trace entry types that are tracked but not replayed. I'm thinking not to assert it but do continue.

zhichao-cao: @riversand963 Actually, it is safe to use continue. For example, some one implemented more tracing types in the tracer, but replay has not been developed yet. It will switch to "else", we should not stop replay. If the trace is broken and there is no KTraceEnd in the entry, ReadTrace the Status will be "Corrupted" and break out.

riversand963: Thanks for the clarification. If the program can reach this point, then we should not assert(false).

trace_replay/trace_replay.cc
```diff
@@ -329,23 +328,29 @@ Status Replayer::MultiThreadReplay(uint32_t threads_num) {
...
       ops++;
     } else if (ra->trace_entry.type == kTraceIteratorSeekForPrev) {
-      thread_pool.Schedule(&Replayer::BGWorkIterSeekForPrev, ra, nullptr,
-                           nullptr);
+      thread_pool.Schedule(&Replayer::BGWorkIterSeekForPrev, ra.release(),
+                           nullptr, nullptr);
       ops++;
     } else if (ra->trace_entry.type == kTraceEnd) {
-      // Do nothing for now.
+      // Do nothing for now, free the ReplayerWorkerArg.
```
riversand963: This comment does not seem to apply any more.

zhichao-cao: Thanks pointing out. Revised.

Issue #5848, opened 2019-09-24
`db_crashtest` verification fails on file ingestion heavy workload


ajkr:
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://www.facebook.com/groups/rocksdb.dev

When running super heavy file ingestion stress test, verifying the state of the database sometimes fails. This might be a problem with the stress test, or an actual problem with file ingestion, but either way it seems worth figuring out.

### Expected behavior

We'd like to see this message:

```
Crash-recovery verification passed :)
```

### Actual behavior

But instead sometimes we see a message like this:
```
Verification failed for column family 0 key 7341: Length of value read is not equal
Crash-recovery verification failed :(
2019/09/24-19:30:12  Starting database operations
2019/09/24-19:30:12  Starting verification
Verification failed :(
```

### Steps to reproduce the behavior

build `db_stress`, preferably with `DEBUG_LEVEL=0`, then run the below command. On my machine it fails in ~10 seconds.

```
$ python tools/db_crashtest.py blackbox --simple --compression_type=none --interval=3 --max_key=10000 --ingest_external_file_one_in=100 --nooverwritepercent=0 --ingest_external_file_width=10
```

ajkr:
I'm not able to add an assignee, but @sumeerbhola plans to investigate this.

sumeerbhola:
@petermattis found a bug in db_stress.cc where it was not acquiring a lock for a key being updated in TestPut(), if there was a preceding TestIngestExternalFile(), which allowed multiple threads to concurrently mutate the same key using TestPuts. I will send a PR to fix it.

PR #5958, opened 2019-10-23
Fix corruption with intra-L0 on ingested files

Little-Wallace:
## Problem Description

Our process was abort when it call `CheckConsistency`. And the information in  `stderr` show that "`L0 files seqno 3001491972 3004797440 vs. 3002875611 3004524421` ".  Here are the causes of the accident I investigated.

* RocksDB will call `CheckConsistency` whenever `MANIFEST` file is update. It will check sequence number interval of every file, except files which were ingested.
* When one file is ingested into RocksDB, it will be assigned the value of global sequence number, and the minimum and maximum seqno of this file are equal, which are both equal to global sequence number. 
* `CheckConsistency`  determines whether the file is ingested by whether the smallest and largest seqno of an sstable file are equal.
* If IntraL0Compaction picks one sst which was ingested just now and compacted it into another sst,  the `smallest_seqno` of this new file will be smaller than his `largest_seqno`.
    * If more than one ingested file was ingested before memtable schedule flush,  and they all compact into one new sstable file by `IntraL0Compaction`. The sequence interval of this new file will be included in the interval of the memtable.  So `CheckConsistency` will return a `Corruption`.
    * If a sstable was ingested after the memtable was schedule to flush, which would assign a larger seqno to it than memtable. Then the file was compacted with other files (these files were all flushed before the memtable) in L0 into one file. This compaction start before the flush job of memtable start,  but completed after the flush job finish. So this new file produced by the compaction (we call it s1) would have a larger interval of sequence number than the file produced by flush (we call it s2).  **But there was still some data in s1  written into RocksDB before the s2, so it's possible that some data in s2 was cover by old data in s1.** Of course, it would also make a `Corruption` because of overlap of seqno. There is the relationship of the files:
    > s1.smallest_seqno < s2.smallest_seqno < s2.largest_seqno  < s1.largest_seqno

So I skip pick sst file which was ingested in function `FindIntraL0Compaction ` 

## Reason

Here is my bug report: https://github.com/facebook/rocksdb/issues/5913

There are two situations that can cause the check to fail.

### First situation：
- First we ingest five external sst into Rocksdb, and they happened to be ingested in L0. and there had been some data in memtable, which make the smallest sequence number of memtable is less than which of sst that we ingest.

- If there had been one compaction job which compacted sst from L0 to L1, `LevelCompactionPicker` would trigger a `IntraL0Compaction` which would compact this five sst from L0 to L0. We call this sst A, which was merged from five ingested sst. 

- Then some data was put into memtable, and memtable was flushed to L0. We called this sst B.
- RocksDB check consistency , and find the `smallest_seqno` of B is  less than that of A and crash. Because A was merged from five sst, the smallest sequence number of it was less than the biggest sequece number of itself, so RocksDB could not tell if A was produce by ingested.

### Secondary situaion

- First we have flushed many sst in L0,  we call them [s1, s2, s3].

- There is an immutable memtable request to be flushed, but because flush thread is busy, so it has not been picked. we call it m1.  And at the moment, one sst is ingested into L0. We call it s4. Because s4 is ingested after m1 became immutable memtable, so it has a larger log sequence number than m1.

- m1 is flushed in L0. because it is small, this flush job finish quickly. we call it s5.

- [s1, s2, s3, s4] are compacted into one sst to L0, by IntraL0Compaction.  We call it s6.
  - compacted 4@0 files to L0
- When s6 is added into manifest,  the corruption happened. because the largest sequence number of s6 is equal to s4, and they are both larger than that of s5.  But because s1 is older than m1, so the smallest sequence number of s6 is smaller than that of s5.
   - s6.smallest_seqno < s5.smallest_seqno < s5.largest_seqno < s6.largest_seqno


siying:
Assume we have L0 files
`[3 8] and [5 5]`
Is it possible that every time we pick [3 8], [5 5] is always picked up and vice verse? Is it going to make IntraL0 compaction code too complicated?

siying:
Or say, expand Intra-L0 compaction in both directions until a clean cut.

ajkr:
Hm, what if an intra-L0 is picked including [5 5] (and maybe another file like [4 4]) before [3 8] is flushed? I think then when [3 8] is flushed it'll trigger the same consistency failure.

siying:
@ajkr that is a very good point...

siying:
Then I think remove the assertion feels like a logical solution. One thing worries me is that, stress test seems not in a circumstance where Intra-L0 compaction will happen, so we might not be able to uncover bugs like this in the future.

ajkr:
I think after we have @sumeerbhola's fix for https://github.com/facebook/rocksdb/issues/5848 (if it isn't already fixed, I haven't checked recently), then we can set `-ingest_external_file_one_in` to an aggressively low number and it should trigger frequent intra-L0s.

ajkr:
High file ingestion rate could be randomly turned on in db_crashtest.py so only once in a while there'd be high ingestion rate. Then we wouldn't need to add new test setups, and it wouldn't take much coverage away from other features.

Little-Wallace:
> Hm, what if an intra-L0 is picked including [5 5] (and maybe another file like [4 4]) before [3 8] is flushed? I think then when [3 8] is flushed it'll trigger the same consistency failure.

Yes, this is what I have described in **First situatiom**

@siying  I think intra-L0 should skip files which largest seqno smaller than smallest seqno in imm or memtable.  That  is 
- max(largest_seqno(picked file)) < smallest_seqno(imm, mem)

This solution could solve the problem and do not disable intral-l0 compaction for ingested file. 
- If the largest seqno of the picked file is larger than smallest_seqno of imm, it only needs to wait for the next time to be picked, after the imm table has been flushed into L0.

riversand963:
> > Hm, what if an intra-L0 is picked including [5 5] (and maybe another file like [4 4]) before [3 8] is flushed? I think then when [3 8] is flushed it'll trigger the same consistency failure.
> 
> Yes, this is what I have described in **First situatiom**
> 
> @siying I think intra-L0 should skip files which largest seqno smaller than smallest seqno in imm or memtable. That is
> 
> * max(largest_seqno(picked file)) < smallest_seqno(imm, mem)
> 
> This solution could solve the problem and do not disable intral-l0 compaction for ingested file.
> 
> * If the largest seqno of the picked file is larger than smallest_seqno of imm, it only needs to wait for the next time to be picked, after the imm table has been flushed into L0.

Thanks @Little-Wallace for the proposal. I have a question. Let me use an example.
```
Mem                 [4, 5]
L0         [1, 3]                 [6, 6]    [7, 7]
L1
```
Originally we have L0 [1, 3] and memtable [4, 5] in which 1, 3, 4, 5 are sequence numbers. Then we ingest [6, 6] and [7, 7]. The external SSTs do not have overlapping keys with [4, 5], but overlaps with [1, 3]. Suppose no further flush is triggered. Is it possible for [6, 6] and [7, 7] be compacted using intra-L0 compaction and **then moved to L1**? Will this be a problem if we search for a key which exists in both [1, 3] and [6, 6]? Just asking because currently I am not fully familiar with all the compaction boundary picking logic.

Little-Wallace:
@riversand963  IntralL0Compaction only move files into L0 ,  rather than L1. And  key in one sst could only show as its latest version.

db/column_family.cc
```diff
@@ -391,6 +391,11 @@ void SuperVersion::Init(MemTable* new_mem, MemTableListVersion* new_imm,
   refs.store(1, std::memory_order_relaxed);
 }
 
+SequenceNumber SuperVersion::GetSmallestMemSeqno() const {
```
petermattis: Nit: I think this method should be named `GetEarliestMemTableSequenceNumber` for consistency with the naming of similar methods elsewhere. Also, this looks very similar to `DBImpl::GetEarliestMemTableSequenceNumber`.

PS I wouldn't mind if all of the uses of `earliest` where switched to `smallest` for consistency with `FileMetadata::{smallest,largest}_seqno`, but I'll leave it to the FB RocksDB folks to decide if they want to do that.

db/column_family.cc
```diff
@@ -929,8 +934,9 @@ bool ColumnFamilyData::NeedsCompaction() const {
 
 Compaction* ColumnFamilyData::PickCompaction(
     const MutableCFOptions& mutable_options, LogBuffer* log_buffer) {
+  SequenceNumber oldest_mem_seqno = super_version_->GetSmallestMemSeqno();
```
petermattis: Nit: now we have 3 terms in play: `earliest`, `smallest`, and `oldest`. For consistency, I think this should be `earliest_mem_seqno`.

db/compaction/compaction_picker.cc
```diff
@@ -52,7 +59,9 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
     compact_bytes += static_cast<size_t>(level_files[span_len]->fd.file_size);
     compensated_compact_bytes += level_files[span_len]->compensated_file_size;
     new_compact_bytes_per_del_file = compact_bytes / span_len;
-    if (level_files[span_len]->being_compacted ||
+    if ((level_files[0]->fd.smallest_seqno == level_files[0]->fd.largest_seqno &&
+      level_files[span_len]->fd.largest_seqno >= oldest_mem_seqno) ||
```
petermattis: Is this check necessary? My understanding is that `level_files` is sorted in descending `largest_seqno` order. So if the check at the start of this method succeeded, the check here must also be true. Am I missing something?

I wasn't aware that intra-L0 compactions always started with the newest sstable in L0. Rather than the approach take here, could we instead skip the prefix of `level_files` that contains sstables with `largest_seqno >= oldest_mem_seqno` and then build a span from there? Cc @ajkr.

db/compaction/compaction_picker.cc
```diff
@@ -39,7 +39,14 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
                            size_t min_files_to_compact,
                            uint64_t max_compact_bytes_per_del_file,
                            uint64_t max_compaction_bytes,
-                           CompactionInputFiles* comp_inputs) {
+                           CompactionInputFiles* comp_inputs,
+                           SequenceNumber oldest_mem_seqno) {
+
+  // Do not pick ingested file when there is at least one memtable not flushed which of seqno is overlap with the sst.
+  if (level_files[0]->fd.smallest_seqno == level_files[0]->fd.largest_seqno &&
```
petermattis: Would it be better to remove this conditional? Is there ever a case where we want to pick an sstable for intra-L0 compaction that has a largest seqno that is larger than the smallest unflushed seqno?

db/compaction/compaction_picker.cc
```diff
@@ -39,7 +39,14 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
                            size_t min_files_to_compact,
                            uint64_t max_compact_bytes_per_del_file,
                            uint64_t max_compaction_bytes,
-                           CompactionInputFiles* comp_inputs) {
+                           CompactionInputFiles* comp_inputs,
+                           SequenceNumber oldest_mem_seqno) {
```
petermattis: Another naming nit: `smallest_unflushed_seqno` would make it clearer that it is the fact that this seqno is unflushed which is important.

db/compaction/compaction_picker.cc
```diff
@@ -39,7 +39,14 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
...
+
+  // Do not pick ingested file when there is at least one memtable not flushed which of seqno is overlap with the sst.
```
ajkr: Do you think it would work to pick the span of files for intra-L0 starting at the smallest index `i` for which `level_files[i]->fd.largest_seqno < oldest_mem_seqno`?

petermattis: That's a nicer phrasing of a suggestion I had in a comment. I don't see anything wrong with this right now, but I'd like to think harder about it.

ajkr: Sorry I loaded the page before you reviewed and didn't notice the repetition. I haven't thought of any problems with it. I also plan to experiment with relaxing the restriction more generally on intra-L0 picking from the newest file as it might be related to our imports getting stuck.

db/column_family.cc
Little-Wallace: Ok， I'll fix it.

db/compaction/compaction_picker.cc
Little-Wallace: Yes， it is not necessary. I am just worried about other issues make it not keep order.

Little-Wallace: Sorry,  I make a mistake. My intention is to check every file, rather than `level_files[0]`

db/compaction/compaction_picker.cc
Little-Wallace: Yes. I have explained it in my description.  If the sstable was produced by ingested, and its key was not overlap with memtable, it would use the global seqno, which is just in [smallest seqno, largest seqno] of memtable.

db/compaction/compaction_picker.cc
Little-Wallace: I would try it.

db/column_family.cc
```diff
@@ -929,8 +934,9 @@ bool ColumnFamilyData::NeedsCompaction() const {
 
 Compaction* ColumnFamilyData::PickCompaction(
     const MutableCFOptions& mutable_options, LogBuffer* log_buffer) {
+  SequenceNumber oldest_mem_seqno = super_version_->GetEarliestMemTableSequenceNumber();
```
petermattis: For naming consistency, can you name this variable `earliest_mem_seqno`?

db/column_family.cc
```diff
@@ -391,6 +391,11 @@ void SuperVersion::Init(MemTable* new_mem, MemTableListVersion* new_imm,
   refs.store(1, std::memory_order_relaxed);
 }
 
+SequenceNumber SuperVersion::GetEarliestMemTableSequenceNumber() const {
+  SequenceNumber smallest_seqno = std::min(mem->GetEarliestSequenceNumber(), imm->GetEarliestSequenceNumber(false));
```
petermattis: For naming consistency, can you name this variable `earliest_seqno`?

db/compaction/compaction_picker.cc
```diff
@@ -39,32 +39,46 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
...
-                           CompactionInputFiles* comp_inputs) {
-  size_t compact_bytes = static_cast<size_t>(level_files[0]->fd.file_size);
-  uint64_t compensated_compact_bytes = level_files[0]->compensated_file_size;
+                           CompactionInputFiles* comp_inputs,
+                           SequenceNumber earliest_mem_seqno) {
+
+  // Do not pick ingested file when there is at least one memtable not flushed which of seqno is overlap with the sst.
+  size_t start = 0;
+  while (start < level_files.size() && level_files[start]->being_compacted &&
+      level_files[start]->fd.smallest_seqno == level_files[start]->fd.largest_seqno &&
```
petermattis: @Little-Wallace I realize that the check for `smallest_seqno == largest_seqno` is identifying ingested sstables, but I'm not sure we actually need or want that. If there is some other way for an L0 sstable to get created where `sstable.largest_seqno >= earliest_mem_seqno` we wouldn't want to involve that sstable in an intra-L0 compaction.

db/compaction/compaction_picker.cc
```diff
@@ -39,32 +39,46 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
...
                            uint64_t max_compaction_bytes,
-                           CompactionInputFiles* comp_inputs) {
-  size_t compact_bytes = static_cast<size_t>(level_files[0]->fd.file_size);
-  uint64_t compensated_compact_bytes = level_files[0]->compensated_file_size;
+                           CompactionInputFiles* comp_inputs,
+                           SequenceNumber earliest_mem_seqno) {
+
+  // Do not pick ingested file when there is at least one memtable not flushed which of seqno is overlap with the sst.
+  size_t start = 0;
+  while (start < level_files.size() && level_files[start]->being_compacted &&
```
petermattis: I think there should be `being_compacted ||`. I think it might be clearer to perform the conditionals one at a time in the loop:

```
for (; start < level_files_.size(); ++start) {
  if (level_files[start]->being_compacted) {
    continue;
  }
  if (level_files[start]->fd.largest_seqno >= earliest_mem_seqno) {
    continue;
  }
  break;
}
```

db/compaction/compaction_picker.cc
petermattis: This approach looks good to me.

db/compaction/compaction_picker.cc
sumeerbhola: If anything in the prefix that is being skipped due to earliest_mem_seqno is already being_compacted, I think we can return false. And I am not sure if that can even happen:
- if it is an intra-L0 compaction then earliest_mem_seqno must have not been monotonic
- if it is an L0 to a lower level compaction we must have skipped over the non-flushed memtable to include a later sstable which seems incorrect.

db/compaction/compaction_picker.cc
```diff
@@ -39,32 +39,46 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
...
+  size_t compact_bytes = static_cast<size_t>(level_files[start]->fd.file_size);
+  uint64_t compensated_compact_bytes = level_files[start]->compensated_file_size;
   size_t compact_bytes_per_del_file = port::kMaxSizet;
   // Compaction range will be [0, span_len).
   size_t span_len;
   // Pull in files until the amount of compaction work per deleted file begins
   // increasing or maximum total compaction size is reached.
   size_t new_compact_bytes_per_del_file = 0;
-  for (span_len = 1; span_len < level_files.size(); ++span_len) {
+  for (span_len = start + 1; span_len < level_files.size(); ++span_len) {
```
sumeerbhola: I think the statement below should be
```
new_compact_bytes_per_del_file = compact_bytes / (span_len - start);
```

db/compaction/compaction_picker.cc
ajkr: I think `return false` when we reach a `start` that is undergoing compaction is good. It is aligned with what we are doing in `LevelCompactionBuilder::PickIntraL0Compaction()` when the file at index zero is undergoing compaction. Perhaps the latter can be removed once we have the check here.

db/compaction/compaction_picker.cc
ajkr: Good catch. `span_len` isn't really the span length anymore, it's the exclusive end index of the span. Maybe we should call it `limit` or something.

ajkr:
@siying please let us know if you see any issues other than the ones mentioned in our reviews. We're planning to backport/release this soon (probably before it lands upstream).

siying:
Please update the title and summary to more directly explain what the fix does. The title now is too abstract for people to understand the fix, and the description doesn't explain what the fix is and how it fixes the problem.

db/compaction/compaction_picker_fifo.h
```diff
@@ -22,7 +22,8 @@ class FIFOCompactionPicker : public CompactionPicker {
   virtual Compaction* PickCompaction(const std::string& cf_name,
                                      const MutableCFOptions& mutable_cf_options,
                                      VersionStorageInfo* version,
-                                     LogBuffer* log_buffer) override;
+                                     LogBuffer* log_buffer,
+                                     SequenceNumber earliest_memtable_seqno = std::numeric_limits<uint64_t>::max()) override;
```
siying: use the one defined in port.h.

db/column_family.cc
siying: This approach is complicated. Instead of remembering earliest sequence number in memtable the moment when the previous flush happens in Version, it passes the recent one when PickCompaction happens. Do you have a proof that it would be the same value? My worry is that, if there is a concurrent flush going on as this PickCompaction(), is it possible that the memtable sequence number has been updated but PickCompaction() still can't see the newly flushed file?
Generally, I see it safer if the earliest memtable sequence number is remembered in Version when a new Version is generated.

petermattis: `PickCompaction` is called with `DB::mutex_` held, right? If there is a concurrent flush going on the memtables are either still present in `super_version_->imm`, or the flush has completed in which case it will have installed in `super_version_->current`. It looks like `ColumnFamily::current_` and `ColumnFamily::super_version_` are both updated with `DB::mutex_` held and should be identical. I'm not seeing what would be different about the approach taken in this PR and memoizing the earliest unflushed sequence number when the version is installed. I'm just reading code, though. Very possible that I'm missing something. 

Is there always an earliest memtable sequence number when a new version is installed. If I write a single key to the DB and call `Flush`, a new version will be created, and a new memtable will be created, but there won't be an earliest memtable sequence number.

ajkr: I do not have a proof, but my reading of the code suggests the following sequence of events in a single thread:

- `LogAndApply()` is called on one or more memtables that have finished flushing.
  - `ColumnFamilyData::current_` is updated which will include L0 files corresponding to the flushed memtables.
- The flushed memtables are removed from the `MemTableListVersion`.
- `InstallSuperVersion()` writes a new `ColumnFamilyData::super_version_` with the updated memtable list that excludes the flushed memtables.

Now, when this sequence of events happens concurrently, I believe the problematic sequence you mentioned should not happen.  In particular because a flushed L0 file is installed to `ColumnFamilyData::current_` *before* its corresponding memtable is dropped from `ColumnFamilyData::super_version_`. For a given memtable, these two updates happen in the same thread, so even if lock is released between the two updates, it should not be possible for another thread to drop our memtable in `super_version_`.

db/column_family.cc
Little-Wallace: Ok. Maybe I should `grep` all world like `smallest_seqno`

db/column_family.cc
Little-Wallace: These operations all happens during held mutex. And even if we get earliest seqno from an older  super_version, it does not matter. Because smaller seqno would pick less file. We only need to wait for next time to pick these files.

db/compaction/compaction_picker_fifo.h
Little-Wallace: Ok, I have replaced it.

db/column_family.cc
siying: Can we directly use ColumnFamilyData::mem_, ColumnFamilyData::imm_ to calculate the value instead of getting from SuperVersion? If everything is under DB mutex, then current_, mem_ and imm_ should all be from the same snapshot. Does it work?

riversand963:
Have a few questions about the description.

> ### First situation：
> * First we ingest five external sst into Rocksdb, and they happened to be ingested in L0. and there had been some data in memtable, which make the smallest sequence number of memtable is less than which of sst that we ingest.
> * If there had been one compaction job which compacted sst from L0 to L1, `LevelCompactionPicker` would trigger a `IntraL0Compaction` which would compact this five sst from L0 to L0. We call this sst A, which was merged from five ingested sst.
> * Then some data was put into memtable, and memtable was flushed to L0. We called this sst B.
> * RocksDB check consistency , and find the `smallest_seqno` of A is  less than that of B and crash. Because B was merged from five sst, the smallest sequence number of it was less than the biggest sequece number of it. So RocksDB could not tell if B was produce by ingested.
> 
Should A's smallest_seqno be larger than B's? B is the result of flush, while A results from compaction.

db/compaction/compaction_picker.cc
yiwu-arbug: @ajkr It seems to me it could still order the output to be newer than `L0_(i-1)` if `L0_i` is an ingested file and `L0_i->largest_seqno < oldest_mem_seqno`.

db/compaction/compaction_picker.h
```diff
@@ -57,7 +57,8 @@ class CompactionPicker {
   virtual Compaction* PickCompaction(const std::string& cf_name,
                                      const MutableCFOptions& mutable_cf_options,
                                      VersionStorageInfo* vstorage,
-                                     LogBuffer* log_buffer) = 0;
+                                     LogBuffer* log_buffer,
+                                     SequenceNumber earliest_memtable_seqno = port::kMaxUint64) = 0;
```
yiwu-arbug: s/port::kMaxUint64/kMaxSequenceNumber ?

db/compaction/compaction_picker_fifo.h
```diff
@@ -22,7 +22,8 @@ class FIFOCompactionPicker : public CompactionPicker {
   virtual Compaction* PickCompaction(const std::string& cf_name,
                                      const MutableCFOptions& mutable_cf_options,
                                      VersionStorageInfo* version,
-                                     LogBuffer* log_buffer) override;
+                                     LogBuffer* log_buffer,
+                                     SequenceNumber earliest_memtable_seqno = port::kMaxUint64) override;
```
yiwu-arbug: ditto

db/compaction/compaction_picker.h
Little-Wallace: Yes. I didn't know that there was such a variable before.

db/compaction/compaction_picker.cc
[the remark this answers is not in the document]
Little-Wallace: Maybe not. If a file is `being_compacted`,  we must skip it.

Little-Wallace:
@riversand963 No. B was merged from ingested file b and other L0 files. b was ingested when A had become immutable memtable. So b has a larger sequence number than A.

db/column_family.cc
Little-Wallace: Yes, it works. But  there has been an available interface of SuperVersion. I just can not understand why we do not use it. I thought `ColumnFamilyData` would call `InstallSuperVersion` every time when switch memtable or flush job was finished. Is it right?

riversand963:
@Little-Wallace  Did I miss anything? From the PR description I quote
"
If there had been one compaction job which compacted sst from L0 to L1, LevelCompactionPicker would trigger a IntraL0Compaction which would compact this five sst from L0 to L0. **We call this sst A, which was merged from five ingested sst**.

Then some data was put into memtable, and memtable was flushed to L0. **We called this sst B**.
"

db/compaction/compaction_picker.cc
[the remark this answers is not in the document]
siying: Correct me if I'm wrong: previously, no Inra-L0 compaction is issued if any file scanned is in compaction. Now with the change, we are skipping them instead. I'm not sure whether the new behavior introduces bug or not, but it is really hard to be reason about. In theory, a file can be picked up for L0=>L1 compaction, or another Intro-L0 compaction. Now we are picking older L0 files to an intro-L0 compaction. We are bringing ourselves to a scenario that is really had to reason about. Unless I see a strong case that relaxing the condition can improve performance significantly, my vote would be to keep the logic simple, which is to skip intro-L0 compaction if we see a file in compaction here.

ajkr:
I think there is some confusion about which is sst A and sst B.

@riversand963 later he mentions `Because B was merged from five sst, the smallest sequence number of it was less than the biggest sequece number of it.` which suggests B is the result of intra-L0. I think if you swap A and B in the part you quoted then everything makes sense.

riversand963:
> I think there is some confusion about which is sst A and sst B.
> 
> @riversand963 later in the description he mentions `Because B was merged from five sst, the smallest sequence number of it was less than the biggest sequece number of it.` which suggests B is the result of intra-L0. I think if you swap A and B in the part you quoted then everything makes sense.

Agree. Thanks for the suggestion.

db/compaction/compaction_picker.cc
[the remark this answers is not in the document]
Little-Wallace: Ok, I agree. This is a less risky suggestion.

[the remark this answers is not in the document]
yiwu-arbug: @siying I think it might be a good idea to allow picking *newer* files for intra-L0 compaction while there are L0=>L1 compaction, which makes intra-L0 compaction more useful than before: avoid write stall if there's a slow L0=>L1 compaction.

db/column_family.cc
siying: I am not 100% sure about this. I would need to check source code to double check. When we rely on such an assumption that is applicable to change in the future, we are making the code prone to bugs in the future. Another reason is, like what you have said, SuperVersion is created after updating current_, imm_, or mem_ but it doesn't have to be atomic to it. It's perfectly OK to update current_, imm_ or mem_, release the DB mutex, do something like and then install super version. So SuperVersion is not guaranteed to be in sync with imm_ and mem_. I saw the argument that a stale SuperVersion should still be OK in this case, the argument seems to be plausible to me, but if we don't have to rely on such proof to make sure the code is correct, it is even better.

db/compaction/compaction_picker.cc
```diff
@@ -39,32 +39,49 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
...
+      break;
+    }
+  }
+  if (start >= level_files.size()) {
+    return false;
+  }
+  size_t compact_bytes = static_cast<size_t>(level_files[start]->fd.file_size);
+  uint64_t compensated_compact_bytes = level_files[start]->compensated_file_size;
   size_t compact_bytes_per_del_file = port::kMaxSizet;
   // Compaction range will be [0, span_len).
```
riversand963: This comment seems to be out-dated because we are compacting [start + 1, limit)

yiwu-arbug:
@Little-Wallace you need to run `make format` once. Some lines are still over 80 chars.

riversand963:
> ### First situation：
> * First we ingest five external sst into Rocksdb, and they happened to be ingested in L0. and there had been some data in memtable, which make the smallest sequence number of memtable is less than which of sst that we ingest.
> * If there had been one compaction job which compacted sst from L0 to L1, `LevelCompactionPicker` would trigger a `IntraL0Compaction` which would compact this five sst from L0 to L0. We call this sst A, which was merged from five ingested sst.
> * Then some data was put into memtable, and memtable was flushed to L0. We called this sst B.
> * RocksDB check consistency , and find the `smallest_seqno` of A is  less than that of B and crash. Because B was merged from five sst, the smallest sequence number of it was less than the biggest sequece number of it. So RocksDB could not tell if B was produce by ingested.
> 

As discussed earlier, this seems a typo (if not correct me). Is SST A or B generated from compaction or flush? This is a minor thing, but still worth fixing to avoid confusing people since this message will be in the commit log, and users may rely on it to understand the context.

yiwu-arbug:
> > ### First situation：
> > 
> > * First we ingest five external sst into Rocksdb, and they happened to be ingested in L0. and there had been some data in memtable, which make the smallest sequence number of memtable is less than which of sst that we ingest.
> > * If there had been one compaction job which compacted sst from L0 to L1, `LevelCompactionPicker` would trigger a `IntraL0Compaction` which would compact this five sst from L0 to L0. We call this sst A, which was merged from five ingested sst.
> > * Then some data was put into memtable, and memtable was flushed to L0. We called this sst B.
> > * RocksDB check consistency , and find the `smallest_seqno` of A is  less than that of B and crash. Because B was merged from five sst, the smallest sequence number of it was less than the biggest sequece number of it. So RocksDB could not tell if B was produce by ingested.
> 
> As discussed earlier, this seems a typo (if not correct me). Is SST A or B generated from compaction or flush? This is a minor thing, but still worth fixing to avoid confusing people since this message will be in the commit log, and users may rely on it to understand the context.

updated.

db/compaction/compaction_picker.cc
```diff
@@ -39,32 +39,51 @@ bool FindIntraL0Compaction(const std::vector<FileMetaData*>& level_files,
...
-  for (span_len = 1; span_len < level_files.size(); ++span_len) {
-    compact_bytes += static_cast<size_t>(level_files[span_len]->fd.file_size);
-    compensated_compact_bytes += level_files[span_len]->compensated_file_size;
-    new_compact_bytes_per_del_file = compact_bytes / span_len;
-    if (level_files[span_len]->being_compacted ||
+  for (limit = start + 1; limit < level_files.size(); ++limit) {
+    compact_bytes += static_cast<size_t>(level_files[limit]->fd.file_size);
+    compensated_compact_bytes += level_files[limit]->compensated_file_size;
+    new_compact_bytes_per_del_file = compact_bytes / (limit - start);
+    if (level_files[limit]->fd.largest_seqno >= earliest_mem_seqno ||
```
siying: I don't know how `level_files[limit]->fd.largest_seqno >= earliest_mem_seqno` could possibly be true here, but if this is true, and we picked up some files supposed to be newer than this file and compact them together, it feels hard to reason about whether this is correct.

I don't think `level_files[limit]->fd.largest_seqno > earliest_mem_seqno` is ever true here (we can even add an assertion), but in case it does, can we just return false?

Regarding the `==` case, this is mystery. It feels inconsistent with the previous check of
```
    if (level_files[start]->fd.largest_seqno <= earliest_mem_seqno) {
      break;
    }
```
Can we treat the equality case consistently? Should we include the file whose largest_seqno equals earliest_mem_seqno or not? We may already included level_files[start] which satisfies the equality condition.

Little-Wallace:
@siying  any more comments?

yiwu-arbug:
@siying ping

db/compaction/compaction_picker_test.cc
```diff
@@ -1516,12 +1516,12 @@ TEST_F(CompactionPickerTest, IntraL0MaxCompactionBytesHit) {
...
-  Add(0, 3U, "201", "250", 200000U);
-  Add(0, 4U, "251", "300", 200000U);
-  Add(0, 5U, "301", "350", 200000U);
-  Add(1, 6U, "100", "350", 200000U);
+  Add(0, 1U, "100", "150", 200000U, 0, 100, 101);
+  Add(0, 2U, "151", "200", 200000U, 0, 102, 103);
+  Add(0, 3U, "201", "250", 200000U, 0, 104, 105);
+  Add(0, 4U, "251", "300", 200000U, 0, 106, 107);
+  Add(0, 5U, "301", "350", 200000U, 0, 108, 109);
+  Add(1, 6U, "100", "350", 200000U, 0, 109, 110);
```
siying: Can you add several tests to cover more scenarios in FindIntraL0Compaction()? For example, when being_compacted shows up in L0? Also, it will be nice to directly cover the earliest_mem_seqno scenarios in tests here.

db/db_compaction_test.cc
```diff
@@ -4683,6 +4685,155 @@ TEST_F(DBCompactionTest, ConsistencyFailTest) {
...
+  ASSERT_EQ(10, NumTableFilesAtLevel(0));
+
+  // Wake up flush job
+  sleeping_tasks.WakeUp();
+  sleeping_tasks.WaitUntilDone();
+
+  dbfull()->TEST_WaitForCompact();
+  rocksdb::SyncPoint::GetInstance()->DisableProcessing();
+  uint64_t error_count = 0;
+  db_->GetIntProperty("rocksdb.background-errors", &error_count);
```
siying: In these two tests, other than asserting no error, can we validate Get() the keys we inserted returns the correct  results?

Little-Wallace: Good idea.

db/compaction/compaction_picker_test.cc
Little-Wallace: Yes, I have add a tests for `FindIntraL0Compaction`, which make sure that it would skip sst which of largest lsn is larger than earliest_seqno.

yiwu-arbug:
@ajkr ping

ajkr:
Sorry didn't notice I was tagged before. Taking a look.

Little-Wallace:
@ajkr  force_consistency_checks does not work for process compiled in `DEBUG` mode. If you revert this fix, this ut would failed to check consistency in Compaction thread.

ajkr:
Oh right, that's a good point. It was only possible on our fork as we deleted the part of the consistency check that fails with intra-L0. Anyways, this LGTM.

siying:
btw I see this failure once:
```
Note: Google Test filter = DBCompactionTestWithParam/DBCompactionTestWithParam.FlushAfterL0IntraCompactionCheckConsistencyFail/0
[----------] 1 test from DBCompactionTestWithParam/DBCompactionTestWithParam
[ RUN      ] DBCompactionTestWithParam/DBCompactionTestWithParam.FlushAfterL0IntraCompactionCheckConsistencyFail/0
db/db_compaction_test.cc:4921: Failure
  i + 1
    Which is: 10
  NumTableFilesAtLevel(0)
    Which is: 5
terminate called after throwing an instance of 'testing::internal::GoogleTestFailureException'
  what():  db/db_compaction_test.cc:4921: Failure
  i + 1
    Which is: 10
  NumTableFilesAtLevel(0)
    Which is: 5
```
It doesn't reproduce though, but I wonder whether you have any thoughts on what could be the reason.

siying:
And another run is hanging:
```
(gdb) thread apply all bt

Thread 7 (Thread 0x7fca00ff9700 (LWP 15225)):
#0  0x00007fca155e7f85 in futex_abstimed_wait_cancelable (private=<optimized out>, abstime=0x7fca00ff8a30, expected=0,
    futex_word=0x5580d3ee3f98) at ../sysdeps/unix/sysv/linux/futex-internal.h:205
#1  __pthread_cond_wait_common (abstime=0x7fca00ff8a30, mutex=0x5580d3ee3f28, cond=0x5580d3ee3f70) at pthread_cond_wait.c:539
#2  __pthread_cond_timedwait (cond=cond@entry=0x5580d3ee3f70, mutex=0x5580d3ee3f28, abstime=abstime@entry=0x7fca00ff8a30)
    at pthread_cond_wait.c:667
#3  0x00005580d2d191e9 in rocksdb::port::CondVar::TimedWait (this=this@entry=0x5580d3ee3f70, abs_time_us=<optimized out>)
    at port/port_posix.cc:120
#4  0x00005580d2ce5154 in rocksdb::InstrumentedCondVar::TimedWaitInternal (this=0x5580d3ee3f70, abs_time_us=<optimized out>)
    at monitoring/instrumented_mutex.cc:66
#5  0x00005580d2ce522d in rocksdb::InstrumentedCondVar::TimedWait (this=this@entry=0x5580d3ee3f70,
    abs_time_us=abs_time_us@entry=1574277484269966) at monitoring/instrumented_mutex.cc:55
#6  0x00005580d2b6d97b in rocksdb::RepeatableThread::wait (delay=<optimized out>, this=0x5580d3ee3ed0) at ./util/repeatable_thread.h:90
#7  rocksdb::RepeatableThread::thread (this=0x5580d3ee3ed0) at ./util/repeatable_thread.h:126
#8  rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}::operator()() const (__closure=<optimized out>)
    at ./util/repeatable_thread.h:38
#9  std::__invoke_impl<void, rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}>(std::__invoke_other, rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}&&) (__f=...) at /usr/include/c++/9/bits/invoke.h:60
#10 std::__invoke<rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}>(std::__invoke_result&&, (rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}&&)...) (__fn=...) at /usr/include/c++/9/bits/invoke.h:95
#11 std::thread::_Invoker<std::tuple<rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}> >::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=<optimized out>) at /usr/include/c++/9/thread:244
#12 std::thread::_Invoker<std::tuple<rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}> >::operator()() (
    this=<optimized out>) at /usr/include/c++/9/thread:251
#13 std::thread::_State_impl<std::thread::_Invoker<std::tuple<rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}> > >::_M_run() (this=<optimized out>) at /usr/include/c++/9/thread:195
#14 0x00007fca14c9e870 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#15 0x00007fca155e16db in start_thread (arg=0x7fca00ff9700) at pthread_create.c:463
#16 0x00007fca1434888f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 6 (Thread 0x7fc9f67fc700 (LWP 15224)):
#0  0x00007fca155e7f85 in futex_abstimed_wait_cancelable (private=<optimized out>, abstime=0x7fc9f67fba30, expected=0,
    futex_word=0x5580d3ee1238) at ../sysdeps/unix/sysv/linux/futex-internal.h:205
#1  __pthread_cond_wait_common (abstime=0x7fc9f67fba30, mutex=0x5580d3ee11c8, cond=0x5580d3ee1210) at pthread_cond_wait.c:539
#2  __pthread_cond_timedwait (cond=cond@entry=0x5580d3ee1210, mutex=0x5580d3ee11c8, abstime=abstime@entry=0x7fc9f67fba30)
    at pthread_cond_wait.c:667
#3  0x00005580d2d191e9 in rocksdb::port::CondVar::TimedWait (this=this@entry=0x5580d3ee1210, abs_time_us=<optimized out>)
    at port/port_posix.cc:120
#4  0x00005580d2ce5154 in rocksdb::InstrumentedCondVar::TimedWaitInternal (this=0x5580d3ee1210, abs_time_us=<optimized out>)
    at monitoring/instrumented_mutex.cc:66
#5  0x00005580d2ce522d in rocksdb::InstrumentedCondVar::TimedWait (this=this@entry=0x5580d3ee1210,
    abs_time_us=abs_time_us@entry=1574277484296472) at monitoring/instrumented_mutex.cc:55
#6  0x00005580d2b6d97b in rocksdb::RepeatableThread::wait (delay=<optimized out>, this=0x5580d3ee1170) at ./util/repeatable_thread.h:90
#7  rocksdb::RepeatableThread::thread (this=0x5580d3ee1170) at ./util/repeatable_thread.h:126
#8  rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}::operator()() const (__closure=<optimized out>)
    at ./util/repeatable_thread.h:38
#9  std::__invoke_impl<void, rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}>(std::__invoke_other, rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}&&) (__f=...) at /usr/include/c++/9/bits/invoke.h:60
#10 std::__invoke<rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}>(std::__invoke_result&&, (rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}&&)...) (__fn=...) at /usr/include/c++/9/bits/invoke.h:95
#11 std::thread::_Invoker<std::tuple<rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}> >::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=<optimized out>) at /usr/include/c++/9/thread:244
#12 std::thread::_Invoker<std::tuple<rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}> >::operator()() (
    this=<optimized out>) at /usr/include/c++/9/thread:251
#13 std::thread::_State_impl<std::thread::_Invoker<std::tuple<rocksdb::RepeatableThread::RepeatableThread(std::function<void ()>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rocksdb::Env*, unsigned long, unsigned long)::{lambda()#1}> > >::_M_run() (this=<optimized out>) at /usr/include/c++/9/thread:195
#14 0x00007fca14c9e870 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#15 0x00007fca155e16db in start_thread (arg=0x7fc9f67fc700) at pthread_create.c:463
#16 0x00007fca1434888f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 5 (Thread 0x7fc9f77fe700 (LWP 15142)):
#0  0x00007fca155e79f3 in futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x5580d3ecd7c8)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:88
#1  __pthread_cond_wait_common (abstime=0x0, mutex=0x5580d3ecd6c0, cond=0x5580d3ecd7a0) at pthread_cond_wait.c:502
#2  __pthread_cond_wait (cond=cond@entry=0x5580d3ecd7a0, mutex=0x5580d3ecd6c0) at pthread_cond_wait.c:655
#3  0x00005580d2d19169 in rocksdb::port::CondVar::Wait (this=this@entry=0x5580d3ecd7a0) at port/port_posix.cc:106
#4  0x00005580d2ce4f4f in rocksdb::InstrumentedCondVar::WaitInternal (this=0x5580d3ecd7a0) at monitoring/instrumented_mutex.cc:48
#5  rocksdb::InstrumentedCondVar::Wait (this=this@entry=0x5580d3ecd7a0) at monitoring/instrumented_mutex.cc:41
#6  0x00005580d2cc3f40 in rocksdb::DeleteScheduler::BackgroundEmptyTrash (this=0x5580d3ecd6a8) at file/delete_scheduler.cc:194
#7  0x00007fca14c9e870 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007fca155e16db in start_thread (arg=0x7fc9f77fe700) at pthread_create.c:463
#9  0x00007fca1434888f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 4 (Thread 0x7fca13224700 (LWP 15141)):
#0  0x00007fca155e79f3 in futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x5580d3ec9930)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:88
#1  __pthread_cond_wait_common (abstime=0x0, mutex=0x5580d3ec98e0, cond=0x5580d3ec9908) at pthread_cond_wait.c:502
#2  __pthread_cond_wait (cond=0x5580d3ec9908, mutex=0x5580d3ec98e0) at pthread_cond_wait.c:655
#3  0x00007fca14c98f0c in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00005580d2dccf25 in rocksdb::ThreadPoolImpl::Impl::BGThread (this=0x5580d3ec9870, thread_id=1) at util/threadpool_imp.cc:196
#5  0x00005580d2dcd25e in rocksdb::ThreadPoolImpl::Impl::BGThreadWrapper (arg=0x5580d3ee0a80) at util/threadpool_imp.cc:306
#6  0x00007fca14c9e870 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007fca155e16db in start_thread (arg=0x7fca13224700) at pthread_create.c:463
#8  0x00007fca1434888f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 3 (Thread 0x7fca13a25700 (LWP 15078)):
#0  0x00007fca155e79f3 in futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x7fff3cefd388)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:88
#1  __pthread_cond_wait_common (abstime=0x0, mutex=0x7fff3cefd330, cond=0x7fff3cefd360) at pthread_cond_wait.c:502
#2  __pthread_cond_wait (cond=cond@entry=0x7fff3cefd360, mutex=0x7fff3cefd330) at pthread_cond_wait.c:655
#3  0x00005580d2d19169 in rocksdb::port::CondVar::Wait (this=0x7fff3cefd360) at port/port_posix.cc:106
#4  0x00005580d2a97b18 in rocksdb::test::SleepingBackgroundTask::DoSleep (this=0x7fff3cefd330) at ./test_util/testutil.h:411
#5  rocksdb::test::SleepingBackgroundTask::DoSleepTask (arg=0x7fff3cefd330) at ./test_util/testutil.h:446
#6  0x00005580d2dcd0cb in std::function<void ()>::operator()() const (this=0x7fca13a24b00) at /usr/include/c++/9/bits/std_function.h:683
#7  rocksdb::ThreadPoolImpl::Impl::BGThread (this=0x5580d3ec9b60, thread_id=0) at util/threadpool_imp.cc:265
#8  0x00005580d2dcd25e in rocksdb::ThreadPoolImpl::Impl::BGThreadWrapper (arg=0x5580d3ecc200) at util/threadpool_imp.cc:306
#9  0x00007fca14c9e870 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#10 0x00007fca155e16db in start_thread (arg=0x7fca13a25700) at pthread_create.c:463
#11 0x00007fca1434888f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 2 (Thread 0x7fca14226700 (LWP 15077)):
#0  0x00007fca155e79f3 in futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x5580d3ec9930)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:88
#1  __pthread_cond_wait_common (abstime=0x0, mutex=0x5580d3ec98e0, cond=0x5580d3ec9908) at pthread_cond_wait.c:502
#2  __pthread_cond_wait (cond=0x5580d3ec9908, mutex=0x5580d3ec98e0) at pthread_cond_wait.c:655
#3  0x00007fca14c98f0c in std::condition_variable::wait(std::unique_lock<std::mutex>&) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00005580d2dccf25 in rocksdb::ThreadPoolImpl::Impl::BGThread (this=0x5580d3ec9870, thread_id=0) at util/threadpool_imp.cc:196
#5  0x00005580d2dcd25e in rocksdb::ThreadPoolImpl::Impl::BGThreadWrapper (arg=0x5580d3ecc060) at util/threadpool_imp.cc:306
#6  0x00007fca14c9e870 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007fca155e16db in start_thread (arg=0x7fca14226700) at pthread_create.c:463
#8  0x00007fca1434888f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 1 (Thread 0x7fca15c12a80 (LWP 15073)):
#0  0x00007fca155e7449 in futex_wait (private=<optimized out>, expected=12, futex_word=0x7fff3cefd384)
    at ../sysdeps/unix/sysv/linux/futex-internal.h:61
#1  futex_wait_simple (private=<optimized out>, expected=12, futex_word=0x7fff3cefd384) at ../sysdeps/nptl/futex-internal.h:135
#2  __pthread_cond_destroy (cond=0x7fff3cefd360) at pthread_cond_destroy.c:54
#3  0x00005580d2d19129 in rocksdb::port::CondVar::~CondVar (this=<optimized out>, __in_chrg=<optimized out>) at port/port_posix.cc:100
#4  0x00005580d29b7c14 in rocksdb::test::SleepingBackgroundTask::~SleepingBackgroundTask (this=0x7fff3cefd330, __in_chrg=<optimized out>)
    at ./test_util/testutil.h:394
#5  rocksdb::DBCompactionTestWithParam_IntraL0CompactionAfterFlushCheckConsistencyFail_Test::TestBody (this=<optimized out>)
    at db/db_compaction_test.cc:4981
#6  0x00005580d2f01730 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void> (
    location=0x5580d2f7f4d5 "the test body", method=<optimized out>, object=0x5580d3ec8c60)
    at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:3880
#7  testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (object=object@entry=0x5580d3ec8c60, method=<optimized out>,
    location=location@entry=0x5580d2f7f4d5 "the test body") at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:3935
#8  0x00005580d2ef5b36 in testing::Test::Run (this=this@entry=0x5580d3ec8c60) at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:3973
#9  0x00005580d2ef5de5 in testing::Test::Run (this=0x5580d3ec8c60) at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:4149
#10 testing::TestInfo::Run (this=0x5580d3ec67b0) at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:4149
#11 0x00005580d2ef5f9c in testing::TestInfo::Run (this=<optimized out>) at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:4124
#12 testing::TestCase::Run (this=0x5580d3eb8a90) at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:4267
#13 0x00005580d2ef666b in testing::TestCase::Run (this=<optimized out>) at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:4253
#14 testing::internal::UnitTestImpl::RunAllTests (this=0x5580d3e7c1f0) at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:6633
#15 0x00005580d2f01cb0 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (
    location=0x5580d2f806f0 "auxiliary test code (environments or event listeners)", method=<optimized out>, object=0x5580d3e7c1f0)
    at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:3880
#16 testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x5580d3e7c1f0,
    method=<optimized out>, location=location@entry=0x5580d2f806f0 "auxiliary test code (environments or event listeners)")
    at third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc:3935
#17 0x00005580d2ef681f in testing::UnitTest::Run (this=0x5580d3261500 <testing::UnitTest::GetInstance()::instance>)
    at ./third-party/gtest-1.8.1/fused-src/gtest/gtest.h:21103
#18 0x00005580d2a0569e in RUN_ALL_TESTS () at ./third-party/gtest-1.8.1/fused-src/gtest/gtest.h:22104
#19 main (argc=<optimized out>, argv=0x7fff3cefdc98) at db/db_compaction_test.cc:5023
```
Can you take a look?

Little-Wallace:
@siying  Ok, It may be conflict with some change of master.

Little-Wallace:
This failure is not always appear every time?   I have not been able to reproduce this failure.

Little-Wallace:
Emmmm,  I  have not been able to reproduce this first failure, but I guess that was caused by IntraL0 compact so quickly that it finished before I had ingested all sst. I set SYNC_POINT in another way to solve this problem. @siying

Little-Wallace:
The second failure was caused by the first failure. If `SleepingBackgroundTask` destruct before calling `WakeUp()`, the flush thread will never be wake up. So if test fails before `WakeUp`, the process will be hanging.

PR #5964, opened 2019-10-24
Rocksdb spdk

ruizhang1230:
It is a wrong operation! Just drop it


riversand963:
> It is a wrong operation! Just drop it

OK, I am closing this one.

PR #5968, opened 2019-10-25
More improvements to filter_bench

pdillinger:
* Adds support for plain table filter. This is not critical right now, but does add a -impl flag that will be useful for new filter implementations initially targeted at block-based table (and maybe later ported to plain table)
* Better mixing of inside vs. outside queries, for more realism
* A -best_case option handy for implementation tuning inner loop
* Option for whether to include hashing time in dry run / net timings

No modifications to production code, just filter_bench.

### DIFF
--- a/util/filter_bench.cc
+++ b/util/filter_bench.cc
@@ -68,6 +68,11 @@ DEFINE_uint32(impl, 0,
               "0 = full filter, 1 = block-based filter. With "
               "-use_plain_table_bloom: 0 = no locality, 1 = locality.");
 
+DEFINE_bool(net_includes_hashing, false,
+            "Whether query net ns/op times should include hashing. "
+            "(if not, dry run will include hashing) "
+            "(build times always include hashing)");
+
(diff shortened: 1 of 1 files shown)


util/filter_bench.cc
```diff
@@ -444,7 +589,8 @@ double FilterBench::RandomQueryTest(bool inside, bool dry_run, TestMode mode) {
     std::cout << "ns/op: " << ns << std::endl;
   }
 
-  if (!inside && !dry_run && mode == kRandomFilter) {
+  if (!dry_run) {
+    fp_rate_report_ = std::ostringstream();
```
merryChris: It may not compatible in gcc 4.9. How about `fp_rate_report_.clear(); fp_rate_report_.str("");` ?

pdillinger: If it's not working for you, please create a pull request. You can assign it to me. Thanks

merryChris: https://github.com/facebook/rocksdb/pull/6043 pr as above, plz make a review.

PR #5976, opened 2019-10-28
Fix IngestExternalFile's bug with two_write_queue

Little-Wallace:
When two_write_queue enable, IngestExternalFile performs EnterUnbatched on both write queues. SwitchMemtable also EnterUnbatched on 2nd write queue when this option is enabled. When the call stack includes IngestExternalFile -> FlushMemTable -> SwitchMemtable, this results into a deadlock.
The implemented solution is to pass on the existing writes_stopped argument in FlushMemTable to skip EnterUnbatched in SwitchMemtable.
Fixes https://github.com/facebook/rocksdb/issues/5974


yiwu-arbug:
@maysamyabandeh PTAL, thanks!

yiwu-arbug:
@Little-Wallace The approach seems overly complicated. Shall we fix it by passing a flag from `IngestExternalFiles()` to `SwitchMemtable()` to indicate we are already in `nonmem_write_thread`?

Little-Wallace:
> @Little-Wallace The approach seems overly complicated. Shall we fix it by passing a flag from `IngestExternalFiles()` to `SwitchMemtable()` to indicate we are already in `nonmem_write_thread`?

It may works. But it must add a param for every function which may call `EnterUnbatch`.

maysamyabandeh:
Thanks @Little-Wallace for working on this. I agree with @yiwu-arbug in favoring simpler solutions. SwitchMemtable is used in a few places and extending its signature with a nonmem_write_stopped should be a simpler change.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1537,7 +1537,7 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
     }
 
     if (!cfd->mem()->IsEmpty() || !cached_recoverable_state_empty_.load()) {
-      s = SwitchMemtable(cfd, &context);
+      s = SwitchMemtable(cfd, &context, nonmem_writes_stopped);
```
maysamyabandeh: I think we can pass writes_stopped here.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1567,7 +1567,7 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
                            "Force flushing stats CF with manual flush of %s "
                            "to avoid holding old logs",
                            cfd->GetName().c_str());
-            s = SwitchMemtable(cfd_stats, &context);
+            s = SwitchMemtable(cfd_stats, &context, nonmem_writes_stopped);
```
maysamyabandeh: I think we can pass writes_stopped here.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1668,7 +1668,7 @@ Status DBImpl::AtomicFlushMemTables(
         continue;
       }
       cfd->Ref();
-      s = SwitchMemtable(cfd, &context);
+      s = SwitchMemtable(cfd, &context, nonmem_writes_stopped);
```
maysamyabandeh: I think we can pass writes_stopped here.

db/db_impl/db_impl.cc
```diff
@@ -3852,7 +3852,7 @@ Status DBImpl::IngestExternalFiles(
         mutex_.Unlock();
         status = AtomicFlushMemTables(cfds_to_flush, flush_opts,
                                       FlushReason::kExternalFileIngestion,
-                                      true /* writes_stopped */);
+                                      true /* writes_stopped */, true /* nonmem_writes_stopped */);
```
maysamyabandeh: I think we do not need a new argument. write_stopped could mean that the writes of both write threads are stopped.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1537,7 +1537,7 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
     }
```
maysamyabandeh: Since we are going to use only one bool, both write threads should be entered unbatched if writes_stopped if false.

maysamyabandeh: What do you think about this comment?

maysamyabandeh:
@Little-Wallace did you get a chance to look at this comment? https://github.com/facebook/rocksdb/pull/5976#pullrequestreview-310500320

db/db_impl/db_impl_compaction_flush.cc
Little-Wallace: Yes, I have merge them into one variable.

Little-Wallace:
@maysamyabandeh I am sorry that I have only seen this message until now. I agree you opinion and I have merged them into one bool in code. Please take a look.

db/db_impl/db_impl_compaction_flush.cc
maysamyabandeh: The comments says:
> Since we are going to use only one bool, both write threads should be entered unbatched if writes_stopped if false.

Meaning that at this lined we need to enter_unbatched for both memtables.

maysamyabandeh:
@Little-Wallace I guess the comment was misunderstood. I tried to clarify it.

db/db_impl/db_impl_compaction_flush.cc
Little-Wallace: I dont's understand. We have enter_unbatched for `write_thread_`  while calling `IngestExternalFiles`, which is before we called `SwitchMemtable`. Why we need to enter_unbatched again for `write_thread_` in `SwitchMemtable`?

maysamyabandeh: Not for the main write_thread_, but for the 2nd one: `nonmem_write_thread_`. This is because we tell `SwitchMemtable` that we have already entered it so we should keep the promise.

Little-Wallace: Do means that I should move `EnterUnbatched` outside of `SwitchMemtable`?

maysamyabandeh: Yes

maysamyabandeh: No. actually. Because inside SwithcMemtable you would skip it if writes_stopped is true.
This conversation is moving too slowly. I can go ahead and apply the change.

Little-Wallace:
@maysamyabandeh  I have changed this. I moved EnterUnbatch outside of `SwitchMemtable`.  Please take a look.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1650,8 +1657,14 @@ Status DBImpl::AtomicFlushMemTables(
     InstrumentedMutexLock guard_lock(&mutex_);
 
     WriteThread::Writer w;
+    WriteThread::Writer nonmem_w;
     if (!writes_stopped) {
       write_thread_.EnterUnbatched(&w, &mutex_);
+      if (two_write_queues_) {
+        // SwitchMemtable is a rare event. To simply the reasoning, we make sure
```
maysamyabandeh: Can you remove this comment?

db/db_impl/db_impl_write.cc
```diff
@@ -1575,12 +1602,6 @@ void DBImpl::NotifyOnMemTableSealed(ColumnFamilyData* /*cfd*/,
 Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
```
maysamyabandeh: Can you add this comment above this line?
// REQUIRES: this thread is currently at the front of the 2nd writer queue if
// two_write_queues_ is true (This is to simplify the reasoning.)

utilities/transactions/write_prepared_transaction_test.cc
```diff
@@ -3497,6 +3497,31 @@ TEST_P(WritePreparedTransactionTest, WC_WP_WALForwardIncompatibility) {
   CrossCompatibilityTest(WRITE_PREPARED, WRITE_COMMITTED, !empty_wal);
 }
 
+TEST_P(WritePreparedTransactionTest, IngestFile) {
```
maysamyabandeh: What are the minimum conditions to trigger the test? two_write_queue and IngestExternalFile? If yes, can we move the test to ./db/external_sst_file_test.cc?

utilities/transactions/write_prepared_transaction_test.cc
```diff
@@ -3497,6 +3497,31 @@ TEST_P(WritePreparedTransactionTest, WC_WP_WALForwardIncompatibility) {
...
+  // Use large buffer to avoid memtable flush after 1024 insertions
+  options.write_buffer_size = 1024 * 1024;
+  txn_db_options.skip_concurrency_control = true;
+  printf("two write que: %d, unordered write: %d\n", options.two_write_queues, options.unordered_write);
```
maysamyabandeh: Can we remove this line?

utilities/transactions/write_prepared_transaction_test.cc
```diff
@@ -3497,6 +3497,31 @@ TEST_P(WritePreparedTransactionTest, WC_WP_WALForwardIncompatibility) {
...
+  ReOpen();
+  ASSERT_OK(db->Put(WriteOptions(), "1000", "v1"));
+  ASSERT_OK(db->Put(WriteOptions(), "1001", "v1"));
+  ASSERT_OK(db->Put(WriteOptions(), "9999", "v1"));
+  ExternalSstFileInfo info;
+  std::string f = test::PerThreadDBPath("tmp_sst_file" + std::to_string(rand()));
```
maysamyabandeh: Once this test moved to external_sst_file_test.cc, there you can find helper functions to ingest a file.

utilities/transactions/write_prepared_transaction_test.cc
```diff
@@ -3497,6 +3497,31 @@ TEST_P(WritePreparedTransactionTest, WC_WP_WALForwardIncompatibility) {
...
+  EnvOptions env_;
+  rocksdb::SstFileWriter writer(env_, options);
+  auto s = writer.Open(f);
+  ASSERT_OK(s);
+// ASSERT_OK(writer.Put(Key(), ""));
```
maysamyabandeh: Is this comment line extra?

utilities/transactions/write_prepared_transaction_test.cc
```diff
@@ -3497,6 +3497,31 @@ TEST_P(WritePreparedTransactionTest, WC_WP_WALForwardIncompatibility) {
   CrossCompatibilityTest(WRITE_PREPARED, WRITE_COMMITTED, !empty_wal);
 }
 
+TEST_P(WritePreparedTransactionTest, IngestFile) {
+  // Use large buffer to avoid memtable flush after 1024 insertions
+  options.write_buffer_size = 1024 * 1024;
+  txn_db_options.skip_concurrency_control = true;
+  printf("two write que: %d, unordered write: %d\n", options.two_write_queues, options.unordered_write);
+  ReOpen();
+  ASSERT_OK(db->Put(WriteOptions(), "1000", "v1"));
```
maysamyabandeh: Can you add inline comment on the rational for the chosen keys in this test?

utilities/transactions/write_prepared_transaction_test.cc
Little-Wallace: ok

utilities/transactions/write_prepared_transaction_test.cc
Little-Wallace: Yes, I do.

maysamyabandeh:
Can you take a look at the travis failures?

Little-Wallace:
@maysamyabandeh  It's OK now. It seems that the travis is not very stable.

PR #5978, opened 2019-10-28
filter_bench not needed for ROCKSDB_LITE

pdillinger:
Summary: filter_bench is a specialized micro-benchmarking tool that
should not be needed with ROCKSDB_LITE. This should fix the LITE build.

Test Plan: make LITE=1 check


pdillinger:
This change is like existing db/forward_iterator_bench.cc, btw

PR #5979, opened 2019-10-29
Support SSTFileWriter constructor with default options

darionyaphet:


java/src/main/java/org/rocksdb/SstFileWriter.java
```diff
@@ -45,6 +45,13 @@ public SstFileWriter(final EnvOptions envOptions, final Options options) {
         envOptions.nativeHandle_, options.nativeHandle_));
   }
 
+  /**
+   * SstFileWriter Constructor.
+   */
+  public SstFileWriter() {
+    this(new EnvOptions(), new Options());
```
adamretter: You can't just create an `Options` in this way, otherwise it will never be cleaned up (deleted).

darionyaphet: @adamretter thanks for you point it out. It seem we could not support a constructor without argument ?

adamretter: That's right, you cannot have a zero-arg constructor here. You need at least `EnvOptions` and `Options` as set out by the underlying C++ API.

adamretter:
@darionyaphet I am closing due to the issue explained in my comments.

PR #6008, opened 2019-11-04
Add clarifying/instructive header to TARGETS and defs.bzl

pdillinger:


buckifier/targets_cfg.py
```diff
@@ -3,7 +3,12 @@
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
-rocksdb_target_header = """load("@fbcode_macros//build_defs:auto_headers.bzl", "AutoHeaders")
+rocksdb_target_header = """# This file @generated by `python buckifier/buckify_rocksdb.py`
```
riversand963: IIRC, there should be an empty line between imports and variable definitions.

Issue #6012, opened 2019-11-07
OpenAsSecondaryWALTailing in rocksdb_db_secondary_test  Failed


CAIQT:
### Expected behavior
Pass test. (Built with vs2017, running in Win10 x64)

### Actual behavior
Raise Exception.  Program crashed.
![Untitled](https://user-images.githubusercontent.com/3953402/68364596-666f8580-0169-11ea-9dc9-3336b8785b86.png)

### Steps to reproduce the behavior
build with vs2017 with debug configuration, and simply run **rocksdb_db_secondary_test**

CAIQT:
I found that it may caused by the following codes in function _RemoveOldMemTables()_ in **db/memtable_list.cc**  :
`
  for (auto it = memlist.rbegin(); it != memlist.rend(); ++it) {
    MemTable* mem = *it;
    if (mem->GetNextLogNumber() > log_number) {
      break;
    }
    current_->Remove(mem, to_delete);
`

The iterator **it** turns invalid after `current_->Remove(mem, to_delete);`

riversand963:
PR is under review. Let's migrate our discussions to PR :). I will be closing this issue to avoid duplication. Thanks @CAIQT for reporting the issue.

PR #6013, opened 2019-11-07
bugfix: MemTableList::RemoveOldMemTables invalid iterator after remov…

CAIQT:
Fix issue #6012.

I found that it may be caused by the following codes in function _RemoveOldMemTables()_ in **db/memtable_list.cc**  :
```
  for (auto it = memlist.rbegin(); it != memlist.rend(); ++it) {
    MemTable* mem = *it;
    if (mem->GetNextLogNumber() > log_number) {
      break;
    }
    current_->Remove(mem, to_delete);
```

The iterator **it** turns invalid after `current_->Remove(mem, to_delete);` 

Test plan:
```
make check
```

### DIFF
--- a/db/memtable_list.cc
+++ b/db/memtable_list.cc
@@ -742,7 +742,7 @@ void MemTableList::RemoveOldMemTables(uint64_t log_number,
     if (mem->GetNextLogNumber() > log_number) {
       break;
     }
-    old_memtables.push_back(mem);    
+    old_memtables.push_back(mem);
   }
 
   for (auto it = old_memtables.begin(); it!= old_memtables.end(); ++it) {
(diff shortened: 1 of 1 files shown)


db/memtable_list.cc
```diff
@@ -736,17 +736,24 @@ void MemTableList::RemoveOldMemTables(uint64_t log_number,
   assert(to_delete != nullptr);
   InstallNewVersion();
   auto& memlist = current_->memlist_;
+  std::list<MemTable*> to_delete_memlist;
```
riversand963: Can we use `autovector` defined in `util/autovector.h`?

CAIQT: Yes, autovector can be use here.

PR #6014, opened 2019-11-07
Fix a buffer overrun problem in BlockBasedTable::MultiGet

anand1976:
The calculation in BlockBasedTable::MultiGet for the required buffer length for reading in compressed blocks is incorrect. It needs to take the 5-byte block trailer into account.

Test plan:
Add a unit test DBBasicTest.MultiGetBufferOverrun that fails in asan_check before the fix, and passes after.

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -2354,11 +2354,11 @@ void BlockBasedTable::RetrieveMultipleBlocks(
     ReadRequest& req = read_reqs[read_req_idx++];
     Status s = req.status;
     if (s.ok()) {
-      if (req.result.size() != block_size(handle)) {
+      if (req.result.size() != req.len) {
         s = Status::Corruption("truncated block read from " +
                                rep_->file->file_name() + " offset " +
                                ToString(handle.offset()) + ", expected " +
-                               ToString(block_size(handle)) +
+                               ToString(req.len) +
                                " bytes, got " + ToString(req.result.size()));
       }
     }
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -496,7 +496,7 @@ class PartitionIndexReader : public BlockBasedTable::IndexReaderCommon {
       return;
     }
     handle = biter.value().handle;
-    uint64_t last_off = handle.offset() + handle.size() + kBlockTrailerSize;
+    uint64_t last_off = handle.offset() + block_size(handle);
     uint64_t prefetch_len = last_off - prefetch_off;
     std::unique_ptr<FilePrefetchBuffer> prefetch_buffer;
     auto& file = rep->file;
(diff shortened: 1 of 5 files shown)


table/block_based/block_based_table_reader.cc
```diff
@@ -2351,11 +2351,11 @@ void BlockBasedTable::RetrieveMultipleBlocks(
     ReadRequest& req = read_reqs[read_req_idx++];
     Status s = req.status;
     if (s.ok()) {
-      if (req.result.size() != handle.size() + kBlockTrailerSize) {
+      if (req.result.size() != block_size(handle)) {
```
ltamasi: I think we could use `req.len` here to make the intent clear, right?

table/block_based/block_based_table_reader.cc
```diff
@@ -2351,11 +2351,11 @@ void BlockBasedTable::RetrieveMultipleBlocks(
...
         s = Status::Corruption("truncated block read from " +
                                rep_->file->file_name() + " offset " +
                                ToString(handle.offset()) + ", expected " +
-                               ToString(handle.size() + kBlockTrailerSize) +
+                               ToString(block_size(handle)) +
```
ltamasi: And here.

table/block_based/block_based_table_reader.cc
```diff
@@ -2910,7 +2910,7 @@ void BlockBasedTableIterator<TBlockIter, TValue>::InitDataBlock() {
           if (!rep->file->use_direct_io() &&
               (data_block_handle.offset() +
                    static_cast<size_t>(data_block_handle.size()) +
-                   kBlockTrailerSize >
+                    kBlockTrailerSize >
```
ltamasi: How about `block_size(data_block_handle)` here?

table/block_based/block_based_table_reader.h
```diff
@@ -461,8 +461,18 @@ class BlockBasedTable : public TableReader {
   void DumpKeyValue(const Slice& key, const Slice& value,
                     WritableFile* out_file);
 
+  // A cumulative data block file read in MultiGet lower than this size will
+  // use a stack buffer
+  static const size_t kMultiGetReadStackBufSize = 8192;
```
ltamasi: We could consider making this `static constexpr`.

table/block_based/block_based_table_reader.cc
anand1976: The definition is not available here (BlockBasedTableIterator). Maybe a follow-on PR. It'd be good to use a helper function in all places where kBlockTrailerSize is referenced.

ltamasi: `block_size` could be a public helper in `BlockBasedTable`, right? On the other hand, if you're going to leave this as-is, could you please revert the whitespace change?

table/block_based/block_based_table_reader.h
```diff
@@ -461,8 +461,18 @@ class BlockBasedTable : public TableReader {
...
+
+  // Make block size calculation for IO less error prone
+  uint64_t block_size(const BlockHandle& handle) const {
```
ltamasi: I think this should be marked `static`.

table/block_based/block_based_table_reader.cc
anand1976: I like the public helper idea. It could be in format.h and available to all.

PR #6043, opened 2019-11-17
Compatible changes with g++4.9.2 and cmake.

merryChris:


CMakeLists.txt
```diff
@@ -1070,6 +1069,7 @@ if(WITH_TESTS)
   add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND})
   set(TESTUTILLIB testutillib${ARTIFACT_SUFFIX})
   add_library(${TESTUTILLIB} STATIC ${TESTUTIL_SOURCE})
+  target_link_libraries(${TESTUTILLIB} ${LIBS})
```
tchaikov: @merryChris this makes sense, as some source files (`db/db_test_util.h` for instance) in `${TESTUTILLIB}` actually references headers used by `ROCKSDB_STATIC_LIB/ROCKSDB_SHARED_LIB`, but  i'd suggest move the cmake changes into a separated commit. and explain the reason why we need this change in the commit message.

cmake/modules/ReadVersion.cmake
```diff
@@ -1,7 +1,7 @@
 # Read rocksdb version from version.h header file.
 
 function(get_rocksdb_version version_var)
-  file(READ "${CMAKE_SOURCE_DIR}/include/rocksdb/version.h" version_header_file)
+  file(READ "${CMAKE_CURRENT_SOURCE_DIR}/include/rocksdb/version.h" version_header_file)
```
tchaikov: why?

db/compaction/compaction_iterator_test.cc
```diff
@@ -184,7 +184,7 @@ class TestSnapshotChecker : public SnapshotChecker {
  public:
   explicit TestSnapshotChecker(
       SequenceNumber last_committed_sequence,
-      const std::unordered_map<SequenceNumber, SequenceNumber>& snapshots = {})
+      const std::unordered_map<SequenceNumber, SequenceNumber>& snapshots = {{}})
```
tchaikov: please move this change to a separated commit, and explain the rationale behind this change.

util/filter_bench.cc
```diff
@@ -606,7 +606,8 @@ double FilterBench::RandomQueryTest(uint32_t inside_threshold, bool dry_run,
   }
 
   if (!dry_run) {
-    fp_rate_report_ = std::ostringstream();
+    fp_rate_report_.clear();
+    fp_rate_report_.str("");
```
tchaikov: ditto.

CMakeLists.txt
merryChris: Fine, I will pick up the cmake changes in a separated commit. Close this pr temporarily.

cmake/modules/ReadVersion.cmake
merryChris: Since I'm trying to build a distributed storage system based on rocksdb, and my codebase directory structure just like this.

```
/path/to/rocskdb_xxx$ tree -L 1
.
|-build.sh
|-my_code/
|-rocksdb/
|-something_else
```

My project name is rocksdb_xxx. While I'm building it at the root path, the cmake file path above is `/path/to/rocskdb_xxx/include/rocksdb/version.h`, which doesn't work. `${CMAKE_CURRENT_SOURCE_DIR}` is a friendly option path, isn't it?

db/compaction/compaction_iterator_test.cc
merryChris: OK, I will separate them into two pr. Plz refer to the issue https://github.com/facebook/rocksdb/issues/3522, it may not friendly for g++4.9.

util/filter_bench.cc
merryChris: Reson as above, and here is the conversation with @pdillinger https://github.com/facebook/rocksdb/pull/5968.

CMakeLists.txt
tchaikov: @merryChris JFYI, there is no need to create a new PR for addressing comments in existing one. you can always `git push -f` to your remote branch for updating a PR.

cmake/modules/ReadVersion.cmake
tchaikov: i see. presumably you include rocksdb project using `add_subdirectory()` which points `CMAKE_SOURCE_DIR` to `/path/to/rocskdb_xxx`.

db/compaction/compaction_iterator_test.cc
tchaikov: i think the existing code is compliant to C++11. and why would you want to compile compile the test with [4 years old compiler](https://gcc.gnu.org/gcc-4.9/)?

CMakeLists.txt
merryChris: @tchaikov thanks for reminding.

cmake/modules/ReadVersion.cmake
merryChris: yeah, that's it.

db/compaction/compaction_iterator_test.cc
merryChris: Sigh, it seems a sad history problem. If you would not like to make a compatibility, it's okay to make a local fit for me.

tchaikov: i think it's the maintainers' call to include your fix or not. i am just wondering why. the reason i was curious is that, normally, the reason why we need to stick to an older version of compiler is just we have to compile our executables/libraries for supported older platforms/distros. but the tests is not in this scope.

CMakeLists.txt
merryChris: @tchaikov I am not sure why I cannot update PR automatically. Sorry for that, and plz have a review at https://github.com/facebook/rocksdb/pull/6045.

db/compaction/compaction_iterator_test.cc
merryChris: Briefly speaking, almost all projects/libs in our company are complied with g++4.9.2. As for me, I have to maintain compatible api lib version under g++4.9.2, 5.4.0, and 6.3.0, which also troubles me a lot. But this is life anyway. ;)

tchaikov: but `compaction_iterator_test.cc` is not included by rocksdb library. so even if it does not compile, you can still compile and link against rocksdb using GCC 4.9.2 .

merryChris: Since we want to make use of built-in `db_bench_tool` as data generator, with the option `WITH_TESTS=ON`. But as you see, `${TESTS}` sources like `compaction_iterator_test` is involved.

pdillinger:
If you get the cmake updates out of this PR (now in #6045) I'll pull in the compaction_iterator_test.cc and filter_bench.cc updates.

db/compaction/compaction_iterator_test.cc
pdillinger: https://stackoverflow.com/questions/26947704/implicit-conversion-failure-from-initializer-list

tchaikov: @pdillinger thanks for the link! so it's C++14 compliant.

merryChris:
> If you get the cmake updates out of this PR (now in #6045) I'll pull in the compaction_iterator_test.cc and filter_bench.cc updates.

That's it. Do I need to separate these two changes into another PR, since sorry that I cannot update this PR anymore?

pdillinger:
Moved to #6053

merryChris:
> Moved to #6053

Thanks a lot.

PR #6047, opened 2019-11-18
Add operator[] to autovector::iterator_impl.

StephanTLavavej:
This is a required operator for random-access iterators, and an upcoming update for Visual Studio 2019 will change the C++ Standard Library's heap algorithms to use this operator.

### DIFF
--- a/util/autovector.h
+++ b/util/autovector.h
@@ -140,6 +140,14 @@ class autovector {
       return &(*vect_)[index_];
     }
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
I've filed #6048 regarding the `const`-overloading in this iterator (which doesn't conform to the C++ Standard's requirements and can cause problems in generic algorithms). In this PR, I have `const`-overloaded the `operator[]` for consistency with the existing code.

PR #6051, opened 2019-11-19
Fix blob context when db_iter uses seek

tabokie:
Fix: when `db_iter` falls back to using seek by `FindValueForCurrentKeyUsingSeek`, `is_blob_` flag is not properly set on encountering BlobIndex.
Also patch existing test for the mentioned code path.
Signed-off-by: tabokie <xy.tao@outlook.com>

### DIFF
--- a/db/db_blob_index_test.cc
+++ b/db/db_blob_index_test.cc
@@ -398,6 +398,26 @@ TEST_F(DBBlobIndexTest, Iterate) {
     verify(15, Status::kOk, get_value(16, 0), get_value(14, 0),
            create_blob_iterator, check_is_blob(false));
 
+    // Iterator with blob support and using seek.
+    ASSERT_OK(dbfull()->SetOptions(cfh(), {{"max_sequential_skip_in_iterations", "0"}}));
+    verify(1, Status::kOk, get_value(1, 0), get_value(1, 0),
+           create_blob_iterator, check_is_blob(true));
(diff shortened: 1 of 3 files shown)


yiwu-arbug:
@ltamasi PTAL. This is a BlobDB bug.

db/db_blob_index_test.cc
```diff
@@ -398,6 +398,26 @@ TEST_F(DBBlobIndexTest, Iterate) {
     verify(15, Status::kOk, get_value(16, 0), get_value(14, 0),
            create_blob_iterator, check_is_blob(false));
 
+    // Iterator with blob support and using seek.
```
yiwu-arbug: how do you make sure `FindValueForCurrentKeyUsingSeek` is called in this test?

yiwu-arbug:
@tabokie please also update HISTORY.md to mention the fix.

db/db_blob_index_test.cc
tabokie: From this [line](https://github.com/facebook/rocksdb/blob/3b3b96d6933c64b9c47dba19020e0dcce214c668/db/db_iter.cc#L724), we know that when `max_skip_==0` backward iteration must call `FindValueForCurrentKeyUsingSeek` to retrieve value. I verify the case by running the test without the fix, and it failed.

PR #6061, opened 2019-11-21
fix  unstable unittest caused by #5958

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

This PR is to fix unstable unit test added by  (https://github.com/facebook/rocksdb/pull/5958).
I set SYNC_POINT in PickCompaction before. If IntraL0Compaction was trigger,  the compact job which compact sst to base level would start instantly. If the compaction thread run faster than unittest main thread, we may observe the number of files in L0 reduce.


anand1976:
Cc @Little-Wallace This test is still flaky in our internal FB tests. It fails occasionally as follows -

```
[ RUN      ] DBCompactionTestWithParam/DBCompactionTestWithParam.FlushAfterIntraL0CompactionCheckConsistencyFail/0
db/db_compaction_test.cc:5014: Failure
Expected: (pick_intra_l0_count.load()) > (0), actual: 0 vs 0
```

PR #6067, opened 2019-11-22
Fix UniversalCompaction trivial move bug

mm304321141:
`curr.level` is `c->inputs_` index, not real level.


mm304321141:
@siying pls ...

PR #6069, opened 2019-11-22
Perform post-flush updates of memtable list in a callback

riversand963:
Currently, the following interleaving of events can lead to SuperVersion containing both immutable memtables as well as the resulting L0. This can cause Get to return incorrect result if there are merge operands. This may also affect other operations such as single deletes.

```
  time  main_thr  bg_flush_thr  bg_compact_thr  compact_thr  set_opts_thr
0  |                                                         WriteManifest:0
1  |                                           issue compact
2  |                                 wait
3  |   Merge(counter)
4  |   issue flush 
5  |                   wait
6  |                                                         WriteManifest:1
7  |                                 wake up
8  |                                 write manifest
9  |                  wake up
10 |  Get(counter)
11 |                  remove imm
   V
```

The reason behind is that: one bg flush thread's installing new `Version` can be batched and performed by another thread that is the "leader" MANIFEST writer. This bg thread removes the memtables from current super version only after `LogAndApply` returns. After the leader MANIFEST writer signals (releasing mutex) this bg flush thread, it is possible that another thread sees this cf with both memtables (whose data have been flushed to the newest L0) and the L0 before this bg flush thread removes the memtables.

To address this issue, each bg flush thread can pass a callback function to `LogAndApply`. The callback is responsible for removing the memtables. Therefore, the leader MANIFEST writer can call this callback and remove the memtables before releasing the mutex.

Test plan (devserver)
```
$make merge_test
$./merge_test --gtest_filter=MergeTest.MergeWithCompactionAndFlush
$make check
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,6 +7,7 @@
 * Since 6.14, fix false positive flush/compaction `Status::Corruption` failure when `paranoid_file_checks == true` and range tombstones were written to the compaction output files.
 * Since 6.14, fix a bug that could cause a stalled write to crash with mixed of slowdown and no_slowdown writes (`WriteOptions.no_slowdown=true`).
 * Fixed a bug which causes hang in closing DB when refit level is set in opt build. It was because ContinueBackgroundWork() was called in assert statement which is a no op. It was introduced in 6.14.
+* Perform post-flush updates of in-memory data structures of memtable list in a callback that will be called by the thread writing to MANIFEST file before signaling other threads and releasing db mutex.
 
 ### Public API Change
 * Deprecate `BlockBasedTableOptions::pin_l0_filter_and_index_blocks_in_cache` and `BlockBasedTableOptions::pin_top_level_index_and_filter`. These options still take effect until users migrate to the replacement APIs in `BlockBasedTableOptions::metadata_cache_options`. Migration guidance can be found in the API comments on the deprecated options.
--- a/db/memtable_list.cc
+++ b/db/memtable_list.cc
@@ -481,83 +481,18 @@ Status MemTableList::TryInstallMemtableFlushResults(
             vset, *cfd, edit_list, memtables_to_flush, prep_tracker));
       }
 
+      const auto manifest_write_cb = [this, cfd, batch_count, log_buffer,
+                                      to_delete](const Status& status) {
+        RemoveMemTablesOrRestoreFlags(status, cfd, batch_count, log_buffer,
+                                      to_delete);
+      };
+
       // this can release and reacquire the mutex.
       s = vset->LogAndApply(cfd, mutable_cf_options, edit_list, mu,
-                            db_directory);
+                            db_directory, /*new_descriptor_log=*/false,
+                            /*column_family_options=*/nullptr,
+                            manifest_write_cb);
       *io_s = vset->io_status();
-
-      // we will be changing the version in the next code path,
-      // so we better create a new one, since versions are immutable
-      InstallNewVersion();
-
-      // All the later memtables that have the same filenum
-      // are part of the same batch. They can be committed now.
-      uint64_t mem_id = 1;  // how many memtables have been flushed.
-
-      // commit new state only if the column family is NOT dropped.
-      // The reason is as follows (refer to
-      // ColumnFamilyTest.FlushAndDropRaceCondition).
-      // If the column family is dropped, then according to LogAndApply, its
-      // corresponding flush operation is NOT written to the MANIFEST. This
-      // means the DB is not aware of the L0 files generated from the flush.
-      // By committing the new state, we remove the memtable from the memtable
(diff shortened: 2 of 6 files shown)


riversand963:
Cleaned up from #6064

mm304321141:
@riversand963 
https://github.com/riversand963/rocksdb/pull/3

> We wrote some code illustrating this SuperVersion bug.
> By the way, we do have a trivial fix of the bug.
> Since it's not elegant, we are expecting your recommendations about this bug.

create a pull to show our trivial fix

riversand963:
> @riversand963
> [riversand963#3](https://github.com/riversand963/rocksdb/pull/3)
> 
> > We wrote some code illustrating this SuperVersion bug.
> > By the way, we do have a trivial fix of the bug.
> > Since it's not elegant, we are expecting your recommendations about this bug.
> 
> create a pull to show our trivial fix ...
> it works ...

I am hesitant about adding a member variable, 'callback' of type `std::function` to `VersionEdit`. It is true that we can distinguish data that should be serialized and persisted from volatile data, but not sure if it is ideal.

mm304321141:
> > @riversand963
> > [riversand963#3](https://github.com/riversand963/rocksdb/pull/3)
> > > We wrote some code illustrating this SuperVersion bug.
> > > By the way, we do have a trivial fix of the bug.
> > > Since it's not elegant, we are expecting your recommendations about this bug.
> > 
> > 
> > create a pull to show our trivial fix ...
> > it works ...
> 
> I am hesitant about adding a member variable, 'callback' of type `std::function` to `VersionEdit`. It is true that we can distinguish data that should be serialized and persisted from volatile data, but not sure if it is ideal.

Our online service needs to be fixed right away :(
The fix is rude and I don't prefer it ...
So we just report this issue to you (official) -.-

Adding a callback to `LogAndApply` is the same workaround ...
That I think adding callback to `VersionEdit` is more adaptable (in the future?).

db/memtable_list.cc
```diff
@@ -466,60 +466,19 @@ Status MemTableList::TryInstallMemtableFlushResults(
...
       }
 
+      auto manifest_write_cb = std::bind(
+          [this, cfd, batch_count, log_buffer,
+           to_delete](const Status& status) {
+            RemoveMemTablesOrRestoreFlags(status, cfd, batch_count, log_buffer,
+                                          to_delete);
+          },
+          std::placeholders::_1);
+
```
rockeet: The `std::bind` call is redundant.

ajkr:
Can this bug also cause issue for `SingleDelete()`? For example, if a SINGLEDEL is in the mutable memtable, and the one PUT it covers is visible in both an immutable memtable and an L0 file, then I'd expect `Get()` to return the PUT from the L0 file.

riversand963:
> Can this bug also cause issue for `SingleDelete()`? For example, if a SINGLEDEL is in the mutable memtable, and the one PUT it covers is visible in both an immutable memtable and an L0 file, then I'd expect `Get()` to return the PUT from the L0 file.

I think it's possible, though I have not used sync points to verify.

db/merge_test.cc
```diff
@@ -488,6 +580,19 @@ TEST_F(MergeTest, MergeDbTtlTest) {
   runTest(test::PerThreadDBPath("merge_testdbttl"),
           true);  // Run test on TTL database
 }
+
+TEST_F(MergeTest, MergeWithCompactionAndFlush) {
```
cheng-chang, line 584: The bug may also affect delete? Can we add a test case for delete too?

cheng-chang (approved):

db/merge_test.cc
riversand963: It should not. Can you elaborate on how delete can be affected?

riversand963:
Thanks @cheng-chang for the review!

HISTORY.md
```diff
@@ -7,6 +7,7 @@
 * Since 6.14, fix false positive flush/compaction `Status::Corruption` failure when `paranoid_file_checks == true` and range tombstones were written to the compaction output files.
 * Since 6.14, fix a bug that could cause a stalled write to crash with mixed of slowdown and no_slowdown writes (`WriteOptions.no_slowdown=true`).
 * Fixed a bug which causes hang in closing DB when refit level is set in opt build. It was because ContinueBackgroundWork() was called in assert statement which is a no op. It was introduced in 6.14.
+* Perform post-flush updates of in-memory data structures of memtable list in a callback that will be called by the thread writing to MANIFEST file before signaling other threads and releasing db mutex.
```
ajkr, line 10: When looking at the release notes, I wasn't able to think of the user-visible bug that this referred to. Only by looking at the PR I learned it's a `Get()` correctness issue. Is it possible to write the note more in terms of what a user would observe?

riversand963: Sure, will submit a PR to address your comment.

PR #6080, opened 2019-11-24
Fix naming of library on PPC64LE

adamretter:
**NOTE**: This also needs to be back-ported to be 6.4.6

Fix a regression introduced in f2bf0b2 by https://github.com/facebook/rocksdb/pull/5674 whereby the compiled library would get the wrong name on PPC64LE platforms.

On PPC64LE, the regression caused the library to be named `librocksdbjni-linux64.so` instead of `librocksdbjni-linux-ppc64le.so`.

This PR corrects the name back to `librocksdbjni-linux-ppc64le.so` and also corrects the ordering of conditional arguments in the Makefile to match the expected order as defined in the documentation for Make.


adamretter:
Note that the CI on the branches is failing because there is no local branch. Travis seems to have got confused as I initially pushed to github.com/rocksdb but then moved the branch to github.com/adamretter.

PR #6081, opened 2019-11-24
Fix compilation under MSVC VS2015

adamretter:
**NOTE**: this also needs to be back-ported to 6.4.6 and possibly older branches if further releases from them is envisaged.

### DIFF
--- a/tools/block_cache_analyzer/block_cache_trace_analyzer.cc
+++ b/tools/block_cache_analyzer/block_cache_trace_analyzer.cc
@@ -14,6 +14,7 @@
 #include <fstream>
 #include <iomanip>
 #include <iostream>
+#include <memory>
 #include <random>
 #include <sstream>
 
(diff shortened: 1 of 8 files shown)


tools/block_cache_analyzer/block_cache_trace_analyzer.cc
```diff
@@ -1209,11 +1209,11 @@ void BlockCacheTraceAnalyzer::WriteBlockReuseTimeline(
   TraverseBlocks(block_callback);
 
   // A cell is the number of blocks accessed in a reuse window.
-  uint64_t reuse_table[reuse_vector_size][reuse_vector_size];
+  uint64_t* reuse_table = new uint64_t[reuse_vector_size * reuse_vector_size];
```
zhichao-cao: Is it necessary to replace the stack space with heap? If so, it would be better to use unique_ptr to avoid potential memory leak.

adamretter: @zhichao-cao Agreed. I have made that change now.

PR #6089, opened 2019-11-25
Merge adjacent file block reads in RocksDB MultiGet() and Add uncompressed block to cache

zhichao-cao:
In the current MultiGet, if the KV-pairs do not belong to the data blocks in the block cache, multiple blocks are read from a SST. It will trigger one block read for each block request and read them in parallel. In some cases, if some data blocks are adjacent in the SST, the reads for these blocks can be combined to a single large read, which can reduce the system calls and reduce the read latency if possible.

Considering to fill the block cache, if multiple data blocks are in the same memory buffer, we need to copy them to the heap separately. Therefore, only in the case that 1) data block compression is enabled, and 2) compressed block cache is null, we can do combined read. Otherwise, extra memory copy is needed, which may cause extra overhead. In the current case, data blocks will be uncompressed to a new memory space.

Also, in the case that 1) data block compression is enabled, and 2) compressed block cache is null, it is possible the data block is actually not compressed. In the current logic, these data blocks will not be added to the uncompressed_cache. So if memory buffer is shared and the data block is not compressed, the data block are copied to the head and fill the cache.

Test plan: Added test case to ParallelIO.MultiGet. Pass make asan_check

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -9,6 +9,7 @@
 #include "table/block_based/block_based_table_reader.h"
 #include <algorithm>
 #include <array>
+#include <iostream>
 #include <limits>
 #include <string>
 #include <utility>
@@ -2457,13 +2458,14 @@ void BlockBasedTable::RetrieveMultipleBlocks(
       // to the heap such that it can be added to the block cache.
       CompressionType compression_type =
           raw_block_contents.get_compression_type();
-      if (rep_->blocks_maybe_compressed && compression_type == kNoCompression) {
-      }
-      if (blocks_share_scratch && compression_type == kNoCompression) {
+      if (rep_->blocks_maybe_compressed &&
+          rep_->table_options.block_cache_compressed == nullptr &&
+          compression_type == kNoCompression && blocks_share_scratch) {
         Slice raw = Slice(req.scratch + req_offset, handle.size());
         raw_block_contents = BlockContents(
             CopyBufferToHeap(GetMemoryAllocator(rep_->table_options), raw),
             handle.size());
+
 #ifndef NDEBUG
         raw_block_contents.is_raw_block = true;
 #endif
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -2458,13 +2458,12 @@ void BlockBasedTable::RetrieveMultipleBlocks(
       // to the heap such that it can be added to the block cache.
       CompressionType compression_type =
           raw_block_contents.get_compression_type();
-      if (rep_->blocks_maybe_compressed &&
-          rep_->table_options.block_cache_compressed == nullptr &&
-          compression_type == kNoCompression && blocks_share_scratch) {
+      if (scratch != nullptr && compression_type == kNoCompression) {
         Slice raw = Slice(req.scratch + req_offset, handle.size());
         raw_block_contents = BlockContents(
             CopyBufferToHeap(GetMemoryAllocator(rep_->table_options), raw),
             handle.size());
+        std::cout<<req.result.size()<<" "<<handle.size()<<" rare case, compress enabled but not block not compressed\n";
 
 #ifndef NDEBUG
         raw_block_contents.is_raw_block = true;
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -2504,12 +2504,12 @@ void BlockBasedTable::RetrieveMultipleBlocks(
                                     handle.size(), &contents, footer.version(),
                                     rep_->ioptions, memory_allocator);
       } else {
-        if (scratch != nullptr || block_size(handle) != req.result.size()) {
+        if (scratch != nullptr || blocks_share_scratch) {
           // If we used the scratch buffer, then the contents need to be
           // copied to heap
           // If the read buffer is shared but the block is not compressed, copy
           // to the head
-          Slice raw = Slice(req.result.data(), handle.size());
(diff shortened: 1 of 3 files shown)


riversand963:
Assigning on behalf of @zhichao-cao who currently has no access.

table/block_based/block_based_table_reader.cc
```diff
@@ -2351,41 +2393,80 @@ void BlockBasedTable::RetrieveMultipleBlocks(
...
 #ifndef NDEBUG
       raw_block_contents.is_raw_block = true;
 #endif
       if (options.verify_checksums) {
         PERF_TIMER_GUARD(block_checksum_time);
         const char* data = req.result.data();
-        uint32_t expected = DecodeFixed32(data + handle.size() + 1);
-        s = rocksdb::VerifyChecksum(footer.checksum(), req.result.data(),
+        uint32_t expected =
+            DecodeFixed32(data + req_offset + handle.size() + 1);
```
anand1976: Add a comment here explaining the calculation.

table/block_based/block_based_table_reader.cc
```diff
@@ -2351,41 +2393,80 @@ void BlockBasedTable::RetrieveMultipleBlocks(
...
+        s = rocksdb::VerifyChecksum(footer.checksum(),
+                                    req.result.data() + req_offset,
                                     handle.size() + 1, expected);
       }
     }
+
+    if (s.ok()) {
```
anand1976: Shouldn't this whole block be guarded by ```rep_->blocks_maybe_compressed && !rep_->table_options.block_cache_compressed```?

table/block_based/block_based_table_reader.cc
```diff
@@ -2351,41 +2393,80 @@ void BlockBasedTable::RetrieveMultipleBlocks(
...
+      // It handles a rare case: compression is set and these is no compressed
+      // cache (enable combined read), some block is actually not compressed
+      // since its compression space saving is smaller than the threshold. In
+      // this case, if the block shares the scratch memory, we need to copy it
+      // to the heap such that it can be added to the block cache.
+      CompressionType compression_type =
+          raw_block_contents.get_compression_type();
+      if ((scratch != nullptr && compression_type == kNoCompression) ||
+          (scratch == nullptr && blocks_share_scratch)) {
```
anand1976: See previous comment. Alternatively, you could check for ```compression_type == kNoCompression``` here.

table/block_based/block_based_table_reader.cc
```diff
@@ -2351,41 +2393,80 @@ void BlockBasedTable::RetrieveMultipleBlocks(
...
     BlockContents raw_block_contents;
+    size_t cur_read_end = req_offset + block_size(handle);
+    if (cur_read_end > req.result.size()) {
+      s = Status::Corruption(
+          "truncated block read from " + rep_->file->file_name() + " offset " +
+          ToString(handle.offset()) + ", expected " + ToString(req.len) +
+          " bytes, got " + ToString(req.result.size()));
+    }
+
+    bool blocks_share_scratch = (req.result.size() != block_size(handle));
```
anand1976: ```blocks_share_scratch``` name is a bit confusing, since this can be true even if the ```scratch``` argument is null. Maybe ```blocks_share_buffer``` is more appropriate?

table/block_based/block_based_table_reader.cc
```diff
@@ -2319,30 +2320,71 @@ void BlockBasedTable::RetrieveMultipleBlocks(
...
+      }
+
+      // Step 2, remeber the previous block info
+      prev_offset = handle.offset();
+      prev_len = block_size(handle);
+      req_offset_for_block.emplace_back(0);
+    }
+    req_idx_for_block.emplace_back(read_reqs.size());
+  }
+  // Hanle the last block and process the pending last request
```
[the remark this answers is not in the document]
zhichao-cao: corrected

table/block_based/block_based_table_reader.cc
zhichao-cao: rewrite the condition, scratch!=nullptr ensures the condition

table/block_based/block_based_table_reader.cc
```diff
@@ -2319,30 +2320,71 @@ void BlockBasedTable::RetrieveMultipleBlocks(
...
     if (handle.IsNull()) {
       continue;
     }
 
+    size_t prev_end = static_cast<size_t>(prev_offset) + prev_len;
+
+    // If current block is adjacent to the previous one, at the same time,
+    // compression is enabled and there is no compressed cache, we combine
+    // the two block read as one.
+    if (rep_->blocks_maybe_compressed &&
```
anand1976: Can we change this to ```scratch != nullptr && prev_end == handle.offset()```? Since the caller would have already checked, let's eliminate the code duplication and the risk of the two getting out of sync.

table/block_based/block_based_table_reader.cc
```diff
@@ -2351,41 +2393,86 @@ void BlockBasedTable::RetrieveMultipleBlocks(
...
       }
     }
+
+    if (s.ok()) {
+      // It handles a rare case: compression is set and these is no compressed
+      // cache (enable combined read). In this case, the scratch != nullptr.
+      // At the same time, some blocks are actually not compressed,
+      // since its compression space saving is smaller than the threshold. In
+      // this case, if the block shares the scratch memory, we need to copy it
+      // to the heap such that it can be added to the regular elock cache.
```
anand1976: Typo "elock" -> "block"

table/block_based/block_based_table_reader.cc
zhichao-cao: @anand1976  I will correct them before landing. Thanks for reviewing the PR!

PR #6094, opened 2019-11-26
Fix BlobDB compilation on older GCC versions

adamretter:


adamretter:
@ltamasi Done :-)

ltamasi:
@adamretter There's actually one (newly introduced) place in the code where `BlobFile::compression()` is called; could you update that one as well? I assume that would take care of the build issues.

adamretter:
@ltamasi Okay that's done now too :-) Thanks

PR #6095, opened 2019-11-26
Improve detection and recovery from incomplete backup deletion

sebastianopeluso:
Prioritize deletion of canary file on DeleteBackup and PurgeOldBackups to detect partial deletions.

Deletion and GC of backups is supported by a new canary file. Whenever a backup is created, it is associated with a new canary file. In the process of a deletion of a backup, its canary file is the first one to be deleted (even before any metadata). The advantage is that backups with no canary file are safe to be deleted by a garbage collection. Hence garbage collection is faster because it can exclude private folders of backups that are associated with a canary file.

Note: Test case DeleteTmpFiles of BackupableDBTest needs to be deleted because it is not valid with the new logic of deletion and GC.


pdillinger:
Consulting with Sebastiano on this and then partially reviewing the change, I reconsidered whether this change was "worth it," especially with no evidence (so far) of BackupEngine auto-GC being a performance concern. The complexity of the proposed new code carries its own inherent risk and maintenance burden, and confronting that gave me a new appreciation for the relative simplicity and self-correcting behavior of the GC code. So as not to mistake motion for progress, let's keep the hack we have, at least until there's good evidence we need a principled approach.

Should circumstances change, we can resurrect this work.

PR #6098, opened 2019-11-27
Isolate building db_bench from tests with `WITH_BENCHMARK_TOOLS` option.

merryChris:
Summary:
Isolate `db_bench` from building tests, out of respect for the related comments.
Let building tests yields to `WITH_TEST=ON` AND `CMAKE_BUILD_TYPE=Debug` both,
and building `db_bench` yields to `WITH_BENCHMARK_TOOLS=ON`.

Test Plan: cmake -DCMAKE_BUILD_TYPE=Debug/Release -DWITH_TESTS=ON/OFF -DWITH_BENCHMARK_TOOLS=ON/OFF -DWITH_TOOLS=ON/OFF && make

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 #        cd build
 # 3. Run cmake to generate project files for Windows, add more options to enable required third-party libraries.
 #    See thirdparty.inc for more information.
-#        sample command: cmake -G "Visual Studio 15 Win64" -DWITH_GFLAGS=1 -DWITH_SNAPPY=1 -DWITH_JEMALLOC=1 -DWITH_JNI=1 ..
+#        sample command: cmake -G "Visual Studio 15 Win64" -DCMAKE_BUILD_TYPE=Release -DWITH_GFLAGS=1 -DWITH_SNAPPY=1 -DWITH_JEMALLOC=1 -DWITH_JNI=1 ..
 # 4. Then build the project in debug mode (you may want to add /m[:<N>] flag to run msbuild in <N> parallel threads
 #                                          or simply /m to use all avail cores)
 #        msbuild rocksdb.sln
@@ -174,7 +174,7 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -W -Wextra -Wall")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wsign-compare -Wshadow -Wno-unused-parameter -Wno-unused-variable -Woverloaded-virtual -Wnon-virtual-dtor -Wno-missing-field-initializers -Wno-strict-aliasing")
   if(MINGW)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-format")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-format -fno-asynchronous-unwind-tables")
     add_definitions(-D_POSIX_C_SOURCE=1)
   endif()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
@@ -885,7 +885,9 @@ if(NOT WIN32 OR ROCKSDB_INSTALL_ON_WINDOWS)
 endif()
 
 option(WITH_TESTS "build with tests" ON)
-if(WITH_TESTS)
+# For test libraries, utilities, and exes that are build iff WITH_TESTS=ON and
+# in Debug mode. Add test only code that is not #ifdefed for Release here.
+if(WITH_TESTS AND CMAKE_BUILD_TYPE STREQUAL "Debug")
   add_subdirectory(third-party/gtest-1.8.1/fused-src/gtest)
   set(TESTS
         cache/cache_test.cc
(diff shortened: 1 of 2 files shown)


tools/flags.cc
```diff
@@ -0,0 +1,322 @@
...
+    "\t--enable_io_prio\n"
+    "\t--dump_malloc_stats\n"
+    "\t--num_multi_db\n"
+    "db_stress will IGNORE all the options that passed via the options file.");
+
+DEFINE_int32(threads, 1, "Number of concurrent threads to run.");
+
+DEFINE_bool(disable_wal, false, "If true, do not write WAL for write.");
+
+DEFINE_bool(enable_pipelined_write, true,  // db_stress use false as default
```
merryChris: @siying I re-arrange the common options in `flags.cc`, and as you see the most default values are the same. While several of the them differs, I just comment as above. I don't know how to make it better. Do you have more ideas?

CMakeLists.txt
```diff
@@ -884,9 +886,11 @@ if(NOT WIN32 OR ROCKSDB_INSTALL_ON_WINDOWS)
   )
 endif()
 
+add_subdirectory(third-party/gtest-1.8.1/fused-src/gtest)
 option(WITH_TESTS "build with tests" ON)
-if(WITH_TESTS)
-  add_subdirectory(third-party/gtest-1.8.1/fused-src/gtest)
+# For test libraries, utilities, and exes that are build iff WITH_TESTS=ON and
+# in Debug mode. Add test only code that is not #ifdefed for Release here.
+if(WITH_TESTS AND CMAKE_BUILD_TYPE STREQUAL "Debug")
```
merryChris: @siying @riversand963 Finally passed the ci checks, perhaps you may confuse about `CMakeLists.txt` changes. To be brief, I'm not convinced of including benchmark part in test part. As a rocksdb user, I'm trying to build in release mode along with `db_bench`, but I **HAVE TO** work with the option `WITH_TESTS=ON`, while comments like _For test util library that is build only in DEBUG mode_ (maybe `CMAKE_BUILD_TYPE=Debug`)diverge with it at the same time. So I prefer to isolate the benchmark part from the test part with a new option `WITH_BENCHMARK_TOOLS`, and make use of release/debug mode with respect to the following comments. Could you plz have a review?

merryChris:
@siying @riversand963 Could you plz review this PR, or some suggestions for better solution? Thx in advance.

riversand963:
Thanks @merryChris for the fix. Will take a look.

tools/flags.cc
riversand963: Many of our and (possibly) other users' scripts may rely on the previous default value of such flags to function properly. Adding a comment is not sufficient to remind users to check all these flags with default values.

simpkins:
FYI, I submitted PR #6117 which has just a simple fix for the build by also including db_stress_tool.cc in the db_stress binary, which is closer to what the Makefile-based build does.  That can help fix the CMake build in the short term and could be landed independently from flag refactoring.

merryChris:
```
add_executable(db_stress${ARTIFACT_SUFFIX} db_stress.cc db_stress_tool.cc)
target_link_libraries(db_stress${ARTIFACT_SUFFIX} ${LIBS})
```

@simpkins Cool as it. But I'm curious about why it work fine, since flags conflicts seems haven't been solved between `db_stress_tool.cc` and `${LIBS}`(built from `db_bench_tool.cc`).

merryChris:
@riversand963 Known about your concerns. Certainly, it may not be a perfect solution that abstracting common flags, because of the disadvantages as you say. Just for the purpose of avoiding building the two together, we HAVE TO remove `db_stress.cc`, or fix like #6117.

> Furthermore, you seem to be addressing a similar issue involving db_bench and unit tests. I suggest that it be addressed in a separate PR.

It's my pleasure. And I don't think `BENCHMARKS` should be involved with the option `WITH_TESTS=ON`. PR updates on the way. ;)

merryChris:
@simpkins travis-ci fails, just the same as yours. ^lol^

CMakeLists.txt
```diff
@@ -174,7 +174,7 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -W -Wextra -Wall")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wsign-compare -Wshadow -Wno-unused-parameter -Wno-unused-variable -Woverloaded-virtual -Wnon-virtual-dtor -Wno-missing-field-initializers -Wno-strict-aliasing")
   if(MINGW)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-format")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-format -fno-asynchronous-unwind-tables")
```
riversand963: Curious to know if `-fno-asynchronous-unwind-tables` is intentional here?

merryChris: Yeah, the ci `JOB_NAME=cmake-mingw` will fail without this option. Plz refer to the [comment](https://github.com/facebook/rocksdb/pull/6098#issuecomment-561232056) above. BTW, I'm not a windows user. LoL

riversand963: Looks like this is flaky, but has been reported at https://sourceforge.net/p/mingw-w64/mailman/message/36287627/ and a few other places.

CMakeLists.txt
```diff
@@ -884,9 +884,11 @@ if(NOT WIN32 OR ROCKSDB_INSTALL_ON_WINDOWS)
   )
 endif()
 
+add_subdirectory(third-party/gtest-1.8.1/fused-src/gtest)
```
riversand963: Should we add subdirectory only when we want to build with tests?

CMakeLists.txt
merryChris: Yeah, that's seems a bug of `mingw`, I have to add it in order to pass the ci checks.

CMakeLists.txt
merryChris: Emm, that's fine. At first I put it beforehand just for convenience, since the `gtest` target is depended on `test` and `benchmarks`. I will make fix it to the original.

PR #6104, opened 2019-11-27
Fix a potential bug scheduling unnecessary threads

riversand963:
Summary:
RocksDB should decrement the counter `unscheduled_flushes_` as soon as the bg
thread is scheduled. Before this fix, the counter is decremented only when the
bg thread starts and picks an element from the flush queue. This may cause more
than necessary bg threads to be scheduled. Not a correctness issue, but may
affect flush thread count.

Test Plan:
```
make check
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,4 +1,7 @@
 # Rocksdb Change Log
+## Unreleased
+### Buf Fixes
+* Fix a bug that can cause unnecessary bg thread to be scheduled(#6104).
 ## 6.6.0 (11/25/2019)
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


riversand963:
> Do you remember why it was moved in the first place?

@siying 
I think it was moved to the bg flush thread because we introduced `FlushRequest` that can contain multiple column families. Each time we request a flush, we increase `unscheduled_flushes_` by `flush_request.size()`. Consequently, each time we need to decrease it by `flush_request.size()`. Therefore, we need to peek the flush queue to determine the flush request size. In the bg thread, this  information is readily available, thus I mistakenly moved it there.

PR #4714, opened 2018-11-24
Remove enable_internal_stats

adam-singer:
Simple patch to address comments in [statistics.h#L65](https://github.com/facebook/rocksdb/blob/master/monitoring/statistics.h#L65|statistics.h#L65)  `TODO(ajkr): clean this up since there are no internal stats anymore`

PR #4691, opened 2018-11-16
FIX #3278: Move global const object definitions from .h to .cc

solicomo:
Summary

We should declare constants in headers and define them in source files.
But this commit is only aimed at compound types.

I don't know if it is necessary to do the same thing to fundamental types.

I used this command to find all of the constant definitions in header files.

`find . -name "*.h" | xargs grep -e "^const .*=.*"`

And here is what I found:

```
./db/version_edit.h:const uint64_t kFileNumberMask = 0x3FFFFFFFFFFFFFFF;
./include/rocksdb/env.h:const size_t kDefaultPageSize = 4 * 1024;
./include/rocksdb/statistics.h:const std::vector<std::pair<Tickers, std::string>> TickersNameMap = {
./include/rocksdb/statistics.h:const std::vector<std::pair<Histograms, std::string>> HistogramsNameMap = {
./include/rocksdb/table.h:const uint32_t kPlainTableVariableLength = 0;
./include/rocksdb/utilities/transaction_db.h:const uint32_t kInitialMaxDeadlocks = 5;
./port/port_posix.h:const uint32_t kMaxUint32 = std::numeric_limits<uint32_t>::max();
./port/port_posix.h:const int kMaxInt32 = std::numeric_limits<int32_t>::max();
./port/port_posix.h:const uint64_t kMaxUint64 = std::numeric_limits<uint64_t>::max();
./port/port_posix.h:const int64_t kMaxInt64 = std::numeric_limits<int64_t>::max();
./port/port_posix.h:const size_t kMaxSizet = std::numeric_limits<size_t>::max();
./port/win/port_win.h:const uint32_t kMaxUint32 = UINT32_MAX;
./port/win/port_win.h:const int kMaxInt32 = INT32_MAX;
./port/win/port_win.h:const int64_t kMaxInt64 = INT64_MAX;
./port/win/port_win.h:const uint64_t kMaxUint64 = UINT64_MAX;
./port/win/port_win.h:const size_t kMaxSizet = UINT64_MAX;
./port/win/port_win.h:const size_t kMaxSizet = UINT_MAX;
./port/win/port_win.h:const uint32_t kMaxUint32 = std::numeric_limits<uint32_t>::max();
./port/win/port_win.h:const int kMaxInt32 = std::numeric_limits<int>::max();
./port/win/port_win.h:const uint64_t kMaxUint64 = std::numeric_limits<uint64_t>::max();
[...]

PR #6107, opened 2019-11-28
statistics: make tickers and histograms extendible

Connor1996:
For outer plugins, they may want to store its statistics into Rocksdb's statistics to leverage the cache-friendly optimization. So making `StatisticsImpl` as a template to let outer users reserve a larger place in `StatisticsData`.

The changes are:
- make `StatisticsImpl` as a template
- move implementation to `statistics.h`
- remove internal stats due to https://github.com/facebook/rocksdb/pull/4714
- keep `TickersNameMap` and `HistogramsNameMap` in `statistics.cc` due to https://github.com/facebook/rocksdb/pull/4691


include/rocksdb/statistics.h
```diff
@@ -541,6 +541,8 @@ class Statistics {
 };
 
 // Create a concrete DBStatistics object
+template <uint32_t TICKER_MAX = TICKER_ENUM_MAX,
```
yiwu-arbug: Can we keep the existing API unchanged and add the template version of `CreateDBStatistics` to another header file under include/rocksdb/utilities?

yiwu-arbug:
CI failure seems to be travis error only.

yiwu-arbug:
@siying @miasantreble any comment on this one? This would allow application to reuse the core-local statistics implementation. Thanks.

yiwu-arbug:
ping

PR #6113, opened 2019-12-02
wait pending memtable writes on file ingestion or compact range

### LINKED ISSUE #6026: File ingest may skip necessary memtable flush if unordered_write is enabled
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://www.facebook.com/groups/rocksdb.dev

### Expected behavior
File ingestion job need to flush memtable if key range of the file being ingest overlap with key wrote before the job and haven't flush.

### Actual behavior
When unordered_write is enabled, a key can have seqno assigned and wrote to WAL before an ingestion job starts, and pending writing to memtable. If the ingestion job check memtable key range before the key landing in memtable, the ingestion job may skip the necessary memtable flush, causing data corruption.

See test in https://github.com/yiwu-arbug/rocksdb/commit/27399777f4280b291aa3e3c5733d61a3c62624e8

Connor1996:
**Summary:**
This PR fixes two unordered_write related issues:
- ingestion job may skip the necessary memtable flush. Fixes #6026
- compact range may cause memtable is flushed before pending unordered write finished
    1. `CompactRange` triggers memtable flush but doesn't wait for pending-writes
    2.  there are some pending writes but memtable is already flushed
    3.  the memtable related WAL is removed( note that the pending-writes were recorded in that WAL).
    4.  pending-writes write to newer created memtable
    5. there is a restart  
    6. missing the previous pending-writes because WAL is removed but they aren't included in SST.

**How to solve:**
- Wait pending memtable writes before ingestion job check memtable key range
- Wait pending memtable writes before flush memtable.
**Note that: `CompactRange` calls `RangesOverlapWithMemtables` too without waiting for pending waits, but I'm not sure whether it affects the correctness.**

**Test Plan:**
make check


db/external_sst_file_test.cc
```diff
@@ -1700,6 +1700,43 @@ TEST_F(ExternalSSTFileTest, SstFileWriterNonSharedKeys) {
   ASSERT_OK(DeprecatedAddFile({file_path}));
 }
 
+TEST_F(ExternalSSTFileTest, WithUnorderedWrite) {
```
yiwu-arbug: Thanks for including the test!

db/external_sst_file_test.cc
```diff
@@ -1700,6 +1700,43 @@ TEST_F(ExternalSSTFileTest, SstFileWriterNonSharedKeys) {
...
+  SyncPoint::GetInstance()->EnableProcessing();
+  port::Thread writer1([&]() { Put("bar", "v2"); });
+  port::Thread writer2([&]() {
+    TEST_SYNC_POINT("ExternalSSTFileTest::WithUnorderedWrite:WaitWriteWAL");
+    ASSERT_OK(GenerateAndAddExternalFile(options, {{"bar", "v3"}}, -1,
+                                         true /* allow_global_seqno */));
+  });
+
+  // make sure `Put` and `GenerateAndAddExternalFile` are all called.
+  env_->SleepForMicroseconds(100000);
```
yiwu-arbug: Sleep almost always lead to flaky test. Can we get rid of it? Are we able to use the test from https://github.com/yiwu-arbug/rocksdb/commit/27399777f4280b291aa3e3c5733d61a3c62624e8 as is?

yiwu-arbug:
@maysamyabandeh PTAL. Thanks!

db/external_sst_file_test.cc
Connor1996: No, cause after the fix, `Put` and `GenerateAndAddExternalFile` will block each other due to sync-point `NeedFlush` only can be called after `WaitForPendingWrites`. I tried multiple ways but only found sleeping can solve that.

Connor1996:
one more case, PTAL @maysamyabandeh

db/external_sst_file_test.cc
yiwu-arbug: I think you can add a sync point in the beginning of `WaitForPendingWrites`(say P), and let the "DBImpl::WriteImpl:BeforeUnorderedWriteMemtable" wait for P instead. Remove the "ExternalSSTFileTest::WithUnorderedWrite:ContinueWriteMemtable" sync point. That way you don't need the sleep.

maysamyabandeh:
Thanks for working on this @Connor1996. It LGTM overall.
Did you get a chance to explore @yiwu-arbug's suggested approach to avoid sleep in the unit test? Sleeps are a major long-term pain point in maintaining tests and it would be great to avoid them in the first place.

db/external_sst_file_test.cc
```diff
@@ -1700,6 +1700,43 @@ TEST_F(ExternalSSTFileTest, SstFileWriterNonSharedKeys) {
...
+  TEST_SYNC_POINT(
+      "ExternalSSTFileTest::WithUnorderedWrite:ContinueWriteMemtable");
+
+  writer1.join();
+  writer2.join();
+  ASSERT_EQ("v3", Get("bar"));
```
maysamyabandeh: ditto about the order.

db/external_sst_file_test.cc
Connor1996: if that, after removing the fix, it may still pass the test

Connor1996:
`CompactRange` calls `RangesOverlapWithMemtables` too without waiting for pending waits. Seems that it doesn't affect the correctness, but it may deviate from the intention of `CompactRange`. I'm not sure whether I need to fix that? @maysamyabandeh

db/external_sst_file_test.cc
yiwu-arbug: Removing the fix will remove `WaitForPendingWrites`, which makes the sync point waiting forever. Surely this is not an ideal unit test, but still better than a flaky one.

maysamyabandeh:
@Connor1996 You are right but this case seems more complicated. Even without unordered_write compact range might miss pending writes to the memtable even if they overlap with the range. Perhaps we can leave this case for future work.

maysamyabandeh:
@Connor1996 Did you get a chance to look into @yiwu-arbug's suggestion?

db/external_sst_file_test.cc
Connor1996: oh, I misunderstood your meaning of `beginning`, seems reasonable.

Connor1996:
@maysamyabandeh changed that, PTAL again

PR #5950, opened 2019-10-21
Include db_stress_tool in rocksdb tools lib

riversand963:
Summary: include db_stress_tool in rocksdb tools lib

Test Plan (on devserver):
```
$make db_stress
$./db_stress
$make all && make check
```

PR #6117, opened 2019-12-03
Fix build failure for db_stress tool when building with CMake

simpkins:
Summary:
PR #5937 changed the db_stress tool to also require db_stress_tool.cc,
and updated the Makefile but not the CMakeLists.txt file.  This updates
the CMakeLists.txt file so that the CMake build succeeds again.

PR #5950 updated the Makefile build to package db_stress_tool.cc into
its own librocksdb_stress.a library.  I haven't done that here since
there didn't really seem to be much benefit: the Makefile-based build
does not install this library.

Test Plan:
Confirmed the CMake build succeeds on an Ubuntu 18.04 system.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -1,7 +1,6 @@
 set(TOOLS
   sst_dump.cc
   db_sanity_test.cc
-  db_stress.cc
   write_stress.cc
   ldb.cc
   db_repl_stress.cc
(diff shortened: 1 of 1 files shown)


simpkins:
Note that the travis-ci build failure for the cmake-mingw build looks unrelated to the changes in PR.  It appears the same build failure exists on master.  e.g., master build with the same failure: https://travis-ci.org/facebook/rocksdb/builds/619863802

riversand963:
Thanks @simpkins for fixing Cmake

PR #6125, opened 2019-12-05
Use SpecialSkipListFactory in RecalculateScoreAfterPicking

### LINKED ISSUE #5736: DBTestUniversalCompaction.RecalculateScoreAfterPicking (db_universal_compaction_test) sometimes fails on ARM
We are now testing Rocksdb on ARM. When running `make check_some` on master branch. There a test failed:

db_universal_compaction_test

db/db_universal_compaction_test.cc:1781: Failure
Value of: 5
Expected: NumSortedRuns()
Which is: 6
terminate called after throwing an instance of 'testing::internal::GoogleTestFailureException'
  what():  db/db_universal_compaction_test.cc:1781: Failure
Value of: 5
Expected: NumSortedRuns()
Which is: 6

I'm not sure how to fix them. Could Rocksdb team take a look at this issue? Thanks.

pdillinger:
Summary: Test DBTestUniversalCompaction.RecalculateScoreAfterPicking was
flaky on ARM, so it now uses SpecialSkipListFactory (like other tests)
for predictable memtable flushes.

Fixes #5736

Test Plan: while ./db_universal_compaction_test; do :; done # for a
while on ARM and on Intel (both Linux)

### DIFF
--- a/db/db_universal_compaction_test.cc
+++ b/db/db_universal_compaction_test.cc
@@ -1753,13 +1753,14 @@ TEST_P(DBTestUniversalCompaction, RecalculateScoreAfterPicking) {
   // scheduling more; otherwise, other CFs/DBs may be delayed unnecessarily.
   const int kNumFilesTrigger = 8;
   Options options = CurrentOptions();
+  options.memtable_factory.reset(
+      new SpecialSkipListFactory(KNumKeysByGenerateNewFile - 1));
   options.compaction_options_universal.max_merge_width = kNumFilesTrigger / 2;
   options.compaction_options_universal.max_size_amplification_percent =
       static_cast<unsigned int>(-1);
   options.compaction_style = kCompactionStyleUniversal;
   options.level0_file_num_compaction_trigger = kNumFilesTrigger;
   options.num_levels = num_levels_;
-  options.write_buffer_size = 100 << 10;  // 100KB
   Reopen(options);
 
   std::atomic<int> num_compactions_attempted(0);


db/db_universal_compaction_test.cc
```diff
@@ -1759,7 +1759,8 @@ TEST_P(DBTestUniversalCompaction, RecalculateScoreAfterPicking) {
   options.compaction_style = kCompactionStyleUniversal;
   options.level0_file_num_compaction_trigger = kNumFilesTrigger;
   options.num_levels = num_levels_;
-  options.write_buffer_size = 100 << 10;  // 100KB
+  // 110 KB = big enough to require manual flush after GenerateNewFile
+  options.write_buffer_size = 110 << 10;
```
ajkr: It feels the root problem before this change was inconsistency -- the `write_buffer_size` here is trying to match the amount of data written by `GenerateNewTable()`. Some other test cases eliminated the duplication thus eliminating risk of inconsistency by using `SpecialSkipListFactory` configured with `KNumKeysByGenerateNewFile`. That memtable factory is different in that the memtables it generates flush after the configured number of keys are added (plus one) rather than after a certain data size is exceeded. The `KNumKeysByGenerateNewFile` is also used by `GenerateNewTable()` to decide how many keys to add, so using that constant eliminates the logic duplication between flushing decision and `GenerateNewTable()`. For example https://github.com/facebook/rocksdb/blob/master/db/db_universal_compaction_test.cc#L1262-L1263

pdillinger: Seems like a good suggestion, though your description seems to focus on fragility under code changes and compilation environment changes ("duplication"). Those are "nice to have" compared to fragility under runtime conditions / random chance, which is what I observed. My observations suggest there is no `write_buffer_size` that is reliably consistent with `GenerateNewTable()` from run to run with the same binary on the same machine. I'm only assuming that's not a bug.

ajkr: Got it, that is good to know. You are right that I wasn't thinking about non-determinism across runs. One possibility is multiple calls to `GenerateNewFile()` can happen between background flushes. That's usually solved by calling `TEST_WaitForFlushMemTable()` between calls to `GenerateNewFile()`. I'm not sure about other causes of non-determinism here, but wouldn't be surprised if there are more.

PR #6128, opened 2019-12-06
Disable new Bloom filter assertion

pdillinger:
Summary: A longstanding bug in our C interface can trigger this
assertion; see issue #6129. Disabling the assertion for now
(for 6.6.0) and will re-enable on fix of that bug.

### DIFF
--- a/table/block_based/filter_policy.cc
+++ b/table/block_based/filter_policy.cc
@@ -443,7 +443,8 @@ void BloomFilterPolicy::CreateFilter(const Slice* keys, int n,
                                      std::string* dst) const {
   // We should ideally only be using this deprecated interface for
   // appropriately constructed BloomFilterPolicy
-  assert(mode_ == kDeprecatedBlock);
+  // FIXME disabled because of bug in C interface; see issue #6129
+  //assert(mode_ == kDeprecatedBlock);
 
   // Compute bloom filter size (in both bits and bytes)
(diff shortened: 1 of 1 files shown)


table/block_based/filter_policy.cc
```diff
@@ -443,7 +443,8 @@ void BloomFilterPolicy::CreateFilter(const Slice* keys, int n,
                                      std::string* dst) const {
   // We should ideally only be using this deprecated interface for
   // appropriately constructed BloomFilterPolicy
-  assert(mode_ == kDeprecatedBlock);
+  // FIXME disabled because of bug in C interface; see issue #6129
+  //assert(mode_ == kDeprecatedBlock);
```
riversand963: Does it make sense to use `assert(mode_ == kDeprecatedBlock || mode_ == kAuto)`?

pdillinger: No, you have to explicitly opt in to using the block-based filter format, which we call "deprecated" in some places

pdillinger: In fact, we would not have caught the bug if the assertion were so permissive.

riversand963: I think the long term solution is to fix the C-API related code. At the moment as a temporary fix, adding an `or` seems stricter than removing the whole assertion.

PR #6139, opened 2019-12-09
Use Env::GetChildren() instead of readdir

riversand963:
Summary:
For more portability, switch from readdir to Env::GetChildren() in ldb's
manifest_dump subcommand.

Test Plan:
```
$make check
```
Manually check ldb command.


riversand963:
Looks like std::regex is not recommended. Will update the impl.

PR #6155, opened 2019-12-11
Enable txn in crash tests

maysamyabandeh:
Start daily crash tests with use_txn flag.
Test:
```
make crash_test_with_txn
```


Makefile
```diff
@@ -1006,6 +1021,11 @@ ubsan_crash_test_with_atomic_flush:
 	COMPILE_WITH_UBSAN=1 $(MAKE) crash_test_with_atomic_flush
 	$(MAKE) clean
 
+ubsan_crash_test_with_txn:
```
riversand963: Looks like tsan_crash_test_with_txn is missing?

maysamyabandeh: It is missing for atomic flush as well. Apparently it sets the tsan flag directly in the lego-determinator instead of relying it to be set from within Makefile: https://github.com/facebook/rocksdb/blob/master/build_tools/rocksdb-lego-determinator#L695

PR #6158, opened 2019-12-12
Enable all txn write policies in crash test

maysamyabandeh:
Currently the default txn write policy in crash tests is WRITE_PREPARED. The patch randomly picks the write policy at the start of the crash test.
Test:
```
make -j32 crash_test_with_txn
```

### DIFF
--- a/db_stress_tool/db_stress_common.h
+++ b/db_stress_tool/db_stress_common.h
@@ -161,7 +161,7 @@ DECLARE_int32(range_deletion_width);
 DECLARE_uint64(rate_limiter_bytes_per_sec);
 DECLARE_bool(rate_limit_bg_reads);
 DECLARE_bool(use_txn);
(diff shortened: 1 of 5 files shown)


db_stress_tool/db_stress_common.h
```diff
@@ -161,6 +161,7 @@ DECLARE_int32(range_deletion_width);
 DECLARE_uint64(rate_limiter_bytes_per_sec);
 DECLARE_bool(rate_limit_bg_reads);
 DECLARE_bool(use_txn);
+DECLARE_uint64(txn_write_policy);
```
riversand963: Is it easier to understand if we use string here, and use a mapping from string to policy?

maysamyabandeh: Makes sense. Let me leave it as a TODO.

PR #6165, opened 2019-12-13
Generate variable length keys in db_stress

anand1976:
Currently, db_stress generates fixed length keys of 8 bytes. This patch adds the ability to generate variable length keys. Most of the db_stress code continues to work with a numeric key randomly generated, and the numeric key also acts as an index into the values_ array. The numeric key is mapped to a variable length string key in a deterministic way. Furthermore, the ordering is preserved.

Test plan:
run make crash_test


db_stress_tool/db_stress_common.h
```diff
@@ -327,16 +330,97 @@ extern inline std::string Key(int64_t val) {
...
+  // big endian format
+  unsigned int levels;
+  // Number of keys of each possible length in a given window
+  std::vector<uint64_t> weights;
+  // Size of window for varying the key length
+  uint64_t window;
+};
+extern KeyGenContext key_gen_ctx;
+
+// Generate a variable length key string from the given int64 val. The
```
pdillinger: I think you should mention that order is preserved, though I find it odd that it takes int64_t and we coerce to uint64_t

db_stress_tool/db_stress_common.h
```diff
@@ -327,16 +330,97 @@ extern inline std::string Key(int64_t val) {
   return big_endian_key;
 }
 
+// A struct for maintaining the parameters for generating variable length keys
+struct KeyGenContext {
+  // Max length of the key in 8-byte units. Each level is a int64 encoded in
+  // big endian format
+  unsigned int levels;
```
pdillinger: Why not use weights.size()?

db_stress_tool/db_stress_common.h
```diff
@@ -327,16 +330,97 @@ extern inline std::string Key(int64_t val) {
...
+  // Number of keys of each possible length in a given window
+  std::vector<uint64_t> weights;
+  // Size of window for varying the key length
+  uint64_t window;
```
pdillinger: It took me a long time to figure out what window meant here. It's really the number of adjacent keys in a cycle of key lengths.

db_stress_tool/db_stress_gflags.cc
```diff
@@ -28,6 +28,18 @@ DEFINE_bool(read_only, false, "True if open DB in read-only mode during tests");
 DEFINE_int64(max_key, 1 * KB * KB,
              "Max number of key/values to place in database");
 
+DEFINE_int32(max_key_len, 3,
```
pdillinger: I don't see the point of this option if it has to agree with the number of values in key_len_percent_dist.

It also seems easy and valuable to use key lengths that are not multiples of 8, such as appending 0-7 'x's depending on the bottom bits of offset.

db_stress_tool/db_stress_common.h
```diff
@@ -327,16 +330,97 @@ extern inline std::string Key(int64_t val) {
...
+  *key_p = key;
+  return true;
+}
+
+extern inline uint64_t GetPrefixKeyCount(const std::string prefix,
+                                         const std::string ub) {
+  uint64_t start = 0;
+  uint64_t end = 0;
+
+  if (!GetIntVal(prefix, &start) || !GetIntVal(ub, &start)) {
```
pdillinger: &end?

db_stress_tool/db_stress_common.h
anand1976: Yes, it is a bit strange. The call sites expect int64_t as max_key is int64_t. The inverse, ```GetIntVal()``` takes a uint64_t*. I think it should be uint64_t throughout.

db_stress_tool/db_stress_gflags.cc
anand1976: I meant to make the key_len_percent_dist optional, but just realized I haven't taken care of initializing it if not specified.

The idea of key lengths that are not multiples of 8 byb appending 0-7 'x's is good. I'll implement it.

Issue #6183, opened 2019-12-15
Android Java: keyMayExist crashes when reading non UTF8 valid character


jurmous:
Using a keyMayExist on a key with a value with invalid UTF8 characters it crashes on Android.

In this case I stored a value which was an encoded number which was not meant to be represented as String.

### Expected behavior

No crash using keyMayExist with a non UTF8 valid value

### Actual behavior

On Android:
```
2019-12-14 19:15:49.948 4756-4756/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
A/DEBUG: Build fingerprint: 'google/sdk_gphone_x86/generic_x86:10/QSR1.190920.001/5891938:user/release-keys'
A/DEBUG: Revision: '0'
A/DEBUG: ABI: 'x86'
A/DEBUG: Timestamp: 2019-12-14 19:15:49+0100
A/DEBUG: pid: 4693, tid: 4730, name: roidJUnitRunner  >>> org.rocksdb.test <<<
A/DEBUG: uid: 10134
A/DEBUG: signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
A/DEBUG: Abort message: 'JNI DETECTED ERROR IN APPLICATION: input is not valid Modified UTF-8: illegal start byte 0x80
        string: '�'
        input: '<0x80>'
        in call to NewStringUTF
        from boolean org.rocksdb.RocksDB.keyMayExist(long, byte[], int, int, java.lang.StringBuilder)'
```

(On the JVM it works fine)

### Steps to reproduce the behavior

put a value of 0x80 in the database. Use keyMayExist to check if key exists. Then get the exception.

Use this project to compile for Android: 
https://github.com/marykdb/rocksdb-android

The git submodule can be changed to refer to different RocksDB commits.

adamretter:
Hmm it would seem to me that `StringBuilder` is a poor choice to use for the value parameter in `keyMayExist`, as we would have to expect a unicode value. I will work on an alternative

adamretter:
@jurmous I think this should solve the problem - https://github.com/facebook/rocksdb/pull/6186

In particular see `KeyMayExistTest#keyMayExistNonUnicodeString()`.

PR #6184, opened 2019-12-15
Java: Enable keyMayExist to use a null StringBuilder

jurmous:
Currently StringBuilder is mandatory in the keyMayExist method. This way a StringBuilder needs to be created even when not needed and causes a crash on Android platform is the value is a non UTF8 valid String. 

This PR enables to pass a null StringBuilder to keyMayExist.

The API is not changed but it introduces another check before appending to the StringBuilder.

This MR introduces a way to circumvent a crash on Android: #6183

### DIFF
--- a/java/rocksjni/rocksjni.cc
+++ b/java/rocksjni/rocksjni.cc
@@ -1618,7 +1618,7 @@ jboolean key_may_exist_helper(JNIEnv* env, rocksdb::DB* db,
   delete[] key;
 
   // extract the value
-  if (value_found && !value.empty()) {
+  if (value_found && !value.empty() && jstring_builder != nullptr) {
     jobject jresult_string_builder =
         rocksdb::StringBuilderJni::append(env, jstring_builder, value.c_str());
     if (jresult_string_builder == nullptr) {


java/rocksjni/rocksjni.cc
```diff
@@ -1618,7 +1618,7 @@ jboolean key_may_exist_helper(JNIEnv* env, rocksdb::DB* db,
   delete[] key;
 
   // extract the value
-  if (value_found && !value.empty()) {
+  if (value_found && !value.empty() && jstring_builder != nullptr) {
```
adamretter: I am not sure that this is a good idea, as it deviates from the C++ API which we try to stay close to. In the C++ API passing `null` as the `value` (in this case the equivalent of `jstring_builder`) would cause an invalid pointer de-reference and presumably result in a SIGSEGV.

The C++ function implementation , simply contains `assert(value != nullptr);`. Perhaps instead of this we should just add an `assert value != null;` in the Java `keyMayExist` functions?

jurmous:
> Not sure about this, as it changes the expected behaviour where the value is always returned _only if_ it is in the block cache. With this change, users could mistakenly pass a null-reference and then think the value is not in the block cache. I think it might be better to add an assert and stay with the underlying C++ API behaviour.

The Java API uses a StringBuilder to "catch" the value if set. So the user needs to read the StringBuilder it passed to see if the value was set in block cache. So I do not understand how they can think the wrong thing because it does not work with reference reading.

> That said, if it was argued that we really do want to do this, then we should at least add Unit tests to this PR.

Those can be added. Let's first see if we get agreement on the implementation.

adamretter:
@jurmous All being well, if you agree, I think https://github.com/facebook/rocksdb/pull/6186 will supersede this

pdillinger:
Going with #6186

jurmous:
Thanks for the work @adamretter!

PR #6186, opened 2019-12-15
RocksDB#keyMayExist should not assume database values are unicode strings

adamretter:
Closes https://github.com/facebook/rocksdb/issues/6183


adamretter:
@pdillinger Thanks for the feedback. I just pushed some changes inline with your suggestions

PR #6190, opened 2019-12-16
Make Env into a std::shared_ptr<Env>.

mrambacher:
This change allows the LoadEnv to work properly without leaking Env structures.  As previously implemented, a LoadEnv called for loading from the DBOptions might leak an Env structure if the Env was not a static object.


adamretter:
@siying we would need to update the RocksJava code to work with this too.

PR #6192, opened 2019-12-16
Fix crash in Transaction::MultiGet() when num_keys > 32

anand1976:
When ```Transaction::MultiGet``` is called with > 32 keys, the internally allocated ```autovector<KeyContext>``` for tracking the key lookups would end up using a dynamically allocated ```std::vector``` in addition to automatic storage. The vector might get reallocated as it grows, causing pointers stored to its elements in ```sorted_keys``` to become invalid. To fix this, we must not store those pointers until we're sure the autovector has been sized correctly.

Test plan:
Add a unit test that fails without the fix and passes now
make check


utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -1004,10 +1004,13 @@ void WriteBatchWithIndex::MultiGetFromBatchAndDB(
     assert(result == WriteBatchWithIndexInternal::Result::kMergeInProgress ||
            result == WriteBatchWithIndexInternal::Result::kNotFound);
     key_context.emplace_back(column_family, keys[i], &values[i], &statuses[i]);
-    sorted_keys.emplace_back(&key_context.back());
     merges.emplace_back(result, std::move(merge_context));
   }
 
+  for (KeyContext& key : key_context) {
+    sorted_keys.emplace_back(&key);
```
lth: Alternatively, we could preallocate key_context if you wanted to do just one pass.

utilities/transactions/transaction_test.cc
```diff
@@ -2823,6 +2823,93 @@ TEST_P(TransactionTest, MultiGetBatchedTest) {
...
+  s = db->Write(write_options, &batch);
+  ASSERT_OK(s);
+
+  WriteBatchWithIndex wb;
+  // Write some data to the db
+  s = wb.Delete(handles[1], std::to_string(1));
+  ASSERT_OK(s);
+  s = wb.Put(handles[1], std::to_string(2), "new_val" + std::to_string(2));
+  ASSERT_OK(s);
+  for (int i = 8; i < 56; ++i) {
```
maysamyabandeh: Can you add some inline comment about how these numbers are picked?

utilities/transactions/transaction_test.cc
```diff
@@ -2823,6 +2823,93 @@ TEST_P(TransactionTest, MultiGetBatchedTest) {
...
+    s = wb.Merge(handles[1], std::to_string(i), "merge");
+    ASSERT_OK(s);
+  }
+
+  std::vector<Slice> keys;
+  for (int i = 0; i < 64; ++i) {
```
maysamyabandeh: ditto

lth:
Looks good to me.

utilities/transactions/transaction_test.cc
```diff
@@ -2823,6 +2823,93 @@ TEST_P(TransactionTest, MultiGetBatchedTest) {
   }
 }
 
+TEST_P(TransactionTest, MultiGetLargeBatchedTest) {
```
maysamyabandeh: It seems that the test is more complicated than simply using a large batch. Can you add a few inline comments of what the test does?

PR #6193, opened 2019-12-16
Fix use-after-free and double-deleting files in BackgroundCallPurge()

al13n321:
The bad code was:

```
mutex.Lock(); // `mutex` protects `container`
for (auto& x : container) {
  mutex.Unlock();
  // do stuff to x
  mutex.Lock();
}
```

It's incorrect because both `x` and the iterator may become invalid if another thread modifies the container while this thread is not holding the mutex.

Broken by https://github.com/facebook/rocksdb/pull/5796 - it replaced a `while (!container.empty())` loop with a `for (auto x : container)`.

(RocksDB code does a lot of such unlocking+re-locking of mutexes, and this type of bugs comes up a lot :/ )

Test plan: Ran some logdevice integration tests that were crashing without this fix.

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -1321,19 +1321,26 @@ void DBImpl::BackgroundCallPurge() {
     delete sv;
     mutex_.Lock();
   }
-  for (const auto& file : purge_files_) {
-    const PurgeFileInfo& purge_file = file.second;
+
+  // Can't use iterator to go over purge_files_ because inside the loop we're
+  // unlocking the mutex that protects purge_files_.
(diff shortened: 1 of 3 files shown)


riversand963:
Thanks @al13n321  for fixing this. Will take a look.

riversand963:
@al13n321, do you plan to land this PR?

al13n321:
> do you plan to land this PR?

Yes, will land after checks complete.

al13n321:
The travis failure looks unrelated (some invalid HTML in some java file), and https://github.com/facebook/rocksdb/pull/6206 has the same error. Landing.

PR #6195, opened 2019-12-16
Refactor pulling out parts of StressTest::OperateDb

pdillinger:
Complete some refactoring called for in #6148. Somehow I got some 'make format' in here for files I didn't change, but that should be OK. (I'm not sure why "hide whitespace changes" doesn't seem to help in review.)

Not addressed in this PR: some operations simply print to stdout rather than failing on discovering a bad status or inconsistency.


pdillinger:
> Can you separate the formatting change that is not related to db_stress to a separate PR?

Done, except for separate PR

PR #6205, opened 2019-12-17
Update file indexer to take timestamp into consideration

hliu18:
Exclude timestamp in key comparison during boundary calculation to avoid key versions being excluded.


db/db_basic_test.cc
```diff
@@ -2015,22 +2015,54 @@ INSTANTIATE_TEST_CASE_P(
...
+    return result;
+  }
+  void ClearFlushedFiles() {
+    std::lock_guard<std::mutex> lock(mutex_);
+    flushed_files_.clear();
+  }
+
+ private:
+  std::vector<std::string> flushed_files_;
+  std::mutex mutex_;
```
riversand963: Nit: we usually use `InstrumentedMutex` instead of directly using `std::mutex`. Similarly, we use `InstrumentedMutexLock` instead of `std::lock_guard`.

db/db_basic_test.cc
```diff
@@ -2015,22 +2015,54 @@ INSTANTIATE_TEST_CASE_P(
...
+
+  virtual void OnFlushCompleted(DB* /*db*/, const FlushJobInfo& info) override {
+    std::lock_guard<std::mutex> lock(mutex_);
+    flushed_files_.push_back(info.file_path);
+  }
+
+  std::vector<std::string> GetFlushedFiles() {
+    std::lock_guard<std::mutex> lock(mutex_);
+    std::vector<std::string> result;
+    for (auto fname : flushed_files_) {
```
riversand963: Since this is just a test, it's not a big issue, but seems like you can just do
```
  std::vector<std::string> result;
  {
    InstrumentedMutexLock lock(mutex_);
    result = flushed_files_;
  }
  return result:
```

db/db_basic_test.cc
```diff
@@ -2015,22 +2015,54 @@ INSTANTIATE_TEST_CASE_P(
     ::testing::Combine(::testing::Bool(), ::testing::Bool(),
                        ::testing::Bool(), ::testing::Bool()));
 
+// A class which remembers the name of each flushed file.
+class FlushedFileCollector : public EventListener {
+ public:
+  FlushedFileCollector() {}
+  ~FlushedFileCollector() {}
+
+  virtual void OnFlushCompleted(DB* /*db*/, const FlushJobInfo& info) override {
```
riversand963: Remove `virtual` but keep `override`.

db/db_basic_test.cc
```diff
@@ -2015,22 +2015,54 @@ INSTANTIATE_TEST_CASE_P(
     ::testing::Combine(::testing::Bool(), ::testing::Bool(),
                        ::testing::Bool(), ::testing::Bool()));
 
+// A class which remembers the name of each flushed file.
+class FlushedFileCollector : public EventListener {
+ public:
+  FlushedFileCollector() {}
+  ~FlushedFileCollector() {}
```
riversand963: `~FlushedFileCollector() override {}`.

db/db_basic_test.cc
```diff
@@ -2015,22 +2015,54 @@ INSTANTIATE_TEST_CASE_P(
...
-      public testing::WithParamInterface<bool> {
+      public testing::WithParamInterface<int> {
  public:
   DBBasicTestWithTimestampWithParam()
       : DBTestBase("/db_basic_test_with_timestamp") {}
 
  protected:
   class TestComparator : public Comparator {
    private:
+    bool number_order_;
```
riversand963: Looks like you are trying to modify the comparison logic of the comparator. Will it be simpler if you just create another `Comparator` class?

db/db_basic_test.cc
```diff
@@ -2104,14 +2142,20 @@ class DBBasicTestWithTimestampWithParam
 TEST_P(DBBasicTestWithTimestampWithParam, PutAndGet) {
   const int kNumKeysPerFile = 8192;
   const size_t kNumTimestamps = 6;
-  bool memtable_only = GetParam();
+  const size_t kNumKeysPerTimestamp = (kNumKeysPerFile - 1) / kNumTimestamps;
+  const size_t kSplitPosBase = kNumKeysPerTimestamp / 2;
+  int flush_compaction = GetParam();
```
riversand963: The meaning of `flush_compaction` can be clearer?

db/db_basic_test.cc
hliu18: Will update. (The FlushedFileCollector class was copy-pasted from other tests.)

db/db_basic_test.cc
hliu18: Will update.

db/db_basic_test.cc
hliu18: Will update.

db/db_basic_test.cc
hliu18: Will refactor.

db/db_basic_test.cc
hliu18: Will refactor.

db/db_basic_test.cc
hliu18: Will update.

db/db_basic_test.cc
```diff
@@ -2101,6 +2097,47 @@ class DBBasicTestWithTimestampWithParam
...
+ public:
+  DBBasicTestWithTimestamp() {}
+
+ protected:
+  class TestComparator : public TestComparatorBase {
+   public:
+    explicit TestComparator(size_t ts_sz) : TestComparatorBase(ts_sz) {}
+
+    int CompareImpl(const Slice& a, const Slice& b) const override {
+      int n1 = atoi(a.data()+3);  // 3: length of "Key", to be skipped
```
riversand963: Is the value `3` specific to your test case? If so, can you use a public static const in TestComparator or somewhere else, to indicate the length of key?
`a` is a `Slice` object whose underlying `data()` does not have to be a `\0`-terminated, C-string.

hliu18: Made '3' a const. Good catch on the Slice data type. Fixed.
(I'm on vacation and response can be slow.)

db/db_basic_test.cc
```diff
@@ -2198,6 +2236,116 @@ TEST_P(DBBasicTestWithTimestampWithParam, PutAndGet) {
...
+          EncodeTimestamp(1 + i * 2, 0, &read_ts_strs[i]));
+    const Slice& write_ts = write_ts_list.back();
+    WriteOptions wopts;
+    wopts.timestamp = &write_ts;
+    for (int cf = 0; cf != static_cast<int>(num_cfs); ++cf) {
+      for (size_t j = 0; j != (kNumKeysPerFile - 1) / kNumTimestamps; ++j) {
+        ASSERT_OK(Put(
+          cf, "key" + std::to_string(j),
+          "value_" + std::to_string(j) + "_" + std::to_string(i), wopts));
+        if (j == kSplitPosBase + i || j == kNumKeysPerTimestamp - 1) {
```
riversand963: `i` can be 0 or 1 in this test. Any specific reason for `j == kSplitPosBase + i`?

db/db_basic_test.cc
```diff
@@ -2101,6 +2097,48 @@ class DBBasicTestWithTimestampWithParam
...
 
+class DBBasicTestWithTimestamp
+    : public DBBasicTestWithTimestampBase {
+ public:
+  DBBasicTestWithTimestamp() {}
+
+ protected:
+  class TestComparator : public TestComparatorBase {
+   public:
+    const int tKeyPrefixLength = 3; // 3: length of "key" in generated keys ("key" + std::to_string(j))
```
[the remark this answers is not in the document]
hliu18: My mistake. Changing 't' to 'k'.
Formatting to make lines shorter.

db/db_basic_test.cc
```diff
@@ -2213,4 +2361,4 @@ int main(int argc, char** argv) {
   ::testing::InitGoogleTest(&argc, argv);
   RegisterCustomObjects(argc, argv);
   return RUN_ALL_TESTS();
-}
+}
```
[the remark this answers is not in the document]
hliu18: removed.

db/db_basic_test.cc
```diff
@@ -2198,6 +2236,116 @@ TEST_P(DBBasicTestWithTimestampWithParam, PutAndGet) {
...
+          // such that lowerlevel[1].smallest.userkey == higherlevel[0].largest.userkey
+          ASSERT_OK(Flush(cf));
+  
+          // compact files (2 at each level) to a lower level such that all keys with the same
+          // timestamp is at one level, with newer versions at higher levels.
+          CompactionOptions compact_opt;
+          compact_opt.compression = kNoCompression;
+          db_->CompactFiles(compact_opt, handles_[cf],
+                            collector->GetFlushedFiles(),
+                              (int)(kNumTimestamps - i));
```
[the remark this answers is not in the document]
hliu18: Changed

db/db_basic_test.cc
```diff
@@ -2198,6 +2236,116 @@ TEST_P(DBBasicTestWithTimestampWithParam, PutAndGet) {
...
+  for (size_t i = 0; i != kNumTimestamps; ++i) {
+      write_ts_list.emplace_back(
+          EncodeTimestamp(i * 2, 0, &write_ts_strs[i]));
+      read_ts_list.emplace_back(
+          EncodeTimestamp(1 + i * 2, 0, &read_ts_strs[i]));
+    const Slice& write_ts = write_ts_list.back();
+    WriteOptions wopts;
+    wopts.timestamp = &write_ts;
+    for (int cf = 0; cf != static_cast<int>(num_cfs); ++cf) {
+      for (size_t j = 0; j != (kNumKeysPerFile - 1) / kNumTimestamps; ++j) {
```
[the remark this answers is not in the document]
hliu18: Good catch. Lost during refactor.

db/db_basic_test.cc
hliu18: It's explained in the comment in the lines below. "+ i" shifts the split position to the right (larger key). For example, if keys on level 2 are split as (1 2 3, 4 5 6), on level 1 they are split as (1 2 3 4, 5 6). key '4' is in the second sst file on level 2, but in the first sst file on level 1. When compared with timestamp, '4'+ts' on level 1 is smaller than '4'+ts on level 2 (ts' > ts). When the code is timestamp aware, the second sst file on level 2 overlaps with the first sst file on level 1 (compare without timestamp); otherwise, there is no overlap.

db/db_basic_test.cc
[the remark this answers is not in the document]
riversand963: Not quite.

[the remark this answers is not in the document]
riversand963: Screenshot 2020-01-07 16 51 02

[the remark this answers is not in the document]
riversand963: nvm. I'll do final formatting.

db/db_basic_test.cc
```diff
@@ -2101,6 +2095,52 @@ class DBBasicTestWithTimestampWithParam
   }
 };
 
+class DBBasicTestWithTimestamp : public DBBasicTestWithTimestampBase {
```
riversand963: Maybe it's better to let `DBBasicTestWithTimestamp`, `DBBasicTestWithTimestampWithParam` have different db directory. To achieve this, the constructor of `DBBasicTestWithTimestampBase ` can take directory path as argument.
```
class DBBasicTestWithTimestampBase ... {
  DBBasicTestWithTimestampBase(const std::string& dbname);
};
..
class DBBasicTestWithTimestamp : public DBBasicTestWithTimestampBase {
  DBBasicTestWithTimestamp(): DBBasicTestWithTimestampBase("/db_basic_test_with_timestamp") {}
};
...
class DBBasicTestWithTimestampWithParam : 
{
  DBBasicTestWithTimestampWithParam(): DBBasicTestWithTimestampBase("/db_basic_test_with_timestamp") {}
};
```

hliu18: Pushed a new change to use different db directories.

riversand963:
@hliu18 could you please rebase? Our internal landing system complained that it may be  out-dated and aborted the merge.

hliu18:
Done.

hliu18:
Rebased.

hliu18:
Hold on. I might have overwritten the last format change.

hliu18:
@riversand963 , can you do "make format" again? I acted too fast without pulling your earlier push.

riversand963:
Appveyor error irrelevant.

PR #6209, opened 2019-12-18
Syntax check python files on testing

pdillinger:
Summary: Adds a python script to syntax check all python files in the
repository and report any errors.

Test Plan: 'make check' with and without seeded syntax errors. Also look
for "No syntax errors in 34 .py files" on success, and in java_test CI output

### DIFF
--- a/tools/check_all_python.py
+++ b/tools/check_all_python.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python2
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.
+import glob
+
+# Checks that all python files in the repository are at least free of syntax
+# errors. This provides a minimal pre-/post-commit check for python file
+# modifications.
+
+count = 0
+# Clean this up when we finally upgrade to Python 3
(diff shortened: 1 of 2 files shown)


pdillinger:
> Also look
> for "No syntax errors in 34 .py files" on success, and in java_test CI output

Found in Travis java_test build

tools/check_all_python.py
```diff
@@ -0,0 +1,17 @@
...
+# Checks that all python files in the repository are at least free of syntax
+# errors. This provides a minimal pre-/post-commit check for python file
+# modifications.
+
+count = 0
+# Clean this up when we finally upgrade to Python 3
+for filename in glob.glob("*.py") + glob.glob("*/*.py") + glob.glob("*/*/*.py") + glob.glob("*/*/*/*.py"):
+    source = open(filename, 'r').read() + '\n'
+    # Parses and syntax checks the file, throwing on error. (No pyc written.)
+    _ = compile(source, filename, 'exec')
```
anand1976: I assume make would abort if there's an error in this step?

pdillinger: Yes, compile() will throw an exception, python will return non-zero, and make will fail

PR #6210, opened 2019-12-18
Add kHashSearch to stress tests

maysamyabandeh:
Beside extending index_type to kHashSearch, it clarifies in the code base that this feature is incompatible with index_block_restart_interval > 1.
Test: 
```
make -j32 crash_test

### DIFF
--- a/table/block_based/block_based_table_factory.cc
+++ b/table/block_based/block_based_table_factory.cc
@@ -157,6 +157,8 @@ size_t TailPrefetchStats::GetSuggestedPrefetchSize() {
   return std::min(kMaxPrefetchSize, max_qualified_size);
 }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: table/block_based/block_based_table_factory.cc [hunk 1]
```diff
@@ -184,6 +186,11 @@ BlockBasedTableFactory::BlockBasedTableFactory(
   if (table_options_.index_block_restart_interval < 1) {
     table_options_.index_block_restart_interval = 1;
   }
+  if (table_options_.index_type == BlockBasedTableOptions::kHashSearch &&
+      table_options_.index_block_restart_interval == 1) {
```
ltamasi: This should be `!=`, right?

PR #6216, opened 2019-12-19
Checksum for each SST file and stores in MANIFEST

zhichao-cao:
In the current code base, RocksDB generate the checksum for each block and verify the checksum at usage. Current PR enable SST file checksum. After a SST file is generated by Flush or Compaction, RocksDB generate the SST file checksum and store the checksum value and checksum method name in the vs_info and MANIFEST as part for the FileMetadata.

Added the enable_sst_file_checksum to Options to enable or disable file checksum. Added sst_file_checksum to Options such that user can plugin their own SST file checksum calculate method via overriding the SstFileChecksum class. The checksum information inlcuding uint32_t checksum value and a checksum name (string).  A new tool is added to LDB such that user can dump out a list of file checksum information from MANIFEST. If user enables the file checksum but does not provide the sst_file_checksum instance, RocksDB will use the default crc32checksum implemented in table/sst_file_checksum_crc32c.h

Test plan: Added the testing case in table_test and ldb_cmd_test to verify checksum is correct in different level. Pass make asan_check.


tools/ldb_cmd_impl.h
```diff
@@ -171,6 +171,25 @@ class ManifestDumpCommand : public LDBCommand {
...
+class FileChecksumDumpCommand : public LDBCommand {
+ public:
+  static std::string Name() { return "file_checksum_dump"; }
+
+  FileChecksumDumpCommand(const std::vector<std::string>& params,
+                          const std::map<std::string, std::string>& options,
+                          const std::vector<std::string>& flags);
+
+  static void Help(std::string& ret);
+  virtual void DoCommand() override;
```
riversand963: Remove `virtual` but keep `override`.

db/builder.cc
```diff
@@ -131,9 +131,17 @@ Status BuildTable(
       file->SetIOPriority(io_priority);
       file->SetWriteLifeTimeHint(write_hint);
 
-      file_writer.reset(
-          new WritableFileWriter(std::move(file), fname, file_options, env,
-                                 ioptions.statistics, ioptions.listeners));
+      if (ioptions.enable_sst_file_checksum &&
```
riversand963: What if `enable_sst_file_checksum && ioptions.sst_file_checksum == nullptr`? Looks like the default crc32 impl should be used.

db/version_set.cc
```diff
@@ -4694,6 +4695,84 @@ Status VersionSet::ReduceNumberOfLevels(const std::string& dbname,
       mutable_cf_options, &ve, &dummy_mutex, nullptr, true);
 }
 
+// Get the checksum information including the checksum and checksum method
+// name of all SST files of this Manifest. Store the information in
+// FileChecksumLis which contains a map from file id to its checksum info
+Status VersionSet::GetAllFileCheckSumInfo(Options& options,
```
riversand963: Do you need to update `VersionSet::GetLiveFilesMetaData()`, too?

include/rocksdb/metadata.h
```diff
@@ -117,6 +120,16 @@ struct SstFileMetaData {
   // Timestamp when the SST file is created, provided by Env::GetCurrentTime().
   // 0 if the information is not available.
   uint64_t file_creation_time;
+
+  // The checksum of a SST file, the value is decided by the file content and
+  // the checksum algorithm used for this SST file. The checksum algorithm is
+  // identified by the file_checksum_name. If no method is provided or file
+  // checksum is disabled, it is 0.
+  uint32_t file_checksum;
```
riversand963: Any reason these two fields are not added to `struct FileMetaData`?

file/writable_file_writer.h
```diff
@@ -141,6 +149,14 @@ class WritableFileWriter {
 
   bool TEST_BufferIsEmpty() { return buf_.CurrentSize() == 0; }
 
+  void SetFileChecksumMethod(SstFileChecksum* checksum_cal) {
+    checksum_cal_ = checksum_cal;
+  }
+
+  uint32_t GetFileChecksum() { return file_checksum_; }
```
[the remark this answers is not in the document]
zhichao-cao: @riversand963 Thanks! Corrected the const. In this case, I prefer to use GetFileChecksum which alights with GetFileSize and GetFileChecksumName.

db/db_impl/db_impl_open.cc
```diff
@@ -136,6 +137,10 @@ DBOptions SanitizeOptions(const std::string& dbname, const DBOptions& src) {
     result.avoid_flush_during_recovery = false;
   }
 
+  if (result.enable_sst_file_checksum && !result.sst_file_checksum) {
```
zhichao-cao: @riversand963 Answer the previous question, if enable_sst_file_checksum but sst_file_checksum is null. When open the db, it will check the flag and the pointer. If flag is set but checksum method is not specified, the default one will be used. In table_builder and compaction_job, we still need to check both in case something like sst_file_checksum is freed unexpected.

db/version_set.cc
zhichao-cao: @riversand963 Yes, thanks.

db/db_impl/db_impl_open.cc
riversand963: Is this a bug?
> something like sst_file_checksum is freed unexpected.

zhichao-cao: @riversand963 There are two cases: 1) user does not provide the sst_file_checksum method, just enables the enable_sst_file_checksum; 2) user provide initiate the pointer but sst_file_checksum is freed unexpected before this check. For case 1, it is not a bug. For case 2, it can be a bug or crash happens before this check. But db cannot decide what case should be. So we follow the flag and reset to the default Crc32. Do you mean we should log a potential error message here?

include/rocksdb/sst_file_checksum.h
```diff
@@ -0,0 +1,69 @@
...
+#include <cassert>
+#include <map>
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace rocksdb {
+
+class Slice;
```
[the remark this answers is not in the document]
zhichao-cao: @riversand963 removed

include/rocksdb/sst_file_checksum.h
```diff
@@ -0,0 +1,66 @@
...
+  // Returns a name that identifies the current file checksum method.
+  virtual const char* Name() const = 0;
+};
+
+// The data structure that stores the mapping from file number to its checksum
+// It is used to store the file checksum from Manifest.
+struct FileChecksumList {
+  std::unordered_map<uint64_t, std::pair<uint32_t, std::string>> checksum_map;
+
+  void AddChecksumUnit(uint64_t f_id, uint32_t checksum,
```
riversand963: Let's use `file_number` instead of `f_id` for consistency across the codebase.

options/db_options.cc
```diff
@@ -242,6 +243,8 @@ void ImmutableDBOptions::Dump(Logger* log) const {
   ROCKS_LOG_HEADER(
       log, "                Options.log_readahead_size: %" ROCKSDB_PRIszt,
       log_readahead_size);
+  ROCKS_LOG_HEADER(log, "                Options.sst_file_checksum: %s",
+                   sst_file_checksum ? sst_file_checksum->Name() : "None");
```
riversand963: Replace hard-coded "None" with the constant string defined in version_edit.cc/h

include/rocksdb/sst_file_checksum.h
```diff
@@ -0,0 +1,66 @@
...
+  // Return a processed value of the checksum for store in somewhere
+  virtual uint32_t ProcessChecksum(const uint32_t checksum) = 0;
+
+  // Returns a name that identifies the current file checksum method.
+  virtual const char* Name() const = 0;
+};
+
+// The data structure that stores the mapping from file number to its checksum
+// It is used to store the file checksum from Manifest.
+struct FileChecksumList {
```
riversand963: Maybe do not expose the implementation of `FileChecksumList` in the public header.
1. They are subject to change.
2. They are implementation detail. It is possible that different data structures are preferred as the underlying storage for file checksums. A vector, a set, unordereset, etc.

table/table_test.cc
```diff
@@ -1171,6 +1172,123 @@ class PlainTableTest : public TableTest {};
...
+  Status CalculateFileChecksum(SstFileChecksum* file_checksum_cal,
+                               uint32_t* checksum) {
+    assert(file_checksum_cal != nullptr);
+    cur_uniq_id_ = checksum_uniq_id_++;
+    test::StringSink* ss_rw =
+        rocksdb::test::GetStringSinkFromLegacyWriter(file_writer_.get());
+    file_reader_.reset(test::GetRandomAccessFileReader(
+        new test::StringSource(ss_rw->contents())));
+    char* scratch = nullptr;
+    scratch = new char[2048];
```
riversand963: Nit: you can use `std::unique_ptr<char[]>` to avoid explicitly deleting.

include/rocksdb/sst_file_checksum.h
riversand963: Do we need to expose this class in public header in this PR? There is only one member function of `VersionSet` that uses this.

tools/ldb_cmd.cc
```diff
@@ -1139,6 +1144,141 @@ void ManifestDumpCommand::DoCommand() {
...
+  //  sst file numer, checksum method name, checksum value
+  //  sst file numer, checksum method name, checksum value
+  //  ......
+
+  std::unique_ptr<FileChecksumList> checksum_list(new FileChecksumList);
+  GetAllFileCheckSumInfoFromManifest(options_, manifestfile,
+                                     checksum_list.get());
+  if (checksum_list != nullptr) {
+    for (auto it : checksum_list->checksum_map) {
+      printf("%" PRId64 ", %s, %u\n", it.first, it.second.second.c_str(),
```
riversand963: use `fprintf)`

tools/ldb_cmd.cc
```diff
@@ -1139,6 +1144,141 @@ void ManifestDumpCommand::DoCommand() {
   }
 }
 
+// ----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
zhichao-cao: @riversand963 This is used in ldb_cmd.cc file only to separate the implementations of different CMDs in ldb.

include/rocksdb/sst_file_checksum.h
zhichao-cao: @riversand963 Looks like it would be better to directly return an unordered_map from the function in VersionSet to avoid the introducing new class to user.

db/builder.cc
```diff
@@ -131,9 +131,16 @@ Status BuildTable(
...
+      if (ioptions.sst_file_checksum != nullptr) {
+        // Calculate the file checksum when do Append
+        file_writer.reset(new WritableFileWriter(
+            std::move(file), fname, file_options, env, ioptions.statistics,
+            ioptions.listeners, ioptions.sst_file_checksum));
+      } else {
+        file_writer.reset(
+            new WritableFileWriter(std::move(file), fname, file_options, env,
+                                   ioptions.statistics, ioptions.listeners));
+      }
```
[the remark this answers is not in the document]
zhichao-cao: @riversand963 I will unify the the variable names as well as the constant.

include/rocksdb/sst_file_checksum.h
riversand963: Well, I think introducing a new class is fine, as long as we do not impose a certain data structure or implementation on the application. The public header should be the "interface", not implementation. Keep in mind that, it is much more difficult to remove something from public header once added.
Additionally, you can provide a default implementation of `FileChecksumList` whose implementation is also hidden. Applications can call `NewDefaultChecksumList()` to create it. The application can pass this to the upcoming API `DB::GetFileChecksums()`, which you will probably add in the next PR.
If the application is OK with default implementation of `FileChecksumList`, they can use it conveniently. Otherwise, they can create their own FileChecksumList by subclassing the base class defined in the public header.

file/writable_file_writer.h
```diff
@@ -68,13 +71,17 @@ class WritableFileWriter {
   RateLimiter* rate_limiter_;
   Statistics* stats_;
   std::vector<std::shared_ptr<EventListener>> listeners_;
+  SstFileChecksum* checksum_method_;
```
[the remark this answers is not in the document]
zhichao-cao: @riversand963 If it is const SstFileChecksum*, we cannot call its member function checksum->Value, and checksum->extend which are not const functions. Also, there is a function to set the file checksum function, we cannot make it SstFileChecksum* const.

include/rocksdb/sst_file_checksum.h
zhichao-cao: @riversand963 I will change accordingly. Thanks!

table/table_test.cc
zhichao-cao: @riversand963 changed accordingly.

include/rocksdb/sst_file_checksum.h
```diff
@@ -0,0 +1,82 @@
...
+#include <vector>
+
+#include "rocksdb/status.h"
+
+namespace rocksdb {
+
+// SstFileChecksumFunc is the function class to generates the checksum value
+// for each SST file when the file is written to the file system. The checksum
+// is stored in VersionSet as well as MANIFEST.
+class SstFileChecksumFunc {
```
riversand963: This is kind of nitpick: does this have to be specific to sst? Asking because file is named 'sst_file_checksum.h', but I think the implementation can work for any WritableFileWriter, correct?

include/rocksdb/sst_file_checksum.h
```diff
@@ -0,0 +1,82 @@
...
+// store the checksum information of a list of SST files to be ingested.
+class FileChecksumList {
+ public:
+  virtual ~FileChecksumList() {}
+
+  // Clean the previously stored file checksum information.
+  virtual void reset() = 0;
+
+  // Get the number of checksums in the checksum list
+  virtual size_t size() = 0;
```
riversand963: `virtual size_t size() const`?

include/rocksdb/sst_file_checksum.h
```diff
@@ -0,0 +1,82 @@
...
+
+  // Given the file_number, it searches if the file checksum information is
+  // stored.
+  virtual Status SearchOneFileChecksum(const uint64_t file_number,
+                                       uint32_t* checksum,
+                                       std::string* checksum_func_name) = 0;
+
+  // Insert the checksum information of one file to the FileChecksumList.
+  virtual Status InsertOneFileChecksum(
+      const uint64_t file_number, const uint32_t checksum,
```
riversand963: No need to use const if you pass POD types by value. This applies to other methods as well.

table/table_test.cc
```diff
@@ -1171,6 +1172,119 @@ class PlainTableTest : public TableTest {};
...
+
+  void SetFileChecksumFunc(SstFileChecksumFunc* checksum_func) {
+    if (file_writer_ != nullptr) {
+      file_writer_->TEST_SetFileChecksumFunc(checksum_func);
+    }
+  }
+
+  WritableFileWriter* GetFileWriter() { return file_writer_.get(); }
+
+  Status ResetTableBuilder(std::unique_ptr<TableBuilder> builder) {
```
riversand963: Maybe consider passing `builder` as rvalue reference?

include/rocksdb/sst_file_checksum.h
```diff
@@ -0,0 +1,82 @@
...
+      const std::string checksum_func_name) = 0;
+
+  // Remove the checksum information of one SST file.
+  virtual Status RemoveOneFileChecksum(const uint64_t file_number) = 0;
+};
+
+// Create a new file checksum list.
+extern FileChecksumList* NewFileChecksumList();
```
riversand963: What will be the API to create a default `SstFileChecksumFunc` object, e.g. `SstFileChecksumFunccrc32`? Exposing class `SstFileChecksumCrc32c` may not be a good idea.

tools/ldb_cmd.cc
```diff
@@ -1139,6 +1144,151 @@ void ManifestDumpCommand::DoCommand() {
...
+                                        options.table_cache_numshardbits));
+  // Notice we are using the default options not through SanitizeOptions(),
+  // if VersionSet::GetAllFileCheckSumInfo depends on any option done by
+  // SanitizeOptions(), we need to initialize it manually.
+  options.db_paths.emplace_back("dummy", 0);
+  options.num_levels = 64;
+  WriteController wc(options.delayed_write_rate);
+  WriteBufferManager wb(options.db_write_buffer_size);
+  ImmutableDBOptions immutable_db_options(options);
+  VersionSet versions(dbname, &immutable_db_options, sopt, tc.get(), &wb, &wc,
```
riversand963: I think you need to call `versions.Recover()`.

include/rocksdb/sst_file_checksum.h
zhichao-cao: @riversand963 You mean it is not suggested that user directly use SstFileCheckSumCrc32c? Instead, we expose an interface like CreateDefaultSstFileChecksumFunc?

db/repair.cc
```diff
@@ -586,7 +586,8 @@ class Repairer {
             table->meta.largest, table->meta.fd.smallest_seqno,
             table->meta.fd.largest_seqno, table->meta.marked_for_compaction,
             table->meta.oldest_blob_file_number,
-            table->meta.oldest_ancester_time, table->meta.file_creation_time);
+            table->meta.oldest_ancester_time, table->meta.file_creation_time,
+            table->meta.file_checksum, table->meta.file_checksum_func_name);
```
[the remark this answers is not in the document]
zhichao-cao: @riversand963 They are not populated. In ScanTable, we are not able to know the two. So, when doing the repair, the file_checksum and file_checksum_func_name are the value initialized when create a new TableInfo.

include/rocksdb/sst_file_checksum.h
zhichao-cao: @riversand963 Yes, directly using FileChecksumFunc make scenes. Should I also change the file name as file_checksum.h?

tools/ldb_cmd.cc
zhichao-cao: @riversand963 In this case, the tool just dumps out the existing information from Manifest file. Why versions.Recover() is needed? In this case, we do not have information about the CF and db_id.

db/repair.cc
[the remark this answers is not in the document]
riversand963: Thanks for clarification

include/rocksdb/sst_file_checksum.h
riversand963: SG.

include/rocksdb/sst_file_checksum.h
riversand963: How about `CreateDefaultFileChecksumFunc()`, or `CreateFileChecksumFuncCrc32`? The implementation of SstFileChecksumFuncCrc32c should not be exposed.

tools/ldb_cmd.cc
```diff
@@ -1139,6 +1144,151 @@ void ManifestDumpCommand::DoCommand() {
...
+                      /*block_cache_tracer=*/nullptr);
+  Status s = versions.GetAllFileCheckSumInfo(options, file, checksum_list);
```
riversand963: Nit: Checksum?

tools/ldb_cmd.cc
riversand963: Oh, I see, `GetFllFileChecksumInfo` scans the MANIFEST, so no need to call Recover.

include/rocksdb/sst_file_checksum.h
zhichao-cao: @riversand963 Updated as CreateFileChecksumFuncCrc32c(), I removed Default.

db/version_set.cc
```diff
@@ -4708,6 +4709,92 @@ Status VersionSet::ReduceNumberOfLevels(const std::string& dbname,
       mutable_cf_options, &ve, &dummy_mutex, nullptr, true);
 }
 
+// Get the checksum information including the checksum and checksum function
+// name of all SST files in this Manifest. Store the information in
+// FileChecksumList which contains a map from file number to its checksum info.
+Status VersionSet::GetAllFileChecksumInfo(const Options& options,
```
riversand963: Gave it another thought. Do you intend to call `DB::GetAllFileChecksumsInfo()` on a live DB? If so, the file metadata info should have already be in memory, thus you don't have to read the MANIFEST.
If you also use this for ldb too, you can always call `Recover()` first.
My suggestion is, maybe you can remove the MANIFEST reading part from this method. Instead, can you populate the checksum info from column families' version storage info?

tools/ldb_cmd.cc
riversand963: Please see my most recent comment regarding `VersionSet::GetAllFileChecksumInfo`.

dhruba:
Hi folks, can somebody describe the use-case for this feature? Is a per-file-checksum better than having a per-block checksum?

zhichao-cao:
> a

Hi Dhruba, the full file checksum has the follow use-cases: 1) to provide better correctness guarantee of SST files. Comparing the name and size in the file system with the information in Manifest does not provide strong enough information to ensure the correctness of the SST file. 2) When a file is copied to a remote host or file system, file might be broken during the transmission. Check the SST file checksum will be a way of validation. 3) When RocksDB is used for disaggregate storage (i.g., remote file system or object storage), the storage system can provide the checksum of a file being stored, comparing the checksum of an SST file when it is generated and the checksum when it is store will provide better protection. Files are more vulnerable in this case than storing directly in local file system.

db/version_edit.cc
```diff
@@ -18,7 +18,9 @@
 #include "util/string_util.h"
 
 namespace rocksdb {
-// The default sst file checksum function name.
+// The unknown file checksum.
+const std::string kUnknownFileChecksum("0");
```
riversand963: Can we use empty string?

table/block_based/block_based_table_builder.h
```diff
@@ -92,10 +92,10 @@ class BlockBasedTableBuilder : public TableBuilder {
   // Get table properties
   TableProperties GetTableProperties() const override;
 
-  // Get checksum value
-  uint32_t GetFileChecksum() const override { return file_checksum_; }
+  // Get file checksum
+  std::string GetFileChecksum() const override { return file_checksum_; }
```
riversand963: Nit: for all table builder classes, the return type of `GetFileChecksum()` can actually be `const std::string&`, given that the file checksum won't change after file is written.

util/file_checksum_helper.h
```diff
@@ -9,25 +9,29 @@
...
 // will be used as the default checksum method for SST file checksum
 class FileChecksumFuncCrc32c : public FileChecksumFunc {
  public:
-  uint32_t Extend(uint32_t init_checksum, const char* data, size_t n) override {
+  std::string Extend(const std::string& init_checksum, const char* data,
+                     size_t n) override {
     assert(data != nullptr);
-    return crc32c::Extend(init_checksum, data, n);
+    uint32_t checksum_value = ParseUint32(init_checksum);
+    return ToString(crc32c::Extend(checksum_value, data, n));
```
riversand963: Let's not use `ToString()` which uses `std::to_string()`.  We should consider endianness and convert the result to byte array of 8. An example is https://github.com/facebook/rocksdb/blob/master/examples/multi_processes_example.cc#L74. We probably have something similar for uint32_t, but you may want to search the codebase.

riversand963: Just realized https://github.com/facebook/rocksdb/blob/master/port/port_posix.h#L99 can be used to detect endianness instead of writing a simple function.

zhichao-cao:
@riversand963 Thanks for the comments and review!

PR #6227, opened 2019-12-20
Add Transaction::MultiGet to db_stress

anand1976:
Call Transaction::MultiGet from TestMultiGet() in db_stress. We add some Puts/Merges/Deletes into the transaction in order to exercise MultiGetFromBatchAndDB. There is no data verification on read, just check status. There is currently no read data verification in db_stress as it requires synchronization with writes. It needs to be tackled separately.

Test plan:
make crash_test_with_txn

### DIFF
--- a/db_stress_tool/no_batched_ops_stress.cc
+++ b/db_stress_tool/no_batched_ops_stress.cc
@@ -181,15 +181,17 @@ class NonBatchedOpsStressTest : public StressTest {
       keys.emplace_back(key_str.back());
 #ifndef ROCKSDB_LITE
       if (FLAGS_use_txn) {
+        // With a 1 in 10 probability, insert the just added key in the batch
+        // into the transaction. This will create an overlap with the MultiGet
+        // keys and exercise some corner cases in the code
         if (thread->rand.OneIn(10)) {
           int op = thread->rand.Uniform(2);
           Status s;
           switch (op) {
             case 0:
-            case 1:
-            {
-              uint32_t value_base = thread->rand.Next() %
-                                    thread->shared->UNKNOWN_SENTINEL;
+            case 1: {
+              uint32_t value_base =
+                  thread->rand.Next() % thread->shared->UNKNOWN_SENTINEL;
               char value[100];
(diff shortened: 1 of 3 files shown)


db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -166,12 +166,64 @@ class NonBatchedOpsStressTest : public StressTest {
...
+              Slice v(value, sz);
+              if (op == 0) {
+                s = txn->Put(cfh, keys.back(), v);
+              } else {
+                s = txn->Merge(cfh, keys.back(), v);
+              }
+              break;
+            }
+            default:
+              assert(op == 2);
```
maysamyabandeh: Would not be more clear if we add "case 2" explicitly?

db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -166,12 +166,64 @@ class NonBatchedOpsStressTest : public StressTest {
...
+      }
+#endif
     }
-    db_->MultiGet(read_opts, cfh, num_keys, keys.data(), values.data(),
-                  statuses.data());
+
+    if (!FLAGS_use_txn) {
+      db_->MultiGet(read_opts, cfh, num_keys, keys.data(), values.data(),
+                    statuses.data());
+     } else {
```
maysamyabandeh: Seems like it needs a "make format"

db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -166,12 +166,64 @@ class NonBatchedOpsStressTest : public StressTest {
...
+            case 0:
+            case 1:
+            {
+              uint32_t value_base = thread->rand.Next() %
+                                    thread->shared->UNKNOWN_SENTINEL;
+              char value[100];
+              size_t sz = GenerateValue(value_base, value, sizeof(value));
+              Slice v(value, sz);
+              if (op == 0) {
+                s = txn->Put(cfh, keys.back(), v);
```
maysamyabandeh: Can you document the rationale behind choosing keys.back() as the key? Would any key work here?

PR #6228, opened 2019-12-20
Add oldest snapshot sequence property

Connor1996:
Add oldest snapshot sequence property, so we can use `db.GetProperty("rocksdb.oldest-snapshot-sequence")` to get the sequence number of the oldest snapshot.

### DIFF
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -524,6 +524,7 @@ TEST_F(DBBasicTest, Snapshot) {
     ASSERT_EQ(1U, GetNumSnapshots());
     uint64_t time_snap1 = GetTimeOldestSnapshots();
     ASSERT_GT(time_snap1, 0U);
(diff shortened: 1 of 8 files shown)


maysamyabandeh:
@yiwu-arbug  A PR with no description?

yiwu-arbug:
@Connor1996 update description?

Connor1996:
@maysamyabandeh @yiwu-arbug updated.

yiwu-arbug:
Use case: when we want to drop a region in tikv (think of it as a shard) through `DeleteFilesInRange`, we want to make sure the the oldest snapshot is newer than the timestamp when we drop the region logically, thus we need the API.

maysamyabandeh:
@Connor1996 could you do a rebase?

Connor1996:
@maysamyabandeh Done

PR #6244, opened 2019-12-24
Fix a clang analyzer report, and 'analyze' make rule

pdillinger:
Summary: Clang analyzer was falsely reporting on use of txn=nullptr.
Added a new const variable so that it can properly prune impossible
control flows.

Also, 'make analyze' previously required setting USE_CLANG=1 as an
environment variable, not a make variable, or else compilation errors
like

g++: error: unrecognized command line option ‘-Wshorten-64-to-32’

Now USE_CLANG is not required for 'make analyze' (it's implied) and you
can do an incremental analysis (recompile what has changed) with
'USE_CLANG=1 make analyze_incremental'

Test Plan: 'make -j24 analyze', 'make crash_test'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1097,6 +1097,9 @@ parallel_check: $(TESTS)
 		TMPD=$(TMPD) J=$(J) db_test=0 parloop;
 
 analyze: clean
+	USE_CLANG=1 $(MAKE) analyze_incremental
+
+analyze_incremental:
 	$(CLANG_SCAN_BUILD) --use-analyzer=$(CLANG_ANALYZER) \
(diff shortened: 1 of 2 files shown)


pdillinger:
> Thanks @pdillinger for the fix. LGTM.
> If the clang analyzer can deduce the value of txn after `NewTxn(.., &txn)`?

I believe it just assumes no bug on that path if address passed to another function (a conservative assumption). It just also assumes the value of FLAGS_use_txn might change in a callee or other thread (a liberal assumption). (Coverity would not report this FP.)

> If not, do we need to check the value of `NewTxn()`?

The analyzer doesn't care, but we probably should, so I will.

PR #6258, opened 2020-01-03
Increase max_log_size in FlushJob to 1024 bytes

maysamyabandeh:
When measure_io_stats_ is enabled, the volume of logging is beyond the default limit of 512 size. The patch allows the EventLoggerStream to change the limit, and also sets it to 1024 for FlushJob.

### DIFF
--- a/db/flush_job.cc
+++ b/db/flush_job.cc
@@ -249,7 +249,7 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker,
   }
   RecordFlushIOStats();
 
-  auto stream = event_logger_->LogToBuffer(log_buffer_);
+  auto stream = event_logger_->LogToBuffer(log_buffer_, 1024);
   stream << "job" << job_context_->job_id << "event"
          << "flush_finished";
   stream << "output_compression"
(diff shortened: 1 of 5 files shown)


logging/event_logger.h
```diff
@@ -182,12 +183,16 @@ class EventLogger {
 
   explicit EventLogger(Logger* logger) : logger_(logger) {}
   EventLoggerStream Log() { return EventLoggerStream(logger_); }
-  EventLoggerStream LogToBuffer(LogBuffer* log_buffer) {
-    return EventLoggerStream(log_buffer);
+  EventLoggerStream LogToBuffer(
+      LogBuffer* log_buffer,
+      const size_t max_log_size = LogBuffer::kDefaultMaxLogSize) {
```
anand1976: Can we avoid default values for function parameters? Either explicitly specify the value at call sites or overload the function.

db/flush_job.cc
```diff
@@ -249,7 +249,7 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker,
   }
   RecordFlushIOStats();
 
-  auto stream = event_logger_->LogToBuffer(log_buffer_);
+  auto stream = event_logger_->LogToBuffer(log_buffer_, 1024);
```
anand1976: Add a comment to explain the reason for 1024.

PR #6277, opened 2020-01-09
Fix compilation under LITE

riversand963:
Fix compilation under LITE by putting `#ifndef ROCKSDB_LITE` around a code block.


riversand963:
The `DefaultEnvWithoutDirectIO/EnvPosixTestWithParam.DecreaseNumBgThreads` is known to be flaky.

PR #6278, opened 2020-01-09
Remove earlier partial BlobDB GC implementation

ltamasi:
Summary:
In addition to removing the earlier partially implemented garbage collection
logic from the BlobDB codebase, the patch also removes the test cases (as well as
the related sync points, as appropriate) that were only relevant for the old
implementation, and reworks the remaining ones so they use the new GC logic.

Test Plan:
`make check`


zhiliaoniu:
Hello, after you remove the previous garbage collection logic, if I write a large number of kvs with a ttl of 1 year into rocksdb, most of the kvs will be overwritten repeatedly, so that the disk will be full in a few days However, because the ttl setting is 1 year, the entire blob file has not expired. According to the current logic, the blob file cannot be compressed and deleted. What should I do in this scenario?

ltamasi:
> Hello, after you remove the previous garbage collection logic, if I write a large number of kvs with a ttl of 1 year into rocksdb, most of the kvs will be overwritten repeatedly, so that the disk will be full in a few days However, because the ttl setting is 1 year, the entire blob file has not expired. According to the current logic, the blob file cannot be compressed and deleted. What should I do in this scenario?

Not entirely sure what you mean. GC was never really supported for TTL blob files, since the old code this PR removed was never fully functional. It was actually dead code; there was no way for users to enable it and it was only called in tests.

PR #6279, opened 2020-01-09
Consider all compaction input files to compute the oldest ancestor time

sagar0:
Look at all compaction input files to compute the oldest ancestor time. 

In #5992 we changed how creation_time (aka oldest-ancestor-time) table property of compaction output files is computed from max(creation-time-of-all-compaction-inputs) to min(creation-time-of-all-inputs). This exposed a bug where, during compaction, the creation_time:s of only the L0 compaction inputs were being looked at, and all other input levels were being ignored. This PR fixes the issue. 
Some TTL compactions when using Level-Style compactions might not have run due to this bug. 

Test plan:
Enhanced the unit tests to validate that the correct time is propagated to the compaction outputs.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -7,6 +7,7 @@
 * A new `OptimisticTransactionDBOptions` Option that allows users to configure occ validation policy. The default policy changes from kValidateSerial to kValidateParallel to reduce mutex contention.
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


sagar0:
The Visual Studio 14 test failure is not related to this PR.

Issue #6280, opened 2020-01-10
unable to set WAL disable or enable correctly when concurrently write 


chenyou-fdu:
Not sure if I misuse or misunderstand the way to enable / disable WAL when uses multiple column families:
I try to concurrently write to CF A (with FIFO style compaction) with WAL disabled and CF default with WAL enabled, and in large concurrency situation, I found that WAL files is appended with content from CF A (with WAL disable)

I try to dig into source code a bit, found that when MergeBatch happens, it seems only care if leader writer batch has WAL enabled, and simply merged all other batches together regardless if these batches have WAL enabled or not, and then write to WAL files. Not sure if I have understand this part correctly. Could some expert to help me out on this issue, thanks!

### Expected behavior
only the data with WAL enable will be written to wal file
 
### Actual behavior
the data with WAL disable is also writtent to wal file

### Steps to reproduce the behavior
version v6.4.6, uses 
```c++
int main(int argc, char** argv) {
    rocksdb::DB* db;
    rocksdb::DBOptions db_options;
    db_options = *db_options.IncreaseParallelism(16);
    db_options.create_if_missing = true;
    db_options.max_background_jobs = 32;
    db_options.db_write_buffer_size = 268435456;
    db_options.stats_dump_period_sec = 10;
    db_options.max_open_files = -1;
    db_options.max_subcompactions = 16;
    db_options.keep_log_file_num = 20;
    Options options;
    options.OptimizeLevelStyleCompaction();
    options.IncreaseParallelism(16);
    options.create_if_missing = true;
    options.max_background_jobs = 32;
    options.db_write_buffer_size = 268435456;
    options.stats_dump_period_sec = 10;
    options.max_open_files = -1;
    options.max_subcompactions = 16;
    options.max_log_file_size = 25 * 1024 * 1024;
    options.keep_log_file_num = 20;
    options.max_background_jobs = 32;
    options.db_write_buffer_size = 671088640;
    rocksdb::Status status =
            rocksdb::DB::Open(options, "/home/rocksdb/data_data", &db);
    assert(status.ok());
    // create column family
    rocksdb::ColumnFamilyHandle* cf;
    ColumnFamilyOptions cf_option;
    CompactionOptionsFIFO cp_option_fifo;
    cf_option.compaction_options_fifo = cp_option_fifo;
    cf_option.compression = kNoCompression;
    cf_option.write_buffer_size = 6291456;
    cf_option.max_write_buffer_number = 2;
[...]

PR #6290, opened 2020-01-13
Separate enable-WAL and disable-WAL writer to avoid unwanted data in log files

chenyou-fdu:
When we do concurrently writes, and different write operations will have WAL enable or disable.
But the data from write operation with WAL disabled will still be logged into log files, which will lead to extra disk write/sync since we do not want any guarantee for these part of data.

Detail can be found in https://github.com/facebook/rocksdb/issues/6280. This PR avoid mixing the two types in a write group. The advantage is simpler reasoning about the write group content

### DIFF
--- a/db/db_impl/db_impl_write.cc
+++ b/db/db_impl/db_impl_write.cc
@@ -945,9 +945,6 @@ WriteBatch* DBImpl::MergeBatch(const WriteThread::WriteGroup& write_group,
     // interface
     merged_batch = tmp_batch;
     for (auto writer : write_group) {
-      if(writer->disable_wal) {
-          continue;
-      }
       if (!writer->CallbackFailed()) {
         WriteBatchInternal::Append(merged_batch, writer->batch,
                                    /*WAL_only*/ true);
@@ -1012,9 +1009,6 @@ Status DBImpl::WriteToWAL(const WriteThread::WriteGroup& write_group,
     write_group.leader->log_used = logfile_number_;
   } else if (write_with_wal > 1) {
     for (auto writer : write_group) {
-      if(writer->disable_wal) {
-          continue;
-      }
       writer->log_used = logfile_number_;
     }
   }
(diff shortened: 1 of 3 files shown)


maysamyabandeh:
Have you tried to reproduce the problem with a unit test?

chenyou-fdu:
@maysamyabandeh I've added UT to cover this issue, which is quite navie that I read the log file, and simply to judge if the disabled WAL data in log files.

maysamyabandeh:
@chenyou-fdu I am not sure what this unit test does. But I think there are simpler unit tests to verify that. One could measure WAL_FILE_BYTES to see if there is no extra data written to the WAL.

chenyou-fdu:
@maysamyabandeh Thanks for the advise! I've updated UT using WAL_FILE_BYTES.
I set the No WAL data quite large (100KB in value) while WAL data a small one, then uses this statistic to verify is there extra large data written to wal files.
Please help to review.

maysamyabandeh:
Thanks @chenyou-fdu. Could you also confirm that this unit test fails on master?

chenyou-fdu:
@maysamyabandeh Please see this [PR](https://github.com/facebook/rocksdb/pull/6303), I've ported this unit test to master, and the CI fails on this unit test, please check [log link](https://travis-ci.org/facebook/rocksdb/jobs/637765911?utm_medium=notification&utm_source=github_status).

chenyou-fdu:
more unit test failed result in master branch [link](https://travis-ci.org/facebook/rocksdb/jobs/637851330?utm_medium=notification&utm_source=github_status)

maysamyabandeh:
Thanks @chenyou-fdu. No need for proof, your confirmation is sufficient.

What do you think about this fix instead? The comment throughout the code assumes that disableWAL and !disableWAL writes are not mixed and we would need a through check to ensure that having them mixed does not cause other problems. I think the logic becomes simpler if we simply do not mix the two.

```
diff --git a/db/write_thread.cc b/db/write_thread.cc
index 1ded68fde..046d804d5 100644
--- a/db/write_thread.cc
+++ b/db/write_thread.cc
@@ -444,7 +444,7 @@ size_t WriteThread::EnterAsBatchGroupLeader(Writer* leader,
       break;
     }

-    if (!w->disable_wal && leader->disable_wal) {
+    if (w->disable_wal != leader->disable_wal) {
       // Do not include a write that needs WAL into a batch that has
       // WAL disabled.
       break;

chenyou-fdu:
@maysamyabandeh Sure, this fix is conciser and safer. I've updated PR using this way.

PR #6306, opened 2020-01-16
Make DBCompactionTest.SkipStatsUpdateTest more robust

ltamasi:
Summary:
Currently, this test case tries to infer whether
`VersionStorageInfo::UpdateAccumulatedStats` was called during open by
checking the number of files opened against an arbitrary threshold (10).
This makes the test brittle and results in sporadic failures. The patch
changes the test case to use sync points to directly test whether
`UpdateAccumulatedStats` was called.

Test Plan:
`make check`

### DIFF
--- a/db/db_compaction_test.cc
+++ b/db/db_compaction_test.cc
@@ -400,6 +400,7 @@ TEST_F(DBCompactionTest, SkipStatsUpdateTest) {
   // The test will need to be updated if the internal behavior changes.
 
   Options options = DeletionTriggerOptions(CurrentOptions());
+  options.disable_auto_compactions = true;
   options.env = env_;
   DestroyAndReopen(options);
   Random rnd(301);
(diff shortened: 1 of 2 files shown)


db/version_set.cc
```diff
@@ -2112,6 +2112,8 @@ bool Version::MaybeInitializeFileMetaData(FileMetaData* file_meta) {
 }
 
 void VersionStorageInfo::UpdateAccumulatedStats(FileMetaData* file_meta) {
+  TEST_SYNC_POINT("VersionStorageInfo::UpdateAccumulatedStats");
```
riversand963: Nit: maybe use `TEST_SYNC_POINT_CALL_BACK`.

db/db_compaction_test.cc
```diff
@@ -400,6 +400,7 @@ TEST_F(DBCompactionTest, SkipStatsUpdateTest) {
   // The test will need to be updated if the internal behavior changes.
 
   Options options = DeletionTriggerOptions(CurrentOptions());
+  options.disable_auto_compactions = true;
```
riversand963: Is this necessary?

db/db_compaction_test.cc
```diff
@@ -410,31 +411,32 @@ TEST_F(DBCompactionTest, SkipStatsUpdateTest) {
...
   options.skip_stats_update_on_db_open = false;
-  env_->random_file_open_counter_.store(0);
   Reopen(options);
 
-  // Since we do a normal stats update on db-open, there
-  // will be more random open files.
-  ASSERT_GT(env_->random_file_open_counter_.load(), kMaxFileOpenCount);
+  ASSERT_GT(update_acc_stats_called, 0);
+
+  SyncPoint::GetInstance()->DisableProcessing();
```
riversand963: It's good practice to clear all callbacks after test.

db/version_set.cc
ltamasi: `TEST_SYNC_POINT_CALLBACK` is only needed when you want to pass a pointer to the callback, right? At least, that's my understanding based on the header :)

```
  // The argument to the callback is passed through from
  // TEST_SYNC_POINT_CALLBACK(); nullptr if TEST_SYNC_POINT or
  // TEST_IDX_SYNC_POINT was used.
  void SetCallBack(const std::string& point,
                   const std::function<void(void*)>& callback);
```

riversand963: That's correct. In this situation, I usually pass a `nullptr` and use `TEST_SYNC_POINT_CALL_BACK`. In this PR, this is not necessary.

db/db_compaction_test.cc
ltamasi: Yes; the flakiness is in part caused by a compaction starting right after the open and creating a new `Version` (which results in those stats getting updated).

db/db_compaction_test.cc
ltamasi: Will do, thanks!

PR #6307, opened 2020-01-17
Expose atomic flush option in C API

mappum:
This PR adds a `rocksdb_options_set_atomic_flush` function to the C API.


include/rocksdb/c.h
```diff
@@ -737,6 +737,8 @@ rocksdb_block_based_options_set_pin_top_level_index_and_filter(
     rocksdb_block_based_table_options_t*, unsigned char);
 extern ROCKSDB_LIBRARY_API void rocksdb_options_set_block_based_table_factory(
     rocksdb_options_t* opt, rocksdb_block_based_table_options_t* table_options);
+extern ROCKSDB_LIBRARY_API void rocksdb_options_set_atomic_flush(
```
ltamasi: Thanks for contributing @mappum !

The patch looks good to me, just one comment: could you move this declaration to right below `rocksdb_options_set_ratelimiter` so the order is the same in the header and the `.cc` file?

mappum: :heavy_check_mark: Updated as requested.

PR #6309, opened 2020-01-17
fix build warnnings on MSVC

burtonli:
Fix build warnings on MSVC. @siying

### DIFF
--- a/db_stress_tool/db_stress_common.cc
+++ b/db_stress_tool/db_stress_common.cc
@@ -62,7 +62,7 @@ void InitializeHotKeyGenerator(double alpha) {
 int64_t GetOneHotKeyID(double rand_seed, int64_t max_key) {
   int64_t low = 1, mid, high = zipf_sum_size, zipf = 0;
   while (low <= high) {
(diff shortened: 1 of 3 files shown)


db_stress_tool/db_stress_common.cc
```diff
@@ -62,7 +62,7 @@ void InitializeHotKeyGenerator(double alpha) {
 int64_t GetOneHotKeyID(double rand_seed, int64_t max_key) {
   int64_t low = 1, mid, high = zipf_sum_size, zipf = 0;
   while (low <= high) {
-    mid = std::floor((low + high) / 2);
+    mid = static_cast<int64_t>(std::floor((low + high) / 2));
```
Orvid: This entire line doesn't make sense. `(low + high) / 2` is already a truncating integer division. All the call to `std::floor` here does is lose precision by converting the result to a `double` then back to an `int64_t`.

burtonli: yes, the std::floor does nothing here. updated the PR

PR #6313, opened 2020-01-18
Double Crash in kPointInTimeRecovery with TransactionDB

maysamyabandeh:
In WritePrepared there could be gap in sequence numbers. This breaks the trick we use in kPointInTimeRecovery which assume the first seq in the log right after the corrupted log is one larger than the last seq we read from the logs. To let this trick keep working, we add a dummy entry with the expected sequence to the first log right after recovery.
Also in WriteCommitted, if the log right after the corrupted log is empty, since it has no sequence number to let the sequential trick work, it is assumed as unexpected behavior. This is however expected to happen if we close the db after recovering from a corruption and before writing anything new to it. To remedy that, we apply the same technique by writing a dummy entry to the log that is created after the corrupted log.

### DIFF
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -1361,7 +1361,7 @@ class DBImpl : public DB {
   // corrupted_log_found is set to true if we recover from a corrupted log file.
   Status RecoverLogFiles(const std::vector<uint64_t>& log_numbers,
                          SequenceNumber* next_sequence, bool read_only,
-                         bool* corrupted_log_found = nullptr);
+                         bool* corrupted_log_found);
 
   // The following two methods are used to flush a memtable to
   // storage. The first one is used at database RecoveryTime (when the
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -1113,7 +1113,7 @@ class DBImpl : public DB {
   virtual Status Recover(
       const std::vector<ColumnFamilyDescriptor>& column_families,
       bool read_only = false, bool error_if_log_file_exist = false,
-      bool error_if_data_exists_in_logs = false);
+      bool error_if_data_exists_in_logs = false, uint64_t* next_sequence_ptr = nullptr);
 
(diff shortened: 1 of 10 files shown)


db/db_impl/db_impl_open.cc
```diff
@@ -973,6 +980,9 @@ Status DBImpl::RecoverLogFiles(const std::vector<uint64_t>& log_numbers,
         status = Status::OK();
         stop_replay_for_corruption = true;
         corrupted_log_number = log_number;
+        if (corrupted_log_found != nullptr) {
```
anand1976: Should this be here or at line# 837? IIUC, we need to insert a dummy record only if the newest log is corrupted, right?

db/db_impl/db_impl.h
```diff
@@ -1355,8 +1358,10 @@ class DBImpl : public DB {
       JobContext* job_context, LogBuffer* log_buffer, Env::Priority thread_pri);
 
   // REQUIRES: log_numbers are sorted in ascending order
+  // corrupted_log_found is set to true if we recover from a corrupted log file.
   Status RecoverLogFiles(const std::vector<uint64_t>& log_numbers,
-                         SequenceNumber* next_sequence, bool read_only);
+                         SequenceNumber* next_sequence, bool read_only,
+                         bool* corrupted_log_found = nullptr);
```
anand1976: Make this parameter non-optional. Its only called from 1 or 2 places I think, so should be easy to do.

db/db_impl/db_impl_open.cc
maysamyabandeh: Line #837 is reached when we have a double crash, and the logs after corrupted one are created after the 1st recovery. This is actually a case that we don't need to insert dummy since the continuity of sequence numbers are already taken care of in the 1st recovery.
You are right that the cases that do insert the dummy record could be more limited but limiting it would not bring any performance advantage yet bring more complexity on reasoning which case is which. It is simpler to always start the first log after recovery with a dummy record if there is a corrupted log that we have skipped.

PR #6315, opened 2020-01-18
Correct pragma once problem with Bazel on Windows

matthewvon:
This is a simple edit to have two #include file paths be consistent within range_del_aggregator.{h,cc} with everywhere else.  

The impact of this inconsistency is that it actual breaks a Bazel based build on the Windows platform.  The same pragma once failure occurs with both Windows Visual C++ 2019 and clang for Windows 9.0.  Bazel's "sandboxing" of the builds causes both compilers to not properly recognize "rocksdb/types.h" and "include/rocksdb/types.h" to be the same file (also comparator.h).  My guess is that the backslash versus forward slash mixing within path names is the underlying issue.  

But, everything builds fine once the include paths in these two source files are consistent with the rest of the repository.


matthewvon:
Both failures related to Travis java build.  First failure was something about bad tail block on jar files.  Second failure, with my changes removed, appears to be a download issue for missing jar files.  Adding back my edits.

matthewvon:
Merging HEAD from Facebook/rocksdb moved Travis appveyor from Java bugs to Windows bugs.  This is infrastructure ... or some unrelated change in HEAD.

PR #6320, opened 2020-01-21
Update version for next release, 6.7.0

gfosco:
Adjusted history for 6.6.1 and 6.6.2, switched master version to 6.7.0.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,7 @@
 # Rocksdb Change Log
 ## Unreleased
+
+## 6.7.0 (01/21/2020)
 ### Public API Change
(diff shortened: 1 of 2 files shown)


HISTORY.md
```diff
@@ -26,6 +23,20 @@
...
+### Bug Fixes
+* Fixed a bug where non-L0 compaction input files were not considered to compute the `creation_time` of new compaction outputs.
+
+## 6.6.1 (01/02/2020)
+### Bug Fixes
+* Fix a bug in WriteBatchWithIndex::MultiGetFromBatchAndDB, which is called by Transaction::MultiGet, that causes due to stale pointer access when the number of keys is > 32
+* Fixed two performance issues related to memtable history trimming. First, a new SuperVersion is now created only if some memtables were actually trimmed. Second, trimming is only scheduled if there is at least one flushed memtable that is kept in memory for the purposes of transaction conflict checking.
+* BlobDB no longer updates the SST to blob file mapping upon failed compactions.
+* Fix a bug in which a snapshot read through an iterator could be affected by a DeleteRange after the snapshot (#6062).
+* Fixed a bug where BlobDB was comparing the `ColumnFamilyHandle` pointers themselves instead of only the column family IDs when checking whether an API call uses the default column family or not.
```
ltamasi: Not sure if we usually deduplicate fixes that get cherry-picked into patch releases like this but if we do, we should also remove this item from the 6.7.0 section.

gfosco: Removed the duplicated line from 6.7.. Thanks for catching that.

PR #6321, opened 2020-01-21
Correct pragma once problem with Bazel on Windows

matthewvon:
This is a simple edit to have two #include file paths be consistent within range_del_aggregator.{h,cc} with everywhere else.

The impact of this inconsistency is that it actual breaks a Bazel based build on the Windows platform. The same pragma once failure occurs with both Windows Visual C++ 2019 and clang for Windows 9.0. Bazel's "sandboxing" of the builds causes both compilers to not properly recognize "rocksdb/types.h" and "include/rocksdb/types.h" to be the same file (also comparator.h). My guess is that the backslash versus forward slash mixing within path names is the underlying issue.

But, everything builds fine once the include paths in these two source files are consistent with the rest of the repository.


ltamasi:
Hi @matthewvon ! Did you revert your changes intentionally?

matthewvon:
@ltamasi I need help.  I reverted the changes to prove that the appveyor test fails both with my changes and without my changes.  I am open to any solution to get this PR certified.

ltamasi:
The AppVeyor issues are unrelated to your changes; feel free to put them back. Also, if you could take care of the same issue in these two files under `utilities/persistent_cache`, it would be great!

```
$ git grep "#include \"include/rocksdb"
db/range_del_aggregator.cc:#include "include/rocksdb/comparator.h"
db/range_del_aggregator.cc:#include "include/rocksdb/types.h"
db/range_del_aggregator.h:#include "include/rocksdb/comparator.h"
db/range_del_aggregator.h:#include "include/rocksdb/types.h"
utilities/persistent_cache/block_cache_tier_file_buffer.h:#include "include/rocksdb/comparator.h"
utilities/persistent_cache/hash_table.h:#include "include/rocksdb/env.h"
```

matthewvon:
@ltamasi I was aware of the two additional files you mentioned.  But they were not part of my build, so changing them was not something I could readily prove / disprove as helpful.  However, I have submitted my two files and those additional two files in the latest commit.

PR #6322, opened 2020-01-21
Fix queue manipulation in WriteThread::BeginWriteStall()

anand1976:
When there is a write stall, the active write group leader calls ```BeginWriteStall()``` to walk the queue of writers and remove any with the ```no_slowdown``` option set. There was a bug in the code which updated the back pointer but not the forward pointer (```link_newer```), corrupting the list and causing some threads to wait forever. This PR fixes it.

Test plan:
Add a unit test in db_write_test

### DIFF
--- a/db/db_write_test.cc
+++ b/db/db_write_test.cc
@@ -39,6 +39,101 @@ TEST_P(DBWriteTest, SyncAndDisableWAL) {
   ASSERT_TRUE(dbfull()->Write(write_options, &batch).IsInvalidArgument());
 }
 
+TEST_P(DBWriteTest, WriteThreadHangOnWriteStall) {
+  Options options = GetOptions();
+  options.level0_stop_writes_trigger = options.level0_slowdown_writes_trigger = 4;
+  std::vector<port::Thread> threads;
+  std::atomic<int> thread_num(0);
+  port::Mutex mutex;
+  port::CondVar cv(&mutex);
+
+  Reopen(options);
+
+  std::function<void()> write_slowdown_func = [&]() {
+    int a = thread_num.fetch_add(1);
+    std::string key = "foo" + std::to_string(a);
+    WriteOptions wo;
+    wo.no_slowdown = false;
+    dbfull()->Put(wo, key, "bar");
+  };
+  std::function<void()> write_no_slowdown_func = [&]() {
+    int a = thread_num.fetch_add(1);
+    std::string key = "foo" + std::to_string(a);
+    WriteOptions wo;
+    wo.no_slowdown = true;
+    dbfull()->Put(wo, key, "bar");
+  };
+  std::function<void(void *)> unblock_main_thread_func = [&](void *) {
+    cv.SignalAll();
+  };
+
+  // Create 3 L0 files and schedule 4th without waiting
+  Put("foo" + std::to_string(thread_num.fetch_add(1)), "bar");
+  Flush();
+  Put("foo" + std::to_string(thread_num.fetch_add(1)), "bar");
+  Flush();
+  Put("foo" + std::to_string(thread_num.fetch_add(1)), "bar");
+  Flush();
+  Put("foo" + std::to_string(thread_num.fetch_add(1)), "bar");
+
+  rocksdb::SyncPoint::GetInstance()->SetCallBack(
+    "WriteThread::JoinBatchGroup:Start", unblock_main_thread_func);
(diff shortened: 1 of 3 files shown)


db/write_thread.cc
```diff
@@ -345,6 +345,7 @@ void WriteThread::BeginWriteStall() {
       w->status = Status::Incomplete("Write stall");
       SetState(w, STATE_COMPLETED);
       w = prev->link_older;
+      w->link_newer = (prev == &write_stall_dummy_) ? nullptr : prev;
```
maysamyabandeh: What if w is nullptr? Why link_newer cannot be assigned to write_stall_dummy_ when prev is write_stall_dummy_?
I am thinking the following change would be easier to read as the meaning of "w" should not change through the code block.
```
if (prev->link_older != nullptr) {
   prev->link_older->link_newer = prev;
} 
w = prev->link_older;

Comment (anand1976) [reply db/write_thread.cc]: w will not be nullptr, as this function is called by a write group leader and the loop will stop at the first (or last, depending on how you look at it) writer in the write group. But I agree its safer to check for nullptr in case this gets called in other contexts.

link_newer cannot be assigned to write_stall_dummy_ as elsewhere in the code, traversal of link_newer assumes it terminates with a nullptr.

Comment (maysamyabandeh) [reply db/write_thread.cc]: If we don't enter the loop because w->write_group == nullptr the last writer before write_stall_dummy_ will have a link_newer pointed to write_stall_dummy_, right?

Comment (anand1976) [reply db/write_thread.cc]: Not necessarily. CreateMissingNewerLinks() may not have had a chance to run.

File: db/write_thread.cc [hunk 2]
```diff
@@ -355,7 +358,12 @@ void WriteThread::BeginWriteStall() {
 void WriteThread::EndWriteStall() {
   MutexLock lock(&stall_mu_);
 
+  // Unlink write_stall_dummy_ from the write queue. This will unblock
+  // pending write threads to enqueue themselves
   assert(newest_writer_.load(std::memory_order_relaxed) == &write_stall_dummy_);
+  if (write_stall_dummy_.link_older) {
```
maysamyabandeh: can we assert(write_stall_dummy_.link_older != nullptr)? If we are in the write thread, then there must be at least one write in the queue.

db/db_write_test.cc
```diff
@@ -39,6 +39,101 @@ TEST_P(DBWriteTest, SyncAndDisableWAL) {
...
+  rocksdb::SyncPoint::GetInstance()->SetCallBack(
+    "WriteThread::JoinBatchGroup:Start", unblock_main_thread_func);
+  rocksdb::SyncPoint::GetInstance()->LoadDependency(
+    {{"DBWriteTest::WriteThreadHangOnWriteStall:1",
+      "DBImpl::BackgroundCallFlush:start"},
+     {"DBWriteTest::WriteThreadHangOnWriteStall:2",
+      "DBImpl::WriteImpl:BeforeLeaderEnters"},
+     // Make compaction start wait for the write stall to be detected and
+     // implemented by a write group leader
+     //{"DBImpl::DelayWrite:Wait",
```
maysamyabandeh: If this line is not necessary, should we remove it?

db/db_write_test.cc
```diff
@@ -39,6 +39,101 @@ TEST_P(DBWriteTest, SyncAndDisableWAL) {
...
+  FlushOptions fopt;
+  fopt.wait = false;
+  dbfull()->Flush(fopt);
+
+  // Create a mix of slowdown/no_slowdown write threads
+  mutex.Lock();
+  // First leader
+  threads.emplace_back(write_slowdown_func);
+  cv.Wait();
+  //mutex.Lock();
```
maysamyabandeh: ditto

db/write_thread.cc
```diff
@@ -355,7 +358,13 @@ void WriteThread::BeginWriteStall() {
 void WriteThread::EndWriteStall() {
   MutexLock lock(&stall_mu_);
 
+  // Unlink write_stall_dummy_ from the write queue. This will unblock
+  // pending write threads to enqueue themselves
   assert(newest_writer_.load(std::memory_order_relaxed) == &write_stall_dummy_);
+  assert(write_stall_dummy_.link_older != nullptr);
```
maysamyabandeh: I guess after this assert, the next conditional branch is unnecessary?

anand1976:
Appveyor failure seems unrelated

PR #6330, opened 2020-01-24
Implement PinnableSlice::remove_prefix

maysamyabandeh:
The function was left unimplemented. Although we currently don't have a use for that it was declared with an assert(0) to prevent mistakenly using the remove_prefix of the parent class. The function body  with only assert(0) however causes issues with some compiler's warning levels. The patch implements the function to avoid the warning.
It also piggybacks some minor code warning for unnecessary semicolons after the function definition.s

### DIFF
--- a/include/rocksdb/slice.h
+++ b/include/rocksdb/slice.h
@@ -195,8 +195,15 @@ class PinnableSlice : public Slice, public Cleanable {
     }
   }
 
-  void remove_prefix(size_t /*n*/) {
-    assert(0);  // Not implemented
+  void remove_prefix(size_t n) {
+    assert(n <= size());
+    if (pinned_) {
+      data_ += n;
+      size_ -= n;
+    } else {
+      buf_->erase(0, n);
+      PinSelf();
+    }
   }
 
   void Reset() {
--- a/include/rocksdb/cache.h
+++ b/include/rocksdb/cache.h
@@ -255,7 +255,7 @@ class Cache {
   // Always delete the DB object before calling this method!
   virtual void DisownData(){
       // default implementation is noop
-  };
+  }
 
   // Apply callback to all entries in the cache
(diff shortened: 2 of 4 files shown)


include/rocksdb/slice.h
```diff
@@ -195,8 +195,15 @@ class PinnableSlice : public Slice, public Cleanable {
...
 
-  void remove_prefix(size_t /*n*/) {
-    assert(0);  // Not implemented
+  void remove_prefix(size_t n) {
+    assert(n <= size());
+    if (pinned_) {
+      data_ += n;
+      size_ -= n;
+    } else {
+      buf_->erase(0, n);
```
anand1976: Is it safe to do this if buf was passed to the constructor and thus not owned by the PinnableSlice? Also, it might be inefficient. Why not just change data_ and size_ in this case too?

maysamyabandeh: The logic was to keep data_, and size_ consistent with what buf_ represents. There is a GetSelf function that returns the buf_ and if what they get from GetSelf is different than what what data_ and size_ indicate that might cause bugs in future. It does not have to be like that and we could have had defined the contract of GetSelf not to have to be exactly what data() and size() indicate. But the existing implementation of remove_suffix does respect this consistency, so it is more sensible to follow the pattern.

About efficiency, I am not concerned about existing usages as there is none. And we dont expect anyone to use it either. One might say we can postpone efficiency/clarity tradeoff decision for future, and update both remove_prefix and remove_suffix implementations.

About safety, it follows what we already have in remove_suffix implementation. So I dont expect the users to be surprised about what happens to the buffer that they pass to PinnableSlice.

anand1976: Its a good point about GetSelf(). It is a problematic function and requires callers to know whether the slice is pinned or not. So we can follow the pattern of remove_suffix() and revisit this at some point.

PR #6331, opened 2020-01-24
Force a new manifest file if append to current one fails

anand1976:
Fix for issue #6316 

When an append/sync of the manifest file fails due to an IO error such
as NoSpace, we don't always put the DB in read-only mode. This is true
for flush and compactions, as well as foreground operatons such as column family
add/drop, CompactFiles etc. Subsequent changes to the DB will be
recorded in the same manifest file, which would have a corrupted record
in the middle due to the previous failure. On next DB::Open(), it will
fail to process the full manifest and data will be lost.

To fix this, we reset VersionSet::descriptor_log_ on append/sync
failure, which will force a new manifest file to be written on the next
append.

Test Plan:
Add new unit tests in error_handler_test.cc

### DIFF
--- a/db/error_handler.cc
+++ b/db/error_handler.cc
@@ -207,7 +207,6 @@ Status ErrorHandler::SetBGError(const Status& bg_err, BackgroundErrorReason reas
   bool auto_recovery = auto_recovery_;
   if (new_bg_err.severity() >= Status::Severity::kFatalError && auto_recovery) {
     auto_recovery = false;
-    ;
   }
 
   // Allow some error specific overrides
--- a/db/error_handler.cc
+++ b/db/error_handler.cc
@@ -166,12 +166,6 @@ Status ErrorHandler::SetBGError(const Status& bg_err, BackgroundErrorReason reas
     return Status::OK();
   }
 
-  // Check if recovery is currently in progress. If it is, we will save this
-  // error so we can check it at the end to see if recovery succeeded or not
-  if (recovery_in_prog_ && recovery_error_.ok()) {
-    recovery_error_ = bg_err;
-  }
-
   bool paranoid = db_options_.paranoid_checks;
   Status::Severity sev = Status::Severity::kFatalError;
   Status new_bg_err;
@@ -203,6 +197,12 @@ Status ErrorHandler::SetBGError(const Status& bg_err, BackgroundErrorReason reas
   }
 
   new_bg_err = Status(bg_err, sev);
+
+  // Check if recovery is currently in progress. If it is, we will save this
+  // error so we can check it at the end to see if recovery succeeded or not
+  if (recovery_in_prog_ && recovery_error_.ok()) {
+    recovery_error_ = new_bg_err;
+  }
 
   bool auto_recovery = auto_recovery_;
   if (new_bg_err.severity() >= Status::Severity::kFatalError && auto_recovery) {
--- a/db/error_handler_test.cc
+++ b/db/error_handler_test.cc
@@ -22,6 +22,21 @@ namespace rocksdb {
 class DBErrorHandlingTest : public DBTestBase {
  public:
   DBErrorHandlingTest() : DBTestBase("/db_error_handling_test") {}
+
+  std::string GetManifestNameFromLiveFiles() {
+    std::vector<std::string> live_files;
+    uint64_t manifest_size;
+
+    dbfull()->GetLiveFiles(live_files, &manifest_size, false);
+    for (auto& file : live_files) {
+      uint64_t num = 0;
+      FileType type;
+      if (ParseFileName(file, &num, &type) && type == kDescriptorFile) {
+        return file;
+      }
+    }
+    return "";
+  }
 };
 
 class DBErrorHandlingEnv : public EnvWrapper {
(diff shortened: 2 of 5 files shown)


db/error_handler_test.cc
```diff
@@ -161,6 +161,205 @@ TEST_F(DBErrorHandlingTest, FLushWriteError) {
...
+  for (auto& file : live_files) {
+    if (file.find("MANIFEST") != std::string::npos) {
+      new_manifest = file;
+      break;
+    }
+  }
+  ASSERT_NE(new_manifest, old_manifest);
+
+  Reopen(options);
+  ASSERT_EQ("val", Get(Key(0)));
```
zhichao-cao: Can we check the value of key(1) also?

db/error_handler.cc
```diff
@@ -204,6 +198,12 @@ Status ErrorHandler::SetBGError(const Status& bg_err, BackgroundErrorReason reas
 
   new_bg_err = Status(bg_err, sev);
 
+  // Check if recovery is currently in progress. If it is, we will save this
```
zhichao-cao: Can we move this block just before if (new_bg_err == Status::NoSpace()) ?

anand1976: We could, but is there a particular reason? The reason for moving this down here is so recovery_error_ will have the correct severity. This way, the caller of DB::Resume() will receive an error indicating whether they can retry resume later or not. We don't necessarily care how new_bg_err_ is handled from this point on.

db/error_handler_test.cc
```diff
@@ -161,6 +161,205 @@ TEST_F(DBErrorHandlingTest, FLushWriteError) {
...
+  std::vector<std::string> live_files;
+  std::string old_manifest;
+  std::string new_manifest;
+  uint64_t manifest_size;
+
+  listener->EnableAutoRecovery(false);
+  DestroyAndReopen(options);
+  ASSERT_OK(dbfull()->GetLiveFiles(live_files, &manifest_size, false));
+  for (auto& file : live_files) {
+    if (file.find("MANIFEST") != std::string::npos) {
```
riversand963: Can we use `ParseFileName()` here and in a few other places?

db/error_handler_test.cc
```diff
@@ -161,6 +161,205 @@ TEST_F(DBErrorHandlingTest, FLushWriteError) {
...
+      new_manifest = file;
+      break;
+    }
+  }
+  ASSERT_NE(new_manifest, old_manifest);
+  Reopen(options);
+  ASSERT_EQ("val", Get(Key(0)));
+  Destroy(options);
```
riversand963: I suggest that we let `DBErrorHandlingTest`, a subclass of `DBTestBase` observe `KEEP_DB` env variable. See `~DBTestBase()`. This allows us to optionally preserve the db directory for debugging purpose.

db/error_handler.cc
zhichao-cao: Thanks for the explanation. Can you remove the ";" in line 210 in this PR?

anand1976:
@riversand963 
> What if the process crashes after closing the previous, potentially corrupted MANIFEST, but before creating the new MANIFEST? Will the db still be able to recover?

Isn't it the same as basic crash recovery?

riversand963:
> @riversand963
> 
> > What if the process crashes after closing the previous, potentially corrupted MANIFEST, but before creating the new MANIFEST? Will the db still be able to recover?
> 
> Isn't it the same as basic crash recovery?

It is, and will cause later recovery to fail iiuc. Just trying to understand the scope of this fix.

anand1976:
> 
> It is, and will cause later recovery to fail iiuc. Just trying to understand the scope of this fix.

Actually, I think there is a difference, especially if the writes are batched. A crash while writing manifest will only result in the last record being partially written, and that will be skipped on open I think. The previous records in the batch will be valid. However, if the DB stays up (either read-only or read-write), the files referenced by the previous records will be deleted and recovery on re-open will fail. I think this is the issue you were trying to fix in https://github.com/facebook/rocksdb/pull/5379?

riversand963:
> > It is, and will cause later recovery to fail iiuc. Just trying to understand the scope of this fix.
> 
> Actually, I think there is a difference, especially if the writes are batched. A crash while writing manifest will only result in the last record being partially written, and that will be skipped on open I think. The previous records in the batch will be valid. However, if the DB stays up (either read-only or read-write), the files referenced by the previous records will be deleted and recovery on re-open will fail. I think this is the issue you were trying to fix in #5379?

Thanks for reminding me of this. I checked log::Reader, and you are right, the trailing incomplete version edit in the MANIFEST will not be reported as failure.

db/error_handler_test.cc
```diff
@@ -200,18 +207,14 @@ TEST_F(DBErrorHandlingTest, ManifestWriteError) {
...
-  }
+
+  new_manifest = GetManifestNameFromLiveFiles();
   ASSERT_NE(new_manifest, old_manifest);
 
   Reopen(options);
   ASSERT_EQ("val", Get(Key(0)));
-  Destroy(options);
+  ASSERT_EQ("val", Get(Key(1)));
+  Close();
```
riversand963: This `Close()` at the end of each test is not needed, but it's harmless.

anand1976:

anand1976:
The FB internal failures are due to an infra issue and unrelated to this PR. So going ahead with landing this PR.

PR #6334, opened 2020-01-27
Attempt to recover from db with missing table files

riversand963:
There are situations when RocksDB tries to recover, but the db is in an inconsistent state due to SST files referenced in the MANIFEST being missing. In this case, previous RocksDB will just fail the recovery and return a non-ok status.
This PR enables another possibility. During recovery, RocksDB checks possible MANIFEST files, and try to recover to the most recent state without missing table file. `VersionSet::Recover()` applies version edits incrementally and "materializes" a version only when this version does not reference any missing table file. After processing the entire MANIFEST, the version created last will be the latest version.
`DBImpl::Recover()` calls `VersionSet::Recover()`. Afterwards, WAL replay will *not* be performed.
To use this capability, set `options.best_efforts_recovery = true` when opening the db. Best-efforts recovery is currently incompatible with atomic flush.

Test plan (on devserver):
```
$make check
$COMPILE_WITH_ASAN=1 make all && make check
```

### DIFF
--- a/db/db_impl/db_impl_files.cc
+++ b/db/db_impl/db_impl_files.cc
@@ -14,6 +14,7 @@
 #include "db/event_helpers.h"
 #include "db/memtable_list.h"
 #include "file/file_util.h"
+#include "file/filename.h"
 #include "file/sst_file_manager_impl.h"
 #include "util/autovector.h"
 
@@ -662,6 +663,51 @@ uint64_t PrecomputeMinLogNumberToKeep(
     min_log_number_to_keep = min_log_refed_by_mem;
   }
   return min_log_number_to_keep;
+}
+
+Status DBImpl::CleanupFilesAfterRecovery() {
+  mutex_.AssertHeld();
+  std::set<std::string> paths;
+  for (const auto& db_path : immutable_db_options_.db_paths) {
+    paths.insert(db_path.path);
+  }
+  for (const auto* cfd : *versions_->GetColumnFamilySet()) {
+    for (const auto& cf_path : cfd->ioptions()->cf_paths) {
+      if (paths.find(cf_path.path) == paths.end()) {
+        paths.insert(cf_path.path);
+      }
+    }
+  }
+  uint64_t next_file_number = versions_->current_next_file_number();
+  uint64_t largest_file_number = next_file_number;
+  std::set<std::string> files_to_delete;
+  for (const auto& path : paths) {
+    std::vector<std::string> files;
+    env_->GetChildren(path, &files);
+    for (const auto& fname : files) {
+      uint64_t number = 0;
+      FileType type;
+      if (!ParseFileName(fname, &number, &type)) {
+        continue;
+      }
+      largest_file_number = std::max(largest_file_number, number);
+      if (type == kTableFile && number >= next_file_number &&
+          files_to_delete.find(fname) == files_to_delete.end()) {
+        files_to_delete.insert(NormalizePath(path + fname));
+      }
+    }
+  }
+  versions_->next_file_number_.store(largest_file_number + 1);
+  mutex_.Unlock();
+  Status s;
+  for (const auto& fname : files_to_delete) {
+    s = env_->DeleteFile(fname);
+    if (!s.ok()) {
+      break;
+    }
+  }
+  mutex_.Lock();
+  return s;
 }
 
 }  // namespace ROCKSDB_NAMESPACE
--- a/db/db_impl/db_impl_files.cc
+++ b/db/db_impl/db_impl_files.cc
@@ -667,17 +667,20 @@ uint64_t PrecomputeMinLogNumberToKeep(
 
 Status DBImpl::CleanupFilesAfterRecovery() {
   mutex_.AssertHeld();
-  std::set<std::string> paths;
+  std::vector<std::string> paths;
+  paths.push_back(dbname_);
   for (const auto& db_path : immutable_db_options_.db_paths) {
-    paths.insert(db_path.path);
+    paths.push_back(db_path.path);
   }
   for (const auto* cfd : *versions_->GetColumnFamilySet()) {
     for (const auto& cf_path : cfd->ioptions()->cf_paths) {
-      if (paths.find(cf_path.path) == paths.end()) {
-        paths.insert(cf_path.path);
-      }
+      paths.push_back(cf_path.path);
     }
   }
+  // Dedup paths
+  std::sort(paths.begin(), paths.end());
+  paths.erase(std::unique(paths.begin(), paths.end()), paths.end());
+
   uint64_t next_file_number = versions_->current_next_file_number();
   uint64_t largest_file_number = next_file_number;
   std::set<std::string> files_to_delete;
--- a/db/db_impl/db_impl_files.cc
+++ b/db/db_impl/db_impl_files.cc
@@ -693,14 +693,17 @@ Status DBImpl::CleanupFilesAfterRecovery() {
       if (!ParseFileName(fname, &number, &type)) {
         continue;
       }
+      const std::string normalized_fpath = NormalizePath(path + fname);
       largest_file_number = std::max(largest_file_number, number);
       if (type == kTableFile && number >= next_file_number &&
-          files_to_delete.find(fname) == files_to_delete.end()) {
-        files_to_delete.insert(NormalizePath(path + fname));
+          files_to_delete.find(normalized_fpath) == files_to_delete.end()) {
+        files_to_delete.insert(normalized_fpath);
       }
     }
   }
-  versions_->next_file_number_.store(largest_file_number + 1);
+  if (largest_file_number > next_file_number) {
+    versions_->next_file_number_.store(largest_file_number + 1);
+  }
   mutex_.Unlock();
   Status s;
   for (const auto& fname : files_to_delete) {
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -253,9 +253,9 @@ Status DBImpl::ValidateOptions(const DBOptions& db_options) {
   }
 
   // TODO remove this restriction
-  if (db_options.atomic_flush && db_options.incremental_recovery) {
+  if (db_options.atomic_flush && db_options.best_efforts_recovery) {
     return Status::InvalidArgument(
-        "atomic_flush is currently incompatible with incremental recovery");
+        "atomic_flush is currently incompatible with best-efforts recovery");
   }
 
   return Status::OK();
@@ -427,7 +427,7 @@ Status DBImpl::Recover(
   assert(db_id_.empty());
   Status s;
   bool missing_table_file = false;
-  if (!immutable_db_options_.incremental_recovery) {
+  if (!immutable_db_options_.best_efforts_recovery) {
     s = versions_->Recover(column_families, read_only, &db_id_);
   } else {
     s = versions_->TryRecover(column_families, read_only, &db_id_,
@@ -515,7 +515,7 @@ Status DBImpl::Recover(
     // attention to it in case we are recovering a database
     // produced by an older version of rocksdb.
     std::vector<std::string> filenames;
-    if (!immutable_db_options_.incremental_recovery) {
+    if (!immutable_db_options_.best_efforts_recovery) {
       s = env_->GetChildren(immutable_db_options_.wal_dir, &filenames);
     }
     if (s.IsNotFound()) {
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -419,7 +419,15 @@ Status DBImpl::Recover(
     }
   }
   assert(db_id_.empty());
-  Status s = versions_->Recover(column_families, read_only, &db_id_);
+  Status s;
+  if (!immutable_db_options_.incremental_recovery) {
+    s = versions_->Recover(column_families, read_only, &db_id_);
+  } else {
+    s = versions_->TryRecover(column_families, read_only, &db_id_);
+    if (s.ok()) {
+      s = CleanupFilesAfterRecovery();
+    }
+  }
   if (!s.ok()) {
     return s;
   }
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -252,6 +252,12 @@ Status DBImpl::ValidateOptions(const DBOptions& db_options) {
         "atomic_flush is incompatible with enable_pipelined_write");
   }
 
+  // TODO remove this restriction
+  if (db_options.atomic_flush && db_options.incremental_recovery) {
+    return Status::InvalidArgument(
+        "atomic_flush is currently incompatible with incremental recovery");
+  }
+
   return Status::OK();
 }
 
@@ -509,7 +515,9 @@ Status DBImpl::Recover(
     // attention to it in case we are recovering a database
     // produced by an older version of rocksdb.
     std::vector<std::string> filenames;
-    s = env_->GetChildren(immutable_db_options_.wal_dir, &filenames);
+    if (!immutable_db_options_.incremental_recovery) {
+      s = env_->GetChildren(immutable_db_options_.wal_dir, &filenames);
+    }
     if (s.IsNotFound()) {
       return Status::InvalidArgument("wal_dir not found",
                                      immutable_db_options_.wal_dir);
@@ -554,7 +562,7 @@ Status DBImpl::Recover(
       }
     }
 
-    if (!logs.empty() && !missing_table_file) {
+    if (!logs.empty()) {
       // Recover in the order in which the logs were generated
       std::sort(logs.begin(), logs.end());
       bool corrupted_log_found = false;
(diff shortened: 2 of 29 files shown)


db/version_set.cc
```diff
@@ -4553,6 +4592,410 @@ Status VersionSet::Recover(
...
+        auto fiter = missing_files.find(file_num);
+        if (fiter != missing_files.end()) {
+          missing_files.erase(fiter);
+        }
+      }
+      for (const auto& elem : edit.GetNewFiles()) {
+        const FileMetaData& meta = elem.second;
+        const FileDescriptor& fd = meta.fd;
+        uint64_t file_num = fd.GetNumber();
+        const std::string fname = MakeTableFileName(dbname_, file_num);
```
riversand963: Should consider path id when constructing table file name.

db/db_impl/db_impl_files.cc
```diff
@@ -664,4 +665,49 @@ uint64_t PrecomputeMinLogNumberToKeep(
   return min_log_number_to_keep;
 }
 
+Status DBImpl::CleanupFilesAfterRecovery() {
```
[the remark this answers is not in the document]
riversand963: Done.

db/db_impl/db_impl_files.cc
```diff
@@ -664,4 +665,49 @@ uint64_t PrecomputeMinLogNumberToKeep(
...
+  mutex_.AssertHeld();
+  std::set<std::string> paths;
+  for (const auto& db_path : immutable_db_options_.db_paths) {
+    paths.insert(db_path.path);
+  }
```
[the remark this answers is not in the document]
riversand963: Good catch. Checked `SanitizeOptions()`, and it should be added here. I will also need to dedup.

db/db_impl/db_impl_open.cc
```diff
@@ -543,7 +553,7 @@ Status DBImpl::Recover(
       }
     }
 
-    if (!logs.empty()) {
+    if (!logs.empty() && !missing_table_file) {
```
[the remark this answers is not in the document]
riversand963: No. It means whether the last version contains missing file, or put it another way, whether the db recovers *all* version edits in the MANIFEST. If not, there is a gap between MANIFEST and WAL, and we should not replay WAL. If missing_files = false, then we can replay WAL.

db/version_set.cc
```diff
@@ -5536,6 +5997,21 @@ uint64_t VersionSet::GetTotalSstFilesSize(Version* dummy_versions) {
   return total_files_size;
 }
 
+Status VersionSet::VerifyFileMetadata(const std::string& fpath,
+                                      const FileMetaData& meta) const {
+  Status status = env_->FileExists(fpath);
+  uint64_t fsize = 0;
+  if (status.ok()) {
+    status = fs_->GetFileSize(fpath, IOOptions(), &fsize, nullptr);
```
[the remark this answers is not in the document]
riversand963: I think it is an IOError with PathNotFound subcode. See https://github.com/facebook/rocksdb/blob/master/env/io_posix.h#L54:L56.
The `GetFileSize` does make `FileExists` unnecessary.

db/version_set.cc
```diff
@@ -5536,6 +5997,21 @@ uint64_t VersionSet::GetTotalSstFilesSize(Version* dummy_versions) {
   return total_files_size;
 }
 
+Status VersionSet::VerifyFileMetadata(const std::string& fpath,
+                                      const FileMetaData& meta) const {
+  Status status = env_->FileExists(fpath);
```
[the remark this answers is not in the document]
riversand963: This can be resolved with the next comment.

db/version_set.cc
```diff
@@ -4553,6 +4592,422 @@ Status VersionSet::Recover(
...
+      }
+    }
+  }
+  return s;
+}
+
+// TODO Make this function shorter.
+// Also note that LoadTableHandlers should be called after we determine the
+// final version, thus may likely be outside this function.
+Status VersionSet::TryRecoverFromOneManifest(
```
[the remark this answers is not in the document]
riversand963: I agree. Let me try again. The logic of handling each version edit is pretty different from what we do in regular recovery though. I expect a refactoring to combine `TryRecoverFromOneManifest` and `Recover` as a follow up. At this point, not sure if it is a good idea to modify `VersionSet::Recover` unnecessarily. 
but let me try again.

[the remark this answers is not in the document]
riversand963: @anand1976, after checking the code, I believe we should have a `VersionEditHandler` class and subclasses to make the code simpler. Should we do it here or in another PR?

riversand963:
Another way to test is to temporarily set the default value of options.incremental_recovery to true and make check.

db/version_set.cc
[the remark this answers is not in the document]
riversand963: @anand1976, I tried to address this comment to shorten the function. I can see a way of doing it, as in a temp branch: https://github.com/facebook/rocksdb/compare/master...riversand963:version-edit-handler.
If we pursue this path, I will add VersionEditHandler class in this PR, and use it only for `TryRecoverFromOneManifest`, but not in `Recover` and `ReadAndApply`. Afterwards, I will use this class and its subclasses to refactor `Recover` and `ReadAndApply`.
Could you let me know what you think? Thanks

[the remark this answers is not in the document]
riversand963: Done with refactoring that can be found in recent commits. Ready for another round of review.

db/version_edit_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+          "MANIFEST record referencing unknown column family");
+    }
+    ColumnFamilyData* tmp_cfd = nullptr;
+    if (s.ok()) {
+      auto builder_iter = builders_.find(edit.column_family_);
+      assert(builder_iter != builders_.end());
+      tmp_cfd = version_set_->GetColumnFamilySet()->GetColumnFamily(
+          edit.column_family_);
+      assert(tmp_cfd != nullptr);
+      s = CreateVersion(edit, tmp_cfd, false);
```
anand1976: Why create a new Version each time? Why not just apply to current?

db/version_edit_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+  // in builders means that user supplied that column family
+  // option AND that we encountered column family add record
+  bool in_builders = builders_.find(edit.column_family_) != builders_.end();
+  // They cannot both be true
+  assert(!(in_not_found && in_builders));
+  *cf_in_not_found = in_not_found;
+  *cf_in_builders = in_builders;
+}
+
+void VersionEditHandler::CheckIterateResult(const log::Reader& reader,
```
anand1976: The name CheckIterateResult was a little confusing to me, as I thought it implied another round of iteration. Maybe CheckIterationResult would be clearer.

db/version_edit_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+    uint64_t file_num = fd.GetNumber();
+    const std::string fpath =
+        MakeTableFileName(cfd->ioptions()->cf_paths[0].path, file_num);
+    s = version_set_->VerifyFileMetadata(fpath, meta);
+    if (s.IsPathNotFound() || s.IsNotFound()) {
+      missing_files.insert(file_num);
+      s = Status::OK();
+    }
+  }
+  bool missing_info = !version_edit_params_.has_log_number_ ||
```
anand1976: Wouldn't ```missing_info_ == true``` indicate a corrupt manifest entry? Maybe we should log a message atleast since its not expected.

db/version_edit_handler.h
```diff
@@ -0,0 +1,96 @@
...
+#include "db/version_edit.h"
+#include "db/version_set.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+typedef std::unique_ptr<BaseReferencedVersionBuilder> VersionBuilderUPtr;
+
+// Not thread-safe, external synchronization is necessary if an object of
+// VersionEditHandler is shared by multiple threads.
+class VersionEditHandler {
```
anand1976: Some comments here about what the class does would be good, especially how to call it and how it interacts with VersionSet.

db/version_edit_handler.h
```diff
@@ -0,0 +1,96 @@
...
+      cf_to_missing_files_;
+  bool ignore_missing_files_;
+
+ private:
+  Status ExtractInfoFromVersionEdit(ColumnFamilyData* cfd,
+                                    const VersionEdit& edit);
+  bool initialized_;
+};
+
+class VersionEditHandlerPointInTime : public VersionEditHandler {
```
anand1976: Comments here too

db/version_edit_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+VersionEditHandlerPointInTime::~VersionEditHandlerPointInTime() {
+  for (const auto& elem : versions_) {
+    delete elem.second;
+  }
+  versions_.clear();
+}
+
+void VersionEditHandlerPointInTime::CheckIterateResult(
+    const log::Reader& reader, Status* s) {
+  VersionEditHandler::CheckIterateResult(reader, s);
```
anand1976: Should this be called here or after appending the versions to ```version_set_```?

db/version_edit_handler.cc
riversand963: No it does not indicate one corrupted MANIFEST entry.
If missing_info is true after processing the entire MANIFEST, then the MANIFEST file is incomplete, and we cannot recover.

db/version_edit_handler.cc
riversand963: Which function do you mean?
`VersionEditHandler::CheckIterateResult` should be called before appending the versions to version_set_ because it is necessary to perform some checks to see if the recovery is successful.

anand1976: Ok. I got confused because ```VersionEditHandler::CreateVersion``` creates and appends a new version to ```version_set_```, so it would already have a set of valid versions when ```VersionEditHandler::CheckIterateResult``` is called.

db/version_edit_handler.cc
riversand963: We currently rely on `VersionBuilder::SaveTo(version->strorage_info())` to create versions. 
The version builder keeps a history of all added and delete files. Calling `version_builder->SaveTo(version->storage_info())` repeatedly for the same version will cause the same files be added.
This is also our current way of building versions
1. Create a version builder
2. version builder applies multiple version edits
3. Finally, create a version, call version_builder->SaveTo(version)
4. Append version
I am following this practice.

anand1976:
@riversand963 Thanks for the explanation about the reasoning behind handling CF add/drop. With this recovery, we don't support cross CF consistency. We may have to revisit later for cases when WAL is used or atomic flush is enabled. For now, should we just not try to recover if either of those is true?

riversand963:
> @riversand963 Thanks for the explanation about the reasoning behind handling CF add/drop. With this recovery, we don't support cross CF consistency. We may have to revisit later for cases when WAL is used or atomic flush is enabled. For now, should we just not try to recover if either of those is true?

Agree.
In fact, I realize that `TryRecover` from an inconsistent MANIFEST can result in `DB::Open` failure if WAL is enabled and replayed. So let's disable it for now.

db/version_edit_handler.h
riversand963: Done

db/version_edit_handler.h
riversand963: Done

db/version_edit_handler.cc
riversand963: Done

riversand963:
@anand1976 , could you PTAL? Thanks

db/db_impl/db_impl_open.cc
```diff
@@ -419,7 +425,17 @@ Status DBImpl::Recover(
     }
   }
   assert(db_id_.empty());
-  Status s = versions_->Recover(column_families, read_only, &db_id_);
+  Status s;
+  bool missing_table_file = false;
```
anand1976: Where is ```missing_table_file``` used?

db/db_impl/db_impl_open.cc
```diff
@@ -499,7 +515,9 @@ Status DBImpl::Recover(
     // attention to it in case we are recovering a database
     // produced by an older version of rocksdb.
     std::vector<std::string> filenames;
-    s = env_->GetChildren(immutable_db_options_.wal_dir, &filenames);
+    if (!immutable_db_options_.best_efforts_recovery) {
```
anand1976: Is just ignoring WAL files when best_efforts_recovery is enabled correct? Or should the recovery fail if there are non-empty WAL files?

db/version_builder.h
```diff
@@ -44,5 +46,21 @@ class VersionBuilder {
...
+ public:
+  explicit BaseReferencedVersionBuilder(ColumnFamilyData* cfd);
+  BaseReferencedVersionBuilder(ColumnFamilyData* cfd, Version* v);
+  ~BaseReferencedVersionBuilder();
+  VersionBuilder* version_builder() const { return version_builder_.get(); }
+
+ private:
+  std::unique_ptr<VersionBuilder> version_builder_;
+  Version* version_;
+  const bool should_ref_cfd_version_;
```
anand1976: I suggest adding a comment to clarify what this is used for.

db/version_edit_handler.cc
```diff
@@ -0,0 +1,591 @@
...
+    if (!s.ok()) {
+      break;
+    }
+    if (edit.has_db_id_) {
+      version_set_->db_id_ = edit.GetDbId();
+      if (db_id != nullptr) {
+        *db_id = version_set_->db_id_;
+      }
+    }
+    s = read_buffer_.AddEdit(&edit);
```
anand1976: We add every version edit to ```read_buffer_```, but only clear it if its an atomic group and the buffer is replayed?

db/version_set_test.cc
```diff
@@ -1279,6 +1331,704 @@ INSTANTIATE_TEST_CASE_P(
...
+    ColumnFamilyData* cfd = versions_->GetColumnFamilySet()->GetColumnFamily(
+        kUnknownColumnFamilyName);
+    ASSERT_EQ(nullptr, cfd);
+  }
+}
+
+INSTANTIATE_TEST_CASE_P(
+    BestEffortRecovery, VersionSetTestEmptyDb,
+    testing::Combine(
+        testing::Bool(), testing::Bool(),
```
anand1976: Can you add comments on what each parameter is for?

include/rocksdb/options.h
```diff
@@ -1130,6 +1130,16 @@ struct DBOptions {
   //
   // Default: nullptr
   std::shared_ptr<FileChecksumFunc> sst_file_checksum_func = nullptr;
+
+  // By default, RocksDB recovery fails if any table file referenced in
+  // MANIFEST are missing after scanning the MANIFEST.
+  // Best-efforts recovery is another recovery mode that
+  // tries to restore the database to the most recent point in time without
+  // missing file.
+  // Currently not compatible with atomic flush. Furthermoare, WAL files will
```
anand1976: Typo - Furthermoare

riversand963: Will fix

db/db_impl/db_impl_open.cc
riversand963: It's used in testing.

db/version_builder.h
riversand963: I am actually rethinking this. Maybe I can try to remove this variable. If it's non-trivial, will add a comment.

db/db_impl/db_impl_open.cc
riversand963: On the one hand, WAL contains data ahead of SSTs (not considering WAL with prepared entries). On the other hand, it is very common to have non-empty WAL even if db closed cleanly.
I think we should not report best-efforts recovery as failure.

db/version_set_test.cc
riversand963: Sure.

db/db_impl/db_impl_open.cc
riversand963: Furthermore, I think we need to bump up cfd->log_number and min_log_number_to_keep after we ignore the existing WALs.

db/version_edit_handler.cc
riversand963: As https://github.com/facebook/rocksdb/blob/master/db/version_set.cc#L3513:L3518 shows, AtomicGroupReadBuf does not actually buffer a non-atomic edit.

db/version_builder.h
riversand963: Removed, thus no need for comment any more.

db/db_impl/db_impl_open.cc
anand1976: I was thinking about cross-CF consistency. Do we implicitly guarantee it if the user uses WAL? We don't guarantee it with the best efforts recovery right? Maybe we can just mention it in READ.md.

riversand963: Good point. User may very likely assume that cross-CF consistency is guaranteed if they enable WAL. We should call this out in HISTORY.md.

Issue #6338, opened 2020-01-28
BlobDB crashes in sequentially fill test


burtonli:
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://www.facebook.com/groups/rocksdb.dev

### Expected behavior
Test finish without crash

### Actual behavior
DB crashes at 
    BlobFile::LinkSstFile  
    BlobDBImpl::LinkSstToBlobFileImpl
    BlobDBImpl::ProcessCompactionJobInfo

### Steps to reproduce the behavior
Run db bench with:
--benchmarks=fillseq --use_blob_db=1 --blob_db_enable_gc=1 --blob_db_gc_cutoff=0.25 --write_buffer_size=2097152 --target_file_size_base=2097152 --max_bytes_for_level_base=2097152 --num_levels=3  --compaction_style=0 --mmap_read=0 --statistics=1 --histogram=1 --num=2400000 --threads=1 --value_size=32768 --block_size=4096 --cache_size=1048576 --bloom_bits=10 --cache_numshardbits=4 --open_files=-1 --verify_checksum=1 --db=F:\RocksDBPerfTest --sync=0 --disable_wal=1 --compression_type=none --stats_interval=1000000 --compression_ratio=0.25 --max_write_buffer_number=3  --level0_file_num_compaction_trigger=4 --level0_slowdown_writes_trigger=48 --level0_stop_writes_trigger=64 --stats_per_interval=1 --use_existing_db=0 --use_direct_reads=1 --use_direct_io_for_flush_and_compaction=1 --compaction_readahead_size=262144 --writable_file_max_buffer_size=524288 --new_table_reader_for_compaction_inputs=1

Root cause:
@ltamasi @siying 

There are two code paths to maintain blob->sst mappings:
1. Flush
BackgroundCallFlush->BackgroundFlush->FlushMemTablesToOutputFiles->FlushMemTableToOutputFile->NotifyOnFlushCompleted->ProcessFlushJobInfo->LinkSstToBlobFile->LinkSstToBlobFileImpl

2. Compaction
BackgroundCallFlush->BackgroundFlush->FlushMemTablesToOutputFiles->FlushMemTableToOutputFile->SchedulePendingCompaction and MaybeScheduleFlushOrCompaction ->BlobDBListenerGC::OnCompactionCompleted->
ProcessCompactionJobInfo

Since these paths are running in different threads, and not synchronized, so it's possible 2nd path
[...]

PR #6341, opened 2020-01-29
BlobDB: handle the case when OnCompactionCompleted precedes OnFlushCompleted

ltamasi:
Summary:
BlobDB keeps track of the mapping between SSTs and blob files using
the `OnFlushCompleted` and `OnCompactionCompleted` callbacks of
the `Listener` interface: upon receiving a flush notification, a link is added
between the newly flushed SST and the corresponding blob file; for
compactions, links are removed for the inputs and added for the outputs.
In certain cases, it is possible for the compaction notification that results
in a link being removed to precede the flush notification that establishes
the link (see #6338).
The earlier code did not handle this case, resulting in assertion failures.
With the patch, BlobDB stores unmatched link removals in a separate set,
and reconciles them with link additions as they arrive.

Test Plan:
Added a unit test that simulates this case.

### DIFF
--- a/utilities/blob_db/blob_db_impl.cc
+++ b/utilities/blob_db/blob_db_impl.cc
@@ -519,7 +519,8 @@ bool BlobDBImpl::MarkBlobFileObsoleteIfNeeded(
   // blob file obsolete if there are still unflushed memtables from before
   // the time the blob file was closed.
   if (blob_file->GetImmutableSequence() > flush_sequence_ ||
-      !blob_file->GetLinkedSstFiles().empty()) {
+      !blob_file->GetLinkedSstFiles().empty() ||
+      !blob_file->GetUnlinkedSstFiles().empty()) {
     return false;
   }
 
--- a/utilities/blob_db/blob_db_test.cc
+++ b/utilities/blob_db/blob_db_test.cc
@@ -1857,6 +1857,48 @@ TEST_F(BlobDBTest, MaintainBlobFileToSstMapping) {
   }
 }
 
+TEST_F(BlobDBTest, MaintainBlobFileToSstMappingRace) {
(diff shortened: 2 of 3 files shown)


burtonli:
I think handling OnCompactionCompleted precedes OnFlushCompleted inside blob db_impl may create issue for MarkUnreferencedBlobFilesObsolete(). If OnCompactionCompleted arrives first and the blob file happens to be the oldest file (e.g.: the sst file is the only file in the db, and it is triggered compaction by reaching the level base size), then the blob file may be marked as obsolete in BlobDBImpl::ProcessCompactionJobInfo() before OnFlushCompleted call.

I have a proposal to fix the issue by always make sure OnFlushCompleted is always called ahead of OnCompactionBegin and OnCompactionCompleted for the same file, which seems to be a more nature way for other external consumers to built their similar logic like BlobDB. Think about how does an external service to handle a file starts compaction before flushed.
[https://github.com/facebook/rocksdb/pull/6342](url)

ltamasi:
Considering that this is not a threading issue as originally thought (since the notification for the trivial move compaction deterministically precedes the notification for the flush), it would be better to fix this in RocksDB itself. Closing in favor of #6342 .

ltamasi:
> I think handling OnCompactionCompleted precedes OnFlushCompleted inside blob db_impl may create issue for MarkUnreferencedBlobFilesObsolete(). If OnCompactionCompleted arrives first and the blob file happens to be the oldest file (e.g.: the sst file is the only file in the db, and it is triggered compaction by reaching the level base size), then the blob file may be marked as obsolete in BlobDBImpl::ProcessCompactionJobInfo() before OnFlushCompleted call.
> 

We actually have protection in place against that but I do agree that this should ideally be fixed in the RocksDB core.

PR #6349, opened 2020-01-30
Fail recovery on loading missing file

riversand963:
VersionSet::Recovery() should fail if trying to load missing or corrupted SST files. It is better to fail early than late. To achieve this, we need to check the return value of `LoadTableHandlers()` and propagate the failure if necessary.

Test plan (devserver):
```
$make check
```

### DIFF
--- a/db/cuckoo_table_db_test.cc
+++ b/db/cuckoo_table_db_test.cc
@@ -282,7 +282,7 @@ TEST_F(CuckooTableDBTest, SameKeyInsertedInTwoDifferentFilesAndCompacted) {
   }
 }
 
-TEST_F(CuckooTableDBTest, AdaptiveTable) {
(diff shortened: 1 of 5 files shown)


db/cuckoo_table_db_test.cc
```diff
@@ -282,7 +282,7 @@ TEST_F(CuckooTableDBTest, SameKeyInsertedInTwoDifferentFilesAndCompacted) {
   }
 }
 
-TEST_F(CuckooTableDBTest, AdaptiveTable) {
+TEST_F(CuckooTableDBTest, DISABLED_AdaptiveTable) {
```
riversand963: Not sure whether we should keep supporting using DB with one type of table factory to open a previous db with sst files of another type.

riversand963:
Closing for now.

PR #6352, opened 2020-01-30
Shorten certain test names to avoid infra failure

riversand963:
Unit test names, together with other components,  are used to create log files
during some internal testing. Overly long names cause infra failure due to file
names being too long.

Look for internal tests.


riversand963:
Infra tests that used to fail now pass.

riversand963:
AppVeyor failure is irrelevant.

Issue #4803, opened 2018-12-20
DB::GetDbIdentity has wrong name or description


adamretter:
From `db.h`:

```C++
  // Sets the globally unique ID created at database creation time by invoking
  // Env::GenerateUniqueId(), in identity. Returns Status::OK if identity could
  // be set properly
  virtual Status GetDbIdentity(std::string& identity) const = 0;
```

I am not sure if the function should be renamed `SetDbIdentity` or if the description is wrong?

riversand963:
I think the description is inaccurate. The function reads the IDENTITY file and stores the id in the variable `identity`.

atulagrwl:
PR Created - https://github.com/facebook/rocksdb/pull/6354

atulagrwl:
Request has been merged. The issue can be closed now.

PR #6354, opened 2020-01-31
Fixing the documentation of the function (#4803)

atulagrwl:


include/rocksdb/db.h
```diff
@@ -1410,9 +1410,8 @@ class DB {
 
 #endif  // ROCKSDB_LITE
 
-  // Sets the globally unique ID created at database creation time by invoking
-  // Env::GenerateUniqueId(), in identity. Returns Status::OK if identity could
-  // be set properly
+  // Reads the IDENTITY file and stores the id in the identity variable.
```
riversand963: I believe calling `GetIdentify(std::string&)` will not read from IDENTITY file. The db id is recovered during db recovery and stored in a variable `db_id_`.

atulagrwl: Checked the code and you are correct. It returns db_id_ which is set during opening. Should I change the comment of documentation reflecting the same?
It could says
"Returns the ID which is read from IDENTITY file during the opening of database by setting in the identity variable"

riversand963: Mostly LGTM. Maybe worthwhile to say "unique ID" as well.

atulagrwl:
Updated as per comments.

riversand963:
The failures are unrelated.

riversand963:
Sorry I forgot to merge yesterday, and now this PR is based on an old master, causing merging to fail.
Could you please rebase? @atulagrwl

atulagrwl:
Merged master into issue branch.

PR #6363, opened 2020-02-03
Avoid lots of calls to Env::GetFileSize() in SstFileManagerImpl when opening DB

al13n321:
Before this PR it calls GetFileSize() once for each sst file in the DB. This can take a long time if there are be tens of thousands of sst files (e.g. in thousands of column families), and even longer if Env is talking to some remote service rather than local filesystem. This PR makes DB::Open() use sst file sizes that are already known from manifest (typically almost all files in the DB) and only call GetFileSize() for non-sst or obsolete files. Note that GetFileSize() is also called and checked against manifest in CheckConsistency(), so the calls in SstFileManagerImpl were completely redundant.

Test plan: deployed to a test cluster, looked at a dump of Env calls (from a custom instrumented Env) - no more thousands of GetFileSize()s.

### DIFF
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -1545,6 +1545,27 @@ Status DBImpl::Open(const DBOptions& db_options, const std::string& dbname,
   if (s.ok() && sfm) {
     // Notify SstFileManager about all sst files that already exist in
     // db_paths[0] and cf_paths[0] when the DB is opened.
+
+    // SstFileManagerImpl needs to know sizes of the files. For files whose size
+    // we already know (sst files that appear in manifest - typically that's the
+    // vast majority of all files), we'll pass the size to SstFileManager.
+    // For all other files SstFileManager will query the size from filesystem.
+
+    std::vector<LiveFileMetaData> metadata;
+
+    impl->mutex_.Lock();
+    impl->versions_->GetLiveFilesMetaData(&metadata);
+    impl->mutex_.Unlock();
+
+    std::unordered_map<std::string, uint64_t> known_file_sizes;
+    for (const auto& md : metadata) {
+      std::string name = md.name;
+      if (!name.empty() && name[0] == '/') {
+        name = name.substr(1);
+      }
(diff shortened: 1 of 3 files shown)


db/db_impl/db_impl_open.cc
```diff
@@ -1564,11 +1585,26 @@ Status DBImpl::Open(const DBOptions& db_options, const std::string& dbname,
...
+            known_file_sizes.erase(file_name);
+          } else {
+            sfm->OnAddFile(file_path);
+          }
         }
       }
     }
 
+    for (const auto& kv : known_file_sizes) {
+      ROCKS_LOG_WARN(impl->immutable_db_options_.info_log,
```
anand1976: This might be a misleading message. Since the calls to sst_file_manager are happening outside the DB mutex, a compaction happening in parallel could have deleted files after we retrieved the list of live files from the current version. In any case, missing files would have been caught by the consistency check.

file/sst_file_manager_impl.h
```diff
@@ -36,7 +36,11 @@ class SstFileManagerImpl : public SstFileManager {
   ~SstFileManagerImpl();
 
   // DB will call OnAddFile whenever a new sst file is added.
-  Status OnAddFile(const std::string& file_path, bool compaction = false);
+  // Parameter `file_size` is for optimization: if the caller knows the size of
+  // the file, they can provide it, otherwise OnAddFile() will query it from
+  // the filesystem.
+  Status OnAddFile(const std::string& file_path, bool compaction = false,
```
anand1976: I prefer to just overload OnAddFile() rather than add another default parameter.

db/db_impl/db_impl_open.cc
al13n321: This is mostly a sanity check that this code handles file names correctly. E.g. if I missed that I need to strip the leading '/' from `md.name` a few lines above, that could go unnoticed for a long time without this warning.

al13n321:
Removed the log message and added the overload.

PR #6364, opened 2020-02-03
Prevent a case of WriteBufferManager flush thrashing

ajkr:
Previously, the flushes triggered by `WriteBufferManager` could affect
the same CF repeatedly if it happens to get consecutive writes. Such
flushes are not particularly useful for reducing memory usage since
they switch nearly-empty memtables to immutable while they've just begun
filling their first arena block. In fact they may not even reduce the 
mutable memory count if they involve replacing one mutable memtable containing
one arena block with a new mutable memtable containing one arena block.
Further, if such switches happen even a few times before a flush finishes,
the immutable memtable limit will be reached and writes will stall.
    
This PR adds a heuristic to not switch memtables to immutable for CFs 
that already have one or more immutable memtables awaiting flush. There
is a memory usage regression if the user continues writing to the same
CF, that DB does not have any CFs eligible for switching, flushes
are not finishing, and the `WriteBufferManager` was constructed with
`allow_stall=false`. Before, it would grow by switching nearly empty
memtables until writes stall. Now, it would grow by filling memtables
until writes stall. This feels like an acceptable behavior change because
users who prefer to stall over violate the memory limit should be using
`allow_stall=true`, which is unaffected by this PR.
    
Test Plan:
    
- Command:
    
`rm -rf /dev/shm/dbbench/ && TEST_TMPDIR=/dev/shm ./db_bench -benchmarks=fillrandom -num_multi_db=8 -num_column_families=2 -write_buffer_size=4194304 -db_write_buffer_size=16777216 -compression_type=none -statistics=true -target_file_size_base=4194304 -max_bytes_for_level_base=16777216`
    
- `rocksdb.db.write.stall` count before this PR: 175 
- `rocksdb.db.write.stall` count after this PR: 0

### DIFF
--- a/db/db_impl/db_impl_write.cc
+++ b/db/db_impl/db_impl_write.cc
@@ -1648,12 +1648,6 @@ Status DBImpl::HandleWriteBufferManagerFlush(WriteContext* write_context) {
   // thread is writing to another DB with the same write buffer, they may also
   // be flushed. We may end up with flushing much more DBs than needed. It's
   // suboptimal but still correct.
-  ROCKS_LOG_INFO(
-      immutable_db_options_.info_log,
-      "Flushing column family with oldest memtable entry. Write buffers are "
-      "using %" ROCKSDB_PRIszt " bytes out of a total of %" ROCKSDB_PRIszt ".",
-      write_buffer_manager_->memory_usage(),
-      write_buffer_manager_->buffer_size());
   // no need to refcount because drop is happening in write thread, so can't
   // happen while we're in the write thread
   autovector<ColumnFamilyData*> cfds;
@@ -1667,9 +1661,11 @@ Status DBImpl::HandleWriteBufferManagerFlush(WriteContext* write_context) {
       if (cfd->IsDropped()) {
         continue;
       }
-      if (!cfd->mem()->IsEmpty()) {
-        // We only consider active mem table, hoping immutable memtable is
-        // already in the process of flushing.
+      if (!cfd->mem()->IsEmpty() && cfd->imm()->NumNotFlushed() == 0) {
+        // We only consider flush on CFs with bytes in the mutable memtable,
+        // and no immutable memtables for which flush has yet to finish. If
+        // we triggered flush on CFs already pending flush, we would risk
+        // creating too many immutable memtables leading to write stalls.
         uint64_t seq = cfd->mem()->GetCreationSeq();
         if (cfd_picked == nullptr || seq < seq_num_for_cf_picked) {
           cfd_picked = cfd;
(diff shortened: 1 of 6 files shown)


db/db_impl/db_impl_write.cc
```diff
@@ -1309,9 +1303,15 @@ Status DBImpl::HandleWriteBufferFull(WriteContext* write_context) {
       if (cfd->IsDropped()) {
         continue;
       }
-      if (!cfd->mem()->IsEmpty()) {
+      if (!cfd->mem()->IsEmpty() &&
+          cfd->mem()->ApproximateMemoryUsageFast() >
+              cfd->imm()->ApproximateUnflushedMemTablesMemoryUsage()) {
```
ajkr: after experimenting with it a bit today I realized this heuristic is confusing. Instead of this, how about only let write buffer manager trigger flush when `cfd->imm()->ApproximateUnflushedMemTablesMemoryUsage() == 0`?

ajkr: Done.

ajkr: This still isn't my favorite heuristic though. IMO it'd be good if (1) we permitted more cases of `WriteBufferManager` triggering flushes when there are already immutable memtables awaiting flush, and (2) we didn't risk iterating the list of CFs in every write for many consecutive writes. To achieve both we could maintain a flag in `ColumnFamilyData` indicating whether the next switch would trigger write stall, and another flag in `DBImpl` indicating whether the previously-mentioned flag is true in every CF. We would only call `HandleWriteBufferFull()` when the `DBImpl` flag is false. Within `HandleWriteBufferFull()`, we would only trigger flush on a CF where the `ColumnFamilyData` flag is false.

If that's desirable or there's any other suggested change, I am happy to try it.

ajkr: ping @siying

ajkr:
@cheng-chang I talked to Siying and he prefers `WriteBufferManager` to be strict, while this PR would make it less strict. so I am not sure this change is desirable, although we didn't discuss this PR specifically.

cheng-chang:
@ajkr I thought it should be imported, in that case, I'll discard the diff in Phabricator.

db/db_impl/db_impl_write.cc
ajkr: > (1) we permitted more cases of WriteBufferManager triggering flushes when there are already immutable memtables awaiting flush

This doesn't seem like a good idea to me anymore. Even if it won't lead to write stall, there's no need to trigger more flushes on the same CF.

jay-zhuang (approved):
LGTM. The change does make sense to me, plus the user is able to explicitly stall the write with `allow_stall=true` in WBM.

It **is** a behavior change, should update the doc like:
https://github.com/facebook/rocksdb/blob/b283f041f58deb87a7a0fa21005c963d0f0dee88/include/rocksdb/options.h#L906-L908

PR #6368, opened 2020-02-04
Fail fast in paranoid mode when LoadTableHandlers fail during recovering

cheng-chang:
Summary:
Previously, when recovering version set, LoadTableHandlers failures are ignored.
If paranoid_checks is true, this failure should not be ignored, otherwise, the opened db might be in an inconsistent state.

Test Plan:
make check


db/version_set.cc
```diff
@@ -3788,11 +3788,21 @@ Status VersionSet::ProcessManifestWrites(
...
                builder_guards.size() == versions.size());
         ColumnFamilyData* cfd = versions[i]->cfd_;
-        builder_guards[i]->version_builder()->LoadTableHandlers(
+        s = builder_guards[i]->version_builder()->LoadTableHandlers(
             cfd->internal_stats(), cfd->ioptions()->optimize_filters_for_hits,
             true /* prefetch_index_and_filter_in_cache */,
             false /* is_initial_load */,
             mutable_cf_options_ptrs[i]->prefix_extractor.get());
+        if (!s.ok()) {
+          if (db_options_->paranoid_checks) {
```
riversand963: Just break on !s.ok() can simplify the code.

db/version_set.cc
```diff
@@ -4506,11 +4516,17 @@ Status VersionSet::Recover(
...
       // unlimited table cache. Pre-load table handle now.
       // Need to do it out of the mutex.
-      builder->LoadTableHandlers(
+      s = builder->LoadTableHandlers(
           cfd->internal_stats(), db_options_->max_file_opening_threads,
           false /* prefetch_index_and_filter_in_cache */,
           true /* is_initial_load */,
           cfd->GetLatestMutableCFOptions()->prefix_extractor.get());
+      if (!s.ok()) {
+        if (db_options_->paranoid_checks) {
```
riversand963: Ditto.

cheng-chang:
@riversand963 PTAL

db/version_set.cc
```diff
@@ -3788,41 +3788,49 @@ Status VersionSet::ProcessManifestWrites(
...
+              new log::Writer(std::move(file_writer), 0, false));
+          s = WriteCurrentStateToManifest(curr_state, descriptor_log_.get());
+        }
       }
-    }
 
-    if (!first_writer.edit_list.front()->IsColumnFamilyManipulation()) {
-      for (int i = 0; i < static_cast<int>(versions.size()); ++i) {
-        versions[i]->PrepareApply(*mutable_cf_options_ptrs[i], true);
+      if (!first_writer.edit_list.front()->IsColumnFamilyManipulation()) {
```
riversand963: Looks like your intention is to avoid calling `versions[i]->PrepareApply()` when `!s.ok()`. If so, you may also need to check the result of `WriteCurrentStateToManifest()`.

cheng-chang: right, I guarded this if block with another s.ok() check.

cheng-chang:
@riversand963 PTAL

db/version_set.cc
```diff
@@ -3788,20 +3788,26 @@ Status VersionSet::ProcessManifestWrites(
...
-    // This is fine because everything inside of this block is serialized --
-    // only one thread can be here at the same time
-    if (new_descriptor_log) {
+    if (s.ok() && new_descriptor_log) {
+      // This is fine because everything inside of this block is serialized --
+      // only one thread can be here at the same time
       // create new manifest file
       ROCKS_LOG_INFO(db_options_->info_log, "Creating manifest %" PRIu64 "\n",
-                     pending_manifest_file_number_);
+                    pending_manifest_file_number_);
```
riversand963: I believe original format is the result of `make format`. Can you run make format to fix the indentation of `pending_manifest_file_number_`?

cheng-chang:
Ignoring the failure of the known flaky test `EnvPosixTestWithParam.DecreaseNumBgThreads` and landing now.

PR #6370, opened 2020-02-04
Be able to read compatible leveldb sst files

cheng-chang:
Summary:
In `DBSSTTest.SSTsWithLdbSuffixHandling`, some sst files are renamed to ldb files, the original intention of the test is to test that the ldb files can be loaded along with the sst files.

The original test checks this by `ASSERT_NE("NOT_FOUND", Get(Key(k)))`, but the problem is `Get(Key(k))` returns IO error due to path not found instead of NOT_FOUND, so the success of ASSERT_NE does not mean the key can be retrieved.

This PR updates the test to make sure Get(Key(k)) returns the original value.

Test Plan:
make db_sst_test && ./db_sst_test

### DIFF
--- a/db/table_cache.cc
+++ b/db/table_cache.cc
@@ -100,6 +100,10 @@ Status TableCache::GetTableReader(
   std::unique_ptr<FSRandomAccessFile> file;
   Status s = ioptions_.fs->NewRandomAccessFile(fname, file_options, &file,
                                                nullptr);
+  if (s.IsPathNotFound()) {
+    fname = Rocks2LevelTableFileName(fname);
+    s = ioptions_.fs->NewRandomAccessFile(fname, file_options, &file, nullptr);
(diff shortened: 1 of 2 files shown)


riversand963:
Just a reminder, please run `make format` before merging.

db/table_cache.cc
```diff
@@ -100,6 +100,10 @@ Status TableCache::GetTableReader(
   std::unique_ptr<FSRandomAccessFile> file;
   Status s = ioptions_.fs->NewRandomAccessFile(fname, file_options, &file,
                                                nullptr);
+  if (s.IsPathNotFound()) {
```
riversand963: One minor thing, you probably need to call `RecordTick(ioptions_.statistics, NO_FILE_OPENS)` even for failed file open attempts.

PR #6374, opened 2020-02-05
Support move semantics for PinnableSlice

cheng-chang:
Summary:
It's logically correct for PinnableSlice to support move semantics to transfer ownership of the pinned memory region. This PR adds both move constructor and move assignment to PinnableSlice.

Test Plan:
A set of unit tests for the move semantics are added in slice_test.
So `make slice_test && ./slice_test`.

### DIFF
--- a/util/slice.cc
+++ b/util/slice.cc
@@ -209,4 +209,33 @@ const SliceTransform* NewNoopTransform() {
   return new NoopTransform;
 }
 
+PinnableSlice::PinnableSlice(PinnableSlice&& other) {
+  *this = std::move(other);
+}
+
+PinnableSlice& PinnableSlice::operator=(PinnableSlice&& other) {
+  if (this != &other) {
+    Cleanable::Reset();
+    Cleanable::operator=(std::move(other));
+    size_ = other.size_;
+    pinned_ = other.pinned_;
+    if (pinned_) {
+      data_ = other.data_;
+      // When it's pinned, buf should no longer be of use.
+    } else {
+      if (other.buf_ == &other.self_space_) {
+        self_space_ = std::move(other.self_space_);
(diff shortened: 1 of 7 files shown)


util/slice.cc
```diff
@@ -209,4 +209,33 @@ const SliceTransform* NewNoopTransform() {
...
+      if (other.buf_ == &other.self_space_) {
+        self_space_ = std::move(other.self_space_);
+        buf_ = &self_space_;
+        data_ = buf_->data();
+      } else {
+        buf_ = other.buf_;
+        data_ = other.data_;
+      }
+    }
+    other.size_ = 0;
```
anand1976: Would be a good idea to also reset ```other.buf_``` to its ```self_space_```. What would be the value of ```other.data_```?

util/slice_test.cc
```diff
@@ -0,0 +1,159 @@
...
+    PinnableSlice v2(std::move(v1));
+
+    AssertEmpty(v1);
+    // Since v1's Cleanable has been moved to v2,
+    // no cleanup should happen in Reset.
+    v1.Reset();
+    ASSERT_EQ(1, res);
+
+    AssertSameData(const_str1, v2);
+  }
```
anand1976: Can we try reusing ```v1``` to make sure its in a valid state?

util/slice.cc
cheng-chang: to be consistent, after resetting `other.buf_` to `other.self_space_`, `other.data_` needs to be set to `other.buf_`.

cheng-chang:
@anand1976 PTAL

cheng-chang:
@anand1976 updated the build files, thanks.

cheng-chang:
The failed test is the known flakey env_test.

PR #6375, opened 2020-02-05
Let new threads get to work first

cheng-chang:
Summary:
AppVeyor fails consistently with error 
```
[ RUN      ] DefaultEnvWithoutDirectIO/EnvPosixTestWithParam.DecreaseNumBgThreads/0
c:\projects\rocksdb\env\env_test.cc(590): error: Expected equality of these values:
  1U
    Which is: 1
  env_->GetThreadPoolQueueLen(Env::Priority::HIGH)
    Which is: 2
```
This might be because after increasing the number of threads, the queued task is not scheduled onto the new thread during the time of delay set in the test. This PR tries to let the new threads get task from queue ASAP.

Test Plan:
Wait for results from AppVeyor.


pdillinger:
The test still failed in AppVeyor.

Same goal as #6232

PR #6378, opened 2020-02-05
Fix Buck build on macOS

chadaustin:
liburing is a Linux-specific dependency, so make sure it's configured in the Linux-only Buck rules.


chadaustin:
> btw, a n00b question: what is the difference between liburing in external deps vs "third-party//liburing:uring"?

I don't think there is a difference. Just two syntaxes for the same thing. Andrew Gallagher might know.

PR #6380, opened 2020-02-06
Add status checks during DB::Open

cheng-chang:
Summary:
Several statuses were not checked during DB::Open.

Test Plan:
make check

### DIFF
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -442,22 +442,25 @@ Status DBImpl::Recover(
       assert(s.IsIOError());
       return s;
     }
-    GetDbIdentityFromIdentityFile(&db_id_);
-    if (immutable_db_options_.write_dbid_to_manifest) {
+    s = GetDbIdentityFromIdentityFile(&db_id_);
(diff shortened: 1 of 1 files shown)


db/db_impl/db_impl_open.cc
```diff
@@ -442,22 +442,25 @@ Status DBImpl::Recover(
...
-      versions_->LogAndApply(versions_->GetColumnFamilySet()->GetDefault(),
+      s = versions_->LogAndApply(versions_->GetColumnFamilySet()->GetDefault(),
                              mutable_cf_options, &edit, &mutex_, nullptr,
                              false);
     }
   } else {
-    SetIdentityFile(env_, dbname_, db_id_);
+    s = SetIdentityFile(env_, dbname_, db_id_);
+  }
+  if (!s.ok()) {
```
ajkr: Instead of returning early here in case of non-ok status, can you revert line 463? Then a non-ok status should fall through to the end of this function where it will be returned. It is convenient to minimize the number of exit points a function has in case common cleanup work needs to be done before returning.

PR #6382, opened 2020-02-06
Add utility class Defer

cheng-chang:
Summary:
Add a utility class `Defer` to defer the execution of a function until the Defer object goes out of scope.
Used in VersionSet:: ProcessManifestWrites as an example.
The inline comments for class `Defer` have more details.

Test Plan:
`make defer_test version_set_test && ./defer_test && ./version_set_test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -595,6 +595,7 @@ TESTS = \
 	db_secondary_test \
 	block_cache_tracer_test \
 	block_cache_trace_analyzer_test \
+	defer_test \
 
 ifeq ($(USE_FOLLY_DISTRIBUTED_MUTEX),1)
 	TESTS += folly_synchronization_distributed_mutex_test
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -595,6 +595,7 @@ TESTS = \
 	db_secondary_test \
 	block_cache_tracer_test \
 	block_cache_trace_analyzer_test \
+	defer_test \
```
riversand963: Nit: you also need to update src.mk and  CMakelists.txt. Then run `python buckifier/buckify_rocksdb.py` in order to update TARGETS file.

util/defer.h
```diff
@@ -0,0 +1,52 @@
...
+// The above code ensures that cleanups will always happen on returning.
+//
+// Without the help of Defer, you can
+// 1. every time when !s.ok(), do the cleanup;
+// 2. instead of returning when !s.ok(), continue the work only when s.ok(),
+//    but sometimes, this might lead to nested blocks of "if (s.ok()) {...}".
+//
+// With the help of Defer, you can centralize the cleanup logic inside the
+// lambda passed to Defer, and you can return immediately on failure when necessary.
+class Defer {
```
riversand963: Nit: do you intend  this class to be inherited? If not, can we make it final? Do you see any potential issue with this? Just a thought.

cheng-chang: Yes, I think this should be final.

Makefile
cheng-chang: updated, thanks.

cheng-chang:
The failed test is the flaky test in env_test.

PR #6383, opened 2020-02-06
Clean up VersionEdit a bit

ltamasi:
Summary:
This is a bunch of small improvements to `VersionEdit`. Namely, the patch

* Makes the names and order of variables, methods, and code chunks related
  to the various information elements more consistent, and adds missing
  getters for the sake of completeness.
* Initializes previously uninitialized stack variables.
* Marks all getters const to improve const correctness.
* Adds in-class initializers and removes the default ctor that would
  create an object with uninitialized built-in fields and call `Clear`
  afterwards.
* Adds a new type alias for new files and changes the existing `typedef`
  for deleted files into a type alias as well.
* Makes the helper method `DecodeNewFile4From` private.
* Switches from long-winded iterator syntax to range based loops in a
  couple of places.
* Fixes a couple of assignments where an integer 0 was assigned to
  boolean members.

Test Plan:
make check

### DIFF
--- a/db/version_edit.cc
+++ b/db/version_edit.cc
@@ -118,28 +118,28 @@ void FileMetaData::UpdateBoundaries(const Slice& key, const Slice& value,
 }
 
 void VersionEdit::Clear() {
+  max_level_ = 0;
   db_id_.clear();
   comparator_.clear();
-  max_level_ = 0;
   log_number_ = 0;
   prev_log_number_ = 0;
-  last_sequence_ = 0;
   next_file_number_ = 0;
   max_column_family_ = 0;
   min_log_number_to_keep_ = 0;
+  last_sequence_ = 0;
   has_db_id_ = false;
   has_comparator_ = false;
   has_log_number_ = false;
   has_prev_log_number_ = false;
   has_next_file_number_ = false;
-  has_last_sequence_ = false;
   has_max_column_family_ = false;
   has_min_log_number_to_keep_ = false;
+  has_last_sequence_ = false;
   deleted_files_.clear();
   new_files_.clear();
   column_family_ = 0;
-  is_column_family_add_ = 0;
-  is_column_family_drop_ = 0;
+  is_column_family_add_ = false;
+  is_column_family_drop_ = false;
   column_family_name_.clear();
   is_in_atomic_group_ = false;
   remaining_entries_ = 0;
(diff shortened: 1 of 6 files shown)


db/version_edit.cc
```diff
@@ -463,6 +455,14 @@ Status VersionEdit::DecodeFrom(const Slice& src) {
         }
         break;
 
+      case kLastSequence:
```
cheng-chang: why should this case be moved here?

db/version_edit.cc
```diff
@@ -163,12 +163,12 @@ bool VersionEdit::EncodeTo(std::string* dst) const {
   if (has_next_file_number_) {
     PutVarint32Varint64(dst, kNextFileNumber, next_file_number_);
   }
-  if (has_last_sequence_) {
-    PutVarint32Varint64(dst, kLastSequence, last_sequence_);
-  }
   if (has_max_column_family_) {
     PutVarint32Varint32(dst, kMaxColumnFamily, max_column_family_);
   }
+  if (has_last_sequence_) {
```
cheng-chang: why move this block here?

db/version_edit.cc
```diff
@@ -648,13 +652,11 @@ std::string VersionEdit::DebugString(bool hex_key) const {
     r.append("\n  LastSeq: ");
     AppendNumberTo(&r, last_sequence_);
   }
-  for (DeletedFileSet::const_iterator iter = deleted_files_.begin();
-       iter != deleted_files_.end();
-       ++iter) {
+  for (const auto& deleted_file : deleted_files_) {
```
cheng-chang: use structured binding here to give `first` and `second` intuitive names directly?

db/version_edit.cc
```diff
@@ -686,10 +688,6 @@ std::string VersionEdit::DebugString(bool hex_key) const {
   if (is_column_family_drop_) {
     r.append("\n  ColumnFamilyDrop");
   }
-  if (has_max_column_family_) {
```
cheng-chang: why move this `if` block up?

db/version_edit.cc
```diff
@@ -767,12 +769,6 @@ std::string VersionEdit::DebugJSON(int edit_num, bool hex_key) const {
   if (is_column_family_drop_) {
     jw << "ColumnFamilyDrop" << column_family_name_;
   }
-  if (has_max_column_family_) {
```
cheng-chang: why move these `if` blocks up?

db/version_edit.h
```diff
@@ -237,56 +237,74 @@ struct LevelFilesBrief {
...
+    has_last_sequence_ = true;
+    last_sequence_ = seq;
+  }
+  bool HasLastSequence() const { return has_last_sequence_; }
+  SequenceNumber GetLastSequence() const { return last_sequence_; }
 
-  bool has_next_file_number() const { return has_next_file_number_; }
+  // Delete the specified "file" from the specified "level".
+  void DeleteFile(int level, uint64_t file) {
+    deleted_files_.insert({level, file});
```
cheng-chang: how about `emplace(level, file)` to construct in-place and save the copy in the initializer list?

db/version_edit.cc
ltamasi: That's a C++17 feature; we generally use C++11 because we have to support fairly old compilers.

db/version_edit.h
ltamasi: Good call.

PR #6386, opened 2020-02-07
db_bench supports for generating random variable sized value.

burtonli:
1. `db_bench` now supports `value_size_distribution_type`, `value_size_min`, `value_size_max` options for generating random variable sized value. 
2. Added `blob_db_compression_type` option for BlobDB to enable blob compression.


tools/db_bench_tool.cc
```diff
@@ -245,7 +245,15 @@ DEFINE_int32(threads, 1, "Number of concurrent threads to run.");
 DEFINE_int32(duration, 0, "Time in seconds for the random-ops tests to run."
              " When 0 then num & reads determine the test duration");
 
-DEFINE_int32(value_size, 100, "Size of each value");
+DEFINE_string(distribution_type, "fixed",
```
zhichao-cao: `value_size_distribution_type` might be more clear for users.

tools/db_bench_tool.cc
```diff
@@ -1416,20 +1429,126 @@ class ReportFileOpEnv : public EnvWrapper {
...
+      val = std::min(max_value_size_, val);
+    }
+    return val;
+  }
+ private:
+  virtual unsigned int Get(std::mt19937&) = 0;
+  virtual bool NeedTruncate() {
+    return false;
+  }
+  unsigned int min_value_size_;
```
zhichao-cao: Maybe just provide the interface in the BaseDistribution. Other distributions like Zipfian or Exp might can also inherited from this class. In that case, the private member such as  std::random_device, std::mt19937 cannot be used.

tools/db_bench_tool.cc
```diff
@@ -1447,13 +1566,9 @@ class RandomGenerator {
     return Slice(data_.data() + pos_ - len, len);
   }
 
-  Slice GenerateWithTTL(unsigned int len) {
```
zhichao-cao: Why this function is removed?

burtonli: 1. GenerateWithTTL is dead code, no one calls it.
2. GenerateWithTTL does the same with thing as 'Slice Generate(unsigned int len)'

tools/db_bench_tool.cc
burtonli: Thanks, I will update

tools/db_bench_tool.cc
```diff
@@ -1416,20 +1429,129 @@ class ReportFileOpEnv : public EnvWrapper {
...
 class RandomGenerator {
  private:
   std::string data_;
   unsigned int pos_;
+  std::unique_ptr<BaseDistribution> dist_;
 
  public:
+
   RandomGenerator() {
+    auto max_value_size = FLAGS_value_size_max;
```
zhichao-cao: Using switch can make it easier to extend to other distributions

tools/db_bench_tool.cc
```diff
@@ -1416,20 +1429,129 @@ class ReportFileOpEnv : public EnvWrapper {
...
+    if (NeedTruncate()) {
+      val = std::max(min_value_size_, val);
+      val = std::min(max_value_size_, val);
+    }
+    return val;
+  }
+ private:
+  virtual unsigned int Get() = 0;
+  virtual bool NeedTruncate() {
+    return false;
```
zhichao-cao: It is safer to enable value size truncate in BaseDistribution. I suggest to set it to "true".

tools/db_bench_tool.cc
```diff
@@ -5343,24 +5472,24 @@ class Benchmark {
       } else if (query_type == 1) {
         // the Put query
         puts++;
-        int64_t value_size = ParetoCdfInversion(
+        int64_t size = ParetoCdfInversion(
```
zhichao-cao: `size` can be misleading. Maybe v_size or others.

tools/db_bench_tool.cc
```diff
@@ -6106,21 +6237,22 @@ class Benchmark {
       GenerateKeyFromInt(key_rand, merge_keys_, &key);
 
       Status s;
+      auto val = gen.Generate();
```
zhichao-cao: I suggest not to use `auto` here for val. The same in other places.

zhichao-cao:
LGTM, thanks for the contribution.

PR #6387, opened 2020-02-07
Check KeyContext status in MultiGet

anand1976:
Currently, any IO errors and checksum mismatches while reading data
blocks, are being ignored by the batched MultiGet. Its only looking at
the GetContext state. Fix that.

Test plan:
Add unit tests


anand1976:
@siying I updated HISTORY.md in the initial commit.

PR #6388, opened 2020-02-08
Pass info_log_level to the inner logger of AutoRollLogger

riversand963:
Before this fix, the info_log_level passed from CreateLoggerFromOptions() will
be ignored by AutoRollLogger::logger_. This PR fixes it by setting the info log
level of logger_ during ResetLogger().

Test plan (dev server):
```
COMPILE_WITH_TSAN=1 make all && make check
make all && make check
```


riversand963:
Appveyor failure seems unrelated.

PR #6393, opened 2020-02-09
Fix flaky test DecreaseNumBgThreads

cheng-chang:
Summary:
The DecreaseNumBgThreads test keeps failing on Windows in AppVeyor.
It fails because it depends on a timed wait for the tasks to be dequeued from the threadpool's internal queue, but within the specified time, the task might have not been scheduled onto the newly created threads.
https://github.com/facebook/rocksdb/pull/6232 tries to fix this by waiting for longer time to let the threads scheduled.
This PR tries to fix this by replacing the timed wait with a synchronization on the task's internal conditional variable.
When the number of threads increases, instead of guessing the time needed for the task to be scheduled, it directly blocks on the conditional variable until the task starts running.
But when thread number is reduced, it still does a timed wait, but this does not lead to the flakiness now, will try to remove these timed waits in a future PR.

Test Plan:
Wait to see whether AppVeyor tests pass.


riversand963:
Could you reference #6232 so that people looking at either of these two PRs can be aware of both PRs. @cheng-chang

env/env_test.cc
```diff
@@ -567,26 +567,27 @@ TEST_P(EnvPosixTestWithParam, TwoPools) {
...
   env_->SetBackgroundThreads(1, Env::Priority::HIGH);
-  Env::Default()->SleepForMicroseconds(kDelayMicros);
 
   // Schedule 3 tasks. 0 running; Task 1, 2 waiting.
   for (size_t i = 0; i < 3; i++) {
     env_->Schedule(&test::SleepingBackgroundTask::DoSleepTask, &tasks[i],
                    Env::Priority::HIGH);
-    Env::Default()->SleepForMicroseconds(kDelayMicros);
   }
+  ASSERT_FALSE(tasks[0].TimedWaitUntilSleeping(kWaitMicros));
```
riversand963: In extreme cases, this assertion can still fail if the task is not scheduled onto a thread after 1 min, right?

riversand963:
Also, ln609 of env_test.cc. Should the test thread wait on the cond-var before asserting `IsSleep()` of task[1]?

test_util/testutil.h
```diff
@@ -429,6 +429,20 @@ class SleepingBackgroundTask {
       bg_cv_.Wait();
     }
   }
+  // Waits for the status to change to sleeping,
+  // otherwise times out.
+  // wait_time is in microseconds.
+  // Returns true when times out, false otherwise.
+  bool TimedWaitUntilSleeping(uint64_t wait_time) {
+    auto abs_time = Env::Default()->NowMicros() + wait_time;
```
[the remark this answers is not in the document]
cheng-chang: `SleepingBackgroundTask` does not have `env_`.

env/env_test.cc
cheng-chang: That's true.

cheng-chang: It's not an ideal fix. I'm thinking about how to make the scheduling logic inside threadpool_impl more deterministically testable.

cheng-chang:
@riversand963 Since `Wakeup` is internally synchronized on the cond-var, after `Wakeup` returns, `IsSleep` will always be true.

test_util/testutil.h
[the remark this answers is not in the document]
riversand963: okay then.

cheng-chang:
Had an offline discussion with @riversand963, after `WakeUp` signaling the cond var, to ensure that the task has received the signal before checking its status, there needs another `TimedWaitOnWokenUp` method.

PR #6403, opened 2020-02-11
Ensure all MultiGet IO errors are propagated to user

anand1976:
Unrevert the previous fix to propagate error status, and an additional fix to not treat a memtable lookup MergeInProgress status as an error.

Test plan:
Unit tests
Tried running stress tests but couldn't repro the stress failure


anand1976:

PR #6404, opened 2020-02-11
WriteUnPrepared: Untracked keys

lth:
For write unprepared, some applications may bypass the transaction api, and write keys directly into the write batch. However, since they are not tracked, rollbacks (both for savepoint and transaction) are not aware that these keys have to be rolled back.

The fix is to track them in `WriteUnpreparedTxn::untracked_keys_`. This is populated whenever we flush unprepared batches into the DB.

### DIFF
--- a/utilities/transactions/write_unprepared_txn.cc
+++ b/utilities/transactions/write_unprepared_txn.cc
@@ -95,6 +95,7 @@ void WriteUnpreparedTxn::Initialize(const TransactionOptions& txn_options) {
   largest_validated_seq_ = 0;
   assert(active_iterators_.empty());
   active_iterators_.clear();
+  untracked_keys_.clear();
 }
 
 Status WriteUnpreparedTxn::HandleWrite(std::function<Status()> do_write) {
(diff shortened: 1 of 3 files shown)


utilities/transactions/write_unprepared_txn.cc
```diff
@@ -286,6 +287,65 @@ Status WriteUnpreparedTxn::FlushWriteBatchToDBInternal(bool prepared) {
...
 
+  struct UntrackedKeyHandler : public WriteBatch::Handler {
+    WriteUnpreparedTxn* txn_;
+    bool rollback_merge_operands_;
+
+    UntrackedKeyHandler(WriteUnpreparedTxn* txn, bool rollback_merge_operands)
+        : txn_(txn), rollback_merge_operands_(rollback_merge_operands) {}
+
+    Status AddUntrackedKey(uint32_t cf, const Slice& key) {
+      auto str = std::string(key.data(), key.size());
```
maysamyabandeh: There is also a Slice::ToString function that we can use.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -577,49 +636,75 @@ Status WriteUnpreparedTxn::CommitInternal() {
   return s;
 }
 
+Status WriteUnpreparedTxn::WriteRollbackKeys(
+    const TransactionKeyMap& tracked_keys, WriteBatchWithIndex* rollback_batch,
+    ReadCallback* callback, const ReadOptions& roptions) {
+  const auto& cf_map = *wupt_db_->GetCFHandleMap();
+  auto fn = [&](const std::string& key, uint32_t cfid) {
```
maysamyabandeh: You might want to consider a name more descriptive than "fn".

utilities/transactions/write_unprepared_txn.h
```diff
@@ -315,6 +319,16 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
   // batch, it is possible that the delta iterator on the iterator will point to
   // invalid memory.
   std::vector<Iterator*> active_iterators_;
+
+  // Untracked keys that we have to rollback.
+  //
+  // TODO(lth): Currently we we do not record untracked keys per-savepoint.
+  // This means that when rolling back to savepoints, we have to check all
+  // keys in the current transaction for rollback. A more optimal solution
```
maysamyabandeh: I am confused how RollbackToSavepoint works. Would not this approach also rollback the keys before the save point?

lth: Yes it would, but that would only be inefficient, not incorrect because for RollbackToSavepoint, we do the rollback by reading the keys at the savepoint (not at the beginning of the transaction). This means that the rollback batch would just contain a reissue of the marker (ie. either a Put/Delete) and is not actually rolled back.

maysamyabandeh: Oh because of WriteUnpreparedTxnReadCallback? If yes, it would be helpful to add inline comments in two places: i) in RollbackToSavePointInternal to briefly the implication of using the callback, ii) here to mention the argument that you presented above.

Issue #6281, opened 2020-01-10
The proposal of Timestamp Ordering Transactions Over RocksDB


wolfkdy:
# Timestamp Ordering Transactions Over RocksDB

by @riversand963 & @wolfkdy

## Abstraction

In this article, we introduce the timestamp ordering transactions over RocksDB. It's a cooperation of three layers, the application-layer commits a transaction with an user-defined timestamp, certainly, the timestamp has to satisfy several restrictions. The transaction layer, which mainly handles write-write conflict check. The rocksdb layers, we augmented the internalkey with an extra commit-timestamp, Necessary modifications have to be added into read/write/flush/iterate/compaction paths to support the correct semantics.

## Introduction

Timestamp Ordering(T/O for short) transaction is a rising star in industrial database circles these two years. Although  it has been academicly investigated since the 1980s, its acceptance and popular use starts not long ago. CockroachDB[2], based on rocksdb, keeps a timestamp-based version-meta for every single key. Thus each read or write has to take extra cost to maintain the meta data. MongoDB4.0, based on wiredTiger3.0, fulfills T/O txns in an more elegant way[3], wiredTiger encodes timestamp in each key, The internal processes, such as version garbage collection, visibility check, have to consider both timestamps and snapshots.

Base on @riversand963 's current work about timestamps on RocksDB, I think it's time to make a conclusion, and move on to another greater stage. We proposal the t/o txn manager over RocksDB in this article,  In chapter 1, We will describe this functionality with a few code pieces. In this way, a brief impression can be gained. In the later chapters, we will show what is added into the rocksdb codebase to support this functionality. then, we will describe several restrictions that user-level has to obey.

## 1. Basic APIs

Timestamp Ordering Transaction provides these apis below, they are quite alike the tranditional transaction, except two timestamp-related apis. SetReadTimeStamp provides the ability of time-travel read. SetCommitTimeStamp provides the ability of user-defined asif-commit-timestamp.

```c++
class TOTransaction {
 public:
  Status SetCommitTimeStamp(const RocksTimeStamp& timestamp);
  Status GetReadTimeStamp(RocksTimeStamp&  timestamp) const;
  Status Commit();
  Status Rollback();
  Status Get(ReadOptions& options, const Slice& key, std::string& value);
  Iterator GetIterator(ReadOptions& read_options);
  Status Put(const Slice& key, const Slice& value);
  Status Delete(const Slice& key);
}
```

[...]

PR #6407, opened 2020-02-12
[WIP] timestamp ordering transaction

wolfkdy:
It's a txn-mgr implementation of proposal  https://github.com/facebook/rocksdb/issues/6281, which highly depends on the fundamental apis currently under development by @riversand963 .
Some api names are not compatible with rocksdb's current master branch, and will be fixed in the future commits.

### DIFF
--- a/include/rocksdb/utilities/totransaction.h
+++ b/include/rocksdb/utilities/totransaction.h
@@ -1,0 +1,96 @@
+#pragma once
+#ifndef ROCKSDB_LITE
+
+#include <string>
+#include <vector>
+
+#include "rocksdb/comparator.h"
+#include "rocksdb/db.h"
+#include "rocksdb/status.h"
+
+namespace rocksdb {
+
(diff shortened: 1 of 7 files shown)


wolfkdy:
+ @igorcanadi if you are interested.

BradBarnich:
Is it possible for this to use `WriteBatch->AssignTimestamps` instead of extending `WriteOptions`?

wolfkdy:
> Is it possible for this to use `WriteBatch->AssignTimestamps` instead of extending `WriteOptions`?
sure it is possible. TxnMgr layer should follow official rocksdb's underneath get/put API.
At this point of time(20200311), underneath part is not yet ready. And I think official guys are not ready to consider about this.
A neat runnable timestamp txn api can be found here: https://github.com/wolfkdy/rocksdb/tree/fb5.18_totdb, which follows official rocksdb's 5.18.3 branch.

PR #6410, opened 2020-02-13
Revert usage of Defer.

cheng-chang:
Summary:
Seems like this caused the following test failure on AppVeyor:
DBTest2.CrashInRecoveryMultipleCF
c:\projects\rocksdb\db\db_test_util.cc(107): error: DestroyDB(dbname_, options)
IO error: Failed to delete: C:\projects\rocksdb\db_tests\\testrocksdb-3112//db_test2_10791409581227174103/000013.sst: Access is denied.

Test Plan:
Wait to see whether the AppVeyor test passes.

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -50,7 +50,6 @@
 #include "table/two_level_iterator.h"
 #include "test_util/sync_point.h"
 #include "util/coding.h"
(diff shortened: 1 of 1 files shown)


riversand963:
Will keep on investigating, but probably not right now.

PR #6416, opened 2020-02-13
Add blob file state to VersionEdit

ltamasi:
Summary:
BlobDB currently does not keep track of blob files: no records are written to
the manifest when a blob file is added or removed, and upon opening a database,
the list of blob files is populated simply based on the contents of the blob directory.
This means that lost blob files cannot be detected at the moment. We plan to solve
this issue by making blob files a part of `Version`; as a first step, this patch makes
it possible to store information about blob files in `VersionEdit`. Currently, this information
includes blob file number, total number and size of all blobs, and total number and size
of garbage blobs. However, the format is extensible: new fields can be added in
both a forward compatible and a forward incompatible manner if needed (similarly
to `kNewFile4`).

Test Plan:
`make check`

### DIFF
--- a/db/blob_file_state.cc
+++ b/db/blob_file_state.cc
@@ -40,6 +40,8 @@ void BlobFileState::EncodeTo(std::string* output) const {
   PutVarint64(output, total_blob_bytes_);
   PutVarint64(output, garbage_blob_count_);
   PutVarint64(output, garbage_blob_bytes_);
+  PutLengthPrefixedSlice(output, checksum_method_);
+  PutLengthPrefixedSlice(output, checksum_value_);
 
   // Encode any custom fields here. The format to use is a Varint32 tag (see
   // CustomFieldTags above) followed by a length prefixed slice. Unknown custom
@@ -73,6 +75,18 @@ Status BlobFileState::DecodeFrom(Slice* input) {
   if (!GetVarint64(input, &garbage_blob_bytes_)) {
     return Status::Corruption(class_name, "Error decoding garbage blob bytes");
   }
+
+  Slice checksum_method;
+  if (!GetLengthPrefixedSlice(input, &checksum_method)) {
+    return Status::Corruption(class_name, "Error decoding checksum method");
+  }
+  checksum_method_ = checksum_method.ToString();
+
+  Slice checksum_value;
+  if (!GetLengthPrefixedSlice(input, &checksum_value)) {
+    return Status::Corruption(class_name, "Error decoding checksum value");
+  }
+  checksum_value_ = checksum_value.ToString();
 
   while (true) {
     uint32_t custom_field_tag = 0;
@@ -122,7 +136,9 @@ bool operator==(const BlobFileState& lhs, const BlobFileState& rhs) {
          lhs.GetTotalBlobCount() == rhs.GetTotalBlobCount() &&
          lhs.GetTotalBlobBytes() == rhs.GetTotalBlobBytes() &&
          lhs.GetGarbageBlobCount() == rhs.GetGarbageBlobCount() &&
-         lhs.GetGarbageBlobBytes() == rhs.GetGarbageBlobBytes();
+         lhs.GetGarbageBlobBytes() == rhs.GetGarbageBlobBytes() &&
+         lhs.GetChecksumMethod() == rhs.GetChecksumMethod() &&
+         lhs.GetChecksumValue() == rhs.GetChecksumValue();
 }
 
 bool operator!=(const BlobFileState& lhs, const BlobFileState& rhs) {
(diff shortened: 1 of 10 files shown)


db/blob_file_state.h
```diff
@@ -0,0 +1,93 @@
...
+    return !(garbage_blob_count_ < total_blob_count_);
+  }
+
+  void EncodeTo(std::string* output) const;
+  Status DecodeFrom(Slice* input);
+
+  std::string DebugString() const;
+  std::string DebugJSON() const;
+
+  bool operator==(const BlobFileState& rhs) const {
```
riversand963: We follow Google C++ style guide, which discourages overloading `==` as member function. https://google.github.io/styleguide/cppguide.html#Operator_Overloading

db/blob_file_state.cc
```diff
@@ -0,0 +1,141 @@
...
+  /////////////////////////////////////////////////////////////////////
+
+  kForwardIncompatibleMask = 1 << 6,
+
+  // Add forward incompatible fields here
+};
+
+}  // anonymous namespace
+
+void BlobFileState::EncodeTo(std::string* output) const {
```
riversand963: I wonder whether returning a boolean is better, assuming sometime in the future we want to notify the caller of success or failure. Note that `VersionEdit::EncodeTo` returns boolean.

db/blob_file_state.h
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  bool operator!=(const BlobFileState& rhs) const { return !(*this == rhs); }
+
+ private:
+  uint64_t blob_file_number_ = kInvalidBlobFileNumber;
+  uint64_t total_blob_count_ = 0;
+  uint64_t total_blob_bytes_ = 0;
+  uint64_t garbage_blob_count_ = 0;
+  uint64_t garbage_blob_bytes_ = 0;
```
riversand963: Should we reserve a field of type std::string for full file checksum of blob files? cc @zhichao-cao

ltamasi:
Thanks for the review @riversand963 !

db/version_edit.cc
```diff
@@ -68,11 +71,14 @@ enum CustomTag : uint32_t {
   kFileCreationTime = 6,
   kFileChecksum = 7,
   kFileChecksumFuncName = 8,
-  kPathId = 65,
+
+  // If this bit for the custom tag is set, opening DB should fail if
+  // we don't know this field.
+  kCustomTagNonSafeIgnoreMask = 1 << 6,
```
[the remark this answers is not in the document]
ltamasi: Not necessarily: `CustomTag` in `version_edit.cc` actually represents the custom fields of "NewFile4" records, while `CustomFieldTags` in `blob_file_state.cc` represents the custom fields of the new "BlobFileState" records, so neither the tags nor the mask need to be shared. I'm actually thinking about renaming the NewFile4 one to something like `NewFileCustomTag` to make this clearer, lmk what you think.

db/blob_file_state.h
ltamasi: That only makes a difference if implicit conversions are in play (which is not the case here) but will change it to be in sync. Thanks!

db/blob_file_state.cc
ltamasi: I'm actually thinking about changing it to `Status` to make it more flexible; there might be more than one failure case.

db/blob_file_state.h
ltamasi: Good point, thanks! Will reserve a field.

zhichao-cao: @riversand963 That's a good point. I think yes, it will be good to add full file checksum for each blob file. @ltamasi Thanks for reserve that. I will discuss with you later and see how to add the checksum for blob files.

ltamasi: @zhichao-cao Sounds good. Note that this is just groundwork; we can work on checksumming once blob files have been integrated into `Version`.

zhichao-cao: @ltamasi Sure!

db/blob_file_state.cc
ltamasi: Actually, I'd say let's leave this for now; we can always change the signature later if it becomes necessary. I would also argue that `VersionEdit::EncodeTo` returning a `bool` is not a good design: it only returns `false` if a newly added file does not have a valid smallest or largest key, which could have (should have) been detected when the file was added to the edit.

riversand963: It does not sound like an error in "encoding", maybe we should change VersionEdit::EncodeTo.
OK with your approach here.

db/version_edit.cc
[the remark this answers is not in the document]
riversand963: SGTM.

PR #6418, opened 2020-02-14
WriteUnPrepared: Enable WAL during crash recovery

lth:
Unfortunately, it seems like mysqld reuses xids across machine restarts. When that happens, we could have something like the following happening:

```
BEGIN_PREPARE(unprepared) Put(a) END_PREPARE(xid = 1)
-- crash and recover with Put(a) rolled back as it was not prepared
BEGIN_PREPARE(prepared) Put(b) END_PREPARE(xid = 1)
COMMIT(xid = 1)
-- crash and recover with both a, b
```

To solve this, we will have to log the rollback batch into the WAL during recovery.

WritePrepared already logs the rollback batch into the WAL, if a rollback happens after prepare, so there is no problem there.

### DIFF
--- a/utilities/transactions/write_unprepared_txn_db.cc
+++ b/utilities/transactions/write_unprepared_txn_db.cc
@@ -21,10 +21,23 @@ Status WriteUnpreparedTxnDB::RollbackRecoveredTransaction(
   assert(rtxn->unprepared_);
   auto cf_map_shared_ptr = WritePreparedTxnDB::GetCFHandleMap();
   auto cf_comp_map_shared_ptr = WritePreparedTxnDB::GetCFComparatorMap();
+  // In theory we could write with disableWAL = true during recovery, and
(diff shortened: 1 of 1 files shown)


maysamyabandeh:
Even if the xid reuse problem did not exist, this scenario says that we might have ended up with a prepared transaction that is never rolled back or committed. Is that correct?

lth:
I don't think so? unless I'm misunderstanding the scenario. You would have:

```
BEGIN_PREPARE(unprepared) Put(a) END_PREPARE(xid = 1)
-- crash
BEGIN_PREPARE(prepared) Put(b) END_PREPARE(xid = 2)
-- crash
```

On recovery, you would detect that xid 1 was not prepared, and thus should be rolled back. You would detect that xid 2 was prepared, and we would return that to the application to decide on its fate.

PR #6419, opened 2020-02-14
WriteUnPrepared: Fix assertion during recovery

lth:
Multiple (un)prepared batches could exist in the same WAL record due to group commit. This breaks an assertion in `MemTableInserter::MarkBeginPrepare` during recovery.

To fix, reset unprepared_batch_ to false after `MarkEndPrepare`.


lth:
I've updated the description to clarify, but yeah I meant group commit during normal writes (not during recovery).

PR #6422, opened 2020-02-16
Add Functions to OptionTypeInfo

mrambacher:
Added functions for parsing, serializing, and comparing elements to OptionTypeInfo.  These functions allow all of the special cases that could not be handled directly in the map of OptionTypeInfo to be moved into the map.  Using these functions, every type can be handled via the map rather than special cased.

By adding these functions, the code for handling options can become more standardized (fewer special cases) and (eventually) handled completely by common classes.


table/block_based/filter_policy.cc
```diff
@@ -756,4 +756,34 @@ FilterBuildingContext::FilterBuildingContext(
 
 FilterPolicy::~FilterPolicy() { }
 
+Status FilterPolicy::CreateFromString(
```
zhichao-cao, line 759: You mean give the example of the input string "Value" and also how and when to use the CreateFromString in the comments.

options/cf_options.cc
```diff
@@ -297,10 +375,69 @@ std::unordered_map<std::string, OptionTypeInfo>
...
+            // Nested options
+            auto tf = reinterpret_cast<std::shared_ptr<TableFactory>*>(addr);
+            BlockBasedTableOptions table_opts, base_opts;
+            BlockBasedTableFactory* bbtf =
+                static_cast_with_check<BlockBasedTableFactory, TableFactory>(
+                    tf->get());
+            if (bbtf != nullptr) {
+              base_opts = bbtf->table_options();
+            }
+            Status s = GetBlockBasedTableOptionsFromString(base_opts, value,
```
zhichao-cao, line 412: If we based on the original logic if !GetBlockBasedTableOptionsFromString().ok(), it should returns: 
```
return Status::InvalidArgument(
            "unable to parse the specified CF option " + name);
```
But here, s is directly returned. The behavior will be different from the current code. Other cases such as  plain_table_factory, memtable also has the similar mismatches.

options/cf_options.cc
```diff
@@ -280,9 +348,19 @@ std::unordered_map<std::string, OptionTypeInfo>
           OptionType::kCompressionType, OptionVerificationType::kNormal,
           OptionTypeFlags::kMutable,
           offsetof(struct MutableCFOptions, bottommost_compression)}},
-        {kNameComparator,
+        {"comparator",
          {offset_of(&ColumnFamilyOptions::comparator), OptionType::kComparator,
-          OptionVerificationType::kByName, OptionTypeFlags::kNone, 0}},
+          OptionVerificationType::kByName, OptionTypeFlags::kCompareLoose, 0,
+          [](const ConfigOptions& /*opts*/, const std::string& /*name*/,
+             const std::string& value, char* addr) {
```
zhichao-cao, line 355: Could you add the comments for each Lambda expression to explain what it is used for? Otherwise it is hard for user to follow.

options/cf_options.cc
```diff
@@ -297,10 +375,69 @@ std::unordered_map<std::string, OptionTypeInfo>
...
+            return s;
+          }}},
+        {"plain_table_factory",
+         {offset_of(&ColumnFamilyOptions::table_factory),
+          OptionType::kTableFactory, OptionVerificationType::kAlias,
+          OptionTypeFlags::kCompareLoose, 0,
+          [](const ConfigOptions& /*opts*/, const std::string& /*name*/,
+             const std::string& value, char* addr) {
+            // Nested options
+            auto tf = reinterpret_cast<std::shared_ptr<TableFactory>*>(addr);
```
zhichao-cao, line 426: In the Lambda expression, can you keep the original variable names such as plain_table_factory, table_opt? ptf and tf are hard to follow. Also for other cases like block_based_table_factory, memtable.

options/cf_options.cc
mrambacher: In the original code, you would lose whatever error you got from the lower-level call.  For example, if you passed in a bad block-based table option, you would just get "Unable to parse block_based_table_factory" and have no idea or means to figure out why.  This code preserves the same error semantics -- returns success or failure under the same scenarios -- but gives the true underlying reason for the error rather than a generic, ambiguous wrapper.

The semantics are the same as the existing code but we return a more informative error.  Isn't that a good thing?

options/db_options.cc
```diff
@@ -332,6 +333,30 @@ std::unordered_map<std::string, OptionTypeInfo>
...
+            return Status::OK();
+          }}},
+        {"env",
+         {offsetof(struct DBOptions, env), OptionType::kUnknown,
+          OptionVerificationType::kNormal,
+          (OptionTypeFlags::kStringNone | OptionTypeFlags::kCompareNever), 0,
+          [](const ConfigOptions& /*opts*/, const std::string& /*name*/,
+             const std::string& value, char* addr) {
+            auto env = reinterpret_cast<Env**>(addr);
+            return Env::LoadEnv(value, env);
```
zhichao-cao, line 358: Although in LoadEnv, if ObjectRegistry::NewInstance()->NewStaticObject<Env>(value, &env) fails, env will not be changed. To be safe (e.g., LoadEnv might be changed in the future and does not provide such protection), it would be better to follow the original logic.

```
    Env* env = new_options->env;
      Status status = Env::LoadEnv(value, &env);
      // Only support static env for now.
      if (status.ok()) {
        new_options->env = env;
      }
```

options/options_helper.cc
```diff
@@ -1015,28 +916,26 @@ Status StringToMap(const std::string& opts_str,
...
-    bool result =
-        SerializeSingleOptionHelper(opt_address, opt_info.type, &value);
-    if (result) {
-      opt_string->append(iter.first + "=" + value + cfg_options.delimiter);
-    } else {
-      return Status::InvalidArgument("failed to serialize %s\n",
-                                     iter.first.c_str());
+    // If the option is no longer used in rocksdb and marked as deprecated,
+    // we skip it in the serialization.
+    if (opt_info.ShouldSerialize()) {
```
zhichao-cao, line 928: Add comments to explicitly explain the case of IsDeprecated(). Also, if !IsDeprecated() but IsAlias() or IsEnabled(OptionTypeFlags::kStringNone), ShouldSerialize() also returns False, has the behavior been changed?

options/options_helper.cc
```diff
@@ -1015,28 +916,26 @@ Status StringToMap(const std::string& opts_str,
...
+      const char* opt_addr =
+          reinterpret_cast<const char*>(opt_ptr) + opt_info.offset;
+      std::string value;
+      Status s = opt_info.SerializeOption(config_options, iter.first, opt_addr,
+                                          &value);
+      if (s.ok()) {
+        opt_string->append(iter.first + "=" + value + config_options.delimiter);
+      } else {
+        return s;
```
zhichao-cao, line 937: Should we return return `Status::InvalidArgument("failed to serialize %s\n", iter.first.c_str())` here?

mrambacher: If we do that, we lose information.  The OptionTypeInfo::SerializeOption returned a specific status.  If we wrap it with a generic status, we have no idea what the lower level status returned.

options/options_helper.cc
```diff
@@ -1102,41 +1001,14 @@ static Status ParseDBOption(const ConfigOptions& config_options,
...
-      } else {
-        return Status::InvalidArgument(
-            "Unable to parse the specified DB option " + name);
-      }
-    }
-  } catch (const std::exception&) {
-    return Status::InvalidArgument("Unable to parse DBOptions:", name);
+  auto iter = db_options_type_info.find(name);
+  if (iter == db_options_type_info.end()) {
+    return Status::InvalidArgument("Unrecognized option DBOptions:", name);
```
zhichao-cao, line 1006: Would be better to ensure the error messanges are the same. For CF cases and also DB cases. After the change, the error message will be like "Deserializing the option xxxx is not supported" and "Error parsing xxxx". But the original one will be "Deserializing the specified DB option xxxx is not supported" or "Deserializing the specified CF option xxxx is not supported". Similar for "Unable to parse the specified DB option" and others

options/options_helper.cc
mrambacher: IsAlias and IsEnabled(kStringNone) are new.  There were some options previously that were listed in the "parse" code but not in the "serialize" code.  When I moved them into the OptionTypeInfo, I marked them as "kStringNone".

IsAlias is also new and is used when there are multiple names referring to the same potential object -- like "memtable" and "memtable_factory" or "table_factory" and "block_based_table_factory".  In those cases, we do not want to serialize the alias as it will be handled by the definition.  

I will add a comment in "ShouldSerialize" explaining this intended behavior

options/options_helper.cc
mrambacher: - If the option is not found (iter==end), the output is exactly the same
- If we caught an exception, the status is the same (InvalidArgument) but now also contains information about the exception.
- If we failed to parse but it was a "Name" option, the status is the same (NotSupported) but no longer contains "DBOption" in the text.  This allows the unification of the code 
- If the value could not be parsed, the status is the same but now says "Error parsing"  rather than  "unable to parse specified DBOption".

If you look around the code base, this nearly identical code was in several places and each treated the errors slightly differently and printed slightly different messages.  This code is a first pass at an attempt to unify the messages and errors that are returned.  In the end, I would like to see NotSupported, NotFound, and InvalidArgument mean specific things for all options:
- NotSupported: It is a valid option but we do not know how to parse the value
- InvalidArgument: It is a valid option name but we cannot parse the value for some reason
- NotFound: It is not a valid option name
Unifying these errors will make it easier to document them, expand the use/scope of options, and allow the caller to know what to expect.

options/cf_options.cc
zhichao-cao: > I am not sure what you mean by "original variable name"? Do you mean the name in the options structure or something else?

I mean, just replace tf, ptf with table_factory, plain_table_factory to make it more readable

options/options_helper.cc
zhichao-cao: > IsAlias and IsEnabled(kStringNone) are new. There were some options previously that were listed in the "parse" code but not in the "serialize" code. When I moved them into the OptionTypeInfo, I marked them as "kStringNone".
> 
> IsAlias is also new and is used when there are multiple names referring to the same potential object -- like "memtable" and "memtable_factory" or "table_factory" and "block_based_table_factory". In those cases, we do not want to serialize the alias as it will be handled by the definition.
> 
> I will add a comment in "ShouldSerialize" explaining this intended behavior

Maybe use another name instead of "kStringNone", it is hard to understand.

options/cf_options.cc
zhichao-cao: > In the original code, you would lose whatever error you got from the lower-level call. For example, if you passed in a bad block-based table option, you would just get "Unable to parse block_based_table_factory" and have no idea or means to figure out why. This code preserves the same error semantics -- returns success or failure under the same scenarios -- but gives the true underlying reason for the error rather than a generic, ambiguous wrapper.
> 
> The semantics are the same as the existing code but we return a more informative error. Isn't that a good thing?

I see. Now I got your argument. It would be good to return the lower-level call status.

options/options_helper.cc
zhichao-cao: > * If the option is not found (iter==end), the output is exactly the same
> * If we caught an exception, the status is the same (InvalidArgument) but now also contains information about the exception.
> * If we failed to parse but it was a "Name" option, the status is the same (NotSupported) but no longer contains "DBOption" in the text.  This allows the unification of the code
> * If the value could not be parsed, the status is the same but now says "Error parsing"  rather than  "unable to parse specified DBOption".
> 
> If you look around the code base, this nearly identical code was in several places and each treated the errors slightly differently and printed slightly different messages. This code is a first pass at an attempt to unify the messages and errors that are returned. In the end, I would like to see NotSupported, NotFound, and InvalidArgument mean specific things for all options:
> 
> * NotSupported: It is a valid option but we do not know how to parse the value
> * InvalidArgument: It is a valid option name but we cannot parse the value for some reason
> * NotFound: It is not a valid option name
>   Unifying these errors will make it easier to document them, expand the use/scope of options, and allow the caller to know what to expect.

It is clear for me now. It would be better to add these explanation of error message as the comments above ParseOption.

mrambacher: I added the error status codes to the ParseOption and GetXXXFromMap methods.

options/options_helper.cc
mrambacher: Renamed to kDontSerialize

options/cf_options.cc
```diff
@@ -47,8 +47,79 @@ int offset_of(T1 AdvancedColumnFamilyOptions::*member) {
...
+      return Status::InvalidArgument(
+          "unable to parse the specified CF option " + name);
+    }
+    compression_opts.enabled =
+        ParseBoolean("", value.substr(start, value.size() - start));
+  }
+  return Status::OK();
+}
+
+const std::string kOptNameBMCompOpts = "bottommost_compression_opts";
```
zhichao-cao, line 121: Any reasons that not using the string directly in the map? Some of the enum are replaced by the string name in the map.

table/block_based/block_based_table_factory.cc
```diff
@@ -244,8 +243,46 @@ static std::unordered_map<std::string, OptionTypeInfo>
...
+              *value = (*policy)->Name();
+            } else {
+              *value = kNullptrString;
+            }
+            return Status::OK();
+          },
+          // Compares two FilterPolicy objects for equality
+          [](const ConfigOptions&, const std::string&, const char* addr1,
+             const char* addr2, std::string*) {
+            const auto* thisOne =
```
zhichao-cao, line 271: Better rename thisOne and thatOne to casted_addr1, casted_addr2.

options/cf_options.cc
mrambacher: I left those names as constants as I will use the constants later when doing save/restore operations for the names.  For the moment, they are not really needed.

include/rocksdb/cache.h
```diff
@@ -142,6 +143,13 @@ class Cache {
   Cache(const Cache&) = delete;
   Cache& operator=(const Cache&) = delete;
 
+  // Creates a new Cache based on the input value string and returns the result
+  // The value might be an ID, and ID with properties, or an old-style cache
+  // initialization string.
+  static Status CreateFromString(const ConfigOptions& config_options,
```
zhichao-cao, line 149: Also give an example of the string value here to help understand how it can be used in other situations.

options/cf_options.cc
```diff
@@ -309,11 +457,24 @@ std::unordered_map<std::string, OptionTypeInfo>
...
+          OptionTypeFlags::kCompareLoose, 0,
+          // Parses the input value as a MergeOperator, updating the value
+          [](const ConfigOptions& /*opts*/, const std::string& /*name*/,
+             const std::string& value, char* addr) {
+            auto mop = reinterpret_cast<std::shared_ptr<MergeOperator>*>(addr);
+            Status status =
+                ObjectRegistry::NewInstance()->NewSharedObject<MergeOperator>(
+                    value, mop);
+            // Only support static comparator for now.
+            if (status.ok()) {
```
zhichao-cao, line 473: Do we need the logic here? It will always return Status OK().

options/options_type.h
```diff
@@ -123,18 +206,93 @@ class OptionTypeInfo {
...
+
+  // Compares the "addr1" and "addr2" values according to the rules of this
+  // class and returns true if they match.  On a failed match, mismatch is the
+  // name of the option that failed to match.
+  bool MatchesOption(const ConfigOptions& config_options,
+                     const std::string& opt_name, const char* addr1,
+                     const char* addr2, std::string* mismatch) const;
+
+  // Used to override the match rules for "ByName" options.
+  bool CheckByName(const ConfigOptions& config_options,
```
zhichao-cao, line 278: CheckByName is unclear to me. Maybe replace it with another name which shows its usage.

mrambacher: Renamed to MatchesByName

PR #6423, opened 2020-02-16
Add OptionTypeInfo::Enum and related methods

mrambacher:
Add methods and constructors for handling enums to the OptionTypeInfo.  This change allows enums to be converted/compared without adding a special "type" to the OptionType.  

This change addresses a couple of issues:
- It allows new enumerated types to be added to the options without editing the OptionType base class (and related methods)
- It standardizes the procedure for adding enumerated types to the options, reducing potential mistakes
- It moves the enum maps to the location where they are used, allowing them to be static file members rather than global values
- It reduces the number of types and cases that need to be handled in the various OptionType methods

### DIFF
--- a/options/db_options.cc
+++ b/options/db_options.cc
@@ -20,6 +20,29 @@
 
 namespace ROCKSDB_NAMESPACE {
 #ifndef ROCKSDB_LITE
+static std::unordered_map<std::string, WALRecoveryMode>
+    wal_recovery_mode_string_map = {
+        {"kTolerateCorruptedTailRecords",
+         WALRecoveryMode::kTolerateCorruptedTailRecords},
+        {"kAbsoluteConsistency", WALRecoveryMode::kAbsoluteConsistency},
+        {"kPointInTimeRecovery", WALRecoveryMode::kPointInTimeRecovery},
+        {"kSkipAnyCorruptedRecords",
+         WALRecoveryMode::kSkipAnyCorruptedRecords}};
+
+static std::unordered_map<std::string, DBOptions::AccessHint>
+    access_hint_string_map = {{"NONE", DBOptions::AccessHint::NONE},
+                              {"NORMAL", DBOptions::AccessHint::NORMAL},
+                              {"SEQUENTIAL", DBOptions::AccessHint::SEQUENTIAL},
+                              {"WILLNEED", DBOptions::AccessHint::WILLNEED}};
+
+static std::unordered_map<std::string, InfoLogLevel> info_log_level_string_map =
+    {{"DEBUG_LEVEL", InfoLogLevel::DEBUG_LEVEL},
+     {"INFO_LEVEL", InfoLogLevel::INFO_LEVEL},
+     {"WARN_LEVEL", InfoLogLevel::WARN_LEVEL},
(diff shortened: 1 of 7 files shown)


options/options_type.h
```diff
@@ -198,6 +226,44 @@ class OptionTypeInfo {
...
 
+  // Creates an OptionTypeInfo for an enum type.  Enums use an additional
+  // map to convert the enums to/from their string representation.
+  // @param _offset The offset in the options for this enum
+  // @param map The string to enum mapping for this enum
+  template <typename T>
+  static OptionTypeInfo Enum(
+      int _offset, const std::unordered_map<std::string, T>* const map) {
+    return OptionTypeInfo(
+        _offset, OptionType::kEnum, OptionVerificationType::kNormal,
```
zhichao-cao, line 237: Better to add comments for each lambda function here to explain what they are used for.

options/db_options.cc
```diff
@@ -20,6 +20,29 @@
 
 namespace ROCKSDB_NAMESPACE {
 #ifndef ROCKSDB_LITE
+static std::unordered_map<std::string, WALRecoveryMode>
```
zhichao-cao, line 23: If these enum maps are used in other .cc files, user may need to add the declaration. Do you think if you add the declaration of these enums to the corresponding .h files will be easy to use (similar as in options_help.h)?

mrambacher: I do not believe the enum maps will be required outside of this file.  For the most part, I do not think I changed the scope of where any of these enum maps were required/used.

options/db_options.cc
```diff
@@ -268,12 +290,12 @@ std::unordered_map<std::string, OptionTypeInfo>
           OptionType::kUInt64T, OptionVerificationType::kNormal,
           OptionTypeFlags::kNone, 0}},
         {"access_hint_on_compaction_start",
-         {offsetof(struct DBOptions, access_hint_on_compaction_start),
-          OptionType::kAccessHint, OptionVerificationType::kNormal,
-          OptionTypeFlags::kNone, 0}},
-        {"info_log_level",
-         {offsetof(struct DBOptions, info_log_level), OptionType::kInfoLogLevel,
```
zhichao-cao, line 275: Can you write a comment of instruction for developer who want to add new enum member or enum, where to change the code. Originally, it is straight forward: add the member, add the entry to the <string, optiontypeinfo> map with parameters for construction. But now, it has one more layer of OptionTypeInfo::Enum. May cause confusion.

options/db_options.cc
zhichao-cao: > I do not believe the enum maps will be required outside of this file. For the most part, I do not think I changed the scope of where any of these enum maps were required/used.

Got it. Thanks!

options/db_options.cc
mrambacher: I updated the comment in OptionTypeInfo to describe how to create an OptionTypeInfo::Enum.

Note that the new procedure (write a map and use OptionTypeInfo::Enum is much simpler than the old process (write a map, add a OptionType, add an OptionTypeInfo to the options, add handling of that new OptionType to ParseOptionHelper, SerializeSingleOptionHelper, and AreOptionsEqual) -- 4 of the steps have been eliminated.

PR #6424, opened 2020-02-16
Add OptionTypeInfo::Vector to parse/serialize vectors

mrambacher:
The OptionTypeInfo::Vector method allows a vector<T> to be converted to/from strings via the options.  

The kVectorInt and kVectorCompressionType vectors were replaced with this methodology.

As part of this change, the NextToken method was added to the OptionTypeInfo.  This method was refactored from code within the StringToMap function.

Future types that could use this functionality include the EventListener vectors.

### DIFF
--- a/options/options_helper.cc
+++ b/options/options_helper.cc
@@ -271,59 +271,6 @@ std::vector<CompressionType> GetSupportedCompressions() {
 }
 
 #ifndef ROCKSDB_LITE
-
-namespace {
-bool SerializeVectorCompressionType(const std::vector<CompressionType>& types,
-                                    std::string* value) {
-  std::stringstream ss;
-  bool result;
-  for (size_t i = 0; i < types.size(); ++i) {
-    if (i > 0) {
-      ss << ':';
-    }
-    std::string string_type;
-    result = SerializeEnum<CompressionType>(compression_type_string_map,
-                                            types[i], &string_type);
-    if (result == false) {
-      return result;
-    }
-    ss << string_type;
-  }
-  *value = ss.str();
-  return true;
-}
-
-bool ParseVectorCompressionType(
-    const std::string& value,
-    std::vector<CompressionType>* compression_per_level) {
-  compression_per_level->clear();
-  size_t start = 0;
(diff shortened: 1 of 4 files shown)


options/options_helper.cc
```diff
@@ -1086,6 +978,59 @@ std::unordered_map<std::string, CompactionStopStyle>
...
+      }
+      if (pos < opts.size() && opts[pos] != delimiter) {
+        return Status::InvalidArgument("Unexpected chars after nested options");
+      }
+      *end = pos;
+    } else {
+      return Status::InvalidArgument(
+          "Mismatched curly braces for nested options");
+    }
+  } else {
```
zhichao-cao, line 1022: Better to give an example in the comments, in which case the count!=0.

options/options_test.cc
```diff
@@ -3441,6 +3441,32 @@ TEST_F(OptionTypeInfoTest, TestStruct) {
...
+
+TEST_F(OptionTypeInfoTest, TestVectorType) {
+  OptionTypeInfo vec_info = OptionTypeInfo::Vector<std::string>(
+      0, OptionVerificationType::kNormal, OptionTypeFlags::kNone, 0,
+      {0, OptionType::kString});
+  std::vector<std::string> vec1, vec2;
+  std::string mismatch;
+
+  ConfigOptions config_options;
+  TestAndCompareOption(config_options, vec_info, "v", "a:b:c:d", &vec1, &vec2);
```
zhichao-cao, line 3453: Better to add the test case for other possible input string format (like {A=a:a1:a2;B=b:b1}?)

mrambacher: I am willing to add more test cases, but am not sure what exactly you are asking for.  Do you mean "A={a:a1:a2};" or A=a:{a1}:{a2}" or something else?   Your comment shows two separate options.

zhichao-cao: Yeah, you are right. Sorry about the misleading in the example.

mrambacher: Added more test cases for vectors with nested brackets and nested vectors

PR #6425, opened 2020-02-16
Add Struct Type to OptionsTypeInfo

mrambacher:
Added code for generically handing structs to OptionTypeInfo.  A struct is a collection of variables handled by their own map of OptionTypeInfos.  Examples of structs include Compaction and Cache options.

### DIFF
--- a/options/options_helper.cc
+++ b/options/options_helper.cc
@@ -322,80 +322,6 @@ bool ParseVectorCompressionType(
   }
   return true;
 }
-
-// This is to handle backward compatibility, where compaction_options_fifo
-// could be assigned a single scalar value, say, like "23", which would be
-// assigned to max_table_files_size.
-bool FIFOCompactionOptionsSpecialCase(const std::string& opt_str,
-                                      CompactionOptionsFIFO* options) {
-  if (opt_str.find("=") != std::string::npos) {
-    // New format. Go do your new parsing using ParseStructOptions.
-    return false;
-  }
-
-  // Old format. Parse just a single uint64_t value.
-  options->max_table_files_size = ParseUint64(opt_str);
-  return true;
-}
-
-static bool SerializeStruct(
-    const void* const opt_ptr, std::string* value,
-    const std::unordered_map<std::string, OptionTypeInfo>& type_info_map) {
-  ConfigOptions config_options;
-  config_options.delimiter = ";";
-
-  std::string opt_str;
-  Status s =
-      GetStringFromStruct(config_options, opt_ptr, type_info_map, &opt_str);
-  if (!s.ok()) {
-    return false;
-  }
-  *value = "{" + opt_str + "}";
-  return true;
-}
-
-static bool ParseSingleStructOption(
-    const ConfigOptions& config_options, const std::string& opt_val_str,
-    void* options,
-    const std::unordered_map<std::string, OptionTypeInfo>& type_info_map) {
-  size_t end = opt_val_str.find('=');
-  std::string key = opt_val_str.substr(0, end);
-  std::string value = opt_val_str.substr(end + 1);
-  auto iter = type_info_map.find(key);
-  if (iter == type_info_map.end()) {
-    return false;
-  }
-  const auto& opt_info = iter->second;
-  Status s = opt_info.ParseOption(
-      config_options, key, value,
-      reinterpret_cast<char*>(options) + opt_info.mutable_offset);
-  return s.ok();
-}
-
-static bool ParseStructOptions(
-    const std::string& opt_str, void* options,
-    const std::unordered_map<std::string, OptionTypeInfo>& type_info_map) {
-  assert(!opt_str.empty());
-  ConfigOptions config_options;
-
-  size_t start = 0;
-  if (opt_str[0] == '{') {
-    start++;
-  }
-  while ((start != std::string::npos) && (start < opt_str.size())) {
-    if (opt_str[start] == '}') {
-      break;
-    }
-    size_t end = opt_str.find(';', start);
-    size_t len = (end == std::string::npos) ? end : end - start;
-    if (!ParseSingleStructOption(config_options, opt_str.substr(start, len),
-                                 options, type_info_map)) {
-      return false;
-    }
-    start = (end == std::string::npos) ? end : end + 1;
-  }
-  return true;
-}
 }  // anonymouse namespace
 
 bool ParseSliceTransformHelper(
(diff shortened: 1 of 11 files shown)


options/options_type.h
```diff
@@ -283,6 +281,78 @@ class OptionTypeInfo {
...
+  //
+  // @param _offset The offset in the option object for this enum
+  // @param map The string to enum mapping for this enum
+  static OptionTypeInfo Struct(
+      const std::string& struct_name,
+      const std::unordered_map<std::string, OptionTypeInfo>* struct_map,
+      int _offset, OptionVerificationType _verification, OptionTypeFlags _flags,
+      int _mutable_offset) {
+    return OptionTypeInfo(
+        _offset, OptionType::kStruct, _verification, _flags, _mutable_offset,
```
zhichao-cao, line 312: Usually, we do not name a variable starting with underscore. Also, for Enum, I missed the _offset in previous PR. You may change them?

include/rocksdb/convenience.h
```diff
@@ -26,6 +26,10 @@ struct Options;
 // of the serialization (e.g. delimiter), and how to compare
 // options (sanity_level).
 struct ConfigOptions {
+  // Returns a ConfigOptions suitable for nested/embedded
+  // options.
+  ConfigOptions Embedded() const;
```
zhichao-cao, line 31: The usage of this function is unclear to me. Can you explain? Based on its implementation
` ConfigOptions embedded = *this;
  embedded.delimiter = ";";
  return embedded;`
Looks like we do not need this function?

options/options_type.h
```diff
@@ -283,6 +281,78 @@ class OptionTypeInfo {
...
+                                  const char* addr2, std::string* mismatch) {
+          return MatchesStruct(opts, struct_name, struct_map, name, addr1,
+                               addr2, mismatch);
+        });
+  }
+  static OptionTypeInfo Struct(
+      const std::string& struct_name,
+      const std::unordered_map<std::string, OptionTypeInfo>* struct_map,
+      int _offset, OptionVerificationType _verification, OptionTypeFlags _flags,
+      int _mutable_offset, const ParserFunc& _pfunc) {
```
zhichao-cao, line 338: Directly use parse_func. instead of _pfunc.

options/options_type.h
mrambacher: I renamed all of the member variables to end in "_" and removed any leading "_" from the parameters to the constructor

options/options_type.h
```diff
@@ -367,18 +439,56 @@ class OptionTypeInfo {
                      const std::string& opt_name, const char* this_ptr,
                      const std::string& that_value) const;
 
+  // Parses the input value according to the map for the struct at opt_addr
+  // struct_name is the name of the struct option as registered
+  // opt_name is the name of the option being evaluated.  This may
+  // be the whole struct or a sub-element of it, based on struct_name and
+  // opt_name.
```
zhichao-cao, line 446: It would be better to give an example to explain "This maybe the whole struct or a sub-element of it"

options/options_type.h
```diff
@@ -283,7 +281,79 @@ class OptionTypeInfo {
...
+  //
+  // To create an OptionTypeInfo that is a Struct, one should:
+  // - Create a static map of string-OptionTypeInfo corresponding to the
+  //   properties of the object that can be set via the options.
+  // - Call this method passing the name and map in as parameters.
+  // Note that it is not necessary to add a new OptionType or make any
+  // other changes -- the returned object handles parsing, serialization, and
+  // comparisons.
+  //
+  // @param _offset The offset in the option object for this enum
```
zhichao-cao, line 304: remove underscore of _offset

options/options_type.h
```diff
@@ -239,9 +237,9 @@ class OptionTypeInfo {
   // @param map The string to enum mapping for this enum
   template <typename T>
   static OptionTypeInfo Enum(
-      int _offset, const std::unordered_map<std::string, T>* const map) {
```
zhichao-cao, line 242: @param _offset to @param offset at L236

options/options_type.h
```diff
@@ -164,67 +162,67 @@ using EqualsFunc = std::function<bool(
...
-        parser_func(_pfunc),
-        string_func(_sfunc),
-        equals_func(_efunc),
-        type(_type),
-        verification(_verification),
-        flags(_flags) {}
+  int offset_;
+  int mutable_offset_;
+
+  // A simple "normal", non-mutable Type "_type_" at _offset
```
zhichao-cao, line 168: Also some of the comments need to be checked like _type_, _offset

options/options_type.h
```diff
@@ -367,18 +439,56 @@ class OptionTypeInfo {
...
+  static bool MatchesStruct(
+      const ConfigOptions& config_options, const std::string& struct_name,
+      const std::unordered_map<std::string, OptionTypeInfo>* map,
+      const std::string& opt_name, const char* this_offset,
+      const char* that_offset, std::string* mismatch);
+
+  // Finds the entry for the opt_name in the opt_map, returning
+  // nullptr if not found.
+  // If found, elem_name will be the name of option to find.
+  // This may be opt_name, or a substring of opt_name.
```
zhichao-cao, line 474: Do you mean elem_name will be the option name or the field name?

options/options_helper.cc
```diff
@@ -1438,6 +1329,61 @@ bool OptionTypeInfo::MatchesOption(const ConfigOptions& config_options,
   return false;
 }
 
+bool OptionTypeInfo::MatchesStruct(
+    const ConfigOptions& config_options, const std::string& struct_name,
+    const std::unordered_map<std::string, OptionTypeInfo>* struct_map,
+    const std::string& opt_name, const char* this_offset,
+    const char* that_offset, std::string* mismatch) {
```
zhichao-cao, line 1336: Would be better to use this_addr and that_addr instead of this_offset and that_offset.

mrambacher:
Based on the comments, I plan to do the following:
- The "function types" off of OptionTypeInfo are now ParseFunc, SerializeFunc, and EqualsFunc, with corresponding variable names.  They were Parse, String, And Equals.
- The OptionTypeInfo methods that use these functions are now "Parse", "Serialize", and "AreEqual" (they were ParseOption, SerializeOption, MatchesOption.  I dropped the word "Option" as it is redundant since it is already an Option object
- The equivalent methods for structs are ParseStruct, SerializeStruct, and StructsAreEqual.
- The helper "Matches" functions off OptionTypeInfo (like MatchesByName) that are now AreEqualByName

If this is agreeable, I will push out an update to the PR accordingly.  If it is not, please let me know what you prefer to see.

zhichao-cao:
> Based on the comments, I plan to do the following:
> 
> * The "function types" off of OptionTypeInfo are now ParseFunc, SerializeFunc, and EqualsFunc, with corresponding variable names.  They were Parse, String, And Equals.
> * The OptionTypeInfo methods that use these functions are now "Parse", "Serialize", and "AreEqual" (they were ParseOption, SerializeOption, MatchesOption.  I dropped the word "Option" as it is redundant since it is already an Option object
> * The equivalent methods for structs are ParseStruct, SerializeStruct, and StructsAreEqual.
> * The helper "Matches" functions off OptionTypeInfo (like MatchesByName) that are now AreEqualByName
> 
> If this is agreeable, I will push out an update to the PR accordingly. If it is not, please let me know what you prefer to see.

I'm ok with these changes. Thanks!

options/options_type.h
zhichao-cao: @mrambacher The _offset-> offset in comment

PR #6428, opened 2020-02-18
Ignore .vs/.vscode dir used by visual studio/ vscode

riversand963:


riversand963:
Appveyor test failure is not caused by this PR.

PR #6436, opened 2020-02-20
CI: add Arm support to travis CI matrix

guyuqi:
This patch based on #5932 offers a better solution to add arm64 to TravisCI matrix.
Really thank @adamretter for initiating Arm CI setup.

Difference comparing to amd64:
1. For CMake, as no official arm64 release ready on Kitware page,
a third party (conda-forge) released one is used instead of
building from source. The main reason is to save CI time.
2. Explicit export JAVA_HOME on arm64
3. Disable mingw test

Signed-off-by: Yuqi Gu <yuqi.gu@arm.com>

PR #6437, opened 2020-02-20
Fix db_wal_test::TruncateLastLogAfterRecoverWithoutFlush failure

guyuqi:
`TruncateLastLogAfterRecoverWithoutFlush` case depends on fallocate support
of underlying file system. 

On a file system which lacks of this feature, like zfs, it will fail to allocate predefined file size as this test case intends to do;

So a check block is added to detect fallocate support and skip test if not.
The related work is done by @JunHe77. Thanks!

Signed-off-by: Yuqi Gu <yuqi.gu@arm.com>

### DIFF
--- a/db/db_wal_test.cc
+++ b/db/db_wal_test.cc
@@ -1535,17 +1535,15 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
     fd = open(fname_test_fallocate.c_str(), O_CREAT | O_RDWR | O_TRUNC, 0644);
   } while (fd < 0 && errno == EINTR);
   ASSERT_GT(fd, 0);
-  int err_number = 0;
   int alloc_status = fallocate(fd, 0, 0, 1);
-  if (alloc_status != 0) {
-    err_number = errno;
-  }
+  int err_number = errno;
   close(fd);
   ASSERT_OK(options.env->DeleteFile(fname_test_fallocate));
-  if (alloc_status != 0) {
-    fprintf(stderr, "             Skipped preallocated space check: %s\n", strerror(err_number));
+  if (err_number == ENOSYS || err_number == EOPNOTSUPP) {
+    fprintf(stderr, "Skipped preallocated space check: %s\n", strerror(err_number));
     return;
   }
+  ASSERT_EQ(0, alloc_status);
 
   DestroyAndReopen(options);
   size_t preallocated_size =
--- a/db/db_wal_test.cc
+++ b/db/db_wal_test.cc
@@ -1527,6 +1527,26 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
   constexpr size_t kKB = 1024;
   Options options = CurrentOptions();
   options.avoid_flush_during_recovery = true;
+  // Test fallocate support of running file system.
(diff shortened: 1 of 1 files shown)


adamretter:
@guyuqi So we have a compile time flag `ROCKSDB_FALLOCATE_PRESENT` which is calculated, and controls aspects of the code that depend on fallocate support. See - https://github.com/facebook/rocksdb/blob/master/build_tools/build_detect_platform#L248

The code that you patch is only enabled when that flag is true. So it would seem to me that your patch indicates a problem with detecting a working fallocate (or some feature of it - Is this some combination of fallocate and truncate?).

Either way, it would seem prudent to fix the detection, either:
1) modifying the existing conditions for ROCKSDB_FALLOCATE_PRESENT, or
2) if this is about fallocate and truncate working together, then adding an additional flag

What do you think?

guyuqi:
@adamretter  
The `ROCKSDB_FALLOCATE_PRESENT` is the flag for **compile-time**.
This PR is to check whether fallocate is supported or not for **run-time**.

So this patch is needed when binary code run on a node which doesn't support fallocate though the code was compiled properly on another node where `ROCKSDB_FALLOCATE_PRESENT` was detected as 'true'.

adamretter:
@guyuqi I am a bit confused, as you said it is needed for https://github.com/facebook/rocksdb/pull/6436

But in that PR (i.e. for Travis) the code is compiled and run on the same platform.

guyuqi:
> the code is compiled and run on the same platform.

Yes, it's the specific case.
We should also add the run-time check to cover the scenario that compilation and code-running are in different places.

JunHe77:
@adamretter as you mentioned in [build_detect_platform](https://github.com/facebook/rocksdb/blob/master/build_tools/build_detect_platform#L249-L257), this flag is validated in compile time by checking if the test prog can be built or not, which means as long as `<linux/falloc.h>` is presented, this macro `ROCKSDB_FALLOCATE_PRESENT` will be defined.
But if underlying file system doesn't support fallocate feature, this test will fail to execute even on the very same machine where build is done.
Example:
Built passed but execution failed: https://travis-ci.org/JunHe77/rocksdb/jobs/654724257#L329-L367

guyuqi:
@siying , could you please have a look into this PR?

db/db_wal_test.cc
```diff
@@ -1527,6 +1527,26 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
...
+  ASSERT_GT(fd, 0);
+  int err_number = 0;
+  int alloc_status = fallocate(fd, 0, 0, 1);
+  if (alloc_status != 0) {
+    err_number = errno;
+  }
+  close(fd);
+  ASSERT_OK(options.env->DeleteFile(fname_test_fallocate));
+  if (alloc_status != 0) {
+    fprintf(stderr, "             Skipped preallocated space check: %s\n", strerror(err_number));
```
pdillinger: I would be more comfortable if we only skip when `err_number == ENOSYS || err_number == EOPNOTSUPP` and otherwise fail the test on `alloc_status != 0`.

guyuqi: Updated.
Thanks for your comments.

PR #6440, opened 2020-02-21
Some minor fix-ups

riversand963:
Cleanup some code without any real change in functionality.

### DIFF
--- a/db/arena_wrapped_db_iter.h
+++ b/db/arena_wrapped_db_iter.h
@@ -45,26 +45,26 @@ class ArenaWrappedDBIter : public Iterator {
   // Set the internal iterator wrapped inside the DB Iterator. Usually it is
   // a merging iterator.
   virtual void SetIterUnderDBIter(InternalIterator* iter) {
-    static_cast<DBIter*>(db_iter_)->SetIter(iter);
+    db_iter_->SetIter(iter);
   }
 
-  virtual bool Valid() const override { return db_iter_->Valid(); }
-  virtual void SeekToFirst() override { db_iter_->SeekToFirst(); }
-  virtual void SeekToLast() override { db_iter_->SeekToLast(); }
-  virtual void Seek(const Slice& target) override { db_iter_->Seek(target); }
-  virtual void SeekForPrev(const Slice& target) override {
+  bool Valid() const override { return db_iter_->Valid(); }
+  void SeekToFirst() override { db_iter_->SeekToFirst(); }
+  void SeekToLast() override { db_iter_->SeekToLast(); }
+  void Seek(const Slice& target) override { db_iter_->Seek(target); }
+  void SeekForPrev(const Slice& target) override {
     db_iter_->SeekForPrev(target);
   }
-  virtual void Next() override { db_iter_->Next(); }
-  virtual void Prev() override { db_iter_->Prev(); }
-  virtual Slice key() const override { return db_iter_->key(); }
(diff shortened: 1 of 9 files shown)


zhichao-cao:
@riversand963 One question, for the class `ArenaWrappedDBIter` and `BlockIter`, if virtual is removed, the future derived class functions cannot be called by the base. So you assume the derivation will not happen or the derivation member function will not be called by the base pointer?

riversand963:
Thanks @zhichao-cao for the review. I think `override` without `virtual` in the derived class provides more check, as [post](https://stackoverflow.com/questions/39932391/virtual-override-or-both-c) and [post](https://stackoverflow.com/questions/13880205/is-the-override-keyword-just-a-check-for-a-overridden-virtual-method) explain. The methods will still remain virtual because they are marked as `virtual` in the base class, i.e. `Iterator` and `InternalIteratorBase`.
Can you give an example to illustrate when it goes wrong?

zhichao-cao:
> Thanks @zhichao-cao for the review. I think `override` without `virtual` in the derived class provides more check, as [post](https://stackoverflow.com/questions/39932391/virtual-override-or-both-c) and [post](https://stackoverflow.com/questions/13880205/is-the-override-keyword-just-a-check-for-a-overridden-virtual-method) explain. The methods will still remain virtual because they are marked as `virtual` in the base class, i.e. `Iterator` and `InternalIteratorBase`.
> Can you give an example to illustrate when it goes wrong?

Thanks for the explanation. Yes, it can provide more check during compile. I'm thinking the following case, not sure it will happen or not:

```
class BlockIter:public Iter{
virtual seek() override{ func 1};
};

class DerBlockIter:public BlockIter {
seek() override{ func 2};
}

BlockIter* bp = new DerBlockIter();
bp->seek(); 
```

In this case, if virtual is removed, it can only call func1 instead of func2. I'm not sure, if the two class might be used in this way or not? Or we always use the abstract class pointer for all of them, if so, removing virtual is fine.

riversand963:
I don't think that is going to happen due to C++ polymorphism.
```
#include <cstdio>

class Base {
 public:
  virtual void foo() {
    fprintf(stdout, "base foo\n");
  }
};

class Derived : public Base {
 public:
   void foo() override {
    fprintf(stdout, "derived foo\n");
   }
};


int main() {
  Base* ptr = new Derived();
  ptr->foo();
  delete ptr;
  return 0;
}
```


```
$g++ -std=c++11 test.cc -o test
$./test
derived foo
```

zhichao-cao:
> I don't think that is going to happen due to C++ polymorphism.
> 
> ```
> #include <cstdio>
> 
> class Base {
>  public:
>   virtual void foo() {
>     fprintf(stdout, "base foo\n");
>   }
> };
> 
> class Derived : public Base {
>  public:
>    void foo() override {
>     fprintf(stdout, "derived foo\n");
>    }
> };
> 
> class Derived2 : public Derived {
>  public:
>   void foo() override {
>     fprintf(stdout, "derived2 foo\n");
>   }
> };
> 
> int main() {
>   Base* ptr = new Derived2();
>   ptr->foo();
>   delete ptr;
>   return 0;
> }
> ```
> 
> ```
> $g++ -std=c++11 test.cc -o test
> $./test
> derived2 foo
> ```

Thanks for the detailed example. You are right, removing virtual has no influence here.

riversand963:
Thanks again!

PR #6442, opened 2020-02-21
Add function to set row cache on rocksdb_options_t

morningsend:
Adding a C API function to set `row_cache` on `rocksdb_options_t` as this functionality is missing.

### DIFF
--- a/db/c.cc
+++ b/db/c.cc
@@ -2765,6 +2765,12 @@ void rocksdb_ratelimiter_destroy(rocksdb_ratelimiter_t *limiter) {
   delete limiter;
 }
 
+void rocksdb_options_set_row_cache(rocksdb_options_t* opt, rocksdb_cache_t* cache) {
+  if(cache) {
+    opt->rep.row_cache = cache->rep;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


db/c.cc
```diff
@@ -2765,6 +2765,12 @@ void rocksdb_ratelimiter_destroy(rocksdb_ratelimiter_t *limiter) {
   delete limiter;
 }
 
+void rocksdb_options_set_row_cache(rocksdb_options_t* opt, rocksdb_cache_t* cache) {
+  if(cache) {
```
riversand963: NIt: how about `opt->rep.row_cache = cache ? cache->rep : nullptr`?

morningsend: Thanks for your suggestion, I will update PR

riversand963: Sorry, on second thought, this may not be the desired behavior. What if opt->rep.row_cache is not nullptr before calling this function?

morningsend: Yeah, I was wondering the same thing, shall I leave it as it is?

riversand963: SGTM.

morningsend: Awesome!

PR #6444, opened 2020-02-21
Misc filter_bench improvements

pdillinger:
Summary: Useful in validating/testing internal fragmentation changes (#6427)

Test Plan: manual (no changes to production code)

### DIFF
--- a/util/filter_bench.cc
+++ b/util/filter_bench.cc
@@ -322,7 +322,7 @@ void FilterBench::Go() {
                       : FLAGS_quick ? quickTestModes : allTestModes;
 
   m_queries_ = FLAGS_m_queries;
-  uint32_t working_mem_size_mb = FLAGS_working_mem_size_mb;
+  double working_mem_size_mb = FLAGS_working_mem_size_mb;
   if (FLAGS_quick) {
     m_queries_ /= 7.0;
   } else if (FLAGS_best_case) {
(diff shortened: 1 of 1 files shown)


util/filter_bench.cc
```diff
@@ -491,8 +529,7 @@ double FilterBench::RandomQueryTest(uint32_t inside_threshold, bool dry_run,
 
   uint32_t num_infos = static_cast<uint32_t>(infos_.size());
   uint32_t dry_run_hash = 0;
-  uint64_t max_queries =
-      static_cast<uint64_t>(FLAGS_m_queries * 1000000 + 0.50);
+  uint64_t max_queries = static_cast<uint64_t>(m_queries_ * 1000000 + 0.50);
```
riversand963: It is possible that `m_queries_ * 1000000` overflows. Cast m_queries_ to uint64_t first.

util/filter_bench.cc
```diff
@@ -392,7 +430,7 @@ void FilterBench::Go() {
     std::cout << "Verifying..." << std::endl;
 
     uint32_t outside_q_per_f =
-        static_cast<uint32_t>(FLAGS_m_queries * 1000000 / infos_.size());
+        static_cast<uint32_t>(m_queries_ * 1000000 / infos_.size());
```
riversand963: Is this ever going to overflow?

util/filter_bench.cc
pdillinger: The real bug is accidentally declaring the new m_queries_ member as uint32_t instead of double like the command line parameter. (Sometimes I like to do fractions of millions of queries, but usually millions.)

And for whatever reason, neither our gcc or clang compilers complain about implicit narrowing conversion on `m_queries_ = FLAGS_m_queries;` 😠

util/filter_bench.cc
```diff
@@ -701,7 +738,11 @@ int main(int argc, char **argv) {
         << "\n      of queries." << std::endl;
   } else {
     FilterBench b;
-    b.Go();
+    for (uint32_t i = 0; i < FLAGS_runs; ++i) {
+      b.Go();
+      FLAGS_seed += 100;
```
riversand963: Nit: it is not recommended to change the value of `FLAGS_` variables, but this is truly minor.

PR #6446, opened 2020-02-22
Support direct IO in RandomAccessFileReader::MultiRead

cheng-chang:
Summary:
By supporting direct IO in RandomAccessFileReader::MultiRead, the benefits of parallel IO (IO uring) and direct IO can be combined.

In direct IO mode, read requests are aligned and merged together before being issued to RandomAccessFile::MultiRead, so blocks in the original requests might share the same underlying buffer, the shared buffers are returned in `aligned_bufs`, which is a new parameter of the `MultiRead` API.

For example, suppose alignment requirement for direct IO is 4KB, one request is (offset: 1KB, len: 1KB), another request is (offset: 3KB, len: 1KB), then since they all belong to page (offset: 0, len: 4KB), `MultiRead` only reads the page with direct IO into a buffer on heap, and returns 2 Slices referencing regions in that same buffer. See `random_access_file_reader_test.cc` for more examples.

Test Plan:
Added a new test `random_access_file_reader_test.cc`.


cheng-chang:
A future PR will build on top of this PR to enable direct IO in `BlockBasedTableReader::MultiGet`.

CMakeLists.txt
```diff
@@ -993,6 +993,7 @@ if(WITH_TESTS)
         env/env_test.cc
         env/mock_env_test.cc
         file/delete_scheduler_test.cc
+        file/random_access_file_reader_test.cc
```
cheng-chang: add to src.mk too

file/random_access_file_reader.cc
```diff
@@ -146,18 +147,77 @@ Status RandomAccessFileReader::Read(uint64_t offset, size_t n, Slice* result,
...
     StopWatch sw(env_, stats_, hist_type_,
                  (stats_ != nullptr) ? &elapsed : nullptr, true /*overwrite*/,
                  true /*delay_enabled*/);
     auto prev_perf_level = GetPerfLevel();
     IOSTATS_TIMER_GUARD(read_nanos);
 
+    FSReadRequest* fs_reqs = read_reqs;
+    size_t num_fs_reqs = num_reqs;
+#ifndef ROCKSDB_LITE
+    std::vector<FSReadRequest> aligned_reqs;
```
anand1976: Can we use ```rocksdb::autovector``` instead of ```std::vector```?

file/random_access_file_reader.h
```diff
@@ -22,6 +22,8 @@ namespace ROCKSDB_NAMESPACE {
 class Statistics;
 class HistogramImpl;
 
+using AlignedBuffers = std::vector<std::unique_ptr<const char[]>>;
```
anand1976: ```rocksdb::autovector``` instead of ```std::vector```

file/random_access_file_reader.cc
```diff
@@ -146,18 +147,77 @@ Status RandomAccessFileReader::Read(uint64_t offset, size_t n, Slice* result,
...
+      aligned_reqs.push_back(Align(read_reqs[0], alignment));
+      for (size_t i = 1; i < num_reqs; i++) {
+        const auto& r = Align(read_reqs[i], alignment);
+        if (!TryMerge(&aligned_reqs.back(), r)) {
+          aligned_reqs.push_back(r);
+        }
+      }
+      AlignedBuffer buf;
+      buf.Alignment(alignment);
+      for (auto& r : aligned_reqs) {
```
anand1976: Can we simplify even more by allocating one contiguous buffer for the combined length?

file/random_access_file_reader.cc
```diff
@@ -166,8 +226,27 @@ Status RandomAccessFileReader::MultiRead(FSReadRequest* read_reqs,
...
+      size_t aligned_i = 0;
+      for (size_t i = 0; i < num_reqs; i++) {
+        auto& r = read_reqs[i];
+        if (static_cast<size_t>(r.offset) > End(aligned_reqs[aligned_i])) {
+          aligned_i++;
+        }
+        const auto& fs_r = fs_reqs[aligned_i];
+        uint64_t offset = r.offset - fs_r.offset;
+        size_t len = std::min(r.len, static_cast<size_t>(fs_r.len - offset));
+        r.result = Slice(fs_r.scratch + offset, len);
```
anand1976: If status is non-ok, this should be set to empty Slice

env/composite_env_wrapper.h
```diff
@@ -521,8 +521,8 @@ class CompositeEnvWrapper : public Env {
   unsigned int GetThreadPoolQueueLen(Priority pri = LOW) const override {
     return env_target_->GetThreadPoolQueueLen(pri);
   }
-  Status GetTestDirectory(std::string* path) override {
-    return env_target_->GetTestDirectory(path);
+  Status GetTestDirectory(std::string* path, bool ignore_env = false) override {
```
cheng-chang: consider not setting the test dir env parameter to /dev/shm for random_access_file_reader_test  in Makefile, then get rid of `ignore_env`

anand1976: I like the option of not setting test dir env parameter in Makefile, or even calling ```setenv()``` in the test to clear ```TEST_TMPDIR```. Modifying the signature of ```GetTestDirectory()``` is a problem as there are custom env implementations outside the Github repo that'll break as a result.

file/random_access_file_reader.cc
cheng-chang: My concern is that the aligned blocks in `aligned_reqs` might be far away, assume 2 aligned blocks [0, 4KB], and [4000KB, 4004KB], in this case, if we allocate one contiguous buffer, it would be [0, 4004KB], then (4KB, 4000KB) is wasted.

file/random_access_file_reader.cc
cheng-chang: `fs_reqs` will be set to `aligned_reqs.data()` later, the pointer returned by `data()` should be the starting address of a contiguous address space. But `autovector` cannot have a `data()` method because if number of items is large, part of the data is stored on stack, others are stored in an internal `vector`, they don't constitute a contiguous space.

env/composite_env_wrapper.h
cheng-chang: got rid of `ignore_env` and used `unsetenv` to clear `TEST_TMPDIR`

cheng-chang:
@anand1976 thanks for comments, PTAL.

file/random_access_file_reader.cc
cheng-chang: reserved the space to `num_reqs`

file/random_access_file_reader.cc
anand1976: What I meant was can we allocate an 8KB buffer instead of 2 4KB buffers? Just track the total buffer space required and make one allocation.

cheng-chang:
@anand1976 thanks for the suggestion of allocating once, I applied that idea in this commit: https://github.com/facebook/rocksdb/pull/6446/commits/5f2cc7736fbd8a1f3128239e512163cafcb7d290

PR #6447, opened 2020-02-22
Introduce value type to merge operator

tabokie:
Primary goal of this PR is to support `kBlobIndex` type in merge operator. We add value type to merge operation input and output for that purpose, so that user can take value with different types and mutate value type through merge. 
Currently `kDeletion` is not supported as output type.


anand1976:
Cc @ltamasi Seems to be related to BlobDB.

yiwu-arbug:
Hi Levi, this is attempt to support merge operator BlobDB - at least make it to function. Do you mind take a look? Thanks.

include/rocksdb/merge_operator.h
```diff
@@ -74,18 +74,28 @@ class MergeOperator {
...
 
+  enum MergeValueType : unsigned char {
+    kDeletion = 0x0,
+    kValue = 0x1,
+    kBlobIndex = 0x2,
+  };
+
   struct MergeOperationInput {
-    explicit MergeOperationInput(const Slice& _key,
+    explicit MergeOperationInput(const Slice& _key, MergeValueType _value_type,
```
yiwu-arbug: Despite this changes the API, I think it is fine since only rocksdb itself will construct `MergeOperationInput`.

PR #6453, opened 2020-02-25
Return early on failure when constructing CuckooTableReader

cheng-chang:
Summary:
If file is not mmaped, CuckooTableReader should not try to read table properties from the file.

Test Plan:
Added a new unit test


cheng-chang:
appveyor test timed out, not failed.

PR #6454, opened 2020-02-25
support SstFileManager in db_stress

ajkr:
Summary: Add some flags to db_stress for configuring SstFileManager.

Test Plan:

- ran it a while:

```
$ python ./tools/db_crashtest.py blackbox --simple -max_key=100000 -write_buffer_size=131072 -target_file_size_base=131072 -max_bytes_for_level_base=524288 -value_size_mult=33 --interval=10 -max_background_compactions=4 -max_background_flushes=2 -sst_file_manager_bytes_per_sec=1048576
```

- verified with strace the SstFileManager is behaving as configured:

```
$ strace -fp `pidof db_stress` -e ftruncate,unlink
...
[pid 3074805]
ftruncate(9</tmp/rocksdb_crashtest_blackbox6OJywh/000070.sst.trash>,
67423) = 0
[pid 3074805]
ftruncate(9</tmp/rocksdb_crashtest_blackbox6OJywh/000070.sst.trash>,
51039) = 0
[pid 3074805]
ftruncate(9</tmp/rocksdb_crashtest_blackbox6OJywh/000070.sst.trash>,
34655) = 0
[pid 3074805]
ftruncate(9</tmp/rocksdb_crashtest_blackbox6OJywh/000070.sst.trash>,
18271) = 0
[pid 3074805]
ftruncate(9</tmp/rocksdb_crashtest_blackbox6OJywh/000070.sst.trash>,
1887) = 0
[pid 3074805]
unlink("/tmp/rocksdb_crashtest_blackbox6OJywh/000070.sst.trash") = 0
...
```

### DIFF
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -12,6 +12,7 @@
 #include "db_stress_tool/db_stress_common.h"
 #include "db_stress_tool/db_stress_driver.h"
 #include "rocksdb/convenience.h"
+#include "rocksdb/sst_file_manager.h"
 
 namespace ROCKSDB_NAMESPACE {
 StressTest::StressTest()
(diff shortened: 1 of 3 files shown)


db_stress_tool/db_stress_test_base.cc
```diff
@@ -1839,6 +1840,21 @@ void StressTest::Open() {
       options_.new_table_reader_for_compaction_inputs = true;
     }
   }
+  if (FLAGS_sst_file_manager_bytes_per_sec > 0 ||
```
riversand963: Just to clarify, if user does not specify these two newly-added flags, `options_` will not have an sst file manager created here, but `SanitizeOptions` will create an SstFileManager anyway with default values for these two knobs. Should we say this loud while defining these two flags?

ajkr: Yeah good point, I didn't notice it in `SanitizeOptions()` as I forgot the code organization lately. Will fix everything that implies `SstFileManager` is only created by setting these options.

PR #6457, opened 2020-02-27
Cache result of GetLogicalBufferSize in Linux

cheng-chang:
Summary:
In Linux, when reopening DB with many SST files, profiling shows that 100% system cpu time spent for a couple of seconds for `GetLogicalBufferSize`. This slows down MyRocks' recovery time when site is down.

This PR introduces two new APIs:
1. `Env::RegisterDbPaths` and `Env::UnregisterDbPaths` lets `DB` tell the env when it starts or stops using its database directories . The `PosixFileSystem` takes this opportunity to set up a cache from database directories to the corresponding logical block sizes.
2. `LogicalBlockSizeCache` is defined only for OS_LINUX to cache the logical block sizes. 

Other modifications:
1. rename `logical buffer size` to `logical block size` to be consistent with Linux terms.
2. declare `GetLogicalBlockSize` in `PosixHelper` to expose it to `PosixFileSystem`.
3. change the functions `IOError` and `IOStatus` in `env/io_posix.h` to have external linkage since they are used in other translation units too.

Test Plan:
1. A new unit test is added for `LogicalBlockSizeCache` in `env/io_posix_test.cc`.
2. A new integration test is added for `DB` operations related to the cache in `db/db_logical_block_size_cache_test.cc`.

`make check`


db/db_impl/db_impl_open.cc
```diff
@@ -1366,6 +1385,14 @@ Status DBImpl::Open(const DBOptions& db_options, const std::string& dbname,
     return s;
   }
 
+#ifdef OS_LINUX
+  auto db_paths = DbPaths(dbname, db_options, column_families);
+  s = FileSystem::Default()->HintDbPaths(db_paths);
```
siying: Can we always call env_->HintDbPaths(), and leave the implementation as empty?

env/io_posix.h
```diff
@@ -31,39 +36,81 @@
...
+  // result, otherwise, call GetLogicalBufferSize(fd) and cache the result.
+  // If multiple threads try to call GetLogicalBufferSize concurrently,
+  // only one active thread will do the actual work, other threads will return
+  // when the active thread returns and see the size cached by the active
+  // thread.
+  //
+  // 2. The file is not under any specified directory, the size is always
+  // computed by GetLogicalBufferSize.
+  size_t size(const std::string& fname, int fd) {
+    ReadLock lock(&cache_mutex_);
```
siying: I'm a little bit concerned about the potential overhead of the lock. Can you benchmark on a case where max_open_files is much smaller than number of files and see whether it impacts performance?

env/io_posix.h
```diff
@@ -31,39 +36,81 @@
...
+
+  // Logical buffer size of files under each specified directory will be cached.
+  // If the logical buffer size for a directory has been cached, the original
+  // cached size will keep being used.
+  void AddCacheDirectories(const std::unordered_set<std::string>& directories) {
+    WriteLock lock(&cache_mutex_);
+    for (const auto& d : directories) {
+      auto& s = cache_[d];
+      (void) s;
+    }
```
siying: Paths need to be removed when a DB is closed. I may have missed where it happens. Can you point me to the code?

db/db_impl/db_impl_open.cc
```diff
@@ -255,6 +255,25 @@ Status DBImpl::ValidateOptions(const DBOptions& db_options) {
...
+  for (const auto& p : opt.db_paths) {
+    paths.insert(p.path);
+  }
+  for (const auto& cfd : cfds) {
+    for (const auto& p : cfd.options.cf_paths) {
+      paths.insert(p.path);
+    }
+  }
+  return paths;
+}
```
siying: Do we handle DB::CreateColumnFamily() case somewhere?

env/io_posix.h
```diff
@@ -31,39 +36,81 @@
...
+      : get_logical_buffer_size_(get_logical_buffer_size) {}
+
+  // Logical buffer size of files under each specified directory will be cached.
+  // If the logical buffer size for a directory has been cached, the original
+  // cached size will keep being used.
+  void AddCacheDirectories(const std::unordered_set<std::string>& directories) {
+    WriteLock lock(&cache_mutex_);
+    for (const auto& d : directories) {
+      auto& s = cache_[d];
+      (void) s;
```
siying: Why don't we call get_logical_buffer_size_() here so that we don't need to call it in size()?

env/io_posix.h
```diff
@@ -31,39 +36,81 @@
...
+  // 1. The file is under a specified directory. If size is cached, return the
+  // result, otherwise, call GetLogicalBufferSize(fd) and cache the result.
+  // If multiple threads try to call GetLogicalBufferSize concurrently,
+  // only one active thread will do the actual work, other threads will return
+  // when the active thread returns and see the size cached by the active
+  // thread.
+  //
+  // 2. The file is not under any specified directory, the size is always
+  // computed by GetLogicalBufferSize.
+  size_t size(const std::string& fname, int fd) {
```
siying: I suggest we change it to a better name. Size() feels like number of entries in the cache.

Based on the naming convention we are following https://google.github.io/styleguide/cppguide.html#Function_Names , the first letter needs to be capital.

env/io_posix.h
cheng-chang: I was concerning of potentially large number of dbpaths, so get_logical_buffer_size_ was called on demand. But since number of dbpaths in practice is not huge, I updated the PR to call  get_logical_buffer_size_ when adding the dbpaths.

db/db_impl/db_impl_open.cc
cheng-chang: Added HintDbPaths to Env too

env/io_posix.h
cheng-chang: Updated to `GetLogicalBufferSize` to be more explicit

env/io_posix.h
cheng-chang: added logic to remove cached sizes

db/db_impl/db_impl_open.cc
cheng-chang: added logic in `DBImpl::CreateColumnFamilyImpl`

db/db_impl/db_impl.cc
```diff
@@ -562,6 +562,19 @@ Status DBImpl::CloseHelper() {
     delete txn_entry.second;
   }
 
+  std::unordered_set<std::string> db_paths;
```
siying: My 2 cents are to use std::set<> rather than unordered set. Most likely there is only one DB path.

db/db_impl/db_impl.cc
```diff
@@ -562,6 +562,19 @@ Status DBImpl::CloseHelper() {
...
+  db_paths.insert(dbname_);
+  for (auto& db_path : initial_db_options_.db_paths) {
+    db_paths.insert(db_path.path);
+  }
+  auto cfs = versions_->GetColumnFamilySet();
+  for (auto it = cfs->begin(); it != cfs->end(); ++it) {
+    for (auto& cf_path : (*it)->ioptions()->cf_paths) {
```
siying: For those auto in this patch, my vote is to spell out the type. Google C++ Style has a rule of auto which allows auto if it make the code clearer to readers: https://google.github.io/styleguide/cppguide.html#Type_deduction

I personally feels those auto makes it harder for me to read, rather than easier.

db/db_impl/db_impl.cc
```diff
@@ -2425,6 +2440,11 @@ Status DBImpl::DropColumnFamilyImpl(ColumnFamilyHandle* column_family) {
     assert(cfd->IsDropped());
     ROCKS_LOG_INFO(immutable_db_options_.info_log,
                    "Dropped column family with id %u\n", cfd->GetID());
+    std::unordered_set<std::string> paths;
+    for (auto& cf_path : cfd->ioptions()->cf_paths) {
+      paths.insert(cf_path.path);
+    }
+    s = env_->HintDbPathsRemoved(paths);
```
siying: I'm not sure it is a good idea. Different CFs can share the same path. We don't want to remove them unless it is used by another column family.

env/io_posix.cc
```diff
@@ -409,13 +370,112 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
...
+  WriteLock lock(&cache_mutex_);
+  for (auto& dir : directories) {
+    cache_.erase(dir);
+  }
+}
+
+size_t LogicalBufferSizeCache::GetLogicalBufferSize(const std::string& fname,
+                                                    int fd) {
+  ReadLock lock(&cache_mutex_);
+  std::string dir = fname.substr(0, fname.find_last_of("/"));
```
siying: Can we move it before read lock?

env/io_posix.cc
```diff
@@ -409,13 +370,112 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
...
+  if (cache_.find(dir) != cache_.end()) {
+    return cache_[dir];
```
siying: The result should already fetched by `cache_.find(dir)`. No need to recheck it.

env/io_posix.cc
```diff
@@ -409,13 +370,112 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
...
+  }
+  return get_logical_buffer_size_(fd);
```
siying: Can we move it outside the read lock?

env/io_posix.cc
cheng-chang: Also made the critical section less expensive in CacheLogicalBufferSize

db/db_impl/db_impl.cc
cheng-chang: good point, the cost of hashing may not be worth it.

env/io_posix.cc
cheng-chang: I thought mostly there is no read-write contention, so the critical section for the read lock won't cause performance drop, and leveraging `ReadLock` at the beginning makes the code cleaner.

But I'm ok with taking extra effort to make all critical sections more dense.

db/db_impl/db_impl.cc
```diff
@@ -2402,6 +2417,7 @@ Status DBImpl::DropColumnFamilyImpl(ColumnFamilyHandle* column_family) {
                                     mutable_cf_options->max_write_buffer_number;
     }
 
+    bool cf_support_snapshot = cfd->mem()->IsSnapshotSupported();
```
siying: Is the logic related to dropping column family removed?
I know it might be challenging to handle drop column family, but it might need to be handled.

cheng-chang: Yes, I'm still thinking about how to only remove the uniquely owned cf_paths.

db/db_impl/db_impl.cc
cheng-chang: Removed the logic temporarily, will think about how to do it efficiently other than looping through all cf_paths to check conflict.

env/fs_posix.cc
```diff
@@ -833,7 +845,24 @@ class PosixFileSystem : public FileSystem {
...
+  Status OnDbPathsAdded(const std::set<std::string>& paths) override {
+    std::set<std::pair<std::string, int>> fname_fds;
+    for (auto& path : paths) {
+      int fd = open(path.c_str(), O_DIRECTORY | O_RDONLY);
+      if (fd == -1) {
+        return Status::IOError("Cannot open directory " + path);
+      }
+      fname_fds.emplace(path, fd);
+    }
+    logical_buffer_size_cache_.CacheLogicalBufferSize(fname_fds);
```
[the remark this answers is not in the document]
cheng-chang: Yes, I tested [this gist](https://gist.github.com/cheng-chang/7956a11c8f672fcbe560418a9a651c40) on a Linux server.

db/db_impl/db_impl.cc
```diff
@@ -562,6 +561,18 @@ Status DBImpl::CloseHelper() {
     delete txn_entry.second;
   }
 
+  std::set<std::string> db_paths;
+  db_paths.insert(dbname_);
+  for (const DbPath& db_path : initial_db_options_.db_paths) {
+    db_paths.insert(db_path.path);
+  }
+  ColumnFamilySet* cfs = versions_->GetColumnFamilySet();
+  for (ColumnFamilySet::iterator it = cfs->begin(); it != cfs->end(); ++it) {
```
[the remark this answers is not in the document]
cheng-chang: changed to `for (ColumnFamilyData* cf : *versions_->GetColumnFamilySet())` since `GetColumnFamilySet` returns a pointer.

db/db_impl/db_impl.cc
cheng-chang: `DropColumnFamilyImpl` might not be the most appropriate place to call `OnDbPathsRemoved`, because even when CF is dropped, as long as the user hasn't deleted `ColumnFamilyHandle`, the user might still be reading data from the CF through the handle.

cheng-chang: I think `~ColumnFamilyData` is a better place to call `OnDbPathsRemoved`, because when `~ColumnFamilyData` is called, it means there is no `ColumnFamilyHandle` referencing to the CFD, so there will be no user reading from the SST files in the CF's cf_paths any more.

siying: That's a good point.

cheng-chang:
Since CFs might share data paths, when a CF is destroyed, the data path might still be used by other CFs. I added reference counting to the cache to only remove the cached value when the data path's reference count reaches 0.

cheng-chang:
@siying PTAL

env/io_posix.cc
```diff
@@ -409,13 +370,135 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
   return static_cast<size_t>(rid - id);
 }
 #endif
+
+#ifdef OS_LINUX
+void LogicalBufferSizeCache::RefAndCacheLogicalBufferSize(
+    const std::set<std::pair<std::string, int>>& directories) {
+  std::map<int, size_t> fd_sizes;
```
cheng-chang: Some applications will `CreateColumnFamily` and `DropColumnFamily` frequently, but I assume the `cf_paths` for the `CreateColumnFamily` would not be too different, so I made the optimization here to only acquire write lock when there are new paths.

env/io_posix.cc
```diff
@@ -409,13 +370,135 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
...
+      cache_.erase(it);
+    }
+  }
+}
+
+size_t LogicalBufferSizeCache::GetLogicalBufferSize(const std::string& fname,
+                                                    int fd) {
+  std::string dir = fname.substr(0, fname.find_last_of("/"));
+  {
+    ReadLock lock(&cache_mutex_);
```
cheng-chang: I think we can even get rid of the read lock if we maintain multiple versions of the map, and always read from a consistent snapshot, kind of like mvcc. But since cache is read heavy, I think the read lock here is fine. I'll post a benchmark result from db_bench later.

include/rocksdb/env.h
```diff
@@ -163,6 +164,33 @@ class Env {
...
+  // so OnDbPathsRegistered might be called with the same data paths.
+  // For example, when CreateColumnFamily is called multiple times with the same
+  // data path, OnDbPathsRegistered will also be called with the same data path.
+  //
+  // When a path is called in OnDbPathsRegistered, it must be correspondingly
+  // called in OnDbPathsUnregistered.
+  //
+  // Different implementations may take different actions.
+  // By default, it's a no-op.
+  virtual Status OnDbPathsRegistered(const std::set<std::string>& /*paths*/) {
```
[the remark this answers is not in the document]
cheng-chang: agree, maybe in the future, there are certain handlers that assume the paths can be duplicated.

env/io_posix.cc
```diff
@@ -409,13 +370,135 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
...
+
+#ifdef OS_LINUX
+void LogicalBufferSizeCache::RefAndCacheLogicalBufferSize(
+    const std::set<std::pair<std::string, int>>& directories) {
+  std::map<int, size_t> fd_sizes;
+  {
+    ReadLock lock(&cache_mutex_);
+    for (auto& fname_fd : directories) {
+      if (cache_.find(fname_fd.first) == cache_.end()) {
+        fd_sizes.emplace(fname_fd.second, 0);
```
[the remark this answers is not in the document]
cheng-chang: I chose `emplace` because I think `fb_sizes[fname_fd.second]` alone already constructs and inserts the `{fname_fd.second, 0}` pair, then ` = 0` is an extra copy, I want to avoid that extra copy.

db/db_impl/db_impl.cc
```diff
@@ -3597,6 +3609,14 @@ Status DestroyDB(const std::string& dbname, const Options& options,
     soptions.sst_file_manager.reset();
 
     env->DeleteDir(dbname);  // Ignore error in case dir contains other files
+
+    paths.insert(dbname);
+    Status s = env->OnDbPathsUnregistered(paths);
```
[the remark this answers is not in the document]
cheng-chang: make sense to me

env/io_posix.cc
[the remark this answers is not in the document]
siying: I am not expert on it but I feel that assigning an integer has a very minor cost. Maybe ask someone who is more familiar with it. But it's your call here.

include/rocksdb/env.h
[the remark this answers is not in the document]
cheng-chang: According to the method's doc:  `Handles the event when a new DB or a new ColumnFamily starts using the specified data paths.` Using a `set` here enforces that the data paths should not be duplicated in the parameter.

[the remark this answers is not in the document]
cheng-chang: @siying  If it's a more general API like "OnPathsRegistered", then I think it's more natural for the paths to be a `vector`. I'm ok with making this API more general, what do you think?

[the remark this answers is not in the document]
cheng-chang: After second thought, I'm ok with making it a `vector`, then also in `DBImpl`, we do not deduplicate paths via `set`, we always respect what users have set in `dbname`, `db_paths` and `cf_paths`.

env/io_posix.cc
siying: That might be too hard to maintain. I hope performacne with this ReadLock is enough.

include/rocksdb/env.h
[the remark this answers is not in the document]
siying: It makes sense. You can decide.

db/db_impl/db_impl.cc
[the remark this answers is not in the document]
siying: I'm confused. Is the code here from DestroyDB()?

include/rocksdb/env.h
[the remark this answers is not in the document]
cheng-chang: Considering a case: Open a db with specified 2 `ColumnFamilyDescriptor`s which share a `cf_path` , then if using a `set` of paths, only 1 path and 1 reference to that path is added to the cache. Later, `DropColumnFamily` on one of the column, then the cached path will be evicted since ref count reaches 0, but there is still another column using the path, so the cached size should be kept there.
Using a `vector` of paths and do not deduplicate them can solve this problem.

db/db_impl/db_impl.cc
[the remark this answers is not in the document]
cheng-chang: Yes, I haven't updated the PR, will be removed.

include/rocksdb/env.h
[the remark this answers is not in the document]
siying: I didn't suggest that DBImpl should pass duplicated value with std::vector<>. We can dedup in DBImpl and pass a std::vector<>. But you have a point that, as a general API, users of the API could pass duplicated value and we need to handle it. So it's your call.

db/db_impl/db_impl.cc
cheng-chang: With the new reference counting mechanism in the cache, each column family handle will increase the reference count, so unregister is changed to be called in `~ColumnFamilyHandle` instead of `~ColumnFamilyData` .

include/rocksdb/env.h
```diff
@@ -1184,6 +1184,14 @@ class EnvWrapper : public Env {
   Env* target() const { return target_; }
 
   // The following text is boilerplate that forwards all methods to target()
+  Status OnDbPathsRegistered(const std::vector<std::string>& paths) override {
```
siying: Do we need "On" in the function name? I think we use "on" for listener callbacks, which indicates that the function is called while the operation happens outside the function, while in this case, registration happens inside the function. Can we call it `RegisterDirectories()` or something like that?

db/db_logical_block_size_cache_test.cc
```diff
@@ -0,0 +1,366 @@
...
+    ASSERT_EQ(11, cache_->GetLogicalBlockSize(data_path_1_ + "/sst", 200));
+    ASSERT_EQ(2, ncall_);
+    ASSERT_OK(db->Close());
+    ASSERT_EQ(2, ncall_);
+    // Logical block size of data paths are removed from cache during Close.
+    ASSERT_EQ(100, cache_->GetLogicalBlockSize(data_path_0_ + "/sst", 100));
+    ASSERT_EQ(3, ncall_);
+    ASSERT_EQ(200, cache_->GetLogicalBlockSize(data_path_1_ + "/sst", 200));
+    ASSERT_EQ(4, ncall_);
+    delete db;
```
siying: Can we assert the the cache is empty after deleting the DB?

db/db_logical_block_size_cache_test.cc
```diff
@@ -0,0 +1,366 @@
...
+  // Logical block size of dbname_ is cached during Open.
+  ASSERT_EQ(1, ncall_);
+  ASSERT_EQ(0, cache_->GetLogicalBlockSize(dbname_ + "/sst", 100));
+  ASSERT_EQ(1, ncall_);
+
+  delete db;
+  ASSERT_EQ(1, ncall_);
+  // Logical block size of dbname_ is removed from cache during Close.
+  ASSERT_EQ(100, cache_->GetLogicalBlockSize(dbname_ + "/sst", 100));
+  ASSERT_EQ(2, ncall_);
```
siying: Can we validate in a more straight-forward way: adding a function in cache that determine whether it is empty, or return number of paths in the cache?

db/db_logical_block_size_cache_test.cc
```diff
@@ -0,0 +1,366 @@
...
+  ASSERT_EQ(3, ncall_);
+
+  // Delete handle will drop cache.
+  ASSERT_OK(db->DestroyColumnFamilyHandle(cf));
+  ASSERT_EQ(100, cache_->GetLogicalBlockSize(cf_path_0_ + "/sst", 100));
+  ASSERT_EQ(4, ncall_);
+  ASSERT_EQ(200, cache_->GetLogicalBlockSize(cf_path_0_ + "/sst", 200));
+  ASSERT_EQ(5, ncall_);
+
+  delete db;
```
siying: Same here validate cache emptiness.

db/db_logical_block_size_cache_test.cc
```diff
@@ -0,0 +1,366 @@
...
+    ASSERT_OK(db->DestroyColumnFamilyHandle(cfs[0]));
+    ASSERT_EQ(20, cache_->GetLogicalBlockSize(cf_path_0_ + "/sst1", 200));
+    ASSERT_EQ(2, ncall_);
+    ASSERT_OK(db->DestroyColumnFamilyHandle(cfs[1]));
+    ASSERT_EQ(0, cache_->GetLogicalBlockSize(dbname_ + "/sst1", 201));
+    ASSERT_EQ(2, ncall_);
+
+    delete db;
+  }
+}
```
siying: I know multiple DB test is harder to write, but if you are not 100% sure about the cache when dealing with multiple DBs, I encourage you to write one test case for multiple DBs.

include/rocksdb/env.h
cheng-chang: Sounds good, I thought of them as event handlers, I changed the name to `RegisterDbPaths` and `UnregisterDbPaths`.

db/db_logical_block_size_cache_test.cc
cheng-chang: Added an `IsEmpty()` method to the cache.

cheng-chang: And asserts that cache IsEmpty after every `delete db`

cheng-chang: Added `Size()` to return number of cached directories

env/io_posix.cc
```diff
@@ -409,13 +370,178 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
...
+    close(fd);
+    return Status::IOError("Cannot open directory " + directory);
+  }
+  *size = PosixHelper::GetLogicalBlockSizeOfFd(fd);
+  close(fd);
+  return Status::OK();
+}
+
+size_t PosixHelper::GetLogicalBlockSizeOfFd(int __attribute__((__unused__))
+                                            fd) {
```
siying: Instead of having an attribute, you can also in non OS_LINUX case, call `(void) fd`.

include/rocksdb/file_system.h
```diff
@@ -171,6 +171,17 @@ class FileSystem {
   // The result of Default() belongs to rocksdb and must never be deleted.
   static std::shared_ptr<FileSystem> Default();
 
+  // See Env::OnDbPathsRegistered.
+  virtual Status OnDbPathsRegistered(
```
siying: Rename this too?

include/rocksdb/file_system.h
```diff
@@ -171,6 +171,17 @@ class FileSystem {
   // The result of Default() belongs to rocksdb and must never be deleted.
   static std::shared_ptr<FileSystem> Default();
 
+  // See Env::OnDbPathsRegistered.
```
siying: Can we do it the other way around? Comment here, and write en Env.h to see it here? FileSystem is the newer API and we should encourage users to implement this instead.

include/rocksdb/env.h
```diff
@@ -163,6 +163,33 @@ class Env {
...
+  // Handles the event when a new DB or a new ColumnFamily starts using the
+  // specified data paths.
+  //
+  // The data paths might be shared by different DBs or ColumnFamilies,
+  // so OnDbPathsRegistered might be called with the same data paths.
+  // For example, when CreateColumnFamily is called multiple times with the same
+  // data path, OnDbPathsRegistered will also be called with the same data path.
+  //
+  // When a path is called in OnDbPathsRegistered, it must be correspondingly
+  // called in OnDbPathsUnregistered.
```
siying: Change name in comments too.

include/rocksdb/env.h
```diff
@@ -163,6 +163,33 @@ class Env {
...
+  //
+  // Different implementations may take different actions.
+  // By default, it's a no-op.
```
siying: I think it is better to document here the behavior when Status returned is not OK. I assume what we want is that, if it is not OK, nothing should change and users don't have to call UnregisterDbPaths() to unregister them?

db/db_logical_block_size_cache_test.cc
cheng-chang: added tests for situations for multiple DBs.

db/db_logical_block_size_cache_test.cc
```diff
@@ -0,0 +1,366 @@
...
+    DB* db;
+    if (!i) {
+      printf("Open\n");
+      ASSERT_OK(DB::Open(options, dbname_, &db));
+    } else {
+      printf("OpenForReadOnly\n");
+      ASSERT_OK(DB::OpenForReadOnly(options, dbname_, &db));
+    }
+    // Logical block size of data paths are cached during Open.
+    ASSERT_EQ(2, ncall_);
```
cheng-chang: After second thought, for this test, it's sufficient to check what directories are cached. Checking `ncall_` to verify which directory is cached is indirect, instead, I added `Size` and `Contains` to the cache to check the content of the cache directly. 

From DB's perspective, if I can verify that a directory is cached, then I don't need to verify whether `GetLogicalBlockSize` on that cached directory will increase `ncall_` because that's ensured by the cache's implementation, and is already verified in the cache's unit test in `io_posix_test.cc`.

db/db_logical_block_size_cache_test.cc
cheng-chang: added two more tests regarding multiple DBs.

cheng-chang:
@siying I just found a much cleaner way to call `RegisterDbPaths` and `UnregisterDbPaths`. Previously, we call these methods in DB methods such as `Open`, `CreateColumnFamily`, etc. Now, `RegisterDbPaths` is called when constructing `ColumnFamilyData`, and `UnregisterDbPaths` is called in `~ColumnFamilyData`. With this approach, all DB implementations such as `DBReadOnly`, `DBSecondary` support the register and unregister semantic automatically.

env/io_posix.h
cheng-chang: [Here](https://gist.github.com/cheng-chang/11bf2f255c0769dd4890dffdfe7c8e79) is a benchmark to show that the read lock in the cache is not a concern.

cheng-chang:
[Here](https://gist.github.com/cheng-chang/11bf2f255c0769dd4890dffdfe7c8e79) is a benchmark to show that the read lock in the cache is not a concern.

env/io_posix.h
cheng-chang: ran a benchmark, result is [here](https://gist.github.com/cheng-chang/11bf2f255c0769dd4890dffdfe7c8e79).

db/column_family.cc
```diff
@@ -471,6 +472,17 @@ void SuperVersionUnrefHandle(void* ptr) {
...
+std::vector<std::string> ColumnFamilyData::GetDbPaths() const {
+  std::vector<std::string> paths;
+  paths.reserve(ioptions_.cf_paths.size());
+  for (const DbPath& db_path : ioptions_.cf_paths) {
+    paths.emplace_back(db_path.path);
+  }
+  return paths;
+}
+
+const uint32_t ColumnFamilyData::kDummyColumnFamilyDataId = UINT32_MAX;
```
siying: port::kkMaxUint32

db/column_family.cc
```diff
@@ -507,7 +519,19 @@ ColumnFamilyData::ColumnFamilyData(
       queued_for_compaction_(false),
       prev_compaction_needed_bytes_(0),
       allow_2pc_(db_options.allow_2pc),
-      last_memtable_id_(0) {
+      last_memtable_id_(0),
+      db_paths_registered_(false) {
+  if (id_ != kDummyColumnFamilyDataId) {
+    Status s = ioptions_.env->RegisterDbPaths(GetDbPaths());
```
siying: If possible, avoid potentially expensive operation like this in side constructors. But I assume that you already weighed in those considerations and feel this is the cleanest way we can get.

Do we have access to FileSystem object here? If we have, it will be a good idea to call FileSystem rather than Env here.

db/column_family.cc
```diff
@@ -601,6 +625,16 @@ ColumnFamilyData::~ColumnFamilyData() {
   for (MemTable* m : to_delete) {
     delete m;
   }
+
+  if (db_paths_registered_) {
+    Status s = ioptions_.env->UnregisterDbPaths(GetDbPaths());
```
siying: Same here. It will be ideal to call FileSystem rather than Env, as it is the new API.

env/fs_posix.cc
```diff
@@ -833,7 +839,15 @@ class PosixFileSystem : public FileSystem {
...
     return optimized;
   }
-
+#ifdef OS_LINUX
+  Status RegisterDbPaths(const std::vector<std::string>& paths) override {
+    return logical_block_size_cache_.RefAndCacheLogicalBlockSize(paths);
+  }
+  Status UnregisterDbPaths(const std::vector<std::string>& paths) override {
+    logical_block_size_cache_.UnrefAndTryRemoveCachedLogicalBlockSize(paths);
+    return Status::OK();
```
siying: Can we avoid the return code? In that way we don't have to handle the hypothesis case that it fails in ColumnFamilyData's destructor.

cheng-chang: Since `UnregisterDbPaths` is a general API, in case other implementations might need to return a status, I prefer to keep the status here.

db/column_family.cc
cheng-chang: This is the cleanest way, the downside is that this might be called with db mutex held. In `Open`, it's not big deal since no other threads will start writing during Open, but for dynamically creating column families, this will bring performance down when there are a lot of data paths for a CF.

cheng-chang: replaced env with fs.

cheng-chang: I cannot find a cleaner way right now, but I'll try to remove the RegisterDbPaths call from the code path with db mutex held in another PR.

db/column_family.cc
cheng-chang: Some tests rely on `EnvWrapper` right now, using `fs` here would break those tests.

Also, I find that most of the code now still uses `env` as a public API, `FileSystem` is more like a subset or implementation detail of `Env`.

cheng-chang: left a TODO in the code

db/column_family.cc
cheng-chang: left a TODO in the code

PR #6468, opened 2020-02-29
Replace Directory with FSDirectory in DB

zhichao-cao:
In the current code base, we can use Directory from Env to manage directory (e.g, Fsync()). The PR #5761  introduce the File System as a new Env API. So we further replace the Directory class in DB with FSDirectory such that we can have more IO information from IOStatus returned by FSDirectory.


zhichao-cao:
> LGTM. Thanks for the PR!

Thanks for the review!

PR #6479, opened 2020-03-04
build_tools/build_detect_platform: fix C++ tests

ffontaine:
Replace `-o /dev/null` by `-o test.o` when testing for C++ features such as
-faligned-new otherwise tests will fail with some bugged binutils
(https://sourceware.org/bugzilla/show_bug.cgi?id=19526):

```
output/host/bin/xtensa-buildroot-linux-uclibc-g++ -faligned-new -x c++ - -o /dev/null <<EOF
            struct alignas(1024) t {int a;};
            int main() {}
EOF
/home/fabrice/buildroot/output/host/lib/gcc/xtensa-buildroot-linux-uclibc/8.3.0/../../../../xtensa-buildroot-linux-uclibc/bin/ld: final link failed: file truncated

```
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


riversand963:
According to https://sourceware.org/bugzilla/show_bug.cgi?id=19526, the bug motivating this PR has been fixed for 5 years.

ffontaine:
There is still embedded toolchains out in the field that are affected by this bug.

For example, this patch was needed on buildroot to fix build failures on nios2 and xtensa: https://git.buildroot.net/buildroot/commit/package/rocksdb?id=1c85b1d63eb9c9536188c3767caffcf74a6e220e. 

But perhaps, I should disable rocksdb on those architectures?

riversand963:
@ffontaine thanks for letting us know. Just curious any reason why the fix has not been picked up?

ffontaine:
[buildroot](https://buildroot.org/) is a tool to generate embedded Linux systems through cross-compilation. We have autobuilders running tests with a wide variety of toolchains. However some of them are just publicly retrieved and we have no way to patch them.

riversand963:
Thanks @ffontaine for the info. I will keep this open then and see if we can resolve it in the near future. It will be nice to make sure RocksDB builds on multiple platforms unless we explicitly say we do not support the said platforms, which we currently don't.

riversand963:
@ffontaine can you try rebasing this PR?

ffontaine:
OK, I'll try to rebase it.

riversand963:
@ffontaine I tested this, and the change mostly LGTM.
One minor thing, can you cleanup the test.o instead of relying on `make clean`?

riversand963 (approved):

riversand963:
Thanks @ffontaine for the fix

Issue #6276, opened 2020-01-09
RockDB plugin build in MariaDB is not reproducible


ottok:
Reproducible builds are important for software supply chain security. See https://reproducible-builds.org/

All packages in Debian are tested for reproducibility. Currently the latest MariaDB 10.3 build in Debian fails due to the build of the Mroonga, RocksDB and TokuDB plugins. See https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/mariadb-10.3.html

Based on the diffoscope data is seems RocksDB is embedding a random build ID in every build:
![image](https://user-images.githubusercontent.com/668724/72098881-127b4b80-3328-11ea-9323-90b0593f5d2e.png)

Could you please not do that, or alternative provide a build flag to turn it off?

Greetings,
MariaDB maintainer in Debian

anand1976:
Can you set the NO_UPDATE_BUILD_VERSION env variable and see if it helps?

ottok:
I added this env to the build of mariadb-10.3 in Debian in: https://salsa.debian.org/mariadb-team/mariadb-10.3/commit/5eb8f505b9861f0d374663d8f2c857c8b5d1cd09

Unfortunately this did not seem to solve it, nor can I even see in the output if it helped at all:
https://user-images.githubusercontent.com/668724/73429956-89af6880-4345-11ea-9976-36cdb9727d0b.png

ottok:
Why was this issue closed? Is RocksDB at all in any release reproducible somewhere? I suggest to keep this open until somebody manages to build it so that reproducibility is proven.

ottok:
Did you consider the text here?
![image](https://user-images.githubusercontent.com/668724/73599403-75d15580-4543-11ea-8439-273774c9894d.png)

`​rocksdb_build_git_date:​@2020-​01-​28·​08:​12:​28@`

Can this be disabled somehow?

ottok:
Why was this issue closed? Is RocksDB at all in any release reproducible somewhere? I suggest to keep this open until somebody manages to build it so that reproducibility is proven.

mdcallag:
AFAIK it was closed after you didn't reply to [a question](https://github.com/facebook/rocksdb/issues/6276#issuecomment-574370196)

PR #6481, opened 2020-03-04
Fix spelling: commited -> committed

ottok:
In most places in the code the variable names are spelled correctly as
COMMITTED but in a couple places not. This fixes them and ensures the
variable is always called COMMITTED everywhere.

### DIFF
--- a/include/rocksdb/utilities/transaction.h
+++ b/include/rocksdb/utilities/transaction.h
@@ -491,7 +491,8 @@ class Transaction {
     AWAITING_PREPARE = 1,
     PREPARED = 2,
     AWAITING_COMMIT = 3,
-    COMMITED = 4,
+    COMMITTED = 4,
+    COMMITED = COMMITTED, // old misspelled name
     AWAITING_ROLLBACK = 5,
     ROLLEDBACK = 6,
     LOCKS_STOLEN = 7,
--- a/java/src/main/java/org/rocksdb/Transaction.java
+++ b/java/src/main/java/org/rocksdb/Transaction.java
@@ -1797,7 +1797,7 @@ public class Transaction extends RocksObject {
      * Keep old misspelled variable as alias
      * Tip from https://stackoverflow.com/a/37092410/454544
      */
-    public static final Browser COMMITED=COMMITTED;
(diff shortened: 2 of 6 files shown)


ottok:
BTW, please re-open https://github.com/facebook/rocksdb/issues/6276 as the issue itself is still unfixed just like it was then the issue was opened, there was no logical reason to close the issue.

include/rocksdb/utilities/transaction.h
```diff
@@ -491,7 +491,7 @@ class Transaction {
     AWAITING_PREPARE = 1,
     PREPARED = 2,
     AWAITING_COMMIT = 3,
-    COMMITED = 4,
+    COMMITTED = 4,
```
pdillinger: This is a public API, so I think it's worth keeping the old name as an alias for the new name, which IIRC can be done with `COMMITED = COMMITTED, // old misspelled name`

java/src/main/java/org/rocksdb/Transaction.java
```diff
@@ -1788,7 +1788,7 @@ public long getId() {
     AWAITING_PREPARE((byte)1),
     PREPARED((byte)2),
     AWAITING_COMMIT((byte)3),
-    COMMITED((byte)4),
+    COMMITTED((byte)4),
```
pdillinger: This is also a public API, so let's do our best to keep the old name as an alias for the new name. It's not 100% possible for Java enums, but close enough, IMHO: https://stackoverflow.com/a/37092410/454544

include/rocksdb/utilities/transaction.h
ottok: Added

java/src/main/java/org/rocksdb/Transaction.java
ottok: Added

ottok:
Updated and force pushed new version with aliases for old COMMITED and a pending change in HISTORY.md.

ottok:
.. and `allow edits from maintainers` is selected in case you want to further edit this in the same commit, but you can also iterate it more in later commits.

PR #6487, opened 2020-03-06
Pass IOStatus to write path and set retryable IO Error as hard error in BG jobs

zhichao-cao:
In the current code base, we use Status to get and store the returned status from the call. Specifically, for IO related functions, the current Status cannot reflect the IO Error details such as error scope, error retryable attribute, and others. With the implementation of #5761, we have the new Wrapper for IO, which returns IOStatus instead of Status. However, the IOStatus is purged at the lower level of write path and transferred to Status.

The first job of this PR is to pass the IOStatus to the write path (flush, WAL write, and Compaction). The second job is to identify the Retryable IO Error as HardError, and set the bg_error_ as HardError. In this case, the DB Instance becomes read only. User is informed of the Status and need to take actions to deal with it (e.g., call db->Resume()).

Test plan: Added the testing case to error_handler_fs_test. Pass make asan_check


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -405,6 +417,20 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
     s = error_status.ok() ? s : error_status;
   }
 
+  if (io_s.ok()) {
+    IOStatus io_error = IOStatus::OK();
+    for (int i = 0; i != num_cfs; i++) {
```
anand1976: Nit: Can we use ```io_status.size()``` instead of ```num_cfs```?

db/db_impl/db_impl_write.cc
```diff
@@ -823,6 +840,17 @@ void DBImpl::WriteStatusCheck(const Status& status) {
   }
 }
 
+void DBImpl::IOStatusCheck(const IOStatus& io_status) {
```
anand1976: It could be in another PR, but I believe that ```IOStatusCheck()``` and ```WriteStatusCheck()``` should propagate the status returned by ```SetBGError``` to ```w.status```.

db/error_handler.cc
```diff
@@ -238,6 +238,33 @@ Status ErrorHandler::SetBGError(const Status& bg_err, BackgroundErrorReason reas
...
+  if (bg_io_err.ok()) {
+    return Status::OK();
+  }
+  if (recovery_in_prog_ && recovery_error_.ok()) {
+    recovery_error_ = bg_io_err;
+  }
+  Status new_bg_io_err = bg_io_err;
+  Status s;
+  // First, check if the error is a retryable IO error or not.
+  if (bg_io_err.GetRetryable()) {
```
anand1976: We should check for ```bg_io_err.GetDataLoss()``` also and set severity to ```kUnrecoverableError```.

db/error_handler.cc
```diff
@@ -238,6 +238,33 @@ Status ErrorHandler::SetBGError(const Status& bg_err, BackgroundErrorReason reas
...
+    // In current stage, treat retryable error as HardError. No automatic
+    // recovery.
+    bool auto_recovery = false;
+    Status bg_err(new_bg_io_err, Status::Severity::kHardError);
+    bg_error_ = bg_err;
```
anand1976: We might end up overwriting a more severe error with a less severe one. Check the current severity of bg_error_ before overwriting.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -359,14 +366,17 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
     // TODO (yanqin) investigate whether we should sync the closed logs for
     // single column family case.
     s = SyncClosedLogs(job_context);
+    s = io_s;
```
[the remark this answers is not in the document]
zhichao-cao: @anand1976 Yes, missed here.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -206,8 +208,12 @@ Status DBImpl::FlushMemTableToOutputFile(
   }
 
   if (!s.ok() && !s.IsShutdownInProgress() && !s.IsColumnFamilyDropped()) {
```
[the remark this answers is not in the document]
zhichao-cao: @anand1976 Sure

db/error_handler.cc
zhichao-cao: @anand1976 I see, will change

db/error_handler.cc
zhichao-cao: @anand1976 I see. I missed this possibility. Thanks!

db/compaction/compaction_job.cc
```diff
@@ -713,9 +718,13 @@ Status CompactionJob::Install(const MutableCFOptions& mutable_cf_options) {
   cfd->internal_stats()->AddCompactionStats(
       compact_->compaction->output_level(), thread_pri_, compaction_stats_);
 
+  versions_->SetIOStatusOK();
```
anand1976: Could this be called once in DBImpl::ResumeImpl() instead of at the start of every compaction/flush?

rohansuri:
Hi @zhichao-cao, 
I wanted to understand this change better. It seems that autorecovery in case of `ENOSPC` errors using`SstFileManager.StartErrorRecovery` won't be done now. What is the reason behind turning off this autorecovery?

zhichao-cao:
> ENOSPC

@rohansuri Correct me If I understand your answer in a wrong direction. Currently, ENOSPC is auto recovered in the old SetBGError logic (https://github.com/facebook/rocksdb/blob/master/db/error_handler.cc#L235). The new SetBGError interface is only for the specific IO Status and this function only handles the Retryable error (map it to hard error). No space error will be pass to the old SetBGError to handle.

rohansuri:
Thanks for your reply @zhichao-cao. 

> No space error will be pass to the old SetBGError to handle

I think in case of `kFlush` background errors, `ENOSPC` won't be passed to old `SetBGError`, but rather to the new one which won't trigger the auto recovery using `SstFileManager.StartErrorRecovery`. I haven't written a test to verify this or looked at existing ones if they check this path. But from the looks of it, it seems valid. Apologies if that's not the case.

Consider the code path (function stack downwards, but might be easier to read bottom up):
* DBImpl::BackgroundFlush
* DBImpl::FlushMemTablesToOutputFiles
* [DBImpl::FlushMemTableToOutputFile](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/db/db_impl/db_impl_compaction_flush.cc#L196) takes the `FlushJob.io_status_` and calls the new [ErrorHandle::SetBGError](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/db/db_impl/db_impl_compaction_flush.cc#L213). The new logic will handle `ENOSPC` as `kHardError` since the `retryable_` flag is true for [IOStatus:NoSpace()](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/env/io_posix.cc#L63).
* FlushJob::Run
* [FlushJob::WriteLevel0Table](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/db/flush_job.cc#L398) sets the `FlushJob.io_status_`.
* [BuildTable](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/db/builder.cc#L194) takes the builder's `io_status` and sends up the stack to `FlushJob::WriteLevel0Table`.
* [BlockBasedTableBuilder::Add](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/table/block_based/block_based_table_builder.cc#L753)
* [BlockBasedTableBuilder::Flush](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/table/block_based/block_based_table_builder.cc#L895) timely flush based on the policy.
* BlockBasedTableBuilder::WriteBlock
* [BlockBasedTableBuilder::WriteRawBlock](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/table/block_based/block_based_table_builder.cc#L1067) failure to write `block_contents` into the file when calling `WritableFileWriter::Append` due to `ENOSPC`. This leads to setting both [io_status](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/table/block_based/block_based_table_builder.cc#L1171) and [status](https://github.com/facebook/rocksdb/blob/91bc0130fa0790a77694c4a9c352fbbaa55b9f80/table/block_based/block_based_table_builder.cc#L1174) of `BlockBasedTableBuilder::Rep`.
 

> The new SetBGError interface is only for the specific IO Status and this function only handles the Retryable error (map it to hard error)

Could you elaborate on which specific IOStatuses is the new codepath for? I see the new classification under it with `data_loss_` and `retryable_`. What IO errors would fall under these?

zhichao-cao:
> * FlushMemTableToOutputFile

Hi @rohansuri ,

The code path you find is the logic that IO status will be set and pass to the new SetBGError(IOStatus& io_s). However, in this function, the io_s.GetRetryable() will be tested. Only in a special case: 1) we use the FileSystem, 2) the IO related function returns IOStatus io_s and it explicitly set the retryable_ (io_s.SetRetryable()) to true. Then, this error will be map to the HardError. In all other cases, including the out of space (Status::SpaceLimit()), old SetBGError will be called in the new SetBGError to handle them.

PR #6488, opened 2020-03-06
add transaction multi get to C API

morningsend:
Hi, I understand MultiGet now uses `io_uring` APIs from linux kernel, which can bring significant speed up to queries. I have created this PR to make C multiget API more uniform. Following things are added:

* Expose variants of `MultiGet` methods from `Transaction` to C API.
* Expose function to return base db of `transactiondb`, which is currently only present for `optimistictransactiondb`.
* Added unit tests.
* Refactor `c_test` utilities.


ryoqun:
@ajkr @morningsend what's remaining to proceed thi pr to be merged? context: https://github.com/rust-rocksdb/rust-rocksdb/pull/565

Issue #6471, opened 2020-03-02
expose `GetCompactionReasonString()` publicly


ajkr:
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://groups.google.com/forum/#!forum/rocksdb or https://www.facebook.com/groups/rocksdb.dev

### Expected behavior

Users of compaction event listeners should be able to get a human-readable string indicating the reason for a compaction.

### Actual behavior

Instead they get `CompactionJobInfo::compaction_reason`, which is a `CompactionReason` enum, with no way to translate it to something human-readable.

### Steps to reproduce the behavior
N/A

ajkr:
Thanks for taking it. Let me know if you have any questions. Coincidentally we are in the same bootcamp class.

cbi42:
Completed in #11778.

PR #6489, opened 2020-03-06
Expose GetCompactionReasonString() publicly

thepulkitagarwal:
For Issue #6471 

Add function `GetStringFromCompactionReason(...)` to get a string value from compaction reason, similar to compression type in `listener.h`.

### DIFF
--- a/db/compaction/compaction.h
+++ b/db/compaction/compaction.h
@@ -281,6 +281,8 @@ class Compaction {
 
   int GetInputBaseLevel() const;
 
+  const char* GetCompactionReasonString() const;
+
   CompactionReason compaction_reason() { return compaction_reason_; }
 
   const std::vector<FileMetaData*>& grandparents() const {
--- a/db/compaction/compaction.h
+++ b/db/compaction/compaction.h
@@ -281,8 +281,6 @@ class Compaction {
 
   int GetInputBaseLevel() const;
 
-  const char* GetCompactionReasonString() const;
-
   CompactionReason compaction_reason() { return compaction_reason_; }
 
   const std::vector<FileMetaData*>& grandparents() const {
--- a/options/options_helper.cc
+++ b/options/options_helper.cc
@@ -14,7 +14,6 @@
 #include "rocksdb/compaction_filter.h"
 #include "rocksdb/convenience.h"
 #include "rocksdb/filter_policy.h"
-#include "rocksdb/listener.h"
 #include "rocksdb/memtablerep.h"
 #include "rocksdb/merge_operator.h"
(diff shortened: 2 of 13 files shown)


db/compaction/compaction.h
```diff
@@ -281,6 +281,8 @@ class Compaction {
 
   int GetInputBaseLevel() const;
 
+  const char* GetCompactionReasonString() const;
```
ajkr: In order to be publicly visible (i.e., usable by RocksDB users), the function needs to be declared in a header file under "include/" directory. For example take a look at how `GetStringFromCompressionType()` is declared: https://github.com/facebook/rocksdb/blob/e171a219d592b1bda9b654db126f04d4d834a815/include/rocksdb/convenience.h#L296-L297.

options/options_helper.cc
```diff
@@ -253,6 +254,25 @@ std::unordered_map<std::string, CompressionType>
         {"kZSTD", kZSTD},
         {"kZSTDNotFinalCompression", kZSTDNotFinalCompression},
         {"kDisableCompressionOption", kDisableCompressionOption}};
+
+std::unordered_map<std::string, CompactionReason>
```
ajkr: I think we shouldn't follow `GetStringFromCompressionType()`'s implementation pattern - just following the declaration pattern (declaring `GetStringFromCompactionReason()` in a file under an "include/" directory) is enough. `CompactionReason` is not an option so its logic doesn't really belong here. We also don't do lookup based on string so the map isn't necessary.

Also if you want to make analogous changes for the other enums in "include/rocksdb/listener.h", that'd be helpful (e.g., `TableFileCreationReason`).

db/compaction/compaction.h
ajkr: Sorry I probably shouldn't have provided this example here, as the declaration itself is not substantial. Meanwhile, the example function looks too similar to what our function does so it's tempting to follow its implementation pattern.

options/options_helper.cc
thepulkitagarwal: I'm not sure where it should be defined. I thought of putting the definition in `compaction.h`, but then the other enums' definitions wouldn't fit in there. What file would you suggest for that?

ajkr: Looks like everything declared in `include/rocksdb/listener.h` so far is defined inline. We could either continue that pattern or create a `db/listener.cc` file for these definitions. Or, if you have the declarations in `include/rocksdb/convenience.h`, then the definitions should go in `db/convenience.cc`.

cbi42:
Implemented in #11778

PR #6497, opened 2020-03-07
V5.18.4 all platforms

adamretter:
@pdillinger Sorry about this, but there were 11 additional commits I had to pull in to build 5.18.4.

These are because of external changes over the time since I created the last PR for 5.18.4:
1. Some dependencies are no longer available from their previous locations
2. Newer version of macOS, has stricter compiler checks.

If you could merge this, and then re-cut the `v5.18.4` tag. Then I can check that out and rebuild and release. I think this should be fine now, I have built this branch on all release platforms:
1. macOS x64
2. Linux x86
3. Linux x64
4. Linux ppc64le
5. Linux aarch64
6. Windows x64


pdillinger:
I'm calling a timeout on making Travis happy and merging+tagging b9a4a10 as requested.

pdillinger:
For the record, example force update of an existing tag AND make sure it's not listed as "latest release" in github project:

    $ git tag -d v5.18.4
    Deleted tag 'v5.18.4' (was 6ea4477cf)
    $ GIT_COMMITTER_DATE="Mon, 27 Jan 2020 08:21:13 -0800" git tag -f -a v5.18.4 5.18.fb
    <enter a message>
    $ git push -f origin refs/tags/v5.18.4

Suggested message info: (Note: the originally tagged commit for this release was wrong but the tag has been updated. You might need to delete your copy of the tag with `git tag -d v5.18.4` to get the new one. See https://git-scm.com/docs/git-tag#_on_re_tagging)

PR #6512, opened 2020-03-11
Force Java version on Travis CI

adamretter:
In the `.travis.yml` file the `jdk: openjdk7` element is ignored when `language: cpp`. So whatever version of the JDK that was installed in the Travis container was used - typically JDK 11.

To ensure our RocksJava builds are working, we now instead install and use OpenJDK 8. Ideally we would use OpenJDK 7, as RocksJava supports Java 7, but many of the newer Travis containers don't support Java 7, so Java 8 is the next best thing.


pdillinger:
Should we be using -source to javac to enforce a source code compatibility level?

adamretter:
@pdillinger Hmm... good idea! Normally I am using a Java build tool (instead of Make) and so that stuff is taken care of for me. I will add it in now...

pdillinger:
I guess we could run into newer versions of javac not supporting such an old source level, but at least in the javac version I have handy, that appears to be a warning and not an error.

adamretter:
@pdillinger Hmm the `JAVAC_ARGS` isn't being picked up... I will take a further look...

adamretter:
@pdillinger Okay I updated this PR, assuming the CI is good. Could you re-import it please?

adamretter:
@pdillinger can you re-import please? I fixed the merge conflict

.travis.yml
```diff
@@ -98,8 +96,19 @@ install:
       TAR_OPT="--strip-components=1 -xj";
       mkdir cmake-dist && curl --silent --fail --show-error --location "${CMAKE_DIST_URL}" | tar -C cmake-dist ${TAR_OPT} && export PATH=$PWD/cmake-dist/bin:$PATH;
     fi
-  - if [[ "${JOB_NAME}" == java_test || "${JOB_NAME}" == cmake* ]]; then
-      java -version && echo "JAVA_HOME=${JAVA_HOME}";
+  - |
```
pdillinger: before_install is now redendant, right? Can't we remove it?

.travis.yml
```diff
@@ -98,8 +96,19 @@ install:
...
+      if [ "${TRAVIS_OS_NAME}" == osx ]; then
+        brew tap AdoptOpenJDK/openjdk
+        brew cask install adoptopenjdk8
+        export JAVA_HOME=$(/usr/libexec/java_home)
+      else
+        sudo apt-get install -y openjdk-8-jdk
+        export PATH=/usr/lib/jvm/java-8-openjdk-$(dpkg --print-architecture)/bin:$PATH
+        export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-$(dpkg --print-architecture)
+      fi
+      java -version && echo "JAVA_HOME=${JAVA_HOME}"
```
pdillinger: I suggest

    echo "JAVA_HOME=${JAVA_HOME}"
    which java && java -version
    which javac && javac -version

.travis.yml
adamretter: Yup

.travis.yml
adamretter: Okie

adamretter:
@pdillinger Okay I pushed your suggestions.

pdillinger:
> @pdillinger Okay I pushed your suggestions.

Let's hope the Travis gods are still happy ;)

PR #6517, opened 2020-03-11
fix MSVC build failures

hliu18:
fix a few build warnings that are treated as failures with more strict MSVC warning settings


adamretter:
LGTM

PR #6523, opened 2020-03-12
OpenForReadOnly is not supported in LITE mode

cheng-chang:
In DBLogicalBlockSizeCacheTest, do not test OpenForReadOnly in LITE mode.

Test Plan:
watch test for LITE mode


pdillinger:
> Test Plan:
> watch test for LITE mode

My bad. I should have rejected your test plan as insufficient. Local build and test with lite mode is easy. make clean; make LITE=1 check

PR #6529, opened 2020-03-12
fix: release log handles to allow deletion

hliu18:
It is observed in tests on Windows that background compaction fails to clean log files in PurgeObsoleteFiles() because job_context still holds open handles. job_context.Clean() releases those handles but is called after PurgeObsoleteFiles(). This fix releases those log handles before trying to clean the log files.


hliu18:
It should work. The issue was noticed on an earlier version. I wasn't aware about the fix.

PR #6530, opened 2020-03-13
Added the safe-to-ignore tag to version_edit

zhichao-cao:
Each time RocksDB switches to a new MANIFEST file from old one, it calls WriteCurrentStateToManifest() which writes a 'snapshot' of the current in-memory state of versions to the beginning of the new manifest as a bunch of version edits. We can distinguish these version edits from other version edits written during normal operations with a custom, safe-to-ignore tag.

Test plan: added test to version_edit_test, pass make asan_check

### DIFF
--- a/db/version_edit.cc
+++ b/db/version_edit.cc
@@ -60,7 +60,7 @@ enum Tag : uint32_t {
   kBlobFileAddition,
   kBlobFileGarbage,
   kStateUponManifestSwitch,
-  kManifestSwitchFinished,
+  kManifestSwitched,
 };
 
 enum NewFileCustomTag : uint32_t {
@@ -163,7 +163,7 @@ void VersionEdit::Clear() {
   is_in_atomic_group_ = false;
   remaining_entries_ = 0;
   state_upon_manifest_switch_ = false;
-  manifest_switch_finished_ = false;
+  manifest_switched_ = false;
 }
 
 bool VersionEdit::EncodeTo(std::string* dst) const {
@@ -301,8 +301,8 @@ bool VersionEdit::EncodeTo(std::string* dst) const {
     PutVarint32(dst, kStateUponManifestSwitch);
   }
 
-  if (manifest_switch_finished_) {
-    PutVarint32(dst, kManifestSwitchFinished);
+  if (manifest_switched_) {
+    PutVarint32(dst, kManifestSwitched);
   }
 
   if (is_column_family_add_) {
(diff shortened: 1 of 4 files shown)


db/version_edit.cc
```diff
@@ -59,6 +59,7 @@ enum Tag : uint32_t {
   kDbId,
   kBlobFileAddition,
   kBlobFileGarbage,
+  kSafeToIgnore,
```
riversand963: Sorry for the confusion during offline communication. I suggest that we rename it to something like `kStateUponManifestSwitch`.

zhichao-cao:
> Thanks @zhichao-cao for working on this. I wonder whether an end marker is needed to denote the end of such version edits written when switching MANIFESTs.

Thanks for the review. The new commit addresses the comments.

db/version_edit.h
```diff
@@ -423,6 +423,19 @@ class VersionEdit {
...
+  bool GetStateUponManifestSwitchTag() const {
+    return state_upon_manifest_switch_;
+  }
+  void SetManifestSwitchFinishedTag(bool tag) {
+    manifest_switch_finished_ = tag;
+  }
+  bool GetManifestSwitchFinishedTag() const {
+    return manifest_switch_finished_;
+  }
+
```
riversand963: For this group of functions, I suggest that we rename them by removing `SwitchTag` or `FinishedTag`.

db/version_edit.h
```diff
@@ -470,6 +483,13 @@ class VersionEdit {
 
   bool is_in_atomic_group_ = false;
   uint32_t remaining_entries_ = 0;
+  // To distinguish the version edit written by WriteCurrentStateToManifest
+  // and other redgular writes. Default is false.
```
riversand963: Nit: typo regular

db/version_set.cc
```diff
@@ -5072,6 +5075,18 @@ Status VersionSet::WriteCurrentStateToManifest(
...
+  end_flag.SetManifestSwitchFinishedTag(true);
+  std::string end_record;
+  if (!end_flag.EncodeTo(&end_record)) {
+    return Status::Corruption("Unable to Encode VersionEdit:" +
+                              end_flag.DebugString(true));
+  }
+  Status s_end_record = log->AddRecord(end_record);
+  if (!s_end_record.ok()) {
+    return s_end_record;
+  }
```
riversand963: How about just returning `s_end_of_record` directly.

db/version_edit.h
zhichao-cao: @riversand963 I will rename to SetStateUponManifest/GetStateUponManifest and SetManifestSwitched/GetManifestSwitched

db/version_set.cc
zhichao-cao: @riversand963 Sure

PR #6532, opened 2020-03-13
Destroy DB at the end of each test in db_logical_block_size_cache_test

cheng-chang:
If DB is not deleted, in concurrent test, the tests might fail because of the previously existing DB.

Test Plan:
make clean && make -j24 LITE=1  db_logical_block_size_cache_test && ./db_logical_block_size_cache_test
make clean && make -j24 db_logical_block_size_cache_test && ./db_logical_block_size_cache_test

### DIFF
--- a/db/db_logical_block_size_cache_test.cc
+++ b/db/db_logical_block_size_cache_test.cc
@@ -90,6 +90,7 @@ TEST_F(DBLogicalBlockSizeCacheTest, OpenClose) {
     ASSERT_EQ(0, cache_->Size());
     delete db;
   }
+  ASSERT_OK(DestroyDB(dbname_, options, {}));
 }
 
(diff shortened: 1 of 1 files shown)


db/db_logical_block_size_cache_test.cc
```diff
@@ -90,6 +90,7 @@ TEST_F(DBLogicalBlockSizeCacheTest, OpenClose) {
     ASSERT_EQ(0, cache_->Size());
     delete db;
   }
+  ASSERT_OK(DestroyDB(dbname_, options, {}));
```
riversand963: How about in the destructor?

cheng-chang: different tests have different setups of dbname and column families. So cannot just call DestroyDB with the same setup in destructor.

PR #6516, opened 2020-03-11
Separate timestamp related test from db_basic_test

zhichao-cao:
In some of the test, db_basic_test may cause time out due to its long running time. Separate the timestamp related test from db_basic_test to avoid the potential issue.

Test plan: pass make asan_check

PR #6535, opened 2020-03-13
Fix build bug caused by PR 6516

zhichao-cao:
Fix the build corruption caused by PR #6516

Testing plan: make make asan_check


zhichao-cao:
> LGTM

Thanks for helping with it!

PR #6538, opened 2020-03-13
Fault injection in db_stress

anand1976:
This PR implements a fault injection mechanism for injecting errors in reads in db_stress. The FaultInjectionTestFS is used for this purpose. A thread local structure is used to track the errors, so that each db_stress thread can independently enable/disable error injection and verify observed errors against expected errors. This is initially enabled only for Get and MultiGet, but can be extended to iterator as well once its proven stable.

Test plan:
crash_test
make check


test_util/fault_injection_test_fs.h
```diff
@@ -238,14 +278,91 @@ class FaultInjectionTestFS : public FileSystemWrapper {
...
+  char** GetFaultBacktrace(int* frames);
+
  private:
   port::Mutex mutex_;
   std::map<std::string, FSFileState> db_file_state_;
   std::set<std::string> open_files_;
   std::unordered_map<std::string, std::set<std::string>>
       dir_to_new_files_since_last_sync_;
   bool filesystem_active_;  // Record flushes, syncs, writes
+  bool filesystem_writable_;  // Record flushes, syncs, writes
```
riversand963: The code comment for `filesystem_writable_` is exactly the same as that of `filesystem_active_`. What is the difference between the two?

test_util/fault_injection_test_fs.cc
```diff
@@ -427,4 +468,70 @@ void FaultInjectionTestFS::UntrackFile(const std::string& f) {
...
+                                           char* scratch) {
+  struct ErrorContext* ctx =
+        static_cast<struct ErrorContext*>(thread_local_error_->Get());
+  if (ctx == nullptr || !ctx->enable_error_injection || !ctx->one_in) {
+    return IOStatus::OK();
+  }
+
+  if (ctx->rand.OneIn(ctx->one_in)) {
+    ctx->count++;
+    ctx->frames = backtrace(ctx->callstack, 128);
```
riversand963: This seems Linux-specific.

test_util/fault_injection_test_fs.cc
```diff
@@ -427,4 +468,70 @@ void FaultInjectionTestFS::UntrackFile(const std::string& f) {
   open_files_.erase(f);
 }
 
+IOStatus FaultInjectionTestFS::InjectError(ErrorOperation op,
+                                           Slice* slice,
```
riversand963: Better name than `slice`?

test_util/fault_injection_test_fs.cc
```diff
@@ -427,4 +468,70 @@ void FaultInjectionTestFS::UntrackFile(const std::string& f) {
...
+                                           char* scratch) {
+  struct ErrorContext* ctx =
```
riversand963: `struct ErrorContext*` seems to be C-style.

test_util/fault_injection_test_fs.cc
```diff
@@ -427,4 +468,70 @@ void FaultInjectionTestFS::UntrackFile(const std::string& f) {
...
+    ctx->count++;
+    ctx->frames = backtrace(ctx->callstack, 128);
+    switch (op) {
+      case READ:
+      {
+        uint32_t type = ctx->rand.Uniform(3);
+        switch (type) {
+          case 0:
+            return IOStatus::IOError();
+          case 1:
```
riversand963: what does `1` mean?

test_util/fault_injection_test_fs.h
```diff
@@ -238,14 +278,91 @@ class FaultInjectionTestFS : public FileSystemWrapper {
     error_ = io_error;
   }
 
+  // Specify what the operation, so we can inject the right type of error
+  enum ErrorOperation {
```
riversand963: Nit: specify underlying type of this enum and/or use type-safer enum.

test_util/fault_injection_test_fs.h
```diff
@@ -238,14 +278,91 @@ class FaultInjectionTestFS : public FileSystemWrapper {
...
+    READ = 0,
```
riversand963: I guess we follow naming convention such that they are named kRead and kOpen.

test_util/fault_injection_test_fs.h
```diff
@@ -238,14 +278,91 @@ class FaultInjectionTestFS : public FileSystemWrapper {
...
+    OPEN,
+  };
+
+  // Set thread-local parameters for error injection. The first argument,
+  // seed is the seed for the random number generator, and one_in determines
+  // the probability of injecting error (i.e an error is injected with
+  // 1/one_in probability)
+  void SetThreadLocalReadError(uint32_t seed, int one_in) {
+    struct ErrorContext* ctx =
```
riversand963: Any particular reason for C-style `struct ErrorContext` instead of C++ style `ErrorContext`?

test_util/fault_injection_test_fs.h
```diff
@@ -238,14 +278,91 @@ class FaultInjectionTestFS : public FileSystemWrapper {
...
+  enum ErrorOperation {
+    READ = 0,
+    OPEN,
+  };
+
+  // Set thread-local parameters for error injection. The first argument,
+  // seed is the seed for the random number generator, and one_in determines
+  // the probability of injecting error (i.e an error is injected with
+  // 1/one_in probability)
+  void SetThreadLocalReadError(uint32_t seed, int one_in) {
```
riversand963: Nit: rename to `SetThreadLocalReadErrorContext`?

test_util/fault_injection_test_fs.cc
anand1976: Yes. Made it conditional.

test_util/fault_injection_test_fs.cc
anand1976: Added comments

test_util/fault_injection_test_fs.cc
```diff
@@ -427,4 +468,80 @@ void FaultInjectionTestFS::UntrackFile(const std::string& f) {
...
+            if (result->data() == scratch) {
+              uint64_t offset = ctx->rand.Uniform((uint32_t)result->size());
+              uint32_t len = (uint32_t)std::min(result->size() - offset, 64UL);
+              assert(offset < result->size());
+              assert(offset + len <= result->size());
+              std::string str = DBTestBase::RandomString(&ctx->rand, len);
+              memcpy(scratch + offset, str.data(), len);
+              break;
+            } else {
+              [[fallthrough]];
```
zhichao-cao: Are we use "FALLTHROUGH_INTENDED;"?

test_util/fault_injection_test_fs.h
```diff
@@ -238,14 +278,92 @@ class FaultInjectionTestFS : public FileSystemWrapper {
...
+          static_cast<ErrorContext*>(thread_local_error_->Get());
+    int count = 0;
+    if (ctx != nullptr) {
+      count = ctx->count;
+      ctx->count = 0;
+    }
+    return count;
+  }
+
+  void EnableErrorInjection() {
```
zhichao-cao: EnableErrorInjection is only used for read? The error is returned from InjectionError instead of GetError. In this case, we do not need to call SetFilesystemActive to set the IO Error. Can we unify them?

test_util/fault_injection_test_fs.h
```diff
@@ -119,6 +124,25 @@ class TestFSRandomRWFile : public FSRandomRWFile {
   FaultInjectionTestFS* fs_;
 };
 
+class TestFSRandomAccessFile : public FSRandomAccessFile {
+ public:
+  explicit TestFSRandomAccessFile(const std::string& fname,
+                              std::unique_ptr<FSRandomAccessFile>&& f,
+                              FaultInjectionTestFS* fs);
+  virtual ~TestFSRandomAccessFile();
```
riversand963: Nit: remove `virtual` and add `override`.

db_stress_tool/db_stress_tool.cc
```diff
@@ -31,6 +32,21 @@ static std::shared_ptr<ROCKSDB_NAMESPACE::DbStressEnvWrapper> env_wrapper_guard;
 }  // namespace
 
 KeyGenContext key_gen_ctx;
+std::shared_ptr<FaultInjectionTestFS> fault_fs_guard;
```
riversand963: According to https://google.github.io/styleguide/cppguide.html#Variable_Comments, some comments can be helpful.

test_util/fault_injection_test_fs.cc
```diff
@@ -195,6 +196,30 @@ IOStatus TestFSRandomRWFile::Sync(const IOOptions& options,
...
 }
 
+TestFSRandomAccessFile::TestFSRandomAccessFile(const std::string& /*fname*/,
+                                       std::unique_ptr<FSRandomAccessFile>&& f,
+                                       FaultInjectionTestFS* fs)
+    : target_(std::move(f)), fs_(fs) {
+  assert(target_ != nullptr);
+}
+
+TestFSRandomAccessFile::~TestFSRandomAccessFile() {
```
riversand963: Nit: move this to .h?

db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -144,17 +145,48 @@ class NonBatchedOpsStressTest : public StressTest {
...
+    }
     Status s = db_->Get(read_opts, cfh, key, &from_db);
+    if (fault_fs_guard) {
+      error_count = fault_fs_guard->GetAndResetErrorCount();
+    }
     if (s.ok()) {
+      if (fault_fs_guard) {
+        if (error_count && !filter_read_error) {
+#ifndef NDEBUG
+          fprintf(stderr, "Didn't get expected error from Get\n");
```
riversand963: Since this should not happen, we may want to assert.

test_util/fault_injection_test_fs.cc
anand1976: Nice. Wasn't aware of that.

test_util/fault_injection_test_fs.h
anand1976: They serve different purposes though. SetFileSystemActive() will set an error globally, whereas EnableErrorInjection() only does it for the calling thread. The former is difficult to use in db_stress, whereas the latter is difficult to use when you want to inject an error in, say, a background thread. So I think both are needed.

db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -144,17 +145,48 @@ class NonBatchedOpsStressTest : public StressTest {
...
+      fault_fs_guard->EnableErrorInjection();
+      filter_read_error = false;
+    }
     Status s = db_->Get(read_opts, cfh, key, &from_db);
+    if (fault_fs_guard) {
+      error_count = fault_fs_guard->GetAndResetErrorCount();
+    }
     if (s.ok()) {
+      if (fault_fs_guard) {
+        if (error_count && !filter_read_error) {
```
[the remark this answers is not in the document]
anand1976: How so? ```filter_read_error``` is set to true if the corresponding sync point is triggered.

db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -243,20 +283,42 @@ class NonBatchedOpsStressTest : public StressTest {
 
     for (const auto& s : statuses) {
       if (s.ok()) {
-        // found case
-        thread->stats.AddGets(1, 1);
+        if (fault_fs_guard && error_count && !filter_read_error) {
```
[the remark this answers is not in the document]
anand1976: See previous comment.

db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -243,20 +283,42 @@ class NonBatchedOpsStressTest : public StressTest {
...
+#ifndef NDEBUG
+          fprintf(stderr, "Didn't get expected error from MultiGet\n");
+          int frames = 0;
+          char** strs = fault_fs_guard->GetFaultBacktrace(&frames);
```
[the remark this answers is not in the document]
anand1976: Its for debugging purposes. By the time we detect the swallowed error in TestGet, its too late to figure out where it should have been caught. Backtrace will help in that case.

table/block_based/block_based_table_reader.cc
```diff
@@ -1584,6 +1584,11 @@ Status BlockBasedTable::PrefetchIndexAndFilterBlocks(
     auto filter = new_table->CreateFilterBlockReader(
         prefetch_buffer, use_cache, prefetch_filter, pin_filter,
         lookup_context);
+#ifndef NDEBUG
+    if (rep_->filter_type != Rep::FilterType::kNoFilter && !filter) {
+      TEST_SYNC_POINT("FilterReadError");
```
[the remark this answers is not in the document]
anand1976: I kept it generic so db_stress wouldn't have to register for all possible sync points. Probably not that much value in having distinct sync points.

test_util/fault_injection_test_fs.h
zhichao-cao: > They serve different purposes though. SetFileSystemActive() will set an error globally, whereas EnableErrorInjection() only does it for the calling thread. The former is difficult to use in db_stress, whereas the latter is difficult to use when you want to inject an error in, say, a background thread. So I think both are needed.

I see. Thanks for the explanation.

db_stress_tool/db_stress_tool.cc
```diff
@@ -69,6 +85,19 @@ int db_stress_tool(int argc, char** argv) {
   } else {
     raw_env = Env::Default();
   }
+#ifndef NDEBUG
+  if (FLAGS_read_fault_one_in) {
+    FaultInjectionTestFS* fs = new FaultInjectionTestFS(FileSystem::Default().get());
```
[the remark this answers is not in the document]
riversand963: Looks like we need to call `Env::LoadEnv()` to obtain underlying custom env if necessary. Then create a FS from the underlying Env, wrap the created FS with FaultInjectionTestFS, created a CompositeEnvWrapper, and pass to options?

db_stress_tool/db_stress_common.h
```diff
@@ -259,6 +261,14 @@ inline enum RepFactory StringToRepFactory(const char* ctype) {
 extern enum RepFactory FLAGS_rep_factory;
 
 namespace ROCKSDB_NAMESPACE {
+extern std::shared_ptr<FaultInjectionTestFS> fault_fs_guard;
+extern std::shared_ptr<CompositeEnvWrapper> fault_env_guard;
+#if defined(ROCKSDB_SUPPORT_THREAD_LOCAL) && defined(OS_LINUX)
+extern thread_local bool filter_read_error;
```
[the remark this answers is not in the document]
anand1976: The perf_context uses a different syntax for ```OS_SOLARIS```. Do we need to check for that or its not needed?

db_stress_tool/db_stress_tool.cc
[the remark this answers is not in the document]
anand1976: Good point! We need to wrap the underlying FS and not necessarily the default FS.

anand1976:
@riversand963 PTAL

db_stress_tool/db_stress_common.h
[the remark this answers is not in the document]
riversand963: I see. Maybe we can keep this. Do you think we should add OS_SOLARIS here just to be consistent?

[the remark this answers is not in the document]
anand1976: Yes, makes sense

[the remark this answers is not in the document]
riversand963: Sorry about this late comment. I am actually not a big fan of global variables. Can we make them thread local, class static or the `ThreadLocalPtr`?

db_stress_tool/db_stress_common.h
```diff
@@ -261,6 +265,20 @@ inline enum RepFactory StringToRepFactory(const char* ctype) {
 extern enum RepFactory FLAGS_rep_factory;
 
 namespace ROCKSDB_NAMESPACE {
+#ifndef NDEBUG
+extern std::shared_ptr<FaultInjectionTestFS> fault_fs_guard;
```
riversand963: Can this not be global?

db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -243,20 +292,43 @@ class NonBatchedOpsStressTest : public StressTest {
...
-        // found case
-        thread->stats.AddGets(1, 1);
+#ifndef NDEBUG
+        if (fault_fs_guard && error_count && !filter_read_error) {
+          // Grab mutex so multiple thread don't try to print the
+          // stack trace at the same time
+          MutexLock l(thread->shared->GetMutex());
+          fprintf(stderr, "Didn't get expected error from MultiGet\n");
+          fprintf(stderr, "Callstack that injected the error\n");
+          fault_fs_guard->PrintFaultBacktrace();
```
riversand963: Now I get that, different `status` can have different stack trace when errors occur.

db_stress_tool/db_stress_tool.cc
```diff
@@ -31,6 +34,25 @@ static std::shared_ptr<ROCKSDB_NAMESPACE::DbStressEnvWrapper> env_wrapper_guard;
...
+std::shared_ptr<FaultInjectionTestFS> fault_fs_guard;
+#endif // NDEBUG
+std::shared_ptr<CompositeEnvWrapper> fault_env_guard;
+
+// Errors when reading filter blocks are ignored, so we use a thread
+// local variable updated via sync points to keep track of errors injected
+// while reading filter blocks in order to ignore the Get/MultiGet result
+// for those calls
+#if defined(ROCKSDB_SUPPORT_THREAD_LOCAL) && defined(OS_LINUX)
+thread_local bool filter_read_error;
```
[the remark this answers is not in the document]
anand1976: It means filter block. If it was the verb, ```filter_error``` would be a more appropriate variable name.

PR #6539, opened 2020-03-13
Use correct Env for DestroyDB in stress test

riversand963:
Summary:
When using custom Env, trying to call DestroyDB() with default Options will
fail.

Test Plan:
./db_stress


riversand963:
@cheng-chang could you PTAL?

PR #6540, opened 2020-03-14
Remove XXH3(preview) streaming APIs

### LINKED ISSUE #6508: Alignment bug in XXH3 (possible segfault)
See https://github.com/Cyan4973/xxHash/issues/175#issuecomment-596254843

pdillinger:
Summary: There was an alignment bug in our copy of the streaming APIs
for XXH3 (which we dubbed "XXH3p" for "preview" release). Since those
APIs are unused and some values for XXH3 have changed since XXH3p, I'm
simply removing those APIs, expecting it's better to use finalized XXH3
function if/when we decide to use those APIs (e.g. for checksums).

Fixes #6508

Test Plan: make check


ajkr:
hm, I was looking to use these for calculating checksum on `SliceParts` objects. What would you think about reverting most of this PR except leaving `XXH3p_createState()` / `XXH3p_freeState()` unimplemented? That should avoid the alignment problem and works for me (`XXH3p_state_t` will be on the stack).

PR #6552, opened 2020-03-17
Simplify migration to FileSystem API

anand1976:
The current Env/FileSystem API separation has a couple of issues -
1. It requires the user to specify 2 options - ```Options::env``` and ```Options::file_system``` - which means they have to make code changes to benefit from the new APIs. Furthermore, there is a risk of accessing the same APIs in two different ways, through Env in the old way and through FileSystem in the new way. The two may not always match, for example, if env is ```PosixEnv``` and FileSystem is a custom implementation. Any stray RocksDB calls to env will use the ```PosixEnv``` implementation rather than the file_system implementation.
2. There needs to be a simple way for the FileSystem developer to instantiate an Env for backward compatibility purposes.

This PR solves the above issues and simplifies the migration in the following ways -
1. Embed a shared_ptr to the ```FileSystem``` in the ```Env```, and remove ```Options::file_system``` as a configurable option. This way, no code changes will be required in application code to benefit from the new API. The default Env constructor uses a ```LegacyFileSystemWrapper``` as the embedded ```FileSystem```.
1a. - This also makes it more robust by ensuring that even if RocksDB
  has some stray calls to Env APIs rather than FileSystem, they will go
  through the same object and thus there is no risk of getting out of
  sync.
2. Provide a ```NewCompositeEnv()``` API that can be used to construct a
PosixEnv with a custom FileSystem implementation. This eliminates an
indirection to call Env APIs, and relieves the FileSystem developer of
the burden of having to implement wrappers for the Env APIs.
3. Add a couple of missing FileSystem APIs - ```SanitizeEnvOptions()``` and
```NewLogger()```

Tests:
1. New unit tests
2. make check and make asan_check

### DIFF
--- a/db/error_handler_fs_test.cc
+++ b/db/error_handler_fs_test.cc
@@ -150,12 +150,12 @@ class ErrorHandlerFSListener : public EventListener {
 };
 
 TEST_F(DBErrorHandlingFSTest, FLushWriteError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
+  std::unique_ptr<Env> fault_fs_env(NewCompositeEnv(fault_fs));
   std::shared_ptr<ErrorHandlerFSListener> listener(
       new ErrorHandlerFSListener());
   Options options = GetDefaultOptions();
-  options.file_system.reset(fault_fs);
+  options.env = fault_fs_env.get();
   options.create_if_missing = true;
   options.listeners.emplace_back(listener);
   Status s;
@@ -181,12 +181,12 @@ TEST_F(DBErrorHandlingFSTest, FLushWriteError) {
 }
 
 TEST_F(DBErrorHandlingFSTest, ManifestWriteError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
+  std::unique_ptr<Env> fault_fs_env(NewCompositeEnv(fault_fs));
   std::shared_ptr<ErrorHandlerFSListener> listener(
       new ErrorHandlerFSListener());
   Options options = GetDefaultOptions();
-  options.file_system.reset(fault_fs);
+  options.env = fault_fs_env.get();
   options.create_if_missing = true;
   options.listeners.emplace_back(listener);
   Status s;
@@ -223,12 +223,12 @@ TEST_F(DBErrorHandlingFSTest, ManifestWriteError) {
 }
 
 TEST_F(DBErrorHandlingFSTest, DoubleManifestWriteError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
+  std::unique_ptr<Env> fault_fs_env(NewCompositeEnv(fault_fs));
   std::shared_ptr<ErrorHandlerFSListener> listener(
       new ErrorHandlerFSListener());
   Options options = GetDefaultOptions();
-  options.file_system.reset(fault_fs);
+  options.env = fault_fs_env.get();
   options.create_if_missing = true;
   options.listeners.emplace_back(listener);
   Status s;
@@ -272,12 +272,12 @@ TEST_F(DBErrorHandlingFSTest, DoubleManifestWriteError) {
 }
 
 TEST_F(DBErrorHandlingFSTest, CompactionManifestWriteError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
+  std::unique_ptr<Env> fault_fs_env(NewCompositeEnv(fault_fs));
   std::shared_ptr<ErrorHandlerFSListener> listener(
       new ErrorHandlerFSListener());
   Options options = GetDefaultOptions();
-  options.file_system.reset(fault_fs);
+  options.env = fault_fs_env.get();
   options.create_if_missing = true;
   options.level0_file_num_compaction_trigger = 2;
   options.listeners.emplace_back(listener);
@@ -344,12 +344,12 @@ TEST_F(DBErrorHandlingFSTest, CompactionManifestWriteError) {
 }
 
 TEST_F(DBErrorHandlingFSTest, CompactionWriteError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
+  std::unique_ptr<Env> fault_fs_env(NewCompositeEnv(fault_fs));
   std::shared_ptr<ErrorHandlerFSListener> listener(
       new ErrorHandlerFSListener());
   Options options = GetDefaultOptions();
-  options.file_system.reset(fault_fs);
+  options.env = fault_fs_env.get();
   options.create_if_missing = true;
   options.level0_file_num_compaction_trigger = 2;
   options.listeners.emplace_back(listener);
@@ -387,10 +387,10 @@ TEST_F(DBErrorHandlingFSTest, CompactionWriteError) {
 }
 
 TEST_F(DBErrorHandlingFSTest, CorruptionError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
+  std::unique_ptr<Env> fault_fs_env(NewCompositeEnv(fault_fs));
   Options options = GetDefaultOptions();
-  options.file_system.reset(fault_fs);
+  options.env = fault_fs_env.get();
   options.create_if_missing = true;
   options.level0_file_num_compaction_trigger = 2;
   Status s;
@@ -426,12 +426,12 @@ TEST_F(DBErrorHandlingFSTest, CorruptionError) {
 }
 
 TEST_F(DBErrorHandlingFSTest, AutoRecoverFlushError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
+  std::unique_ptr<Env> fault_fs_env(NewCompositeEnv(fault_fs));
   std::shared_ptr<ErrorHandlerFSListener> listener(
       new ErrorHandlerFSListener());
   Options options = GetDefaultOptions();
-  options.file_system.reset(fault_fs);
+  options.env = fault_fs_env.get();
   options.create_if_missing = true;
   options.listeners.emplace_back(listener);
   Status s;
@@ -460,12 +460,12 @@ TEST_F(DBErrorHandlingFSTest, AutoRecoverFlushError) {
 }
 
 TEST_F(DBErrorHandlingFSTest, FailRecoverFlushError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
+  std::unique_ptr<Env> fault_fs_env(NewCompositeEnv(fault_fs));
   std::shared_ptr<ErrorHandlerFSListener> listener(
       new ErrorHandlerFSListener());
   Options options = GetDefaultOptions();
-  options.file_system.reset(fault_fs);
+  options.env = fault_fs_env.get();
   options.create_if_missing = true;
   options.listeners.emplace_back(listener);
   Status s;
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: env/composite_env_wrapper.h [hunk 1]
```diff
@@ -293,6 +293,10 @@ class CompositeEnvWrapper : public Env {
   // calls to env, and all file operations to fs
   explicit CompositeEnvWrapper(Env* env, FileSystem* fs)
       : env_target_(env), fs_env_target_(fs) {}
+  explicit CompositeEnvWrapper(Env* env, std::unique_ptr<FileSystem>&& fs)
+      : env_target_(env), fs_env_ptr_(std::move(fs)), fs_env_target_(fs_env_ptr_.get()) {}
```
riversand963: Can `CompositeEnvWrapper` keep `fs_env_ptr_` and remove `fs_env_target_`, since the latter is just `fs_env_ptr_.get()`?

env/composite_env_wrapper.h
```diff
@@ -293,6 +293,10 @@ class CompositeEnvWrapper : public Env {
   // calls to env, and all file operations to fs
   explicit CompositeEnvWrapper(Env* env, FileSystem* fs)
       : env_target_(env), fs_env_target_(fs) {}
+  explicit CompositeEnvWrapper(Env* env, std::unique_ptr<FileSystem>&& fs)
```
riversand963: Should `fs` be passed to `Env` (base class constructor) as well?

include/rocksdb/env.h
```diff
@@ -530,12 +533,19 @@ class Env {
...
+
   // If you're adding methods here, remember to add them to EnvWrapper too.
 
  protected:
   // The pointer to an internal structure that will update the
   // status of each thread.
   ThreadStatusUpdater* thread_status_updater_;
+
+  // Pointer to the underlying FileSystem implementation
+  std::shared_ptr<FileSystem> file_system_;
```
riversand963: What is the difference between `Env::file_system_` and `CompositeEnv::fs_env_ptr`?

include/rocksdb/env.h
```diff
@@ -530,12 +533,19 @@ class Env {
 
   virtual void SanitizeEnvOptions(EnvOptions* /*env_opts*/) const {}
 
+  // Get the FileSystem implementation this Env was constructed with. It
+  // could be a fully implemented one, or a wrapper class around the Env
+  std::shared_ptr<FileSystem> GetFileSystem();
```
mrambacher: shouldn't this be const std::shared_ptr<FileSystem>& GetFileSystem() const?

env/env_posix.cc
```diff
@@ -406,18 +382,34 @@ class PosixEnv : public CompositeEnvWrapper {
...
   //  has no access to db files.
-  bool allow_non_owner_access_;
+  bool& allow_non_owner_access_;
 };
 
 PosixEnv::PosixEnv()
-    : CompositeEnvWrapper(this, FileSystem::Default().get()),
-      thread_pools_(Priority::TOTAL),
-      allow_non_owner_access_(true) {
+    : CompositeEnvWrapper(this, FileSystem::Default()),
```
mrambacher: Why doesn't PosixEnv just inherit from Env?  What is gained by the Wrapper?

env/env_posix.cc
```diff
@@ -428,6 +420,15 @@ PosixEnv::PosixEnv()
   thread_status_updater_ = CreateThreadStatusUpdater();
 }
 
+PosixEnv::PosixEnv(const PosixEnv* default_env, std::shared_ptr<FileSystem> fs)
```
mrambacher: I do not think this constructor is or should be necessary.  Who calls it and when?

env/env_posix.cc
```diff
@@ -519,9 +520,12 @@ Env* Env::Default() {
...
-  static CompositeEnvWrapper composite_env(&default_env,
-                                           FileSystem::Default().get());
-  return &composite_env;
+  return &default_env;
+}
+
+std::unique_ptr<Env> NewCompositeEnv(std::shared_ptr<FileSystem> fs) {
+  PosixEnv* default_env = static_cast<PosixEnv*>(Env::Default());
+  return std::unique_ptr<Env>(new PosixEnv(default_env, fs));
 }
```
mrambacher: This seems like the wrong approach to me.  Can you explain why we need this method at all?

db/version_set.cc
```diff
@@ -4771,9 +4771,10 @@ Status VersionSet::DumpManifest(Options& options, std::string& dscname,
   Status s;
   {
     std::unique_ptr<FSSequentialFile> file;
-    s = options.file_system->NewSequentialFile(
+    std::shared_ptr<FileSystem> fs = options.env->GetFileSystem();
```
mrambacher: auto & fs = ... ?

include/rocksdb/env.h
anand1976: ```CompositeEnv``` can use ```Env::file_system_```. I believe there's no need for ```fs_env_ptr_```.

env/composite_env_wrapper.h
anand1976: Yes, it can. Will make the change.

db/error_handler_fs_test.cc
```diff
@@ -150,12 +150,12 @@ class ErrorHandlerFSListener : public EventListener {
 };
 
 TEST_F(DBErrorHandlingFSTest, FLushWriteError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default().get()));
```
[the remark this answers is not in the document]
anand1976: Yes, it can be a shared_ptr. Let me see if I can change it. If its not that easy, maybe a separate PR.

env/composite_env_wrapper.h
```diff
@@ -293,6 +293,10 @@ class CompositeEnvWrapper : public Env {
...
+      : env_target_(env), fs_env_ptr_(std::move(fs)), fs_env_target_(fs_env_ptr_.get()) {}
+  explicit CompositeEnvWrapper(Env* env, std::shared_ptr<FileSystem> fs)
```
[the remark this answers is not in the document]
anand1976: EnvWrapper redirects to a base Env, CompositeEnvWrapper redirects to a base Env and a base FileSystem. EnvWrapper could be made to conditionally redirect file operations to an Env or a FileSystem depending on how its constructed, but I choose to keep the two separate.

env/env_posix.cc
anand1976: Backward compatibility. The old/to-be-deprecated Env APIs should still work for the time being.

env/env_posix.cc
anand1976: See previous comment about backward compatibility.

anand1976:
@mrambacher I took a look at MockEnv (can't find SleepEnv in the codebase). Agree that EnvWrapper should have a constructor that takes an Env pointer and a FileSystem.

include/rocksdb/env.h
anand1976: Makes sense, since we can avoid some overhead if the caller just wants to dereference it immediately.

env/env.cc
```diff
@@ -472,4 +480,14 @@ Status NewEnvLogger(const std::string& fname, Env* env,
   return Status::OK();
 }
 
+const std::shared_ptr<FileSystem>& Env::GetFileSystem() const {
+  return file_system_;
+}
+
+#ifdef OS_WIN
```
riversand963: I guess this function is here temporarily. It will eventually be moved to port/win?

env/composite_env_wrapper.h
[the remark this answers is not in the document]
mrambacher: Why does an EnvWrapper need to do anything with a FileSystem?  I do not see the need for both classes and think it just adds an extra layer and more confusion.  Or is the plan to deprecate the EnvWrapper code?

Since an Env has a FileSystem, there is need to wrap the FileSystem or related calls in the EnvWrapper class (just like the CompositeEnvWrapper does).  You can create an EnvWrapper with just an Env (in which case it gets the FileSystem from the input arg) or with both an Env and a FileSystem.  In both cases, it passes the FileSystem to the Env parent.

env/env_posix.cc
mrambacher: Can you explain the compatibility?  I do not see where that is coming from.

env/env_posix.cc
mrambacher: This logic is already accomplished through the CompositeEnv class and doesn't need to be pushed into the PosixEnv one.

I think what you are after is that there should be an Env::Default(std::shared_ptr<FileSystem>) API added.  This way, you could create a "PosixEnv" using a non-Posix File system.  Additionally, there should be a std::shared_ptr<FileSystem> & FileSystem::Default() call that returns a PosixFileSystem.

env/env_posix.cc
anand1976: ```Env::New*File``` APIs should call the corresponding ```FileSystem::New*File``` APIs. Since the former are pure virtual functions in Env, CompositeEnvWrapper provides the necessary wrappers to redirect those calls to the underlying FileSystem. Eventually, WinEnv should also inherit from it.

env/env_posix.cc
anand1976: The FileSystem::Default() API already exists. NewCompositeEnv(std::shared_ptr<FileSystem>) is basically the Env::Default(std::shared_ptr) API that you're referring to.

env/composite_env_wrapper.h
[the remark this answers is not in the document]
anand1976: The plan is to deprecate the file related APIs in both Env and EnvWrapper eventually. 

You have a good point about wrapping the FileSystem related calls in EnvWrapper. Taking MockEnv as an example, it could be split into two -

```
class MockFileSystem : public FileSystemWrapper {
  MockFileSystem(std::shared_ptr<FileSystem> base_fs)
    :  FileSystemWrapper(base_fs) {}
```
and
```
class MockEnv : public EnvWrapper {
  MockEnv(Env* base_env, std::shared_ptr<FileSystem> base_fs)
    :  Env(std::make_shared<MockFileSystem>(base_fs)) {}
```
and FileSystem related calls in EnvWrapper would look something like this -
```
class EnvWrapper : public Env {
  EnvWrapper(Env* base_env, std::shared<FileSystem> base_fs)
    :  Env(base_fs), target_(base_env), constructed_with_fs_(true) {}

  Status NewSequentialFile(...) {
    if (constructed_with_fs_) {
      return file_system_->NewSequentialFile(...);
    } else {
      return target_->NewSequentialFile(...);
    }
  }
```

There may be other ways of doing it - need to think about it more. Will defer this to a separate PR. Right now, my focus is on porting an Env+FS implementation that inherits directly from Env, very similar to HdfsEnv.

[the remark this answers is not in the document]
mrambacher: Why isn't the model:
- In the base env classes, the "legacy file system" calls are no longer pure virtual.  Instead, they are wrapped as "file_system->DoSomething()"
- There is a "LegacyEnvWrapper" that converts a legacy env into a file system.  
- The Env class has a constructor that takes a FileSystem as an argument
- The no-arg Env constructor creates a LegacyEnvWrapper from "this"
- The EnvWrapper ONLY wraps the non-legacy FileSystem calls.  All other calls are handled by the base Env class, which will redirect it to the underlying FileSystem
- The EnvWrapper was two constructors, one that takes an Env and one that takes an Env and a FileSystem.  

There is no need for three classes (CompositeWrapper, EnvWrapper, and Env) as they all overlap.  

And there should not need to be modifications to the base Posix class, other than to separate out the PosixFileSystem, but it sounds like that is already done.

I would also make the case that it is very hard to prove that what you done works until you have covered a few more of the cases -- like making a few of the existing implementations follow this model.  Ideally, I would think you would do ALL of them to prevent code rot...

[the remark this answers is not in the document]
anand1976: > * In the base env classes, the "legacy file system" calls are no longer pure virtual.  Instead, they are wrapped as "file_system->DoSomething()"
> * The EnvWrapper ONLY wraps the non-legacy FileSystem calls.  All other calls are handled by the base Env class, which will redirect it to the underlying FileSystem

I don't disagree. This may be what we end up with. However, I'm not comfortable with making the calls non-pure virtual yet until the new APIs are proven with a FileSystem implementation that actually uses some of the features of the new API (its WIP currently). MockEnv etc are not good candidates. I understand they need to be accommodated, but they're not useful in proving the new API.

CompositeEnvWrapper and PosixEnv are internal to RocksDB. If/when we go with the approach you suggested, it should be very cheap to refactor them.

env/env.cc
anand1976: Yes. Eventually, WinEnv should be ported to the new APIs and this function will have a proper implementation.

db/error_handler_fs_test.cc
```diff
@@ -181,12 +181,12 @@ TEST_F(DBErrorHandlingFSTest, FLushWriteError) {
 }
 
 TEST_F(DBErrorHandlingFSTest, ManifestWriteError) {
-  FaultInjectionTestFS* fault_fs =
-      new FaultInjectionTestFS(FileSystem::Default().get());
+  std::shared_ptr<FaultInjectionTestFS> fault_fs(new FaultInjectionTestFS(FileSystem::Default()));
```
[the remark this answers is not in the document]
anand1976: Maybe leave it like that for now, since efficiency is not that much of a concern here.

FelipeGarcia8704:
> The current Env/FileSystem API separation has a couple of issues -
> 
> 1. It requires the user to specify 2 options - `Options::env` and `Options::file_system` - which means they have to make code changes to benefit from the new APIs. Furthermore, there is a risk of accessing the same APIs in two different ways, through Env in the old way and through FileSystem in the new way. The two may not always match, for example, if env is `PosixEnv` and FileSystem is a custom implementation. Any stray RocksDB calls to env will use the `PosixEnv` implementation rather than the file_system implementation.
> 2. There needs to be a simple way for the FileSystem developer to instantiate an Env for backward compatibility purposes.
> 
> This PR solves the above issues and simplifies the migration in the following ways -
> 
> 1. Embed a shared_ptr to the `FileSystem` in the `Env`, and remove `Options::file_system` as a configurable option. This way, no code changes will be required in application code to benefit from the new API. The default Env constructor uses a `LegacyFileSystemWrapper` as the embedded `FileSystem`.
>    1a. - This also makes it more robust by ensuring that even if RocksDB
>    has some stray calls to Env APIs rather than FileSystem, they will go
>    through the same object and thus there is no risk of getting out of
>    sync.
> 2. Provide a `NewCompositeEnv()` API that can be used to construct a
>    PosixEnv with a custom FileSystem implementation. This eliminates an
>    indirection to call Env APIs, and relieves the FileSystem

PR #6554, opened 2020-03-18
Eliminate the mutable_offset in OptionsTypeInfo

mrambacher:
This PR separates the OptionTypeInfo maps into two -- one containing the mutable options and one for the immutable ones.  This separation will make it easier to manage the maps going forward -- there is no need to keep track of if an object is mutable or not.  Additionally, it means that the options map points to only one structure -- either the mutable or immutable one -- eliminating any confusion over which one to use.

The old methodology could lead to some confusion, especially when dealing with nested objects.  For example, the options of the cache had to all be marked as mutable so that they would be picked up by some other method.  This separation will eliminate any confusion.

The kMutable flag was not removed, as it might have utility in the future.

Some methods were added as "bridge" methods until the Configurable class is completed and merged.


options/options_helper.cc
```diff
@@ -577,6 +595,90 @@ Status GetMutableOptionsFromStrings(
...
+        unused_opts->insert(o);
+      } else {
+        invalid_opts.insert(o);
+        invalid = s;
+      }
+    }
+  }
+  // While there are unused properties and we processed at least one,
+  // go through the remaining unused properties and attempt to configure them.
+  while (found_one && !unused_opts->empty()) {
```
zhichao-cao, line 642: I'm confused here, if we cannot parse the unused options in the first round, why we are able to parse it in the second round?

options/options_helper.cc
```diff
@@ -676,6 +774,13 @@ Status GetStringFromStruct(
   return Status::OK();
 }
 
+Status GetStringFromMutableDBOptions(const ConfigOptions& config_options,
```
zhichao-cao, line 777: If we provide GetStringFromMutableDBOptions, I think it is natural to provide GetStringFromImmutableDBOptions. Otherwise, we do not need the interface, we can directly call GetStringFromStruct when we need to serialize the mutable or immutable options. Similar for GetStringFromMutableCFOptions. What do you think?

options/options_helper.cc
mrambacher: There are a few ways this can happen.

1.  When we support dynamic loading of objects, it is possible the factories are not loaded during the first pass, meaning we cannot instantiate the instance until we instantiate the registry;

2.  If values are encoded independently (table.id = BlockBasedTable; table.checksum=12345;", it is possible that the table.checksum is processed before table.id, in which case the underlying object may be wrong or not created yet. In this case, the checksum property would not exist until the ID was interpreted.

options/options_helper.cc
mrambacher: The GetStringFromMutable is only required for tests.  Also, the GetStringFromStruct will be going away in a future PR

mrambacher:
This change was taken as part of #5753

PR #6564, opened 2020-03-19
Create a thread in DeleteScheduler only when rate limit enabled

akankshamahajan15:
Summary: Create a thread in DeleteScheduler only when delete rate limit is set
	 because when there is no rate limit on deletion, a thread per DeleteScheduler
	 consumes unnecessary resources.

Task Id: T64154614

Test Plan: make -j64 check

### DIFF
--- a/file/delete_scheduler.h
+++ b/file/delete_scheduler.h
@@ -45,6 +45,7 @@ class DeleteScheduler {
   // Set delete rate limit in bytes per second
   void SetRateBytesPerSecond(int64_t bytes_per_sec) {
     rate_bytes_per_sec_.store(bytes_per_sec);
+    MaybeCreateBackgroundThread();
   }
 
   // Mark file as trash directory and schedule it's deletion. If force_bg is
@@ -90,6 +91,8 @@ class DeleteScheduler {
                          uint64_t* deleted_bytes, bool* is_complete);
 
   void BackgroundEmptyTrash();
+
+  void MaybeCreateBackgroundThread();
 
   Env* env_;
   FileSystem* fs_;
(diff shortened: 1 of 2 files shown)


file/delete_scheduler.h
```diff
@@ -45,6 +45,7 @@ class DeleteScheduler {
   // Set delete rate limit in bytes per second
   void SetRateBytesPerSecond(int64_t bytes_per_sec) {
     rate_bytes_per_sec_.store(bytes_per_sec);
+    CreateBackgroundThread();
```
riversand963: How about renaming to `MaybeCreateBackgroundThread()` in which you check the condition and create the thread if necessary? This function can also be called in the constructor, and the check outside this function can be removed.

file/delete_scheduler.h
```diff
@@ -91,6 +92,8 @@ class DeleteScheduler {
 
   void BackgroundEmptyTrash();
 
+  void MaybeCreateBackgroundThread();
```
riversand963: Sorry for this last minute comment. Adding a one-liner comment for this function will be good.

file/delete_scheduler.h
```diff
@@ -45,6 +45,7 @@ class DeleteScheduler {
   // Set delete rate limit in bytes per second
   void SetRateBytesPerSecond(int64_t bytes_per_sec) {
     rate_bytes_per_sec_.store(bytes_per_sec);
+    MaybeCreateBackgroundThread();
```
anand1976: It would be safer to lock the mutex before calling this. I don't expect SetRateBytesPerSecond() to be called by multiple threads, but it doesn't hurt to handle it anyway.

lee-qiu: You are right. Rocksdb maybe crash If we call the SetRateBytesPerSecond() function concurrently without lock. Unfortunately, this code seems to have been merged without fix on it.

PR #6571, opened 2020-03-21
Fix the MultiGet testing failure in  Circleci

zhichao-cao:
The MultiGet test in db_basic_test fails in CircleCI vs2019. The reason is that even Snappy compression is enabled, the first compression type is still kNoCompression. This PR checks the list and ensure that only when compression is enable and the compression type is valid, compression will be enabled. Such that, it will not fail the combined read test in MultiGet.

Test plan: make check, db_basic_test.

### DIFF
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -7,8 +7,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
-#include <iostream>
-
 #include "db/db_test_util.h"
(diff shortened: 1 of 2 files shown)


db/db_basic_test.cc
```diff
@@ -1979,7 +1987,6 @@ class DBBasicTestWithParallelIO
 };
 
 // TODO: fails on CircleCI's Windows env
```
cheng-chang: this TODO can also be removed now

zhichao-cao: @cheng-chang  Sure.

PR #6573, opened 2020-03-22
[RocksJava] Add RocksIterator::Refresh

npepinpe:
This PR exposes the `Iterator::Refresh` method to the Java API by adding it on the `RocksIteratorInterface` interface. There are three concrete implementations: `RocksIterator`, `SstFileReaderIterator`, and `WBWIRocksIterator`. For the first two cases, the JNI side simply delegates to the underlying `Iterator::Refresh` method; in the last case, as it doesn't share an ancestor, and per the discussion in #3465, a `Status::NotSupported` exception is thrown.

As the last PR had no activity in a while, I'm opening a new one - I'm completely fine with merging the previous PR if it gets completed before this is reviewed.

Let me know if there's anything missing or anything else I can do :+1:


pdillinger:
LGTM. Essentially same as #3465. @adamretter?

java/rocksjni/write_batch_with_index.cc
```diff
@@ -860,3 +860,15 @@ jlongArray Java_org_rocksdb_WBWIRocksIterator_entry1(JNIEnv* env,
...
+
+/*
+ * Class:     org_rocksdb_WBWIRocksIterator
+ * Method:    refresh0
+ * Signature: (J)V
+ */
+void Java_org_rocksdb_WBWIRocksIterator_refresh0(JNIEnv* env, jobject /*jobj*/,
+                                            jlong handle) {
+  ROCKSDB_NAMESPACE::Status s = ROCKSDB_NAMESPACE::Status::NotSupported("Refresh() is not supported");
+  (void)handle;
```
adamretter, line 872: What is the `(void)handle;` line for?

adamretter: Apart from that it looks good :-)

pdillinger: Suppress unused parameter warning. Could have commented out the name, but eh

npepinpe: My bad, I wasn't aware you could comment out the name instead :sweat_smile:

adamretter: @npepinpe IMHO it's better to remove the names that are unused rather than commented out - could you update in that manner please? I think @pdillinger has the facility to re-import the PR?

PR #6574, opened 2020-03-23
Fix write_unprepared_transaction_test crash on debug version.

burtonli:
The last key may hit index of out bound exception when id = 9.

### DIFF
--- a/utilities/transactions/write_unprepared_transaction_test.cc
+++ b/utilities/transactions/write_unprepared_transaction_test.cc
@@ -182,8 +182,8 @@ TEST_P(WriteUnpreparedStressTest, ReadYourOwnWriteStress) {
     ReadOptions read_options;
 
     for (uint32_t i = 0; i < kNumIter; i++) {
-      std::set<std::string> owned_keys(&keys[id * kNumKeys],
-                                       &keys[(id + 1) * kNumKeys]);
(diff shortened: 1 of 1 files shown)


utilities/transactions/write_unprepared_transaction_test.cc
```diff
@@ -182,8 +182,8 @@ TEST_P(WriteUnpreparedStressTest, ReadYourOwnWriteStress) {
     ReadOptions read_options;
 
     for (uint32_t i = 0; i < kNumIter; i++) {
-      std::set<std::string> owned_keys(&keys[id * kNumKeys],
-                                       &keys[(id + 1) * kNumKeys]);
+      std::set<std::string> owned_keys(keys.begin() + id * kNumKeys,
+                                       keys.begin() + (id + 1) * kNumKeys);
```
cheng-chang: this iterator can still go out of bounds, can we explicitly compute a valid end?

burtonli: I think it's fine for iterator as keys.begin() + keys.size() = keys.end().

cheng-chang: correct, after reading the context, this cannot go over keys.end, so it's fine.

PR #6579, opened 2020-03-24
Fix msvc debug test failures

burtonli:
1. stats_history_test: one slice of stats history is 12526 Bytes, which is greater than original assumption. 
image
2. table_test: in VerifyBlockAccessTrace function, release trace reader before delete trace file.


burtonli:
@siying could you please help to find someone to review this test fix, thanks!

PR #6581, opened 2020-03-24
Refactor with VersionEditHandler

riversand963:
Summary:
Added a few classes in the same class hierarchy to remove code duplication and
refactor the logic of reading and processing MANIFEST files.

New classes are as follows.
```
class VersionEditHandlerBase;
class ListColumnFamiliesHandler : VersionEditHandlerBase;
class FileChecksumRetriever : VersionEditHandlerBase;
class DumpManifestHandler : VersionEditHandler;
```
Classes that already existed before this PR are as follows.
```
class VersionEditHandler : VersionEditHandlerBase;
```

With these classes, refactored functions: `VersionSet::Recover()`,
`VersionSet::ListColumnFamilies()`, `VersionSet::DumpManifest()`,
`GetFileChecksumFromManifest()`.

Test Plan (devserver):
```
make check
```
These refactored code, especially recovery-related logic, will be tested intensively by
all existing unit tests and stress tests. For example, run
```
make crash_test
```
Verified 3 successful runs on devserver.


riversand963:
> Looks great. Can `DumpManifest()` and `ReactiveVersionSet::Recover()` take the same approach?

Thanks @ajkr for the review. Yes, I believe `DumpManifest()` or even `ListColumnFamilies()` can take this approach.
`ReactiveVersionSet::Recover` is a little different, because it needs to handle the case of MANIFEST switch, and more work is needed there.

riversand963:
One travis failure should be addressed in #7367

riversand963:
Thanks @ajkr for the review. I have updated the PR so that DumpManifest, ListColumnFamilies and GetFileChecksumsFromManifest all use the same code path.

ajkr:
Sorry I did not get back to this in reasonable time (it'd take a while to relearn the MANIFEST writing/processing logic as I can't seem to remember it..), but glad this was merged!

PR #6525, opened 2020-03-12
Update a few scripts to be python3 compatible

riversand963:
Summary:
There are a few scripts with python3 compatibility issues that were not
detected by automated tool before. Update them now.

Test Plan (devserver):
python2 tools/ldb_test.py
python3 tools/ldb_test.py

python2 tools/write_stress_runner.py --runtime_sec=30
python3 tools/write_stress_runner.py --runtime_sec=30

python2 tools/db_crashtest.py --simple --interval=2 --duration=10 blackbox
python3 tools/db_crashtest.py --simple --interval=2 --duration=10 blackbox

python2 tools/db_crashtest.py --simple --duration=10 --random_kill_odd=1000 --ops_per_thread=1000 whitebox
python3 tools/db_crashtest.py --simple --duration=10 --random_kill_odd=1000 --ops_per_thread=1000 whitebox

PR #6583, opened 2020-03-24
Keep building RocksJava on all architectures

adamretter:
Adding solid support for multiple architectures was initially triggered by RocksJava users. As such I would like to keep the CI for RocksJava on all architectures, to ensure we don't break backwards compatibility.

@pdillinger okay let's see how long it takes to complete Travis-CI with this one...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,17 +118,6 @@ matrix:
     env: TEST_GROUP=4
   - if: type = pull_request
     os : osx
-    env: JOB_NAME=java_test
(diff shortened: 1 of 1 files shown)


riversand963:
With current resource quota, running this many Travis tests instances seem to be very expensive. For example, I updated #6525 over one hour ago, but Travis has not even started. Other PRs seem stuck as well.
Before we increase our Travis resource, I suggest that we plan the resource usage accordingly to ensure better development experience.

PR #6590, opened 2020-03-24
Create a Customizable class to load classes and configurations

mrambacher:
The Customizable class is an extension of the Configurable class and allows instances to be created by a name/ID.  Classes that extend customizable can define their Type (e.g. "TableFactory", "Cache") and  a method to instantiate them (TableFactory::CreateFromString).  Customizable objects can be registered with the ObjectRegistry and created dynamically.

Future PRs will make more types of objects extend Customizable.


include/rocksdb/customizable.h
```diff
@@ -0,0 +1,149 @@
...
+ * For custom objects, their configuration and name can be specified by:
+ * [prop]={name=X;option 1 = value1[; option2=value2...]}
+ *
+ * [prop].name=X
+ * [prop].option1 = value1
+ *
+ * [prop].name=X
+ * X.option1 =value1
+ */
+class Customizable : public Configurable {
```
zhichao-cao, line 31: Can you explain the essentialness of the customizable class. Why can't we directly use Configurable? What's the main differences?

include/rocksdb/utilities/options_type.h
```diff
@@ -406,6 +408,99 @@ class OptionTypeInfo {
...
+  // This function will call the T::CreateFromString method to create a new
+  // std::shared_ptr<T> object.
+  //
+  // @param offset The offset for the Customizable from the base pointer
+  // @param ovt How to verify this option
+  // @param flags, Extra flags specifying the behavior of this option
+  // @param _sfunc Optional function for serializing this option
+  // @param _efunc Optional function for comparing this option
+  template <typename T>
+  static OptionTypeInfo AsCustomS(int offset, OptionVerificationType ovt,
```
zhichao-cao, line 421: The name ended with S, U, P do not help to understand the function. Can we directly use like AsCustomSharedPtr, AsCustomUniquPtr, AsCustomRawPtr?

include/rocksdb/customizable.h
mrambacher: Almost every class that is Configurable is also Customizable.   A few that are not are the wrappers that deal with DBOptions and ColumnFamilyOptions.  Another is the OptionsRegistry itself.

When there are a choice of implementations for something (like different Cache or Table implementations), those are Customizable.  When there are choices, each choice has a Name() and there is a factory to create these by name (CreateFromString).

When there is a single implementation class for the options, that is Configurable.  There is no name and no factory, as there is only one choice of things to create.

A Configurable is very similar to a "struct with options" (like CompressionOptions and its associated map of OptionTypeInfo). 

Customizable implies multiple implementations; Configurable implies a single implementation.

I added a comment documenting this but it will be a pretty odd comment.  The "base class" (e.g. Table, Cache) that extends a Customizable is expected to implement certain methods (Type, CreateFromString).  The "implementation classes" are expected to implement certain methods (Name).

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,149 @@
...
+  // Returns the named instance of the Customizable as a T*, or nullptr if not
+  // found. This method uses IsInstanceOf to find the appropriate class instance
+  // and then casts it to the expected return type.
+  template <typename T>
+  const T* CastAs(const std::string& name) const {
+    if (IsInstanceOf(name)) {
+      return static_cast<const T*>(this);
+    } else {
+      const auto inner = static_cast<const Customizable*>(Inner());
+      if (inner != nullptr) {
```
zhichao-cao, line 72: I'm a little bit confused here. If the name does not match, why we need to continue the inner configurable and try if the name matches the instance.

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,149 @@
...
+        return nullptr;
+      }
+    }
+  }
+
+  // Checks to see if this Configurable is equivalent to other.
+  // This method assumes that the two objects are of the same class.
+  // @param config_options Controls how the options are compared.
+  // @param other The other object to compare to.
+  // @param mismatch If the objects do not match, this parameter contains
```
zhichao-cao, line 98: Where is the "mismatch"?

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,149 @@
...
+  bool AreEquivalent(const ConfigOptions& config_options,
+                     const Configurable* other,
+                     std::string* name) const override;
+#ifndef ROCKSDB_LITE
+  Status GetOption(const ConfigOptions& config_options, const std::string& name,
+                   std::string* value) const override;
+
+#endif  // ROCKSDB_LITE
+ protected:
+  std::string GetOptionName(const std::string& long_name) const override;
```
zhichao-cao, line 110: Some simple comments for `Getoption` and `GetOptionName`

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,149 @@
...
+#ifndef ROCKSDB_LITE
+  std::string SerializeOptions(const ConfigOptions& options,
+                               const std::string& prefix) const override;
+#endif  // ROCKSDB_LITE
+ public:
+  // Helper method for configuring a new customizable object.
+  // If base_opts are set, this is the "default" options to use for the new
+  // object whereas "new_opts" are overlaid on the base options. Returns OK if
+  // the object could be successfully configured
+  static Status ConfigureNewObject(
```
zhichao-cao, line 120: It is not very clear for me to understand the `new_opts` and `base_opts`. If `base_opts` is not set, the customizable object is configured with the `new_opts`? If `base_opts` is set, `new_opts` will be updated with the `base_opts`?

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,149 @@
...
+  // The input opt_value can be in the form of "name" or "name=value
+  // [;name=value]". The first form uses the "name" as an id with no options The
+  // latter form converts the input into a map of name=value pairs and sets "id"
+  // to the "id" value from the map.
+  // @param opt_value The value to split into id and options
+  // @param id The id field from the opt_value
+  // @param options The remaining name/value pairs from the opt_value
+  // @param default_id If specified and there is no id field in the map, this
+  // value
+  //      is returned as the ID
```
zhichao-cao, line 135: Adjust the format here.

options/customizable.cc
```diff
@@ -0,0 +1,139 @@
...
+
+std::string Customizable::SerializeOptions(const ConfigOptions& config_options,
+                                           const std::string& prefix) const {
+  std::string result;
+  std::string parent;
+  if (!config_options.IsShallow()) {
+    parent = Configurable::SerializeOptions(config_options, "");
+  }
+  if (parent.empty()) {
+    result = GetId();
```
zhichao-cao, line 46: No prefix here?

options/customizable.cc
```diff
@@ -0,0 +1,139 @@
...
+      // Don't run prepare options on the base, as we would do that on the
+      // overlay opts instead
+      Status status = object->ConfigureFromString(config_options, base_opts);
+      if (!status.ok()) {
+        return status;
+      }
+#endif  // ROCKSDB_LITE
+    }
+    if (!opts.empty()) {
+      return object->ConfigureFromMap(config_options, opts);
```
zhichao-cao, line 93: So here, if base_opts and opts are not empty, we configure the object twice, right? Some of the options in base_opts will be overwritten by opts?

options/customizable_helper.h
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
```
zhichao-cao, line 5: Can we just add those to customizable.h and customizable.cc?

include/rocksdb/customizable.h
mrambacher: There are Customizable classes that have a has-a relationship.  For example, an EncryptedEnv has-a PosixEnv in it (as the inner).  If I needed access to the underlying posix file system, there is no way to get to it without going through the inner class.  Or a DBWithTTL has a DBImpl in it somewhere and you need a means of getting at the underlying implementation.

Perhaps it might be better to rename the "CastAs" method to something like "GetAs"?  Would that help alleviate some of the confusion?

include/rocksdb/customizable.h
mrambacher: First the object is configured using base_opts and then it is configured again using new_opts. 

Say you had something like a BlockBasedTableOptions and you want to create a new one and change only one option.  To do this generically (meaning you do not know you have a BlockBasedTableOptions but have an old class), you first serialize the existing object to a string -- this is your base_opts.  Then you configure the new object using that base_opts string.  Then you override what you wanted to change with the new_opts.

options/customizable.cc
mrambacher: No prefix is correct.  If it is shallow, we are not serializing the options contained in the object and only need to serialize the ID of the object and not the object in its entirety.

In a future PR, I will introduce more ways to serialize the objects.  At the moment, most complex objects are serialized in a "struct" form (  property = { field1 = value1; field2 = value2}.  I would like to be able to serialize things "detached":
   property.field1 = value1;
   property.field2 = value2;
and additionally allow prefixes on the objects:
  table.block_based.opt1 = value1;
  table.plain.opt1 = value2;

I would also like to be able to serialize options in "flags" format for command line options (--field1=value1) or in "Dump" format for log files:
  property.field1        : value1

All of those formats and complexity to the processing of Serialize and were accounted for in a previous iteration.  I removed them for now to simplify the review and processing.

include/rocksdb/customizable.h
zhichao-cao: Maybe `GetAs` is better. CastAs is something like we force to change its type to.

include/rocksdb/customizable.h
zhichao-cao: I see, thanks for the explanation.

include/rocksdb/utilities/options_type.h
```diff
@@ -660,17 +764,26 @@ Status ParseVector(const ConfigOptions& config_options,
   result->clear();
   Status status;
 
+  // Turn off ignore_unknown_objects so we can tell if the returned
+  // object is valid or not.
+  ConfigOptions copy = config_options;
+  copy.ignore_unsupported_options = false;
```
zhichao-cao, line 770: Should we mention this change in HISTORY.md?

options/customizable.cc
mrambacher: Yes, we would configure the object twice, but not necessarily the same parts

include/rocksdb/utilities/options_type.h
mrambacher: I do not think it is necessary.  The options_type is only in the public API to satisfy some of the Configurable interfaces and is not really for public usage.

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,178 @@
...
+    return name == Name();
+  }
+
+  // Returns the named instance of the Customizable as a T*, or nullptr if not
+  // found. This method uses IsInstanceOf to find the appropriate class instance
+  // and then casts it to the expected return type.
+  template <typename T>
+  const T* GetAs(const std::string& name) const {
+    if (IsInstanceOf(name)) {
+      return static_cast<const T*>(this);
```
pdillinger, line 83: What happens if I write

    a.GetAs<B>("A")

? Do we not have enough template magic in C++11 to reject this? I expect we could with constexpr functions.

mrambacher: This is not using any template magic and is trying to rely on the smarts/accuracy of the programmer.

One way to get around it would be to require all of the T to have a static method (like kName()) and then use that as the "name":
const T* GetAs() const {
  if (IsInstanceOf(T::kName()) {
     ...
}

Note that T::kName() would/should also be returned by the const char *Name() method...

I am not good enough at C++ templates to come up with something better.  I would really like to get rid of all of the assumed casts (most are static_cast) in the code/tests though (the tests do a cast all over for DBImpl).

pdillinger: That's basically what I was thinking. For some background / forward-proofing see https://en.wikipedia.org/wiki/Concepts_(C%2B%2B)  I'm using the idea pretty extensively here: https://github.com/pdillinger/rocksdb/blob/ribbon1/util/ribbon_alg.h#L360

mrambacher: I will see if I can get it working with a kName() if that is what you think is preferred.  The "concept" qualifiers seem pretty new to C++ and I am not sure how far to take it

pdillinger: My recommendation is for informal concepts: informally specified in comments.

pdillinger: Seems much better, thanks. (Unless I missed it) more documentation on what's expected of Customizeable classes because of this template would be nice. I might have called it DynamicCast<T> but GetAs<T> is fine with me. :)

mrambacher: I will add documentation on the expectations of the Customizable class that have changed since the PR was originated.

I originally named this method CastAs and was asked to change it as part of the PR earlier.  You and @zhichao-cao can fight out which way it should be :)

zhichao-cao: DynamicCast is ok for me. Maybe change to this?

mrambacher: Internally it does not do a DynamicCast, so that names seems like a poor naming choice to me.  How about simply "Cast"?

pdillinger: The earlier objection to Cast was that "cast" can mean UB if not valid. This is a checked cast returning an appropriate object pointer if valid (ideally) or nullptr if not valid. That is a pointer dynamic cast; we're just doing it without built-in RTTI.

> Internally it does not do a DynamicCast

If we assume there is some logic to the API, why create a function called DynamicCast if you can just dynamic_cast the object in all cases where you can DynamicCast (for the same thing)? (Therefore the function must exist because dynamic_cast might not be available.)

pdillinger: Or you could call it CheckedCast.

HISTORY.md
```diff
@@ -31,6 +31,7 @@
 * Added is_full_compaction to CompactionJobStats, so that the information is available through the EventListener interface.
 * Add more stats for MultiGet in Histogram to get number of data blocks, index blocks, filter blocks and sst files read from file system per level.
 * SST files have a new table property called db_host_id, which is set to the hostname by default. A new option in DBOptions, db_host_id, allows the property value to be overridden with a user specified string, or disable it completely by making the option string empty.
+* Methods to create customizable extensions -- such as TableFactory -- are exposed directly through the Customizable base class (from which these objects inherit).  This change will allow these Customizable classes to be loaded and configured in a standard way (via CreateFromString).  More information on how to write and use Customizable classes is in the header file and on the Wiki.
```
zhichao-cao, line 34: We may not refer Wiki in HISTORY.md. Also, you can directly point out which header file you refer to.

zhichao-cao, line 34: I think we should mention virtual Configurable* Inner() is removed from configurable.h protected functions in "Public API Change". And if any user that are using it, how should they do.

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,189 @@
...
+  // @param True if the objects match, false otherwise.
+  bool AreEquivalent(const ConfigOptions& config_options,
+                     const Configurable* other,
+                     std::string* mismatch) const override;
+  Status PrepareOptions(const ConfigOptions& config_options) override;
+  bool IsPrepared() const override;
+  Status ValidateOptions(const DBOptions& db_opts,
+                         const ColumnFamilyOptions& cf_opts) const override;
+#ifndef ROCKSDB_LITE
+  // Returns the value of the option associated with the input name
```
zhichao-cao, line 134: "Get the value of...."

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,189 @@
+// Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
```
zhichao-cao, line 1: I suggest to indicate clearly the possible return value (e.g., ok case and Non-ok case) and what does that means of each function, especially those returns "Status" type. Even they are nearly the same as Configurable, it would be better to copy the comments here.

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,189 @@
...
+  std::string GetOptionName(const std::string& long_name) const override;
+#ifndef ROCKSDB_LITE
+  std::string SerializeOptions(const ConfigOptions& options,
+                               const std::string& prefix) const override;
+#endif  // ROCKSDB_LITE
+ public:
+  // Helper method for configuring a new customizable object.
+  // If base_opts are set, this is the "default" options to use for the new
+  // object.  Then any values in "new_opts" are applied to the object.
+  // Returns OK if the object could be successfully configured
```
zhichao-cao, line 161: How about not ok case?

include/rocksdb/customizable.h
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  // Checks to see if this Configurable is equivalent to other.
+  // This method assumes that the two objects are of the same class.
+  // @param config_options Controls how the options are compared.
+  // @param other The other object to compare to.
+  // @param mismatch If the objects do not match, this parameter contains
+  //      the name of the option that triggered the match failure.
+  // @param True if the objects match, false otherwise.
+  bool AreEquivalent(const ConfigOptions& config_options,
```
zhichao-cao, line 126: comments for PrepareOptions, bool IsPrepared(), Status ValidateOptions, GetOption. You may directly copy them from Configurable if they are the same.

include/rocksdb/table.h
```diff
@@ -22,7 +22,7 @@
 #include <string>
 #include <unordered_map>
```
zhichao-cao, line 24: Indicate this public API change in HISTORY.md

table/block_based/block_based_table_factory.h
```diff
@@ -46,6 +46,9 @@ class BlockBasedTableFactory : public TableFactory {
 
   ~BlockBasedTableFactory() {}
 
+  // Method to allow GetAs to work for this class
```
zhichao-cao, line 49: Developer may not be able to know GetAs when reading the code here. Better to directly explain what this function is used for. The same as other 2

HISTORY.md
mrambacher: The Inner was introduced when the Configurable class was introduced.  Since this is a new class as of the last release and is used by RocksDB extension developers (without the benefit of the Wiki documentation), I would be amazed if anyone is using this currently.  If you feel strongly it should be documented, I will but I think it is highly unlikely to be used and more likely to cause confusion.

include/rocksdb/customizable.h
mrambacher: For each method that overrides one in Configurable, I put in a see comment to the method in Configurable.  I am not in favor of every method redefining its arguments and returns from the base.  First, the base method defined the contract on what the arguments mean and what can be returned.  Derived classes must follow this contract.  Second, duplicating the comments in the derived classes means that there it is very hard to make sure the documentation stays up to date as things change.  The "see" is standard practice for doxygen/javadoc comments

table/block_based/block_based_table_factory.h
mrambacher: Updated comment to say CheckedCast

include/rocksdb/table.h
mrambacher: It is already mentioned as part of the introduction of the Customizable class.  What else would you like?

include/rocksdb/configurable.h
```diff
@@ -58,6 +58,9 @@ class Configurable {
   Configurable() : prepared_(false) {}
   virtual ~Configurable() {}
 
+  constexpr static const char* kIdPropName = "id";
```
zhichao-cao, line 61: Do we need make the two public?

include/rocksdb/utilities/options_type.h
zhichao-cao: At least, we should mention the behavior changes in the comment of ParseVector.

options/configurable.cc
```diff
@@ -403,6 +409,32 @@ Status ConfigurableHelper::ConfigureOption(
...
+      Status status = object->ConfigureFromString(config_options, base_opts);
+      if (!status.ok()) {
+        return status;
+      }
+#endif  // ROCKSDB_LITE
+    }
+    if (!opts.empty()) {
+      return object->ConfigureFromMap(config_options, opts);
+    }
+  } else if (!opts.empty()) {  // No object but no map.  This is OK
```
zhichao-cao, line 432: The comments "// No object but no map.  This is OK" should move before L435?

options/configurable_helper.h
```diff
@@ -75,6 +75,43 @@ class ConfigurableHelper {
...
+  // value is returned as the ID
+  // @return OK if the value was converted to a map succesfully and an ID was
+  // found.
+  // @return InvalidArgument if the value could not be converted to a map or
+  // there was or there is no id property in the map.
+  static Status GetOptionsMap(
+      const std::string& opt_value, std::string* id,
+      std::unordered_map<std::string, std::string>* options);
+  static Status GetOptionsMap(
+      const std::string& opt_value, const std::string& default_id,
```
zhichao-cao, line 112: I did not see any test or examples that we need to use the defualt_id. Do we need this interface? Can we just use Status GetOptionsMap(
      const std::string& opt_value, std::string* id,
      std::unordered_map<std::string, std::string>* options); and in the function, we set id to ""?

options/customizable_helper.h
```diff
@@ -0,0 +1,165 @@
...
+#pragma once
+#include <functional>
+#include <memory>
+#include <unordered_map>
+
+#include "options/configurable_helper.h"
+#include "options/options_helper.h"
+#include "rocksdb/convenience.h"
+#include "rocksdb/customizable.h"
+#include "rocksdb/status.h"
```
zhichao-cao, line 15: Better to write the comments for each functions in customizable_help.h

include/rocksdb/configurable.h
mrambacher: Moved to Helper class.

HISTORY.md
mrambacher: Removed the Inner altogether for now as it is not currently needed.  Will re-introduce it later if required.

include/rocksdb/utilities/options_type.h
mrambacher: There is no behavior change from the outside of this method.  This allows errors reported from inside the method to be differentiated.  Previously, if ignore_unsupported_options=true, the status below would return OK at times that were not OK (because we ignored the error) -- we would not detect NotFound in the inner loop.  Now the inner loop can return NotFound and the outer code can handle it appropriately,

zhichao-cao (approved):
LGTM, thanks for the contribution and addressing the comments!

PR #6591, opened 2020-03-24
Extensions to the ObjectRegistry 

mrambacher:
The ObjectRegistry is now a member variable of the DBOptions and ConfigOptions.  This allows a registry to be on a per-configuration basis.

The ObjectLibrary was expanded to include a "Program" and a "Loaded" library subclass.  The "Loaded" library is initialized from a shared library.  The "Program" library is initialized from an in-process method.


include/rocksdb/utilities/object_registry.h
```diff
@@ -76,29 +131,75 @@ class ObjectLibrary {
...
   static std::shared_ptr<ObjectRegistry> NewInstance();
 
-  ObjectRegistry();
+  // Makes a copy of this registry.
+  virtual std::shared_ptr<ObjectRegistry> Clone() const = 0;
 
-  void AddLibrary(const std::shared_ptr<ObjectLibrary>& library) {
-    libraries_.emplace_back(library);
-  }
+  // Creates a new local library, registering the factories in registrar
```
zhichao-cao, line 165: What's the difference between local library and dynamic library? May need some explanations and examples.

include/rocksdb/utilities/object_registry.h
```diff
@@ -62,10 +79,48 @@ class ObjectLibrary {
     FactoryFunc<T> factory_;
   };  // End class FactoryEntry
  public:
+  // Loads the ObjectLibrary specified by the input value into result
+  // Libraries can be "default", "local", or "dynamic".
```
zhichao-cao, line 83: How can a user decide whether a local, or dynamic should be used?

include/rocksdb/utilities/object_registry.h
```diff
@@ -62,10 +79,48 @@ class ObjectLibrary {
...
+  static Status CreateFromString(const ConfigOptions& config_options,
+                                 const std::string& value,
+                                 std::shared_ptr<ObjectLibrary>* result);
+  static const char* Type() { return "ObjectLibrary"; }
+  virtual ~ObjectLibrary() {}
+
   // Finds the entry matching the input name and type
   const Entry* FindEntry(const std::string& type,
                          const std::string& name) const;
-  void Dump(Logger* logger) const;
```
zhichao-cao, line 68: Dump is in the public API, if some one is using it, how they can write out the info in registry to the log?

include/rocksdb/utilities/object_registry.h
```diff
@@ -62,10 +79,48 @@ class ObjectLibrary {
...
+  //      - "library" specifies the library to load
+  //      - "method"  specifies the method in that library to invoke
+  //      - "arg"     specifies the argument to pass to that method
+  // @param config_options Controls how the library is loaded
+  // @param value The name and optional properties describing the library
+  //      to load.
+  // @param result On success, returns the loaded library
+  // @return OK if the library  was successfully loaded.
+  // @return not-OK if the load failed.
+  static Status CreateFromString(const ConfigOptions& config_options,
```
zhichao-cao, line 100: It is a little confusion to use "CreateFromString", do you have other better name? Or you want to align it with the "String" in Configurable?

include/rocksdb/utilities/object_registry.h
```diff
@@ -62,10 +79,48 @@ class ObjectLibrary {
...
+
   // Finds the entry matching the input name and type
   const Entry* FindEntry(const std::string& type,
                          const std::string& name) const;
-  void Dump(Logger* logger) const;
+
+  // Returns the number of registered types for this library.
+  // If specified (not-null), types is updated to include the names of the
+  // registered types.
+  size_t GetRegisteredTypes(std::unordered_set<std::string>* types) const;
```
zhichao-cao, line 113: If the API is not used in the current stage, do you think we can add it later to the code base?

include/rocksdb/utilities/object_registry.h
mrambacher: I will add a comment later.

A dynamic library is one that was loaded into the application explicitly.  Methods are located by name in that external library.

A "local library" searches the current executable for the method in question, if enabled/allowed.  Methods in a "local library" could be searched by name (if allowed) or their address might be known.  In the latter case, the local library could not be reconstructed on restart.

If you have a better name for the "local library", feel free to suggest one.

include/rocksdb/utilities/object_registry.h
mrambacher: If you are loading an external method, it will be a dynamic library.  If you are loading a method from the current executable, it could be put into the local or default library.

include/rocksdb/utilities/object_registry.h
mrambacher: All Customizable classes will have a CreateFromString method.  This is the "standard" constructor method used by the options hierarchy

include/rocksdb/utilities/object_registry.h
mrambacher: It could be added later, but is useful if we want to put back Dump...

include/rocksdb/utilities/object_registry.h
zhichao-cao: Oh, yeah, I have the impression that CreateFromString is the member function.

include/rocksdb/utilities/object_registry.h
zhichao-cao: I'm thinking another name to replace "Local". "local" is not an exclusion case of "dynamic"......

include/rocksdb/utilities/object_registry.h
```diff
@@ -62,10 +79,48 @@ class ObjectLibrary {
...
+
+  // Returns the number of registered names for the input type
+  // If specified (not-null), names is updated to include the names for the type
+  size_t GetRegisteredNames(const std::string& type,
```
zhichao-cao, line 117: Is there any order of the names being returned? Maybe just explain it in comment is fine.

include/rocksdb/utilities/object_registry.h
```diff
@@ -76,29 +131,75 @@ class ObjectLibrary {
...
+  virtual std::shared_ptr<ObjectRegistry> Clone() const = 0;
 
-  void AddLibrary(const std::shared_ptr<ObjectLibrary>& library) {
-    libraries_.emplace_back(library);
-  }
+  // Creates a new local library, registering the factories in registrar
+  // @param registrar   The registration function to invoke
+  // @param method      The name of the registration function
+  // @param arg         Argument to supply to the registration function.
+  virtual void AddLocalLibrary(const RegistrarFunc& registrar,
```
zhichao-cao, line 169: I'm little confused by the two "AddLocalLibrary" and two "AddDynamicLibrary". Can you explain more?

include/rocksdb/utilities/object_registry.h
```diff
@@ -189,17 +290,25 @@ class ObjectRegistry {
...
-  // Dump the contents of the registry to the logger
-  void Dump(Logger* logger) const;
+  // Returns the number of registered types for this registry.
+  // If specified (not-null), types is updated to include the names of the
+  // registered types.
+  virtual size_t GetRegisteredTypes(
+      std::unordered_set<std::string>* types) const = 0;
+  // Returns the number of registered names for the input type
+  // If specified (not-null), names is updated to include the names for the type
+  virtual size_t GetRegisteredNames(const std::string& type) const = 0;
```
zhichao-cao, line 300: It is confusion that the function name is "GetRegisteredNames" but returns the number of names of certain type.

include/rocksdb/utilities/object_registry.h
mrambacher: For local, the one signature takes in a name (the method argument) of a registrar function and attempts to find that function in the executable using the Env DynamicLibrary methods.  If the function was not found, an error is returned.  

The other method takes in the RegistrarFunction (a std::function) and invokes it to register the types in the library.  The first method will end up calling this one if the function is found.

Comparable signatures exist for the dynamic libraries, using a DynamicLibrary to locate the registrar function and the Env to locate the dynamic library.

The "string-based" signature must be used in order to recover/restore the ObjectRegistry from a string (or options file).  The "registrar" signature is what does the actual registration, and is useful for programs who wish to register their types at "compile-time".

include/rocksdb/utilities/object_registry.h
mrambacher: Would "GetNamesForType" be better?

utilities/object_registry.cc
```diff
@@ -26,61 +55,412 @@ const ObjectLibrary::Entry *ObjectLibrary::FindEntry(
...
+    for (const auto &library : options_.libraries) {
+      count += library->GetRegisteredNames(type, names);
+    }
+    return count;
+  }
+
+  // Returns the total number of factories registered for this library.
+  // This method returns the sum of all factories registered for all types.
+  // @param num_types returns how many unique types are registered.
+  size_t GetFactoryCount(size_t *num_types) const override {
```
zhichao-cao, line 217: Make it clear that, the returned value is the sum of all factories registered for all types, "num_types" only contains the types

utilities/object_registry.cc
```diff
@@ -26,61 +55,412 @@ const ObjectLibrary::Entry *ObjectLibrary::FindEntry(
...
+         iter != options_.libraries.crend(); ++iter) {
+      const auto *entry = iter->get()->FindEntry(type, name);
+      if (entry != nullptr) {
+        return entry;
+      }
+    }
+    return nullptr;
+  }
+
+  Status ConfigureOptions(
```
zhichao-cao, line 253: Add some simple comments here for reference.

env/env.cc
```diff
@@ -53,18 +61,28 @@ Status Env::LoadEnv(const std::string& value, Env** result) {
...
   std::unique_ptr<Env> uniq_guard;
   std::string err_msg;
   assert(guard != nullptr);
-  env = ObjectRegistry::NewInstance()->NewObject<Env>(value, &uniq_guard,
-                                                      &err_msg);
+  env = config_options.registry->NewObject<Env>(value, &uniq_guard, &err_msg);
   if (!env) {
-    s = Status::NotFound(std::string("Cannot load ") + Env::Type() + ": " +
-                         value);
+    s = Status::NotSupported(std::string("Cannot load ") + Env::Type() + ": " +
```
zhichao-cao, line 84: In HISTORY.md we need to mention the return code changes here.

env/env_posix.cc
```diff
@@ -110,7 +110,7 @@ class PosixDynamicLibrary : public DynamicLibrary {
       return Status::OK();
     } else {
       char* err = dlerror();
-      return Status::NotFound("Error finding symbol: " + sym_name, err);
+      return Status::InvalidArgument("Error finding symbol: " + sym_name, err);
```
zhichao-cao, line 113: This change is not related to current OBjectRegistry right?

include/rocksdb/utilities/object_registry.h
zhichao-cao: Thanks for the explanation. Maybe you can add those to the comments of "AddLocalLibrary" and "AddDynamicLibrary"

include/rocksdb/utilities/object_registry.h
zhichao-cao: Yes, let's use "GetNamesForType"

include/rocksdb/utilities/object_registry.h
```diff
@@ -189,17 +290,25 @@ class ObjectRegistry {
...
+  virtual size_t GetRegisteredNames(const std::string& type,
+                                    std::vector<std::string>* names) const = 0;
 
- private:
-  const ObjectLibrary::Entry* FindEntry(const std::string& type,
-                                        const std::string& name) const;
+  // Returns the total number of factories registered for this library.
+  // This method returns the sum of all factories registered for all types.
+  // @param num_types returns how many unique types are registered.
+  virtual size_t GetFactoryCount(size_t* num_types) const = 0;
```
zhichao-cao, line 307: How about "GetFactoryNums" for this one?

options/db_options.cc
```diff
@@ -465,8 +478,24 @@ class DBOptionsConfigurable : public MutableDBConfigurable {
...
       std::unordered_map<std::string, std::string>* unused) override {
-    Status s = ConfigurableHelper::ConfigureOptions(config_options, *this,
-                                                    opts_map, unused);
+    // Keep a copy of the registry in case we have to revert it
+    ConfigOptions copy = config_options;
+    std::unordered_map<std::string, std::string> left = opts_map;
+
+    if (copy.registry == immutable_.object_registry) {
+      // The registry from the ConfigOptions matches that of the DB
+      immutable_.object_registry = copy.registry->Clone();
```
zhichao-cao, line 487: I'm not very clear about the logic here. If `copy.registry == immutable_.object_registry` why we need 'immutable_.object_registry = copy.registry->Clone()'.

utilities/object_registry.cc
```diff
@@ -26,61 +55,412 @@ const ObjectLibrary::Entry *ObjectLibrary::FindEntry(
...
+                                        &object_registry_library_type_info);
+  }
+
+  ObjectRegistryImpl(const ObjectRegistryImpl &that) : options_(that.options_) {
+    ConfigurableHelper::RegisterOptions(*this, "Libraries", &options_.libraries,
+                                        &object_registry_library_type_info);
+  }
+
+  // Makes a copy of this registry.
+  std::shared_ptr<ObjectRegistry> Clone() const override {
```
zhichao-cao, line 148: I'm not sure `Clone` is a good function name here. We just share the same OBjectRegistry instance pointer right? Not clone a new instance.

utilities/object_registry.cc
```diff
@@ -26,61 +55,412 @@ const ObjectLibrary::Entry *ObjectLibrary::FindEntry(
...
+      Status s = library->PrepareOptions(config_options);
+      if (!s.ok()) {
+        return s;
+      }
+    }
+    return Configurable::PrepareOptions(config_options);
+  }
+
+ protected:
+  // Searches (from back to front) the libraries looking for the
```
zhichao-cao, line 238: In comment "looking for the an entry" no "the"?

utilities/object_registry.cc
```diff
@@ -26,61 +55,412 @@ const ObjectLibrary::Entry *ObjectLibrary::FindEntry(
...
+
+  Status ConfigureOptions(
+      const ConfigOptions &config_options,
+      const std::unordered_map<std::string, std::string> &opt_map,
+      std::unordered_map<std::string, std::string> *unused) override {
+    // We want to control when any libraries are initialized.
+    // Turn off prepare while we are configuring the registry.
+    // Once the registry is completely configured, we can go back and configure
+    // the libraries.
+    ConfigOptions copy = config_options;
```
zhichao-cao, line 261: I several places, the variable name "copy" and "left" are used. Maybe replacing with more explicit name.

include/rocksdb/utilities/object_registry.h
```diff
@@ -13,17 +13,34 @@
...
 
-class ObjectLibrary {
+// The signature of the function for loading factories
+// into an object library.  This method is expected to register
+// factory functions in the supplied ObjectLibrary.
+// @param library   The library to load factories into.
+// @param arg       Argument to the library loader
+using RegistrarFunc = std::function<void(ObjectLibrary&, const std::string&)>;
+
+class ObjectLibrary : public Customizable {
```
zhichao-cao, line 43: Add some comments to explain 1) what is ObejctLibrary, 2) why it is inherited from Customizable.

zhichao-cao (approved):
LGTM, CC @ajkr if he has further comments?

include/rocksdb/utilities/object_registry.h
```diff
@@ -190,16 +234,46 @@ class ObjectRegistry {
...
+  virtual std::shared_ptr<ObjectLibrary> AddProgramLibrary(
+      const std::string& id, const RegistrarFunc& registrar,
+      const std::string& arg) = 0;
+
+  // Creates a new ObjectLibrary within this ObjectRegistry.
+  // Loads the registrar func named "method" from the input library and, upon
+  // success, invokes registrar(arg) to register the appropriate factories in
+  // the new library. Returns OK if method was found in the library and non-OK
+  // otherwise.
+  virtual Status AddLoadedLibrary(
```
ajkr, line 252: Is there a way to verify the plugin was built against compatible headers?
Also I wonder if there is any effort to expand our ABI compatibility guarantees -- that should make this approach more compelling. Currently we only do best-effort to uphold ABI compatibility across patch versions within a minor version.

ajkr (commented):
Is the title up to date? Somehow I couldn't find any of the things mentioned in the title.

edit: Upon further thought I think we should prioritize #6614 over this by a lot. 6614 has general benefit to extension users/implementors such as FB, whereas this PR has less general benefits. I know you already submitted 6614 a year ago, so we just need to review our review priorities...

include/rocksdb/utilities/object_registry.h
mrambacher: This is not part of the current submission.  The PR was broken into more pieces and this will be in a follow-on

mrambacher:
> Is the title up to date? Somehow I couldn't find any of the things mentioned in the title.
> 
> edit: Upon further thought I think we should prioritize #6614 over this by a lot. 6614 has general benefit to extension users/implementors such as FB, whereas this PR has less general benefits. I know you already submitted 6614 a year ago, so we just need to review our review priorities...

The title and comment were out of date.  The scope of the PR was narrowed after the "plugin" work came into existence.

I plan to update #6614 shortly and submit it as a bunch of separate PRs (few Customizable objects per class).

include/rocksdb/utilities/object_registry.h
mrambacher: In #7949 I propose how to do better ABI validation.  I can remove this method until that is approved if that helps this one move forward.

mrambacher: I removed the AddLoadedLibrary method for now.

options/db_options.cc
mrambacher: Code restructured.  No longer needed

env/env_posix.cc
mrambacher: Removed from this PR

ajkr (changes requested):
The title/description appear to not reflect the contents of the PR. They should contain a convincing reason to add complexity to the public API.

edit: I was not able to figure out the use case from reading the code. If it is not strictly needed, it can save reviewer and maintenance effort to not pursue it.

riversand963:
Haven't reviewed the entire series of PRs about configuration and object registry. Read about dlopen/dlsym a bit more..
If dlsym is used to access symbols defined in the shared library, then it seems that we cannot enforce function signature check on functions loaded from the shared library.

Will this be an issue if RocksDB changes the API in a future release?  Can we have some compile/link time check?

riversand963:
Something else not related to this PR.
The use of `std::regex` (https://github.com/facebook/rocksdb/blob/master/include/rocksdb/utilities/object_registry.h#L61) is treated as an error by our internal linter due to its recursion on program stack, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61582, and it should be removed from code that can potentially be used in production.

mrambacher:
Abandoning this PR now in favor of other approaches.

PR #6597, opened 2020-03-25
Add blob files to VersionStorageInfo/VersionBuilder

ltamasi:
Summary:
The patch adds a couple of classes to represent metadata about
blob files: `SharedBlobFileMetaData` contains the information elements
that are immutable (once the blob file is closed), e.g. blob file number,
total number and size of blob files, checksum method/value, while
`BlobFileMetaData` contains attributes that can vary across versions like
the amount of garbage in the file. There is a single `SharedBlobFileMetaData`
for each blob file, which is jointly owned by the `BlobFileMetaData` objects
that point to it; `BlobFileMetaData` objects, in turn, are owned by `Version`s
and can also be shared if the (immutable _and_ mutable) state of the blob file
is the same in two versions.

In addition, the patch adds the blob file metadata to `VersionStorageInfo`, and extends
`VersionBuilder` so that it can apply blob file related `VersionEdit`s (i.e. those
containing `BlobFileAddition`s and/or `BlobFileGarbage`), and save blob file metadata
to a new `VersionStorageInfo`. Consistency checks are also extended to ensure
that table files point to blob files that are part of the `Version`, and that all blob files
that are part of any given `Version` have at least some _non_-garbage data in them.

Test Plan:
`make check`


db/version_builder.cc
```diff
@@ -333,9 +478,104 @@ class VersionBuilder::Rep {
...
+    assert(vstorage);
+    assert(meta);
+
+    if (meta->GetGarbageBlobCount() < meta->GetTotalBlobCount()) {
+      vstorage->AddBlobFile(meta);
+    }
+  }
+
+  void SaveBlobFilesTo(VersionStorageInfo* vstorage) const {
+    // Merge the base version with the changes (edits) applied.
```
riversand963: Should this line be moved outside this method?

db/version_builder.cc
```diff
@@ -140,6 +146,57 @@ class VersionBuilder::Rep {
...
+
+    return std::shared_ptr<BlobFileMetaData>();
+  }
+
+  Status CheckConsistencyOfOldestBlobFileReference(
+      const VersionStorageInfo* vstorage, uint64_t blob_file_number) const {
+    assert(vstorage);
+
+    // TODO: remove this check once we actually start recoding metadata for
+    // blob files in the MANIFEST.
```
riversand963: Not sure I am following this. What metadata in MANIFEST will make this check unnecessary? By this check, I mean line 180 to 182.

db/version_builder.cc
```diff
@@ -217,6 +295,23 @@ class VersionBuilder::Rep {
...
       }
     }
+
+    // Make sure that all blob files in the version have non-garbage data.
+    const auto& blob_files = vstorage->GetBlobFiles();
+    for (const auto& pair : blob_files) {
+      const auto& blob_file_meta = pair.second;
+      assert(blob_file_meta);
+
+      if (!(blob_file_meta->GetGarbageBlobCount() <
```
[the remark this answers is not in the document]
ltamasi: Sure. Will change that.

db/version_builder.cc
ltamasi: Thanks for the review @riversand963 !

ltamasi: > Not sure I am following this. What metadata in MANIFEST will make this check unnecessary? By this check, I mean line 180 to 182.

It refers to the blob file metadata I'm in the process of adding. With this patch, the plumbing will be largely in place; however, the logic is currently dormant. On the other hand, those `oldest_blob_file_number` references have been there for a while, so unless I sort of disable these checks for now, all consistency checks would fail when BlobDB is used.

riversand963: Got it. I guess next step will be modifying `LogAndApply` and related callers to write BlobAdd and BlobGarbage to MANIFEST. Since `Recover` uses `VersionBuilder`, I guess that part is easier to update.

PR #6599, opened 2020-03-26
Add unit test for TransactionLockMgr

cheng-chang:
Although there are tests related to locking in transaction_test, this new test directly tests against TransactionLockMgr.

Test Plan:
make transaction_lock_mgr_test && ./transaction_lock_mgr_test

### DIFF
--- a/utilities/transactions/transaction_lock_mgr_test.cc
+++ b/utilities/transactions/transaction_lock_mgr_test.cc
@@ -257,12 +257,12 @@ TEST_F(TransactionLockMgrTest, Deadlock) {
   ASSERT_EQ(deadlocks.size(), 2u);
 
   ASSERT_EQ(deadlocks[0].m_txn_id, txn1->GetID());
-  ASSERT_EQ(deadlocks[0].m_cf_id, 1);
+  ASSERT_EQ(deadlocks[0].m_cf_id, 1u);
   ASSERT_TRUE(deadlocks[0].m_exclusive);
   ASSERT_EQ(deadlocks[0].m_waiting_key, "k2");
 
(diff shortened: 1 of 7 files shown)


utilities/transactions/transaction_lock_mgr_test.cc
```diff
@@ -0,0 +1,348 @@
...
+}
+
+TEST_F(TransactionLockMgrTest, LockDowngrade) {
+  // Tests that a txn can acquire a shared lock after acquiring an exclusive
+  // lock on the same key.
+  locker_->AddColumnFamily(1);
+  auto txn = NewTxn();
+  ASSERT_OK(locker_->TryLock(txn, 1, "k", env_, true));
+  ASSERT_OK(locker_->TryLock(txn, 1, "k", env_, false));
+  delete txn;
```
lth, line 164: I don't think lock downgrade is actually supported, so would be nice to verify that lock is still exclusively locked.

Might be good to do this for some of the other test cases too.

utilities/transactions/transaction_lock_mgr_test.cc
```diff
@@ -0,0 +1,348 @@
...
+  port::Thread t(f);
+
+  while (!reached.load()) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->DisableProcessing();
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->ClearAllCallBacks();
+
+  return t;
+}
```
lth, line 214: I feel like this section should come after the "SharedLocks" test.

PR #6611, opened 2020-03-30
Fix potential memory leak in table_test

zhichao-cao:
The checksum generator should be released if file_writer fails to reset the pointer.

Test: pass make asan_check


table/table_test.cc
```diff
@@ -1190,6 +1190,10 @@ class FileChecksumTestHelper {
   void SetFileChecksumGenerator(FileChecksumGenerator* checksum_generator) {
     if (file_writer_ != nullptr) {
       file_writer_->TEST_SetFileChecksumGenerator(checksum_generator);
+    } else {
+      if (checksum_generator != nullptr) {
```
riversand963: This `if` can be removed.

PR #6614, opened 2020-03-30
Makes many classes of objects into Customizable, allowing them to be saved and created

mrambacher:
This PR makes the FilterPolicy, CompactionFilter, CompactionFilterFactory, MergeOperator, Cache, MemTableRepFactory, Comparator, Statistics, and FlushBlockPolicyFactory into Customizable objects.  This change allows specializations of these objects to be saved and created in the options file/string.

The base class of these objects all have a Type and CreateFromString methods.  When registered as an option, the AsCustomX methods off OptionTypeInfo can be used to control how these objects are created and stored.


table/block_based/filter_policy_internal.h
```diff
@@ -35,11 +35,7 @@ class BuiltinFilterBitsBuilder : public FilterBitsBuilder {
   virtual double EstimatedFpRate(size_t keys, size_t bytes) = 0;
 };
 
-// RocksDB built-in filter policy for Bloom or Bloom-like filters.
-// This class is considered internal API and subject to change.
-// See NewBloomFilterPolicy.
-class BloomFilterPolicy : public FilterPolicy {
- public:
+struct BloomFilterOptions {
```
pdillinger: As much as possible, I would not like to perpetuate or re-enforce the existing configuration pattern for Bloom filters: which is essentially just a bits/key (but also the option to use old block-based filter ☹️). But I'm in the process of more research and development on what should be done instead. I don't think you're necessarily creating greater friction for upcoming changes but I wanted to bring up the concern.

Also, you've seen that the "fixed implementation" modes are not exposed to the API user. (This header is internal.) Is it weird to have them part of an options object? Is Customizable only supposed to capture user-serviceable options?

mrambacher: I assume you referring to the "Mode" in the BloomFilterOptions?  Originally, I attempted to restrict the mode settings to be only the "UserMode" values but then some of the tests will not pass.

Note that the BloomFilterOptions is still internal, but its "string representation" is now potentially public.  If necessary, it would be possible to keep the "bool" option and hide the enum setting.  If this is important, please make a note in the PR and I will address it when that PR is under more thorough review.

The "goal" of the Customizable/Configurable is to be able to save and restore the settings that the user may specify somehow via the options.  The goal is to be able to restore the state of DBOptions from a string to the equivalent of what it was before.  In other words, if I take any object (such as a DBOptions, ColumnFamilyOptions, TableFactory, FilterPolicy, etc), I can convert that object to a string (GetOptionString) and recreate the equivalent object/settings (ConfigureFromString).  This will allow options to be saved to a file and the exact options be recreated thru configuration rather than programmatically.  And with the addition of ObjectRegistry, Customizable, and such, we can add features to RocksDB without additional programming.

table/block_based/filter_policy.cc
```diff
@@ -756,4 +803,50 @@ FilterBuildingContext::FilterBuildingContext(
...
+Status FilterPolicy::CreateFromString(
+    const std::string& value, const ConfigOptions& opts,
+    std::shared_ptr<const FilterPolicy>* policy) {
+  const std::string kBloomName = "bloomfilter:";
+  Status s;
+  if (value == kNullptrString) {
+    policy->reset();
+#ifndef ROCKSDB_LITE
+  } else if (value.compare(0, kBloomName.size(), kBloomName) == 0) {
+    size_t pos = value.find(':', kBloomName.size());
```
pdillinger: Here the string representation is the same as the already public string representation from ParseBlockBasedTableOption, but can you explain the mapping of "rocksdb.BuiltinBloomFilter" -> new BloomFilterPolicy(0, BloomFilterOptions::kDeprecatedBlock) in LoadFilterPolicy?

mrambacher: - The "value" can look like the old string (bloomfilter:*), in which case it is handled like before;
- If the "value" is "rocksdb.BuiltinBloomFilter" or contains "id=rocksdb.BuiltinBloomFilter", we care a BloomFilterPolicy.
- If the value contains other properties (bits_per_key, mode), those are set in the FilterPolicy (via ConfigureFromMap/String).
- After all of the properties are configured, the PrepareOptions method is invoked to finish the setup of the FilterPolicy. 

A fully qualified BloomFilterPolicy would now look something like "{id=rocksdb.BuiltinBloomFilter; bits_per_key=10; mode=auto;}".  This representation matches how all of the other properties are represented in the options file.

mrambacher:
I believe many (if not all) of these changes and classes were made Customizable in individual PRs.  Closing this PR because of that.

PR #6616, opened 2020-03-31
Hex encode keys in compaction flush logs

sfackler:
The raw key bytes are currently dumped directly into the log messages,
which is not ideal if the keys aren't ASCII strings. Null bytes in
particular can cut off bits of the message early.


sfackler:
As an example, here's the hex dump of the bad log line that caused me to track this down:

```

sfackler:
It looks like the failing CI build just timed out.

PR #6626, opened 2020-04-01
Add a compact option max_compaction_bytes

huachaohuang:
I found that manual compact sometimes picks some very large compactions. I haven't figure out why but using a small `max_compaction_bytes` can prevent it. I think it is reasonable to add `max_compaction_bytes` as a manual compact option too.

### DIFF
--- a/db/compaction/compaction_picker.cc
+++ b/db/compaction/compaction_picker.cc
@@ -775,6 +775,11 @@ Compaction* CompactionPicker::CompactRange(
     return nullptr;
   }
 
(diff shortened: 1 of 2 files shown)


ajkr:
I ran into a similar problem before too. In my case it was because we add files to the input level until it reaches `max_compaction_bytes`, and we completely forgot to count the bytes in files that overlap on the output level (https://github.com/facebook/rocksdb/blob/8a1da56b961baf1b491b00af67f0fe8ab5c9afc5/db/compaction/compaction_picker.cc#L673-L683). It felt somewhat risky to fix it since we have large use cases with low values for `max_compaction_bytes`.

PR #6629, opened 2020-04-01
Revamp cache_bench to resemble a real workload

pdillinger:
Summary: I suspect LRUCache could use some optimization, and to support
such an effort, a good benchmarking tool is needed. The existing
cache_bench was heavily skewed toward insertion and lookup misses, and
did not saturate memory with other work. This change should improve
those things to better resemble a real workload.

(All below using clang compiler, for some consistency, but not
necessarily same version and settings.)

The real workload is from production MySQL on RocksDB, filtering stacks
containing "LRU", "ShardedCache" or "CacheShard."
Lookup inclusive: 66%
Insert inclusive: 17%
Release inclusive: 15%

An alternate simulated workload is MySQL running a LinkBench read test:
Lookup inclusive: 54%
Insert inclusive: 24%
Release inclusive: 21%

cache_bench default settings, prior to this change:
Lookup inclusive: 35.8%
Insert inclusive: 63.6%
Release inclusive: 0%

cache_bench after this change (intended as somewhat "tighter" workload
than average production, more like LinkBench):
Lookup inclusive: 52%
Insert inclusive: 20%
Release inclusive: 26%

And top exclusive stacks (portion of stack samples as filtered above):
Production MySQL:
LRUHandleTable::FindPointer: 25.3%
rocksdb::operator==: 15.1%  <-- Slice ==
LRUCacheShard::LRU_Remove: 13.8%
ShardedCache::Lookup: 8.9%
__pthread_mutex_lock: 7.1%
LRUCacheShard::LRU_Insert: 6.3%
MurmurHash64A: 4.8%  <-- Since upgraded to XXH3p
...

Old cache_bench:
LRUHandleTable::FindPointer: 23.6%
__pthread_mutex_lock: 15.0%
__pthread_mutex_unlock_usercnt: 11.7%
__lll_lock_wait: 8.6%
__lll_unlock_wake: 6.8%
LRUCacheShard::LRU_Insert: 6.0%
ShardedCache::Lookup: 4.4%
LRUCacheShard::LRU_Remove: 2.8%
...
rocksdb::operator==: 0.2%  <-- Slice ==
...

New cache_bench:
LRUHandleTable::FindPointer: 22.8%
__pthread_mutex_unlock_usercnt: 14.3%
rocksdb::operator==: 10.5%  <-- Slice ==
LRUCacheShard::LRU_Insert: 9.0%
__pthread_mutex_lock: 5.9%
LRUCacheShard::LRU_Remove: 5.0%
...
ShardedCache::Lookup: 2.9%
...

So there's a bit more lock contention in the benchmark than in
production, but otherwise looks similar enough to me. At least it's a
big improvement over the existing code.

Test Plan: No production code changes, ran cache_bench with ASAN

### DIFF
--- a/cache/cache_bench.cc
+++ b/cache/cache_bench.cc
@@ -151,7 +151,7 @@ struct KeyGen {
 
 char* createValue(Random64& rnd) {
   char* rv = new char[FLAGS_value_bytes];
-  // And fill with some filler data
+  // Fill with some filler data, and take some CPU time
   for (uint32_t i = 0; i < FLAGS_value_bytes; i += 8) {
     EncodeFixed64(rv + i, rnd.Next());
   }
--- a/cache/cache_bench.cc
+++ b/cache/cache_bench.cc
@@ -14,6 +14,7 @@ int main() {
 #include <stdio.h>
 #include <sys/types.h>
 #include <cinttypes>
+#include <limits>
 
 #include "port/port.h"
 #include "rocksdb/cache.h"
@@ -27,10 +28,12 @@ int main() {
 
 using GFLAGS_NAMESPACE::ParseCommandLineFlags;
 
-static const uint32_t KB = 1024;
+static constexpr uint32_t KiB = uint32_t{1} << 10;
+static constexpr uint32_t MiB = KiB << 10;
+static constexpr uint64_t GiB = MiB << 10;
 
 DEFINE_uint32(threads, 16, "Number of concurrent threads to run.");
-DEFINE_uint64(cache_size, 1 * KB * KB * KB,
+DEFINE_uint64(cache_size, 1 * GiB,
               "Number of bytes to use as a cache of uncompressed data.");
 DEFINE_uint32(num_shard_bits, 6, "shard_bits.");
 
@@ -38,7 +41,7 @@ DEFINE_double(resident_ratio, 0.25,
               "Ratio of keys fitting in cache to keyspace.");
 DEFINE_uint64(ops_per_thread, 0,
               "Number of operations per thread. (Default: 5 * keyspace size)");
-DEFINE_uint32(value_bytes, 8 * KB, "Size of each value added.");
+DEFINE_uint32(value_bytes, 8 * KiB, "Size of each value added.");
 
 DEFINE_uint32(skew, 5, "Degree of skew in key selection");
 DEFINE_bool(populate_cache, true, "Populate cache before operations");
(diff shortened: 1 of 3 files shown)


cache/cache_bench.cc
```diff
@@ -220,40 +265,76 @@ class CacheBench {
...
+        if (handle) {
+          // do something with the data
+          result += NPHash64(reinterpret_cast<char*>(cache_->Value(handle)),
+                             FLAGS_value_bytes);
+        } else {
+          // do insert
+          cache_->Insert(key, createValue(thread->rnd), FLAGS_value_bytes,
+                         &deleter, &handle);
+        }
+      } else if (prob_op -= FLAGS_lookup_insert_percent,
```
ltamasi, line 290: `((prob_op -= FLAGS_lookup_insert_percent) < FLAGS_insert_percent)` ? I think the comma operator would trip a lot of people up. Alternatively, we could build a CDF (i.e. precompute `FLAGS_lookup_insert_percent+FLAGS_insert_percent` etc.) so we don't have to subtract.

cache/cache_bench.cc
```diff
@@ -138,18 +179,21 @@ class CacheBench {
     } else {
       cache_ = NewLRUCache(FLAGS_cache_size, FLAGS_num_shard_bits);
     }
+    max_key_ = static_cast<uint64_t>(FLAGS_cache_size / FLAGS_resident_ratio /
```
ltamasi, line 182: This could be done in the member initializer list above.

cache/cache_bench.cc
```diff
@@ -220,40 +265,76 @@ class CacheBench {
...
     for (uint64_t i = 0; i < FLAGS_ops_per_thread; i++) {
-      uint64_t rand_key = thread->rnd.Next() % FLAGS_max_key;
-      // Cast uint64* to be char*, data would be copied to cache
-      Slice key(reinterpret_cast<char*>(&rand_key), 8);
-      int32_t prob_op = thread->rnd.Uniform(100);
-      if (prob_op >= 0 && prob_op < FLAGS_insert_percent) {
+      Slice key = gen.GetRand(thread->rnd, max_key_);
+      uint32_t prob_op = static_cast<uint32_t>(thread->rnd.Uniform(100U));
+      if (prob_op >= 0 && prob_op < FLAGS_lookup_insert_percent) {
+        if (handle) {
```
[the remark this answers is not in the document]
pdillinger: It's intentional, in order to hold handles for non-trivial time (like happens in the real world!), though it's not so effective with almost all operations now recycling handle.

cache/cache_bench.cc
pdillinger: I only gave this enough attention to fix the buggy construction in the existing code (`&&` before `-=`). ;)

PR #6634, opened 2020-04-02
Allow sst_dump to check size of different compression levels and report time

akankshamahajan15:
Summary : 1. Add two arguments --compression_level_from and --compression_level_to to check
	  the compression size with different compression level in the given range. Users must
          specify one compression type else it will error out. Both from and to levels must
	  also be specified together.
	  2. Display the time taken to compress each file with different compressions by default.

Test Plan : make -j64 check

### DIFF
--- a/tools/sst_dump_tool.cc
+++ b/tools/sst_dump_tool.cc
@@ -9,6 +9,7 @@
 #include "tools/sst_dump_tool_imp.h"
 
 #include <cinttypes>
+#include <chrono>
 #include <iostream>
 #include <map>
 #include <memory>
@@ -207,7 +208,32 @@ uint64_t SstFileDumper::CalculateCompressedTableSize(
 int SstFileDumper::ShowAllCompressionSizes(
     size_t block_size,
     const std::vector<std::pair<CompressionType, const char*>>&
-        compression_types) {
+      compression_types,
+    int32_t compress_level_from,
+    int32_t compress_level_to) {
+
+  fprintf(stdout, "Block Size: %" ROCKSDB_PRIszt "\n", block_size);
+  for (auto& i : compression_types) {
+    if (CompressionTypeSupported(i.first)) {
+      fprintf(stdout, "Compression: %-24s\n", i.second);
+      CompressionOptions compress_opt;
+      for(int32_t j = compress_level_from; j <= compress_level_to;
+          j++) {
+        fprintf(stdout, "Compression level: %d", j);
+        compress_opt.level = j;
+        ShowCompressionSize(block_size, i.first, compress_opt);
+      }
+    } else {
+      fprintf(stdout, "Unsupported compression type: %s.\n", i.second);
+    }
+  }
+  return 0;
+}
+
+int SstFileDumper::ShowCompressionSize(
+    size_t block_size,
+    CompressionType compress_type,
+    const CompressionOptions& compress_opt) {
   ReadOptions read_options;
   Options opts;
   opts.statistics = ROCKSDB_NAMESPACE::CreateDBStatistics();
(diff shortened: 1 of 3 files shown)


tools/sst_dump_tool.cc
```diff
@@ -505,16 +540,24 @@ int SSTDumpTool::Run(int argc, char** argv, Options options) {
...
+  std::string compression_level_to_str;
   size_t block_size = 0;
   std::vector<std::pair<CompressionType, const char*>> compression_types;
   uint64_t total_num_files = 0;
   uint64_t total_num_data_blocks = 0;
   uint64_t total_data_block_size = 0;
   uint64_t total_index_block_size = 0;
   uint64_t total_filter_block_size = 0;
+  int32_t compression_level_from = 0;
+  int32_t compression_level_to = 0;
```
anand1976, line 560: These can be initialized to ```kDefaultCompressionLevel```.

tools/sst_dump_tool.cc
```diff
@@ -205,7 +206,38 @@ uint64_t SstFileDumper::CalculateCompressedTableSize(
...
     size_t block_size,
     const std::vector<std::pair<CompressionType, const char*>>&
-        compression_types) {
+      compression_types,
+    bool show_compression_levels,
+    int32_t compress_level_from,
+    int32_t compress_level_to) {
+
+  fprintf(stdout, "Block Size: %" ROCKSDB_PRIszt "\n", block_size);
+  for (auto& i : compression_types) {
```
anand1976, line 215: I think this logic can be simplified by defaulting ```compress_level_from``` and ```compress_level_to``` to ```kDefaultCompressionLevel``` and not requiring ```show_compression_levels``` to be passed.

tools/sst_dump_tool.cc
```diff
@@ -217,63 +249,58 @@ int SstFileDumper::ShowAllCompressionSizes(
...
+      false /* skip_filters */, column_family_name, unknown_level);
+  uint64_t num_data_blocks = 0;
+  auto start = std::chrono::steady_clock::now();
+  uint64_t file_size =
+      CalculateCompressedTableSize(tb_opts, block_size, &num_data_blocks);
+  auto end = std::chrono::steady_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - start);
+  fprintf(stdout, " Size: %10" PRIu64, file_size);
+  fprintf(stdout, " Blocks: %6" PRIu64, num_data_blocks);
+  fprintf(stdout, " Time Taken: %10s microsecs", std::to_string(duration.count()).c_str());
```
anand1976, line 266: ```make format``` to ensure <= 80 columns

tools/sst_dump_tool.cc
```diff
@@ -219,63 +245,59 @@ int SstFileDumper::ShowAllCompressionSizes(
...
-      fprintf(stdout, "Unsupported compression type: %s.\n", i.second);
-    }
+  std::string column_family_name;
+  int unknown_level = -1;
+  TableBuilderOptions tb_opts(
+      imoptions, moptions, ikc, &block_based_table_factories, compress_type,
+      0 /* sample_for_compression */, compress_opt,
+      false /* skip_filters */, column_family_name, unknown_level);
+  uint64_t num_data_blocks = 0;
+  auto start = std::chrono::steady_clock::now();
```
anand1976, line 255: We try to avoid using ```auto``` - https://google.github.io/styleguide/cppguide.html#Type_deduction. In this context, its better to be explicit so the unit of time is clear

tools/sst_dump_tool.cc
```diff
@@ -219,63 +245,59 @@ int SstFileDumper::ShowAllCompressionSizes(
...
+  uint64_t file_size =
+      CalculateCompressedTableSize(tb_opts, block_size, &num_data_blocks);
+  auto end = std::chrono::steady_clock::now();
```
anand1976, line 258: Same as previous comment

tools/sst_dump_tool.cc
```diff
@@ -219,63 +245,59 @@ int SstFileDumper::ShowAllCompressionSizes(
...
+       opts.statistics->getAndResetTickerCount(NUMBER_BLOCK_COMPRESSED);
+  const uint64_t not_compressed_blocks =
+       opts.statistics->getAndResetTickerCount(NUMBER_BLOCK_NOT_COMPRESSED);
+  // When the option enable_index_compression is true,
+  // NUMBER_BLOCK_COMPRESSED is incremented for index block(s).
+  if ((compressed_blocks + not_compressed_blocks) > num_data_blocks) {
+     num_data_blocks = compressed_blocks + not_compressed_blocks;
   }
+  const uint64_t ratio_not_compressed_blocks =
+       (num_data_blocks - compressed_blocks) - not_compressed_blocks;
```
anand1976, line 274: This formula looks wrong (it'll always evaluate to 0?). What exactly are we trying to calculate here?

akankshamahajan15: From the code it seems "num_data_blocks = compressed_blocks + not_compressed_blocks"  only when 
"compressed_blocks + not_compressed_blocks > num_data_blocks" so that
ratio_not_compressed_blocks will be 0 instead of -ve value.

Also, its mentioned that 
//When the option enable_index_compression is true,
// NUMBER_BLOCK_COMPRESSED is incremented for index block(s).
so I guess in that case, compressed_blocks increases which makes the total sum > num_data_blocks.

When I run the tool, for instance, num_data_blocks = 3320, compressed_blocks = 3321, non_compressed_blocks = 0 (before changing the num_data_blocks value)

anand1976: Interesting. Thanks for digging into this. Its still not clear when "compressed_blocks + not_compressed_blocks < num_data_blocks", which is when "ratio_not_compressed_blocks" will be > 0. Anyways, since it was there before, its a separate issue and we can probably follow it up outside this PR.

PR #6646, opened 2020-04-03
save a key comparison in DataBlockIter::Seek()

ajkr:
This saves an internal key comparison in all cases except when (1)
`block_restart_interval == 1`, or (2) the data block seek key exactly
equals the data key. I expect (2) is never the case since the seek key
has a special value type that is not used in data keys.

The key comparison that is saved is during the transition from binary
search over restart keys to linear scan within a restart interval. The
binary search lands on a key strictly less than the seek key. Given
that we know the start key of the linear scan is less than the seek key,
we do not need to compare them again.

Test Plan:

ran a benchmark with mostly default settings and counted key comparisons

before: `user_key_comparison_count = 19399529`
after: `user_key_comparison_count = 18431498`

setup command:

```
$ TEST_TMPDIR=/dev/shm/dbbench ./db_bench -benchmarks=fillrandom,compact -write_buffer_size=1048576 -target_file_size_base=1048576 -max_bytes_for_level_base=4194304 -max_background_jobs=12 -level_compaction_dynamic_level_bytes=true -num=10000000
```

benchmark command:

```
$ TEST_TMPDIR=/dev/shm/dbbench/ ./db_bench -use_existing_db=true -benchmarks=readrandom -disable_auto_compactions=true -num=10000000 -compression_type=none -reads=1000000 -perf_level=3
```


ajkr:
huh, after ~30 minutes stress test failed. not sure yet if related.

```
$ python ./tools/db_crashtest.py blackbox --simple --max_key=1000000 --value_size_mult=33 --write_buffer_size=1048576 --target_file_size_base=1048576 --max_bytes_for_level_base=4194304
...
***Iterator diverged from control iterator which has value 000000000000270F000000000000012B00000000000002B1 total_order_seek: 0 auto_prefix_mode: 0 S 00000000000022D700000000000001027878 PPNPNPNNNP; total_order_seek: 0 auto_prefix_mode: 0 STL PPPNNNN***
```

ajkr:
alright, known problems are fixed so ready for review.

table/block_based/block.cc
```diff
@@ -243,16 +243,55 @@ void DataBlockIter::Seek(const Slice& target) {
...
+      // to the data keys before comparing them to the seek key. If the seek key
+      // is for a snapshot seqnum then it's possible that the key at the
+      // `index`th restart is actually less than the seek key. So here we have
+      // to consider the case that the `index`th restart is smaller than seek
+      // key once global seqnum is applied. Since user keys in ingested files
+      // must be strictly ascending, we need to advance at most one position.
+      ParseNextDataKey<DecodeEntry>();
+    }
+  } else {
+    SeekToRestartPoint(index - 1);
```
pdillinger, line 269: > This saves one key comparison in block seeks when the seek key does not
exactly equal the found key. This is always the case in data
blocks as we use a special value type for seek that is not present in
data keys, and almost always the case in index blocks.

Nevertheless, what would this branch do if restart point @ index is equal to target? Would this code scan through restart point @ index-1 just to figure out that restart point @ index is equal to the target, which we already figured out in BinarySeek but do not communicate to the caller? (Aside from seqno craziness...) Wouldn't it simplify calling code to have BinarySeek return whether 'index' is an exact or nearest-greater match?

Actually, wouldn't it be even simpler to add that exact vs. non-exact distinction to the old version of the code to optimize away the unnecessary comparison? (So there's no 0 vs. > 0 distinction)

ajkr: > Nevertheless, what would this branch do if restart point @ index is equal to target? Would this code scan through restart point @ index-1 just to figure out that restart point @ index is equal to the target

Yes, it sacrifices that case under the assumption it almost never happens.

> which we already figured out in BinarySeek but do not communicate to the caller? (Aside from seqno craziness...) Wouldn't it simplify calling code to have BinarySeek return whether 'index' is an exact or nearest-greater match?

Agreed. Then we don't have to rely on my unverified assumption about how often it happens.

> Actually, wouldn't it be even simpler to add that exact vs. non-exact distinction to the old version of the code to optimize away the unnecessary comparison? (So there's no 0 vs. > 0 distinction)

Will try it out now and see. It sounds right...

ajkr: > > Actually, wouldn't it be even simpler to add that exact vs. non-exact distinction to the old version of the code to optimize away the unnecessary comparison? (So there's no 0 vs. > 0 distinction)

> Will try it out now and see. It sounds right...

We still end up needing to expand the binary search range by one to distinguish the case where 0th restart key is less than, equal to, or greater than the seek key. Interestingly right now `BinarySeek()` never compares against the 0th restart key (which also explains why they didn't handle the edge case where an empty index block says it has one restart key but actually has zero). Anyways, I think the change ends up similar to this PR.

But still before proceeding with this PR I plan to (1) fix the seqno craziness so `BinarySeek()` actually lands in a consistent place regardless of global seqno/snapshot reads; and (2) change `BinarySeek()` to return whether it found an exact or non-exact match, and use that to skip comparing against restart keys.

table/block_based/block.cc
```diff
@@ -128,17 +128,14 @@ struct DecodeKeyV4 {
 };
 
 void DataBlockIter::Next() {
-  assert(Valid());
```
ajkr, line 131: This is removed because `Next()` is now used to initialize block iterators ( we call it from `BlockIter::ScanAfterBinarySeek()` where we don't know whether the underlying call will be `ParseNextDataKey()` or `ParseNextIndexKey()`).

table/block_based/block.h
```diff
@@ -343,6 +343,8 @@ class BlockIter : public InternalIteratorBase<TValue> {
 
   Cache::Handle* cache_handle() { return cache_handle_; }
 
+  virtual void Next() = 0;
```
ajkr, line 346: This is also in the parent class, just add it here too to solve some compiler error.

table/block_based/block.cc
ajkr: Done. I also redid the `BinarySeek()` to avoid changing the binary search. The number of key comparisons saved lessened as a result, but I am fine with leaving those for a future PR..

table/block_based/block.cc
```diff
@@ -659,17 +648,70 @@ void IndexBlockIter::DecodeCurrentValue(uint32_t shared) {
...
+    } else {
+      max_offset = restarts_;
+    }
+    bool first = true;
+    while (true) {
+      Next();
+      if (!Valid()) {
+        break;
+      }
+      if (first) {
```
pdillinger, line 673: I don't like the "bool first" idiom here, as the amount of "duplicated" code without it is minimal. I think it hurts readability, and if the compiler doesn't properly optimize, could hurt efficiency with extra conditional branch.

And some of the code is already essentially already duplicated on either size of is_index_key_result.

table/block_based/block.cc
```diff
@@ -659,17 +648,70 @@ void IndexBlockIter::DecodeCurrentValue(uint32_t shared) {
...
+  if (is_index_key_result) {
+    SeekToRestartPoint(index);
+    Next();
+  } else {
+    SeekToRestartPoint(index);
+    uint32_t max_offset;
+    if (index + 1 < num_restarts_) {
+      max_offset = GetRestartPoint(index + 1);
+    } else {
+      max_offset = restarts_;
```
pdillinger, line 665: I don't think the `current_ == max_offset` case below is hit under this assignment, so in theory it could be `max_offset = restarts_ + 12345;`

I don't have a suggested code improvement, but maybe worth a code comment about intentions

pdillinger:
MacOS compilation failure

table/block_based/block.cc
ajkr: Done. Made it `kMaxUint32` so it's extra clear it's not intended to be the loop terminating condition.

table/block_based/block.cc
ajkr: Done.

table/block_based/block.cc
```diff
@@ -659,17 +648,66 @@ void IndexBlockIter::DecodeCurrentValue(uint32_t shared) {
...
-// is either the last restart point with a key less than target,
-// which means the key of next restart point is larger than target, or
-// the first restart point with a key = target
+template <class TValue>
+void BlockIter<TValue>::ScanAfterBinarySeek(const Slice& target, uint32_t index,
+                                            bool is_index_key_result,
+                                            const Comparator* comp) {
+  // Linear search (within restart block) for first key >= target
+  SeekToRestartPoint(index);
+  Next();
```
anand1976, line 657: A comment here might help. It took me a little while to figure out that this ```Next()``` actually doesn't advance to the next key, but just ends up decoding the key at the restart point we seek'd to.

table/block_based/block.cc
```diff
@@ -692,11 +730,32 @@ bool BlockIter<TValue>::BinarySeek(const Slice& target, uint32_t left,
...
     } else {
+      *is_index_key_result = true;
       left = right = mid;
     }
   }
 
+  assert(left == right);
   *index = left;
+  if (*index == 0) {
+    // Special case as we land at zero as long as restart key at index 1 is >=
```
anand1976, line 741: We land at 0 if restart key at index 1 > "target", right?

table/block_based/block.cc
```diff
@@ -393,6 +386,7 @@ void IndexBlockIter::Seek(const Slice& target) {
   }
   status_ = Status::OK();
   uint32_t index = 0;
+  bool is_index_key_result = false;
```
anand1976, line 389: Something like ```skip_linear_scan``` might be clearer.

table/block_based/block.cc
```diff
@@ -659,17 +648,66 @@ void IndexBlockIter::DecodeCurrentValue(uint32_t shared) {
...
+      // the next restart key is strictly greater than `target`, we can
+      // terminate upon reaching it without any additional key comparison.
+      max_offset = GetRestartPoint(index + 1);
+    } else {
+      // We are in the last restart interval. The while-loop will terminate by
+      // `Valid()` returning false upon advancing past the block's last key.
+      max_offset = port::kMaxUint32;
+    }
+    while (true) {
+      Next();
```
anand1976, line 671: Can we just use ```ParseNextDataKey``` instead of ```Next``` and ```Valid```?

PR #6652, opened 2020-04-06
Fix crash in JNI getApproximateSizes

koldat:
This change is fixing a crash happening in getApproximateSizes JNI implementation. It also reenables Java test that was crashing most likelly because if this bug.

### DIFF
--- a/java/rocksjni/rocksjni.cc
+++ b/java/rocksjni/rocksjni.cc
@@ -2277,10 +2277,11 @@ jlongArray Java_org_rocksdb_RocksDB_getApproximateSizes(
 
   auto ranges = std::unique_ptr<ROCKSDB_NAMESPACE::Range[]>(
       new ROCKSDB_NAMESPACE::Range[range_count]);
+  size_t range_offset = 0;
   for (jsize i = 0; i < jlen; ++i) {
     auto* start = reinterpret_cast<ROCKSDB_NAMESPACE::Slice*>(jranges[i]);
     auto* limit = reinterpret_cast<ROCKSDB_NAMESPACE::Slice*>(jranges[++i]);
-    ranges.get()[i] = ROCKSDB_NAMESPACE::Range(*start, *limit);
+    ranges.get()[range_offset++] = ROCKSDB_NAMESPACE::Range(*start, *limit);
   }
 
   auto* db = reinterpret_cast<ROCKSDB_NAMESPACE::DB*>(jdb_handle);
--- a/java/src/test/java/org/rocksdb/RocksDBTest.java
+++ b/java/src/test/java/org/rocksdb/RocksDBTest.java
@@ -1170,7 +1170,6 @@ public class RocksDBTest {
     }
   }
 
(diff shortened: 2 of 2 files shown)


koldat:
@adamretter , please take a look

java/rocksjni/rocksjni.cc
```diff
@@ -2277,10 +2277,11 @@ jlongArray Java_org_rocksdb_RocksDB_getApproximateSizes(
 
   auto ranges = std::unique_ptr<ROCKSDB_NAMESPACE::Range[]>(
       new ROCKSDB_NAMESPACE::Range[range_count]);
+  size_t range_offset = 0;
   for (jsize i = 0; i < jlen; ++i) {
     auto* start = reinterpret_cast<ROCKSDB_NAMESPACE::Slice*>(jranges[i]);
     auto* limit = reinterpret_cast<ROCKSDB_NAMESPACE::Slice*>(jranges[++i]);
-    ranges.get()[i] = ROCKSDB_NAMESPACE::Range(*start, *limit);
+    ranges.get()[range_offset++] = ROCKSDB_NAMESPACE::Range(*start, *limit);
```
adamretter, line 2284: So I think that on x64 (macOS):

1. `size_t` is `uint64_t`
2. `jsize` is `int`, which is 32 bits.

So I don't quite understand the change for this issue.... As previously `i` could only be 32 bits, why does increasing it to 64 bits fix this, when the max could only ever be 32 bits.

java/src/test/java/org/rocksdb/RocksDBTest.java
```diff
@@ -1185,7 +1184,7 @@ public void getApproximateSizes() throws RocksDBException {
 
         final long[] sizes = db.getApproximateSizes(
             Arrays.asList(
-                new Range(new Slice(key1), new Slice(key2)),
+                new Range(new Slice(key1), new Slice(key1)),
```
adamretter, line 1187: Why the change in the range, is it a genuine mistake originally?

java/rocksjni/rocksjni.cc
koldat: Issue is that i variable is increasing two times faster thus code writes after allocated memory.

java/src/test/java/org/rocksdb/RocksDBTest.java
koldat: Because test is asserting against 0 size. It cannot happen it is zero size when it gets range that is non empty.

java/rocksjni/rocksjni.cc
adamretter: Doh! Of course!

java/src/test/java/org/rocksdb/RocksDBTest.java
adamretter: Okie dokie :-)

adamretter:
@pdillinger LGTM! Can we merge?

PR #6657, opened 2020-04-06
Add Java API for rocksdb::CancelAllBackgroundWork()

sahibpandori:
Adding a Java API for rocksdb::CancelAllBackgroundWork() so that the user can call this (when required) before closing the DB. This is to **prevent the crashes when manual compaction is running and the user decides to close the DB**.

Calling CancelAllBackgroundWork() seems to be the recommended way to make sure that it's safe to close the DB (according to RocksDB FAQ: https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ#basic-readwrite).

### DIFF
--- a/java/rocksjni/rocksjni.cc
+++ b/java/rocksjni/rocksjni.cc
@@ -2612,6 +2612,17 @@ jobjectArray Java_org_rocksdb_RocksDB_compactFiles(
 
 /*
  * Class:     org_rocksdb_RocksDB
+ * Method:    cancelAllBackgroundWork
+ * Signature: (JZ)V
+ */
+void Java_org_rocksdb_RocksDB_cancelAllBackgroundWork(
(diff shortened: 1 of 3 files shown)


adamretter:
@pdillinger Could you merge this please?

sahibpandori:
Hey @adamretter @pdillinger , I just fixed the unit test. Initially I was using continueBackgroundWork() to check if the cancelAllBackgroundWork() call truly happened. However, the exception that I was expecting to be thrown would be thrown regardless if we call continueBackgroundWork() without calling pauseBackgroundWork().
So I've fixed the unit test to instead attempt to flush() after cancelling all background work and relying on the fact that we should get a ShutdownInProgress exception.

twei05:
Hi all,
  Here is a question. Which release version is checked in this change?

Best,
Tong

Issue #6589, opened 2020-03-24
DB Plugin Architecture Proposal


mrambacher:
This document outlines a mechanism for stacking RocksDB database implementations in novel and interesting ways.  A means of layering different databases and related functionality by compilation or configuration will be described.

RocksDB provides many different database implementations that provide different extended functionality.  The TTL database provides a means of expiring older values.  The TransactionDB provides a means of adding transaction support to RocksDB.  All of these databases are extensions of the basic RocksDB database and are examples of Stackable DBs.

Each database implementation (including the base implementation) follows a basic pattern.  The database is created via a static Open method in the base class (eg, DB::Open, DBTtl::Open) that takes a set of arguments -- including an “options” structure specific to that type of database -- and returns a database.

There are a few problems with this approach.  First, this approach means that the type of database must be decided at compile-time (since the specific static Open call must be determined invoked).  A compile-time decision limits the ability of an end-user to potentially choose a different implementation at a later date.  For example, a user could not choose to use the Cloud, Blob, or TTL databases in their application.  Additionally, there is no mechanism of stacking databases more than one deep.  For example, it is not possible to create a “Blob TTL” database.   Finally, because there is no clear interface for how to extend all of the database characteristics -- such as Clone, Destroy, Repair -- , not all databases implement these functions in the same way (if at all).

Our approach to solving this problem is to introduce a DBPlugin class.  This class will have methods that do the following:
-> PrepareDB: Changes any options (DBOptions, ColumnFamilyOptions) to be suitable for this database implementation
-> SanitizeDB: Checks that the database options are valid for this database implementation
-> WrapDB: Wraps the input database with one for this implementation
-> RepairDB: Does a “Repair” on this database as appropriate
-> DestroyDB: 
Other methods will be added as appropriate.  All methods will have a “no op” default implementation.

The DBOptions structure will be extended to have a vector of DBPlugins.  When a “constructive” operation is invoked, the plugins will be invoked in ascending order of the vector.  When a “destructive” operation (close, destroy) is invoked, the plugins will be invoked in descending order.  This logic would be handled by the base DB class methods.

Each type of database will have its own plugin implementation (e.g. TtlDBPlugin).  The plugins will be created with the options structure that is currently passed to the static Open call for this database implementation.  The plugin classes will be Configurable and Customizable, meaning that they can be created and stored in Options files.  

[...]

Issue #6533, opened 2020-03-13
Towards a more Configurable, Customizable, and Extensible RocksDB


mrambacher:
This document outlines a mechanism for providing configuration-time support for extending and configuring RocksDB via the Options files.  A means of adding functionality to RocksDB via customizable extensions is described and provided.

RocksDB provides a rich set of options and configurations for the programmer, allowing RocksDB to be used in many different environments.  Programmers can experiment with different implementations and settings for various RocksDB features and see what works best for their environment.  

Wouldn’t it be nice to be able to try these configurations out at run-time without rebuilding a new executable?  What about being able to use a HDFS via a configuration-time flag or try out different FilterPolicy or Cache implementations?  What about being able to try out different database implementations (like Backupable or TTL) without recompiling?

======

RocksDB uses an Options file to store its configuration settings.  The Options file outlines how RocksDB was configured at the time and can be used to rehydrate many of the settings within a RocksDB instance on restart.

Unfortunately, not all of the RocksDB properties can be configured via the Options file.  For some objects, the Options file may reflect the name of the class being used but not any configuration properties.  Furthermore, when these classes are loaded from the Options file, the loader does not know how to create or rehydrate many of those classes, meaning that these objects must be configured in code in order to recreate the original settings.

==========

Historically, RocksDB objects have been configured by creating a string to ObjectTypeInfo map where the string represents the name of the property to configure and the ObjectTypeInfo describes the type of the property being configured.  APIs were then added to get/set the properties and compare the options (GetDBOptionsFromMap, GetColumnFamilyOptionsFromMap, etc).  Each object class corresponded to a single map and multiple APIs.  These APIs would then need interjected in proper places to convert the objects to/from strings as appropriate.  Most of these APIs were very similar in nature but not quite identical. Additionally, each additional base type required updating an enumerated type and adding code (to three places) to support that type.  Finally, some types could not be handled by the base map code and were special cased in the APIs.

Presumably because of the number of steps involved and potential complexity, there are a large number of objects that cannot be configured via the Options file.  To address some of the deficiencies in the RocksDB options management, the OptionTypeInfo class was expanded and a Configurable base class was developed.  

======

The OptionTypeInfo class has been improved. The goal for this improvement is to reduce the amount of code required to add new configurable objects. These changes minimize or ideally
eliminate the special cases for objects that cannot be handled via the ObjectTypeInfo map. The improvement also eliminates the need for many of the option types.

[...]

PR #6661, opened 2020-04-07
Create a DBPlugin class to allow StackableDBs without static Open calls

mrambacher:
This PR addressses #6589, where new database implementations can be registered with the DBOptions and invoked without a custom DBXXX:Open call.  The plugin architecture allows the plugin to alter (SanitizeDB) or check (ValidateDB) the options to the open call prior to invoking the open.  If the options are not valid, the open fails with a Status.

The Plugin can wrap the Open call and return the appropriate wrapped (e.g. StackableDB) class as appropriate.  Different wrappers and validations are available for open versus OpenReadonly, versus OpenSecondary.

The Plugin architecture also allows the database to perform tasks when a repair or destroy operation is invoked.

The Plugin class is an extension of the Customizable framework (#6533, #6614), so plugin classes can be read/written/created from their serialization.

This checkin makes the TTL DB conform to the Plugin architecture.  The static Open calls still exist, but the functionality of the Open has been moved into the TtlPlugin layer.


yhchiang:
Oops, I mistakenly imported a wrong diff.  @mrambacher: do you mind updating the PR again with your previous change?

PR #6662, opened 2020-04-07
Fix initializer syntax for old Xcode compiler

pdillinger:
Summary: Example compiler output, from OSX TEST_GROUP=3:

db/flush_job_test.cc:185:7: error: suggest braces around initialization
of subobject [-Werror,-Wmissing-braces]
      kInvalidBlobFileNumber, 5, 103, 17, 102, 101};

Apparently permitted in newer version, but worth working around.
https://stackoverflow.com/questions/31555584/why-is-clang-warning-suggest-braces-around-initialization-of-subobject-wmis

Test Plan: CI (temporarily including OSX TEST_GROUP=3 in Travis)

### DIFF
--- a/db/flush_job_test.cc
+++ b/db/flush_job_test.cc
@@ -181,8 +181,8 @@ TEST_F(FlushJobTest, NonEmpty) {
   // Note: the first two blob references will not be considered when resolving
   // the oldest blob file referenced (the first one is inlined TTL, while the
   // second one is TTL and thus points to a TTL blob file).
-  constexpr std::array<uint64_t, 6> blob_file_numbers{
-      kInvalidBlobFileNumber, 5, 103, 17, 102, 101};
+  constexpr std::array<uint64_t, 6> blob_file_numbers{{
+      kInvalidBlobFileNumber, 5, 103, 17, 102, 101}};
   for (size_t i = 0; i < blob_file_numbers.size(); ++i) {
     std::string key(ToString(i + 10001));
     std::string blob_index;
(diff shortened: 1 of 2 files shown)


db/flush_job_test.cc
```diff
@@ -181,8 +181,8 @@ TEST_F(FlushJobTest, NonEmpty) {
   // Note: the first two blob references will not be considered when resolving
   // the oldest blob file referenced (the first one is inlined TTL, while the
   // second one is TTL and thus points to a TTL blob file).
-  constexpr std::array<uint64_t, 6> blob_file_numbers{
-      kInvalidBlobFileNumber, 5, 103, 17, 102, 101};
+  constexpr std::array<uint64_t, 6> blob_file_numbers{{
```
ltamasi, line 184: Could you try `constexpr std::array<uint64_t, 6> blob_file_numbers = {kInvalidBlobFileNumber, 5, 103, 17, 102, 101};` ? I think that would work too and it looks nicer :)

pdillinger: I thought = for initialization was arcane / taboo. ¯\_(ツ)_/¯

pdillinger:
(double brace passed compiler https://travis-ci.org/github/facebook/rocksdb/jobs/672163016)

db/flush_job_test.cc
ltamasi: Well, the nice (and completely legit) way of doing it is what's on master... Since that doesn't work with this compiler, I say let's pick the least ugly workaround :)

ltamasi: Hm, it seems that this version of `clang` also chokes on the `=` version that was standard even before CWG 1270. Double braces it is then... :(

pdillinger:
Link for not working with = and single braces: https://travis-ci.org/github/facebook/rocksdb/jobs/672187850

PR #6663, opened 2020-04-08
Fix CF import with overlapping SST files

### LINKED ISSUE #6647: [BUG] ImportColumnFamily::Prepare fail
https://github.com/facebook/rocksdb/blob/fdf882ded218344c136c97daf76dfb59e4bc155f/db/import_column_family_job.cc#L67

sstableKeyCompare does not compare LSN, so it is legal that two neighbouring ssts in the same level have have sstableKeyCompare(i.max, (i+1).min) == 0

ajkr:
Invariant checking should use internal key comparator rather than
`sstableKeyCompare()`. The latter was intended for checking whether a
compaction input file's neighboring files need to be included in the
same compaction. Using it for invariant checking was leading to false
positives for files with overlapping endpoints.

Fixes #6647.

Test Plan: regression test

### DIFF
--- a/db/import_column_family_job.cc
+++ b/db/import_column_family_job.cc
@@ -33,7 +33,6 @@ Status ImportColumnFamilyJob::Prepare(uint64_t next_file_number,
     files_to_import_.push_back(file_to_import);
   }
 
-  const auto ucmp = cfd_->internal_comparator().user_comparator();
   auto num_files = files_to_import_.size();
   if (num_files == 0) {
     return Status::InvalidArgument("The list of files is empty");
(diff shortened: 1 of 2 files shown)


db/import_column_family_job.cc
```diff
@@ -55,17 +54,18 @@ Status ImportColumnFamilyJob::Prepare(uint64_t next_file_number,
         }
       }
 
-      std::sort(sorted_files.begin(), sorted_files.end(),
-                [&ucmp](const IngestedFileInfo* info1,
-                        const IngestedFileInfo* info2) {
-                  return sstableKeyCompare(ucmp, info1->smallest_internal_key,
-                                           info2->smallest_internal_key) < 0;
-                });
+      std::sort(
```
zhichao-cao, line 57: If I understand correctly, the fix is to avoid the range tombstone check in sstableKeyCompare, which check the case if the two smallest_internal_key is equal it will sort range tombstone sentinel keys before other keys with the same user-key. This compare is not needed when just sorting SST files in this level based on the smallest_internal_key?

ajkr: Right, the special logic in `sstableKeyCompare()` is not relevant here. For checking that files within a level are non-overlapping, we want a simple internal key comparison.

thanks for reviewing!

Issue #6666, opened 2020-04-08
[Bug] Keys of ingested file are lost


Little-Wallace:
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://groups.google.com/forum/#!forum/rocksdb or https://www.facebook.com/groups/rocksdb.dev

We  queried the keys in which files we ingested just before and put them into another `SstFileWriter`, then we got error "Keys must be added in strict ascending order.".

We have found out the reason why keys are unordered. This is because RocksDB will encode data block and make keys with the same prefix can share prefix string. This feature makes RocksDB save much space when there are many similar keys. RocksDB encodes these keys with `InternalKey` rather than `UserKey`. If a key is very short and after it is coded as InternalKey, the common prefix with his adjacent key is longer than itself (we call it UserKey), BlockBasedTable will reset the seqno of this key if it exists in an ingested file and **make the common prefix changed incorrectly.**
 You can reproduce the bug by this test: [Little-Wallace/rocksdb](https://github.com/Little-Wallace/rocksdb/blob/out-of-order/db/external_sst_file_basic_test.cc#L1153)
This bug is caused by https://github.com/facebook/rocksdb/blob/master/table/block_based/block.cc#L555

Little-Wallace:
Here is test code which reproduces this problem.
```
  Options options = CurrentOptions();
  options.force_consistency_checks = true;

  DestroyAndReopen(options);
  const size_t kValueSize = 1 << 3;
  Random rnd(301);
  std::string value(RandomString(&rnd, kValueSize));

  ExternalSstFileInfo info;
  std::string f = test::PerThreadDBPath("sst_file_0");
  EnvOptions env;
  rocksdb::SstFileWriter writer(env, options);
  // Write some key to make global seqno larger than zero
  for (int i = 0; i < 10; i++) {
    ASSERT_OK(Put("ab" + Key(i), value));
  }
  // Get a Snapshot to make RocksDB assign global seqno to ingested sst files.
  auto snap = dbfull()->GetSnapshot();

  auto s = writer.Open(f);
  ASSERT_OK(s);
  std::string key1 = "ab";
  std::string key2 = "ab";

  // Make the prefix of key2 is same with key1 add zero seqno. The tail of every
  // key is composed as (seqno << 8 | value_type), and here `1` represents
  // ValueType::kTypeValue
  PutFixed64(&key2, 1);
  key2 += "cdefghijkl";

[...]

PR #6669, opened 2020-04-08
Fix wrong key being read on ingested file with global seqno and delta encoding

yiwu-arbug:
Summary:
On reading an ingested SST file, `DataBlockIter` will replace seqno encoded in a key with global seqno. However, if the original seqno was part of the prefix used for the next key, the global seqno is by mistake used as part of the prefix to construct the next key, causing wrong result being returned. Although at this point it is only software error while data in the file is not corrupted, the issue can further cause compaction output out of order and corrupted result when the ingested SST participated in compaction. Fixing the issue by save the actual seqno and restore it before the key being used as prefix to construct next key.

The unit test is by @Little-Wallace from #6666. Fixing #6666.

Test Plan:
New unit test

Signed-off-by: Yi Wu <yiwu@pingcap.com>

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,8 @@
 # Rocksdb Change Log
 ## Unreleased
+### Bug Fixes
+* Fix wrong result being read from ingested file, when a key in the file happen to be prefix of another key also in the file. The issue can further cause more data corruption. The issue exists with rocksdb >= 5.0.0 since DB::IngestExternalFile() was introduced.
+
 ### New Features
 * Added support for pipelined & parallel compression optimization for `BlockBasedTableBuilder`. This optimization makes block building, block compression and block appending a pipeline, and uses multiple threads to accelerate block compression. Users can set `CompressionOptions::parallel_threads` greater than 1 to enable compression parallelism.
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,7 +1,7 @@
 # Rocksdb Change Log
 ## Unreleased
 ### Bug Fixes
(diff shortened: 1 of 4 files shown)


HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # Rocksdb Change Log
 ## Unreleased
+### Bug Fixes
+* Fix wrong result being read from ingested file, when a key in the file happen to be prefix of another key also in the file. The issue can further cause more data corruption. The issue exists with rocksdb >= 5.0.0 since DB::IngestExternalFile() was introduced.
```
ajkr, line 4: I think this implies the problem happens every time a key is a prefix of another key. But in my understanding it requires the shared prefix extends multiple bytes into the footer of one of the keys (the first byte is the value type which isn't manipulated by the call to `UpdateInternalKey()`). Maybe clarify it only happens under certain conditions?

yiwu-arbug: updated.

table/block_based/block.cc
```diff
@@ -525,6 +525,9 @@ bool DataBlockIter::ParseNextDataKey(const char* limit) {
       key_.SetKey(Slice(p, non_shared), false /* copy */);
       key_pinned_ = true;
     } else {
+      if (global_seqno_ != kDisableGlobalSequenceNumber) {
+        key_.UpdateInternalKey(stored_seqno_, stored_value_type_);
```
Connor1996, line 529: can we give an example of the problem here, otherwise others who don't know the bug won't understand this very well

yiwu-arbug: updated comment.

PR #6672, opened 2020-04-09
Fix result slice's  address for direct io read

cheng-chang:
When aligned_buf is provided, the result slice's starting address should take offset advance into account.

Test Plan:
make random_access_file_reader_test && ./random_access_file_reader_test


cheng-chang:
added a test case which will not pass without the fix

anand1976:
@cheng-chang Thanks for adding the test.

PR #6675, opened 2020-04-09
Revert pr 6455

cheng-chang:
Reverts "Remove memcpy from RandomAccessFileReader::Read in direct IO mode (#6455)".
This reverts commit 0a0151f.
The optimization has not been applied to all targeted code paths, so it doesn't bring too much benefit for release 6.9, should be better to target for 6.10 with more code paths coverage and more thorough tests.

Test Plan:
make check


cheng-chang:
Closed since the optimization can now reduce memcpy for reading `Footer` and `BlobDB`.

PR #6680, opened 2020-04-09
Report kFilesMarkedForCompaction for delete triggered compactions

akankshamahajan15:
Summary : Set manual_compaction false in case of DeleteTriggeredCompaction object so that kFilesMarkedForComapaction can be reported.
          Added a DeletionTriggeredUniversalCompactionMarking test case for Deletion Triggered compaction in case of Universal Compaction.

Test Plan : make check -j64

### DIFF
--- a/db/compaction/compaction_picker_universal.cc
+++ b/db/compaction/compaction_picker_universal.cc
@@ -960,7 +960,7 @@ Compaction* UniversalCompactionBuilder::PickDeleteTriggeredCompaction() {
       GetCompressionType(ioptions_, vstorage_, mutable_cf_options_,
                          output_level, 1),
       GetCompressionOptions(mutable_cf_options_, vstorage_, output_level),
-      /* max_subcompactions */ 0, /* grandparents */ {}, /* is manual */ true,
+      /* max_subcompactions */ 0, /* grandparents */ {}, /* is manual */ false,
       score_, false /* deletion_compaction */,
       CompactionReason::kFilesMarkedForCompaction);
 }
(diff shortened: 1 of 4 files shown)


db/db_table_properties_test.cc
```diff
@@ -251,6 +251,96 @@ TEST_F(DBTablePropertiesTest, GetColumnFamilyNameProperty) {
...
+              CompactionReason::kFilesMarkedForCompaction);
+  }
+
+  void OnCompactionCompleted(DB* , const CompactionJobInfo& ci) override {
+    ASSERT_EQ(ci.compaction_reason,
+              CompactionReason::kFilesMarkedForCompaction);
+  }
+};
+
+TEST_F(DBTablePropertiesTest, DeletionTriggeredUniversalCompactionMarking) {
```
anand1976, line 267: Can this be combined with the ```DeletionTriggeredCompactionMarking``` test and parameterized, with compaction style as the parameter?

db/compaction/compaction_picker_universal.cc
```diff
@@ -960,7 +960,7 @@ Compaction* UniversalCompactionBuilder::PickDeleteTriggeredCompaction() {
       GetCompressionType(ioptions_, vstorage_, mutable_cf_options_,
                          output_level, 1),
       GetCompressionOptions(mutable_cf_options_, vstorage_, output_level),
-      /* max_subcompactions */ 0, /* grandparents */ {}, /* is manual */ true,
+      /* max_subcompactions */ 0, /* grandparents */ {}, /* is manual */ false,
```
anand1976, line 963: This needs to be checked and fixed for level compaction as well, so the same compaction reason is consistently reported for both level and universal.

PR #6686, opened 2020-04-10
Log CompactOnDeletionCollectorFactory parameters on DB open

anand1976:
Log it in the info log to help in troubleshooting. It is logged as follows -
```
2020/04/10-10:51:39.886662 7ffff7fef340                   Options.table_properties_collectors: CompactOnDeletionCollector (Sliding window size = 100 Deletion trigger = 90);
```

Tests:
make check

### DIFF
--- a/include/rocksdb/table_properties.h
+++ b/include/rocksdb/table_properties.h
@@ -137,9 +137,10 @@ class TablePropertiesCollectorFactory {
   // The name of the properties collector can be used for debugging purpose.
   virtual const char* Name() const = 0;
 
-  // Can be overridden by sub-classes to return configuration info that will
-  // be logged to the info log when the DB is opened
-  virtual std::string ToString() const { return ""; }
+  // Can be overridden by sub-classes to return the Name, followed by
+  // configuration info that will // be logged to the info log when the
+  // DB is opened
+  virtual std::string ToString() const { return Name(); }
 };
 
 // TableProperties contains a bunch of read-only properties of its associated
--- a/include/rocksdb/table_properties.h
+++ b/include/rocksdb/table_properties.h
@@ -136,6 +136,10 @@ class TablePropertiesCollectorFactory {
 
   // The name of the properties collector can be used for debugging purpose.
   virtual const char* Name() const = 0;
+
+  // Can be overridden by sub-classes to return configuration info that will
+  // be logged to the info log when the DB is opened
+  virtual std::string ToString() const { return ""; }
 };
 
 // TableProperties contains a bunch of read-only properties of its associated
--- a/include/rocksdb/utilities/table_properties_collectors.h
+++ b/include/rocksdb/utilities/table_properties_collectors.h
@@ -18,9 +18,9 @@ namespace ROCKSDB_NAMESPACE {
 class CompactOnDeletionCollectorFactory
     : public TablePropertiesCollectorFactory {
  public:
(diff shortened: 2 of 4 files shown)


include/rocksdb/utilities/table_properties_collectors.h
```diff
@@ -38,6 +38,8 @@ class CompactOnDeletionCollectorFactory
     return "CompactOnDeletionCollector";
   }
 
+  virtual std::string ToString() const override;
```
ltamasi, line 41: Minor: could you please change this and all overriding methods (including the destructor) so that `virtual` is not explicitly specified but `override` is?

include/rocksdb/table_properties.h
```diff
@@ -136,6 +136,10 @@ class TablePropertiesCollectorFactory {
 
   // The name of the properties collector can be used for debugging purpose.
   virtual const char* Name() const = 0;
+
+  // Can be overridden by sub-classes to return configuration info that will
+  // be logged to the info log when the DB is opened
+  virtual std::string ToString() const { return ""; }
```
ltamasi, line 142: Since this does not include the name, I'm wondering if something like `DebugInfo`, `Parameters` or `Configuration` would be a better name. Alternatively, we could keep `ToString` and make it include the name as well, in which case by default it could return what `Name` returns (as opposed to an empty string).

options/options.cc
```diff
@@ -321,7 +321,12 @@ void ColumnFamilyOptions::Dump(Logger* log) const {
                      compaction_options_fifo.allow_compaction);
     std::string collector_names;
     for (const auto& collector_factory : table_properties_collector_factories) {
+      std::string collector_config;
```
ltamasi, line 324: We could generate the entire string that gets logged using a single `std::ostringstream`. I think that would be cleaner and a bit more efficient. (Also, `collector_names` is a bit of a misnomer now since we log the entire collector configuration.)

utilities/table_properties_collectors/compact_on_deletion_collector.cc
```diff
@@ -78,6 +78,12 @@ CompactOnDeletionCollectorFactory::CreateTablePropertiesCollector(
       sliding_window_size_.load(), deletion_trigger_.load());
 }
 
+std::string CompactOnDeletionCollectorFactory::ToString() const {
+  return "Sliding window size = " +
```
ltamasi, line 82: We could consider using `std::ostringstream` here as well.

include/rocksdb/utilities/table_properties_collectors.h
anand1976: I think this class is extendable - specifically, MyRocks is planning to extend and override some methods so they can track such compactions.

include/rocksdb/table_properties.h
anand1976: I like the latter idea.

include/rocksdb/utilities/table_properties_collectors.h
ltamasi: I'm not saying these methods shouldn't be virtual; it's just a coding style comment (see https://google.github.io/styleguide/cppguide.html#Inheritance). `override` in the derived class implies `virtual` (the code does not compile if it doesn't override anything), so `virtual` in the derived class is superfluous. (In the base class, `virtual` is of course necessary.)

PR #6687, opened 2020-04-10
[db_stress] Add env_fault_injection argument to db_stress

yhchiang:
Summary:
Add env_fault_injection argument to db_stress.  When enabled,
FaultInjectionTestEnv will be used instead.  Currently this
option does not support running with other env setting.

This will allow
us to later manually produce error when running db_crashtest.

Test Plan:
make db_stress -j32
./db_stress --env_fault_injection
./db_stress --env_fault_injection --hdfs   // expect error message

Reviewers: ajkr

Subscribers: rocksdb

Tasks: T37386000

Tags: rocksdb

### DIFF
--- a/db_stress_tool/db_stress_gflags.cc
+++ b/db_stress_tool/db_stress_gflags.cc
@@ -674,4 +674,8 @@ DEFINE_int32(approximate_size_one_in, 64,
 
 DEFINE_int32(read_fault_one_in, 1000,
             "On non-zero, enables fault injection on read");
+
+DEFINE_bool(sync_fault_injection, false,
+            "If true, FaultInjectionTestFS will be used for write operations, "
+            " and unsynced data in DB will lost after crash.");
 #endif  // GFLAGS
--- a/db_stress_tool/db_stress_tool.cc
+++ b/db_stress_tool/db_stress_tool.cc
@@ -73,8 +73,9 @@ int db_stress_tool(int argc, char** argv) {
   } else {
     raw_env = Env::Default();
   }
+
 #ifndef NDEBUG
-  if (FLAGS_read_fault_one_in) {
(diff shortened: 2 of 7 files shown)


ajkr:
oh I just remembered there's also ongoing work to add fault injection on the read side of things - #6538. Do you mind checking this change is consistent with what's going on over there?

zhichao-cao:
@anand1976 This PR may related to the one you are working on

yhchiang:
@ajkr :  Sorry for the late reply.  Thanks for the info!  I will take a look.

db_stress_tool/db_stress_gflags.cc
```diff
@@ -674,4 +674,8 @@ DEFINE_int32(approximate_size_one_in, 64,
 
 DEFINE_int32(read_fault_one_in, 1000,
             "On non-zero, enables fault injection on read");
+
+DEFINE_bool(env_fault_injection, false,
```
anand1976, line 678: Can we leave env out of the name? Maybe name it ```write_fault_injection``` or ```sync_fault_injection```. The reason is we're moving to the FileSystem class for file related APIs, and this might become a bit of a misnomer.

db_stress_tool/db_stress_tool.cc
```diff
@@ -84,6 +85,16 @@ int db_stress_tool(int argc, char** argv) {
     raw_env = fault_env_guard.get();
   }
 #endif
+
+  if (FLAGS_env_fault_injection) {
```
anand1976, line 89: Can we combine this block with the one above and use ```FaultInjectionTestFS```, which offers the same functionality?

db_stress_tool/db_stress_gflags.cc
yhchiang: Sure.  Let me update to apply only FaultInjectionFS.

PR #6688, opened 2020-04-10
Add NewFileChecksumGenCrc32cFactory to file checksum

zhichao-cao:
Add NewFileChecksumGenCrc32cFactory to file checksum public interface such that applications can use the build in crc32 checksum factory.

### DIFF
--- a/include/rocksdb/file_checksum.h
+++ b/include/rocksdb/file_checksum.h
@@ -97,8 +97,10 @@ class FileChecksumList {
 // Create a new file checksum list.
 extern FileChecksumList* NewFileChecksumList();
 
-// Create a buildin Crc32 based file checksum generatory factory object, which
-// can be shared to create the Crc32c based checksum generator object.
-extern FileChecksumGenFactory* NewFileChecksumGenCrc32cFactory();
+// Return a shared_ptr of the builtin Crc32 based file checksum generatory
+// factory object, which can be shared to create the Crc32c based checksum
+// generator object.
+extern std::shared_ptr<FileChecksumGenFactory>
+GetFileChecksumGenCrc32cFactory();
 
 }  // namespace ROCKSDB_NAMESPACE
--- a/include/rocksdb/file_checksum.h
+++ b/include/rocksdb/file_checksum.h
@@ -97,4 +97,8 @@ class FileChecksumList {
 // Create a new file checksum list.
 extern FileChecksumList* NewFileChecksumList();
 
+// Create a buildin Crc32 based file checksum generatory factory object, which
+// can be shared to create the Crc32c based checksum generator object.
+extern FileChecksumGenFactory* NewFileChecksumGenCrc32cFactory();
+
 }  // namespace ROCKSDB_NAMESPACE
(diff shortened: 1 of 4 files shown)


include/rocksdb/file_checksum.h
```diff
@@ -97,4 +97,8 @@ class FileChecksumList {
 // Create a new file checksum list.
 extern FileChecksumList* NewFileChecksumList();
 
+// Create a buildin Crc32 based file checksum generatory factory object, which
```
riversand963, line 100: s/buildin/built-in, s/generatory/generator

include/rocksdb/file_checksum.h
```diff
@@ -97,4 +97,8 @@ class FileChecksumList {
...
+// can be shared to create the Crc32c based checksum generator object.
```
riversand963, line 101: Maybe make it more explicit that the factory object of Crc32c is meant to be shared?
In file_checksum_helper.cc
```
std::shared_ptr<FileChecksumGenFactory> NewFileChecksumGenCrc32cFactory() {
  static std::shared_ptr<FileChecksumGenFactory> default_crc32c_gen_factory(new FileChecksumGenCrc32cFactory());
  return default_crc32c_gen_factory;
}
```

riversand963: If you go with this, make sure test with ASAN build.
Two benefits:
1. Application does not have to worry about freeing the memory.
2. If application calls `GetFileChecksumGenCrc32cFactory()` multiple times, only one object of Crc32cGenFactory is allocated.

zhichao-cao: @riversand963 Thanks for the review! I may prefer to keep the current implementation, in which user or application can decide if they want to share or not and when to free the object.

include/rocksdb/file_checksum.h
zhichao-cao: @riversand963 Thanks for checking the typos, will change accordingly

include/rocksdb/file_checksum.h
riversand963: Okay. I will be pretty surprised if an application wants to create multiple `FileChecksumGenCrc32cFactory` objects, given that it has no data member.

zhichao-cao: @riversand963 I'm thinking it is more comfortable for user to create an object and manage its life cycle.

riversand963: @zhichao-cao flexibility also means more care.
It's not uncommon to return pointer to singleton, as what we already do for `BytewiseComparator` and `ReverseByteWiseComparator`.

zhichao-cao: @riversand963 Got it, will change accordingly. It's hard to balance flexibility and safety. But in this case, since there is no variable member, flexibility is not necessary.

zhichao-cao: @riversand963 Issue comes here. raw pointer or shared_ptr cannot be used for the static object here. Since options.file_checksum_factory is a shared PTR type. If we return a raw pointer of the static object, it will be deleted when option is destroyed. If we use static shared_ptr, still, when destroy the shared_ptr, the static object will be destroyed. Both cases will cause segment fault.

PR #6689, opened 2020-04-11
Reduce memory copies when fetching and uncompressing blocks from SST files

cheng-chang:
In https://github.com/facebook/rocksdb/pull/6455, we modified the interface of `RandomAccessFileReader::Read` to be able to get rid of memcpy in direct IO mode.
This PR applies the new interface to `BlockFetcher` when reading blocks from SST files in direct IO mode.

Without this PR, in direct IO mode, when fetching and uncompressing compressed blocks, `BlockFetcher` will first copy the raw compressed block into `BlockFetcher::compressed_buf_` or `BlockFetcher::stack_buf_` inside `RandomAccessFileReader::Read` depending on the block size. then during uncompressing, it will copy the uncompressed block into `BlockFetcher::heap_buf_`.

In this PR, we get rid of the first memcpy and directly uncompress the block from `direct_io_buf_` to `heap_buf_`.

Test Plan:
A new unit test `block_fetcher_test` is added.


table/block_fetcher.cc
```diff
@@ -209,14 +230,28 @@ Status BlockFetcher::ReadBlockContents() {
...
       // Actual file read
-      status_ = file_->Read(handle_.offset(), block_size_ + kBlockTrailerSize,
-                            &slice_, used_buf_, nullptr, for_compaction_);
+      if (file_->use_direct_io()) {
+        status_ =
+            file_->Read(handle_.offset(), block_size_ + kBlockTrailerSize,
+                        &slice_, nullptr, &direct_io_buf_, for_compaction_);
+        used_buf_ = const_cast<char*>(slice_.data());
+      } else {
+        PrepareBufferForBlockFromFile();
```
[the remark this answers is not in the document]
cheng-chang: I chose to duplicate PERF_xxxx in both branches

table/block_fetcher.cc
```diff
@@ -178,12 +192,19 @@ inline void BlockFetcher::GetBlockContents() {
...
+      CopyBufferToHeapBuf();
     } else if (used_buf_ == compressed_buf_.get()) {
       if (compression_type_ == kNoCompression &&
           memory_allocator_ != memory_allocator_compressed_) {
-        CopyBufferToHeap();
+        CopyBufferToHeapBuf();
+      } else {
+        heap_buf_ = std::move(compressed_buf_);
+      }
+    } else if (direct_io_buf_.get() != nullptr) {
```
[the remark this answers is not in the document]
cheng-chang: nope, the data might be in heap_buf_

cheng-chang:
@anand1976 PTAL

table/block_fetcher.cc
```diff
@@ -178,14 +189,21 @@ inline void BlockFetcher::GetBlockContents() {
     // page can be either uncompressed or compressed, the buffer either stack
     // or heap provided. Refer to https://github.com/facebook/rocksdb/pull/4096
     if (got_from_prefetch_buffer_ || used_buf_ == &stack_buf_[0]) {
-      CopyBufferToHeap();
+      CopyBufferToHeapBuf();
```
anand1976, line 192: It might be a good idea to add some comments either at the top of this function or in the header about the various buffers and how data moves between them. Its quite confusing otherwise.

table/block_fetcher.cc
```diff
@@ -178,14 +189,21 @@ inline void BlockFetcher::GetBlockContents() {
...
-        CopyBufferToHeap();
+        CopyBufferToHeapBuf();
       } else {
         heap_buf_ = std::move(compressed_buf_);
       }
+    } else if (direct_io_buf_.get() != nullptr) {
+      if (compression_type_ == kNoCompression) {
+        CopyBufferToHeapBuf();
+      } else {
+        CopyBufferToCompressedBuf();
```
anand1976, line 204: Previously, ```memory_allocator_``` was used if ```do_uncompress_``` was true, and ```memory_allocator_compressed_``` was used if ```do_uncompress_``` was false. Should the same check be done here? My guess is ```memory_allocator_compressed_``` is used if the block needs to be inserted in the compressed block cache?

cheng-chang: If the block needs to be uncompressed, it will be uncompressed in https://github.com/facebook/rocksdb/pull/6689/files#diff-e43ed8357174b6496b131fbf68e138b4R295.

Entering this code path means the block does not need to be uncompressed, so if compression type is NoCompression, then insert to heap buf, otherwise insert to compressed buf.

table/block_fetcher.cc
cheng-chang: left a comment for this method

table/block_fetcher.cc
anand1976: Good point. Thanks for the clarification.

PR #6690, opened 2020-04-11
Fix release build

riversand963:
Fix release build caused by variable defined but unused.

Test plan (devserver)
```
make release
```

### DIFF
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -26,22 +26,8 @@ class DBBasicTestWithTimestampBase : public DBTestBase {
 
  protected:
   static std::string Key1(uint64_t k) {
-    uint32_t x = 1;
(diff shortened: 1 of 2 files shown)


zhichao-cao:
Is 6.9.1 affected by this issue?

riversand963:
> Is 6.9.1 affected by this issue?

No, because 6.9.fb doest not have 0c05624d501aaab577bea821a6124e711b5ccdf0.
I should have avoided the term 'release'. Here, it means `make release` which is different from our internal release process.

zhichao-cao:
> > Is 6.9.1 affected by this issue?
> 
> No, because 6.9.fb doest not have [0c05624](https://github.com/facebook/rocksdb/commit/0c05624d501aaab577bea821a6124e711b5ccdf0).
> I should have avoided the term 'release'. Here, it means `make release` which is different from our internal release process.

I see, the PR just merged today. Thanks for the explanation.

PR #6693, opened 2020-04-13
Fix a few bugs in db_stress fault injection

anand1976:
Summary:
Fix the following issues -
1. Output parsing error in db_crashtest.py
2. Memory leak on exit
3. False alarm on filter block read error

Test Plan:
asan_crash


anand1976:
@riversand963 It looks like candidate files for deletion are not getting deduped. This seems to have been there for a while. The difference now is the ```FaultInjectionTestFS``` prints the error status to stderr - https://github.com/facebook/rocksdb/blob/master/test_util/fault_injection_test_fs.cc#L339

anand1976:
I'll suppress the error message and create a follow-up task for investigating why the candidate files are not getting deduped.

PR #6700, opened 2020-04-14
Fix a couple of bugs in db_stress fault injection

anand1976:
Summary:
1. Fix a memory leak in FaultInjectionTestFS in the stack trace related
code
2. Check status of all MultiGet keys before deciding whether an error
was swallowed, instead of assuming an ok status for any key means an
undetected error

Test Plan:
Run db_stress with asan and fault injection

### DIFF
--- a/db_stress_tool/no_batched_ops_stress.cc
+++ b/db_stress_tool/no_batched_ops_stress.cc
@@ -290,24 +290,31 @@ class NonBatchedOpsStressTest : public StressTest {
 #endif
     }
 
+#ifndef NDEBUG
+    if (fault_fs_guard && error_count && !SharedState::filter_read_error) {
+      int stat_nok = 0;
+      for (const auto& s : statuses) {
+        if (!s.ok() && !s.IsNotFound()) {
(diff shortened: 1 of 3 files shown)


db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -290,24 +290,31 @@ class NonBatchedOpsStressTest : public StressTest {
...
+          stat_nok++;
         }
+      }
+
+      if (stat_nok < error_count) {
+        // Grab mutex so multiple thread don't try to print the
+        // stack trace at the same time
+        MutexLock l(thread->shared->GetMutex());
+        fprintf(stderr, "Didn't get expected error from MultiGet\n");
+        fprintf(stderr, "Callstack that injected the error\n");
```
cheng-chang, line 307: These two `error` keywords will be captured by `db_crashtest.py` by two errors, is it ok?

anand1976: Let me change the text so there's only one error.

PR #6701, opened 2020-04-14
Set max_background_flushes dynamically

akankshamahajan15:
Summary: 1. Add changes so that max_background_flushes can be set dynamically.
                   2. Add a testcase DBOptionsTest.SetBackgroundFlushThreads which set the 
                        max_background_flushes dynamically using SetDBOptions.

TestPlan:  1. make -j64 check
                  2. Using new testcase DBOptionsTest.SetBackgroundFlushThreads

### DIFF
--- a/db/db_options_test.cc
+++ b/db/db_options_test.cc
@@ -402,6 +402,20 @@ TEST_F(DBOptionsTest, SetBackgroundCompactionThreads) {
   ASSERT_EQ(3, dbfull()->TEST_BGCompactionsAllowed());
 }
 
+TEST_F(DBOptionsTest, SetBackgroundFlushThreads) {
+  Options options;
+  options.create_if_missing = true;
+  options.max_background_flushes = 1;
+  options.env = env_;
+  Reopen(options);
+  ASSERT_EQ(1, dbfull()->TEST_BGFlushesAllowed());
+  ASSERT_EQ(1, env_->GetBackgroundThreads(Env::Priority::HIGH));
+  ASSERT_OK(dbfull()->SetDBOptions({{"max_background_flushes", "3"}}));
+  ASSERT_EQ(3, env_->GetBackgroundThreads(Env::Priority::HIGH));
+  ASSERT_EQ(3, dbfull()->TEST_BGFlushesAllowed());
+}
+
+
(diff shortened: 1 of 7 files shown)


db/db_options_test.cc
```diff
@@ -402,6 +402,19 @@ TEST_F(DBOptionsTest, SetBackgroundCompactionThreads) {
...
 
+TEST_F(DBOptionsTest, SetBackgroundFlushThreads) {
+  Options options;
+  options.create_if_missing = true;
+  options.max_background_flushes = 1;
+  options.env = env_;
+  Reopen(options);
+  ASSERT_EQ(1, dbfull()->TEST_BGFlushesAllowed());
+  ASSERT_OK(dbfull()->SetDBOptions({{"max_background_flushes", "3"}}));
+  auto stop_token = dbfull()->TEST_write_controler().GetStopToken();
```
ajkr, line 413: Did this `GetStopToken()` come from the compaction equivalent of this test? I think it is relevant there but not here. It's needed there because RocksDB has a feature to limit compaction concurrency to one until compaction falls behind, after which it increases the limit to `max_background_compactions`. That feature intends to help in multi-process scenarios (like MyRocks) to allocate more compaction bandwidth to the RocksDB instances that really need it.

The way it detects compaction fell behind is via these `WriteController` tokens. So obtaining the token here would cause the number of compactions allowed to increase from one to `max_background_compactions`. But AFAIK there is no analogous feature for flush, so we shouldn't need to acquire a token here.

db/db_options_test.cc
```diff
@@ -402,6 +402,19 @@ TEST_F(DBOptionsTest, SetBackgroundCompactionThreads) {
...
+  ASSERT_EQ(3, dbfull()->TEST_BGFlushesAllowed());
+}
```
ajkr, line 415: can we check `env_->GetBackgroundThreads(Env::Priority::HIGH)` increase from one to three as well?

akankshamahajan15: Yes sure. I will add that check. Thanks!

db/db_options_test.cc
akankshamahajan15: Yes, we don't need this. It is required in compaction test case. I will remove this one. Thanks!

PR #6705, opened 2020-04-15
Remove a printf from db_stress that's not useful info

anand1976:
This was causing db_crashtest.py to wrongly assume an error by parsing the output. Hopefully this will stabilize the crash tests.

Test:
make blackbox_crash_test

### DIFF
--- a/tools/db_crashtest.py
+++ b/tools/db_crashtest.py
@@ -453,8 +453,7 @@ def whitebox_crash_main(args, unknown_args):
 
         stdoutdata = stdoutdata.lower()
         errorcount = (stdoutdata.count('error') -
-                      stdoutdata.count('got errors 0 times') -
-                      stdoutdata.count('got expected errors 0 times'))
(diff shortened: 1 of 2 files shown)


tools/db_crashtest.py
```diff
@@ -453,8 +453,7 @@ def whitebox_crash_main(args, unknown_args):
 
         stdoutdata = stdoutdata.lower()
         errorcount = (stdoutdata.count('error') -
-                      stdoutdata.count('got errors 0 times') -
-                      stdoutdata.count('got expected errors 0 times'))
```
ltamasi, line 457: I think the problem might actually be this line: if the errors are expected, we shouldn't expect the number to be zero, no? That is, we might consider keeping the message in `db_stress_stat.h` and simply remove the `0 times` part from here so that any number of expected errors are ignored.

anand1976: I thought of doing that initially, but I don't think that print tells us anything useful anyway.

PR #6710, opened 2020-04-15
Implement deadline support for MultiGet

anand1976:
Initial implementation of ReadOptions.deadline for MultiGet. If the request takes longer than the deadline, the keys not yet found will be returned with Status::TimedOut(). This
implementation enforces the deadline in DBImpl, which is fairly high
level. Its best effort and may not check the deadline after every key
lookup, but may do so after a batch of keys.

In subsequent stages, we will extend this to passing a timeout down to the FileSystem.

Test Plan:
Add new unit tests

### DIFF
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -2421,7 +2421,7 @@ class DBBasicTestMultiGetDeadline : public DBBasicTestMultiGet {
           file_(std::move(file)) {}
 
     IOStatus Read(uint64_t offset, size_t len, const IOOptions& opts,
-                  Slice* result, char* scratch, IODebugContext* dbg) const {
+          Slice* result, char* scratch, IODebugContext* dbg) const override {
       int delay;
       if (fs_.ShouldDelay(&delay)) {
         Env::Default()->SleepForMicroseconds(delay);
@@ -2431,7 +2431,7 @@ class DBBasicTestMultiGetDeadline : public DBBasicTestMultiGet {
     }
 
     IOStatus MultiRead(FSReadRequest* reqs, size_t num_reqs,
-                       const IOOptions& options, IODebugContext* dbg) {
+          const IOOptions& options, IODebugContext* dbg) override {
       int delay;
       if (fs_.ShouldDelay(&delay)) {
         Env::Default()->SleepForMicroseconds(delay);
@@ -2452,7 +2452,7 @@ class DBBasicTestMultiGetDeadline : public DBBasicTestMultiGet {
     IOStatus NewRandomAccessFile(const std::string& fname,
                                  const FileOptions& opts,
                                  std::unique_ptr<FSRandomAccessFile>* result,
-                                 IODebugContext* dbg) {
+                                 IODebugContext* dbg) override {
       std::unique_ptr<FSRandomAccessFile> file;
       IOStatus s;
 
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -2010,7 +2010,7 @@ class DBBasicTestMultiGet : public DBTestBase {
     return false;
   }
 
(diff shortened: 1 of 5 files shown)


db/db_impl/db_impl.cc
```diff
@@ -1801,6 +1810,19 @@ std::vector<Status> DBImpl::MultiGet(
...
+
+    if (read_options.deadline.count() &&
+        env_->NowMicros() >
+            static_cast<uint64_t>(read_options.deadline.count())) {
+      timedout = true;
+      break;
+    }
+  }
+
+  if (timedout) {
```
riversand963, line 1822: Is `timedout` the equivalent to `keys_read < num_keys`?

db/db_basic_test.cc
```diff
@@ -2363,6 +2396,240 @@ INSTANTIATE_TEST_CASE_P(ParallelIO, DBBasicTestWithParallelIO,
                                            ::testing::Bool(), ::testing::Bool(),
                                            ::testing::Values(1, 4)));
 
+// A test class for intercepting random reads and injecting artificial
+// delays. Used for testing the deadline/timeout feature
+class DBBasicTestMultiGetDeadline : public DBBasicTestMultiGet {
+ public:
+  DBBasicTestMultiGetDeadline()
+      : DBBasicTestMultiGet("db_basic_test_multiget_deadline", 10, false, true,
```
riversand963, line 2404: Would be helpful to comment on these constant values passed to `DBBasicTestMultiGet`.

db/db_impl/db_impl.h
```diff
@@ -1763,7 +1763,7 @@ class DBImpl : public DB {
   // to have acquired the SuperVersion and pass in a snapshot sequence number
   // in order to construct the LookupKeys. The start_key and num_keys specify
   // the range of keys in the sorted_keys vector for a single column family.
-  void MultiGetImpl(
+  Status MultiGetImpl(
```
riversand963, line 1766: Nit: I think the return value of `MultiGetImpl` worth a sentence. Now that it returns a `Status` now, it may be clearer to explicitly say the return value is either OK or timedout, while the result of individual key lookups are stored in `sorted_keys`.

db/db_basic_test.cc
```diff
@@ -2363,6 +2396,240 @@ INSTANTIATE_TEST_CASE_P(ParallelIO, DBBasicTestWithParallelIO,
...
+                             std::unique_ptr<FSRandomAccessFile>& file)
+        : FSRandomAccessFileWrapper(file.get()),
+          fs_(fs),
+          file_(std::move(file)) {}
+
+    IOStatus Read(uint64_t offset, size_t len, const IOOptions& opts,
+                  Slice* result, char* scratch, IODebugContext* dbg) const {
+      int delay;
+      if (fs_.ShouldDelay(&delay)) {
+        Env::Default()->SleepForMicroseconds(delay);
```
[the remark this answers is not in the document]
anand1976: I didn't understand the concern. What is the problem with a longer delay?

db/db_impl/db_impl.cc
anand1976: Yes. I suppose that check should suffice.

db/db_basic_test.cc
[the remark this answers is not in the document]
riversand963: There are limits on how long a unit test can run. If a unit test passes a large value to the deadline FS, and this will cause the test to sleep for a long time. I don't think it is necessary to sleep for long time, can we add a check to avoid this entirely?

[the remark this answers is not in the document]
anand1976: Got it. I'll assert if its greater than, say, 500ms. Just want to give enough buffer to ensure there's no flakiness.

PR #6711, opened 2020-04-15
Add IsDirectory() to Env and FS

riversand963:
Summary:
IsDirectory() is a common API to check whether a path is a regular file or
directory.
POSIX: call stat() and use S_ISDIR(st_mode)
Windows: PathIsDirectoryA() and PathIsDirectoryW()
HDFS: FileSystem.IsDirectory()
Java: File.IsDirectory()
...

Test Plan:
make check

### DIFF
--- a/env/fs_posix.cc
+++ b/env/fs_posix.cc
@@ -879,8 +879,18 @@ class PosixFileSystem : public FileSystem {
 
   IOStatus IsDirectory(const std::string& path, const IOOptions& /*opts*/,
                        bool* is_dir, IODebugContext* /*dbg*/) override {
+    // First open
+    int fd = -1;
+    int flags = cloexec_flags(O_RDONLY, nullptr);
+    {
+      IOSTATS_TIMER_GUARD(open_nanos);
+      fd = open(path.c_str(), flags);
+    }
+    if (fd < 0) {
+      return IOError("While open for IsDirectory()", path, errno);
+    }
     struct stat sbuf;
-    if (stat(path.c_str(), &sbuf) < 0) {
+    if (fstat(fd, &sbuf) < 0) {
       return IOError("While doing stat for IsDirectory()", path, errno);
     }
     if (nullptr != is_dir) {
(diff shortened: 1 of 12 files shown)


include/rocksdb/env.h
```diff
@@ -540,6 +540,9 @@ class Env {
     return Status::NotSupported();
   }
 
+  // Check whether the specified path is a directory
+  virtual Status IsDirectory(const std::string& path, bool* is_dir) = 0;
```
anand1976, line 544: I wonder if we should have a default implementation that returns ```Status::NotSupported()```, rather than pure abstract, in order to avoid disruption. Especially since RocksDB itself doesn't need to use it at the moment.

env/fs_posix.cc
```diff
@@ -875,6 +875,18 @@ class PosixFileSystem : public FileSystem {
     return IOStatus::OK();
   }
 
+  IOStatus IsDirectory(const std::string& path, const IOOptions& /*opts*/,
+                       bool* is_dir, IODebugContext* /*dbg*/) override {
+    struct stat sbuf;
+    if (stat(path.c_str(), &sbuf) < 0) {
+      return IOError("While doing stat", path, errno);
```
anand1976, line 882: Message could be a little more descriptive to indicate the failure is in ```IsDirectory()```, like "While doing stat on directory".

include/rocksdb/env.h
riversand963: In the future, I am thinking whether it's possible to implement these functions as
```
GetFileSystem()->XXX();
```
For now, I will just return `NotSupported()`.

anand1976: Yes, they could be redirected to the ```FileSystem``` implementation.

include/rocksdb/env.h
```diff
@@ -540,6 +540,11 @@ class Env {
     return Status::NotSupported();
   }
 
+  // Check whether the specified path is a directory
+  virtual Status IsDirectory(const std::string& /*path*/, bool* /*is_dir*/) {
+    return Status::NotSupported();
```
anand1976, line 545: Nit: Can we add a message here too? There was a case recently where a NotSupported() error was reported in the info log, but it was hard to figure out where it was coming from.

riversand963: Done.

PR #6144, opened 2019-12-10
Add ppc64le builds to Travis

adamretter:
Let's see how this goes...

PR #6713, opened 2020-04-16
POWER8 compatiblity fix

grooverdan:
When running on a POWER8, the compiler will successfully report POWER9 compatibility. It will however fail to run due to POWER9+ only instructions being used.

As such we perform POWER8 checks with a compile test based on the _ARCH_PWR8 per the ABI specification.

ABI ref: http://openpowerfoundation.org/wp-content/uploads/resources/leabi/content/dbdoclet.50655243_75216.html

To preserve the goals of ensuring an optimization for the local hardware, -mcpu=native is used, but only if we don't already have C{,XX}_FLAGS with cpu/arch/tune set.

This is compatible on POWER8, clang++-6.0.0 and gcc-7.5.0.

This ensures that if a binary distributor compiles on a POWER9, and aims to support POWER8 packages, then hopefully they will have the correct C{,XX}_FLAGS, and as such, will
not result in a broken package on POWER8.

Fixes: 8fc20ac468b266a53083175025375985ec04b796

@adamretter FYI - this changes your commit so I hope your ok with this.

Also I don't quite follow the reason for the CMAKE_SYSTEM_PROCESSOR pattern change https://github.com/facebook/rocksdb/pull/6144/commits/fc96bb9222969543d3f6f4d0c6b7f7050fb63b70. A big endian would be `ppc64` and I'm not sure it will ever be `power64` however I may be wrong. I did fix one case of a cross-compile using `uname -m` incorrectly, is this the case?
http://buildroot-busybox.2317881.n4.nabble.com/PATCH-1-1-package-pkg-cmake-set-CMAKE-SYSTEM-PROCESSOR-correctly-on-powerpc64-le-tc159337.html

I did come across this BE fault https://github.com/MariaDB/server/pull/369 years ago but never resolved if it was a MariaDB or RockDB bug. Probably MariaDB due to other BE fixes I see here.


adamretter:
@grooverdan I will try and take a look soon

grooverdan:
gentle reminder :smiley:

grooverdan:
happy new month. :paintbrush:

adamretter:
@grooverdan Apologies for the delay, and thanks for persisting :-)

> Also I don't quite follow the reason for the CMAKE_SYSTEM_PROCESSOR pattern change fc96bb9. A big endian would be ppc64 and I'm not sure it will ever be power64 however I may be wrong.

The syntax `MATCHES "^(powerpc|ppc)64"` is designed to match the `CMAKE_SYSTEM_PROCESSOR` reagardless of edian'nes. i.e. it matches against the start of the string, so it will also allow a trailing `le`.

...more comments to come shortly...

CMakeLists.txt
```diff
@@ -203,18 +203,20 @@ else()
...
-      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mcpu=power8 -mtune=power8")
-      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcpu=power8 -mtune=power8")
-    endif(HAS_POWER8)
-  endif(HAS_POWER9)
+  CHECK_CXX_SOURCE_COMPILES("
+  #ifndef _ARCH_PWR8
+  #error Not POWER8+
+  #endif
+  int main() { return 0; }"
+  HAS_POWER8)
```
adamretter, line 213: If I understand this correctly `CHECK_CXX_SOURCE_COMPILES` sets the `HAS_POWER8` flag only if the machine support the POWER 8 (or higher) ISA... however it then appears that this flag is not used anywhere else?

CMakeLists.txt
```diff
@@ -203,18 +203,20 @@ else()
...
+  if(NOT CMAKE_C_FLAGS MATCHES "m(cpu|tune|arch)")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mcpu=native")
```
adamretter, line 215: From the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/RS_002f6000-and-PowerPC-Options.html):

> Specifying ‘native’ as cpu type detects and selects the architecture option that corresponds to the host processor of the system performing the compilation. -mcpu=native has no effect if GCC does not recognize the processor.

> The other options specify a specific processor. Code generated under those options runs best on that processor, and may not run at all on others.

So if I understand correctly:
Previously... if building on a POWER9 system then the binaries were optimised for POWER9 ISA. If building on a POWER8 the binaries should have been optimised for POWER8, but the HAS_POWER9 flag was being set to true on a POWER8 and the resultant compiled binaries would not run on said POWER8.

---

I think we need the option to be able to build CPU optimised and non-optimised binaries. For portability practices we already have the `PORTABLE` flag which is used in Makefile and also in CMakeLists.txt.

1. For optimised builds, I think we should combine what we had previously, with your extra CHECK_CXX_SOURCE_COMPILES check on POWER8 to ensure that POWER8 compiled binaries will run on Power8. 

2. For portable binaries, I think that on ppc64/le we should check the presence of the PORTABLE flag and should likely always set a minimum system ISA. Whether that should be POWER8 or something earlier like POWER6 I can't say at the moment... I am not certain what is representative of the installed base of PPC systems out there.

When performing release builds we already set the PORTABLE flag, so this should provide what you need.

In addition, whatever changes we make to CMakeLists.txt should also be similarly reflected in our Makefile setup too.

grooverdan: yes you've understood it correctly. Good idea making more use of `PORTABLE`.

1. native(?) optimised

What was there previously was forcing the compile cpu to POWER9, which isn't going to run on POWER8. The CHECK_CXX_SOURCE_COMPILES ensures a minimum of POWER8. This is why I set mcpu to native. So on P8 we'd get P8 code and on P9 we'd get P9. Do you have a good runtime check for cmake for P8? `grep /proc/cpuinfo`  but that is bit ugly. I'm not sure what this gains over mcpu=native. Maybe I've misunderstood something. Apologies if I have.

2. Portable:

I can check on a POWER7. POWER6 might be harder to find. So sure, can make PORTABLE default to the minimum workable architecture.

2.1 portable, but still optimized.

For functions like crc32 that have a significant performance difference we could use https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes - target_clones maybe? Any other functions you've noticed significant p8 vs p9 gains in?

grooverdan:
updated in #7079 with PORTABLE included

CMakeLists.txt
adamretter: > Do you have a good runtime check for cmake for P8? grep /proc/cpuinfo but that is bit ugly. I'm not sure what this gains over mcpu=native. Maybe I've misunderstood something. Apologies if I have.

I think I am also getting a bit confused. My concern I think was that we were previously setting both `-mcpu` and -`mtune` however we are now only setting `-mcpu`... https://gcc.gnu.org/onlinedocs/gcc/RS_002f6000-and-PowerPC-Options.html

> So sure, can make PORTABLE default to the minimum workable architecture.

Excellent!

> Any other functions you've noticed significant p8 vs p9 gains in?

I don't have enough experience with using Rocks on ppc64le systems to really comment on this. Most of my work was just making sure that it would compile on ppc64le.

PR #6717, opened 2020-04-16
Pluggable Compression for RocksDB

lucagiac81:
This PR introduces pluggable compression in RocksDB. With the new architecture, applications can use custom compressors in addition to the ones available by default, and they can do so with no code changes. Backward compatibility is maintained.

### Architecture Overview

Current architecture
- RocksDB supports a fixed set of compression algorithms (“built-in” algorithms)
- Each algorithm is assigned a fixed numeric type (in CompressionType enum)
- The numeric type is used in column family options (compression, bottommost_compression, compression_per_level)

Pluggable architecture
- Applications can add custom compression algorithms to RocksDB (the built-in ones are still available)
- Each algorithm is still assigned a numeric type
  - Built-in algorithms have the same fixed numeric type as before (backward compatible)
  - Custom algorithms are assigned numeric types dynamically
- Since each algorithm is still identified by a numeric type, existing column family options can be used with no change (backward compatible)

The new architecture is based on two classes: Compressor and CompressorRegistry. 

### Compressor

The Compressor class defines the interface each compression algorithm needs to implement. 

A compressor has 
- a unique name
- Compress/Uncompress methods
- Optional methods to handle dictionary/context (implemented, for example, by zstd)

Built-in algorithms are wrapped in this interface as well.
To ensure compressor names are unique, we need a naming convention. For example, we could use a reverse domain name notation like Java packages (e.g., org.mycompany.mycompressor).

### CompressorRegistry

The CompressorRegistry stores instances of all available compressors (built-in and custom), and it assigns a numeric type to each compressor.

To add a compressor to the registry, the application first needs a Compressor instance. It can obtain an instance in two ways
- Define and instantiate a compressor in the application code.
- Load a compressor from a shared library.

To load compressors from shared libraries, CompressorRegistry provides two methods
- LoadCompressor: it takes a library name and a directory path, and it returns a pointer to a Compressor instance.
- LoadCompressors: it takes a directory path and loads compressors from all the libraries in that directory.

Compressors are added to the registry as follows
- Some compressors are automatically added when the registry is instantiated (by RocksDB, unless the application explicitly instantiates it first)
  - Built-in compressors.
  - Compressors loaded from libraries in a default directory (“./compressors”). 
- Additional compressors can be explicitly added by the application using these methods
  - AddCompressor: it takes a Compressor instance, and it returns the assigned numeric type.
  - LoadAndAddCompressor: it takes a library name and path. This is a convenience method to load a compressor from a library and add it to the registry in one call. There is also a LoadAndAddCompressors method to load and add all compressors in a directory.

Loading compressors automatically from a default directory allows applications that don’t explicitly handle the CompressorRegistry (e.g., existing applications not yet modified for pluggable compression) to access custom compressors with no/minimal code changes. This approach has been tested with MySQL/MariaDB with MyRocks. 
The default directory can be overridden if the application instantiates the registry before RocksDB does internally. If an empty path is specified, no automatic loading takes place.

When using the AddCompressor method, we have two options
- Let the registry assign a type dynamically (most common scenario).
- Force the registry to assign a specific type. For example, this allows overriding a built-in compressor with a compatible implementation.

Once the compressors are in the registry, instances are retrieved using their name or type (using the GetCompressor methods).

The diagram below summarizes the different ways to interact with the registry.

image

### Code Examples

Example 1: define and instantiate a compressor in application code and add it to the registry
```
class MyCompressor : public Compressor {
  // implement at least Name, Compress, Uncompress
  ...
}

auto registry = CompressorRegistry::NewInstance();
std::shared_ptr<Compressor> compressor = std::make_shared<MyCompressor>();
unsigned char type = registry->AddCompressor(compressor);
```

Example 2: load a compressor from a library and add it to the registry
```
auto registry = CompressorRegistry::NewInstance();
std::shared_ptr<Compressor> compressor = registry->LoadCompressor("library_name", "library_path");
// Set any properties of compressor as needed here...
unsigned char type = registry->AddCompressor(compressor);
```

Example 3: load a compressor from a library and add it to the registry in one call
```
auto registry = CompressorRegistry::NewInstance();
unsigned char type = registry->LoadAndAddCompressor("library_name", "library_path");
```

For all the examples above, use the type obtained when adding the compressor to set options
```
Options options;
options.compression = static_cast<CompressionType>(type);
```

### Persistence

The CompressorRegistry assigns numeric types to compressors dynamically. Therefore, the assigned types may change from run to run, depending on which compressors are available and in which order they are added (except for the built-in compressors, which always have a fixed type for backward compatibility). So, how do we identify compressors across different instances of RocksDB? We need to rely on the compressor name, which is unique.

The table below summarizes the compression information stored in SST files and Options file

image

kCustomCompression is a fixed numeric type defined in the CompressionType enum. It is used as the compression type identifier in blocks compressed with custom compressors.

When writing an SST file
- If a block is compressed with a custom compressor, set its compression type identifier to kCustomCompression.
- Record the compressor name in the properties block.

When reading an SST file
- Read the compressor name from the properties block. Query the CompressorRegistry to get the corresponding numeric type (if the required compressor is not present in the registry, an error occurs). This is called the “table compression type”.
- When reading a compressed block, if its compression type identifier is kCustomCompression, overwrite it with the “table compression type”.

So, the kCustomCompression type is like a “flag” in compressed blocks, which tells us to refer to the properties block to find the right compressor.

### CompressionOptions

The CompressionOptions class has a fixed set of options. Some are general (like “level”), and some are specific to built-in compressors.
To pass options to custom compressors, a custom_options member was added
- custom_options is a map of key-value string pairs
- the options will be interpreted by the selected compressor as needed

CompressionOptions can be passed through string/file with this format
`compression_opts=4:5:6:0:0:0:1:opt1=val1:opt2=val2;`

CompressionOptions are serialized to the properties block of SST files with this format
`window_bits=-14; level=32767; strategy=0; max_dict_bytes=0; zstd_max_train_bytes=0; custom_options={opt1=val1;opt2=val2;}; enabled=0;`

The same format is used in the LOG file.

### Performance

Path length/CPI is within 1% of unmodified code for fillseq and readrandom test. Throughput and latency are matched. 

### Summary of Code Changes

- Add Compressor class
- Add CompressorRegistry class
- Wrap existing compression functions into the Compressor interface
  - In particular, any references to zstd outside the compressor have been cleaned up 
- Move Dict/Context/Info to public API (as they are needed in the Compress/Uncompress methods)
- Replace switch statements to select compressors with queries to the CompressorRegistry
- Update db_bench to support custom compressors (using the compressor name for compression_type)
- Update sst_dump tool to support custom compressors
- Add functions to C API to interact with the CompressorRegistry
- Add custom_options member in CompressionOptions
  - Update parse/serialize functions to handle the new member
- Add tests
  - Define a SimpleRLECompressor implementing the Compressor interface
  - SimpleRLECompressor is also built into a shared library to test loading methods in CompressorRegistry

The Java API still has to be updated. It will be part of a follow-up PR.


adamretter:
@lucagiac81 This is very interesting. I am wondering if this is technically limited to just "Compression? It looks like I could implement a Compressor with the compress/decompress methods that would actually do encryption/decryption instead. Perhaps there is scope to generalise your approach a little further?

lucagiac81:
@adamretter That's a great point. In general, this should work with any reversible data transformation. For encryption specifically, RocksDB already supports pluggable providers, but there could be other uses.

include/rocksdb/advanced_options.h
```diff
@@ -132,6 +133,9 @@ struct CompressionOptions {
   // Default: 1.
   uint32_t parallel_threads;
 
+  // Options for custom compressors
+  std::unordered_map<std::string, std::string> custom_options;
+
```
mrambacher, line 138: If possible, it would be better to use the Options/OptionTypeInfo system to define specific options for the extended types.  I can work with you to get that set up.

db/db_impl/db_impl.cc
```diff
@@ -134,11 +134,17 @@ void DumpSupportInfo(Logger* logger) {
   ROCKS_LOG_HEADER(logger, "Compression algorithms supported:");
   for (auto& compression : OptionsHelper::compression_type_string_map) {
     if (compression.second != kNoCompression &&
+        compression.second != kCustomCompression &&
         compression.second != kDisableCompressionOption) {
       ROCKS_LOG_HEADER(logger, "\t%s supported: %d", compression.first.c_str(),
                        CompressionTypeSupported(compression.second));
     }
   }
+  for (auto& compressor :
```
mrambacher, line 143: const auto&

lucagiac81:
@mrambacher Thanks for your feedback! I started by splitting the PR into multiple commits. I'll work on the remaining items.

include/rocksdb/compressor.h
```diff
@@ -0,0 +1,213 @@
...
+    return reinterpret_cast<char*>(allocator->Allocate(size));
+  }
+  return new char[size];
+}
+
+class Compressor {
+ public:
+  virtual ~Compressor() = default;
+
+  virtual std::string Name() const = 0;
```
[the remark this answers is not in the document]
lucagiac81: Return type for Name updated

db/db_impl/db_impl.cc
lucagiac81: Updated

PR #6723, opened 2020-04-17
replace POWER crc32c with C implementation

grooverdan:
The presence of clang compile failures on POWER due to
missing ppc-asm.h prompted the replacement of the ASM
CRC32 implementation with the C implementation.

https://github.com/antonblanchard/crc32-vpmsum/blob/master/vec_crc32.c
only small include path change with the local copyright header maintained.

crc32c_ppc_constants.h per upstream generated, with assembler
compatible code removed. #pragma once per 64324e329eb0a9b4e77241a425a1615ff524c7f1
removed unnecessary header wasn't used in util/crc32c.cc.

util/crc32c.cc removes arch_ppc_crc32 which was only ever
used in a local context. Also incorporated significant advice from
@tchaikov in #2869. HAVE_POWER was never passed from cmake, so replaced
with the architected _ARCH_PWR8 directive. Altivec overloading removed.

Replaced arch_ppc_probe wrapper to used isAltivec directly.

Corrects getauxval detection from 8bbd76edbf by including the right
header (spelling) from the right directory.

From the crc32c_ppc.c (now replaced) comment:

  This wrapper function works around the fact that crc32_vpmsum
  does not gracefully handle the case where the data pointer is NULL.  There
  may be room for performance improvement here.

This isn't applicable provided the length is 0. Added test case for
this.


grooverdan:
FYI @adamretter. Optimized code wasn't being used with CMake based systems previously.

grooverdan:
updated in #7079

PR #6724, opened 2020-04-17
build: add option to turn off tests with debug

gocarlos:
why: currently there is not option to disable the tests with a debug build.

with this patch, we can set `WITH_TESTS` to OFF to not require gtest


gocarlos:
blocked by https://github.com/facebook/rocksdb/issues/6725

PR #6727, opened 2020-04-17
Disable O_DIRECT in stress test when db directory does not support direct IO

cheng-chang:
In crash test, the db directory might be set to /dev/shm or /tmp, in certain environments such as internal testing infrastructure, neither of these directories support direct IO, so direct IO is never enabled in crash test.

This PR sets up SyncPoints in direct IO related code paths to disable O_DIRECT flag in calls to `open`, so the direct IO code paths will be executed, all direct IO related assertions will be checked, but no real direct IO request will be issued to the file system.

Test Plan:
export CRASH_TEST_EXT_ARGS="--use_direct_reads=1 --mmap_read=0"
make -j24 crash_test


tools/db_crashtest.py
```diff
@@ -121,14 +121,32 @@
...
+            params["use_direct_io_for_flush_and_compaction"] == 1) and \
+            params["mmap_read"] == 0
 
 
 def get_dbname(test_name):
+    test_dir_name = "rocksdb_crashtest_" + test_name
     test_tmpdir = os.environ.get(_TEST_DIR_ENV_VAR)
     if test_tmpdir is None or test_tmpdir == "":
-        dbname = tempfile.mkdtemp(prefix='rocksdb_crashtest_' + test_name)
+        dbname = tempfile.mkdtemp(prefix=test_dir_name)
```
anand1976, line 138: This will leave the tmp directory around if dbname is overriden later with another directory for direct IO.

tools/db_crashtest.py
```diff
@@ -121,14 +121,32 @@
 }
 
 _TEST_DIR_ENV_VAR = 'TEST_TMPDIR'
+_DISK_TEMP_DIR_ENV_VAR = "DISK_TEMP_DIR"
+_HOME_ENV_VAR = "HOME"
+
+
+def need_direct_io_support(params):
+    return (params["use_direct_reads"] == 1 or \
```
anand1976, line 129: What if these parameter values are randomly chosen?

tools/db_crashtest.py
```diff
@@ -121,14 +121,32 @@
...
-        dbname = tempfile.mkdtemp(prefix='rocksdb_crashtest_' + test_name)
+        dbname = tempfile.mkdtemp(prefix=test_dir_name)
     else:
-        dbname = test_tmpdir + "/rocksdb_crashtest_" + test_name
+        dbname = test_tmpdir + "/" + test_dir_name
+        shutil.rmtree(dbname, True)
+        os.mkdir(dbname)
+    if need_direct_io_support(default_params) and \
+            not is_direct_io_supported(dbname):
+        shutil.rmtree(dbname, True)
```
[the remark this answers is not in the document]
cheng-chang: it delete the dir created by `os.mkdir(dbname)` in the previous `else` block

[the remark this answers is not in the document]
cheng-chang: also delete the tmp directory in the previous `if` block

tools/db_crashtest.py
cheng-chang: if it's overriden, `shutil.rmtree` will delete the dbname

tools/db_crashtest.py
cheng-chang: no matter how they are chosen, if the params satisfy these conditions, then direct IO support is needed. What's your concern on the random chosen params?

tools/db_crashtest.py
anand1976: This comment is not relevant anymore.

tools/db_crashtest.py
```diff
@@ -215,10 +207,10 @@ def finalize_and_sanitize(src_params):
         dest_params["compression_zstd_max_train_bytes"] = 0
     if dest_params.get("allow_concurrent_memtable_write", 1) == 1:
         dest_params["memtablerep"] = "skip_list"
-    if dest_params["mmap_read"] == 1 or not is_direct_io_supported(
-            dest_params["db"]):
+    if dest_params["mmap_read"] == 1:
```
anand1976, line 210: Sync points are only available in debug mode. We may still have to disable direct IO in release mode.

cheng-chang: do we `make crash_test` in release mode?

anand1976: I don't think we use release mode in the regression tests. But ```make crash_test``` can certainly be run with ```DEBUG_LEVEL=0```. A simple way would be to just check ```DEBUG_LEVEL``` and disable direct IO if necessary.

db_stress_tool/db_stress_tool.cc
```diff
@@ -45,6 +45,10 @@ int db_stress_tool(int argc, char** argv) {
   SanitizeDoubleParam(&FLAGS_memtable_prefix_bloom_size_ratio);
   SanitizeDoubleParam(&FLAGS_max_bytes_for_level_multiplier);
 
+  if (FLAGS_use_direct_reads || FLAGS_use_direct_io_for_flush_and_compaction) {
+    test::SetupSyncPointsToMockDirectIO();
```
anand1976, line 49: Can we do this conditionally through a separate command line argument set by db_crashtest.py? Otherwise, it precludes the ability to run crash_test with end-to-end direct IO.

tools/db_crashtest.py
```diff
@@ -124,25 +124,17 @@
...
+        dbname = tempfile.mkdtemp(prefix=test_dir_name)
     else:
-        dbname = test_tmpdir + "/rocksdb_crashtest_" + test_name
+        dbname = test_tmpdir + "/" + test_dir_name
         shutil.rmtree(dbname, True)
         os.mkdir(dbname)
     return dbname
 
 
-def is_direct_io_supported(dbname):
```
anand1976, line 137: It might be better to keep this, but just call it if ```DEBUG_LEVEL == 0``` .

tools/db_crashtest.py
cheng-chang: in release mode, instead of silently disabling direct IO, I prefer to just exit(1) here. Because if the user specifies direct IO explicitly, we shouldn't silently disable direct IO just because the db directory does not support direct IO.

db_stress_tool/db_stress_tool.cc
cheng-chang: added --mock_direct_io

cheng-chang:
@anand1976  PTAL

tools/db_crashtest.py
```diff
@@ -64,7 +64,8 @@
     "max_bytes_for_level_base": 10485760,
     "max_key": 100000000,
     "max_write_buffer_number": 3,
-    "mmap_read": lambda: random.randint(0, 1),
+    #"mmap_read": 1,lambda: random.randint(0, 1),
+    "mmap_read": 1,
```
anand1976, line 68: Unintentional?

siying:
Can you update the title to represent the new changes?

PR #6728, opened 2020-04-17
Add Github Action to make sure TARGETS is up-to-date and PR is code-style-conformant

riversand963:
Summary:
On the one hand, The TARGETS file is used for internal buck, and we do not
manually update it. On the other hand, we need to run the buckifier scripts to
update TARGETS whenever new files are added, etc. With this Github Action, we
make sure that every PR does not forget this step.

Test Plan:
Watch for Github Action result.


riversand963:
According to https://help.github.com/en/actions/reference/events-that-trigger-workflows,
it says the following.
```
When you create a pull request from a forked repository to the base repository, GitHub sends the pull_request event to the base repository and no pull request events occur on the forked repository
```
hmm...

riversand963:
Moving to #6761 . Closing this one.

PR #6730, opened 2020-04-18
Add default FileOptions::operator=

pallas:
Generation of an implicitly defined copy assignment operator is deprecated
in the presence of a user declared copy constructor.  Therefore, explicitly
add the default operator= to FileOptions.

Signed-off-by: Derrick Pallas <derrick@pallas.us>


pallas:
CLA signed.

PR #6732, opened 2020-04-18
Fix FilterBench when RTTI=0

pallas:
The dynamic_cast in the filter benchmark causes release mode to fail due to
no-rtti.  Replace with static_cast_with_check.

Signed-off-by: Derrick Pallas <derrick@pallas.us>

Addition by peterd: Remove unnecessary 2nd template arg on all static_cast_with_check


util/filter_bench.cc
```diff
@@ -378,7 +381,7 @@ void FilterBench::Go() {
       info.filter_ = info.plain_table_bloom_->GetRawData();
     } else {
       if (!builder) {
-        builder.reset(&dynamic_cast<BuiltinFilterBitsBuilder &>(*GetBuilder()));
+        builder.reset(static_cast_with_check<BuiltinFilterBitsBuilder, FilterBitsBuilder>(GetBuilder()));
```
pdillinger, line 384: The second template argument should not be needed, unless some compiler needs it (which apparently we would not be compiling db_flush_test.cc with).

PR #6746, opened 2020-04-22
Update CURRENT file after best-efforts recovery

riversand963:
After a successful recovery, the CURRENT file should be updated to point to the valid MANIFEST.

Test plan:
make check

### DIFF
--- a/db/db_impl/db_impl_files.cc
+++ b/db/db_impl/db_impl_files.cc
@@ -705,22 +705,21 @@ Status DBImpl::CleanupFilesAfterRecovery() {
     versions_->next_file_number_.store(largest_file_number + 1);
   }
 
-#if 0
   VersionEdit edit;
   edit.SetNextFile(versions_->next_file_number_.load());
   assert(versions_->GetColumnFamilySet());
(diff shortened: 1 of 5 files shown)


db/db_impl/db_impl_files.cc
```diff
@@ -704,8 +704,22 @@ Status DBImpl::CleanupFilesAfterRecovery() {
...
   }
+
+  VersionEdit edit;
+  edit.SetNextFile(versions_->next_file_number_.load());
+  assert(versions_->GetColumnFamilySet());
+  ColumnFamilyData* default_cfd = versions_->GetColumnFamilySet()->GetDefault();
+  assert(default_cfd);
+  // Even if new_descriptor_log is false, we will still switch to a new
+  // MANIFEST and update CURRENT file, since this is in recovery.
+  Status s = versions_->LogAndApply(
```
anand1976, line 715: It feels odd to put this code in DBImpl::CleanupFilesAfterRecovery(). Maybe change the name of the function to better capture this, like FinishRecovery().

Issue #6745, opened 2020-04-22
Undefined behavior (maybe crash?) in MultiGet


pdillinger:
#6483 from @hliu18 adds a 'timestamps' parameter to MultiGet. There appears to be UB in the DBImpl::MultiGet implementation taking a num_keys parameter.

https://github.com/facebook/rocksdb/blob/4cbc19d2a1da1a44775f27c2d167a0d8da0fd9f5/db/db_impl/db_impl.cc#L1999

If 'timestamps' is nullptr, as in the call up just a bit in the code, then this line is doing pointer arithmetic on nullptr, which is technically UB but if never used or compared to nullptr, then probably no crash or logic error in practice.

TBD whether this can cause a crash or logic error

hliu18:
Good catch. I'll create a PR to fix this.

hliu18:
created PR #6747.

PR #6747, opened 2020-04-22
check timestamps pointer for nullptr

hliu18:


hliu18:
@pdillinger @riversand963 
This is fix for issue #6745.

pdillinger:
Even if this is a suitable fix, we need to understand impact on 6.9 internal release, and add a regression test here if there's a way to cause a crash.

hliu18:
When timestamp feature is not enabled, crash cannot happen. The pointer is only dereferenced when timestamp size from user comparator is greater than 0. Current MultiGet() call with no timestamp maps into MultiGet() call with null timestamp pointer. We would have seen crashes if it can happen,

riversand963:
I believe the multiget timestamp support has at least two other separate issues. When construction `LookupKeys` in the constructor of `MultiGetContext`, timestamp in `ReadOptions` has to be taken into account. Otherwise, wrong result will be returned.
Another issue is some overloaded versions of MultiGet() will sort `KeyContext` objects using custom compare logic. In these logic, we need to compare key without timestamp. This means we cannot call `Comparator::Compare` which assumes that the arguments have timestamps when the feature is enabled.
I have put up another fix in #6748 . In order to make the unit test pass, I have to include the fix in this PR. cc @hliu18 @pdillinger

hliu18:
@riversand963 since your #6748 includes this fix, I'll just abandon this one.

PR #6748, opened 2020-04-23
Fix timestamp support for MultiGet

### LINKED ISSUE #6745: Undefined behavior (maybe crash?) in MultiGet

riversand963:
1. Avoid nullptr dereference when passing timestamp to KeyContext creation.
2. Construct LookupKey correctly with timestamp when creating MultiGetContext.
3. Compare without timestamp when sorting KeyContexts.

Fixes #6745 

Test plan (dev server):
make check

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,7 +5,6 @@
 * Finish implementation of BlockBasedTableOptions::IndexType::kBinarySearchWithFirstKey. It's now ready for use. Significantly reduces read amplification in some setups, especially for iterator seeks.
 * Fix a bug by updating CURRENT file so that it points to the correct MANIFEST file after best-efforts recovery.
 * Fixed a bug where ColumnFamilyHandle objects were not cleaned up in case an error happened during BlobDB's open after the base DB had been opened.
-* Fix a bug caused by trying to dereference nullable pointer (timestamp argument) in DB::MultiGet.
 * Fix a potential undefined behavior caused by trying to dereference nullable pointer (timestamp argument) in DB::MultiGet.
 * Fix a bug caused by not including user timestamp in MultiGet LookupKey construction. This can lead to wrong query result since the trailing bytes of a user key, if not shorter than timestamp, will be mistaken for user timestamp.
 * Fix a bug caused by using wrong compare function when sorting the input keys of MultiGet with timestamps.
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,7 @@
 * Finish implementation of BlockBasedTableOptions::IndexType::kBinarySearchWithFirstKey. It's now ready for use. Significantly reduces read amplification in some setups, especially for iterator seeks.
 * Fix a bug by updating CURRENT file so that it points to the correct MANIFEST file after best-efforts recovery.
 * Fixed a bug where ColumnFamilyHandle objects were not cleaned up in case an error happened during BlobDB's open after the base DB had been opened.
+* Fix a bug caused by trying to dereference nullable pointer (timestamp argument) in DB::MultiGet.
 
 ### Public API Change
 * Add a ConfigOptions argument to the APIs dealing with converting options to and from strings and files.  The ConfigOptions is meant to replace some of the options (such as input_strings_escaped and ignore_unknown_options) and allow for more parameters to be passed in the future without changing the function signature.
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,9 @@
 * Fix a bug by updating CURRENT file so that it points to the correct MANIFEST file after best-efforts recovery.
 * Fixed a bug where ColumnFamilyHandle objects were not cleaned up in case an error happened during BlobDB's open after the base DB had been opened.
 * Fix a bug caused by trying to dereference nullable pointer (timestamp argument) in DB::MultiGet.
+* Fix a potential undefined behavior caused by trying to dereference nullable pointer (timestamp argument) in DB::MultiGet.
+* Fix a bug caused by not including user timestamp in MultiGet LookupKey construction. This can lead to wrong query result since the trailing bytes of a user key, if not shorter than timestamp, will be mistaken for user timestamp.
+* Fix a bug caused by using wrong compare function when sorting the input keys of MultiGet with timestamps.
 
 ### Public API Change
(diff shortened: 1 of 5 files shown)


hliu18:
LGTM.

db/db_with_timestamp_basic_test.cc
```diff
@@ -774,6 +774,36 @@ TEST_F(DBBasicTestWithTimestamp, BatchWriteAndMultiGet) {
...
+  const size_t kTimestampSize = Timestamp(0, 0).size();
+  TestComparator test_cmp(kTimestampSize);
+  options.comparator = &test_cmp;
+  DestroyAndReopen(options);
+  WriteOptions write_opts;
+  std::string ts_str = Timestamp(1, 0);
+  Slice ts = ts_str;
+  write_opts.timestamp = &ts;
+  ASSERT_OK(db_->Put(write_opts, "foo", "value"));
+  ASSERT_OK(db_->Put(write_opts, "bar", "value"));
```
pdillinger, line 789: The failure is more interesting if the key is big enough to be mistaken for holding a timestamp: wrong result rather than assertion failure. I threw 16 X's after foo and bar.

Even better: I reproduced a crash:

    ASSERT_OK(db_->Put(write_opts, "foo", "value"));
    ASSERT_OK(db_->Put(write_opts, "bar", "value"));
    // Don't mistake trailing 16 chars in user key as timestamp
    ASSERT_OK(db_->Put(write_opts, "fooXXXXXXXXXXXXXXXX", "value"));
    ASSERT_OK(db_->Put(write_opts, "barXXXXXXXXXXXXXXXX", "value"));
    ...
    Slice keys[] = {"fooXXXXXXXXXXXXXXXX", "barXXXXXXXXXXXXXXXX"};

riversand963: I updated the test adding the two longer keys as shown above, but could not see a crash. Could you share the test that caused the crash?

pdillinger: https://github.com/pdillinger/rocksdb/tree/fix-nullptr-multiget

pdillinger: Can you please include a comment for why the 16 x's, as I did?

pdillinger:
And add to HISTORY.md

db/db_with_timestamp_basic_test.cc
riversand963: Okay, I thought you reproduced a crash with this PR's fix. Will update.

HISTORY.md
```diff
@@ -4,6 +4,7 @@
 * Fix wrong result being read from ingested file. May happen when a key in the file happen to be prefix of another key also in the file. The issue can further cause more data corruption. The issue exists with rocksdb >= 5.0.0 since DB::IngestExternalFile() was introduced.
 * Finish implementation of BlockBasedTableOptions::IndexType::kBinarySearchWithFirstKey. It's now ready for use. Significantly reduces read amplification in some setups, especially for iterator seeks.
 * Fix a bug by updating CURRENT file so that it points to the correct MANIFEST file after best-efforts recovery.
+* Fix a bug caused by trying to dereference nullable pointer (timestamp argument) in DB::MultiGet.
```
pdillinger, line 7: I think this is both too scary and not scary enough.

It is too scary because the bug should only affect cases in which ReadOptions::timestamp is set, right? There's technically UB for all callers using default timestamps argument, but I don't think it's an issue in practice.

It's not scary enough because this wording suggests people are only affected if they've seen a crash. I mentioned how this could but could lead to wrong result. Here's the code: https://github.com/pdillinger/rocksdb/tree/fix-nullptr-multiget-alt

riversand963: Actually there are two bugs here. 1) trying to deref a nullptr, 2) incorrectly construction of LookupKey during MultiGet with timestamp. The wrong result you observed comes from 2).
Will update the HISTORY.md accordingly.

pdillinger: I see on the wiki:

> Note that if the database enables timestamp, then caller of Get() should set ReadOptions.timestamp.

> When using Put with user timestamp, the user needs to set WriteOptions.timestamp.

My quick testing indicates these are not enforced with debug-build assertions. Why not?

riversand963: This PR fixes 3 places, thus updated HISTORY.md with 3 additional entries. Hopefully this is fine. Thanks @pdillinger for the review.

HISTORY.md
```diff
@@ -4,6 +4,9 @@
 * Fix wrong result being read from ingested file. May happen when a key in the file happen to be prefix of another key also in the file. The issue can further cause more data corruption. The issue exists with rocksdb >= 5.0.0 since DB::IngestExternalFile() was introduced.
 * Finish implementation of BlockBasedTableOptions::IndexType::kBinarySearchWithFirstKey. It's now ready for use. Significantly reduces read amplification in some setups, especially for iterator seeks.
 * Fix a bug by updating CURRENT file so that it points to the correct MANIFEST file after best-efforts recovery.
+* Fix a potential undefined behavior caused by trying to dereference nullable pointer (timestamp argument) in DB::MultiGet.
```
pdillinger, line 7: I think it's better to say "caused by pointer arithmetic on a nullable pointer (timestamps argument)" but OK

riversand963:
We should add these assertions. Let me add them.

pdillinger:
───────────────────████
───────────────────█████
───────────────────██████
───────────────────███████
──────────────────████████
──────────────────████████
─────────────────█████████
────────────────█████████
───────────────█████████
──────────────█████████
──────────────██████████████████
────────────█████████████████████
───────────███████████████████████
████████─██████████████████████████
████████─███████████████████████████
████████─████████████████████████████
████████─████████████████████████████
████████─████████████████████████████
████████─████████████████████████████
████████─███████████████████████████
████████─██████████████████████████
████████─█████████████████████████
████████─████████████████████████
████████─███████████████████████

PR #6749, opened 2020-04-23
Make sure (Shared)BlobFileMetaData are owned by shared_ptrs

ltamasi:
Summary:
The patch makes a couple of small cleanups to `SharedBlobFileMetaData` and `BlobFileMetaData`:
* It makes the constructors private and introduces factory methods to ensure these objects are always owned by `shared_ptr`s. Note that `SharedBlobFileMetaData` has an additional factory that takes a deleter object; we can utilize this to e.g. notify `VersionSet` when a blob file becomes obsolete (which is exactly when `SharedBlobFileMetaData` is destroyed).
* It disables move operations explicitly instead of relying on them being suppressed because of a user-declared destructor.

Test Plan:
`make check`

### DIFF
--- a/db/blob/blob_file_meta.h
+++ b/db/blob/blob_file_meta.h
@@ -23,15 +23,25 @@ namespace ROCKSDB_NAMESPACE {
 
 class SharedBlobFileMetaData {
  public:
-  SharedBlobFileMetaData(uint64_t blob_file_number, uint64_t total_blob_count,
-                         uint64_t total_blob_bytes, std::string checksum_method,
-                         std::string checksum_value)
-      : blob_file_number_(blob_file_number),
-        total_blob_count_(total_blob_count),
-        total_blob_bytes_(total_blob_bytes),
-        checksum_method_(std::move(checksum_method)),
-        checksum_value_(std::move(checksum_value)) {
-    assert(checksum_method_.empty() == checksum_value_.empty());
+  static std::shared_ptr<SharedBlobFileMetaData> Create(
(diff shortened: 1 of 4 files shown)


db/blob/blob_file_meta.h
```diff
@@ -68,21 +90,19 @@ std::ostream& operator<<(std::ostream& os,
...
-      : shared_meta_(std::move(shared_meta)),
-        garbage_blob_count_(garbage_blob_count),
-        garbage_blob_bytes_(garbage_blob_bytes) {
-    assert(shared_meta_);
-    assert(garbage_blob_count_ <= shared_meta_->GetTotalBlobCount());
-    assert(garbage_blob_bytes_ <= shared_meta_->GetTotalBlobBytes());
+  static std::shared_ptr<BlobFileMetaData> Create(
+      std::shared_ptr<SharedBlobFileMetaData> shared_meta,
+      uint64_t garbage_blob_count, uint64_t garbage_blob_bytes) {
+    return std::shared_ptr<BlobFileMetaData>(new BlobFileMetaData(
```
riversand963, line 96: How about `std::make_shared`?

ltamasi: As it stands right now, `make_shared` does not work with private constructors, unfortunately. One can't even (portably) work around this because it's not `make_shared` itself that creates the new object but some implementation-specific internal helper.

P.S. I recall seeing a proposal for the C++ standard some time back that would fix this.

PR #6752, opened 2020-04-24
[db_stress] Enable Trace (--trace_path) in db_stress

yhchiang:
Summary:
Add --trace_path to enable trace in db_stress.

  -trace_path (Path to the RocksDB Trace file.  If specified, RocksDB Trace
    feature will be enabled, and the trace will be stored in the specified
    file path.  In case db_stress is run with more than one reopen, the trace
    feature will currently only keep the latest trace.) type: string
    default: ""

Test Plan:
$ ./db_stress --ops_per_thread=100 --reopen=2 --compression_type=zlib --trace_path=/tmp/rocksdbtest-501/trace

Verify the trace file is created after running the above command.

$ ls -l /tmp/rocksdbtest-501/trace
-rw-r--r--  1 yhchiang  wheel  198975928 Apr 23 23:05 /tmp/rocksdbtest-501/trace

Full paste: P129633148

Reviewers: ajkr


db_stress_tool/db_stress_test_base.cc
```diff
@@ -2165,5 +2172,67 @@ void StressTest::Reopen(ThreadState* thread) {
...
+  }
+
+  if (trace_enabled_ == false) {
+    fprintf(stdout, "--trace_path is disabled.\n");
+    return;
+  }
+
+  EnvOptions trace_env_options(options_);
+  std::unique_ptr<TraceWriter> trace_writer;
+
```
zhichao-cao, line 2205: Should we set a trace file limit here?

yhchiang: Sure, I can add an option here.

yhchiang: It seems only the BlockCacheTrace has the file limit while TraceWriter does not.  Am I right at this part?

yhchiang: WIthout setting it it will use the default 64GB limit.

db_stress_tool/db_stress_test_base.cc
```diff
@@ -2165,5 +2170,76 @@ void StressTest::Reopen(ThreadState* thread) {
...
+    trace_enabled_ = false;
+    fprintf(stdout, "NewFileTraceWriter() failed -- %s\n",
+        s.ToString().c_str());
+    return;
+  }
+
+  s = db_->StartTrace(TraceOptions(), std::move(trace_writer));
+  if (!s.ok()) {
+    trace_enabled_ = false;
+    fprintf(stdout, "StartTrace() failed -- %s\n",
```
ajkr, line 2216: Are these failure/incompatibility messages stderr worthy?

yhchiang:
> Does TraceWriter have any in-process buffering? I am wondering if there's possibility of losing the tail of the trace when this is used in crash test and the db_stress is killed randomly.

I can double check that.  But if the way we verify the data is everything before the largest seq_num in the DB, then it probably doesn't matter much?

db_stress_tool/db_stress_common.cc
```diff
@@ -13,6 +13,7 @@
 #include <cmath>
 
 ROCKSDB_NAMESPACE::DbStressEnvWrapper* db_stress_env = nullptr;
+ROCKSDB_NAMESPACE::DbStressEnvWrapper* trace_env = nullptr;
```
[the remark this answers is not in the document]
yhchiang: I can add a comment here.  The main purpose of having a different env is because db_stress_env might enable fault_injection, while we don't want the trace_env to have any fault-injection.

In other words, if the fault-injection is off, then trace_env will be identical to db_stress_env.

db_stress_tool/db_stress_test_base.cc
yhchiang: Do you mean moving from stdout to stderr?  Or completely remove the message?

yhchiang: > That's interesting, I guess we want db_stress to fail if it's configured to enable trace file and the limit is exceeded. Also how does the limit work when the file is closed and reopened?

So it will stop tracing new writes but keep the existing (earlier) trace.

ajkr: I meant moving from stdout to stderr.

PR #6755, opened 2020-04-24
Keep track of obsolete blob files in VersionSet

ltamasi:
Summary:
The patch adds logic to keep track of obsolete blob files. A blob file becomes
obsolete when the last `shared_ptr` that points to the corresponding
`SharedBlobFileMetaData` object goes away, which, in turn, happens when the
last `Version` that contains the blob file is destroyed. No longer needed blob
files are added to the obsolete list in `VersionSet` using a custom deleter to
avoid unnecessary coupling between `SharedBlobFileMetaData` and `VersionSet`.
Obsolete blob files are returned by `VersionSet::GetObsoleteFiles` and stored
in `JobContext`.

Test Plan:
`make check`

### DIFF
--- a/db/version_builder.cc
+++ b/db/version_builder.cc
@@ -90,6 +90,7 @@ class VersionBuilder::Rep {
   Logger* info_log_;
   TableCache* table_cache_;
   VersionStorageInfo* base_vstorage_;
+  VersionSet* version_set_;
   int num_levels_;
   LevelState* levels_;
   // Store states of levels larger than num_levels_. We do this instead of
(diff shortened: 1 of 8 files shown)


db/version_builder.cc
```diff
@@ -388,11 +392,29 @@ class VersionBuilder::Rep {
...
+    VersionSet* const vs = version_set_;
+    const ImmutableCFOptions* const ioptions = ioptions_;
+
+    auto deleter = [vs, ioptions](SharedBlobFileMetaData* shared_meta) {
+      if (vs) {
+        assert(ioptions);
+        assert(!ioptions->cf_paths.empty());
+        assert(shared_meta);
+
+        vs->AddObsoleteBlobFile(shared_meta->GetBlobFileNumber(),
```
riversand963, line 406: So blob file is always in the `cf_paths.front()`?

ltamasi: Yes, that's the plan for now at least, since blob files are not associated with levels. We can always add multiple paths support later if needed.

PR #6756, opened 2020-04-24
Fix a valgrind failure due to DBBasicTestMultiGetDeadline

anand1976:
Fix a valgrind failure.

Test:
valgrind_test


zhichao-cao:
Thanks for the fix. Can you also Fix the 
"env/io_posix.cc:613:28: error: unused variable ‘r’ [-Werror=unused-variable]
       const FSReadRequest& r = reqs[i];
" It fails the Travis and also clang contrun

Issue #6753, opened 2020-04-24
import_column_family bug


x-x-p:
when sorted_files.size() is 0,  sorted_files.size() - 1 is overflow， sorted_files[i + 1] operator[] fatal

https://github.com/facebook/rocksdb/blob/e04f3bce4fbe0453457047ea36e858888f0343e9/db/import_column_family_job.cc#L65-L70

zhichao-cao:
@x-x-p Yes, you are right, size_t is unsigned. Will create a PR to fix it.

PR #6762, opened 2020-04-27
Fix potential size_t overflow in import_column_family

zhichao-cao:
The issue is reported in #6753 . size_t is unsigned and if sorted_file.size() is 0, the end condition of i will be extremely large, cause segment fault in sorted_files[i] and sorted_files[i+1]. Added condition to fix it.

test plan: make asan_check

### DIFF
--- a/db/import_column_family_job.cc
+++ b/db/import_column_family_job.cc
@@ -62,7 +62,7 @@ Status ImportColumnFamilyJob::Prepare(uint64_t next_file_number,
                        info2->smallest_internal_key) < 0;
           });
 
-      for (size_t i = 0; i < sorted_files.size() - 1; i++) {
+      for (size_t i = 0; i + 1 < sorted_files.size(); i++) {
         if (cfd_->internal_comparator().Compare(
                 sorted_files[i]->largest_internal_key,
                 sorted_files[i + 1]->smallest_internal_key) >= 0) {


db/import_column_family_job.cc
```diff
@@ -61,12 +61,13 @@ Status ImportColumnFamilyJob::Prepare(uint64_t next_file_number,
...
                        info2->smallest_internal_key) < 0;
           });
-
-      for (size_t i = 0; i < sorted_files.size() - 1; i++) {
-        if (cfd_->internal_comparator().Compare(
-                sorted_files[i]->largest_internal_key,
-                sorted_files[i + 1]->smallest_internal_key) >= 0) {
-          return Status::InvalidArgument("Files have overlapping ranges");
+      if (sorted_files.size() > 1) {
+        for (size_t i = 0; i < sorted_files.size() - 1; i++) {
```
pdillinger, line 65: How about `i + 1 < sorted_files.size()` instead?

And it looks like there about about 24 more that would could apply this to pro-actively, `git grep -E 'for [(](size_t|uint|unsigned).* (<|!=) .*- 1;'`

pdillinger: > And it looks like there about about 24 more that would could apply this to pro-actively

Feel free to leave this to a separate PR/task so that this one can land.

zhichao-cao: @pdillinger Thanks for the review. Will change to i + 1 < sorted_files.size() in the for loop. Yeah, I will create separate PR for the other potential overflows you mentioned. Thanks!

PR #6765, opened 2020-04-28
Auto resume the DB from Retryable IO Error

zhichao-cao:
In current codebase, in write path, if Retryable IO Error happens, SetBGError is called. The retryable IO Error is converted to hard error and DB is in read only mode. User or application needs to resume it. In this PR, if Retryable IO Error happens in one DB, SetBGError will create a new thread to call Resume (auto resume). otpions.max_bgerror_resume_count controls if auto resume is enabled or not (if max_bgerror_resume_count<=0, auto resume will not be enabled). options.bgerror_resume_retry_interval controls the time interval to call Resume again if the previous resume fails due to the Retryable IO Error. If non-retryable error happens during resume, auto resume will terminate.

Test plan: Added the unit test cases in error_handler_fs_test and pass make asan_check


db/compaction/compaction_job.cc
```diff
@@ -614,14 +618,23 @@ Status CompactionJob::Run() {
...
   IOStatus io_s;
+  for (const auto& state : compact_->sub_compact_states) {
+    if (!state.io_status.ok()) {
+      io_s = state.io_status;
+      break;
+    }
+  }
+  if (!io_s.ok()) {
+    io_status_ = io_s;
+    status = io_s;
```
zhichao-cao, line 629: @anand1976 I'm not sure here, if io_s is not ok, it is possible that status is already set to not ok at L615. Should we also do status = io_s here?

anand1976: Looks like in other places ```status``` takes precedence over ```io_s```, so we should do the same here for consistency

db/error_handler.cc
```diff
@@ -268,10 +270,13 @@ Status ErrorHandler::SetBGError(const IOStatus& bg_io_err,
     Status bg_err(new_bg_io_err, Status::Severity::kHardError);
     EventHelpers::NotifyOnBackgroundError(db_options_.listeners, reason, &s,
                                           db_mutex_, &auto_recovery);
+    if (recovery_in_prog_ && recovery_error_.ok()) {
+      recovery_error_ = bg_err;
+    }
     if (bg_err.severity() > bg_error_.severity()) {
```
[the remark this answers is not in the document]
zhichao-cao: @anand1976 Yes. In the testing, the auto resume is actually not the one that reschedule the compaction. The compaction thread reschedules it.

db/error_handler.cc
```diff
@@ -381,4 +386,125 @@ Status ErrorHandler::RecoverFromBGError(bool is_manual) {
...
+#ifndef ROCKSDB_LITE
+  InstrumentedMutexLock l(db_mutex_);
+  TEST_SYNC_POINT("RecoverFromRetryableBGIOError:BeforeStart");
+  if (recovery_in_prog_) {
+    return;
+  }
+  if (end_recovery_) {
+    return;
+  }
+  recovery_in_prog_ = true;
```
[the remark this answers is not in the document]
zhichao-cao: @anand1976 Combine the previous comments, we always hold the db_mutex through `StartRecoveryFromRetryableBGIOError`, test recovery_in_prog_, and set it. If so, we do not need  `recovery_thread_->join()` at L403 to wait current running thread. recovery_in_prog_ will ensure only one thread will be created at any time. Do I understand correctly?

db/compaction/compaction_job.cc
```diff
@@ -614,14 +618,27 @@ Status CompactionJob::Run() {
   }
 
   IOStatus io_s;
+  for (const auto& state : compact_->sub_compact_states) {
```
anand1976, line 621: Do we need to duplicate this loop here? If the ```io_status``` for any of the ```sub_compact_states``` was non-ok, it would be reflected in their ```status``` right? So can we update ```io_s``` right after line 615?

zhichao-cao: @anand1976 Yes, this loop is unnecessary. Will combine with the previous loop to get io_s.

db/error_handler.cc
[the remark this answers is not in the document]
zhichao-cao: @anand1976 The `bg_error_` is set to Status::Severity::kHardError if the previous one is a Retryable error. Do you mean we need to test `BackgroundErrorReason reason` and set to `kSoftError` for compaction case here?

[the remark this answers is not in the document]
anand1976: Since a compaction failure doesn't immediately affect future writes to the DB, the error severity can be set to ```kSoftError```. Ideally, we can do the same even for flush errors if WAL is not in use, so a user that doesn't use WAL will see no impact due to retryable errors. It can be done in a follow-on PR if that makes it easier.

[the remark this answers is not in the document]
zhichao-cao: @anand1976 Sure, will create a task to track it and do it in a follow-on PR.

db/compaction/compaction_job.cc
zhichao-cao: @anand1976 Updated

db/error_handler.cc
```diff
@@ -381,4 +393,122 @@ Status ErrorHandler::RecoverFromBGError(bool is_manual) {
...
+  }
+  if (db_options_.max_bgerror_resume_count <= 0 || recovery_in_prog_ ||
+      recovery_thread_) {
+    // Auto resume BG error is not enabled, directly return bg_error_.
+    return bg_error_;
+  }
+
+  recovery_in_prog_ = true;
+  recovery_thread_.reset(
+      new port::Thread(&ErrorHandler::RecoverFromRetryableBGIOError, this));
```
riversand963, line 411: Imagine a process with many RocksDB instances. Is it possible that the underlying system starts to return retryable errors to each of them from a certain time? If so, the process will spawn a large number of such unmanaged threads, which can be a concern.

db/error_handler.cc
```diff
@@ -381,4 +393,122 @@ Status ErrorHandler::RecoverFromBGError(bool is_manual) {
...
+  }
+#else
+  (void)io_error;
+  return bg_error_;
+#endif
+}
+
+// Automatic recover from Retryable BG IO error. Must be called after db
+// mutex is released.
+void ErrorHandler::RecoverFromRetryableBGIOError() {
```
riversand963, line 427: The thread executing this function is started while holding the mutex, thus this method should not acquire mutex again.

db/error_handler.cc
zhichao-cao: @riversand963 Yeah. Good point. We have the discussion before about this concern. The decision is, in the current stage, it is ok to have one thread for one DB (for some cases, might be hundred DB in a host), since all DB is in read mode. No flush or compaction will be scheduled. But in the next step, we will put them in a shared thread pool or other way to solve this potential isssue.

db/error_handler.cc
zhichao-cao: The parent thread new this child thread when holding the db mutex. But we do not use the join to wait. So parent will continue hold the db mutex and come back to the caller. In this child thread, db mutex is needed to protect the variables like recovery_io_error_, recovery_error. Also, to call db_->ResumeImpl() and time wait, this thread still need the db mutex. Why it should not acquire mutex again?

include/rocksdb/options.h
```diff
@@ -1144,6 +1144,22 @@ struct DBOptions {
   // not be used for recovery if best_efforts_recovery is true.
   // Default: false
   bool best_efforts_recovery = false;
+
+  // It defines How many times call db resume when retryable BGError happens.
```
ajkr, line 1148: Does somebody want to tune these knobs? How about infinite recovery attempts with a one second gap?

anand1976: Yes, that's a good idea. We probably just need an option to enable/disable the functionality in case of bugs.

zhichao-cao: @anand1976 You mean, I can remove the options of max_bgerror_resume_count and bgerror_resume_retry_interval, set value directly in the resume function. Instead, we give a option for user to enable or disable auto resume? The resume will be infinite attempts as suggested by Andrew?

anand1976: @zhichao-cao Thinking about this some more, I think there's value in allowing the retry interval to be adjusted. The flushes initiated by the recovery threads share the same background thread pool as regular flushes, so if you have a mix of good and bad DBs, changing the retry interval will allow a user to control resource consumption for recovery. The retry count is debatable. We could set the default to infinite and user can set it to 0 to disable. I don't know if there would be a reason to choose some value in between. Again, maybe to limit resource consumption.

Issue #6628, opened 2020-04-01
fix build with gflags 2.1.x


ajkr:
With either make or cmake, building against gflags 2.1.x leads to errors like:

```
In file included from /home/andrewkr/local/rocksdb/tools/trace_analyzer_tool.cc:49:0:
/home/andrewkr/local/rocksdb/util/gflags_compat.h:11:26: error: ‘google’ has not been declared
 #define GFLAGS_NAMESPACE google
```

In gflags 2.1, the namespace changed from google to gflags. Meanwhile, it wasn't until 2.2 that the macro `#define GFLAGS_NAMESPACE ...` was introduced to a header file. So for 2.1 versions, this logic improperly guesses the namespace: https://github.com/facebook/rocksdb/blob/57096ab13e9d373666f099f19d6eb73c6613e797/util/gflags_compat.h#L8-L12.

iFA88:
Hey @ajkr , sorry but i have still issue with this workaround `-D GFLAGS_NAMESPACE=gflags` :
```
CMake Warning:
  Manually-specified variables were not used by the project:

    GFLAGS_NAMESPACE
```
After build i can not use the library:
`/usr/lib/x86_64-linux-gnu/librocksdb.so.6: undefined symbol: _ZN6google21ParseCommandLineFlagsEPiPPPcb`

libgflags-dev has v2.1.2-4 on debian 9.

Rocksdb v6.8.1 works fine.

ajkr:
For the workaround, `-DGFLAGS_NAMESPACE` should be provided as a compiler flag, not as a CMake build option.

PR #6771, opened 2020-04-29
Use GFlags/Snappy config in CMake, with fallback for legacy approach.

xkszltl:
Related to some discussion in https://github.com/facebook/rocksdb/issues/6179


CMakeLists.txt
```diff
@@ -109,16 +109,28 @@ else()
...
+      include_directories(${gflags_INCLUDE_DIR})
+    endif()
     add_definitions(-DGFLAGS=1)
-    include_directories(${gflags_INCLUDE_DIR})
-    list(APPEND THIRDPARTY_LIBS gflags::gflags)
   endif()
 
   if(WITH_SNAPPY)
-    find_package(snappy REQUIRED)
+    find_package(Snappy CONFIG)
```
tchaikov, line 125: @xkszltl since you are here, could you make Snappy `REQUIRED` here? as `WITH_SNAPPY` is OFF by default, so if it's enabled because it is explicitly requested by user, if it's not found, we should bail out and fail.

CMakeLists.txt
```diff
@@ -109,16 +109,28 @@ else()
...
+    if(Snappy_FOUND)
+      list(APPEND THIRDPARTY_LIBS Snappy::snappy)
+    else()
+      message("Snappy config not found. Fallback to legacy find.")
+      find_package(snappy REQUIRED)
```
tchaikov, line 130: could you rename `cmake/modules/Findsnappy.cmake` to `cmake/modules/FindSnappy.cmake` and also apply following patch?
```diff
diff --git a/cmake/modules/Findsnappy.cmake b/cmake/modules/Findsnappy.cmake
index 6ed5fda3d..cb59b5e49 100644
--- a/cmake/modules/Findsnappy.cmake
+++ b/cmake/modules/Findsnappy.cmake
@@ -14,7 +14,7 @@ find_library(SNAPPY_LIBRARIES
   HINTS ${SNAPPY_ROOT_DIR}/lib)
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(snappy DEFAULT_MSG SNAPPY_LIBRARIES SNAPPY_INCLUDE_DIR)
+find_package_handle_standard_args(Snappy DEFAULT_MSG SNAPPY_LIBRARIES SNAPPY_INCLUDE_DIR)
 
 mark_as_advanced(
   SNAPPY_LIBRARIES
```

CMakeLists.txt
xkszltl: No. We should allow CONFIG to failed if missing, but fallback path (general lookup, i.e. try MODULE then CONFIG) should have REQUIRED as final guard.

CMakeLists.txt
xkszltl: Nice! As long as you're OK with that~

cmake/modules/Findsnappy.cmake
```diff
@@ -1,29 +1,29 @@
 # - Find Snappy
 # Find the snappy compression library and includes
 #
-# snappy_INCLUDE_DIRS - where to find snappy.h, etc.
-# snappy_LIBRARIES - List of libraries when using snappy.
-# snappy_FOUND - True if snappy found.
+# Snappy_INCLUDE_DIRS - where to find snappy.h, etc.
```
tchaikov, line 4: please rename this file as well.

xkszltl: Renamed

tchaikov:
```
[ RUN      ] DBBasicTestMultiGetDeadline.MultiGetDeadlineExceeded
1635
1636
1637No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
1638Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
1639
1640The build has been terminated
1641
```

```
Too long with no output (exceeded 10m0s): context deadline exceeded
```

i don't think the failures on MacOS and Windows are related. but they are still annoying though..

xkszltl:
Should be some kind of network timeout. Also get this in other PR.

CMakeLists.txt
```diff
@@ -109,16 +109,26 @@ else()
 
   # No config file for this
   if(WITH_GFLAGS)
-    find_package(gflags REQUIRED)
+    option(GFLAGS_USE_TARGET_NAMESPACE "Use gflags import target with namespace." ON)
+    find_package(gflags CONFIG)
+    if(gflags_FOUND)
+      list(APPEND THIRDPARTY_LIBS ${GFLAGS_TARGET})
```
tchaikov, line 115: why not use `gflags::gflags` directly? as it's documented by https://github.com/gflags/gflags/blob/master/CMakeLists.txt#L19 as well and https://gflags.github.io/gflags/#cmake .

CMakeLists.txt
```diff
@@ -109,16 +109,26 @@ else()
...
+    else()
+      find_package(gflags REQUIRED)
+      list(APPEND THIRDPARTY_LIBS gflags::gflags)
+      include_directories(${gflags_INCLUDE_DIR})
```
tchaikov, line 119: there is no need to add this include globally.

CMakeLists.txt
xkszltl: It's because there're shared/static version of gflags, threaded/nonthreaded, and namespaces/plain (for old system) versions, this redirect to the actual one.
https://github.com/gflags/gflags/blob/addd749114fab4f24b7ea1e0f2f837584389e52c/cmake/config.cmake.in#L89-L109

It's also what gflags used officially: https://github.com/gflags/gflags/blob/addd749114fab4f24b7ea1e0f2f837584389e52c/cmake/config.cmake.in#L163

CMakeLists.txt
xkszltl: Agree, not sure why it was there previously.
It wasn't used anyway because the casing is wrong.

CMakeLists.txt
tchaikov: i am not sure if your reply addresses my question. how is multiple builds of gflags library related not using `gflags::gflags`?

xkszltl: OK I think you're right. The correct variation is also aliased to gflags::gflags.

CMakeLists.txt
```diff
@@ -109,16 +109,23 @@ else()
 
   # No config file for this
   if(WITH_GFLAGS)
-    find_package(gflags REQUIRED)
+    option(GFLAGS_USE_TARGET_NAMESPACE "Use gflags import target with namespace." ON)
```
tchaikov, line 112:
```cmake
  set(GFLAGS_USE_TARGET_NAMESPACE ON CACHE BOOL "" FORCE)
```
would suffice

CMakeLists.txt
```diff
@@ -109,16 +109,23 @@ else()
...
+    find_package(gflags CONFIG)
+    if(NOT gflags_FOUND)
+      find_package(gflags REQUIRED)
```
tchaikov, line 115: could you remind me what we might run into if we just call
```cmake
find_package(gflags REQUIRED)
```
without doing the "CONFIG" then "MODULE" dance?

CMakeLists.txt
xkszltl: Should we leave this to user in case they need it for whatever reason?
For example, if they want to deliver the built binary to system without new gflags (with namespacing available), `option` provides both our recommendation (ON) and freedom (set it to OFF without code change).

CMakeLists.txt
xkszltl: Because the FindXXX we provided is for fallback when XXXConfig is missing, not for overwriting the behavior of XXXConfig.
I think CMake chooses MODULE > CONFIG for backward compatibility (CMake 2 code expecting only FindXXX before XXXConfig is invented still gets FindXXX in CMake 3).

Another option is to set `CMAKE_FIND_PACKAGE_PREFER_CONFIG`, which changes the default to CONFIG > MODULE implicitly.
I'm not doing that because it also changes the behavior of both caller (RocksDB) and callee (gflags/snappy/... and their deps).

xkszltl: Hmmm...Double checked the doc, CONFIG keyword was available much earlier than I expected, but I never seen people use it at that time.

CMakeLists.txt
tchaikov: i assume that you want to support gflags v2.2.1 and lower. as target namespace support was added in v2.2.2. see https://github.com/gflags/gflags/tree/v2.2.2. so instead of exposing a gflag option to user of rocksdb, i'd prefer having something like
```cmake
set(GFLAGS_USE_TARGET_NAMESPACE ON CACHE BOOL "" FORCE)
find_package(gflags REQUIRED)
if(NOT TARGET gflags::gflags)
  add_library (gflags:: gflags ALIAS gflags)
endif()
```

and do the same thing in #6791

CMakeLists.txt
tchaikov: > Because the FindXXX we provided is for fallback when XXXConfig is missing, not for overwriting the behavior of XXXConfig.

i agree with you on this. but does it hurt to allow cmake to use the homebrew `Findfoo.cmake` even if `fooConfig` exists ?

xkszltl: Yes because usually upstream knows better about their own lib.
For example:
1. foo 1.0 and foo 1.1 may use different dir layout and FindXXX designed for 1.1 won't work for 1.0.
2. Here's a gflags compiled for shared (dll) only, but with both threading/serial version.
If user wants to use `gflags_nothreads.lib`, with upstream CONFIG all it take is to `-DGFLAGS_NOTHREADS=ON`.
The `Findgflags.cmake` in RocksDB repo doesn't work for that.
image
3. It's also possible for distro packager to compose special dir layout and export with different set of configs.
Use case include having libs built with multiple compiler runtime (e.g. gcc/llvm/icc, very common in HPC environment).

I think config file is an important feature of cmake ecosystem.
If upstream can tell you what to use, why hard-code it?

CMakeLists.txt
xkszltl: Oops, I think `gflags::gflags` doesn't work without `GFLAGS_USE_TARGET_NAMESPACE=ON`. `${GFLAGS_TARGET}` is still the correct one to use.

CMakeLists.txt
xkszltl: Note that CMake is stateful.
Injection to 3rd-party namespace can potentially mislead whatever `find_package()` call after `find_package(RocksDB)`, and should only be used if nothing else works.
In this case I think
```
if (...)
   list(APPEND ... gflags)
else()
   list(APPEND ... gflags::gflags)
endif()
```
should be the right way, which is equivalent to list(APPEND ... ${GFLAGS_TARGET})

But I agree that GFLAGS_USE_TARGET_NAMESPACE setting should also be written into config file.

xkszltl: Regarding using `set(...)`, I'm confused by the `FORCE`.
`option(XXX)` doesn't make `XXX` a "RocksDB option", it's simply a cached variable and you can also have them show up in your build screen if created by one of you dependency config files. 

Image this use case:
```
option(GFLAGS_USE_TARGET_NAMESPACE "" OFF)
find_package(Foo) # Old lib Foo wants gflags without namespace.
find_package(RocksDB)
# Count-intuitive to say option(GFLAGS_USE_TARGET_NAMESPACE "" OFF) again.
find_package(Bar) # Old lib Bar also wants gflags without namespace.
```
Since RocksDB doesn't have any hard dependency on the namespaced gflags, it's better to leave that to user for compatibility.

xkszltl: To summarize, this is what I'm trying to achieve:
1. Backward compatibility with old gflags (`${GFLAGS_TARGET}` for auto-fallback to `gflags` if `GFLAGS_USE_TARGET_NAMESPACE` is not recognized)
2. Forward compatibility (by not using `GFLAGS_USE_TARGET_NAMESPACE=OFF`) if namespace is required in the future since that's the trend of modern cmake.
3. Compatibility with other libs and allow user to decide if our choice is wrong.

CMakeLists.txt
xkszltl: Added GFLAGS_LIBRARIES support for 2.1

siying:
@tchaikov let me know once the PR is ready to merge.

CMakeLists.txt
tchaikov: > 2. Here's a gflags compiled for shared (dll) only, but with both threading/serial version.
>    If user wants to use `gflags_nothreads.lib`, with upstream CONFIG all it take is to `-DGFLAGS_NOTHREADS=ON`.
>    The `Findgflags.cmake` in RocksDB repo doesn't work for that.

very true, we are not able to compete with upstream config in this perspective.

CMakeLists.txt
```diff
@@ -107,18 +107,33 @@ else()
...
   endif()
 
-  # No config file for this
   if(WITH_GFLAGS)
-    find_package(gflags REQUIRED)
+    # Config with namespace available since gflags 2.2.2
+    option(GFLAGS_USE_TARGET_NAMESPACE "Use gflags import target with namespace." ON)
+    find_package(gflags CONFIG)
+    if(gflags_FOUND)
+      if(DEFINED ${GFLAGS_TARGET})
```
tchaikov, line 115:
```suggestion
      if(DEFINED GFLAGS_TARGET)
```

please note what we want to check is the *name* of "GFLAGS_TARGET", not the value of `GFLAGS_TARGET`.

tchaikov: probably 
```cmake
if(TARGET ${GFLAGS_TARGET})
```

is better?

CMakeLists.txt
```diff
@@ -107,18 +107,33 @@ else()
...
     add_definitions(-DGFLAGS=1)
-    include_directories(${gflags_INCLUDE_DIR})
-    list(APPEND THIRDPARTY_LIBS gflags::gflags)
   endif()
 
   if(WITH_SNAPPY)
-    find_package(snappy REQUIRED)
+    find_package(Snappy CONFIG)
+    if(NOT Snappy_FOUND)
+      message("Snappy config not found. Fallback to legacy find.")
```
tchaikov, line 132: i'd prefer avoid sending non-error messages to stderr. could you drop this line?

tchaikov:
> @tchaikov let me know once the PR is ready to merge.

@siying sure. i think we are close.

@xkszltl thanks for filling me with the details, updated my reviews.

CMakeLists.txt
xkszltl: OK

CMakeLists.txt
xkszltl: Aha good catch!

xkszltl: fixed

tchaikov:
@siying i think this PR is good to go.

@xkszltl thanks for your persistence and patience!

jay-zhuang:
Hi @xkszltl on my linux environment and also CI with gflags 2.1. It's unable to build.
Seems `find_package(gflags CONFIG)` is able find package (`gflags_FOUND`), but `GFLAGS_LIBRARIES` or `GFLAGS_TARGET` is none.

It works with `find_package(gflags REQUIRED)`.

xkszltl:
The casing is different:
```
# rpm -qf /usr/lib64/cmake/gflags/gflags-config.cmake
gflags-devel-2.1.1-6.el7.x86_64
# grep -e _INCLUDE -e _LIBRAR /usr/lib64/cmake/gflags/gflags-config.cmake 
set (gflags_INCLUDE_DIR "${_INSTALL_PREFIX}/include")
set (gflags_LIBRARIES gflags)
```

xkszltl:
You may need to have some aliasing code to fill in the version gap right after package lookup, to create new var, if missing, from old var.

xkszltl:
BTW I don't think https://github.com/facebook/rocksdb/issues/6628 has anything to do with the issue you mentioned here. They're different, but both due to the API change across versions.

jay-zhuang:
Change GFLAGS to lower case works on my local environment: https://github.com/facebook/rocksdb/pull/8324/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR126

Testing it on CI.

xkszltl:
@jay-zhuang 
As you can see from the case-insensitive grep I posted, there's no `gflags_INCLUDE_DIR`/`gflags_LIBRARIES` in 2.2+, and `gflags_TARGET` never existed in any version.
So the change you made in that PR is incorrect and will break 2.2 build.

Regarding CI, I would suggest to increase the coverage beyond Ubuntu 16.04.
CentOS 7/8 and Debian 9 have gflags 2.1.
Several recent Fedora 32/33/34, Debian 10/11, Ubuntu 18.04/20.04 all have gflags 2.2.
You may further look it up in https://pkgs.org/search/?q=gflags

xkszltl:
The correct way is probably to add more elif in https://github.com/facebook/rocksdb/blob/55853de661ce476281170ec90306b944df2234d9/CMakeLists.txt#L124-L127, to further fallback to `gflags_LIBRARIES` if `GFLAGS_LIBRARIES` is missing.
And you may move include_directories to the non-target if-else block, with different casing of var: https://github.com/facebook/rocksdb/blob/55853de661ce476281170ec90306b944df2234d9/CMakeLists.txt#L132

xkszltl:
This was the commit changing `gflags_` (2.1) to `GFLAGS_` (2.2), together with the introduction of `GFLAGS_TARGET`:
https://github.com/gflags/gflags/commit/2de6bf3904c3bc19b4dc40d91bb0c4a55cc52b50
`PACKAGE_NAME` is replaced by upper case `PACKAGE_PREFIX`.
https://github.com/gflags/gflags/blob/f8a0efe03aa69b3336d8e228b37d4ccb17324b88/CMakeLists.txt#L136

xkszltl:
That may safely conclude `GFLAGS_LIBRARIES` co-exists with `GFLAGS_TARGET`.
So a simple fix is to change:
https://github.com/facebook/rocksdb/blob/55853de661ce476281170ec90306b944df2234d9/CMakeLists.txt#L126
to
```
set(GFLAGS_LIB ${GFLAGS_LIBRARIES})
include_directories(${GFLAGS_INCLUDE_DIR})
```
and remove https://github.com/facebook/rocksdb/blob/55853de661ce476281170ec90306b944df2234d9/CMakeLists.txt#L132

xkszltl:
If this works, make sure to test against 2.2 as well before you merge.

jay-zhuang:
@xkszltl based on my test, seems lower-case ones works for both gflags 2.1 and 2.2. So I'm using lower-case ones in #8324, not sure how it works, I just assume it's for backward compatibility.

btw. in 2.2. `gflags_TARGET` is the same as `gflags_LIBRARIES`, should we just just `gflags_LIBRARIES`?
https://github.com/gflags/gflags/blob/0b7f8db2c6b1b0b2451da0923a9ab09cc610e8d1/cmake/config.cmake.in#L163

xkszltl:
The way you change, i.e. a combination of `GFLAGS_TARGET` and `gflags_LIBRARIES`, is correct, except you missed the comment:
image

> btw. in 2.2. gflags_TARGET is the same as gflags_LIBRARIES, should we just use gflags_LIBRARIES?

No, `GFLAGS_TARGET` is the same as `GFLAGS_LIBRARIES`.
The lower case version doesn't exist.

Regarding TARGET vs. LIBRARIES, as you can tell from the name, it's one item vs. multiple.
LIBRARIES can be a list of files, or a list of targets.
A single target can carry more information, e.g. target properties, and it's the preferred way of modern cmake.

PR #6774, opened 2020-04-29
Expose the set of live blob files from Version/VersionSet

ltamasi:
Summary:
The patch adds logic that returns the set of live blob files from
`Version::AddLiveFiles` and `VersionSet::AddLiveFiles` (in addition to
live table files), and also cleans up the code a bit, for example, by
exposing only the numbers of table files as opposed to the earlier
`FileDescriptor`s that no clients used. Moreover, the patch extends
the `GetLiveFiles` API so that it also exposes blob files in the current version.
Similarly to https://github.com/facebook/rocksdb/pull/6755,
this is a building block for identifying and purging obsolete blob files.

Test Plan:
`make check`

### DIFF
--- a/db/db_filesnapshot.cc
+++ b/db/db_filesnapshot.cc
@@ -130,7 +130,7 @@ Status DBImpl::GetLiveFiles(std::vector<std::string>& ret,
 
   ret.clear();
   ret.reserve(live_table_files.size() + live_blob_files.size() +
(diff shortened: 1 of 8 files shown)


ltamasi:
Having technical issues with importing this, will try and create another PR.

ltamasi:
Closing in favor of https://github.com/facebook/rocksdb/pull/6785.

PR #6775, opened 2020-04-29
Make users explicitly be aware of prepare before commit

cheng-chang:
In current commit protocol of pessimistic transaction, if the transaction is not prepared before commit, the commit protocol implicitly assumes that the user wants to commit without prepare.

This PR adds TransactionOptions::skip_prepare, the default value is `true` because if set to `false`, all existing users who commit without prepare need to update their code to set skip_prepare to true. Although this does not force the user to explicitly express their intention of skip_prepare, it at least lets the user be aware of the assumption of being able to commit without prepare.

Test Plan:
added a new unit test TransactionTest::CommitWithoutPrepare

### DIFF
--- a/include/rocksdb/status.h
+++ b/include/rocksdb/status.h
@@ -60,6 +60,7 @@ class Status {
     kTryAgain = 13,
     kCompactionTooLarge = 14,
     kColumnFamilyDropped = 15,
+    kInvalidPrecondition = 16,
     kMaxCode
   };
 
@@ -79,6 +80,7 @@ class Status {
     KMergeOperandsInsufficientCapacity = 10,
     kManualCompactionPaused = 11,
     kOverwritten = 12,
+    kTxnNotPrepared = 13,
     kMaxSubCode
   };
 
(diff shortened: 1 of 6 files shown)


include/rocksdb/status.h
```diff
@@ -60,6 +60,7 @@ class Status {
     kTryAgain = 13,
     kCompactionTooLarge = 14,
     kColumnFamilyDropped = 15,
+    kInvalidPrecondition = 16,
```
lth, line 63: I don't know if there are any general rules in rocksdb on adding new error codes vs just adding subcodes (since this seems similar to kInvalidArgument). Maybe someone on the rocksdb team can comment.

We also use invalid argument in other places when transaction state isn't correct (eg. look at `PessimisticTransaction::Prepare` or `PessimisticTransaction::SetName`)

cheng-chang: I replaced usage of InvalidPrecondition with InvalidArgument

cheng-chang:
@lth PTAL

PR #6785, opened 2020-04-30
Expose the set of live blob files from Version/VersionSet

ltamasi:
Summary:
The patch adds logic that returns the set of live blob files from
`Version::AddLiveFiles` and `VersionSet::AddLiveFiles` (in addition to
live table files), and also cleans up the code a bit, for example, by
exposing only the numbers of table files as opposed to the earlier
`FileDescriptor`s that no clients used. Moreover, the patch extends
the `GetLiveFiles` API so that it also exposes blob files in the current version.
Similarly to https://github.com/facebook/rocksdb/pull/6755,
this is a building block for identifying and purging obsolete blob files.

Test Plan:
`make check`

### DIFF
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -924,6 +924,8 @@ class DBImpl : public DB {
   void TEST_GetFilesMetaData(ColumnFamilyHandle* column_family,
                              std::vector<std::vector<FileMetaData>>* metadata);
 
+  InstrumentedMutex* TEST_GetMutex() const { return &mutex_; }
+
   void TEST_LockMutex();
 
   void TEST_UnlockMutex();
@@ -969,6 +971,8 @@ class DBImpl : public DB {
   void TEST_WaitForPersistStatsRun(std::function<void()> callback) const;
   bool TEST_IsPersistentStatsEnabled() const;
   size_t TEST_EstimateInMemoryStatsHistorySize() const;
+
+  VersionSet* TEST_GetVersionSet() const { return versions_.get(); }
 #endif  // NDEBUG
 
  protected:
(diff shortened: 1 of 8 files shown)


db/version_set.h
```diff
@@ -629,8 +629,10 @@ class Version {
   // and return true. Otherwise, return false.
   bool Unref();
 
-  // Add all files listed in the current version to *live.
-  void AddLiveFiles(std::vector<FileDescriptor>* live);
+  // Add all files listed in the current version to *live_table_files and
+  // *live_blob_files.
+  void AddLiveFiles(std::vector<uint64_t>* live_table_files,
```
riversand963, line 634: I am a little concerned about returning only file numbers. If we should need more information in the future, we would have to change the signature again.

db/db_impl/db_impl.h
```diff
@@ -924,6 +924,8 @@ class DBImpl : public DB {
   void TEST_GetFilesMetaData(ColumnFamilyHandle* column_family,
                              std::vector<std::vector<FileMetaData>>* metadata);
 
+  InstrumentedMutex* TEST_GetMutex() const { return &mutex_; }
```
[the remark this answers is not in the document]
ltamasi: Thanks! Wasn't aware of that.

db/version_set.h
ltamasi: My thought process was that we seem to go to great lengths to make `AddLiveFiles` efficient and release the mutex as soon as possible (e.g. deduplication is done by the caller), yet miss the obvious low-hanging fruit of only copying the data that is actually needed by clients. We can keep it as-is if you see a use for `FileDescriptor`s though; let me know.

ltamasi: Matter of taste I guess but I also like the consistency of using the same type. :)

riversand963: SGTM.

PR #6787, opened 2020-04-30
Fix multiple CF replay failure in db_bench replay

zhichao-cao:
The multiple CF hash map is not passed to the multi-thread worker. When using multi-thread replay for multiple CFs, it will cause segment fault. Pass the cf_map to the argument.

Test plan: pass trace replay test.


yhchiang:
LGTM!

PR #6790, opened 2020-05-01
Mark dependencies as PRIVATE and fix missing dependencies in tools.

xkszltl:
Tools were mistakenly using leaked (PUBLIC) dependencies from main lib.


xkszltl:
This is for https://github.com/facebook/rocksdb/issues/6179
@tchaikov Please help review, thanks!

PR #6795, opened 2020-05-02
Fix issues for reproducing synthetic ZippyDB workloads in the FAST20' paper

celery1124:
Fix issues for reproducing synthetic ZippyDB workloads in the FAST20' paper using db_bench. Details changes as follows.
1, add a separate random mode in MixGraph to produce all_random workload.
2, fix power inverse function for generating prefix_dist workload.
3, make sure key_offset in prefix mode is always unsigned.
note: Need to carefully choose key_dist_a/b to avoid aliasing. Power inverse function range should be close to overall key space.

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -5418,7 +5418,8 @@ class Benchmark {
       if (key_dist_a == 0.0 && key_dist_b == 0.0) {
         key_offset = ini_rand % keyrange_size_;
       } else {
-        double u = static_cast<double>(ini_rand % keyrange_size_) / keyrange_size_;
+        double u =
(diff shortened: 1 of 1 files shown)


tools/db_bench_tool.cc
```diff
@@ -5418,10 +5418,12 @@ class Benchmark {
       if (key_dist_a == 0.0 && key_dist_b == 0.0) {
```
zhichao-cao, line 5418: It would be safe to change it to if (key_dist_a == 0.0 || key_dist_b == 0.0), to avoid x/0.

celery1124: Yeah, you are right! I'll add another commit.

PR #6798, opened 2020-05-04
status check enforcement

ajkr:
Tried making Status object enforce that it is checked in some way. In cases it is not checked, `PermitUncheckedError()` must be called explicitly.

Added a way to run tests (`ASSERT_STATUS_CHECKED=1 make -j48 check`) on a
whitelist. The effort appears significant to get each test to pass with
this assertion, so I only fixed up enough to get one test (`options_test`)
working and added it to the whitelist.

### DIFF
--- a/include/rocksdb/io_status.h
+++ b/include/rocksdb/io_status.h
@@ -170,6 +170,9 @@ inline IOStatus::IOStatus(Code _code, SubCode _subcode, const Slice& msg,
 }
 
 inline IOStatus::IOStatus(const IOStatus& s) : Status(s.code_, s.subcode_) {
+#ifdef ROCKSDB_ASSERT_STATUS_CHECKED
+  s.checked_ = true;
+#endif  // ROCKSDB_ASSERT_STATUS_CHECKED
   retryable_ = s.retryable_;
   data_loss_ = s.data_loss_;
   scope_ = s.scope_;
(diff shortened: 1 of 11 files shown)


include/rocksdb/io_status.h
```diff
@@ -204,14 +211,18 @@ inline IOStatus& IOStatus::operator=(IOStatus&& s)
...
+#endif  // ROCKSDB_ASSERT_STATUS_CHECKED
     code_ = std::move(s.code_);
     s.code_ = kOk;
     subcode_ = std::move(s.subcode_);
     s.subcode_ = kNone;
     retryable_ = s.retryable_;
     data_loss_ = s.data_loss_;
     scope_ = s.scope_;
-    scope_ = kIOErrorScopeFileSystem;
+    s.scope_ = kIOErrorScopeFileSystem;
```
pdillinger, line 225: Nice catch

include/rocksdb/io_status.h
```diff
@@ -204,14 +211,18 @@ inline IOStatus& IOStatus::operator=(IOStatus&& s)
 #endif
 {
   if (this != &s) {
+#ifdef ROCKSDB_ASSERT_STATUS_CHECKED
+    s.checked_ = true;
```
pdillinger, line 215: Why does Status ctor set to false but set to true after move?

pdillinger: OK, that's obvious. Nevermind.

ajkr:
Thanks for the review!

PR #6801, opened 2020-05-04
Do not swallow error returned from SaveTo()

riversand963:
Summary:
With consistency check enabled, VersionBuilder::SaveTo() may return error once
corruption is detected while building versions. We should handle these errors.

Test Plan:
(COMPILE_WITH_ASAN=1) make check

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -10,7 +10,6 @@
 * Fix a bug caused by using wrong compare function when sorting the input keys of MultiGet with timestamps.
 * Upgraded version of bzip library (1.0.6 -> 1.0.8) used with RocksJava to address potential vulnerabilities if an attacker can manipulate compressed data saved and loaded by RocksDB (not normal). See issue #6703.
 * Fix consistency checking error swallowing in some cases when options.force_consistency_checks = true.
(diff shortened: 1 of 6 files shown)


riversand963:
> Wow, there are so many missing cases. I don't always understand the consequence of destroying Version and return early. So what would build more confidence is to back each of the code path with a unit test. I assume it's not yet done?
> Or maybe another pair of eyes will help to gain confidence. Not sure whether @ajkr is interested in helping.

Thanks @siying for the review.
hmm, good point. I chose not to delete version and return early in this PR. In `VersionEditHandler` used by best-efforts recovery, error is stored in `status_` and returned to caller. In `ReactiveVersionSet`, we delete the version and break out of the loop without returning. I added a unit test for each of the following cases:

1. Detected inconsistency (injected) during best-efforts recovery
2. Detected inconsistency (injected) during recovery of secondary instance
3. Detected inconsistency (injected) during secondary trying to catch up with primary

In these cases, the code path going throgh `delete version; break or return;` are tested.
Is there another case you have in mind?

riversand963:
The failures are not related.

PR #6803, opened 2020-05-04
Fix failure to write output in SpecialEnv::GetCurrentTime

pdillinger:
Summary: This very old test code bug was causing a new valgrind failure
in MultiGetDeadlineExceeded

Also fix hang in MultiGetDeadlineExceeded by unifying with some logic from another test.

Test Plan: run that unit test under valgrind, make check


pdillinger:
Need at least one unit test fix

pdillinger:
> Thanks for fixing this! Could this explain the test hanging on Windows?

Maybe, but this gets uglier. A bunch of tests depend on addon_time_ being seconds, and have that returned through GetCurrentTime, and a bunch depend on being able to accumulate addon_time_ in microseconds via SleepForMicroseconds and NowMicros().

pdillinger:
> Maybe, but this gets uglier. A bunch of tests depend on addon_time_ being seconds, and have that returned through GetCurrentTime, and a bunch depend on being able to accumulate addon_time_ in microseconds via SleepForMicroseconds and NowMicros().

That's a bigger fix than I want to tackle in this PR to fix valgrind.

pdillinger:
> Could this explain the test hanging on Windows?

Still timed out on MacOS. :(

PR #6807, opened 2020-05-05
Find/purge obsolete blob files

ltamasi:
Summary:
The patch extends `FindObsoleteFiles` and `PurgeObsoleteFiles` with
support for blob files. The behavior is analogous to SST files: obsolete
blob files are put on the "candidates for deletion" list, while live (and pending)
files are preserved.

Test Plan:
`make check`


ltamasi:
Note that this cleanup logic deals with blob files that are part of `Version` (and handled by the RocksDB core); blob files handled by BlobDB proper are currently stored in a separate directory (by default `blob_dir`) and are thus unaffected.

riversand963:
You may want to rebase and format the code.

PR #6810, opened 2020-05-05
Disallow using the base DB's storage directory as blob_dir in BlobDB

ltamasi:
Summary:
https://github.com/facebook/rocksdb/pull/6807 extends the logic that
identifies and purges obsolete files to blob files handled by RocksDB
itself. In order to prevent that from interfering with the current BlobDB code,
we need to make sure that `BlobDBOptions::blob_dir` is different from
the storage directories used by the base DB. (Note: this is true by default.)
The patch adds a check that explicitly disallows this configuration and
returns `Status::NotSupported` from `BlobDB::Open` in such cases.

Test Plan:
Tested using the BlobDB mode of `db_bench`.

### DIFF
--- a/utilities/blob_db/blob_db_impl.cc
+++ b/utilities/blob_db/blob_db_impl.cc
@@ -211,7 +211,7 @@ Status BlobDBImpl::Open(std::vector<ColumnFamilyHandle*>* handles) {
   }
   db_impl_ = static_cast_with_check<DBImpl>(db_->GetRootDB());
 
-  // Sanitize the blob_dir provided. Using the same directory where the
+  // Sanitize the blob_dir provided. Using a directory where the
   // base DB stores its files for the default CF is not supported.
   const ColumnFamilyData* const cfd =
       static_cast<ColumnFamilyHandleImpl*>(DefaultColumnFamily())->cfd();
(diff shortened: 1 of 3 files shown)


utilities/blob_db/blob_db_impl.cc
```diff
@@ -211,6 +211,33 @@ Status BlobDBImpl::Open(std::vector<ColumnFamilyHandle*>* handles) {
...
+  // base DB stores its files for the default CF is not supported.
+  const ColumnFamilyData* const cfd =
+      static_cast<ColumnFamilyHandleImpl*>(DefaultColumnFamily())->cfd();
+  assert(cfd);
+
+  const ImmutableCFOptions* const ioptions = cfd->ioptions();
+  assert(ioptions);
+
+  for (const auto& cf_path : ioptions->cf_paths) {
+    assert(env_);
```
riversand963, line 224: Nit: asserting once suffices.

ltamasi:
> Thanks @ltamasi for the PR. Is this contract mentioned in any doc/code comment?

Good point, thanks @riversand963 ! It is sort-of implicitly there in the current description of the `blob_dir` option (it refers to `blob_dir` being a directory *under* the DB dir); let me make it clearer/more explicit.

PR #6815, opened 2020-05-06
Enable IO Uring in MultiGet in direct IO mode

cheng-chang:
Currently, in direct IO mode, `MultiGet` retrieves the data blocks one by one instead of in parallel, see `BlockBasedTable::RetrieveMultipleBlocks`.

Since direct IO is supported in `RandomAccessFileReader::MultiRead` in https://github.com/facebook/rocksdb/pull/6446, this PR applies `MultiRead` to `MultiGet` so that the data blocks can be retrieved in parallel.

Also, in direct IO mode and when data blocks are compressed and need to uncompressed, this PR only allocates one continuous aligned buffer to hold the data blocks, and then directly uncompress the blocks to insert into block cache, there is no longer intermediate copies to scratch buffers.

Test Plan:
1. added a new unit test `BlockBasedTableReaderTest::MultiGet`.
2. existing unit tests and stress tests  contain tests against `MultiGet` in direct IO mode.


table/block_based/block_based_table_reader_test.cc
```diff
@@ -0,0 +1,216 @@
...
+    ASSERT_OK(f->Append(content, IOOptions(), nullptr));
+    ASSERT_OK(f->Close(IOOptions(), nullptr));
+  }
+
+  void NewFileWriter(const std::string& filename,
+                     std::unique_ptr<WritableFileWriter>* writer) {
+    std::string path = Path(filename);
+    EnvOptions env_options;
+    std::unique_ptr<WritableFile> file;
+    ASSERT_OK(env_->NewWritableFile(path, &file, env_options));
```
anand1976, line 108: Can we directly call ```fs_->NewWritableFile```?

table/block_based/block_based_table_reader_test.cc
```diff
@@ -0,0 +1,216 @@
...
+    ASSERT_TRUE(table->TEST_KeyInCache(ReadOptions(), keys[i]));
+    ASSERT_EQ(values[i].ToString(), kv[keys[i].ToString()]);
+  }
+}
+
+// Param 1: compression type
+// Param 2: whether to use direct reads
+INSTANTIATE_TEST_CASE_P(
+    MultiGet, BlockBasedTableReaderTest,
+    ::testing::Combine(::testing::ValuesIn(GetSupportedCompressions()),
```
anand1976, line 207: Do we need to do something for the case where compression is enabled, but the block is not compressible?

table/block_based/block_based_table_reader.cc
```diff
@@ -1633,6 +1637,9 @@ void BlockBasedTable::RetrieveMultipleBlocks(
     // If current block is adjacent to the previous one, at the same time,
     // compression is enabled and there is no compressed cache, we combine
     // the two block read as one.
+    // We don't combine block reads here in direct IO mode, because when doing
+    // direct IO read, the block requests will be realigned and merged when
+    // necessary.
     if (scratch != nullptr && prev_end == handle.offset()) {
```
anand1976, line 1643: Maybe its just me, but I think using ```use_shared_buffer``` here and below and then differentiating using ```file->use_direct_io()``` might be easier to reason about, instead of checking ```scratch```.

table/block_based/block_based_table_reader_test.cc
cheng-chang: yes

table/block_based/block_based_table_reader.cc
cheng-chang: make sense. The check on `scratch` contains implicit assumption that if `use_direct_io`, then `scratch` will be set to `nullptr` in `MultiGet`.

table/block_based/block_based_table_reader_test.cc
cheng-chang: `GetSupportedCompressions` can return `kNoCompression`

PR #6817, opened 2020-05-06
Add Support for saving CompressionOptions to Options File

mrambacher:
This PR does a few things:
- The "compression_opts" and "bottom_compression_opts" can now be read/written as name/value pairs of options (instead of only a colon-separated list;
- These options can now be read/written to the Options file;
- The parallel_threads value can now be set (either in the colon or name-value format).

The compression options are now stored and treated as a OptionTypeInfo::Struct by the options system, meaning they can be read and written like the other structs.  This change allows them to be read/written easily to the options file.

Additionally, the colon-format was extended to allow support for setting parallel threads.  Tests were added to test all of the option settings via the optional parameters in the colon format.

### DIFF
--- a/options/cf_options.cc
+++ b/options/cf_options.cc
@@ -94,9 +94,29 @@ static Status ParseCompressionOptions(const std::string& value,
         ParseInt(value.substr(start, value.size() - start));
     end = value.find(':', start);
   }
-  // parallel_threads is not serialized with this format.
-  // We plan to upgrade the format to a JSON-like format.
-  compression_opts.parallel_threads = CompressionOptions().parallel_threads;
+
+  // parallel_threads is optional for backwards compatibility
+  if (end != std::string::npos) {
+    start = end + 1;
+    if (start >= value.size()) {
+      return Status::InvalidArgument(
+          "unable to parse the specified CF option " + name);
+    }
+    // Since parallel_threads comes before enabled but was added optionally
+    // later, we need to check if this is the final token (meaning it is the
+    // enabled bit), or if there is another token (meaning this one is
+    // parallel_threads)
+    end = value.find(':', start);
+    if (end != std::string::npos) {
+      compression_opts.parallel_threads =
+          ParseInt(value.substr(start, value.size() - start));
+    } else {
+      // parallel_threads is not serialized with this format, but enabled is
+      compression_opts.parallel_threads = CompressionOptions().parallel_threads;
+      compression_opts.enabled =
+          ParseBoolean("", value.substr(start, value.size() - start));
+    }
+  }
 
   // enabled is optional for backwards compatibility
   if (end != std::string::npos) {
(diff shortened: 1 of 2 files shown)


zhichao-cao:
@mrambacher Thanks for the fix. We will hold it until 6424, 6425 are merged.

zhichao-cao:
Thanks for working on the CompressionOption part the dealing with the special case! Looks good to me. Just some minor comments.

options/options_test.cc
```diff
@@ -553,6 +553,115 @@ TEST_F(OptionsTest, GetColumnFamilyOptionsFromStringTest) {
...
+  ASSERT_EQ(base_cf_opt.bottommost_compression_opts.zstd_max_train_bytes,
+            dflt.zstd_max_train_bytes);
+  ASSERT_EQ(base_cf_opt.bottommost_compression_opts.parallel_threads,
+            dflt.parallel_threads);
+  ASSERT_EQ(base_cf_opt.bottommost_compression_opts.enabled, dflt.enabled);
+
+  ASSERT_OK(GetColumnFamilyOptionsFromString(
+      config_options, ColumnFamilyOptions(),
+      "compression_opts=4:5:6:7:8:9:true; "
+      "bottommost_compression_opts=5:6:7:8:9:false",
```
zhichao-cao, line 591: Can you add one or two test case that we use ParseStruct instead of the legacy ParseCompressionOptions to handle the pass-in string?

zhichao-cao:
@siying You may take a look at this PR if it is as you expected.

options/options_test.cc
mrambacher: I pushed some new test values in that test.  I can rebase later as required

zhichao-cao:
> It looks good to me. Maybe @ajkr has some comments. But I would defer to @zhichao-cao to approve.

Thanks for reviewing it. @ajkr you may take a look before I approve it?

options/cf_options.cc
```diff
@@ -94,9 +94,29 @@ static Status ParseCompressionOptions(const std::string& value,
...
+          ParseInt(value.substr(start, value.size() - start));
+    } else {
+      // parallel_threads is not serialized with this format, but enabled is
+      compression_opts.parallel_threads = CompressionOptions().parallel_threads;
+      compression_opts.enabled =
+          ParseBoolean("", value.substr(start, value.size() - start));
+    }
+  }
 
   // enabled is optional for backwards compatibility
```
ajkr, line 121: Maybe we should document somewhere that `enabled` is actually required if the user tries to provide `parallel_threads`. It seems a bit difficult to figure out right now.

zhichao-cao: > Maybe we should document somewhere that `enabled` is actually required if the user tries to provide `parallel_threads`. It seems a bit difficult to figure out right now.

You mean, explain it in the HISTORY.md?

options/options_test.cc
```diff
@@ -553,6 +553,172 @@ TEST_F(OptionsTest, GetColumnFamilyOptionsFromStringTest) {
...
+  ASSERT_EQ(base_cf_opt.bottommost_compression_opts.max_dict_bytes, 7u);
+  ASSERT_EQ(base_cf_opt.bottommost_compression_opts.zstd_max_train_bytes,
+            dflt.zstd_max_train_bytes);
+  ASSERT_EQ(base_cf_opt.bottommost_compression_opts.parallel_threads,
+            dflt.parallel_threads);
+  ASSERT_EQ(base_cf_opt.bottommost_compression_opts.enabled, dflt.enabled);
+
+  ASSERT_OK(GetColumnFamilyOptionsFromString(
+      config_options, ColumnFamilyOptions(),
+      "compression_opts=4:5:6:7:8:9:true; "
```
ajkr, line 590: I wonder whether "compression_opts=4:5:6:7:8:9:2" works, as that seems the most likely mistake (trying to specify 2 for `parallel_threads` but actually setting `enabled`).

mrambacher: That actually would not work, since "2" is not a valid boolean.  That case would throw an exception and return an error.  

As it is written, if you want to specify parallel_threads, you must also specify enabled.  The reason for this is that enabled was added in an earlier release.

PR #6819, opened 2020-05-06
Add stress test for best-efforts recovery

riversand963:
Summary:
Add crash test for the case of best-efforts recovery.
After a certain amount of time, we kill the db_stress process, randomly delete some certain table files and restart db_stress. Given the randomness of file deletion, it is difficult to verify against a reference for data correctness. Therefore, we just check that the db can restart successfully.

Test Plan:
```
./db_stress -best_efforts_recovery=true -disable_wal=1 -reopen=0
./db_stress -best_efforts_recovery=true -disable_wal=0 -skip_verifydb=1 -verify_db_one_in=0 -continuous_verification_interval=0
make crash_test_with_best_efforts_recovery
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -790,7 +790,8 @@ endif  # PLATFORM_SHARED_EXT
 	dbg rocksdbjavastatic rocksdbjava install install-static install-shared uninstall \
 	analyze tools tools_lib \
 	blackbox_crash_test_with_atomic_flush whitebox_crash_test_with_atomic_flush  \
-	blackbox_crash_test_with_txn whitebox_crash_test_with_txn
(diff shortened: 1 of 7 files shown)


riversand963:
> @riversand963 A couple of questions -
> 
> 1. Do you plan to have a separate set of regression tests for best efforts recovery?

Yes. Reusing existing test jobs will make them run longer, which is not desirable. Therefore, I prefer to add more jobs that can run in parallel.

> 2. Can we keep a copy of the original DB and atleast verify the keys that are present after recovery? Basically what guarantee does best efforts recovery provide to a user about keys that exist post-recovery and how do we validate that.

I think it can be hard as long as we randomly delete files before restart. Keys deleted can re-appear after deleting some files and restart.

PR #6821, opened 2020-05-07
Fix false NotFound from batched MultiGet with kHashSearch

pdillinger:
Summary: The error is assigning KeyContext::s to NotFound status in a
table reader for a "not found in this table" case, which skips searching
in later tables, like only a delete should. (The hash search index iterator
is the only one that can return status NotFound even if Valid() == false.)

This was detected by intermittent failure in
MultiThreadedDBTest.MultiThreaded/5, a kHashSearch configuration.

Test Plan: modified existing unit test to reproduce problem

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -2461,13 +2461,16 @@ void BlockBasedTable::MultiGet(const ReadOptions& read_options,
                               ExtractUserKey(v.first_internal_key)) < 0)) {
           // The requested key falls between highest key in previous block and
           // lowest key in current block.
-          *(miter->s) = iiter->status();
+          if (!iiter->status().IsNotFound()) {
+            *(miter->s) = iiter->status();
+          }
           data_block_range.SkipKey(miter);
           sst_file_range.SkipKey(miter);
           continue;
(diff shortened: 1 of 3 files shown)


table/block_based/block_based_table_reader.cc
```diff
@@ -2680,7 +2683,7 @@ void BlockBasedTable::MultiGet(const ReadOptions& read_options,
         PERF_COUNTER_BY_LEVEL_ADD(bloom_filter_full_true_positive, 1,
                                   rep_->level);
       }
-      if (s.ok()) {
+      if (s.ok() && !iiter->status().IsNotFound()) {
```
anand1976, line 2686: ```iiter``` may not necessarily have been initialized for this key and might have a stale status. The ```iiter``` status should only be considered if ```!first_block```.

pdillinger: You're suggesting I try to fix a bug in which a failure status (not OK, not NotFound) could falsely be attached to too many keys in a MultiGet batch? That just seems like a case where "best effort" could be better, whereas this PR is about fixing an incorrect result. I'd prefer not to mix the two.

PR #6826, opened 2020-05-08
Allow MultiGet users to limit cumulative value size

akankshamahajan15:
Summary: 1. Add a value_size in read options which limits the cumulative value size of keys read in batches. Once the size exceeds read_options.value_size, all the remaining keys are returned with status Abort without further fetching any key.
2. Add a unit test case MultiGetBatchedValueSizeSimple the reads keys from memory and sst files.

Test Plan: 1. make check -j64
	   2. Add a new unit test case

Reviewers:

Subscribers:

Tasks: T64971827

Tags:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -19,6 +19,7 @@
 ### New Feature
 * sst_dump to add a new --readahead_size argument. Users can specify read size when scanning the data. Sst_dump also tries to prefetch tail part of the SST files so usually some number of I/Os are saved there too.
 * Generate file checksum in SstFileWriter if Options.file_checksum_gen_factory is set. The checksum and checksum function name are stored in ExternalSstFileInfo after the sst file write is finished.
+* Add a value_size_soft_limit in read options which limits the cumulative value size of keys read in batches in MultiGet. Once the cumulative value size of found keys exceeds read_options.value_size_soft_limit, all the remaining keys are returned with status Abort without further finding their values. By default the value_size_soft_limit is std::numeric_limits<uint64_t>::max().
 
 ## 6.10 (5/2/2020)
 ### Bug Fixes
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -1501,6 +1501,221 @@ TEST_F(DBBasicTest, MultiGetBatchedMultiLevelMerge) {
   }
 }
 
+TEST_F(DBBasicTest, MultiGetBatchedValueSizeInMemory) {
+  CreateAndReopenWithCF({"pikachu"}, CurrentOptions());
+  SetPerfLevel(kEnableCount);
+  ASSERT_OK(Put(1, "k1", "v_1"));
+  ASSERT_OK(Put(1, "k2", "v_2"));
+  ASSERT_OK(Put(1, "k3", "v_3"));
+  ASSERT_OK(Put(1, "k4", "v_4"));
+  ASSERT_OK(Put(1, "k5", "v_5"));
+  ASSERT_OK(Put(1, "k6", "v_6"));
+  std::vector<Slice> keys = {"k1", "k2", "k3", "k4", "k5", "k6"};
+  std::vector<PinnableSlice> values(keys.size());
+  std::vector<Status> s(keys.size());
+  std::vector<ColumnFamilyHandle*> cfs(keys.size(), handles_[1]);
+
+  get_perf_context()->Reset();
+  ReadOptions ro;
+  ro.value_size_soft_limit = 11;
+  db_->MultiGet(ro, handles_[1], keys.size(), keys.data(), values.data(),
+                s.data(), false);
+
+  ASSERT_EQ(values.size(), keys.size());
+  for (unsigned int i = 0; i < 4; i++) {
+    ASSERT_EQ(std::string(values[i].data(), values[i].size()),
+              "v_" + std::to_string(i + 1));
+  }
+
+  for (unsigned int i = 4; i < 6; i++) {
+    ASSERT_TRUE(s[i].IsAborted());
+  }
+
+  ASSERT_EQ(12, (int)get_perf_context()->multiget_read_bytes);
+  SetPerfLevel(kDisable);
+}
+
+TEST_F(DBBasicTest, MultiGetBatchedValueSize) {
+  do {
+    CreateAndReopenWithCF({"pikachu"}, CurrentOptions());
+    SetPerfLevel(kEnableCount);
+
+    ASSERT_OK(Put(1, "k6", "v6"));
+    ASSERT_OK(Put(1, "k7", "v7_"));
+    ASSERT_OK(Put(1, "k3", "v3_"));
+    ASSERT_OK(Put(1, "k4", "v4"));
+    Flush(1);
+    ASSERT_OK(Delete(1, "k4"));
+    ASSERT_OK(Put(1, "k11", "v11"));
+    ASSERT_OK(Delete(1, "no_key"));
+    ASSERT_OK(Put(1, "k8", "v8_"));
+    ASSERT_OK(Put(1, "k13", "v13"));
+    ASSERT_OK(Put(1, "k14", "v14"));
+    ASSERT_OK(Put(1, "k15", "v15"));
+    ASSERT_OK(Put(1, "k16", "v16"));
+    ASSERT_OK(Put(1, "k17", "v17"));
+    Flush(1);
+
+    ASSERT_OK(Put(1, "k1", "v1_"));
+    ASSERT_OK(Put(1, "k2", "v2_"));
+    ASSERT_OK(Put(1, "k5", "v5_"));
+    ASSERT_OK(Put(1, "k9", "v9_"));
+    ASSERT_OK(Put(1, "k10", "v10"));
+    ASSERT_OK(Delete(1, "k2"));
+    ASSERT_OK(Delete(1, "k6"));
+
+    get_perf_context()->Reset();
+
+    std::vector<Slice> keys({"k1", "k10", "k11", "k12", "k13", "k14", "k15",
+                             "k16", "k17", "k2", "k3", "k4", "k5", "k6", "k7",
+                             "k8", "k9", "no_key"});
+    std::vector<PinnableSlice> values(keys.size());
+    std::vector<ColumnFamilyHandle*> cfs(keys.size(), handles_[1]);
+    std::vector<Status> s(keys.size());
+
+    ReadOptions ro;
+    ro.value_size_soft_limit = 20;
+    db_->MultiGet(ro, handles_[1], keys.size(), keys.data(), values.data(),
+                  s.data(), false);
+
+    ASSERT_EQ(values.size(), keys.size());
+
+    // In memory keys
+    ASSERT_EQ(std::string(values[0].data(), values[0].size()), "v1_");
+    ASSERT_EQ(std::string(values[1].data(), values[1].size()), "v10");
+    ASSERT_TRUE(s[9].IsNotFound());  // k2
+    ASSERT_EQ(std::string(values[12].data(), values[12].size()), "v5_");
+    ASSERT_TRUE(s[13].IsNotFound());  // k6
+    ASSERT_EQ(std::string(values[16].data(), values[16].size()), "v9_");
+
+    // In sst files
+    ASSERT_EQ(std::string(values[2].data(), values[1].size()), "v11");
+    ASSERT_EQ(std::string(values[4].data(), values[4].size()), "v13");
+    ASSERT_EQ(std::string(values[5].data(), values[5].size()), "v14");
+
+    // Remaining aborted after value_size exceeds.
+    ASSERT_TRUE(s[3].IsAborted());
+    ASSERT_TRUE(s[6].IsAborted());
+    ASSERT_TRUE(s[7].IsAborted());
+    ASSERT_TRUE(s[8].IsAborted());
+    ASSERT_TRUE(s[10].IsAborted());
+    ASSERT_TRUE(s[11].IsAborted());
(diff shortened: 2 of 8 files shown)


include/rocksdb/options.h
```diff
@@ -1356,6 +1356,12 @@ struct ReadOptions {
   // processing a batch
   std::chrono::microseconds deadline;
 
+  // It limits the maximum cumulative value size of the keys in batch while
+  // reading through MultiGet. Once the cumulative value size exceeds this value
+  // then all the remaining keys are returned with status Aborted.
+  // Defualt: 0.
```
pdillinger, line 1362: If you keep 0 as a special value, this comment needs to mention that.

Although our internal checks would be more verbose, I suggest instead using std::numeric_limits<uint64_t>::max() so that the behavior is continuous across the range of settings. (Though this case isn't as weird as -1 for files_size_error_margin.)

Maybe a more descriptive name like "value_size_soft_limit"? I could foresee us adding a hard "value_size_limit" that would apply to Get and MultiGet. A soft limit (allowed to exceed but none after that) obviously doesn't make sense for Get.

akankshamahajan15: Thank you.  Yes it makes sense to be consistent with different range of settings. I will make the changes.

db/memtable.cc
```diff
@@ -929,6 +929,18 @@ void MemTable::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
 
     if (found_final_value) {
       iter->value->PinSelf();
+      if (read_options.value_size > 0) {
+        if (range->GetValueSize() + iter->value->size() >
+            read_options.value_size) {
+          while (iter != temp_range.end()) {
```
anand1976, line 935: ```range``` may not cover all keys in the MultiGet batch, just the ones that could be present in memtable. It would be better to move the logic to abort the request and set the status of remaining keys in ```MultiGetContext```, so it can set status for all remaining keys.

db/version_set.cc
```diff
@@ -2007,6 +2007,18 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
           PERF_COUNTER_BY_LEVEL_ADD(user_key_return_count, 1,
                                     fp.GetHitFileLevel());
           file_range.MarkKeyDone(iter);
+          if (read_options.value_size > 0) {
+            if (file_range.GetValueSize() + iter->value->size() >
+                read_options.value_size) {
+              while (iter != file_range.end()) {
```
anand1976, line 2013: Ditto

db/db_impl/db_impl.cc
```diff
@@ -2315,6 +2328,7 @@ Status DBImpl::MultiGetImpl(
       super_version->current->MultiGet(read_options, &range, callback,
                                        is_blob_index);
     }
+    curr_value_size = range.GetValueSize();
```
anand1976, line 2331: Should we bail here if ```value_size``` exceeds that specified by user, instead of continuing to iterate over the batch?

db/memtable.cc
akankshamahajan15: I'm using MultiGetContext::value_size to hold the value_size of keys processed it now.  VersionSet::MultiGet then checks the MultiGetContext::value_size  (updated by MemTable::MultiGet) and then sets the remaining keys status to Abort if size exceeds. So this way memtable keys and sst file keys both are processed.

But it makes sense to not even call VersionSet::MultiGet if size exceeds and MemTable::MultiGet should update the remaining keys. So i will change the logic to do so. Thanks.

db/memtable.cc
```diff
@@ -929,6 +929,16 @@ void MemTable::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
...
     if (found_final_value) {
       iter->value->PinSelf();
+      range->AddValueSize(iter->value->size());
+      if (range->GetValueSize() > read_options.value_size_soft_limit) {
+        // Set all remaining keys in range to Abort
+        for (auto range_iter = range->begin(); range_iter != range->end();
+             ++range_iter) {
+          range->MarkKeyDone(range_iter);
+          *(range_iter->s) = Status::Aborted();
+        }
```
akankshamahajan15, line 939: This code iterates over all remaining keys in range and set their status to Aborted. I added here because it was simple to add here by iterating whereas in MultiGetContext, I needed to change the index_ as well and code got complicated there.

pdillinger:
Looks good in relation to my earlier review comments. :) Deferring to Anand for final review.

db/db_impl/db_impl.cc
```diff
@@ -2315,6 +2326,21 @@ Status DBImpl::MultiGetImpl(
...
+    if (curr_value_size > read_options.value_size_soft_limit) {
+      break;
```
anand1976, line 2331: Set ```s``` to ```Status::Aborted()```? We have logic to abort the MultiGet for multiple column families around line 2096 that looks at the overall status. You just need to fix the assert in 2096 and 2255 as it currently only expects ```Status::TimedOut()```.

db/db_impl/db_impl.cc
```diff
@@ -2315,6 +2326,21 @@ Status DBImpl::MultiGetImpl(
...
+    }
+  }
+
+  // Set all remaining keys status to abort
+  if (keys_left && s != Status::TimedOut() &&
```
anand1976, line 2336: We can just combine this block with the one in lines 2357-2364 as the logic is the same.

db/version_set.cc
```diff
@@ -2006,6 +2006,17 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
...
                                     fp.GetHitFileLevel());
+          file_range.AddValueSize(iter->value->size());
+          if (file_range.GetValueSize() > read_options.value_size_soft_limit) {
+            // Set all remaining keys in range to Abort
+            for (auto range_iter = range->begin(); range_iter != range->end();
+                 ++range_iter) {
+              range->MarkKeyDone(range_iter);
+              *(range_iter->s) = Status::Aborted();
+            }
+            iter = file_range.end();
```
anand1976, line 2017: Can we set ```s = Status::Aborted()``` and use that to break out of the nested loops? Or maybe checking for ```file_range.empty()``` would work, like in line 2045.

db/version_set.cc
```diff
@@ -2060,6 +2074,17 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
           nullptr /* result_operand */, true);
       if (LIKELY(iter->value != nullptr)) {
         iter->value->PinSelf();
+        range->AddValueSize(iter->value->size());
+        if (range->GetValueSize() > read_options.value_size_soft_limit) {
```
anand1976, line 2078: Can track an overall status and combine this block and the one in lines 2010-2016 into one at the end of the function?

akankshamahajan15:
Addressed comments

db/version_set.cc
```diff
@@ -1960,7 +1962,8 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
       return;
     }
     uint64_t batch_size = 0;
-    for (auto iter = file_range.begin(); iter != file_range.end(); ++iter) {
+    for (auto iter = file_range.begin(); s.ok() && iter != file_range.end();
```
anand1976, line 1965: Nit: No need to check ```Status``` here since there is a ```break``` on line 2040.

db/memtable.cc
```diff
@@ -929,6 +929,16 @@ void MemTable::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
 
     if (found_final_value) {
       iter->value->PinSelf();
+      range->AddValueSize(iter->value->size());
+      if (range->GetValueSize() > read_options.value_size_soft_limit) {
```
anand1976, line 933: Should this block be after ```range->MarkKeyDone(iter)``` is called? Otherwise, we'll discard the value for this key.

db/version_set.cc
```diff
@@ -2060,12 +2068,23 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
           nullptr /* result_operand */, true);
       if (LIKELY(iter->value != nullptr)) {
         iter->value->PinSelf();
+        range->AddValueSize(iter->value->size());
+        if (range->GetValueSize() > read_options.value_size_soft_limit) {
```
anand1976, line 2072: Should this be after calling ```range->MarkKeyDone(iter)```?

db/db_basic_test.cc
```diff
@@ -1501,6 +1501,81 @@ TEST_F(DBBasicTest, MultiGetBatchedMultiLevelMerge) {
...
+    ASSERT_TRUE(s[6].IsAborted());
+    ASSERT_TRUE(s[7].IsAborted());
+    ASSERT_TRUE(s[8].IsAborted());
+    ASSERT_TRUE(s[10].IsAborted());
+    ASSERT_TRUE(s[11].IsAborted());
+    ASSERT_TRUE(s[14].IsAborted());
+    ASSERT_TRUE(s[15].IsAborted());
+    ASSERT_TRUE(s[17].IsAborted());
+
+    // 6 kv pairs * 3 bytes per value (i.e. 18)
```
anand1976, line 1573: I suggest adding another call to ```MultiGet``` with just memtable keys to test the abort logic in ```MemTable::MultiGet```. Also, I suggest adding a few merge values.

db/memtable.cc
akankshamahajan15: I am not marking the key done whose size exceeds the value_size here so that its status can be set to Abort  in the iteration of marking remaining keys to Aborted. But its value is still added to the range->AddValueSize.

Logic:  So for eg if 4th keys value exceeds the value_size limit then its value is still added  to make the range->GetValueSize() exceeds but its status marked Abort.  This way in DBImpl::MultiGet, I know that value_size_limit has already exceeded as range->GetValueSize() > read_options.value_size_soft_limit. If I don't exceed the range->value_size_ then I can't exit from the loop in the DBMultiGet::MultiGet as I don't know if limit has exceeded or not.

db/version_set.cc
akankshamahajan15: Same as above

db/version_set.cc
akankshamahajan15: The break on line 2040 exits the outer loop (for different sst files) and this one is for inner loop that iterates over keys in one sst file.

db/db_basic_test.cc
akankshamahajan15: Sure. I will add them. Thank you!

akankshamahajan15:
Added new testcases 1. One with Merge operation 2. Read all keys from memory with aborting few.

db/version_set.cc
anand1976: Oh that's right! Sorry, misread the code.

db/memtable.cc
anand1976: In that case, should the option name be change to ```value_size_limit```, since its a hard limit rather than a soft limit.

include/rocksdb/options.h
```diff
@@ -1356,6 +1356,13 @@ struct ReadOptions {
   // processing a batch
   std::chrono::microseconds deadline;
 
+  // It limits the maximum cumulative value size of the keys in batch while
+  // reading through MultiGet. Once the cumulative value size exceeds this
```
anand1976, line 1360: The comment can probably be reworded a little to match the behavior - "cumulative size is not allowed to exceed the limit".

anand1976: The comment is still a little confusing as it may be interpreted that the cumulative value size of successful keys can exceed ```value_size_hard_limit```. Maybe we can say "Once a key causes the cumulative size to exceed this hard limit, that key and all remaining keys are returned with status Aborted.".

db/version_set.cc
```diff
@@ -1933,13 +1933,15 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
       &storage_info_.file_indexer_, user_comparator(), internal_comparator());
   FdWithKeyRange* f = fp.GetNextFile();
 
+  Status s = Status::OK();
```
anand1976, line 1936: Nit: Default ```Status``` constructor initializes it to OK.

pdillinger:
Did we determine that hard limit was better for our internal customers than soft limit? (Soft limit seems more "efficient" in principle because when we have a k-v that crosses the limit, we've already done almost all of the work to look it up, and many applications will likely repeat the lookup in a subsequent call if we throw it away.)

pdillinger:
> Did we determine that hard limit was better for our internal customers than soft limit? (Soft limit seems more "efficient" in principle because when we have a k-v that crosses the limit, we've already done almost all of the work to look it up, and many applications will likely repeat the lookup in a subsequent call if we throw it away.)

Internal discussion landed on going with the soft limit.

akankshamahajan15:
Changed the implementation of value_size to soft-limit.

HISTORY.md
```diff
@@ -19,6 +19,7 @@
 ### New Feature
 * sst_dump to add a new --readahead_size argument. Users can specify read size when scanning the data. Sst_dump also tries to prefetch tail part of the SST files so usually some number of I/Os are saved there too.
 * Generate file checksum in SstFileWriter if Options.file_checksum_gen_factory is set. The checksum and checksum function name are stored in ExternalSstFileInfo after the sst file write is finished.
+* Add a value_size_hard_limit in read options which limits the cumulative value size of keys read in batches in MultiGet. Once the cumulative value size of found keys exceeds read_options.value_size_hard_limit, all the remaining keys are returned with status Abort without further finding their values. By default the value_size_hard_limit is std::numeric_limits<uint64_t>::max().
```
anand1976, line 22: Change ```value_size_hard_limit``` to ```value_size_soft_limit```.

PR #6828, opened 2020-05-08
Cherry picks into 6.9.fb (CI validation PR)

pdillinger:


pdillinger:
Also picking

    Do not swallow error returned from SaveTo() (6801)

with local "make check" only for now.

PR #6830, opened 2020-05-08
Make EncryptEnv inheritable

matthewvon:
EncryptEnv class is both declared and defined within env_encryption.cc.  This makes it really tough to derive new classes from that base.  

This branch moves declaration of the class to rocksdb/env_encryption.h.  The change facilitates making new encryption modules (such as an upcoming openssl AES CTR pull request) possible / easy.

The only coding change was to add the EncryptEnv object to env_basic_test.cc.


matthewvon:
mmaszewski@arango2:~/rocksdb.matthewv$ make check-format
$DEBUG_LEVEL is 1
Makefile:169: Warning: Compiling in debug mode. Don't use the resulting binary in production
  GEN      util/build_version.cc
$DEBUG_LEVEL is 1
Makefile:169: Warning: Compiling in debug mode. Don't use the resulting binary in production
  GEN      util/build_version.cc
build_tools/format-diff.sh -c
Nothing needs to be reformatted!

Not sure what to do now.  "make check-format" executes clean on my machine.

matthewvon:
and is there any opposition to adding clang-format-diff.py to .gitignore?

matthewvon:
Did a clean pull to a new directory.  "make check-format" still succeeds.

clang-format -version is:
  clang-format version 3.8.1-24 (tags/RELEASE_381/final)

matthewvon:
@ajkr This pull request is again ready for import.  It started failing a recently added check for LITE builds.  I fixed that issue.

ajkr:
> and is there any opposition to adding clang-format-diff.py to .gitignore?

Sounds good.

riversand963:
> > and is there any opposition to adding clang-format-diff.py to .gitignore?
> 
> Sounds good.

Seems it's already there: https://github.com/facebook/rocksdb/blob/master/.gitignore#L88.

include/rocksdb/env_encryption.h
```diff
@@ -194,13 +196,235 @@ class CTREncryptionProvider : public EncryptionProvider {
...
+  // file given
+  // given name and options. The given prefix is already decrypted.
   virtual Status CreateCipherStreamFromPrefix(
       const std::string& fname, const EnvOptions& options,
       uint64_t initialCounter, const Slice& iv, const Slice& prefix,
       std::unique_ptr<BlockAccessCipherStream>* result);
 };
 
+class EncryptedSequentialFile : public SequentialFile {
+ private:
```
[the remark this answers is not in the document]
matthewvon: I personally prefer protected in nearly every scenario because it simplifies future extension and unit test wrappers.  I believe "private" was from original authors.

matthewvon:
@ajkr Yes, I intend to reuse portions of the existing Encrypted class.  Once this PR merges with develop, I have a second PR waiting that implements OpenSSL CTR encryption ... and creates ability to migrate to/from 
- encryption and not
- different encryption keys
- different encryption code implementations
My basic assumption is that encryption is always changing but end users do not want to export entire dataset to move to next key / code.  Let compactions transition the data over time in non-critical scenarios (which are most scenarios).

matthewvon:
@ajkr  Thank you.  One Facebook Internal test is failing.  What, if anything, should I be doing to fix it?

include/rocksdb/env_encryption.h
```diff
@@ -194,13 +196,235 @@ class CTREncryptionProvider : public EncryptionProvider {
...
+  std::unique_ptr<SequentialFile> file_;
+  std::unique_ptr<BlockAccessCipherStream> stream_;
+  uint64_t offset_;
+  size_t prefixLength_;
+
+ public:
+  // Default ctor. Given underlying sequential file is supposed to be at
+  // offset == prefixLength.
+  EncryptedSequentialFile(SequentialFile* f, BlockAccessCipherStream* s,
```
mrambacher, line 217: I am concerned about the raw pointers here.  This makes it unclear who owns the CipherStream and SequentialFile.  What happens if there are errors thrown while creating these?  A unique or shared pointer would be a better option,

include/rocksdb/env_encryption.h
```diff
@@ -194,13 +196,235 @@ class CTREncryptionProvider : public EncryptionProvider {
...
+
+// A file abstraction for randomly reading the contents of a file.
+class EncryptedRandomAccessFile : public RandomAccessFile {
+ private:
+  std::unique_ptr<RandomAccessFile> file_;
+  std::unique_ptr<BlockAccessCipherStream> stream_;
+  size_t prefixLength_;
+
+ public:
+  EncryptedRandomAccessFile(RandomAccessFile* f, BlockAccessCipherStream* s,
```
mrambacher, line 270: Same issue with pointers.  These should not be passed in as raw pointers

include/rocksdb/env_encryption.h
```diff
@@ -194,13 +196,235 @@ class CTREncryptionProvider : public EncryptionProvider {
...
+// at a time to the file.
+class EncryptedWritableFile : public WritableFileWrapper {
+ private:
+  std::unique_ptr<WritableFile> file_;
+  std::unique_ptr<BlockAccessCipherStream> stream_;
+  size_t prefixLength_;
+
+ public:
+  // Default ctor. Prefix is assumed to be written already.
+  EncryptedWritableFile(WritableFile* f, BlockAccessCipherStream* s,
```
mrambacher, line 334: Should be unique pointers, not raw?

ajkr:
> @ajkr Thank you. One Facebook Internal test is failing. What, if anything, should I be doing to fix it?

I think there's a bug in syncing the test status. The internal test statuses look good to me.

yiwu-arbug:
Great to see the encryption interface continue to evolve!

matthewvon:
@mrambacher The raw pointers are part of the existing codebase.  I merely moved stuff around to enable inheritance.  Should correction be an issue and new PR?

mrambacher:
Previously the constructors that took raw pointers were not part of a public API whereas now they are.  Personally, I think they should be made unique_ptr as part of this effort, but I will leave it to @ajkr for the final word.  

@ajkr : Does the HISTORY.md file need to be updated as part of this PR (since it is a public API change)?

matthewvon:
P.S.  there is a follow-up PR.  The issue of unique_ptr could be addressed there.

include/rocksdb/env_encryption.h
matthewvon: It looks like this was based upon the existing SequentialFileWrapper class in env.h.  That wrapper class still takes raw pointer instead of unique_ptr.  Yes, it too could be updated.

ajkr:
> @ajkr : Does the HISTORY.md file need to be updated as part of this PR (since it is a public API change)?
 
it can be updated. While it isn't something existing use cases strictly need to know about, it could be beneficial for increasing adoption, etc.

matthewvon:
@ajkr Is the HISTORY.md update my action item?  And unique_ptr a requirement now, or next PR, or ???

ajkr:
Thought about it, how about this plan:
- HISTORY.md change is optional - if you want to advertise it there, feel free.
- Let's make the change from raw pointer to unique_ptr in this PR. If we deferred it to a later PR then it'd involve a breaking API change, which we'd rather avoid.

matthewvon:
@ajkr & @mrambacher  I have addressed the recent PR comments:
- std::unique_ptr now used to pass file pointers
- HISTORY.md line added
- private changed to protected in four file classes now exposed

matthewvon:
About the ci/circleci: build-linux-clang-no-test failure:  I cannot find any compile failure in the log, and the exact command line from the log finishes successfully on my Linux box (even with -j32).  Any suggestions would be very welcome: @ajkr

PR #6834, opened 2020-05-11
Use 'make all' in LITE Travis configuration

pdillinger:
Summary: So that we don't miss LITE compilation errors in tests


pdillinger:
Travis failed as expected on bad master. Updating to good master

PR #6835, opened 2020-05-11
Refactor the blob file related logic in VersionBuilder

ltamasi:
Summary:
This patch is groundwork for an upcoming change to store the set of
linked SSTs in `BlobFileMetaData`. With the current code, a new
`BlobFileMetaData` object is created each time a `VersionEdit` touches
a certain blob file. This is fine as long as these objects are lightweight
and cheap to create; however, with the addition of the linked SST set, it would
be very inefficient since the set would have to be copied over and over again.
Note that this is the same kind of problem that `VersionBuilder` is solving
w/r/t `Version`s and files, and we can apply the same solution; that is, we can
accumulate the changes in a different mutable object, and apply the delta in
one shot when the changes are committed. The patch does exactly that by
adding a new `BlobFileMetaDataDelta` class to `VersionBuilder`. In addition,
it turns the existing `GetBlobFileMetaData` helper into `IsBlobFileInVersion`
(which is fine since that's the only thing the method's clients care about now),
and adds a couple of helper methods that can create a `BlobFileMetaData`
object from the `BlobFileMetaData` in the base (if applicable) and the delta
when the `Version` is saved.

Test Plan:
`make check`

### DIFF
--- a/db/version_builder.cc
+++ b/db/version_builder.cc
@@ -559,6 +559,40 @@ class VersionBuilder::Rep {
     return s;
   }
 
+  static std::shared_ptr<BlobFileMetaData> CreateMetaDataForNewBlobFile(
+      const BlobFileMetaDataDelta& delta) {
+    auto shared_meta = delta.GetSharedMeta();
+    assert(shared_meta);
+
+    auto meta = BlobFileMetaData::Create(std::move(shared_meta),
+                                         delta.GetAdditionalGarbageCount(),
+                                         delta.GetAdditionalGarbageBytes());
+
+    return meta;
+  }
+
+  static std::shared_ptr<BlobFileMetaData>
+  GetOrCreateMetaDataForExistingBlobFile(
+      const std::shared_ptr<BlobFileMetaData>& base_meta,
+      const BlobFileMetaDataDelta& delta) {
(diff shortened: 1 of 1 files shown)


db/version_builder.cc
```diff
@@ -551,42 +608,42 @@ class VersionBuilder::Rep {
...
+        assert(base_meta);
         assert(base_meta->GetGarbageBlobCount() <
                base_meta->GetTotalBlobCount());
 
         vstorage->AddBlobFile(base_meta);
 
         ++base_it;
-      } else if (changed_blob_file_number < base_blob_file_number) {
-        AddBlobFileIfNeeded(vstorage, changed_meta);
+      } else if (delta_blob_file_number < base_blob_file_number) {
```
riversand963, line 627: Question:
If it's a new blob file, will it ever be smaller than the blob files already in base storage version?

ltamasi: Good point. No, with ever-increasing blob file numbers it is not possible to hit this branch. I was thinking about keeping this method fully generic but I guess it does not make much sense since we (will) rely on the strictly increasing property elsewhere. Will remove this and add an assertion.

PR #6839, opened 2020-05-12
Fix LITE build failure in compaction_picker_test

anand1976:
Fix LITE build.

Test:
LITE=1 make check


pdillinger:
Related: #6834

PR #6842, opened 2020-05-13
Fix valgrind error by init memory region

riversand963:
Summary:
As title. After allocating a memory buffer, initialize its content to 0s.
This fixes valgrind issue introduced in #6709.

Test Plan:
```
$make valgrind_test
$valgrind --tool=memcheck --track-origins=yes ./db_test2 --gtest_filter=DBTest2.CompressionFailures
```

### DIFF
--- a/db/db_test2.cc
+++ b/db/db_test2.cc
@@ -1317,10 +1317,10 @@ TEST_F(DBTest2, CompressionFailures) {
         if (curr_compression_failure_type == kTestDecompressionCorruption) {
           BlockContents* contents = static_cast<BlockContents*>(arg);
           // Ensure uncompressed data != original data
-          std::unique_ptr<char[]> fake_data(
-              new char[contents->data.size() + 1]);
-          *contents =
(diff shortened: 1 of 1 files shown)


db/db_test2.cc
```diff
@@ -1317,10 +1317,10 @@ TEST_F(DBTest2, CompressionFailures) {
         if (curr_compression_failure_type == kTestDecompressionCorruption) {
           BlockContents* contents = static_cast<BlockContents*>(arg);
           // Ensure uncompressed data != original data
-          std::unique_ptr<char[]> fake_data(
-              new char[contents->data.size() + 1]);
-          *contents =
-              BlockContents(std::move(fake_data), contents->data.size() + 1);
+          const size_t len = contents->data.size() + 1;
+          std::unique_ptr<char[]> fake_data(new char[len]);
+          memset(fake_data.get(), 0, len);
```
pdillinger, line 1322: I believe memset can be avoided with () after new[] (https://stackoverflow.com/a/7546745/454544)

PR #6843, opened 2020-05-13
avoid `IterKey::UpdateInternalKey()` in `BlockIter`

ajkr:
`IterKey::UpdateInternalKey()` is an error-prone API as it's
incompatible with `IterKey::TrimAppend()`, which is used for
decoding delta-encoded internal keys. This PR stops using it in
`BlockIter`. Instead, it assigns global seqno in a separate `IterKey`'s
buffer when needed. The logic for safely getting a Slice with global
seqno properly assigned is encapsulated in `GlobalSeqnoAppliedKey`.
`BinarySeek()` is also migrated to use this API (previously it ignored
global seqno entirely).

Test Plan:

benchmark setup -- single file DBs, in-memory, no compression. "normal_db"
created by regular flush; "ingestion_db" created by ingesting a file. Both
DBs have same contents.

```
$ TEST_TMPDIR=/dev/shm/normal_db/ ./db_bench -benchmarks=fillrandom,compact -write_buffer_size=10485760000 -disable_auto_compactions=true -compression_type=none -num=1000000
$ ./ldb write_extern_sst ./tmp.sst --db=/dev/shm/ingestion_db/dbbench/ --compression_type=no --hex --create_if_missing < <(./sst_dump --command=scan --output_hex --file=/dev/shm/normal_db/dbbench/000007.sst | awk 'began {print "0x" substr($1, 2, length($1) - 2), "==>", "0x" $5} ; /^Sst file format: block-based/ {began=1}')
$ ./ldb ingest_extern_sst ./tmp.sst --db=/dev/shm/ingestion_db/dbbench/
```

benchmark run command:
```
TEST_TMPDIR=/dev/shm/$DB/ ./db_bench -benchmarks=seekrandom -seek_nexts=10 -use_existing_db=true -cache_index_and_filter_blocks=false -num=1000000 -cache_size=1048576000 -threads=1 -reads=40000000
```

results:

| DB | code | throughput |
|---|---|---|
| normal_db | master |  267.9 |
| normal_db   |    PR6843 | 254.2 (-5.1%) |
| ingestion_db |   master |  259.6 |
| ingestion_db |   PR6843 | 250.5 (-3.5%) |

### DIFF
--- a/table/block_based/block.cc
+++ b/table/block_based/block.cc
@@ -188,8 +188,13 @@ void DataBlockIter::Prev() {
     const Slice current_key(key_ptr, current_prev_entry.key_size);
 
     current_ = current_prev_entry.offset;
-    key_.SetKey(current_key, false /* copy */);
+    raw_key_.SetKey(current_key, false /* copy */);
     value_ = current_prev_entry.value;
+    key_ = applied_key_.UpdateAndGetKey();
+    // This is kind of odd in that applied_key_ may say the key is pinned while
+    // key_pinned_ ends up being false. That'll only happen when the key resides
+    // in a transient caching buffer.
+    key_pinned_ = key_pinned_ && applied_key_.IsKeyPinned();
 
     return;
   }
@@ -217,9 +222,9 @@ void DataBlockIter::Prev() {
     if (!ParseNextDataKey<DecodeEntry>()) {
       break;
     }
-    Slice current_key = key();
+    Slice current_key = raw_key_.GetKey();
 
-    if (key_.IsKeyPinned()) {
+    if (raw_key_.IsKeyPinned()) {
       // The key is not delta encoded
       prev_entries_.emplace_back(current_, current_key.data(), 0,
                                  current_key.size(), value());
@@ -252,7 +257,8 @@ void DataBlockIter::Seek(const Slice& target) {
   SeekToRestartPoint(index);
 
   // Linear search (within restart block) for first key >= target
-  while (ParseNextDataKey<DecodeEntry>() && Compare(key_, seek_key) < 0) {
+  while (ParseNextDataKey<DecodeEntry>() &&
+         comparator_->Compare(applied_key_.UpdateAndGetKey(), seek_key) < 0) {
   }
 }
 
@@ -273,8 +279,8 @@ void DataBlockIter::Seek(const Slice& target) {
 //
 // If the return value is TRUE, iter location has two possibilies:
 // 1) If iter is valid, it is set to a location as if set by BinarySeek. In
-//    this case, it points to the first key_ with a larger user_key or a
-//    matching user_key with a seqno no greater than the seeking seqno.
+//    this case, it points to the first key with a larger user_key or a matching
+//    user_key with a seqno no greater than the seeking seqno.
 // 2) If the iter is invalid, it means that either all the user_key is less
 //    than the seek_user_key, or the block ends with a matching user_key but
 //    with a smaller [ type | seqno ] (i.e. a larger seqno, or the same seqno
@@ -330,7 +336,8 @@ bool DataBlockIter::SeekForGetImpl(const Slice& target) {
     //
     // TODO(fwu): check the left and write boundary of the restart interval
     // to avoid linear seek a target key that is out of range.
-    if (!ParseNextDataKey<DecodeEntry>(limit) || Compare(key_, target) >= 0) {
+    if (!ParseNextDataKey<DecodeEntry>(limit) ||
+        comparator_->Compare(applied_key_.UpdateAndGetKey(), target) >= 0) {
       // we stop at the first potential matching user key.
       break;
     }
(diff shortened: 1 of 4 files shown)


ajkr:
There are a few next steps that were too hard to do all at once.

1. migrate IndexBlockIter to use GlobalSeqnoAppliedKey for first_internal_key
2. introduce a separate convenience class like IterKey for non-iterator keys. Currently IterKey is used in  places unrelated to iteration/delta decoding (for example, `GlobalSeqnoAppliedKey::scratch_`).
3. finish migrating everything away from `IterKey::UpdateInternalKey()` and then delete it.

table/block_based/block.cc
```diff
@@ -188,7 +188,7 @@ void DataBlockIter::Prev() {
     const Slice current_key(key_ptr, current_prev_entry.key_size);
 
     current_ = current_prev_entry.offset;
-    key_.SetKey(current_key, false /* copy */);
+    raw_key_.SetKey(current_key, false /* copy */);
```
ajkr, line 191: there's a bug here. the `current_key` came from the entries cached during Prev. Those cached keys have global seqno already assigned. If we assign such a key to `raw_key_` and then the scan direction switches to forward, then we're exposed to the delta-decoding bug we originally set out to avoid.

ajkr: Fixed. Changed Prev's caching to cache raw keys.

ajkr: I think this also fixes an existing bug when `ParseNextDataKey()` uses a cached `Prev()` entry as its starting point. In that case the seqno/value type it restores did not necessarily come from the current key. They came from the most recent key on which we called ParseNextDataKey(), which could be a later key. That means `ParseNextDataKey()` is restoring the wrong key's `ValueType` before it advances. Technically it's restoring the wrong key's seqno too, but it just so happens to work since in ingested files the original seqnos are all the same (zero).

ajkr: The existing bug is a bit different from what I described above. The restoration of the wrong key's seqno/value type just so happens to be OK because it restores them to the wrong buffer (oops). In that case, we're not restoring anything to the cached Prev entry before we use it for delta-decoding, which means the seqno was corrupted by the global seqno.

table/block_based/block.h
```diff
@@ -228,6 +228,49 @@ class Block {
   DataBlockHashIndex data_block_hash_index_;
 };
 
+// A GlobalSeqnoAppliedKey exposes a key with global sequence number applied
+// if configured with `global_seqno != kDisableGlobalSequenceNumber`. It may
+// hold a user key or an internal key since `format_version>=3` index blocks
+// contain user keys. In case it holds user keys, it must be configured with
+// `global_seqno == kDisableGlobalSequenceNumber`.
+class GlobalSeqnoAppliedKey {
```
pdillinger, line 236: At first this looks like a (not particularly efficient) data holder in which repeated calls to GetKey() without re-Initialize() would needlessly recompute the data.

But no, it appears we expect contents of key_ to change, and that GetKey() is really UpdateAndGetKey(). Please drop the const illusion of GetKey(), because I find it rather misleading. This object clearly owns data (otherwise, how does GetKey() return a Slice?) that is modified by GetKey, and mutating data you clearly own (based on public interface) should not be const function.

ajkr: Yeah, it makes sense, I will change it. My initial goal (and still long-term goal actually) is to have a `GetKey() const` that returns a `ParsedInternalKey`, which would point into `key_` for the user key, and have a copy of the global seqno/original value type. But we are not there yet as the way `BlockIter` tries to be agnostic to internal vs. user key makes it complicated...

table/block_based/block.cc
```diff
@@ -717,7 +716,8 @@ int IndexBlockIter::CompareBlockKey(uint32_t block_index, const Slice& target) {
     return 1;  // Return target is smaller
   }
   Slice block_key(key_ptr, non_shared);
-  return Compare(block_key, target);
+  raw_key_.SetInternalKey(block_key, false /* copy */);
+  return comparator_->Compare(applied_key_.UpdateAndGetKey(), target);
```
pdillinger, line 720: We should change the name from CompareBlockKey if it's now (intentionally) comparing with internal key.

ajkr: Hm, I am confused, this function feels like it needs to be able to handle either internal key or user key (index key could be either one depending on the `format_version`). Calling `SetInternalKey()` above looks like a mistake.

ajkr:
> > Maybe also run a seekrandom benchmark (without cache_index_and_filter_blocks) to see what kind of worst-case performance impact this has.
> 
> Sure, will do.

Added to the description. The results aren't that stable, but I do consistently see slowdown in both the "ingestion_db" and "normal_db" case. That is different from what I expected as I thought the main drawback of this PR was introducing a copy into `GlobalSeqnoAppliedKey::scratch_` only in the "ingestion_db" case.

table/block_based/block.cc
ajkr: Done (call `SetKey()` instead).

pdillinger:
(I believe the CI failures are all unrelated. Do another rebase if you want.)

ajkr:
Thanks for the review!

> I think we can accept minor perf regression to help bring some sanity to this part of the code.

The "normal_db" regression can be avoided if we change `UpdateAndGetKey()` to assume it holds non-ingested keys (i.e., `if (global_seqno_ == kDisableGlobalSequenceNumber)` -> `if (LIKELY(global_seqno_ == kDisableGlobalSequenceNumber))`). It's not obvious to me we should do that so I'll leave it as is for now.

The "ingestion_db" regression is due to the extra copy into `GlobalSeqnoAppliedKey::scratch_` that can only be avoided by comparing against `ParsedInternalKey`. More refactoring is necessary first.

PR #6844, opened 2020-05-13
Move checksum calculation ahead of memory copy

zhichao-cao:
Originally, the checksum of appended data in writable file writer is calculated after the data is copied to the buffer. It will not be able to catch the bit flip happens during copy. Move the checksum calculation before it.


zhichao-cao:
> LGTM

Thanks!

PR #6847, opened 2020-05-15
Fix buck target db_stress_lib in opt mode

riversand963:
In buck build with opt mode, target should not include rocksdb_test_lib.

Test plan:
Watch for internal cont build.


riversand963:
Thanks @ajkr for the review!

PR #6849, opened 2020-05-15
Strengthen MultiGet correctness verification in NoBatchedOpsStress

anand1976:
Add MultiGet to VerifyDb and check consistency with Get in TestMultiGet.

Test plan -
make crash_test
ASAN crash test

### DIFF
--- a/db_stress_tool/no_batched_ops_stress.cc
+++ b/db_stress_tool/no_batched_ops_stress.cc
@@ -365,7 +365,10 @@ class NonBatchedOpsStressTest : public StressTest {
         std::string value;
 
         tmp_s = db_->Get(readoptionscopy, cfh, keys[i], &value);
-        if (!s.ok() && tmp_s.ok()) {
+        if (!tmp_s.ok() && !tmp_s.IsNotFound()) {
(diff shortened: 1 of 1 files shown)


db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -309,10 +349,49 @@ class NonBatchedOpsStressTest : public StressTest {
...
+                  keys[i].ToString(true).c_str());
+          fprintf(stderr, "Get returned ok, MultiGet returned not found\n");
+          is_consistent = false;
+        } else if (s.ok() && tmp_s.IsNotFound()) {
+          fprintf(stderr, "MultiGet returned different results with key %s\n",
+                  keys[i].ToString(true).c_str());
+          fprintf(stderr, "MultiGet returned ok, Get returned not found\n");
+          is_consistent = false;
+        } else if (s.ok() && value != values[i].ToString()) {
+          fprintf(stderr, "MultiGet returned different results with key %s\n",
```
pdillinger, line 379: I think we need to check for error status from Get, even though we're outside error injection region. We at least want to see what that status is if it's an error, and hypothetically, the "correct" value could be empty string (which would currently mask an error from Get).

PR #6850, opened 2020-05-15
Compaction filter support for BlobDB

burtonli:
Added compaction filter support for BlobDB non-TTL values. Same as vanilla RocksDB, user compaction filter applies to all k/v pairs of the compaction for non-TTL values. It honors `min_blob_size`, which potentially results value transitions between inlined data and stored-in-blob data when size of value is changed.


utilities/blob_db/blob_compaction_filter.h
```diff
@@ -49,17 +50,27 @@ class BlobIndexCompactionFilterBase : public CompactionFilter {
...
+  Decision FilterV2(int level, const Slice& key, ValueType value_type,
+                    const Slice& value, std::string* new_value,
+                    std::string* skip_until) const override;
+
+  void SetUserCompactionFilter(const CompactionFilter* user_filter);
+  void SetUserCompactionFilter(std::unique_ptr<CompactionFilter>* user_filter);
 
  protected:
   Statistics* statistics() const { return statistics_; }
+  const CompactionFilter* user_data_filter() const { return user_filter_; }
```
mrambacher, line 62: This name appears to violate the naming convention.

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -316,12 +387,23 @@ BlobIndexCompactionFilterFactoryGC::CreateCompactionFilter(
...
+      std::move(bcontext), std::move(bcontext_gc), current_time, statistics());
+
+  // compaction_filter takes precedence over compaction_filter_factory if
+  // client specifies both. (refer to include\rocksdb\options.h)
+  if (user_filter_ != nullptr) {
+    rv->SetUserCompactionFilter(user_filter_);
+  } else if (user_filter_factory_ != nullptr) {
+    auto user_filter = user_filter_factory_->CreateCompactionFilter(context);
+    rv->SetUserCompactionFilter(&user_filter);
+  }
```
mrambacher, line 404: Please look at how the TTLCompactionFilter code is created.  Perhaps we can use a similar model (or common base class)?

ltamasi:
Thanks for the PR @burtonli! I will review it this week.

utilities/blob_db/blob_compaction_filter.cc
burtonli: Good point, add a common class LayeredCompactionFilterBase

burtonli:
@ltamasi yes, the compaction filter only works for BlobDB enables GC, but still this may benefit many scenarios, For example, our scenario high relays compaction filter to do the lazy GC on some patterned data, at the same time, we also has filter on the read path to make sure those data is invisible before they are physically removed.
What about for current stage, we let user know the limitation, and at least for those GC enabled scenario, we could provide the compaction filter support? For long term, I agree that a deeper integration with RocksDB core would make it is possible to do a general support.

ltamasi:
> @ltamasi Thank you for the comments! Yes, the compaction filter only works for BlobDB enables GC, but still this might benefit many scenarios, For example, our scenario highly relies on compaction filter to do the lazy GC for patterned data, at the same time, we also has filter on the read path to make sure those data is invisible before they are physically removed.
> Do you think it's possible to limit the scope for the support and let user know the limitation, and at least for those GC enabled scenario, we can provide compaction filter support? For long term, I agree that a deeper integration with RocksDB core would make it is possible to do a general support.

I think it would be OK to start with partial support; however, we should make sure the rules are relatively straightforward (the semantics of compaction filters are complicated enough as is, see https://github.com/facebook/rocksdb/wiki/Compaction-Filter). Two partial solutions that we could consider:

1) Call the compaction filter for *all* non-TTL values (stored either in blob files, or inlined, in which case they are stored as `kTypeValue`) and make GC a prerequisite for compaction filter support. This should be fairly simple and is relatively close to what's in the PR, however we would need to make sure that a) the compaction filter is called for all non-TTL values in blob files, not just the ones being relocated during GC, and b) the compaction filter is *not* called for any TTL blobs (in particular, it is *not* called for inlined TTL blobs).

2) Similar to the above, call the compaction filter for all non-TTL values but do *not* make GC a prerequisite. This would make it possible to use the feature with BlobDB's FIFO logic. Implementation-wise, it would involve factoring out the read/write blob machinery currently in `BlobIndexCompactionFilterGC`.

ltamasi:
Couple more points regarding the implementation:

1) Since the compaction filter can change the value (and thus the length of the value), we would need to support "inlined -> blob file" and "blob file -> inlined" transitions.
2) `PrepareBlobValue` is meant to be exclusively used for relocation during GC; dropping KVs at that point is intentionally unsupported. Ideally we would perform all calls to the user compaction filter in `FilterV2`; I feel this would be a cleaner/more decoupled design.

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -13,57 +13,128 @@
...
+    }
   }
-  return Decision::kKeep;
+  // If value changed, update value either inline or into blob file.
+  if (decision == Decision::kChangeValue) {
+    BlobDBImpl* const blob_db_impl = context_.blob_db_impl;
+    assert(blob_db_impl);
+    if (new_filter_value.size() > blob_db_impl->bdb_options_.min_blob_size) {
+      if (!OpenNewBlobFileIfNeeded()) {
+        return Decision::kKeep;
```
ltamasi, line 104: How to handle I/O errors here is a good question; we don't want to end up with partially written/potentially corrupt blob files. Note that if we've already put some blobs in a new blob file and emitted the corresponding `BlobIndex`es, we cannot simply give up on the file (because then you would have dangling pointers). For GC, we simply abort the entire compaction in such cases. Will think about this more.

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -281,9 +303,90 @@ bool BlobIndexCompactionFilterGC::CloseAndRegisterNewBlobFile() const {
...
+  return BlobDecision::kChangeValue;
+}
+
+bool BlobIndexCompactionFilterGC::OpenNewBlobFileWithStatsIfNeeded() const {
+  if (IsBlobFileOpened()) {
+    return true;
+  }
+  bool result = OpenNewBlobFileIfNeeded();
+  if (result) {
+    gc_stats_.AddNewFile();
```
ltamasi, line 382: So with this code, only the files opened by `BlobIndexCompactionFilterGC` itself would be counted in the GC statistics, even though relocated blobs might also end up in files opened by the base class, right?

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -295,16 +398,24 @@ BlobIndexCompactionFilterFactory::CreateCompactionFilter(
 
   assert(blob_db_impl());
 
-  BlobCompactionContext context;
-  blob_db_impl()->GetCompactionContext(&context);
+  BlobCompactionContext bcontext;
+  blob_db_impl()->GetCompactionContext(&bcontext);
```
ltamasi, line 402: Nit: please either leave the name as-is, or rename to `blob_compaction_context`. But personally I prefer the shorter `context`. Also affects the other class below.

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -295,16 +398,24 @@ BlobIndexCompactionFilterFactory::CreateCompactionFilter(
...
+
+  std::unique_ptr<const CompactionFilter> user_comp_filter_from_factory =
+      nullptr;
+  if (user_comp_filter_factory_) {
+    user_comp_filter_from_factory =
+        user_comp_filter_factory_->CreateCompactionFilter(context);
+  }
```
ltamasi, line 409: We could consider factoring this out into a helper in the base class.

utilities/blob_db/blob_compaction_filter.h
```diff
@@ -11,55 +11,72 @@
...
 // Compaction filter that deletes expired blob indexes from the base DB.
 // Comes into two varieties, one for the non-GC case and one for the GC case.
-class BlobIndexCompactionFilterBase : public CompactionFilter {
+class BlobIndexCompactionFilterBase : public LayeredCompactionFilterBase {
  public:
-  BlobIndexCompactionFilterBase(BlobCompactionContext&& context,
-                                uint64_t current_time, Statistics* stats)
-      : context_(std::move(context)),
+  BlobIndexCompactionFilterBase(
+      BlobCompactionContext&& blob_comp_context,
```
ltamasi, line 38: Again, `context` would be preferred, here and in the other class below.

utilities/blob_db/blob_db_impl.cc
```diff
@@ -1123,11 +1127,17 @@ Status BlobDBImpl::PutBlobValue(const WriteOptions& /*options*/,
 
 Slice BlobDBImpl::GetCompressedSlice(const Slice& raw,
                                      std::string* compression_output) const {
-  if (bdb_options_.compression == kNoCompression) {
+  return GetCompressedSlice(raw, bdb_options_.compression, compression_output);
```
ltamasi, line 1130: I was wondering why this overload was needed? We exclusively use `bdb_options_.compression`, right?

db/compaction/compaction_iterator.cc
```diff
@@ -651,6 +654,12 @@ void CompactionIterator::PrepareOutput() {
       } else if (blob_decision == CompactionFilter::BlobDecision::kIOError) {
         status_ = Status::IOError("Could not relocate blob during GC");
         valid_ = false;
+      } else if (blob_decision == CompactionFilter::BlobDecision::kRemove) {
+        ikey_.type = kTypeDeletion;
+        current_key_.UpdateInternalKey(ikey_.sequence, kTypeDeletion);
+        value_.clear();
+        iter_stats_.num_record_drop_user++;
+        return;
```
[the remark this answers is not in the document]
burtonli: Yes, I forgot to remove it from original approach. :)

include/rocksdb/compaction_filter.h
```diff
@@ -158,7 +165,8 @@ class CompactionFilter {
                             const Slice& existing_value, std::string* new_value,
                             std::string* /*skip_until*/) const {
     switch (value_type) {
-      case ValueType::kValue: {
+      case ValueType::kValue:
+      case ValueType::kBlobIndex: {
```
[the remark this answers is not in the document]
burtonli: The reason we need to add ValueType::kBlobIndex is to support scenario when user only provide overridden Filter() api not FilterV2(). See the implementation for BlobIndexCompactionFilterBase::FilterV2() line:94, we only call ucf->FilterV2(), so we need this to wrapper to redirect the call to Filter() in user compaction filter. 
But we can also pass kValue into ucf->FilterV2() instead of value_type, since user may not expect internal concept kBlobIndex. Let me try that approach.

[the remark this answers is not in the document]
ltamasi: Oh yeah, we should definitely pass `kValue` even if the value originally came from a blob file.

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -13,57 +13,128 @@
...
+        // Keep the blob index in case fails to read the blob value.
+        return Decision::kKeep;
+      }
+      decision = ucf->FilterV2(level, ikey.user_key, value_type, blob,
+                               &new_filter_value, skip_until);
+    }
   }
-  return Decision::kKeep;
+  // If value changed, update value either inline or into blob file.
+  if (decision == Decision::kChangeValue) {
```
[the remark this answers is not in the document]
burtonli: Good suggestion, thanks!

utilities/blob_db/blob_compaction_filter.cc
burtonli: yes, the stats for BlobDBGarbageCollectionStats.

utilities/blob_db/blob_compaction_filter.h
burtonli: Change to context would have build warnings, as we have context() in BlobIndexCompactionFilterBase. Let me use _context instead.

utilities/blob_db/blob_db_impl.cc
burtonli: You are right, we don't need this anymore.

utilities/blob_db/blob_compaction_filter.cc
burtonli: Same here, use the name context would have build warning, use _context instead.

db/compaction/compaction_iterator.cc
```diff
@@ -231,6 +231,9 @@ void CompactionIterator::InvokeFilterIfNeeded(bool* need_skip,
       compaction_filter_skip_until_.ConvertFromUserKey(kMaxSequenceNumber,
                                                        kValueTypeForSeek);
       *skip_until = compaction_filter_skip_until_.Encode();
+    } else if (filter == CompactionFilter::Decision::kChangeBlobIndex) {
+      ikey_.type = kTypeBlobIndex;
```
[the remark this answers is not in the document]
ltamasi: One more thing: we should change the `kChangeValue` code path similarly to this to account for the fact that a value that used to be in a blob file can potentially be inlined after the compaction filter is done with it. Also, in both cases, we only need to update the type and call `UpdateInternalKey` if it's different from the original type (i.e. it went either from stored-in-blob-file to inlined, or vice versa).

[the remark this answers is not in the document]
burtonli: Good catch, thanks!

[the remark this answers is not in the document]
burtonli: > One more thing: we should change the `kChangeValue` code path similarly to this to account for the fact that a value that used to be in a blob file can potentially be inlined after the compaction filter is done with it. Also, in both cases, we only need to update the type and call `UpdateInternalKey` if it's different from the original type (i.e. it went either from stored-in-blob-file to inlined, or vice versa).
Good catch! I will update, thanks!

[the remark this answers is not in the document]
burtonli: Enrich the unit test to cover data transitions for both way.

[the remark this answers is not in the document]
burtonli: @ltamasi could you help to proceed the new iteration at your convenience, thanks!

utilities/blob_db/blob_db_test.cc
```diff
@@ -1111,27 +1111,112 @@ TEST_F(BlobDBTest, InlineSmallValues) {
...
+        *new_value = value.ToString();
+        // double size by duplicating value
+        *new_value += *new_value;
+        *value_changed = true;
+        return false;
+      } else if (value.size() % 3 == 1) {
+        *new_value = value.ToString();
+        // trancate value size by half
+        *new_value = new_value->substr(0, new_value->size() / 2);
+        *value_changed = true;
```
ltamasi, line 1132: `return false;` seems to be missing from here.

utilities/blob_db/blob_db_test.cc
```diff
@@ -1111,27 +1111,112 @@ TEST_F(BlobDBTest, InlineSmallValues) {
...
+  constexpr uint64_t kMaxValueSize = 1 << 8;
+  constexpr uint64_t kMinBlobSize = 1 << 7;
+  static_assert(kMinValueSize < kMinBlobSize, "");
+  static_assert(kMaxValueSize > kMinBlobSize, "");
+
+  BlobDBOptions bdb_options;
+  bdb_options.min_blob_size = kMinBlobSize;
+  bdb_options.blob_file_size = kMaxValueSize * 10;
+  bdb_options.enable_garbage_collection = true;
+  bdb_options.garbage_collection_cutoff = 0.25;
```
ltamasi, line 1161: GC is no longer necessary, right?

burtonli: > GC is no longer necessary, right?

Correct, we can run test even with garbage_collection_cutoff = 0.  I can remove it to use default optoin.

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -44,26 +65,90 @@ CompactionFilter::Decision BlobIndexCompactionFilterBase::FilterV2(
...
+    }
+    Slice new_blob_value(new_value_from_user_filter);
+    std::string compression_output;
+    if (blob_db_impl->bdb_options_.compression != kNoCompression) {
+      new_blob_value =
+          blob_db_impl->GetCompressedSlice(new_blob_value, &compression_output);
+    }
+    uint64_t new_blob_file_number = 0;
+    uint64_t new_blob_offset = 0;
+    // For kBlobIndex 'key' is internal key; Otherwise, it is user key.
```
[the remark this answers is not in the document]
burtonli: correct, I'm removing this leftover comment.

burtonli:
@ltamasi Thanks again for reviewing, all comments are addressed.

db/compaction/compaction_iterator.cc
```diff
@@ -225,12 +225,28 @@ void CompactionIterator::InvokeFilterIfNeeded(bool* need_skip,
...
+      if (ikey_.type == kTypeValue) {
+        // value transfer from inlined data to blob file
+        ikey_.type = kTypeBlobIndex;
+        current_key_.UpdateInternalKey(ikey_.sequence, ikey_.type);
+      }
+      value_ = compaction_filter_value_;
+    } else if (filter == CompactionFilter::Decision::kIOError) {
+      status_ =
+          Status::IOError("Failed to access blob during compaction filter");
+      valid_ = false;
```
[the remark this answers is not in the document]
burtonli: Yes, only valid_ == true would break the loop. I will update and add UT to cover.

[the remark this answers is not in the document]
burtonli: Added unit tests, and found some defects in original implementation.
1. blob_compactoin_filter.cc, line:298 blob_file_ is reset before error log, which uses blob_file_->PathName().
2. blob_db_impl.cc, line:830, if CheckOrCreateWriterLocked fails, we need to reset blob_file, otherwise ~BlobIndexCompactionFilterBase() will call CloseAndRegisterNewBlobFile() and it hits some assert eventually (e.g. log_writer_->AppendFooter(footer)).

utilities/blob_db/blob_db_test.cc
```diff
@@ -1111,27 +1111,111 @@ TEST_F(BlobDBTest, InlineSmallValues) {
   ASSERT_TRUE(ttl_file->HasTTL());
 }
 
-TEST_F(BlobDBTest, CompactionFilterNotSupported) {
-  class TestCompactionFilter : public CompactionFilter {
-    const char *Name() const override { return "TestCompactionFilter"; }
+TEST_F(BlobDBTest, UserCompactionFilter) {
+  class CustomerFilter : public CompactionFilter {
+   public:
+    bool Filter(int /*level*/, const Slice & /*key*/, const Slice &value,
```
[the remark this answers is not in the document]
burtonli: Interesting, I always run clang-format before submit PR.

db/compaction/compaction_iterator.cc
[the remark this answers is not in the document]
ltamasi: Great catch, thanks @burtonli !

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -6,15 +6,17 @@
...
 
+#include "db/dbformat.h"
+#include "test_util/sync_point.h"
+
 namespace ROCKSDB_NAMESPACE {
 namespace blob_db {
 
 BlobIndexCompactionFilterBase::~BlobIndexCompactionFilterBase() {
-  if (blob_file_) {
+  if (blob_file_ && blob_file_->GetWriter()) {
```
ltamasi, line 19: The original idea was for the class to maintain the invariant `!!blob_file_ == !!writer_` (that is, either both are valid OR neither is). Instead of this change, can we fix `OpenNewBlobFileIfNeeded` so it clears both `shared_ptr`s if `CreateBlobFileAndWriter` returns failure?

utilities/blob_db/blob_db_test.cc
```diff
@@ -1219,6 +1221,80 @@ TEST_F(BlobDBTest, UserCompactionFilter) {
   }
 }
 
+// Test user comapction filter when there is IO error on blob data.
+TEST_F(BlobDBTest, UserCompactionFilter_BlobIOError) {
```
ltamasi, line 1225: This is awesome, thanks for adding it!

utilities/blob_db/blob_db_test.cc
```diff
@@ -1159,7 +1159,9 @@ TEST_F(BlobDBTest, UserCompactionFilter) {
   bdb_options.min_blob_size = kMinBlobSize;
   bdb_options.blob_file_size = kMaxValueSize * 10;
   bdb_options.disable_background_tasks = true;
+#ifdef SNAPPY
```
ltamasi, line 1162: Very minor: could you do `if (Snappy_Supported()) { ... }` instead?

utilities/blob_db/blob_compaction_filter.cc
burtonli: Done

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -180,7 +180,7 @@ bool BlobIndexCompactionFilterBase::OpenNewBlobFileIfNeeded() const {
     ROCKS_LOG_ERROR(blob_db_impl->db_options_.info_log,
                     "Error opening new blob file during GC, status: %s",
                     s.ToString().c_str());
-
+    blob_file_.reset();
```
ltamasi, line 183: Could you also reset `writer_`? It is possible for `CreateBlobFileAndWriter` to fail after the `Writer` has been created (namely in `WriteHeader`).

burtonli: > Could you also reset `writer_`? It is possible for `CreateBlobFileAndWriter` to fail after the `Writer` has been created (namely in `WriteHeader`).

Make sense, done.

PR #6851, opened 2020-05-15
Eliminate the creation of ImmutableDBOptions in WBWI::GetFromBatch

mrambacher:
1. Made `WriteBatchWithIndexInternal` into a class that stores the `DB*` or `DBOptions*`.

2. Changed the `GetFromBatch` method to be non-static and use an instance of the class.  Added `MergeKey` methods to perform the merge itself and return any status.

This change unifies the multiple calls to the `MergeHelper` under a single wrapped API. 

Closes https://github.com/facebook/rocksdb/issues/6683


utilities/write_batch_with_index/write_batch_with_index_internal.h
```diff
@@ -122,8 +124,92 @@ class WriteBatchEntryComparator {
   const ReadableWriteBatch* write_batch_;
 };
 
+typedef SkipList<WriteBatchIndexEntry*, const WriteBatchEntryComparator&>
+    WriteBatchEntrySkipList;
+
+class WBWIIteratorImpl : public WBWIIterator {
```
adamretter, line 130: I see this has moved from the `.c` to the `.h` file. I am wondering if we should have just moved the definition and kept the declaration in the `.cc` file? I don't know the answer, I just thought I would query it...

mrambacher: I am fine with putting only the definition in the header file.  I did a copy/paste move.  There does not seem to be a good standard for RocksDB on what goes in header versus source files.

adamretter: Yeah indeed I have seen a mix of styles in the RocksDB code-base. I wasn't sure what the "house style" was supposed to be though...

utilities/write_batch_with_index/write_batch_with_index_internal.cc
```diff
@@ -151,34 +153,105 @@ int WriteBatchEntryComparator::CompareKey(uint32_t column_family,
...
+          logger, statistics, env, result_operand);
+    } else if (db_options_ != nullptr) {
+      Statistics* statistics = db_options_->statistics.get();
+      Env* env = db_options_->env;
+      Logger* logger = db_options_->info_log.get();
+      return MergeHelper::TimedFullMerge(
+          merge_operator, key, value, merge_context.GetOperands(), result,
+          logger, statistics, env, result_operand);
+    } else {
+      return MergeHelper::TimedFullMerge(
```
adamretter, line 226: I don't think this is a valid state in which to call `TimedFullMerge`. If you call `TimedFullMerge` with a `nullptr` as the `Logger`. Then the MergeOperator implementation may SIGSEGV because AFAIK, Logger should not be null when calling `MergeOperator::FullMergeV2` or `MergeOperator::FullMerge`.

adamretter: Hmm perhaps not... It seems that its perfectly valid for the `Logger*` to be null. So don't worry ;-)

pdillinger:
Facebook lint results:
A single-argument constructor for WriteBatchWithIndexInternal is missing the explicit specifier.

Possible null deref calling wbwii.MergeKey with null merge_data. (I did not really investigate.)

mrambacher:
> Facebook lint results:
> A single-argument constructor for WriteBatchWithIndexInternal is missing the explicit specifier.
> 
> Possible null deref calling wbwii.MergeKey with null merge_data. (I did not really investigate.)

Can you give more info on the MergeKey deref?  I do not see a variable "merge_data" or have an idea what is the potential null deref by looking at the code.

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -1008,27 +888,15 @@ void WriteBatchWithIndex::MultiGetFromBatchAndDB(
...
-          *key.s = MergeHelper::TimedFullMerge(
-              merge_operator, *key.key, merge_data,
-              merge_result.second.GetOperands(), key.value->GetSelf(), logger,
-              statistics, env);
-          key.value->PinSelf();
-        } else {
-          *key.s =
-              Status::InvalidArgument("Options::merge_operator must be set");
-        }
+        *key.s = wbwii.MergeKey(*key.key, merge_data, merge_result.second,
```
pdillinger, line 897: Lint reports possible null deref of merge_data reported here. (I did not dig far enough into MergeKey to confirm or refute.)

mrambacher: The merge_data argument is "value" in MergeKey.  The "value" is never accessed in MergeKey but is passed through as the third argument to MergeHelper::TimedFullMerge.  A quick grep of the code base finds lots of places passing nullptr in for this argument to TimedFullMerge.  

I believe a nullptr is meant to indicate the value does not exist in the database.

mrambacher: I am not sure exactly what this error is or how to solve it.  I re-worked the code slightly to hopefully workaround the lint report.

mrambacher:
I have updated this PR to master.  Can someone please re-review it and merge it?  Thanks!

utilities/write_batch_with_index/write_batch_with_index_internal.cc
```diff
@@ -155,34 +156,105 @@ int WriteBatchEntryComparator::CompareKey(uint32_t column_family,
...
+  uint32_t cf_id = GetColumnFamilyID(column_family_);
   *s = Status::OK();
-  WriteBatchWithIndexInternal::Result result =
-      WriteBatchWithIndexInternal::Result::kNotFound;
+  Result result = kNotFound;
 
-  std::unique_ptr<WBWIIterator> iter =
-      std::unique_ptr<WBWIIterator>(batch->NewIterator(column_family));
+  std::unique_ptr<WBWIIteratorImpl> iter(
+      static_cast<WBWIIteratorImpl*>(batch->NewIterator(column_family_)));
```
pdillinger, line 246: static_cast_with_check is preferred

utilities/write_batch_with_index/write_batch_with_index_internal.cc
```diff
@@ -155,34 +156,105 @@ int WriteBatchEntryComparator::CompareKey(uint32_t column_family,
...
+    ColumnFamilyHandle* column_family)
+    : db_(nullptr), db_options_(nullptr), column_family_(column_family) {}
+
+Status WriteBatchWithIndexInternal::MergeKey(const Slice& key,
+                                             const Slice* value,
+                                             MergeContext& merge_context,
+                                             std::string* result,
+                                             Slice* result_operand) {
+  if (column_family_ != nullptr) {
+    auto cfh = reinterpret_cast<ColumnFamilyHandleImpl*>(column_family_);
```
pdillinger, line 205: reinterpret_cast is WRONG(-ish) for downcasting!

https://stackoverflow.com/a/35528811/454544

utilities/write_batch_with_index/write_batch_with_index_internal.h
```diff
@@ -440,8 +442,98 @@ class WriteBatchEntryComparator {
...
+  uint32_t column_family_id_;
+  WriteBatchEntrySkipList::Iterator skip_list_iter_;
+  const ReadableWriteBatch* write_batch_;
+  WriteBatchEntryComparator* comparator_;
+};
+
 class WriteBatchWithIndexInternal {
  public:
+  explicit WriteBatchWithIndexInternal(
+      DB* db, ColumnFamilyHandle* column_family = nullptr);
```
pdillinger, line 531: It seems like providing "convenience" to omit column_family on an internal API is inviting bugs. How good is our test coverage of WBWI outside of default column family?

utilities/write_batch_with_index/write_batch_with_index_internal.cc
```diff
@@ -155,34 +156,105 @@ int WriteBatchEntryComparator::CompareKey(uint32_t column_family,
...
+    return comparator_->CompareKey(cf_id, key, Entry().key) == 0;
+  } else {
+    return false;
+  }
+}
+
+WriteBatchWithIndexInternal::WriteBatchWithIndexInternal(
+    DB* db, ColumnFamilyHandle* column_family)
+    : db_(db), db_options_(nullptr), column_family_(column_family) {
+  if (db_ != nullptr && column_family_ == nullptr) {
```
pdillinger, line 186: It seems like providing "convenience" to omit column_family on an internal API is inviting bugs.

utilities/write_batch_with_index/write_batch_with_index_internal.cc
```diff
@@ -155,34 +156,105 @@ int WriteBatchEntryComparator::CompareKey(uint32_t column_family,
...
+    column_family_ = db_->DefaultColumnFamily();
+  }
+}
+
+WriteBatchWithIndexInternal::WriteBatchWithIndexInternal(
+    const DBOptions* db_options, ColumnFamilyHandle* column_family)
+    : db_(nullptr), db_options_(db_options), column_family_(column_family) {}
```
pdillinger, line 193: What if column_family is nullptr as in default? Below I see `Status::InvalidArgument("Must provide a column_family")`

utilities/write_batch_with_index/write_batch_with_index_internal.cc
```diff
@@ -155,34 +156,105 @@ int WriteBatchEntryComparator::CompareKey(uint32_t column_family,
...
+
+WriteBatchWithIndexInternal::WriteBatchWithIndexInternal(
+    ColumnFamilyHandle* column_family)
+    : db_(nullptr), db_options_(nullptr), column_family_(column_family) {}
```
pdillinger, line 197: What if column_family is nullptr as in default? Below I see `Status::InvalidArgument("Must provide a column_family")`

utilities/write_batch_with_index/write_batch_with_index_internal.h
```diff
@@ -452,11 +544,25 @@ class WriteBatchWithIndexInternal {
...
+                      bool overwrite_key, Status* s);
+  Status MergeKey(const Slice& key, const Slice* value, std::string* result,
+                  Slice* result_operand = nullptr) {
+    return MergeKey(key, value, merge_context_, result, result_operand);
+  }
+  Status MergeKey(const Slice& key, const Slice* value, MergeContext& context,
+                  std::string* result, Slice* result_operand = nullptr);
+
+ private:
+  DB* db_;  // Not owned
```
pdillinger, line 562: I don't think any of these pointers are owned. Why only mark this one as unowned and not the others?

utilities/write_batch_with_index/write_batch_with_index_internal.cc
```diff
@@ -155,34 +156,105 @@ int WriteBatchEntryComparator::CompareKey(uint32_t column_family,
...
+
+Status WriteBatchWithIndexInternal::MergeKey(const Slice& key,
+                                             const Slice* value,
+                                             MergeContext& merge_context,
```
pdillinger, line 201: Do we really need to pass in the MergeContext here rather than accessing `merge_context_`?

pdillinger (changes requested):

utilities/write_batch_with_index/write_batch_with_index_internal.cc
mrambacher: If there is no column family supplied, you cannot do merge operations because there the merge operator is associated with the column family.  This is the documented behavior in include/rocksdb/utilitiies/write_batch_with_index.h:
 // Similar to previous function but does not require a column_family.
  // Note:  An InvalidArgument status will be returned if there are any Merge
  // operators for this key.  Use previous method instead.
  Status GetFromBatch(const DBOptions& options, const Slice& key,
                      std::string* value) {
    return GetFromBatch(nullptr, options, key, value);
  }

utilities/write_batch_with_index/write_batch_with_index_internal.cc
mrambacher: Yes, there are two signatures of MergeKey, one that takes a MergeContext as a parameter and one that does not but instead uses the internal merge_context_.  The signature requiring a MergeContext is used by the latter, as well as in the case where we do a MultiGet, where the MergeContext needs to be per-key and is not the object internal one.

utilities/write_batch_with_index/write_batch_with_index_internal.cc
mrambacher: This is not a convenience for an internal API but is based on how the external API is written.  Sometimes the external API takes in a ColumnFamily and a DB and sometimes it does not.   The internal workings are based on the external API, which I did not change here.

utilities/write_batch_with_index/write_batch_with_index_internal.h
mrambacher: There are tests in write_batch_with_index_test that use multiple column families.  The "default" versus "non-default" and "convenience" of not specifying a ColumnFamily are not controlled by these internal interfaces but the fact that the public interfaces do not require a ColumnFamily be supplied.

utilities/write_batch_with_index/write_batch_with_index_internal.cc
mrambacher: What do you suggest be used instead?

FWIW, this code has not changed, only moved from one location to another (previously it was in WriteBatchWithIndexInternal::GetFromBatch around line 252-ish).

utilities/write_batch_with_index/write_batch_with_index_internal.cc
pdillinger: Can you point me to that other call site (for MultiGet)? I didn't see it

utilities/write_batch_with_index/write_batch_with_index_internal.cc
pdillinger: Why does this constructor fill in an empty column family and the others leave it nullptr? If the constructors are materially different, there need to be comments clarifying that. These are new constructors.

utilities/write_batch_with_index/write_batch_with_index_internal.cc
mrambacher: write_batch_with_index.cc line 525-ish

utilities/write_batch_with_index/write_batch_with_index_internal.cc
mrambacher: There are three constructors,
- DB + ColumnFamilyHandle
- DBOptions + ColumnFamilyHandle
- ColumnFamilyHandle
For any of the constructors, ColumnFamilyHandle could theoretically be nullptr. When the argument is not specified (e.g. DB for the DBOptions constructor), the unspecified argument could be nullptr.

This could all be done via one constructor taking all three arguments and specifying nullptr for the unused argument but I do not know that helps with code clarity at all.  If you feel that it does, I can make that change.  There is no material difference between the constructors other than not needed to specify nullptr and remember the order of the arguments.

utilities/write_batch_with_index/write_batch_with_index_internal.h
```diff
@@ -440,8 +442,98 @@ class WriteBatchEntryComparator {
...
+ private:
+  uint32_t column_family_id_;
+  WriteBatchEntrySkipList::Iterator skip_list_iter_;
+  const ReadableWriteBatch* write_batch_;
+  WriteBatchEntryComparator* comparator_;
+};
+
 class WriteBatchWithIndexInternal {
  public:
+  explicit WriteBatchWithIndexInternal(DB* db,
```
pdillinger, line 530: Suggest

    // For GetFromBatchAndDb or similar

utilities/write_batch_with_index/write_batch_with_index_internal.h
```diff
@@ -440,8 +442,98 @@ class WriteBatchEntryComparator {
...
+                                       ColumnFamilyHandle* column_family);
+  explicit WriteBatchWithIndexInternal(const DBOptions* db_options,
```
pdillinger, line 532: Suggest

    // For GetFromBatch or similar

utilities/write_batch_with_index/write_batch_with_index_internal.h
```diff
@@ -452,11 +544,25 @@ class WriteBatchWithIndexInternal {
...
-      std::string* value, bool overwrite_key, Status* s);
+  Result GetFromBatch(WriteBatchWithIndex* batch, const Slice& key,
+                      std::string* value, bool overwrite_key, Status* s) {
+    return GetFromBatch(batch, key, &merge_context_, value, overwrite_key, s);
+  }
+  Result GetFromBatch(WriteBatchWithIndex* batch, const Slice& key,
+                      MergeContext* merge_context, std::string* value,
+                      bool overwrite_key, Status* s);
+  Status MergeKey(const Slice& key, const Slice* value, std::string* result,
+                  Slice* result_operand = nullptr) {
```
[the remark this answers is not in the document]
mrambacher: Where is it unused?  The result_operand at line 554 is passed as an argument to the other MergeKey function.  Inside this function the result_operand is passed to the MergeHelper.

pdillinger (approved):
Thanks for the clarification and clean-up

PR #6853, opened 2020-05-15
Re-read the whole request in direct IO mode when IO uring returns partial result

cheng-chang:
If both direct IO and IO uring are enabled, when IO uring returns partial result, we'll try to read the remaining part of the request, but the starting address/offset of the remaining part might not be aligned to the block size, in direct IO mode, the unaligned offset causes bug.

Test Plan:
run make check with both direct IO and IO uring enabled, this is covered by one of the continuous tests.

### DIFF
--- a/env/io_posix.cc
+++ b/env/io_posix.cc
@@ -711,13 +711,19 @@ IOStatus PosixRandomAccessFile::MultiRead(FSReadRequest* reqs,
           // comment
           // https://github.com/facebook/rocksdb/pull/6441#issuecomment-589843435
           // Fall back to pread in this case.
-          Slice tmp_slice;
-          req->status =
-              Read(req->offset + req_wrap->finished_len,
-                   req->len - req_wrap->finished_len, options, &tmp_slice,
-                   req->scratch + req_wrap->finished_len, dbg);
-          req->result =
-              Slice(req->scratch, req_wrap->finished_len + tmp_slice.size());
+          if (use_direct_io()) {
+            // offset must be aligned, so read from the original offset again.
+            req->status = Read(req->offset, req->len, options, &req->result,
+                               req->scratch, dbg);
+          } else {
+            Slice tmp_slice;
+            req->status =
+                Read(req->offset + req_wrap->finished_len,
+                     req->len - req_wrap->finished_len, options, &tmp_slice,
+                     req->scratch + req_wrap->finished_len, dbg);
+            req->result =
+                Slice(req->scratch, req_wrap->finished_len + tmp_slice.size());
+          }
         } else if (bytes_read < req_wrap->iov.iov_len) {
           assert(bytes_read > 0);
           assert(bytes_read + req_wrap->finished_len < req->len);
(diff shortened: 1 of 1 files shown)


env/io_posix.cc
```diff
@@ -711,13 +711,19 @@ IOStatus PosixRandomAccessFile::MultiRead(FSReadRequest* reqs,
...
-          Slice tmp_slice;
-          req->status =
-              Read(req->offset + req_wrap->finished_len,
-                   req->len - req_wrap->finished_len, options, &tmp_slice,
-                   req->scratch + req_wrap->finished_len, dbg);
-          req->result =
-              Slice(req->scratch, req_wrap->finished_len + tmp_slice.size());
+          if (use_direct_io()) {
+            // offset must be aligned, so read from the original offset again.
+            req->status = Read(req->offset, req->len, options, &req->result,
```
anand1976, line 716: With direct IO, I think this might be a more common occurrence, since if we're reading a block close to the end of the file, rounding up the size may extend the IO past the file. Why not just complete the request here instead of calling Read? Otherwise, we'll end up making 4 system calls overall for one IO, 2 in MultiRead and 2 in read.

cheng-chang: According to https://github.com/facebook/rocksdb/pull/6441#issuecomment-589843435, it might return partial result, even if the block is not near the end of file. So if we just complete the request, then we might have remaining bytes not read. If we can complete the request here, then logically we should be able to complete the request in non-direct IO mode too.

cheng-chang: Added a check for whether it's EOF.

env/io_posix.cc
```diff
@@ -711,13 +711,30 @@ IOStatus PosixRandomAccessFile::MultiRead(FSReadRequest* reqs,
...
-          Slice tmp_slice;
-          req->status =
-              Read(req->offset + req_wrap->finished_len,
-                   req->len - req_wrap->finished_len, options, &tmp_slice,
-                   req->scratch + req_wrap->finished_len, dbg);
-          req->result =
-              Slice(req->scratch, req_wrap->finished_len + tmp_slice.size());
+          if (use_direct_io()) {
+            if (req_wrap->finished_len %
+                    static_cast<size_t>(GetRequiredBufferAlignment()) !=
```
anand1976, line 716: The alignment is nearly always going to be a power of 2. Can we use bitwise operations like here - https://github.com/facebook/rocksdb/blob/0a0151fb99b306dc0b1f6b22df8073f8084da795/util/aligned_buffer.h#L24?

env/io_posix.cc
```diff
@@ -711,13 +711,30 @@ IOStatus PosixRandomAccessFile::MultiRead(FSReadRequest* reqs,
...
+                0) {
+              // Bytes reads don't fill sectors. Should only happen at the end
+              // of the file.
+              req->result = Slice(req->scratch, req_wrap->finished_len);
+              req->status = IOStatus::OK();
+            } else {
+              // Returns partial result, need to re-read with pread.
+              // Since offset must be aligned,
+              // read from the original offset again.
+              req->status = Read(req->offset, req->len, options, &req->result,
```
anand1976, line 726: In this else case, ```finished_len``` would be aligned right? So it should be ok to call Read with ```req->offset + req_wrap->finished_len```? Not a big deal, but I think it would make this whole block simpler.

PR #6808, opened 2020-05-05
Update googletest from 1.8.1 to 1.10.0

adamretter:

PR #6858, opened 2020-05-19
Use in-repo gtest in buck build

pdillinger:
Summary: ... so that we have freedom to upgrade it (see #6808).

As a side benefit, gtest will no longer be linked into main library in
buck build.

Test Plan: fb internal build & link


TARGETS
```diff
@@ -114,6 +109,15 @@ ROCKSDB_LIB_DEPS = [
     ":rocksdb_test_lib",
 ] if not is_opt_mode else [":rocksdb_lib"]
 
+# Add a directory for #include
```
riversand963, line 112: How about moving these closer to line 60?

TARGETS
```diff
@@ -446,6 +453,19 @@ cpp_library(
     external_deps = ROCKSDB_EXTERNAL_DEPS,
 )
 
+cpp_library(
+    name = "rocksdb_third_party_gtest",
+    srcs = ["third-party/gtest-1.8.1/fused-src/gtest/gtest-all.cc"],
+    headers = [u'third-party/gtest-1.8.1/fused-src/gtest/gtest.h'],
```
riversand963, line 459: I think `u'<string>'` is not allowed.

pdillinger: > Thanks @pdillinger for the PR. Mostly LGTM. Could you run `python buckifier/buckify_rocksdb.py`?

Considering I've modified the generation code and there are corresponding changes to TARGETS, I obviously did, but who knows what's different in the GitHub build. Maybe it's just the problem with the existing implementation of add_library generating a u string for headers.

PR #6859, opened 2020-05-20
Generate file checksum in SstFileWriter

zhichao-cao:
If Option.file_checksum_gen_factory is set, rocksdb generates the file checksum during flush and compaction based on the checksum generator created by the factory and store the checksum and function name in vstorage and Manifest. 

This PR enable file checksum generation in SstFileWrite and store the checksum and checksum function name in the  ExternalSstFileInfo, such that application can use them for other purpose, for example, ingest the file checksum with files in IngestExternalFile().

Test plan: add unit test and pass make asan_check.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -18,6 +18,7 @@
 
 ### New Feature
 * sst_dump to add a new --readahead_size argument. Users can specify read size when scanning the data. Sst_dump also tries to prefetch tail part of the SST files so usually some number of I/Os are saved there too.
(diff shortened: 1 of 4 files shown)


zhichao-cao:
> LGTM!
> 
> > such that application can use them for other purpose.
> 
> I guess this other purpose is providing them as arguments to `IngestExternalFile()`?

Thanks for the review! Yes, IngestExternalFile(). Ingest file with checksum will be in the next PR...

zhichao-cao:
> LGTM.

Thanks!

PR #6862, opened 2020-05-20
Make it possible to look up files by number in VersionStorageInfo

ltamasi:
Summary:
Does what it says on the can: the patch adds a hash map to `VersionStorageInfo`
that maps file numbers to file locations, i.e. (level, position in level) pairs. This
will enable stricter consistency checks in `VersionBuilder`. The patch also fixes
all the unit tests that used duplicate file numbers in a version (which would trigger
an assertion with the new code).

Test Plan:
`make check`
`make whitebox_crash_test`

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -2621,10 +2621,9 @@ void VersionStorageInfo::AddFile(int level, FileMetaData* f, Logger* info_log) {
 
   const uint64_t file_number = f->fd.GetNumber();
 
-  FileLocation& file_location = file_locations_[file_number];
-  assert(!file_location.IsValid());
-
-  file_location = FileLocation(level, level_files->size() - 1);
+  assert(file_locations_.find(file_number) == file_locations_.end());
+  file_locations_.emplace(FileLocations::value_type(
+      file_number, FileLocation(level, level_files->size() - 1)));
 }
 
 void VersionStorageInfo::AddBlobFile(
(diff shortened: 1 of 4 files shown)


db/version_set.h
```diff
@@ -283,6 +283,47 @@ class VersionStorageInfo {
...
+      return level_ == rhs.level_ && position_ == rhs.position_;
+    }
+
+    bool operator!=(const FileLocation& rhs) const { return !(*this == rhs); }
+
+    static FileLocation Invalid() { return FileLocation(); }
+
+   private:
+    int level_ = -1;
+    size_t position_ = 0;
```
riversand963, line 307: Not important. Can these two members be const?

db/version_set.cc
```diff
@@ -2618,6 +2618,13 @@ void VersionStorageInfo::AddFile(int level, FileMetaData* f, Logger* info_log) {
 #endif
   f->refs++;
   level_files->push_back(f);
+
+  const uint64_t file_number = f->fd.GetNumber();
+
+  FileLocation& file_location = file_locations_[file_number];
+  assert(!file_location.IsValid());
```
riversand963, line 2625: Should we assert `file_locations_.find(file_number) == file_locations_.end()`?

ltamasi: Sure, that also works. Will change.

db/version_set.h
ltamasi: I'd rather keep them as-is; making the members const would mean that the type becomes non-assignable.

PR #6864, opened 2020-05-21
Update format check script

riversand963:
Update format check script so that unformatted code can be printed in the log of GH Action.

Test plan:
Watch for GH Action.


pdillinger:
Should be obsolete by https://github.com/facebook/rocksdb/pull/6861. See e.g. https://github.com/pdillinger/rocksdb/runs/693921556

PR #6870, opened 2020-05-22
Fix/expand ASSERT_STATUS_CHECKED build, add to Travis

pdillinger:
Summary: Fixed some option handling code that recently broke the
ASSERT_STATUS_CHECKED build for options_test.

Added all other existing tests that pass under ASSERT_STATUS_CHECKED to
the whitelist.

Added a Travis configuration to run all whitelisted tests with
ASSERT_STATUS_CHECKED. (Someday we might enable this check by default in
debug builds.)

Test Plan: ASSERT_STATUS_CHECKED=1 make check, Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ env:
   - JOB_NAME=cmake-gcc9 # 3-5 minutes
   - JOB_NAME=cmake-gcc9-c++20 # 3-5 minutes
   - JOB_NAME=cmake-mingw # 3 minutes
+  - JOB_NAME=status_checked
 
 matrix:
   exclude:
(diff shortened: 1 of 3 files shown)


pdillinger:
The new Travis status_checked build failed https://travis-ci.org/github/facebook/rocksdb/jobs/690119387, but @ajkr landed this with accept2ship.

I think the problem is that it's up to the compiler whether

    Status s;
    ... // initial value of s not read
    if (x) {
      s = something;
    } else {
      s = something else;
    }
    check(s);

constructs and destructs a Status that goes unchecked. I think debug builds need to disallow any reading from default-initialized statuses, thus (in whitelisted code) forcing explicit initialization to OK() if a status might be read.

pdillinger:
> I think the problem is that it's up to the compiler whether

Not so sure about that. We'll see in #6871, where we now have stack traces on failures 😄

ajkr:
> The new Travis status_checked build failed https://travis-ci.org/github/facebook/rocksdb/jobs/690119387, but @ajkr landed this with accept2ship.

Sorry, didn't know what checks were in place with that button. Seems not the ones we want.

I think we need to add a check for `status` at the top of this loop body: https://github.com/facebook/rocksdb/blob/35a25a3fb9b5a61a389e853f8be37932487edd03/options/options_helper.cc#L1126-L1135. I haven't figured out why it's compiler dependent...

ajkr:
Ha, I think the order of the unordered_map changed between compiler versions. When I run in older versions, the `status` is non-ok on the second iteration. In newer versions the `status` is non-ok on the first iteration (and we check `status` just above the loop).

PR #6875, opened 2020-05-23
Fix invalid xxhash64 usage

edouarda:
The size of the input is cast to 32-bit, which not only truncates the data to hash, thus not properly hashing all the content if the data is more than 4 GiB, but it can also be cast to a signed 32-bit integer, which will result in an integer overflow if the file is larger than 2 GiB.

xxhash has a similar problem, however there is no simple fix. I suggest deprecating xxhash in favor of xxhash64.


table/block_based/reader_common.cc
```diff
@@ -33,7 +33,7 @@ Status VerifyChecksum(const ChecksumType type, const char* buf, size_t len,
       actual = XXH32(buf, static_cast<int>(len), 0);
       break;
     case kxxHash64:
-      actual = static_cast<uint32_t>(XXH64(buf, static_cast<int>(len), 0) &
+      actual = static_cast<uint32_t>(XXH64(buf, len, 0) &
```
pdillinger, line 36: Not a big deal, but we now have a function Lower32of64() that better captures the intent of the static cast on the return value.

table/block_fetcher.cc
```diff
@@ -49,7 +49,7 @@ inline void BlockFetcher::CheckBlockChecksum() {
         break;
       case kxxHash64:
         actual = static_cast<uint32_t>(
```
pdillinger, line 51: Same

pdillinger:
> LGTM. Do you mind changing the branch to have just this one commit ahead of facebook/rocksdb master? The "Commits" tab shows 531 commits, which is somewhat confusing.

Will not matter since we collapse each PR to single commit in a linear history.

table/block_based/reader_common.cc
pdillinger: ... which I'm aware you didn't write. Thanks for the fix!

pdillinger:
The avalanche (e.g. XXH64_avalanche) might help for trimming to full entropy 32 bits for RocksDB checksum, from 64 bits, though I'm doubtful there's a meaningful difference vs. taking top 32 bits after the multiplication that seems to precede the avalanche in XXH64 and XXH3. Hmm.

Cyan4973:
Skipping the final avalanche will not impact the collision ratio, as long as the entire hash value is stored and used in the comparison.

However, as soon as a part of the hash is used for anything, be it for sorting, dispatching, grouping into buckets, bit index, well whatever else, the absence of avalanche is likely going to create a distribution imbalance.
So it really depends on what usages depend on this hash.

Now, extracting 32-bit from a 64-bit hash : well, that's indeed using a part of the hash, so I would expect an avalanche to make a difference, unless the source is already well randomized.

I understand that, even without the final avalanche, the intermediate accumulator is already mixed and shuffled quite a bit. So it's tempting to consider that it might be good enough. But if it was, I would probably not have added this last avalanche step.

Now, it could be that strict randomness requirements introduced by avalanche quality tests are a bit overkill for some real-world applications, that would work well enough even with a bit less shuffling. Maybe. This is now application-dependent unfortunately, and will have to be tested. 
Plus, there is always a risk that, in some future, the application change, with a new contributor trying to employ the hash for some unforeseen usage.

edouarda:
> Actually, this is a schema change. 

Yes, but anyone using blocks greater than 4 GiB currently crashes, so it doesn't matter.

Blocks greater than 4 GiB can happen in degenerate cases where the user would write very large keys.

pdillinger:
> Yes, but anyone using blocks greater than 4 GiB currently crashes, so it doesn't matter.

SSTFileWriter can generate them, but checksum check fails as I explain in #6978

PR #6877, opened 2020-05-24
Add a new int64 subtract merge operator

adamretter:
This merge operator subtracts `uint64_t` values. My use-case for this is testing the Merge operator on `WBWI` and `BaseDeltaIterator`. This merge operator has an advantage over the other available merge operators, in that a Get after a Merge will not return the same value that was passed to the Merge if the merge operator was executed. The same is not true of the other merge operators, for example:

UInt64AddOperator -> WBWI.Merge(K1, 19) -> WBWI.Get(K1) == 19. However you cannot tell if the merge operator was called or not... If the merge operator was called, the value will be 19, if it was not called, the value is also 19!

At a cursory glance, this may appear similar to `UInt64AddOperator`, but this merge operator handles signed numbers.

This merge operator may also add value for general use by others that want a decrementing counter or similar. However, that is not my use-case. I need this for further testing and making improvements to Merge in WBWI scenarios.

### DIFF
--- a/utilities/merge_operators/int64subtract.cc
+++ b/utilities/merge_operators/int64subtract.cc
@@ -1,0 +1,65 @@
+// Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#include <memory>
+
+#include "logging/logging.h"
+#include "rocksdb/env.h"
+#include "rocksdb/merge_operator.h"
+#include "rocksdb/slice.h"
+#include "util/coding.h"
+#include "utilities/merge_operators.h"
+
+using namespace ROCKSDB_NAMESPACE;
+
+namespace {  // anonymous namespace
+
+// A 'model' merge operator with int64 subtraction semantics
+// operands and database value should be variable length encoded
+// int64_t values, as encoded/decoded by `util/coding.h`.
+class Int64SubtractMergeOperator : public AssociativeMergeOperator {
+ public:
+  bool Merge(const Slice&, const Slice* existing_value, const Slice& value,
+             std::string* new_value, Logger* logger) const override {
+    int64_t orig_value = 0;
+    if (existing_value) {
+      Slice ev(*existing_value);
+      if (!GetVarsignedint64(&ev, &orig_value)) {
+        ROCKS_LOG_ERROR(logger,
+                        "int64 value corruption, size: %" ROCKSDB_PRIszt,
+                        existing_value->size());
+        return false;
+      }
+    }
+
+    int64_t operand = 0;
+    Slice v(value);
+    if (!GetVarsignedint64(&v, &operand)) {
+      ROCKS_LOG_ERROR(logger,
+                      "int64 operand corruption, size: %" ROCKSDB_PRIszt,
+                      value.size());
+      return false;
+    }
+
+    assert(new_value);
+    new_value->clear();
+    const int64_t new_number = orig_value - operand;
+    PutVarsignedint64(new_value, new_number);
+
+    return true;
+  }
+
+  const char* Name() const override { return "Int64SubtractMergeOperator"; }
+};
+
+}  // namespace
+
+namespace ROCKSDB_NAMESPACE {
+
+std::shared_ptr<MergeOperator> MergeOperators::CreateInt64SubtractOperator() {
+  return std::make_shared<Int64SubtractMergeOperator>();
+}
+
+}  // namespace ROCKSDB_NAMESPACE
--- a/utilities/merge_operators/int64subtract_test.cc
+++ b/utilities/merge_operators/int64subtract_test.cc
@@ -1,0 +1,257 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+//
+// Copyright (c) 2011 The LevelDB Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file. See the AUTHORS file for names of contributors.
+
+#ifndef ROCKSDB_LITE
+
+#include "port/stack_trace.h"
+#include "rocksdb/db.h"
+#include "test_util/testharness.h"
+#include "util/coding.h"
+#include "utilities/merge_operators.h"
+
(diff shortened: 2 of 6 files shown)


adamretter:
This could perhaps be refactored from `Int64SubtractOperator` to `Int64AddOperator`. Passing negative numbers to it would then have the effect of doing a subtraction, at the moment they cause an addition.

utilities/merge_operators/int64subtract.cc
```diff
@@ -0,0 +1,65 @@
...
+#include "utilities/merge_operators.h"
+
+using namespace ROCKSDB_NAMESPACE;
+
+namespace {  // anonymous namespace
+
+// A 'model' merge operator with int64 subtraction semantics
+// operands and database value should be variable length encoded
+// int64_t values, as encoded/decoded by `util/coding.h`.
+class Int64SubtractMergeOperator : public AssociativeMergeOperator {
```
ajkr, line 22: It shouldn't inherit from `AssociativeMergeOperator` because subtraction is non-associative, e.g., `5 - (4 - 3) != (5 - 4) - 3`.

utilities/merge_operators/int64subtract_test.cc
```diff
@@ -0,0 +1,257 @@
...
+                        // db, and then merge subtracted merge_num.
+}
+
+TEST_F(Int64SubtractMergeOperatorTest, MergeMultipleValues) {
+  std::string value;
+
+  ASSERT_OK(OpenDB());
+
+  PutVarsignedint64(&value, 123);
+  Status s = db_->Merge(write_opts_, "key", value);  // Merging 123 under key
```
ajkr, line 202: I think we should treat the first merge same as a base put. If we don't, we risk getting inconsistent results because compacting the earliest merge to the bottommost level will convert it to a put. Then, this will count as +123 if the compaction to bottommost level has already happened, or -123 if the compaction to bottommost level has not yet happened.

utilities/merge_operators/int64subtract.cc
pdillinger: Or closer to the implementation:

    ((/*implicit*/ 0 - 5) - 4) - 3

!=

    (/*implicit*/ 0 - 5) - ((/*implicit*/ 0 - 4) - 3)

For an operator to be associative, you have to be able to "bootstrap" from any input element without messing up the merge semantics. I think it's completely misleading, if not wrong, that the operands to AssociativeMergeOperator::Merge are called `existing_value` and `value` rather than `left` (optional) and `right`.

adamretter:
@ajkr when you say:

> a more natural way of expressing, e.g., 5 - 4 -3 would be using the regular addition operator, and calling Merge("key", "5"), Merge("key", "-4"), Merge("key", "-3").

By `regular addition operator` do you mean the `UInt64AddOperator`? I can understand that, but an advantage of this one is that you can store negative numbers. I think it best if I refactor mine from a `Int64SubtractOperator` to a `Int64AddOperator`. Do you agree?

adamretter:
@ajkr Oops! Yes... I agree that it should not inherit from `AssociativeMergeOperator`. I will fix that.

utilities/merge_operators/int64subtract.cc
adamretter: @pdillinger Good point. Should I send a separate PR to rename the existing ones to `left` and `right`?

utilities/merge_operators/int64subtract_test.cc
adamretter: > the earliest merge to the bottommost level will convert it to a put

@ajkr This is something I would like to understand and discuss in more detail...

Are you saying that if I do a single `Merge(K1, V1)`, and there is no other preceding op for `K1`. Then at compaction time this will be rewritten to `Put(K1, V1)`.

If that is the case, I am very surprised as it is not mentioned in the documentation: https://github.com/facebook/rocksdb/wiki/Merge-Operator, and there are three places in the API docs that, at least to me, seem to imply the opposite:

1. https://github.com/facebook/rocksdb/blob/master/include/rocksdb/merge_operator.h#L57
2. https://github.com/facebook/rocksdb/blob/master/include/rocksdb/merge_operator.h#L126
3. https://github.com/facebook/rocksdb/blob/master/include/rocksdb/merge_operator.h#L208

In addition that would seem to violate the "read-modify-write" pattern that the MergeOperator claims to implement.

I guess I am not understanding something correctly here... can you clarify?

utilities/merge_operators/int64subtract.cc
pdillinger: @adamretter I'll take it to put in my preferred wording ;)

utilities/merge_operators/int64subtract_test.cc
ajkr: > Are you saying that - If I do a single Merge(K1, V1), and there is no other preceding op for K1... then at compaction time this will be rewritten to Put(K1, V1)?

I was, but was actually mistaken. https://github.com/facebook/rocksdb/blob/fdf882ded218344c136c97daf76dfb59e4bc155f/db/merge_helper.cc#L316-L324 appears to work the way you expect. The earliest merge operand will be merged against null base value (aka `existing_value`) and the result will be written as a Put, which I believe is what you expect.

ajkr:
> @ajkr when you say:
> 
> > a more natural way of expressing, e.g., 5 - 4 -3 would be using the regular addition operator, and calling Merge("key", "5"), Merge("key", "-4"), Merge("key", "-3").
> 
> By `regular addition operator` do you mean the `UInt64AddOperator`? I can understand that, but an advantage of this Merge Operator is that you can store negative numbers. I think it best if I refactor mine from a `Int64SubtractOperator` to a `Int64AddOperator`. Do you agree?

`Int64AddOperator` sounds fine with me, but then again, it feels like it might defeat your original purpose. My understanding with `Int64AddOperator` is we won't be able to tell whether a single operand went through the merge operator as a `Get()` would expect the same result either way. I think @pdillinger's whitespace trimming suggestion is promising to achieve your original goal.

utilities/merge_operators/int64subtract_test.cc
adamretter: @ajkr Okay cool! Thanks for checking that, I was beginning to doubt my sanity ;-)

adamretter:
> My understanding with Int64AddOperator is we won't be able to tell whether a single operand went through the merge operator as a Get() would expect the same result either way

@ajkr Indeed :-( I will close this PR and open two new ones. One to add the `Int64AddOperator` and then another for another merge operator that will give me what I need for testing.

pdillinger:
> One to add Int64AddOpterator

Also consider my suggested work-around: https://groups.google.com/d/msg/rocksdb/YZYp6kxXZVU/EVGFpmXLAwAJ

PR #6878, opened 2020-05-25
Add a file system parameter: --fs_uri to db_stress and db_bench

yhr:
Summary:

This pull request adds the parameter --fs_uri to db_bench and db_stress, creating a composite env combining the default env with a specified registered rocksdb file system.

This makes it easier to develop and test new RocksDB FileSystems.

The pull request also registers the posix file system for testing purposes.

Examples:
```
$./db_bench --fs_uri=posix:// --benchmarks=fillseq

$./db_stress --fs_uri=zenfs://nullb1
```

zenfs is a RocksDB FileSystem I'm developing to add support for zoned block devices, and in that case the zoned block device is specified in the uri (a zoned null block device in the above example).

### DIFF
--- a/db_stress_tool/db_stress_tool.cc
+++ b/db_stress_tool/db_stress_tool.cc
@@ -34,6 +34,11 @@ static std::shared_ptr<ROCKSDB_NAMESPACE::DbStressEnvWrapper> env_wrapper_guard;
 static std::shared_ptr<CompositeEnvWrapper> fault_env_guard;
 }  // namespace
 
+static Env* GetCompositeEnv(std::shared_ptr<FileSystem> fs) {
+  static std::shared_ptr<Env> composite_env = NewCompositeEnv(fs);
+  return composite_env.get();
+}
+
 KeyGenContext key_gen_ctx;
 
 int db_stress_tool(int argc, char** argv) {
@@ -63,11 +68,15 @@ int db_stress_tool(int argc, char** argv) {
 
   Env* raw_env;
 
+  int env_opts =
+      !FLAGS_hdfs.empty() + !FLAGS_env_uri.empty() + !FLAGS_fs_uri.empty();
+  if (env_opts > 1) {
+    fprintf(stderr,
+            "Error: --hdfs, --env_uri and --fs_uri are mutually exclusive\n");
+    exit(1);
+  }
+
   if (!FLAGS_hdfs.empty()) {
-    if (!FLAGS_env_uri.empty()) {
-      fprintf(stderr, "Cannot specify both --hdfs and --env_uri.\n");
-      exit(1);
-    }
     raw_env = new ROCKSDB_NAMESPACE::HdfsEnv(FLAGS_hdfs);
   } else if (!FLAGS_env_uri.empty()) {
     Status s = Env::LoadEnv(FLAGS_env_uri, &raw_env, &env_guard);
@@ -75,6 +84,14 @@ int db_stress_tool(int argc, char** argv) {
       fprintf(stderr, "No Env registered for URI: %s\n", FLAGS_env_uri.c_str());
       exit(1);
     }
+  } else if (!FLAGS_fs_uri.empty()) {
+    std::shared_ptr<FileSystem> fs;
+    Status s = FileSystem::Load(FLAGS_fs_uri, &fs);
+    if (!s.ok()) {
+      fprintf(stderr, "Error: %s\n", s.ToString().c_str());
+      exit(1);
+    }
+    raw_env = GetCompositeEnv(fs);
   } else {
     raw_env = Env::Default();
   }
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -952,11 +952,17 @@ static bool ValidateTableCacheNumshardbits(const char* flagname,
 DEFINE_int32(table_cache_numshardbits, 4, "");
 
 #ifndef ROCKSDB_LITE
-DEFINE_string(env_uri, "", "URI for registry Env lookup. Mutually exclusive"
-              " with --hdfs.");
+DEFINE_string(env_uri, "",
+              "URI for registry Env lookup. Mutually exclusive"
+              " with --hdfs and --fs_uri");
+DEFINE_string(fs_uri, "",
+              "URI for registry Filesystem lookup. Mutually exclusive"
+              " with --hdfs and --env_uri."
+              " Creates a default environment with the specified filesystem.");
 #endif  // ROCKSDB_LITE
-DEFINE_string(hdfs, "", "Name of hdfs environment. Mutually exclusive with"
-              " --env_uri.");
+DEFINE_string(hdfs, "",
+              "Name of hdfs environment. Mutually exclusive with"
+              " --env_uri and --fs_uri");
 
 static std::shared_ptr<ROCKSDB_NAMESPACE::Env> env_guard;
 
@@ -985,6 +991,15 @@ DEFINE_int32(thread_status_per_interval, 0,
 
 DEFINE_int32(perf_level, ROCKSDB_NAMESPACE::PerfLevel::kDisable,
              "Level of perf collection");
+
+#ifndef ROCKSDB_LITE
+static ROCKSDB_NAMESPACE::Env* GetCompositeEnv(
+    std::shared_ptr<ROCKSDB_NAMESPACE::FileSystem> fs) {
+  static std::shared_ptr<ROCKSDB_NAMESPACE::Env> composite_env =
+      ROCKSDB_NAMESPACE::NewCompositeEnv(fs);
+  return composite_env.get();
+}
+#endif
 
 static bool ValidateRateLimit(const char* flagname, double value) {
   const double EPSILON = 1e-10;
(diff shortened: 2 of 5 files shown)


yhr:
Is that failing CI test for visual studio 2014 anything to worry about? It does not seem related to my patch.

env/env_posix.cc
```diff
@@ -528,4 +532,12 @@ std::unique_ptr<Env> NewCompositeEnv(std::shared_ptr<FileSystem> fs) {
   return std::unique_ptr<Env>(new PosixEnv(default_env, fs));
 }
 
+std::unique_ptr<Env> NewCompositeEnv(std::shared_ptr<FileSystem> fs,
```
mrambacher, line 535: This seems off to me.  Why are we potentially creating two Envs in order to create/return one?

db_stress_tool/db_stress_tool.cc
```diff
@@ -62,18 +64,36 @@ int db_stress_tool(int argc, char** argv) {
...
-    }
     raw_env = new ROCKSDB_NAMESPACE::HdfsEnv(FLAGS_hdfs);
   } else if (!FLAGS_env_uri.empty()) {
     Status s = Env::LoadEnv(FLAGS_env_uri, &raw_env, &env_guard);
     if (raw_env == nullptr) {
       fprintf(stderr, "No Env registered for URI: %s\n", FLAGS_env_uri.c_str());
       exit(1);
     }
+  } else if (!FLAGS_fs_uri.empty()) {
+#ifndef OS_WIN
```
mrambacher, line 84: Why is this in an OS_WIN block?  Can't we load FileSystems on Windows?

tools/db_bench_tool.cc
```diff
@@ -7032,15 +7040,34 @@ int db_bench_tool(int argc, char** argv) {
...
+  }
+
+  if (!FLAGS_env_uri.empty()) {
     Status s = Env::LoadEnv(FLAGS_env_uri, &FLAGS_env, &env_guard);
     if (FLAGS_env == nullptr) {
       fprintf(stderr, "No Env registered for URI: %s\n", FLAGS_env_uri.c_str());
       exit(1);
     }
+  } else if (!FLAGS_fs_uri.empty()) {
+#ifndef OS_WIN
```
mrambacher, line 7058: Why the OS_WIN block?

env/env_posix.cc
yhr: The idea is that we create a base env that can be shared among composite envs, and that we can control the order of destruction of the base env and its dependent composite envs in a separate c-file.

db_stress_tool/db_stress_tool.cc
yhr: See the comment in the source below: - // NewCompositeEnv is not available in Windows
Linking would fail on windows if we would preprocess the call out.

tools/db_bench_tool.cc
yhr: See the comment in the source below: - // NewCompositeEnv is not available in Windows
Linking would fail on windows if we would preprocess the call out.

db_stress_tool/db_stress_tool.cc
yhr: Looking over this again, I found that there is  is an implementation of NewCompositeEnv for OS_WIN in env/env.cc. I'll add support for the new variant there as well, and db_bench/db_stress should compile just fine for windows.

yhr: Implemented NewCompositeEnv for windows and removed the ifdef.

tools/db_bench_tool.cc
yhr: Implemented NewCompositeEnv for windows and removed the ifdef.

env/env.cc
```diff
@@ -476,6 +476,13 @@ const std::shared_ptr<FileSystem>& Env::GetFileSystem() const {
 std::unique_ptr<Env> NewCompositeEnv(std::shared_ptr<FileSystem> fs) {
   return std::unique_ptr<Env>(new CompositeEnvWrapper(Env::Default(), fs));
 }
+
+std::unique_ptr<Env> NewCompositeEnv(std::shared_ptr<FileSystem> fs,
+                                     std::shared_ptr<Env>* shared_env) {
+  assert(shared_env != nullptr);
+  if (!*shared_env) shared_env->reset(Env::Default());
```
ajkr, line 483: Won't this cause `Env::Default()` to be double deleted (once when the last `shared_ptr` is destroyed, again when the static `Env` it points to goes out of scope?

db_stress_tool/db_stress_tool.cc
```diff
@@ -29,6 +29,8 @@
 
 namespace ROCKSDB_NAMESPACE {
 namespace {
+static std::shared_ptr<ROCKSDB_NAMESPACE::FileSystem> fs_guard;
```
ajkr, line 32: If we introduce getter functions for these, we can then make them static local variables. By calling `Env::Default()` before initializing our static local variables, we make sure the `Env` on which they depend is destroyed later.

env/env.cc
yhr: The windows implementation does this (port/win/env_default.cc) : `std::call_once(winenv_once_flag, []() { envptr = new WinEnv(); });`

So it should work as long as Env::Default() is only used for NewComposite envs, right? 

Maybe a better solution would be to add a new Env:Default() that returns a shared pointer, avoiding the out-of-scope-racing entirely?

Thanks for reviewing!

db_stress_tool/db_stress_tool.cc
yhr: Ah, yes. sounds like a better solution. I'll revise the patch.

env/env.cc
ajkr: > So it should work as long as Env::Default() is only used for NewComposite envs, right?

For Windows yes, but not for Linux as its `Env::Default()` uses a function-local static: https://github.com/facebook/rocksdb/blob/722ebba834d550ec03d481ecbb7c1e7857cb9fb7/env/env_posix.cc#L508-L509.

Can we call the `NewCompositeEnv(std::shared_ptr<FileSystem> fs)` overload in case of `!*shared_env`? It seems we do not need a `shared_ptr` to guard the `CompositeEnvWrapper`'s `Env` when it is `Env::Default()`.

db_stress_tool/db_stress_tool.cc
yhr: I've added a getter now, GetCompositeBaseEnv

env/env.cc
yhr: I've now removed the new NewCompositeEnv overload, and added a getter in stead. This allows for composite envs to share thread pools and for the base env to join those threads before the composite envs are destroyed(avoiding the race) - without destroying Env::Default(). Does this make sense?

yhr: Ping @ajkr , could you have a look at the new revision of the branch? Looks neater to me.  Cheers!

db_stress_tool/db_stress_tool.cc
```diff
@@ -29,6 +29,8 @@
...
+static std::shared_ptr<ROCKSDB_NAMESPACE::Env> composite_env_guard;
```
ajkr, line 33: I think `composite_env_guard` depends on `env_guard` for enforcing its dependencies outlive it when `--fs_uri` is provided. Then `env_guard` should be defined above `composite_env_guard` to ensure it is destroyed after.

But actually, this problem only arises because we have `composite_env_guard` as a non-local static. If instead it were a static local variable (maybe in a `GetCompositeEnv()` which could call `FileSystem::Load()` and `NewCompositeEnv()` on first invocation) , the proper destruction order should be enforced without any guards.

yhr: Ping @anand1976  , since you did the refactoring of the env interface and the file system split, do you have time to comment on this? 

Thanks!

anand1976: Its not clear to me why there would be threads running with a dangling reference back to the composite env. The app should be responsible for cancelling/waiting for any such background jobs before destroying the composite env. IMO as @ajkr suggested, we should be able to do without ```GetCompositeBaseEnv``` and use a getter with a static variable instead of ```composite_env_guard```.

env/env_posix.cc
```diff
@@ -509,9 +513,17 @@ Env* Env::Default() {
   return &default_env;
 }
 
+std::shared_ptr<Env> GetCompositeBaseEnv() {
+  static std::shared_ptr<PosixEnv> shared_env;
+  // Call Env::Default to make sure the thread singletons have been setup
+  (void)Env::Default();
+  if (!shared_env) shared_env.reset(new PosixEnv());
```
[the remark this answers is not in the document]
yhr: I had to create a new PosixEnv instance(along with new thread pools) to allow it to be destroyed. The destruction was needed to join the threads before the composite env was destroyed. Since Env::Default is not reference counted it can't be destroyed safely.

db_stress_tool/db_stress_tool.cc
yhr: I wanted to preserve the same behavior as for normal envs to avoid having new weird segfaults crop up when NewCompositeEnv would be used. What I saw(it i remember correctly) was that if something unexpected happens(like running out of disk space) with background work running (like compaction), a thread might continue running for some time with a reference back to the env it is operating in while the app(db_bench) is exiting. The only way(without changing the app code) is to safely destroy that reference is to destroy it's base env, joining all threads.  

Does that make sense as a background?

If the app is doing it wrong, then a getter with a static variable is much more straightforward and I can just fix any dangling-reference/unclean shutdown issues as I discover them.

Cheers!

yhr:
@ajkr, @anand1976 : i've now introduced static getters in db_bench / db_stress in stead. Is this what you had in mind?

Cheers!

db_stress_tool/db_stress_tool.cc
ajkr: Understood the concern now. The way it's supposed to work in a clean shutdown is the app calls `DB::Close()` or delete `DB` on all its DBs. The DB shutdown process involves waiting for all its jobs in the background thread pool to finish (here, finish usually just means failing quickly due to noticing the `shutting_down_` flag is set). Following that, the thread pool should no longer have any opportunity to reference the composite `Env`.

PR #6879, opened 2020-05-26
Fix potential memory leak of scratch buffer

cheng-chang:
If `req.scratch` is an internally allocated buffer, but `raw_block_contents` is not constructed to own `req.scratch`, then `req.scratch` will be leaked.

Test Plan:
make asan_check

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -1717,7 +1717,8 @@ void BlockBasedTable::RetrieveMultipleBlocks(
     FSReadRequest& req = read_reqs[req_idx];
     Status s = req.status;
     if (s.ok()) {
-      if (req.result.size() != req.len) {
+      if ((req.result.size() != req.len) ||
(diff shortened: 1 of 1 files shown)


table/block_based/block_based_table_reader.cc
```diff
@@ -1717,7 +1717,8 @@ void BlockBasedTable::RetrieveMultipleBlocks(
     FSReadRequest& req = read_reqs[req_idx];
     Status s = req.status;
     if (s.ok()) {
-      if (req.result.size() != req.len) {
+      if ((req.result.size() != req.len) ||
+          (req_offset + block_size(handle) > req.result.size())) {
```
anand1976, line 1721: Just making sure, ```req_offset + block_size(handle) == req.result.size()``` should hold even for direct IO with padding right?

cheng-chang: with padding, should be `<=`

PR #6880, opened 2020-05-26
Expose rocksdb_options_copy function to the C API

DarkEld3r:


db/c.cc
```diff
@@ -2154,6 +2154,10 @@ void rocksdb_options_destroy(rocksdb_options_t* options) {
   delete options;
 }
 
+rocksdb_options_t* rocksdb_options_copy(rocksdb_options_t* options) {
```
pdillinger, line 2157: I prefer it to be called `rocksdb_options_create_copy` to be extra clear about allocation.

db/c_test.c
```diff
@@ -1461,6 +1461,18 @@ int main(int argc, char** argv) {
     rocksdb_cuckoo_options_destroy(cuckoo_options);
   }
 
+  StartPhase("options");
+  {
+    rocksdb_options_t* o;
+    o = rocksdb_options_create();
+
```
pdillinger, line 1468: Please change at least one setting in `o` before copying, and assert the change is reflected in `copy`.

DarkEld3r: It was my initial intention, but as far as I can see, all comparison checks use `CheckEqual` that operates on strings (`char*`). Additionally, there are no getters for options in the C API.

pdillinger: Oh, right. You can do the naughty thing based on create_if_missing being the first field or at least near the beginning. Set create_if_missing to true, make a copy, CheckCondition+memcmp that the first ~16 bytes at each pointer (after cast) are same. And then set create_if_missing on one to false, and make sure CheckCondition+memcmp shows a difference.

Although there's a possibility of a future "false positive" test failure, I think it's manageable and preferable to leaving open false negative.

DarkEld3r: Done.

DarkEld3r:
I'm unable to identify the reason of failure for the Appveyor builds. Could you please give a hint?

pdillinger:
> I'm unable to identify the reason of failure for the Appveyor builds. Could you please give a hint?

2 hr timeout. Can ignore.

db/c_test.c
```diff
@@ -93,6 +93,11 @@ static void CheckEqual(const char* expected, const char* v, size_t n) {
   }
 }
 
+static unsigned char CheckOptionsEqual(const rocksdb_options_t* left,
+                                       const rocksdb_options_t* right) {
+  return memcmp(left, right, 16) == 0;
```
wqfish, line 98: what is 16 here?

wqfish, line 98: this still feels like too much magic here... at least you want to add a comment to explain where 16 comes from. otherwise people won't understand this without reading the comments in this PR.

DarkEld3r: That totally makes sense, I have added the comment.

pdillinger:
Unfortunately this test is unreliable on my machine, and I don't have cycles to add a c_test_util that allows c_test to do more introspection than is available in the public C API.

PR #6881, opened 2020-05-27
Improve API documentation of AssociativeMergeOperator

pdillinger:
Summary: In response to discussion in #6877


include/rocksdb/merge_operator.h
```diff
@@ -222,26 +222,38 @@ class MergeOperator {
   }
 };
 
-// The simpler, associative merge operator.
+// A merge operator that can be implemented as an associative binary
+// operator on values/operands. Implementations must satisfy
```
adamretter, line 226: I suggest to use the term `input` when the parameter could be either a `value` or an `operand`:

```
// operator on an input (either a value or a merge operand), and a merge operand. Implementations must satisfy
```

include/rocksdb/merge_operator.h
```diff
@@ -222,26 +222,38 @@ class MergeOperator {
...
+//      merge_two_values(merge_two_values(a, b), c)
+//   == merge_two_values(a, merge_two_values(b, c))
+// as partial merges might substitute the latter for the former.
+//
```
adamretter, line 230:
```
//      assoc_merge(assoc_merge(input_a, operand_b), operand_c)
//   == assoc_merge(input_a, assoc_merge(operand_b, operand_c))
// as an associative merge operator implements both PartialMerge and FullMerge in one function, it may be applied to either:
// 1) null (when there is no existing value) and an operand
// 2) an existing value and an operand
// 3) two operands
```

include/rocksdb/merge_operator.h
```diff
@@ -222,26 +222,38 @@ class MergeOperator {
...
+// A consequence of this assumption is that values must be compatible as
+// operands in associative merge operators. For example, a silly general merge
```
adamretter, line 232:
```
// A consequence of this assumption is that operands to an associative merge operator
// must also be valid values for storing into the database.
```

include/rocksdb/merge_operator.h
```diff
@@ -222,26 +222,38 @@ class MergeOperator {
...
   ~AssociativeMergeOperator() override {}
 
-  // Gives the client a way to express the read -> modify -> write semantics
+  // Merges two values to express the read -> modify -> write semantics
   // key:           (IN) The key that's associated with this merge operation.
-  // existing_value:(IN) null indicates the key does not exist before this op
-  // value:         (IN) the value to update/merge the existing_value with
+  // left_operand:  (IN) null indicates the key does not exist before this op
+  // right_operand: (IN) the value to update/merge the left operand with
   // new_value:    (OUT) Client is responsible for filling the merge result
```
adamretter, line 246:
```
// Merges two inputs to express the read -> modify -> write semantics
// key:           (IN) The key that's associated with this merge operation.
// left_input:  (IN) null indicates the key does not exist before this operation, otherwise an existing value or operand
// right_operand: (IN) the operand to merge the left_input with
// output:    (OUT) Client is responsible for filling the merge result
// here. The string that output is pointing to will be empty initially.
```

adamretter: Or you could just use `left` and `right`, as opposed to `left_input` and `right_operand`.

include/rocksdb/merge_operator.h
pdillinger: That is not strictly true. It is the reverse that is strictly true (values must be valid as operands). My associative merge operation could be trim-whitespace-and-concatenate. I can assume values have trimmed whitespace, but operands might not yet be trimmed.

pdillinger:
Let's see what you guys think of my more ambitious change. (TODO: update the other existing classes deriving from AssociativeMergeOperator, update PR description, update HISTORY.md, update wiki, etc.)

include/rocksdb/merge_operator.h
adamretter: That is true. I felt that it was easier to explain and understand the other way around... although I now see that is less accurate... we need some middle route ;-)

pdillinger: And I think emphasizing the simpler IdentityAssocMergeOperator means we can brush all the weird nuance of AssociativeMergeOperator under the rug.

include/rocksdb/merge_operator.h
```diff
@@ -27,19 +27,28 @@ class Logger;
 //
 // To use merge, the client needs to provide an object implementing one of
 // the following interfaces:
-//  a) AssociativeMergeOperator - for most simple semantics (always take
-//    two values, and merge them into one value, which is then put back
-//    into rocksdb); numeric addition and string concatenation are examples;
+//  a) IdentityAssocMergeOperator - simplest interface for implementing
+//    merge semantics, always taking two inputs (values or merge operands)
+//    and merging them to one value.
+//    Requires: operator is associative, and no distinction between values
```
adamretter, line 33: May I suggest:

```
//    merge semantics; combines an input (value or merge operand) with a merge operand
//    to create a single result.
```

1. The `right` side arg is always a merge operand.
2. The output could be a value or merge_operand - there is no difference ;-)

include/rocksdb/merge_operator.h
```diff
@@ -27,19 +27,28 @@ class Logger;
...
+//    and merge operands.
+//    Examples: numeric addition, and string concatenation.
+//    Non-examples: numeric subtraction (not associative), editable document
+//    (values are documents, merge operands are edits)
 //
 //  b) MergeOperator - the generic class for all the more abstract / complex
```
adamretter, line 39: Perhaps consider re-ordering a, b, and c into an order of complexity/powerfulness: e.g.

a) IdentityAssocMergeOperator
b) AssociativeMergeOperator
c) MergeOperator

adamretter: Consider renaming `IdentityAssocMergeOperator`. I think it would be better to have consistent naming, e.g.`IdentityAssociativeMergeOperator` ...or if we really must shorten it, then `IdentAssociativeMergeOperator`

include/rocksdb/merge_operator.h
```diff
@@ -27,19 +27,28 @@ class Logger;
...
-//    incremental updates.
+//    incremental updates, including if the byte interpretation of merge
+//    operands is different from values.
 //
-// AssociativeMergeOperator is simpler to implement. MergeOperator is simply
-// more powerful.
+//  c) AssociativeMergeOperator - not as simple as IdentityAssocMergeOperator
+//    because left side is optional, but not as general as MergeOperator.
+//
+// IdentityAssocMergeOperator is simplest to implement, and the rest are
```
adamretter, line 50:
```
// IdentityAssocMergeOperator is simplest to implement but is constrained in its use-cases, and the rest are
```

include/rocksdb/merge_operator.h
```diff
@@ -222,39 +231,36 @@ class MergeOperator {
...
-// A consequence of this assumption is that values must be compatible as
-// operands in associative merge operators. For example, a silly general merge
-// operator could be "total length in bytes of all operands, as a fixed
-// uint64." Although the order of operands does not matter, this would not
-// work as an associative merge operator because the operator does not know
-// whether to treat an operand as an integer length (value to add to) or an
-// arbitrary byte sequence (operand to take the length of).
+// (Implementations of this class must satisfy associativity, like
+// IdentityAssocMergeOperator, but null might not be an identity element
+// of the operator, unlike IdentityAssocMergeOperator.)
```
adamretter, line 242: >  but null might not be an identity element  of the operator, unlike IdentityAssocMergeOperator

I think that is a hard statement to follow. Needs to be simplified...

```
// Implementations of this class must satisfy associativity, like
// IdentityAssocMergeOperator. However, unlike IdentityAssocMergeOperator
// the left input can be null to signal that there is no existing binding for the key.
```

include/rocksdb/merge_operator.h
```diff
@@ -222,39 +231,36 @@ class MergeOperator {
...
-  // Merges two values to express the read -> modify -> write semantics
-  // key:           (IN) The key that's associated with this merge operation.
-  // left_operand:  (IN) null indicates the key does not exist before this op
-  // right_operand: (IN) the value to update/merge the left operand with
-  // new_value:    (OUT) Client is responsible for filling the merge result
-  // here. The string that new_value is pointing to will be empty initially.
+  // Merges two values/operands (left side optional) to express the
+  // read -> modify -> write semantics
+  // key:           (IN) The key associated with this merge (often unused)
+  // left:          (IN) earlier value or operand, or null indicating a
```
adamretter, line 250:
```
// left:          (IN) existing value or previous operand...
```

include/rocksdb/merge_operator.h
```diff
@@ -222,39 +231,36 @@ class MergeOperator {
...
+  // merge with a non-existant or deleted value binding.
+  // right:         (IN) later value or operand to update/merge with
```
adamretter, line 252:
```
// right:         (IN) next value or operand to update/merge with
```

include/rocksdb/merge_operator.h
```diff
@@ -266,4 +272,50 @@ class AssociativeMergeOperator : public MergeOperator {
...
+// only) an operand can be simply promoted to a value when there's no
+// existing value to merge it with.
+class IdentityAssocMergeOperator : public AssociativeMergeOperator {
+ public:
+  ~IdentityAssocMergeOperator() override {}
+
+  // Merges two values/operands to express the read -> modify -> write
+  // semantics
+  // key:           (IN) The key associated with this merge (often unused)
+  // left:          (IN) earlier value or operand
```
adamretter, line 301:
```
// left:          (IN) existing value or previous operand
```

include/rocksdb/merge_operator.h
```diff
@@ -266,4 +272,50 @@ class AssociativeMergeOperator : public MergeOperator {
...
+  // right:         (IN) later value or operand to update/merge with
```
adamretter, line 302:
```
// right:         (IN) next value or operand to update/merge with
```

pdillinger:
> I think implementing MergeOperator::PartialMergeMulti() is often a good choice even for basic associative operations, like addition and string concatenation. That's because it can operate on all inputs at once, so doesn't need to allocate/(de)serialize for intermediate results.

I haven't looked at the hard data, but I expect that cost to be small compared to all the other data handling that RocksDB does. You could hit quadratic with string concat, though. We could heuristically partition to balance sizes on each side (assuming sizes might be additive), so we are worst-case log-linear. Hmm

> I'd think it'd be very convenient ...

To implement a merge operator simply by implementing a function with the same interface as PartialMergeMulti()?

> so user doesn't need to be aware of base value vs merge operand distinction.

If they don't need that distinction? (I expect some users might leverage the distinction.)

ajkr:
> I haven't looked at the hard data, but I expect that cost to be small compared to all the other data handling that RocksDB does. You could hit quadratic with string concat, though. We could heuristically partition to balance sizes on each side (assuming sizes might be additive), so we are worst-case log-linear. Hmm

I guess it is hard to generalize though as it depends on both the RocksDB config and what the merge does - I've seen it make a difference once when the merge operands were protobufs and the merged result size was proportional to the number of merge operands. Then each intermediate result involved complicated (de)serialization and allocations of increasing size. It's true that balancing could've also mitigated the problem while still using the same `BinaryMerge()`-like API, although I didn't think of it then, and operating on the full operand list was straightforward enough.

> To implement a merge operator simply by implementing a function with the same interface as PartialMergeMulti()?

Right, or maybe `AssociativeMergeOperator::FullMergeV2()` could just put the existing value in the operand list if it exists, then call `PartialMergeMulti()`. Then a `AssociativeMergeOperator` implementation could work just by implementing `PartialMergeMulti()`.

> If they don't need that distinction? (I expect some users might leverage the distinction.)

Right.

PR #6882, opened 2020-05-27
Fix wrong result read from ingested file from iterator

yiwu-arbug:
Summary:
Fix issue introduced in #6669. The patch doesn't handle the case for `Prev()`, where results are cached in a local buffer. As a result, `DataBlockIter::ParseNextDataKey()` can restore to a wrong value type when restore the original seqno replaced by global seqno. This can cause wrong result being returned from iterators when issue `Next()` after `Prev()` or `SeekForPrev()`.

Test Plan:
Add new unit test. The test will fail on the `!IsKeyPinned()` assert in `IterKey::UpdateInternalKey()`. After comment out the assert the test will fail with iterator returning wrong result.

Signed-off-by: Yi Wu <yiwu@pingcap.com>


ajkr:
Sorry I should have given #6843 a clearer title as it should address the same issue - but I initially created that PR for a different reason and only discovered this issue a bit later. Can you confirm whether it fixes the same problem?

> The wrong result can be propagate further after compaction, result in data corruption.

I feel this one is not so bad as the last one because compaction iterator shouldn't ever do any backward scanning.

yiwu-arbug:
Great. I didn't follow rocksdb updates lately and don't know there's a fix already. I'll close this one.

PR #6883, opened 2020-05-28
Allow missing "unversioned" python, as in CentOS 8

pdillinger:
Summary: RocksDB Makefile was assuming existence of 'python' command,
which is not present in CentOS 8. We avoid using 'python' if 'python3' is available.

Also added fancy logic to format-diff.sh to make clang-format-diff.py for Python2 work even with Python3 only (as some CentOS 8 FB machines come equipped)

Also, now use just 'python3' for PYTHON if not found so that an informative
"command not found" error will result rather than something weird.

Test Plan: manually tried some variants, 'make check' on a fresh CentOS 8 machine without 'python' executable or Python2 but with clang-format-diff.py for Python2.


pdillinger:
Zitan found that format-diff.sh still runs "python" for clang-format-diff.py. Trying to determine best way to resolve

gg814:
Thanks! (also tested on a vm with CentOS 8)

PR #6884, opened 2020-05-28
Add a new int64 add merge operator

adamretter:
Supercedes https://github.com/facebook/rocksdb/pull/6877

This merge operator adds `int64_t` values. At a cursory glance, this may appear similar to `UInt64AddOperator`, but this merge operator handles signed numbers. It can also handle both Addition or Subtraction by merging either positive or negative integers.

Values are stored into the database using variable-byte encoded (zig-zag encoding), so for smaller positive numbers the value stored has the advantage of also consuming less bytes than UInt64AddOperator.


adamretter:
@pdillinger I added the test binary because that seems to be what has been done in the past. I was only trying to adhere to the current convention!
I wonder, how are we to run the tests without it? If we want to change the current convention, perhaps we should also clean up what is already there too in a separate PR?

> Do you want to test with some larger values, including overflow semantics?

I can if you think it is needed. 

> And of course regenerate TARGETS

Can you explain what the TARGETS file is and how one regenerates it? I am not at all clear as to its purpose...

pdillinger:
> Values are stored into the database using variable-byte encoded (zig-zag encoding), so for smaller positive numbers the value stored has the advantage of also consuming less bytes than UInt64AddOperator.

Btw, you're using a prefix encoding[0] with an independently stored length for that single value, which is a little bit wasteful. It would be more efficient to just elide upper bytes when every bit matches the top-most bit of the upper-most kept byte (sign extension). Thus, when 1 byte, it would be like a signed char, 2 bytes like a signed short, etc.

In fact, if we were so inclined, we could add a new backward-compatible version of UInt64AddOperator with a similar variable-length unsigned encoding.

[0] https://en.wikipedia.org/wiki/Prefix_code

adamretter:
> to follow the instructions from 'make check'

Unfortunately `make check` never completes without error on my machine:

```
ETA: 1337s Left: 2268 AVG: 0.59s  local:16/82/100%/0.6s /bin/bash: line 1: 79572 Abort trap: 6           ./column_family_test >&t/log-column_family_test
ETA: 999s Left: 2181 AVG: 0.46s  local:16/169/100%/0.5s /bin/bash: line 1: 79658 Abort trap: 6           ./db_table_properties_test >&t/log-db_table_properties_test
ETA: 0s Left: 0 AVG: 0.24s  local:0/2350/100%/0.2s s
make[1]: *** [check_0] Error 3
make: *** [check] Error 2
```

> "Check buck targets and code format" GitHub Action

I was very unsure what this action was about. I guess because this TARGETS thing is for internal use, unless you are internal then it is still a mystery.

> python buckifier/buckify_rocksdb.py

Okay I have run that and pushed it to this PR.

@pdillinger I have never been asked to run this before, under what circumstances should it be run, so I know which future PRs I need to run it for?

alexgahide:
Hello. When will this feature be available in rocksjni since it is the expected version of the merge operator to be compliant with Java long type ?

PR #6887, opened 2020-05-29
[WIP] Add ExportColumnFamily and CreateColumnFamilyWithImport for Java

runzhiwang:
Still working

### DIFF
--- a/java/rocksjni/checkpoint.cc
+++ b/java/rocksjni/checkpoint.cc
@@ -72,8 +72,7 @@ void Java_org_rocksdb_Checkpoint_createCheckpoint(JNIEnv* env, jobject /*jobj*/,
  * Method:    exportColumnFamily
  * Signature: (JJLjava/lang/String;)J
  */
-jlong Java_org_rocksdb_Checkpoint_exportColumnFamily(JNIEnv* env,
-                                                     jobject /*jobj*/,
+jlong Java_org_rocksdb_Checkpoint_exportColumnFamily(JNIEnv* env, jobject,
                                                      jlong jcheckpoint_handle,
                                                      jlong jcf_handle,
                                                      jstring jexport_dir) {
--- a/java/rocksjni/checkpoint.cc
+++ b/java/rocksjni/checkpoint.cc
@@ -66,3 +66,39 @@ void Java_org_rocksdb_Checkpoint_createCheckpoint(JNIEnv* env, jobject /*jobj*/,
     ROCKSDB_NAMESPACE::RocksDBExceptionJni::ThrowNew(env, s);
   }
 }
+
+/*
+ * Class:     org_rocksdb_ExportColumnFamily
+ * Method:    exportColumnFamily
+ * Signature: (JJLjava/lang/String;)J
+ */
+jlong Java_org_rocksdb_Checkpoint_exportColumnFamily(JNIEnv* env,
+                                                     jobject /*jobj*/,
+                                                     jlong jcheckpoint_handle,
+                                                     jlong jcf_handle,
+                                                     jstring jexport_dir) {
+  const char* export_dir = env->GetStringUTFChars(jexport_dir, nullptr);
+  if (export_dir == nullptr) {
+    // exception thrown: OutOfMemoryError
+    return 0;
+  }
+
+  auto* checkpoint =
+      reinterpret_cast<ROCKSDB_NAMESPACE::Checkpoint*>(jcheckpoint_handle);
+  auto* cfh =
+      reinterpret_cast<ROCKSDB_NAMESPACE::ColumnFamilyHandle*>(jcf_handle);
+
+  ROCKSDB_NAMESPACE::ExportImportFilesMetaData* export_import_files_metadata =
+      nullptr;
+
+  ROCKSDB_NAMESPACE::Status s = checkpoint->ExportColumnFamily(
+      cfh, export_dir, &export_import_files_metadata);
+
+  env->ReleaseStringUTFChars(jexport_dir, export_dir);
+
+  if (!s.ok()) {
+    ROCKSDB_NAMESPACE::RocksDBExceptionJni::ThrowNew(env, s);
+  }
+
+  return reinterpret_cast<jlong>(export_import_files_metadata);
+}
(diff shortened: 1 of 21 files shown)


java/rocksjni/checkpoint.cc
```diff
@@ -66,3 +66,39 @@ void Java_org_rocksdb_Checkpoint_createCheckpoint(JNIEnv* env, jobject /*jobj*/,
...
   }
 }
+
+/*
+ * Class:     org_rocksdb_ExportColumnFamily
+ * Method:    exportColumnFamily
+ * Signature: (JJLjava/lang/String;)J
+ */
+jlong Java_org_rocksdb_Checkpoint_exportColumnFamily(JNIEnv* env,
+                                                     jobject /*jobj*/,
```
adamretter, line 76: Could you remove the long-indenting? I think if you run `make format` it will correct this for you.

Could you also remove the commented out unused arg names like: ` /*jobj*/`?

This occurs in a few different places in the code.

java/rocksjni/export_import_files_metadata.cc
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+//
+// This file implements the "bridge" between Java and C++ for
+// ROCKSDB_NAMESPACE::FilterPolicy.
```
adamretter, line 7: I think `FilterPolicy` here is a copy-paste mistake?

This occurs in a few files.

java/rocksjni/import_column_family_options.cc
```diff
@@ -0,0 +1,60 @@
...
+
+/*
+ * Class:     org_rocksdb_newImportColumnFamilyOptions
+ * Method:    newImportColumnFamilyOptions
+ * Signature: (Z)J
+ */
+jlong Java_org_rocksdb_ImportColumnFamilyOptions_newImportColumnFamilyOptions__Z(
+    JNIEnv*, jclass, jboolean jmove_files) {
+  auto* options = new ROCKSDB_NAMESPACE::ImportColumnFamilyOptions();
+  options->move_files = static_cast<bool>(jmove_files);
```
adamretter, line 33: Could also be written as `jmove_files == JNI_TRUE`

java/rocksjni/sst_file_metadata.cc
```diff
@@ -0,0 +1,241 @@
...
+  has_exception = JNI_FALSE;
+  std::string path =
+      ROCKSDB_NAMESPACE::JniUtil::copyStdString(env, jpath, &has_exception);
+  if (has_exception == JNI_TRUE) {
+    ROCKSDB_NAMESPACE::RocksDBExceptionJni::ThrowNew(
+        env, "Could not copy jstring to std::string");
+    return;
+  }
+
+  has_exception = JNI_FALSE;
```
adamretter, line 56: I don't think you need to keep setting `has_exception` to `JNI_FALSE`. If it was true, the function will have already returned.

java/src/main/java/org/rocksdb/LiveFileMetaData.java
```diff
@@ -9,30 +9,40 @@
...
-      final byte[] largestKey,
-      final long numReadsSampled,
-      final boolean beingCompacted,
-      final long numEntries,
-      final long numDeletions) {
-    super(fileName, path, size, smallestSeqno, largestSeqno, smallestKey,
-        largestKey, numReadsSampled, beingCompacted, numEntries, numDeletions);
-    this.columnFamilyName = columnFamilyName;
-    this.level = level;
+  private LiveFileMetaData(final long nativeHandle) {
```
adamretter, line 16: So do I understand correctly... that your goal is to have calls to LiveFileMetaData return the up-to-date values rather than the values which were cached in the class members previously?

java/src/main/java/org/rocksdb/LiveFileMetaData.java
```diff
@@ -9,30 +9,40 @@
...
+   * @param smallestSeqno the smallest sequence number
+   * @param largestSeqno the largest sequence number
+   * @param smallestKey the smallest key
+   * @param largestKey the largest key
+   * @param numReadsSampled the number of reads sampled
+   * @param beingCompacted true if the file is being compacted, false otherwise
+   * @param numEntries the number of entries
+   * @param numDeletions the number of deletions
+   */
+  public LiveFileMetaData(final byte[] columnFamilyName, final int level, final String fileName,
```
adamretter, line 39: Is this variant constructor still used?

java/src/main/java/org/rocksdb/SstFileMetaData.java
```diff
@@ -36,29 +38,13 @@
...
-    this.size = size;
-    this.smallestSeqno = smallestSeqno;
-    this.largestSeqno = largestSeqno;
-    this.smallestKey = smallestKey;
-    this.largestKey = largestKey;
-    this.numReadsSampled = numReadsSampled;
-    this.beingCompacted = beingCompacted;
-    this.numEntries = numEntries;
-    this.numDeletions = numDeletions;
+  public SstFileMetaData(final String fileName, final String path, final long size,
```
adamretter, line 41: Is this variant constructor still used?

java/src/test/java/org/rocksdb/ExportImportColumnFamilyTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+  public static final Random rand = PlatformRandomHelper.getPlatformSpecificRandomFactory();
+
+  @Test(expected = RocksDBException.class)
+  public void exportColumnFamily() throws RocksDBException {
+    RocksDB db = null;
+    Options options = null;
+    Checkpoint checkpoint = null;
+    ExportImportFilesMetaData metadata = null;
+    try {
```
adamretter, line 42: Could these tests be rewritten in the try-with-resources style of the other tests?

runzhiwang:
@adamretter Thanks for review very much. I have not finished this PR. I will update the PR according to your suggestions. Then I will split this PR into two other PRs: Add ExportColumnFamily for Java, Add CreateColumnFamilyWithImport for Java. Because current PR changes too much, it's difficult to review.

PR #6891, opened 2020-05-30
Ingest SST files with checksum information

zhichao-cao:
Application can ingest SST files with file checksum information, such that during ingestion, DB is able to check data integrity and identify of the SST file. The PR introduces generate_and_verify_file_checksum to IngestExternalFileOption to control if the ingested checksum information should be verified with the generated checksum.

    1. If generate_and_verify_file_checksum options is *FALSE*: *1)* if DB does not enable SST file checksum, the checksum information ingested will be ignored; *2)* if DB enables the SST file checksum and the checksum function name matches the checksum function name in DB, we trust the ingested checksum, store it in Manifest. If the checksum function name does not match, we treat that as an error and fail the IngestExternalFile() call.
    2. If generate_and_verify_file_checksum options is *TRUE*: *1)* if DB does not enable SST file checksum, the checksum information ingested will be ignored; *2)* if DB enable the SST file checksum, we will use the checksum generator from DB to calculate the checksum for each ingested SST files after they are copied or moved. Then, compare the checksum results with the ingested checksum information: _A)_ if the checksum function name does not match, _verification always report true_ and we store the DB generated checksum information in Manifest. _B)_ if the checksum function name mach, and checksum match, ingestion continues and stores the checksum information in the Manifest. Otherwise, terminate file ingestion and report file corruption. 

Test plan: added unit test, pass make asan_check


file/file_util.cc
```diff
@@ -122,4 +122,51 @@ bool IsWalDirSameAsDBPath(const ImmutableDBOptions* db_options) {
...
+      return io_s;
+    }
+    if (slice.size() == 0) {
+      return IOStatus::Corruption("file too small");
+    }
+    checksum_generator->Update(buffer, slice.size());
+    size -= slice.size();
+  }
+  checksum_generator->Finalize();
+  *file_checksum = checksum_generator->GetChecksum();
```
riversand963, line 167: `file_checksum` and `file_checksum_func_name` better to be asserted.

db/external_sst_file_ingestion_job.cc
```diff
@@ -160,6 +165,129 @@ Status ExternalSstFileIngestionJob::Prepare(
...
+            }
+            files_to_ingest_[i].file_checksum = generated_checksum;
+            files_to_ingest_[i].file_checksum_func_name =
+                generated_checksum_func_name;
+          }
+        }
+      } else if (files_checksum.size() != files_to_ingest_.size() ||
+                 files_checksum_func_name.size() != files_to_ingest_.size()) {
+        // The checksum or checksum function name vector is incomplete
+        status = Status::Corruption(
```
riversand963, line 212: This sounds an InvalidArgument to me, and no verification has started.

db/external_sst_file_ingestion_job.cc
```diff
@@ -28,6 +28,8 @@ namespace ROCKSDB_NAMESPACE {
 
 Status ExternalSstFileIngestionJob::Prepare(
     const std::vector<std::string>& external_files_paths,
+    const std::vector<std::string>& files_checksum,
```
riversand963, line 31: Nit: maybe rename to `files_checksums` and `files_checksum_func_names`.

db/external_sst_file_ingestion_job.cc
```diff
@@ -160,6 +165,129 @@ Status ExternalSstFileIngestionJob::Prepare(
...
+        ROCKS_LOG_WARN(db_options_.info_log,
+                       "Sst file checksum verification failed: %s",
+                       status.ToString().c_str());
+      } else {
+        // Generate the checksum via DB checksum generator factory and verify
+        // the checksum and checksum function name.
+        if (ingestion_options_.generate_and_verify_file_checksum) {
+          for (size_t i = 0; i < files_to_ingest_.size(); i++) {
+            if (files_checksum_func_name[i] != db_checksum_func_name) {
+              status = Status::Corruption(
```
riversand963, line 224: This does not mean corruption, does it?

db/external_sst_file_ingestion_job.cc
```diff
@@ -160,6 +165,129 @@ Status ExternalSstFileIngestionJob::Prepare(
...
+            }
+          }
+        } else {
+          // Checksum verification is not enabled, and also the write
+          // global seqno is False, we trust and store the ingested
+          // checksum if the checksum function name matches. If the
+          // checksum function name does not match, fail the ingestion
+          for (size_t i = 0; i < files_to_ingest_.size(); i++) {
+            if (files_checksum_func_name[i] != db_checksum_func_name) {
+              status = Status::Corruption(
```
riversand963, line 271: Similarly, does checksum function name mismatch mean corruption?

db/external_sst_file_ingestion_job.cc
```diff
@@ -687,6 +821,26 @@ Status ExternalSstFileIngestionJob::AssignGlobalSeqnoForIngestedFile(
   return Status::OK();
 }
 
+IOStatus ExternalSstFileIngestionJob::GenerateChecksumForIngestedFile(
+    IngestedFileInfo* file_to_ingest) {
+  if (db_options_.file_checksum_gen_factory == nullptr ||
+      ingestion_options_.generate_and_verify_file_checksum == false ||
+      ingestion_options_.write_global_seqno == false) {
```
riversand963, line 828: Maybe add a comment here to provide a little context on `write_global_seqno` case. Something like "if write_global_seqno is false, it means we are going to use file checksum computed during Prepare()".

zhichao-cao: @riversand963 Sure. Thanks for the review!

HISTORY.md
```diff
@@ -26,6 +26,7 @@
...
+* Enable SST file ingestion with file checksum information. Added files_checksum and files_checksum_func_name to IngestExternalFileArg such that use can call IngestExternalFiles with file checksum information. Add generate_and_verify_file_checksum to IngestExternalFileOptions (default is False). If it is set to True and DB enables file checksum, DB will generate the checksum for each ingested SST file and verify with the ingested checksum informaiton. Any mismatch will fail the ingestion. If the ingested checksum information is non-empty, DB enables file checksum, and generate_and_verify_file_checksum is False, DB will still verify the checksum function name. If the function name matches, the injested checksum will be directly stored in Manifest without verification. Note that, if IngestExternalFileOptions::write_global_seqno is True, the seqno will be changed in the ingested file. Therefore, the checksum of the file will be changed. In this case, no checksum information will be stored in the Manifest, unless generate_and_verify_file_checksum is set to True and a new checksum will be generated after the seqno is updated.
```
ajkr, line 29: How about `verify_file_checksum` instead of `generate_and_verify_file_checksum`? IMO "generate" implies the checksum originates in the `IngestExternalFiles()` layer, but I believe we may not need the option to control whether that happens (see below).

> Note that, if IngestExternalFileOptions::write_global_seqno is True, the seqno will be changed in the ingested file. Therefore, the checksum of the file will be changed. In this case, no checksum information will be stored in the Manifest, unless generate_and_verify_file_checksum is set to True and a new checksum will be generated after the seqno is updated.

I think for this case we can just generate it anyways when the DB enables file checksum, regardless of the value of `generate_and_verify_file_checksum`.

The other case `IngestExternalFiles()` may need to originate a checksum is when user did not provide a `files_checksum`. I think we can also use the DB setting to decide whether to generate a checksum here.

ajkr: My intent behind that suggestion is making it easier to explain. Here's the explanation after that change in my mind (there may be holes in it, please let me know): 

files_checksum contains the file checksum computed by the algorithm specified in files_checksum_func_name. The algorithm name is always checked for compatibility, and the checksum is optionally verified according to verify_files_checksum. Then files_checksum is written to the manifest. However, if files_checksum is not provided or not relevant due to write_global_seqno being set, we fall back to the DB options for the manifest's checksum.

zhichao-cao: > How about `verify_file_checksum` instead of `generate_and_verify_file_checksum`

If I understand correctly: Whether the checksum of ingested file will be calculated is controlled by the DB checksum option. If DB enables the checksum, we will calculate it anyway. `verify_file_checksum` is used to decided if application want to verify whether the calculated checksum matches with the checksum provided by the application. Yeah, in this way the rule will be simpler.

zhichao-cao: > Most of this seems like an API contract. HISTORY.md should describe the capabilities of the new feature, whether/how any existing API calls have changed behavior, and an API pointer in the form of names.

Will change the description in HISTORY.md after I switch to verify_file_checksum.

file/file_util.cc
```diff
@@ -122,4 +122,53 @@ bool IsWalDirSameAsDBPath(const ImmutableDBOptions* db_options) {
...
+    if (!io_s.ok()) {
+      return io_s;
+    }
+    reader.reset(new SequentialFileReader(std::move(r_file), file_path));
+  }
+
+  char buffer[4096];
+  Slice slice;
+  while (size > 0) {
+    size_t bytes_to_read = std::min(sizeof(buffer), static_cast<size_t>(size));
```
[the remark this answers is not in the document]
zhichao-cao: > verify_checksums_readahead_size

This is a general function to generate the checksum of a file (can be any file in DB). So I think we do not need to use` verify_checksums_readahead_size` which is defined in IngestExternalFileOptions specifically.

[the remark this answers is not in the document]
zhichao-cao: > Also, if size_t is 32 bits, bytes_to_read could be 0 (false report of Corruption) or very small (poor performance) for a ~4GB file.

You mean the issue caused by static_cast here? Or 4096 is too small?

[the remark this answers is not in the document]
pdillinger: > You mean the issue caused by static_cast here? Or 4096 is too small?

static cast before min is a bug when size_t = 32 bits. I'm just noting that in some cases it could be a performance bug (and other cases correctness) as bytes_to_read could be some very small value like 1 for a very large file. 

> So I think we do not need to use verify_checksums_readahead_size which is defined in IngestExternalFileOptions specifically.

I would expect full file checksum should benefit more from readahead than block checksums, so it seems weird to me that verify_checksums_readahead_size would apply to verifying block checksums but not full file checksums, especially since the options are right next to each other in the source code.

I'm OK with "I'll fix that soon in another PR," but I don't want to go through a cycle of someone complaining about speed before we make it fast.

> This is a general function to generate the checksum of a file (can be any file in DB).

Perhaps add a parameter.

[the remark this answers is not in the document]
zhichao-cao: @pdillinger Thanks for the explanation! Will add a readahead parameter to the interface. For the static cast, I will change to uint64_t size and cast sizeof(buffer) instead.

zhichao-cao:
> This HISTORY entry is a little long, but not a big deal. I'm satisfied that the issues I raised are resolved. (I hadn't considered the possibility of ingesting checksum info to manifest without verifying.)

@pdillinger Thanks for pointing out the bug in the checksum generation, I have never though about that issue before. A good learn :). Zippy's requirement and use case is a little bit special, otherwise it is more understandable to force checksum generation if DB enable file checksum and do verification if application provides the checksum with ingested files. So the option is kind of special usage for that case.

PR #6892, opened 2020-05-31
Use official CMake snap on Travis Linux

adamretter:
Switch to using the official CMake snap rather than our own from S3.
See https://gitlab.kitware.com/cmake/cmake/-/issues/20122#note_769266

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -226,13 +226,7 @@ install:
       sudo apt-get install -y mingw-w64 ;
     fi
   - if [[ "${JOB_NAME}" == cmake* ]] && [ "${TRAVIS_OS_NAME}" == linux ]; then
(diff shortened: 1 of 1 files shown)


audiofanatic:
For now you need to add the —beta flag when installing the CMake snap for non-amd64 architectures.

audiofanatic:
The failure for the travis-ci build looks to be explained by the following comment about how PATH is handled by snapd:

https://forum.snapcraft.io/t/build-service-fails-to-use-cmake-from-build-snap/17973/10

Basically, you need to explicitly _prepend_ `/snap/bin` to your PATH.

adamretter:
Thanks @audiofanatic I have prepended /snap/bin now... let's see what happens now.

adamretter:
@pdillinger So can we merge this one?

PR #6894, opened 2020-06-01
Pre-work of adding ExportColumnFamily and CreateColumnFamilyWithImport for Java

runzhiwang:
1. This is the Pre-work of [Add ExportColumnFamily for Java](https://github.com/facebook/rocksdb/pull/6895), and Add CreateColumnFamilyWithImport for Java.
2. `ExportImportFilesMetaData` not only as a result returned by API `ExportColumnFamily`, but also as a parameter to API `CreateColumnFamilyWithImport`. 

3. Define of `ExportImportFilesMetaData`:
```
struct ExportImportFilesMetaData {
   std::string db_comparator_name;       // Used to safety check at import.
   std::vector<LiveFileMetaData> files;  // Vector of file metadata.
 };
```

4. So `LiveFileMetaData` and it's parent class `SstFileMetaData` should not only support transforming from C++ to Java, but also support transforming from Java to C++

### DIFF
--- a/java/rocksjni/live_file_metadata.cc
+++ b/java/rocksjni/live_file_metadata.cc
@@ -1,0 +1,145 @@
+// Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+//
+// This file implements the "bridge" between Java and C++ for LiveFileMetaData
+
+#include <jni.h>
+
+#include "include/org_rocksdb_LiveFileMetaData.h"
+#include "rocksdb/metadata.h"
+#include "rocksjni/portal.h"
+
+/*
+ * Class:     org_rocksdb_LiveFileMetaData
+ * Method:    newLiveFileMetaData
+ * Signature: ()J
+ */
+jlong Java_org_rocksdb_LiveFileMetaData_newLiveFileMetaData__(JNIEnv*, jclass) {
+  auto* metadata = new ROCKSDB_NAMESPACE::LiveFileMetaData();
+  return reinterpret_cast<jlong>(metadata);
+}
+
+/*
+ * Class:     org_rocksdb_LiveFileMetaData
+ * Method:    newLiveFileMetaData
+ * Signature: ([BIILjava/lang/String;Ljava/lang/String;JJJ[BI[BIJZJJ)J
+ */
+jlong Java_org_rocksdb_LiveFileMetaData_newLiveFileMetaData___3BIILjava_lang_String_2Ljava_lang_String_2JJJ_3BI_3BIJZJJ(
+    JNIEnv* env, jclass, jbyteArray jcolumn_family_name,
+    jint jcolumn_family_name_len, jint jlevel, jstring jfile_name,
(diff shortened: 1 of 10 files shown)


java/rocksjni/live_file_metadata.cc
```diff
@@ -0,0 +1,145 @@
...
+  return reinterpret_cast<jlong>(metadata);
+}
+
+/*
+ * Class:     org_rocksdb_LiveFileMetaData
+ * Method:    disposeInternal
+ * Signature: (J)V
+ */
+void Java_org_rocksdb_LiveFileMetaData_disposeInternal(JNIEnv*, jobject,
+                                                       jlong jhandle) {
```
runzhiwang, line 116: **make format** add this long-indenting.

java/src/main/java/org/rocksdb/LiveFileMetaData.java
```diff
@@ -9,30 +9,40 @@
...
-      final byte[] largestKey,
-      final long numReadsSampled,
-      final boolean beingCompacted,
-      final long numEntries,
-      final long numDeletions) {
-    super(fileName, path, size, smallestSeqno, largestSeqno, smallestKey,
-        largestKey, numReadsSampled, beingCompacted, numEntries, numDeletions);
-    this.columnFamilyName = columnFamilyName;
-    this.level = level;
+  private LiveFileMetaData(final long nativeHandle) {
```
runzhiwang, line 16: In order to transform LiveFileMetaData not only from C++ to Java, but also from Java to C++, so Java class only hold the C++ pointer.

java/src/main/java/org/rocksdb/LiveFileMetaData.java
```diff
@@ -9,30 +9,40 @@
...
+   * @param smallestSeqno the smallest sequence number
+   * @param largestSeqno the largest sequence number
+   * @param smallestKey the smallest key
+   * @param largestKey the largest key
+   * @param numReadsSampled the number of reads sampled
+   * @param beingCompacted true if the file is being compacted, false otherwise
+   * @param numEntries the number of entries
+   * @param numDeletions the number of deletions
+   */
+  public LiveFileMetaData(final byte[] columnFamilyName, final int level, final String fileName,
```
runzhiwang, line 39: We need this constructor, so that Java can create an object of **LiveFileMetaData**. Because `ExportImportFilesMetaData` which contains `List<LiveFileMetaData>`, will act as a parameter of  API `CreateColumnFamilyWithImport`.

java/src/main/java/org/rocksdb/SstFileMetaData.java
```diff
@@ -8,57 +8,17 @@
...
-    this.size = size;
-    this.smallestSeqno = smallestSeqno;
-    this.largestSeqno = largestSeqno;
-    this.smallestKey = smallestKey;
-    this.largestKey = largestKey;
-    this.numReadsSampled = numReadsSampled;
-    this.beingCompacted = beingCompacted;
-    this.numEntries = numEntries;
-    this.numDeletions = numDeletions;
+  public SstFileMetaData(final long nativeHandle) {
```
runzhiwang, line 16: So that we can convert `LiveFileMetaData` to it's parent class `SstFileMetaData`.

runzhiwang:
@adamretter  Could you help review this patch ? Thank you very much.

cbi42:
Implemented in #11646

PR #6895, opened 2020-06-01
Add ExportColumnFamily for Java

runzhiwang:
This depends on the PR: [Pre-work of adding ExportColumnFamily and CreateColumnFamilyWithImport for Java](https://github.com/facebook/rocksdb/pull/6894).


cbi42:
Implemented in #11646

PR #6900, opened 2020-06-01
API change: DB::OpenForReadOnly will not write to the file system unless create_if_missing is true

gg814:
Summary: DB::OpenForReadOnly will not write anything to the file system (i.e., create directories or files for the DB) unless create_if_missing is true.

This change also fixes some subcommands of ldb, which write to the file system even if the purpose is for readonly.

Two tests for this updated behavior of DB::OpenForReadOnly are also added.

Other minor changes:
1. Updated HISTORY.md to include this API change of DB::OpenForReadOnly;
2. Updated the help information for the put and batchput subcommands of ldb with the option [--create_if_missing];
3. Updated the comment of Env::DeleteDir to emphasize that it returns OK only if the directory to be deleted is empty.

Test plan: Passed make check and some other manual tests.

### DIFF
--- a/db/db_impl/db_impl_readonly.cc
+++ b/db/db_impl/db_impl_readonly.cc
@@ -131,8 +131,21 @@ Status DBImplReadOnly::NewIterators(
 Status DB::OpenForReadOnly(const Options& options, const std::string& dbname,
                            DB** dbptr, bool /*error_if_log_file_exist*/) {
   if (access(dbname.c_str(), F_OK) == -1) {
-    // file does not exist
+    // directory does not exist
     return Status::NotFound("'" + dbname + "': No such file or directory");
+  } else {
+    std::string path_to_current = dbname;
+    if (!dbname.empty() && dbname.back() == '/') {
+      path_to_current += "CURRENT";
+    } else {
+      path_to_current += "/CURRENT";
+    }
+
+    if (access(path_to_current.c_str(), F_OK) == -1) {
+      // not a RocksDB
+      return Status::NotFound("'" + path_to_current +
+                              "': No such file or directory");
+    }
   }
 
   *dbptr = nullptr;
--- a/db/db_impl/db_impl_readonly.cc
+++ b/db/db_impl/db_impl_readonly.cc
@@ -130,28 +130,22 @@ Status DBImplReadOnly::NewIterators(
 
 Status DB::OpenForReadOnly(const Options& options, const std::string& dbname,
                            DB** dbptr, bool /*error_if_log_file_exist*/) {
-  if (access(dbname.c_str(), F_OK) == -1) {
-    // directory does not exist
-    return Status::NotFound("'" + dbname + "': No such file or directory");
-  } else {
-    std::string path_to_current = dbname;
-    if (!dbname.empty() && dbname.back() == '/') {
-      path_to_current += "CURRENT";
-    } else {
-      path_to_current += "/CURRENT";
-    }
 
-    if (access(path_to_current.c_str(), F_OK) == -1) {
-      // not a RocksDB
-      return Status::NotFound("'" + path_to_current +
-                              "': No such file or directory");
-    }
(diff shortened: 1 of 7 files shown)


db/db_impl/db_impl_readonly.cc
```diff
@@ -130,10 +130,20 @@ Status DBImplReadOnly::NewIterators(
 
 Status DB::OpenForReadOnly(const Options& options, const std::string& dbname,
                            DB** dbptr, bool /*error_if_log_file_exist*/) {
+  // Attempt to read "CURRENT" file
```
pdillinger, line 133: This looks good, but if ldb is called with --column-family=foo, then ldb will directly call the other overload of OpenForReadOnly. So we must check for that overload also.

But preferably we would not check twice on calling this overload, because we do not want unnecessary I/Os on DB open.

And since this affects all users of DB::OpenForReadOnly, I think we need to add to HISTORY.md as a "Public API Change". (This is also kind of a bug fix, but I think it's better to call it a "change".) Suggested wording:
* DB::OpenForReadOnly no longer creates files or directories if the named DB does not exist, unless create_if_missing is set to true.

Speaking of which, this check should not be made if create_if_missing is set to true, as weird as that option is for OpenForReadOnly

tools/ldb_test.py
```diff
@@ -424,6 +424,15 @@ def testDumpLiveFiles(self):
         self.assertRunOK("put x3 y3", "OK")
         dumpFilePath = os.path.join(self.TMP_DIR, "dump2")
         self.assertTrue(self.dumpLiveFiles("--db=%s" % dbPath, dumpFilePath))
+        run_err_null("rm -rf %s" % dumpFilePath)
```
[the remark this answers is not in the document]
pdillinger: Considering my other feedback, it might be better to test this functionality directly on OpenForReadOnly in one of our db unit tests. I don't think an ldb test is necessary if we have good unit tests on when OpenForReadOnly creates files/directories.

db/db_impl/db_impl_readonly.cc
gg814: I didn't know the other overload may be called. Thanks for bringing this up!
I'll also try to figure out if there is a way to avoid the checking twice.
(will change HISTORY.md accordingly in the next update.)

tools/ldb_test.py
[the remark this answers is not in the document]
gg814: Indeed, we should test it on OpenForReadOnly directly.

db/db_impl/db_impl_readonly.cc
```diff
@@ -161,6 +174,19 @@ Status DB::OpenForReadOnly(
     const std::vector<ColumnFamilyDescriptor>& column_families,
     std::vector<ColumnFamilyHandle*>* handles, DB** dbptr,
     bool error_if_log_file_exist) {
+  // Attempt to read "CURRENT" file
```
pdillinger, line 177: Looking good, except we now have code duplication and duplicate I/O checking for file existence when the other overload (no specified column family) is called. Since OpenForReadOnly is a static member function, it should be pretty easy to put the existing functionality of this overload into an anonymous namespace helper function (OpenExistingForReadOnly perhaps), and add another anonymous namespace helper function for this new check.

I'm OK with "extra" `if (!s.ok()) { return s; }`in the callers.

tools/ldb_test.py
```diff
@@ -425,6 +425,7 @@ def testDumpLiveFiles(self):
         dumpFilePath = os.path.join(self.TMP_DIR, "dump2")
         self.assertTrue(self.dumpLiveFiles("--db=%s" % dbPath, dumpFilePath))
 
+
```
pdillinger, line 428: Leftover extra space. (Not a big deal.)

db/db_impl/db_impl_readonly.cc
gg814: Will add an anonymous namespace.
(Previously I thought I may need to modify the function signature and resist doing that :)

PR #6901, opened 2020-06-01
Improve consistency checks in VersionBuilder

ltamasi:
Summary:
The patch cleans up the code and improves the consistency checks around
adding/deleting table files in `VersionBuilder`. Namely, it makes the checks
stricter and improves them in the following ways:
1) A table file can now only be deleted from the LSM tree using the level it
resides on. Earlier, there was some unnecessary wiggle room for
trivially moved files (they could be deleted using a lower level number than
the actual one).
2) A table file cannot be added to the tree if it is already present in the tree
on any level (not just the target level). The earlier code only had an assertion
(which is a no-op in release builds) that the newly added file is not already
present on the target level.
3) The above consistency checks around state transitions are now mandatory,
as opposed to the earlier `CheckConsistencyForDeletes`, which was a no-op
in release mode unless `force_consistency_checks` was set to `true`. The rationale
here is that assuming that the initial state is consistent, a valid transition leads to a
next state that is also consistent; however, an *invalid* transition offers no such
guarantee. Hence it makes sense to validate the transitions unconditionally,
and save `force_consistency_checks` for the paranoid checks that re-validate
the entire state.
4) The new checks build on the mechanism introduced in https://github.com/facebook/rocksdb/pull/6862,
which enables us to efficiently look up the location (level and position within level)
of files in a `Version` by file number. This makes the consistency checks much more
efficient than the earlier `CheckConsistencyForDeletes`, which essentially
performed a linear search.

Test Plan:
Extended the unit tests and ran:

`make check`
`make whitebox_crash_test`

### DIFF
--- a/db/version_builder.cc
+++ b/db/version_builder.cc
@@ -134,7 +134,7 @@ class VersionBuilder::Rep {
   // storing them in levels_ to avoid regression in case there are no files
   // on invalid levels. The version is not consistent if in the end the files
   // on invalid levels don't cancel out.
-  std::unordered_map<int, int> invalid_level_sizes_;
+  std::unordered_map<int, size_t> invalid_level_sizes_;
   // Whether there are invalid new files or invalid deletion on levels larger
   // than num_levels_.
   bool has_invalid_levels_;
--- a/db/version_builder.cc
+++ b/db/version_builder.cc
@@ -511,9 +511,22 @@ class VersionBuilder::Rep {
 
     assert(base_vstorage_);
     const auto location = base_vstorage_->GetFileLocation(file_number);
-    // TODO: make sure it got deleted from original location
-    if (location.IsValid() && location.GetLevel() >= level) {
-      return Status::Corruption();  // TODO: message
+    if (location.IsValid()) {
+      const int orig_level = location.GetLevel();
+      if (orig_level >= num_levels_) {
+        auto level_it = invalid_levels_.find(orig_level);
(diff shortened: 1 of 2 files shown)


db/version_builder.cc
```diff
@@ -130,14 +130,16 @@ class VersionBuilder::Rep {
   VersionSet* version_set_;
   int num_levels_;
   LevelState* levels_;
-  // Store states of levels larger than num_levels_. We do this instead of
+  // Store sizes of levels larger than num_levels_. We do this instead of
   // storing them in levels_ to avoid regression in case there are no files
   // on invalid levels. The version is not consistent if in the end the files
   // on invalid levels don't cancel out.
-  std::map<int, std::unordered_set<uint64_t>> invalid_levels_;
+  std::unordered_map<int, size_t> invalid_level_sizes_;
```
ajkr, line 137: I know this is existing code, but do you happen to know under what condition we add or delete files from an invalid level?

db/version_builder.cc
```diff
@@ -479,78 +435,162 @@ class VersionBuilder::Rep {
...
+
+      return Status::OK();
+    }
+
+    auto& level_state = levels_[level];
+
+    auto& del_files = level_state.deleted_files;
+    auto del_it = del_files.find(file_number);
+    if (del_it != del_files.end()) {
+      del_files.erase(del_it);
```
ajkr, line 531: Do we have to allow this case (a file is added after being deleted causing them to cancel out)?

PR #6905, opened 2020-06-01
Fix handling of too-small filter partition size

pdillinger:
Summary: Because ARM and some other platforms have a larger cache line
size, they have a larger minimum filter size, which causes recently
added PartitionedMultiGet test in db_bloom_filter_test to fail on those
platforms. The code would actually end up using larger partitions,
because keys_per_partition_ would be 0 and never == number of keys
added.

The code now attempts to get as close as possible to the small target
size, while fully utilizing that filter size, if the target partition
size is smaller than the minimum filter size.

Also updated the test to break more uniformly across platforms

Test Plan: updated test, tested on ARM


adamretter:
@pdillinger I could also test this on an s390x if that is helpful?

pdillinger:
> @pdillinger I could also test this on an s390x if that is helpful?

I don't think it's needed, thanks. If I've mis-interpolated the test result for s390x, it's only a performance issue under weird config setting + test issue. I vote only to worry about it if and when we see a failure, which I highly doubt.

PR #6915, opened 2020-06-02
Some fixes for gcc 4.8 and add to Travis

pdillinger:
Summary: People keep breaking the gcc 4.8 compilation due to different
warnings for shadowing member functions with locals. Adding to Travis
to keep compatibility. (gcc 4.8 is default on CentOS 7.)

Test Plan: local and Travis


pdillinger:
Try to re-trigger Travis

PR #6917, opened 2020-06-03
fix  IsDirectory function in env_hdfs.cc

hfrt456:
fix IsDirectory function for hdfsEnv


hfrt456:
> Thanks @hfrt456 for the fix.
> Could you sign the CLA so that we can merge this?
Yes，i have sign the CLA

PR #6922, opened 2020-06-03
Fix two core dumps when files are missing

mrambacher:
The LDB create and drop column family commands failed to check if theere was a valid database prior to dereferencing it, leading to a core dump.

The SstFileDumper prefetch code would dereference a file when the file did not exist as part of the Prefetch code.  This dereference was moved inside an st.ok() check.

Tests were added for both failure conditions.

### DIFF
--- a/tools/sst_dump_test.cc
+++ b/tools/sst_dump_test.cc
@@ -286,6 +286,25 @@ TEST_F(SSTDumpToolTest, ReadaheadSize) {
   }
 }
 
+TEST_F(SSTDumpToolTest, NoSstFile) {
+  Options opts;
+  opts.env = env();
+  std::string file_path = MakeFilePath("no_such_file.sst");
+  char* usage[3];
+  PopulateCommandArgs(file_path, "--show_properties", usage);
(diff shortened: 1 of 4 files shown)


tools/sst_dump_test.cc
```diff
@@ -286,6 +286,25 @@ TEST_F(SSTDumpToolTest, ReadaheadSize) {
...
+  opts.env = env();
+  std::string file_path = MakeFilePath("no_such_file.sst");
+  char* usage[3];
+  PopulateCommandArgs(file_path, "--show_properties", usage);
+  ROCKSDB_NAMESPACE::SSTDumpTool tool;
+  for (const auto& command :
+       {"--command=check", "--command=dump", "--command=raw",
+        "--command=verify", "--command=recompress",
+        "--command=verify_checksum"}) {
+    snprintf(usage[1], kOptLength, "%s", command);
```
pdillinger, line 300: Doesn't this overwrite "--show_properties"? Maybe we don't need it, but we should at least start with something like empty string to make it less confusing.

Issue #6919, opened 2020-06-03
Error: bad block type, when running with persistent cache on windows


zhli1142015:
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://groups.google.com/forum/#!forum/rocksdb or https://www.facebook.com/groups/rocksdb.dev

### Expected behavior
rocks_db with "persistent cache" feature can work as expected in windows.
### Actual behavior
db failed with random error, like "bad block type". Log:

2020/06/03-00:40:30.817202 13398 [ERROR] [db\db_impl\db_impl_compaction_flush.cc:2311] Waiting after background flush error: Corruption: **bad block type**
Accumulated background error counts: 1
### Steps to reproduce the behavior
db_bench.exe --benchmarks="fillseq,readrandom"  --cache_index_and_filter_blocks --pin_l0_filter_and_index_blocks_in_cache --key_size=32 --value_size=26000 --read_cache_path="D:\Git\rocksdb\build\Release" --compression_type="zstd" --disable_wal

This issue should be related to function "GenerateCachePrefix" in "block_based_table_reader.cc". "file->GetUniqueId" always return 0 in windows, and "cc" is null, so "persistent_cache_key_prefix" is always "0". Then persistent cache is messed up by multi table readers. I propose to add "NewId()" in "PersistentCache" to solve this issue.
![cache_perfix](https://user-images.githubusercontent.com/10524738/83631991-05aa0d80-a5d1-11ea-874b-6e148e365c4e.JPG)

PR #6932, opened 2020-06-04
Fix persistent cache on windows

zhli1142015:
Summary:
Persistent cache feature caused rocks db crash on windows. I posted a issue for it, #6919. I found this is because no "persistent_cache_key_prefix" is generated for persistent cache. Looking repo history, "GetUniqueIdFromFile" is not implemented on Windows. So my fix is adding "NewId()" function in "persistent_cache" and using it to generate prefix for persistent cache. In this PR, i also re-enable related test cases defined in "db_test2" and "persistent_cache_test" for windows.
Test Plan:
1. run related test cases in "db_test2" and "persistent_cache_test" on windows and see it passed.
2. manually run db_bench.exe with "read_cache_path" and verified.

### DIFF
--- a/include/rocksdb/persistent_cache.h
+++ b/include/rocksdb/persistent_cache.h
@@ -57,6 +57,10 @@ class PersistentCache {
 
   virtual std::string GetPrintableOptions() const = 0;
 
+  // Return a new numeric id.  May be used by multiple clients who are
+  // sharding the same persistent cache to partition the key space.  Typically
+  // the client will allocate a new id at startup and prepend the id to its
+  // cache keys.
   virtual uint64_t NewId() = 0;
 };
 
--- a/include/rocksdb/persistent_cache.h
+++ b/include/rocksdb/persistent_cache.h
@@ -56,6 +56,8 @@ class PersistentCache {
   virtual StatsType Stats() = 0;
 
   virtual std::string GetPrintableOptions() const = 0;
+
+  virtual uint64_t NewId() = 0;
 };
 
 // Factor method to create a new persistent cache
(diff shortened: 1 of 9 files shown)


utilities/persistent_cache/persistent_cache_tier.h
```diff
@@ -266,6 +266,8 @@ class PersistentCacheTier : public PersistentCache {
 
   virtual std::string GetPrintableOptions() const override = 0;
 
+  virtual uint64_t NewId() override;
```
cheng-chang, line 269: does `NewId` override any interface?

zhli1142015: yes, it's defined in "PersistentCache".

include/rocksdb/persistent_cache.h
```diff
@@ -56,6 +56,8 @@ class PersistentCache {
   virtual StatsType Stats() = 0;
 
   virtual std::string GetPrintableOptions() const = 0;
+
+  virtual uint64_t NewId() = 0;
```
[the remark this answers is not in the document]
zhli1142015: added, thanks for reviewing.

table/block_based/block_based_table_reader.h
```diff
@@ -440,6 +440,9 @@ class BlockBasedTable : public TableReader {
                                   char* buffer, size_t* size);
   static void GenerateCachePrefix(Cache* cc, FSWritableFile* file, char* buffer,
                                   size_t* size);
+  static void GeneratePersistentCachePrefix(PersistentCache* cc,
```
riversand963, line 443: The function name `GeneratePersistentCachePrefix` can be `GenerateCachePrefix`. We know it's for persistent cache by looking at its first argument. C++ function overloading will take care of it. The implementation of this function does not have anything specific to persistent cache.
One step forward is to use a template function for all the `GeneratePersistentCachePrefix` functions.

zhli1142015: Good idea, updated, thanks.

zhli1142015:
@riversand963 , @cheng-chang , could you please help to merge this pr?

PR #6934, opened 2020-06-04
Make DestroyDir destroy directories recursively

cheng-chang:
Currently, `DeleteDir` only deletes the directory if there are no other directories under the target dir. This PR makes it delete directories recursively.

Test Plan:
Added a new unit test in testutil_test.cc.
`make testutil_test`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1090,7 +1090,7 @@ if(WITH_TESTS)
         table/sst_file_reader_test.cc
         table/table_test.cc
         table/block_fetcher_test.cc
-        testutil/testutil_test.cc
+        test_util/testutil_test.cc
         tools/block_cache_analyzer/block_cache_trace_analyzer_test.cc
         tools/ldb_cmd_test.cc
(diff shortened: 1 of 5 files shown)


zhichao-cao:
Thanks for the changes! Some comments left.

CMakeLists.txt
```diff
@@ -1090,6 +1090,7 @@ if(WITH_TESTS)
         table/sst_file_reader_test.cc
         table/table_test.cc
         table/block_fetcher_test.cc
+        test_util/testutil_test.cc
```
zhichao-cao, line 1093: run make format

TARGETS
```diff
@@ -1485,6 +1485,13 @@ ROCKS_TESTS = [
         [],
         [],
     ],
+    [
```
zhichao-cao, line 1488: Did you run python buckifier/buckify_rocksdb.py or manually changed it?

cheng-chang: buckify

PR #6938, opened 2020-06-05
Populate cf_id member of CompactionJobInfo for OnCompactionBegin

matthewvon:
Looks like somebody simply missed initializing a member variable. The column family ID, cf_id, is not set during OnCompactionBegin. But it is set properly in the next function for OnCompactionCompleted. Need this cf_id for tracking progress of a Stardog optimize since there may be multiple compactions required for a given column family.


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1164,6 +1164,7 @@ void DBImpl::NotifyOnCompactionBegin(ColumnFamilyData* cfd, Compaction* c,
   TEST_SYNC_POINT("DBImpl::NotifyOnCompactionBegin::UnlockMutex");
   {
     CompactionJobInfo info{};
+    info.cf_id = cfd->GetID();
     info.cf_name = cfd->GetName();
     info.status = st;
     info.thread_id = env_->GetThreadID();
```
ajkr, line 1170: There seems to be an issue with the existing code, not this PR. That is, the `CompactionJobInfo` populating logic looks identical to the code in `BuildCompactionJobInfo()`. Do you mind checking if we can reuse `BuildCompactionJobInfo()`? If it turns out difficult, let us know and we can still take this as is.

matthewvon: I have viewed the two routines side by side.  They appear to be identical except for the missing cod->GetID().  So yes, BuildCompactionJobInfo() could be used here.  Are you asking that I make the change in this PR? @ajkr

ajkr: Yes, I think it'd be good to make the change in this PR.

matthewvon:
I am performing the edit in two steps.  

1. rebase against facebook/rocksdb master ... that is sha 8632983 ... it fails one circle test:  build-linux-clang10-mini-tsan ... going to guess that is not me

2. check-in code that adjusts NotifyOnCompactionBegin to use BuildCompactionJobInfo ... local test suite passes

PR #6939, opened 2020-06-05
Revisit the handling of the case when a file is re-added to the same level

ltamasi:
Summary:
https://github.com/facebook/rocksdb/pull/6901 subtly changed the handling of the corner case
when a table file is deleted from a level, then re-added to the same level. (Note: this
should be extremely rare; one scenario that comes to mind is a trivial move followed by
a call to `ReFitLevel` that moves the file back to the original level.) Before that change,
a new `FileMetaData` object was created as a result of this sequence; after the change,
the original `FileMetaData` was essentially resurrected (since the deletion and the addition
simply cancel each other out with the change). This patch restores the original behavior,
which is more intuitive considering the interface, and in sync with how trivial moves are handled.
(Also note that `FileMetaData` contains some mutable data members, the values of which
might be different in the resurrected object and the freshly created one.)
The PR also fixes a bug in this area: with the original pre-6901 code, `VersionBuilder`
would add the same file twice to the same level in the scenario described above.

Test Plan:
`make check`

### DIFF
--- a/db/version_builder.cc
+++ b/db/version_builder.cc
@@ -485,11 +485,11 @@ class VersionBuilder::Rep {
     if (add_it != add_files.end()) {
       UnrefFile(add_it->second);
       add_files.erase(add_it);
-    } else {
-      auto& del_files = level_state.deleted_files;
-      assert(del_files.find(file_number) == del_files.end());
-      del_files.emplace(file_number);
     }
+
+    auto& del_files = level_state.deleted_files;
+    assert(del_files.find(file_number) == del_files.end());
+    del_files.emplace(file_number);
 
     table_file_levels_[file_number] =
         VersionStorageInfo::FileLocation::Invalid().GetLevel();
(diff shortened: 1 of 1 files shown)


db/version_builder.cc
```diff
@@ -485,12 +485,12 @@ class VersionBuilder::Rep {
...
       add_files.erase(add_it);
-    } else {
-      auto& del_files = level_state.deleted_files;
-      assert(del_files.find(file_number) == del_files.end());
-      del_files.emplace(file_number);
     }
 
+    auto& del_files = level_state.deleted_files;
+    assert(del_files.find(file_number) == del_files.end());
+    del_files.emplace(file_number);
```
ajkr, line 492: I'm not sure we need to keep track of the deletion in the case of addition followed by deletion of the same file (although I also don't know if that case can happen).

db/version_builder.cc
```diff
@@ -529,15 +529,15 @@ class VersionBuilder::Rep {
     auto del_it = del_files.find(file_number);
     if (del_it != del_files.end()) {
       del_files.erase(del_it);
```
ajkr, line 531: In the case of deletion followed by addition, who calls `RemoveCurrentStats()` on the deleted file, or is it not needed?

db/version_builder.cc
ltamasi: During recovery, when we replay the manifest, all edits for a CF go through the same `VersionBuilder`, so it is possible for a single builder to see the deletion of a file that wasn't in the base `Version` but got added in an earlier `VersionEdit`.

Regarding the change here, I think the addition and deletion code have to be symmetrical, otherwise one can end up with incorrect results. As an example, let's assume file X is initially on level Y. If we make the change in `ApplyFileAddition` but not here, then after `del(levelX, fileY) -> add(levelX, fileY) -> del(levelX, fileY)` the file would be still alive due to it being present in the base version, which is not correct.

db/version_builder.cc
ltamasi: It's done by `MaybeAddFile` below (that's the "file appearing in both the base version and the added list" case).

ajkr: Oh ok, understood now, thanks!

db/version_builder.cc
ajkr: Understood now, thanks.

PR #6943, opened 2020-06-05
Implement a new subcommand "identify" for sst_dump

gg814:
Summary:
Implemented a subcommand of sst_dump called identify, which determines whether a file is an SST file or identifies and lists all the SST files in a directory;

This update also fixes the problem that sst_dump exits with a success state even if target file/directory does not exist/is not an SST file/is empty/is corrupted.

One test is added to sst_dump_test.

Test plan:
Passed make check and a few manual tests

### DIFF
--- a/tools/sst_dump.cc
+++ b/tools/sst_dump.cc
@@ -9,8 +9,7 @@
 
 int main(int argc, char** argv) {
   ROCKSDB_NAMESPACE::SSTDumpTool tool;
-  tool.Run(argc, argv);
-  return 0;
+  return tool.Run(argc, argv);
 }
 #else
 #include <stdio.h>
--- a/tools/sst_dump_test.cc
+++ b/tools/sst_dump_test.cc
@@ -298,13 +298,50 @@ TEST_F(SSTDumpToolTest, NoSstFile) {
         "--command=verify", "--command=recompress", "--command=verify_checksum",
         "--show_properties"}) {
     snprintf(usage[1], kOptLength, "%s", command);
-    ASSERT_TRUE(!tool.Run(3, usage, opts));
+    ASSERT_TRUE(tool.Run(3, usage, opts));
   }
   for (int i = 0; i < 3; i++) {
     delete[] usage[i];
   }
 }
 
+TEST_F(SSTDumpToolTest, ValidSSTPath) {
+  Options opts;
+  opts.env = env();
+  char* usage[3];
+  PopulateCommandArgs("", "", usage);
+  SSTDumpTool tool;
+  std::string file_not_exists = MakeFilePath("file_not_exists.sst");
+  std::string sst_file = MakeFilePath("rocksdb_sst_test.sst");
+  createSST(opts, sst_file);
+  std::string text_file = MakeFilePath("text_file");
+  ASSERT_OK(WriteStringToFile(opts.env, "Hello World!", text_file));
+  std::string fake_sst = MakeFilePath("fake_sst.sst");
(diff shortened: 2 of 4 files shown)


tools/sst_dump.cc
```diff
@@ -9,8 +9,7 @@
 
 int main(int argc, char** argv) {
   ROCKSDB_NAMESPACE::SSTDumpTool tool;
-  tool.Run(argc, argv);
-  return 0;
```
pdillinger, line 13: What could possibly go wrong with that old code? 🤪

tools/sst_dump_test.cc
```diff
@@ -305,6 +305,28 @@ TEST_F(SSTDumpToolTest, NoSstFile) {
   }
 }
 
+TEST_F(SSTDumpToolTest, CheckValidSSTPath) {
+  Options opts;
+  opts.env = env();
+  std::string file_path = MakeFilePath("rocksdb_sst_test.sst");
+
+  char* usage[3];
+  PopulateCommandArgs(file_path, "--command=verify", usage);
```
pdillinger, line 314: Right now you test
{--command=verify} * {sst file doesn't exist, sst file is valid}

I'd like that expanded to
{--command=verify, --command=identify} * {sst file doesn't exist, file contains something plain text, sst file is valid}

pdillinger:
We should also add to HISTORY.md under "bug fixes." Suggestion:

- Fixed sst_dump to return non-zero exit code if the specified file is not a recognized SST file or fails requested checks.

I don't think the 'identify' command and directory stuff is important enough to mention in HISTORY.md. I want to mention this part, though, so that people are encouraged to update their command line tools. :)

tools/sst_dump_test.cc
gg814: Will add the tests correspondingly :)

tools/sst_dump_test.cc
```diff
@@ -298,13 +298,74 @@ TEST_F(SSTDumpToolTest, NoSstFile) {
...
+  snprintf(usage[2], kOptLength, "--file=%s", file_path.c_str());
+  ASSERT_TRUE(tool.Run(3, usage, opts));
+  opts.env->DeleteFile(file_path);
+
+  for (int i = 0; i < 3; i++) {
+    delete[] usage[i];
+  }
+}
+
+TEST_F(SSTDumpToolTest, IdentifyValidSSTPath) {
```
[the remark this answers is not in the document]
gg814: In fact, it is not completely copy-pasted besides the command. But yes, the codes should be made compact :)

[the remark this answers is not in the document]
pdillinger: Oh, I see.

tools/sst_dump_test.cc
```diff
@@ -298,13 +298,74 @@ TEST_F(SSTDumpToolTest, NoSstFile) {
...
+  ASSERT_TRUE(!tool.Run(3, usage, opts));
+  cleanup(opts, file_path);
+
+  // Create text file
+  file_path = MakeFilePath("text_file");
+  std::string text = "Hello World!";
+  ASSERT_OK(WriteStringToFile(opts.env, text, file_path));
+  snprintf(usage[2], kOptLength, "--file=%s", file_path.c_str());
+  // Identify should exit successfully although the file is not SST
+  ASSERT_TRUE(!tool.Run(3, usage, opts));
```
pdillinger, line 361: My intent with the identify command was to only return 0 if the file is a valid sst file. (I hadn't really thought about directories, but finding one seems sufficient to report success.)

PR #6944, opened 2020-06-05
Separate internal and user key comparators in `BlockIter`

ajkr:
Replace `BlockIter::comparator_` and `IndexBlockIter::user_comparator_wrapper_` with a concrete `UserComparatorWrapper` and `InternalKeyComparator`. The motivation for this change was the inconvenience of not knowing the concrete type of `BlockIter::comparator_`, which prevented calling specialized internal key comparison functions to optimize comparison of keys with global seqno applied.

Test Plan:

benchmark setup -- single file DBs, in-memory, no compression. "normal_db"
created by regular flush; "ingestion_db" created by ingesting a file. Both
DBs have same contents.

```
$ TEST_TMPDIR=/dev/shm/normal_db/ ./db_bench -benchmarks=fillrandom,compact -write_buffer_size=10485760000 -disable_auto_compactions=true -compression_type=none -num=1000000
$ ./ldb write_extern_sst ./tmp.sst --db=/dev/shm/ingestion_db/dbbench/ --compression_type=no --hex --create_if_missing < <(./sst_dump --command=scan --output_hex --file=/dev/shm/normal_db/dbbench/000007.sst | awk 'began {print "0x" substr($1, 2, length($1) - 2), "==>", "0x" $5} ; /^Sst file format: block-based/ {began=1}')
$ ./ldb ingest_extern_sst ./tmp.sst --db=/dev/shm/ingestion_db/dbbench/
```

benchmark run command:

```
$ TEST_TMPDIR=/dev/shm/$DB/ ./db_bench -benchmarks=seekrandom -seek_nexts=$SEEK_NEXT -use_existing_db=true -cache_index_and_filter_blocks=false -num=1000000 -cache_size=0 -threads=1 -reads=200000000 -mmap_read=1 -verify_checksum=false
```

results: perf increased significantly  for ingestion_db (enough to cover the losses in #6843) and changed negligibly for normal_db:

SEEK_NEXT | DB | code | ops/sec | % change
-- | -- | -- | -- | --
0 | normal_db | master | 378879 |  
0 | normal_db | PR6944 | 376974 | -0.5
0 | ingestion_db | master | 357010 |  
0 | ingestion_db | PR6944 | 394068 | 10.4
10 | normal_db | master | 235596 |  
10 | normal_db | PR6944 | 237640 | 0.9
10 | ingestion_db | master | 216041 |  
10 | ingestion_db | PR6944 | 234371 | 8.5

### DIFF
--- a/table/block_based/block_test.cc
+++ b/table/block_based/block_test.cc
@@ -34,8 +34,9 @@ static std::string RandomString(Random *rnd, int len) {
   test::RandomString(rnd, len, &r);
   return r;
 }
-std::string GenerateKey(int primary_key, int secondary_key, int padding_size,
-                        Random *rnd) {
+
+std::string GenerateInternalKey(int primary_key, int secondary_key,
+                                int padding_size, Random *rnd) {
   char buf[50];
   char *p = &buf[0];
   snprintf(buf, sizeof(buf), "%6d%4d", primary_key, secondary_key);
@@ -62,7 +63,8 @@ void GenerateRandomKVs(std::vector<std::string> *keys,
   for (int i = from; i < from + len; i += step) {
     // generating keys that shares the prefix
     for (int j = 0; j < keys_share_prefix; ++j) {
-      keys->emplace_back(GenerateKey(i, j, padding_size, &rnd));
+      // `DataBlockIter` assumes it reads only internal keys.
+      keys->emplace_back(GenerateInternalKey(i, j, padding_size, &rnd));
 
       // 100 bytes values
       values->emplace_back(RandomString(&rnd, 100));
@@ -177,7 +179,8 @@ void CheckBlockContents(BlockContents contents, const int max_key,
   // simply be set as invalid; whereas the binary search based iterator will
   // return the one that is closest.
   for (int i = 1; i < max_key - 1; i += 2) {
-    auto key = GenerateKey(i, 0, 0, nullptr);
+    // `DataBlockIter` assumes its APIs receive only internal keys.
+    auto key = GenerateInternalKey(i, 0, 0, nullptr);
     regular_iter->Seek(key);
     ASSERT_TRUE(regular_iter->Valid());
   }
--- a/table/block_based/block_test.cc
+++ b/table/block_based/block_test.cc
@@ -43,6 +43,7 @@ std::string GenerateKey(int primary_key, int secondary_key, int padding_size,
   if (padding_size) {
     k += RandomString(rnd, padding_size);
   }
+  AppendInternalKeyFooter(&k, 0 /* seqno */, kTypeValue);
 
   return k;
 }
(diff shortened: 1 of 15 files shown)


util/user_comparator_wrapper.h
```diff
@@ -17,6 +17,8 @@ namespace ROCKSDB_NAMESPACE {
 // perf_context.user_key_comparison_count.
 class UserComparatorWrapper final : public Comparator {
  public:
+  UserComparatorWrapper() = default;
```
pdillinger, line 20: Can we initialize `user_comparator_` to nullptr to trip segfault on misuse, or at least comment this as creating an uninitialized UserComparatorWrapper?

table/block_based/block_test.cc
```diff
@@ -43,6 +43,7 @@ std::string GenerateKey(int primary_key, int secondary_key, int padding_size,
   if (padding_size) {
     k += RandomString(rnd, padding_size);
   }
+  AppendInternalKeyFooter(&k, 0 /* seqno */, kTypeValue);
```
pdillinger, line 46: Purpose? (Extra tricky keys?)

ajkr: It's because `DataBlockIter` no longer offers its user a choice of what comparator to use. Previously this file called `reader.NewDataIterator(options.comparator, options.comparator, ...` to get a `DataBlockIter` that used `options.comparator` for all its comparisons. However, now `DataBlockIter` offers no such choice and will use an `InternalKeyComparator(options.comparator)` for all its comparisons, so our test cases now need to forge data blocks containing internal keys.

The way the new code selects internal comparator is as follows:
1. `DataBlockIter::Initialize()` calls `raw_key_.SetIsUserKey(false);`
2. `BlockIter::CompareCurrentKey()` in that case uses `icmp_`

util/user_comparator_wrapper.h
ajkr: Good idea, will do.

table/block_based/block_test.cc
pdillinger: Renaming the function GenerateInternalKey I think would suffice. (Unqualified 'key' means user key.)

pdillinger:
Perf note in HISTORY.md also might be warranted. Up to you.

ajkr:
> LGTM! :)

Thanks for the review!

> I would request before landing: 'make check' under valgrind, and preemptively run stress test for a little while.

Done, ran:
```
$ make -j48 valgrind_test
$ python tools/db_crashtest.py blackbox --simple -write_buffer_size=1048576 -target_file_size_base=1048576 -max_bytes_for_level_base=4194304 -value_size_mult=33 -max_key=1000000 --duration=1200 --interval=20
```

> Perf note in HISTORY.md also might be warranted. Up to you.

Done.

The test failure (e.g., `./librocksdb_stress_debug.so: undefined reference to `alternative_rocksdb_ns::test::SetupSyncPointsToMockDirectIO()'`) looks unrelated.

PR #6945, opened 2020-06-05
Maintain the set of linked SSTs in BlobFileMetaData

ltamasi:
Summary:
The `FileMetaData` objects associated with table files already contain the
number of the oldest blob file referenced by the SST in question. This patch
adds the inverse mapping to `BlobFileMetaData`, namely the set of table file
numbers for which the oldest blob file link points to the given blob file (these
are referred to as *linked SSTs*). This mapping will be used by the GC logic.

Implementation-wise, the patch builds on the `BlobFileMetaDataDelta`
functionality introduced in https://github.com/facebook/rocksdb/pull/6835: newly linked/unlinked SSTs are
accumulated in `BlobFileMetaDataDelta`, and the changes to the linked SST set
are applied in one shot when the new `Version` is saved. The patch also reworks
the blob file related consistency checks in `VersionBuilder` so they validate the
consistency of the forward table file -> blob file links and the backward blob file ->
table file links for blob files that are part of the `Version`.

Test Plan:
`make check`

### DIFF
--- a/db/version_builder.cc
+++ b/db/version_builder.cc
@@ -472,6 +472,33 @@ class VersionBuilder::Rep {
     return base_vstorage_->GetFileLocation(file_number).GetLevel();
   }
 
+  uint64_t GetBlobFileNumberForTableFile(int level, uint64_t file_number) {
+    assert(level < num_levels_);
+
+    const auto& added_files = levels_[level].added_files;
+
+    auto it = added_files.find(file_number);
+    if (it != added_files.end()) {
+      const FileMetaData* const meta = it->second;
+      assert(meta);
+
+      return meta->oldest_blob_file_number;
+    }
+
+    const auto location = base_vstorage_->GetFileLocation(file_number);
+    assert(location.IsValid());
+    assert(location.GetLevel() == level);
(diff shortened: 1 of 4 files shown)


db/version_builder.cc
```diff
@@ -537,7 +615,23 @@ class VersionBuilder::Rep {
       }
     }
 
-    // Delete files
+    // Add new blob files
```
riversand963, line 618: The implementation implicitly requires addition/deletion of blob files to occur before addition of SSTs. Do you think  it's worth documenting?

ltamasi:
Thanks for the review @riversand963 !

db/version_builder.cc
```diff
@@ -430,6 +472,27 @@ class VersionBuilder::Rep {
     return base_vstorage_->GetFileLocation(file_number).GetLevel();
   }
 
+  uint64_t GetBlobFileNumberForTableFile(int level, uint64_t file_number) {
```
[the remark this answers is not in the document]
ltamasi: Sounds good. Will change.

db/version_builder.cc
ltamasi: Good point, will add a comment to that effect.

PR #6949, opened 2020-06-06
First step towards handling MANIFEST write error

riversand963:
This PR provides preliminary support for handling IO error during MANIFEST write.
File write/sync is not guaranteed to be atomic. If we encounter an IOError while writing/syncing to the MANIFEST file, we cannot be sure about the state of the MANIFEST file. The version edits may or may not have reached the file. During cleanup, if we delete the newly-generated SST files referenced by the pending version edit(s), but the version edit(s) actually are persistent in the MANIFEST, then next recovery attempt will process the version edits(s) and then fail since the SST files have already been deleted.
One approach is to truncate the MANIFEST after write/sync error, so that it is safe to delete the SST files. However, file truncation may not be supported on certain file systems. Therefore, we take the following approach.
If an IOError is detected during MANIFEST write/sync, we disable file deletions for the faulty database. Depending on whether the IOError is retryable (set by underlying file system), either RocksDB or application can call `DB::Resume()`, or simply shutdown and restart. During `Resume()`, RocksDB will try to switch to a new MANIFEST and write all existing in-memory version storage in the new file. If this succeeds, then RocksDB may proceed. If all recovery is completed, then file deletions will be re-enabled.
Note that multiple threads can call `LogAndApply()` at the same time, though only one of them will be going through the process MANIFEST write, possibly batching the version edits of other threads. When the leading MANIFEST writer finishes, all of the MANIFEST writing threads in this batch will have the same IOError. They will all call `ErrorHandler::SetBGError()` in which file deletion will be disabled.

Possible future directions:
- Add an `ErrorContext` structure so that it is easier to pass more info to `ErrorHandler`. Currently, as in this example, a new `BackgroundErrorReason` has to be added.

Test plan (dev server):
make check

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -211,7 +211,15 @@ Status DBImpl::FlushMemTableToOutputFile(
   if (!s.ok() && !s.IsShutdownInProgress() && !s.IsColumnFamilyDropped()) {
     if (!io_s.ok() && !io_s.IsShutdownInProgress() &&
         !io_s.IsColumnFamilyDropped()) {
-      error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush);
+      // Error while writing to MANIFEST.
+      // In fact, versions_->io_status() can also be the result of renaming
+      // CURRENT file. With current code, it's just difficult to tell. So just
+      // be pessimistic and try write to a new MANIFEST.
+      // TODO: distinguish between MANIFEST write and CURRENT renaming
+      auto err_reason = versions_->io_status().ok()
+                            ? BackgroundErrorReason::kFlush
+                            : BackgroundErrorReason::kManifestWrite;
+      error_handler_.SetBGError(io_s, err_reason);
     } else {
       Status new_bg_error = s;
       error_handler_.SetBGError(new_bg_error, BackgroundErrorReason::kFlush);
@@ -575,7 +583,15 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
   // it is not because of CF drop.
   if (!s.ok() && !s.IsColumnFamilyDropped()) {
     if (!io_s.ok() && !io_s.IsColumnFamilyDropped()) {
-      error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush);
+      // Error while writing to MANIFEST.
+      // In fact, versions_->io_status() can also be the result of renaming
+      // CURRENT file. With current code, it's just difficult to tell. So just
+      // be pessimistic and try write to a new MANIFEST.
+      // TODO: distinguish between MANIFEST write and CURRENT renaming
+      auto err_reason = versions_->io_status().ok()
+                            ? BackgroundErrorReason::kFlush
+                            : BackgroundErrorReason::kManifestWrite;
+      error_handler_.SetBGError(io_s, err_reason);
     } else {
       Status new_bg_error = s;
       error_handler_.SetBGError(new_bg_error, BackgroundErrorReason::kFlush);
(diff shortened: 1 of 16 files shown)


db/version_set.cc
```diff
@@ -3992,6 +3996,8 @@ Status VersionSet::ProcessManifestWrites(
       if (!io_s.ok()) {
```
zhichao-cao, line 3996: I'm thinking we should keep the first IO or Status error occurred.
if (io_status_.ok()) {
  io_status_ = io_s;
} 
if (s.ok()) {
  s = io_s;
}

What do you think?

riversand963: Not sure which line this is for?

zhichao-cao: L3981 and L3996. If we check io_status_ is not OK, it will not be updated by io_s. In this way, we keep the earliest error. So we do not need to check `if (io_status_.IsIOError())`

riversand963: In line 3987, we need to update `io_status_` because (1) `io_s.ok()`, (2) `io_status_.IsIOError()`. It means writing MANIFEST has recovered from a prior IOError by writing to a new MANIFEST successfully.

zhichao-cao: I see. I got the logic here. Thanks for the explanation

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -210,7 +210,15 @@ Status DBImpl::FlushMemTableToOutputFile(
   if (!s.ok() && !s.IsShutdownInProgress() && !s.IsColumnFamilyDropped()) {
     if (!io_s.ok() && !io_s.IsShutdownInProgress() &&
         !io_s.IsColumnFamilyDropped()) {
-      error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush);
+      // Error while writing to MANIFEST.
+      // In fact, versions_->io_status() can also be the result of renaming
+      // CURRENT file. With current code, it's just difficult to tell. So just
+      // be pessimistic and try write to a new MANIFEST.
+      // TODO: distinguish between MANIFEST write and CURRENT renaming
+      auto err_reason = versions_->io_status().ok()
```
anand1976, line 218: Where is ```versions_->io_status()``` reset before performing the flush?

db/error_handler.cc
```diff
@@ -209,6 +225,11 @@ Status ErrorHandler::SetBGError(const Status& bg_err, BackgroundErrorReason reas
     auto_recovery = false;
   }
 
+  if (BackgroundErrorReason::kManifestWrite == reason) {
```
anand1976, line 228: Should this be done in the block in lines 282-296 as well?

ajkr:
The description sounds great. It leaves me wondering, what case is not covered? (I am assuming the title is not referring to a single step process.)

db/db_impl/db_impl_compaction_flush.cc
riversand963: It is reset only after a new attempt of `LogAndApply()` successfully writes to a new MANIFEST, in line version_set.cc:L3987 and L4000.

riversand963:
> The description sounds great. It leaves me wondering, what case is not covered? (I am assuming the title is not referring to a single step process.)

Thanks @ajkr.
In this PR, I am trying to cover cases in which SST files are added, i.e. ingestion, compaction and flush.
This is the first step because I feel there can be improvements on how we handle MANIFEST write error. For example, we currently do not distinguish between MANIFEST write error and CURRENT file renaming error. In the latter case, we do not have to disable file deletion because renaming is atomic, and old, valid MANIFEST file contains all the information we need.

Another consideration is how we propagate error information upwards. Currently, we pass error information to `ErrorHandler` so that its member functions handle errors. Such information includes a BackgroundErrorReason, and a status code of the failing operation. This requires adding new BackgroundErrorReason each time we try to handle a new type of error. For example, in this PR, I had to add `BackgroundErrorReason::kManifestWrite`. If we can pass an ErrorContext {file_path, status, reason} to ErrorHandler, we can pass {MANIFEST, IOError, kFlush} to ErrorHandler without having to add a new reason.

db/error_handler.cc
riversand963: Make sense.
Some more thought: even if the error is a data loss w.r.t to the current MANIFEST, the database can still be recovered if next write to a new MANIFEST succeeds. With this being said, I think we can do this in one shot in line 272.

riversand963: Make sense.
Some more thought: even if the error is a data loss w.r.t to the current MANIFEST, the database can still be recovered if next write to a new MANIFEST succeeds. With this being said, I think we can do this in one shot in line 272.

PR #6952, opened 2020-06-08
Remove unnecessary inclusion of version_edit.h in env

riversand963:
In db_options.c, we should avoid including header files in the `db` directory to avoid introducing unnecessary dependency. The reason why `version_edit.h` has been included in `db_options.cc` is because we need two constants, `kUnknownChecksum` and `kUnknownChecksumFuncName`. We can put these two constants as `constexpr` in the public header `file_checksum.h`.

Test plan (devserver):
make check


riversand963:
Appveyor failure is not related.

PR #6953, opened 2020-06-08
Fix a bug in looking up duplicate keys with MultiGet

anand1976:
When MultiGet is called with duplicate keys, and the key matches the
largest key in an SST file and the value type is merge, only the first
instance of the duplicate key is returned with correct results. This is
due to the incorrect assumption that if a key in a batch is equal to the
largest key in the file, the next key cannot be present in that file.

Tests:
Add a new unit test

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -365,6 +365,7 @@ class FilePickerMultiGet {
         range_(range),
         batch_iter_(range->begin()),
         batch_iter_prev_(range->begin()),
+        upper_key_(range->begin()),
         maybe_repeat_key_(false),
         current_level_range_(*range, range->begin(), range->end()),
         current_file_range_(*range, range->begin(), range->end()),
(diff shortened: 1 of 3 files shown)


db/version_set.cc
```diff
@@ -503,6 +514,12 @@ class FilePickerMultiGet {
     *fd = f;
     *file_index = curr_file_index;
     *is_last_key_in_file = cmp_largest == 0;
+    if (!*is_last_key_in_file) {
+      // If the largest key in the batch overlapping the file is not the
+      // largest key in the file, upper_ley_ would not have been updated so
+      // update it here
+      upper_key_ = batch_iter_;
```
cheng-chang, line 521: can we update `upper_key_` in the above `else` block?

anand1976: It may be slightly more efficient to do it here as we don't need to update every iteration, but just in the end.

cheng-chang: didn't notice the wrapping while loop, makes sense to me now

PR #6954, opened 2020-06-08
Expose Cache statistics in JNI

koldat:
This PR exposes basic getters of LRU cache in Java. It allows us to do better monitoring.
```java
long getCapacity() throws RocksDBException;
long getUsage() throws RocksDBException;
long getHighPriorityPoolUsage() throws RocksDBException;
long getPinnedUsage() throws RocksDBException;
long getEntries() throws RocksDBException;
```

### DIFF
--- a/include/rocksdb/cache.h
+++ b/include/rocksdb/cache.h
@@ -255,11 +255,18 @@ class Cache {
   // returns the memory size for the entries residing in the cache.
   virtual size_t GetUsage() const = 0;
 
+  // returns the memory size for the high priority entries residing in the
+  // cache.
+  virtual size_t GetHighPriorityPoolUsage() const = 0;
+
   // returns the memory size for a specific entry in the cache.
   virtual size_t GetUsage(Handle* handle) const = 0;
 
   // returns the memory size for the entries in use by the system
   virtual size_t GetPinnedUsage() const = 0;
+
+  // returns the number of entries in the cache
+  virtual size_t GetEntries() const = 0;
 
   // returns the charge for the specific entry in the cache.
   virtual size_t GetCharge(Handle* handle) const = 0;
--- a/java/rocksjni/lru_cache.cc
+++ b/java/rocksjni/lru_cache.cc
@@ -6,9 +6,11 @@
 // This file implements the "bridge" between Java and C++ for
 // ROCKSDB_NAMESPACE::LRUCache.
 
+#include "cache/lru_cache.h"
+
 #include <jni.h>
 
-#include "cache/lru_cache.h"
+#include "include/org_rocksdb_Cache.h"
 #include "include/org_rocksdb_LRUCache.h"
 
 /*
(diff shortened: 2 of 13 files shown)


koldat:
@adamretter , please take a look

java/rocksjni/lru_cache.cc
```diff
@@ -41,3 +43,59 @@ void Java_org_rocksdb_LRUCache_disposeInternal(JNIEnv* /*env*/,
       reinterpret_cast<std::shared_ptr<ROCKSDB_NAMESPACE::Cache>*>(jhandle);
   delete sptr_lru_cache;  // delete std::shared_ptr
 }
+
+/*
+ * Class:     org_rocksdb_Cache
+ * Method:    getCapacity
+ * Signature: (J)J
+ */
+jlong Java_org_rocksdb_Cache_getCapacity(JNIEnv* /*env*/, jobject /*jobj*/,
```
adamretter, line 52: Please just omit the commented out parameter names.

java/src/main/java/org/rocksdb/Cache.java
```diff
@@ -10,4 +10,50 @@
...
+
+  /**
+   * @return the number of entries in the cache
+   * @throws RocksDBException
+   */
+  public long getEntries() throws RocksDBException {
+    return getEntries(nativeHandle_);
+  }
+
+  private native long getCapacity(final long handle) throws RocksDBException;
```
adamretter, line 54: It looks to me like these methods will never throw `RocksDBException`. If that is the case, perhaps remove the exceptions.

java/src/test/java/org/rocksdb/LRUCacheTest.java
```diff
@@ -14,14 +16,18 @@
   }
 
   @Test
-  public void newLRUCache() {
+  public void newLRUCache() throws RocksDBException {
```
adamretter, line 19: Will an exception be needed?

include/rocksdb/cache.h
```diff
@@ -255,12 +255,19 @@ class Cache {
   // returns the memory size for the entries residing in the cache.
   virtual size_t GetUsage() const = 0;
 
+  // returns the memory size for the high priority entries residing in the
```
adamretter, line 258: Thanks @koldat, in general this looks good to me.

However, this PR is a bit more than just exposing things to RocksJava... I note you are also adding new C++ API methods to expose values in the C++ API. I think therefore we should just check with @siying that he is happy for these values to be added to the public API...

adamretter (changes requested):
LGTM in general. We just need to check about extending the C++ public API...

mrambacher (commented):
I can see the utility of the "GetEntries" counter in the cache but question the GetHighPriorityUsage.  If you are using a ClockCache (which I am not sure you have no way of knowing), the result would be 0, which may be misleading.

I also would question if this is the way you really want to collect stats about the cache.  To collect all of those statistics, you would be making multiple JNI calls.  Each JNI call is going to do a bunch of mutex locks.  A better solution might be to propose the introduction of some sort of cache->GetStats() call in C++ that fills in some stats structure (map<string, size_t>?) and pass that back and forth.

Of course, that is probably outside of the scope of work you are asking for.

Finally, there should probably be some C++ tests for the new APIs somewhere.

koldat:
@siying can you please comment on this API? I can make it cleaner. The thing is that I have found serious performance issue just because of this simple metric. What happened?

- I am using limited max_open_files so I want to use cache. So I use LRU cache + pin_top_level_index_and_filter(true).
- I am using two level index so that pinned data should be tiny
- After long time I see that pinned blocks are just growing so that cache stops to be useful at all (valid blocks does not fit).

Without this metric I was blind. There was no errors, the application performance just dropped a lot. As I was expecting something weird with cache I have added at least this way of doing that.

Now the question is if sum of all top level indices size was really 4GB or if these is some "pinned" leak. Anyway after I stopped pinning it helped a lot with no slowdown. I am just trying to say how important it is.

image

image

image

include/rocksdb/cache.h
```diff
@@ -255,12 +255,19 @@ class Cache {
...
+  // cache.
+  virtual size_t GetHighPriorityPoolUsage() const = 0;
```
siying, line 260: This function looks good to me.

include/rocksdb/cache.h
```diff
@@ -255,12 +255,19 @@ class Cache {
...
+
   // returns the memory size for a specific entry in the cache.
   virtual size_t GetUsage(Handle* handle) const = 0;
 
   // returns the memory size for the entries in use by the system
   virtual size_t GetPinnedUsage() const = 0;
 
+  // returns the number of entries in the cache
+  virtual size_t GetEntries() const = 0;
```
siying, line 269: I'm not sure this information is necessarily available in the cache. I'm not even sure the value returned by LRU cache and clock cache is accurate, and the information might be unavailable from another implementation. Do you really need this stat?

siying (commented):
Sorry I didn't see the notification.

PR #6955, opened 2020-06-08
Rebuild SST file tool that can be used to fix checksum error.

koldat:
I have extended sst_dump tool to be able to repair broken sst file. It tries to copy:
- Table properties
- Column family ID
- All entries in blocks that does not have checksum error. It skips all entries in broken blocks.
- Range deletions

One can also specify compression algorithm (default is no compression). Because SST is changed manifest also needs to be updated (file size) so one need to pass manifest file name as well.

Example:
```
./sst_dump --file=4230207.sst --command=rebuild --rebuild_manifest=MANIFEST-4161457 --rebuild_compression=kZSTD
from [] to []
Process 4230207.sst
Sst file format: block-based
Data rebuilt into the file 4230207_rebuilt.sst
Manifest rebuilt into the file MANIFEST-4161457_rebuild
```

User than just replace files with files that were generated.

### DIFF
--- a/tools/sst_dump_tool.cc
+++ b/tools/sst_dump_tool.cc
@@ -6,10 +6,8 @@
 //
 #ifndef ROCKSDB_LITE
 
-#include "tools/sst_dump_tool_imp.h"
-
-#include <cinttypes>
 #include <chrono>
+#include <cinttypes>
 #include <iostream>
 #include <map>
 #include <memory>
@@ -17,10 +15,16 @@
 #include <vector>
 
 #include "db/blob/blob_index.h"
+#include "db/log_reader.h"
+#include "db/log_writer.h"
 #include "db/memtable.h"
+#include "db/version_set.h"
 #include "db/write_batch_internal.h"
 #include "env/composite_env_wrapper.h"
+#include "file/read_write_util.h"
+#include "file/sequence_file_reader.h"
 #include "options/cf_options.h"
+#include "port/port.h"
 #include "rocksdb/db.h"
 #include "rocksdb/env.h"
 #include "rocksdb/iterator.h"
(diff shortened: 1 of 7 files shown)


koldat:
More information here: https://groups.google.com/d/msg/rocksdb/gUD4kCGTw-0/uLpFwkO5AgAJ

koldat:
Are we planning to do something here? I would even be happy with having "ignore corrupted records" in compaction job (the I can use ldb), but currently there is not way to fix corrupted database.

@siying @ajkr ?

tools/sst_dump_tool.cc
```diff
@@ -182,6 +186,100 @@ Status SstFileDumper::DumpTable(const std::string& out_filename) {
...
+    VersionEdit edit;
+    s = edit.DecodeFrom(record);
+    if (!s.ok()) {
+      break;
+    }
+
+    for (std::vector<std::pair<int, FileMetaData>>::const_iterator it =
+             edit.GetNewFiles().cbegin();
+         it != edit.GetNewFiles().cend(); it++) {
+      if (it->second.fd.GetNumber() == sstNumber) {
```
ajkr, line 260: Can we append a manifest entry that does delete file followed by add file? Note other parts of the file entry may have changed like file endpoints in case the first or last block got dropped.

ajkr (changes requested):
Yes, we are interested in adding this kind of tooling.

It'd probably be good to add this to `ldb` since it affects the manifest which is DB wide. Also we recently added an `ldb unsafe_remove_sst_file` command which has some similarities (https://github.com/facebook/rocksdb/blob/5746767387dfe452e68714d79ba4822284d33d54/tools/ldb_cmd.cc#L3485).

tools/sst_dump_tool.cc
koldat: Sure, sounds good, I will take a look.

PR #6958, opened 2020-06-08
Add IOTracer reader, writer classes for reading/writing IO operations in a binary file

akankshamahajan15:
Summary:1. As part of IOTracing project, Add a class IOTracer,
IOTraceReader and IOTracerWriter that writes the file operations
information in a binary file. IOTrace Record contains record information
and right now it contains access_timestamp, file_operation, file_name,
io_status, len, offset and later other options will be added when file
system APIs will be call IOTracer.

2. Add few unit test cases that verify that reading and writing to a IO
Trace file is working properly and before start trace and after ending
trace nothing is added to the binary file.

Test Plan: 1. make check -j64
                 2. New testcases for IOTracer.

Reviewers:

Subscribers:

Tasks:

Tags:


trace_replay/io_tracer.h
```diff
@@ -0,0 +1,129 @@
...
+  // Required fields for read.
+  std::string file_name;
+  uint64_t len = 0;
+  uint64_t offset = 0;
+
+  IOTraceRecord() {}
+
+  IOTraceRecord(const uint64_t& _access_timestamp, const TraceType& _trace_type,
+                const std::string& _file_operation,
+                const std::string& _io_status, const std::string& _file_name,
```
anand1976, line 34: ```_file_name```, ```_len``` and ```_offset``` are required fields for reads, as indicated by the comment earlier, right? Should we have another constructor with just the minimal information for other operations?

trace_replay/io_tracer.h
```diff
@@ -0,0 +1,129 @@
...
+  IOTraceWriter(Env* env, const TraceOptions& trace_options,
+                std::unique_ptr<TraceWriter>&& trace_writer);
+  ~IOTraceWriter() = default;
+  // No copy and move.
+  IOTraceWriter(const IOTraceWriter&) = delete;
+  IOTraceWriter& operator=(const IOTraceWriter&) = delete;
+  IOTraceWriter(IOTraceWriter&&) = delete;
+  IOTraceWriter& operator=(IOTraceWriter&&) = delete;
+
+  Status WriteIOOp(const IOTraceRecord& record, const Slice& file_operation,
```
anand1976, line 64: ```IOTraceRecord``` already encapsulates the file_operation, io_status and file_name right? Do we still need the other parameters?

trace_replay/io_tracer.h
```diff
@@ -0,0 +1,129 @@
...
+
+  // Start writing IO operations to the trace_writer.
+  Status StartIOTrace(Env* env, const TraceOptions& trace_options,
+                      std::unique_ptr<TraceWriter>&& trace_writer);
+
+  // Stop writing IO operations to the trace_writer.
+  void EndIOTrace();
+
+  bool is_tracing_enabled() const {
+    return writer_.load(std::memory_order_relaxed);
```
anand1976, line 116: Does ```writer_``` need to be an atomic variable? Access to ```writer_``` is already protected by ```trace_writer_mutex_```. ```is_tracing_enabled()``` seems like it would be more frequently called, so it would be better to find a cheaper way to do the check than an atomic variable.

trace_replay/io_tracer.h
akankshamahajan15: WriteIOOp uses PutLengthPrefixedSlice so either pass the Slice arguments in WriteIOOp("" in IOTraceRecord constructor) or convert the string to Slice in the WriteIOOp. I will change it to second option as I am adding FileSystem function to call WriteIOOp, there will be more strings added and not all are needed for all operations.

trace_replay/io_tracer.h
akankshamahajan15: Yes, I will add another constructors that would be more appropriate than calling with all the arguments.

trace_replay/io_tracer.h
akankshamahajan15: @anand1976  `writer_`  is protected by `trace_writer_mutex_ ` in `StartIOTrace` and `EndIOTrace`. So if I change `writer_` to not be atomic, can I access the value of `writer_` without `trace_writer_mutex_ `  in `is_tracing_enabled()`, Or is it fine to do so.

anand1976: Can we just use a ```bool``` to determine if tracing is enabled? ```writer_``` can continue to be an atomic variable, and ```WriteIOOp``` anyway checks ```writer_``` and acquires mutex, so the worst that can happen is a caller calls ```is_tracing_enabled()```, and then calls ```WriteIOOp```, which would simply ignore it if tracing has been disabled in the meantime.

trace_replay/io_tracer.h
```diff
@@ -0,0 +1,150 @@
...
+  bool is_tracing_enabled() const { return tracing_enabled; }
+
+  Status WriteIOOp(const IOTraceRecord& record);
+
+ private:
+  TraceOptions trace_options_;
+  // A mutex protects the writer_.
+  InstrumentedMutex trace_writer_mutex_;
+  std::atomic<IOTraceWriter*> writer_;
+  bool tracing_enabled;
```
anand1976, line 147: Nit: It would be good to add a comment explaining why this is used instead of ```writer_``` in ```is_tracing_enabled()```.

PR #6963, opened 2020-06-11
Fail point-in-time WAL recovery upon IOError reading WAL

riversand963:
If `options.wal_recovery_mode == WALRecoveryMode::kPointInTimeRecovery`, RocksDB stops replaying WAL once hitting an error and discards the rest of the WAL. This can lead to data loss if the error occurs at an offset smaller than the last sync'ed offset.
Ideally, RocksDB point-in-time recovery should permit recovery if the error occurs after last synced offset while fail recovery if error occurs before the last synced offset. However, RocksDB does not track the synced offset of WALs. Consequently, RocksDB does not know whether an error occurs before or after the last synced offset. An error can be one of the following.
- WAL record checksum mismatch. This can result from both corruption of synced data and dropping of unsynced data during shutdown. We cannot be sure which one. In order not to defeat the original motivation to permit the latter case, we keep the original behavior of point-in-time WAL recovery.
- IOError. This means the WAL can be bad, an indicator of whole file becoming unavailable, not to mention synced part of the WAL. Therefore, we choose to modify the behavior of point-in-time recovery and fail the database recovery.

Test plan (devserver):
make check

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 ### Behavior Changes
 * Disable delete triggered compaction (NewCompactOnDeletionCollectorFactory) in universal compaction mode and num_levels = 1 in order to avoid a corruption bug.
 * `pin_l0_filter_and_index_blocks_in_cache` no longer applies to L0 files larger than `1.5 * write_buffer_size` to give more predictable memory usage. Such L0 files may exist due to intra-L0 compaction, external file ingestion, or user dynamically changing `write_buffer_size` (note, however, that files that are already pinned will continue being pinned, even after such a dynamic change).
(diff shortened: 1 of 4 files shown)


riversand963:
> LGTM!
> 
> I wonder if we should reject `Open()` for all errors returned by the underlying `Env`. It's difficult right now because we can't distinguish where `status->IsCorruption()` originates -- it could either be from the underlying `Env` or from RocksDB detecting checksum mismatch. In any case I think this PR is a step forward and should be landed.

In other places such as https://github.com/facebook/rocksdb/blob/master/db/version_set.h#L1162, RocksDB uses a separate status variable to keep track of IO errors. We probably can follow similar approach by adding a member variable to `log::Reader::Reporter`.

yiwu-arbug:
Thanks for the improvement!

PR #6982, opened 2020-06-15
Extend Get/MultiGet deadline support to table open

anand1976:
Current implementation of the ```read_options.deadline``` option only checks the deadline for random file reads during point lookups. This PR extends the checks to file opens, prefetches and preloads as part of table open.

The main changes are in the ```BlockBasedTable```, partitioned index and filter readers, and ```TableCache``` to take ReadOptions as an additional parameter. In ```BlockBasedTable::Open```, in order to retain existing behavior w.r.t checksum verification and block cache usage, we filter out most of the options in ```ReadOptions``` except ```deadline```. However, having the ```ReadOptions``` gives us more flexibility to honor other options like verify_checksums, fill_cache etc. in the future.

Additional changes in callsites due to function signature changes in ```NewTableReader()``` and ```FilePrefetchBuffer```.

Test:
Add new unit tests in db_basic_test


table/block_based/block_based_table_reader.cc
```diff
@@ -595,12 +595,19 @@ Status BlockBasedTable::Open(
...
+  ReadOptions ro;
+  ro.deadline = read_options.deadline;
+
   // prefetch both index and filters, down to all partitions
   const bool prefetch_all = prefetch_index_and_filter_in_cache || level == 0;
   const bool preload_all = !table_options.cache_index_and_filter_blocks;
 
   if (!ioptions.allow_mmap_reads) {
-    s = PrefetchTail(file.get(), file_size, force_direct_prefetch,
+    s = PrefetchTail(ro, file.get(), file_size, force_direct_prefetch,
```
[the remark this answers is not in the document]
anand1976: If the table open, and thus prefetching, happens in the context of a Get/MultiGet, it could cause the request to exceed the deadline. Also, I think we prefetch data that we're relatively certain will be used soon (for example index partitions), so it just shifts the latency around. 

Haven't considered performing prefetch in another thread. I think it already happens in background threads when the table open is initiated by VersionBuilder. If its for a foreground operation, not sure if we would gain much by doing it in another thread since the main operation would just block waiting for it to complete.

db/table_cache.cc
```diff
@@ -102,12 +103,19 @@ Status TableCache::GetTableReader(
...
+  FileOptions fopts = file_options;
+  Status s = PrepareIOFromReadOptions(ro, ioptions_.env, fopts.io_options);
+  if (s.ok()) {
+    s = ioptions_.fs->NewRandomAccessFile(fname, file_options, &file, nullptr);
+  }
   RecordTick(ioptions_.statistics, NO_FILE_OPENS);
   if (s.IsPathNotFound()) {
     fname = Rocks2LevelTableFileName(fname);
-    s = ioptions_.fs->NewRandomAccessFile(fname, file_options, &file, nullptr);
+    s = PrepareIOFromReadOptions(ro, ioptions_.env, fopts.io_options);
```
riversand963, line 114: This line seems duplicate with line 107?

anand1976:
The deadline is for the life of the iterator. The ```ReadOptions``` is cached at iterator creation time. It might make sense to allow the deadline to be updated (through a separate API), but that's complicated as the new value has to be propagated to all the child iterators.

Now that I think about it, we could check the deadline in seek/next as well, which would handle cases where we iterate over in-memory data. But I'm worried that checking the clock too frequently would slow the iterator down.

riversand963:
> The deadline is for the life of the iterator. The `ReadOptions` is cached at iterator creation time. It might make sense to allow the deadline to be updated (through a separate API), but that's complicated as the new value has to be propagated to all the child iterators.
>
 
The life of an iterator can be very long.

> Now that I think about it, we could check the deadline in seek/next as well, which would handle cases where we iterate over in-memory data. But I'm worried that checking the clock too frequently would slow the iterator down.

So the goal here is just to check deadline during iterator initialization?

db/table_cache.cc
anand1976: Recalculating the timeout here, since the previous call to ```NewRandomAccessFile``` may have used up some time.

PR #6983, opened 2020-06-15
Store DB identity and DB session ID in SST files

gg814:
Summary:
`db_id` and `db_session_id` are now part of the table properties for all formats and stored in SST files.

The `TablePropertiesNames` for these two identifiers are `rocksdb.creating.db.identity` and `rocksdb.creating.session.identity`.

In addition, SST files generated from SstFileWriter and Repairer have DB identity “SST Writer” and “DB Repairer”, respectively. Their DB session IDs are generated in the same way as `DB::GetDbSessionId`.

A table property test is added.

The storage overhead of these two identities in an SST file is 99 bytes: 36 bytes for each of the two identifiers and 27 bytes for "DB identity: " and "DB session identity: ".

Test plan:
make check and some manual tests, also ran make check with COMPILE_WITH_ASAN=1 and /tool/check_format_compatible.sh

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,12 @@
 * Best-efforts recovery ignores CURRENT file completely. If CURRENT file is missing during recovery, best-efforts recovery still proceeds with MANIFEST file(s).
 * In best-efforts recovery, an error that is not Corruption or IOError::kNotFound or IOError::kPathNotFound will be overwritten silently. Fix this by checking all non-ok cases and return early.
 
+### Public API Change
+* `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the LOG file on the line starting with "DB Session ID:".
+
+### New Features
+* DB identity (`db_id`) and DB session identity (`db_session_id`) are added to table properties and stored in SST files. SST files generated from SstFileWriter and Repairer have DB identity “SST Writer” and “DB Repairer”, respectively. Their DB session IDs are generated in the same way as `DB::GetDbSessionId`. The session ID for SstFileWriter (resp., Repairer) resets every time `SstFileWriter::Open` (resp., `Repairer::Run`) is called.
+
 ## 6.11 (6/12/2020)
 ### Bug Fixes
 * Fix consistency checking error swallowing in some cases when options.force_consistency_checks = true.
@@ -30,7 +36,6 @@
 * `pin_l0_filter_and_index_blocks_in_cache` no longer applies to L0 files larger than `1.5 * write_buffer_size` to give more predictable memory usage. Such L0 files may exist due to intra-L0 compaction, external file ingestion, or user dynamically changing `write_buffer_size` (note, however, that files that are already pinned will continue being pinned, even after such a dynamic change).
 * In point-in-time wal recovery mode, fail database recovery in case of IOError while reading the WAL to avoid data loss.
 * A new method `Env::LowerThreadPoolCPUPriority(Priority, CpuPriority)` is added to `Env` to be able to lower to a specific priority such as `CpuPriority::kIdle`.
-* `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the `LOG` file on the line starting with `DB Session ID:`.
 
 ### New Features
 * sst_dump to add a new --readahead_size argument. Users can specify read size when scanning the data. Sst_dump also tries to prefetch tail part of the SST files so usually some number of I/Os are saved there too.
--- a/include/rocksdb/table_properties.h
+++ b/include/rocksdb/table_properties.h
@@ -30,6 +30,8 @@ typedef std::map<std::string, std::string> UserCollectedProperties;
 
 // table properties' human-readable names in the property block.
 struct TablePropertiesNames {
+  static const std::string kDbId;
+  static const std::string kDbSessionId;
   static const std::string kDataSize;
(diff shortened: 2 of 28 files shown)


zhichao-cao:
Thanks for the work! Did you run tools/check_format_compatible.sh to check the SST format compatibility?

table/sst_file_writer.cc
```diff
@@ -237,12 +237,17 @@ Status SstFileWriter::Open(const std::string& file_path) {
     r->column_family_name = "";
     cf_id = TablePropertiesCollectorFactory::Context::kUnknownColumnFamily;
   }
-
+  std::string db_session_id = r->ioptions.env->GenerateUniqueId();
```
zhichao-cao, line 240: I'm not very clear here. Assume I'm an application C, C opens a DB and the DB get a session ID. After some time, C calls SstFileWriter and dump out an SST file from the DB. Based on this line `std::string db_session_id = r->ioptions.env->GenerateUniqueId();`, the generated SST file's db_session_id will be different from the DB that C opens?

include/rocksdb/table_properties.h
```diff
@@ -193,6 +195,15 @@ struct TableProperties {
   // Actual SST file creation time. 0 means unknown.
   uint64_t file_creation_time = 0;
 
+  // DB identity
```
zhichao-cao, line 198: Add one or two sentences to explain the identity and session id.

table/sst_file_writer.cc
gg814: Yes. I think there is no way for SstFileWriter to reference back any DB because SstFileWriter is used to create SST files to be added to some DB later. (Please correct me if I'm misunderstood something :)

So here, in this way, we ensure that every time one opens SstFileWriter, it behaves like one opens a "DB".

include/rocksdb/table_properties.h
gg814: Sure!

gg814:
Will run tools/check_format_compatible.sh to check the SST format compatibility. (did not know it exists :)

table/sst_file_writer.cc
zhichao-cao: It would be better to explain this special case in the comment for user to understand.

HISTORY.md
```diff
@@ -3,6 +3,12 @@
 ### Behavior Changes
 * Best-efforts recovery ignores CURRENT file completely. If CURRENT file is missing during recovery, best-efforts recovery still proceeds with MANIFEST file(s).
 
+### Public API Change
+* `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the `LOG` file on the line starting with `DB Session ID:`.
+
+### New Features
+* DB identity (`db_id`) and DB session identity (`db_session_id`) are added to table properties and stored in SST files.
```
zhichao-cao, line 10: I think you should also mention the special case of SstFileWriter in HISTORY.md, so people can be aware of the db_session_id of files being generated from DB is different from those generated from SstFileWriter.

gg814:
Added a few more comments.

pdillinger:
Awesome. The only other thing I would like to see is how much additional space we have added to SST files with this metadata. I'd be satisfied with one simple before-and-after comparison when both a full DB id and session id are stored.

gg814:
The difference is 99 bytes. The details are as below. 

I created two DBs (db_noids and db_ids) with identical contents:
`ldb --db=/tmp/db_noids put key val --create_if_missing`
`ldb --db=/tmp/db_noids put foo bar`
`ldb --db=/tmp/db_noids put a b`
`ldb --db=/tmp/db_noids put 1 2`
`ls -al /tmp/db_noids`
`total 556`
`drwxr-xr-x 1 zitan users    354 Jun 17 09:33 .`
`drwxrwxrwt 1 root  root  319474 Jun 17 09:33 ..`
`-rw-r--r-- 1 zitan users    793 Jun 17 09:32 000004.sst`
`-rw-r--r-- 1 zitan users    793 Jun 17 09:33 000007.sst`
`-rw-r--r-- 1 zitan users    787 Jun 17 09:33 000010.sst`
`-rw-r--r-- 1 zitan users     24 Jun 17 09:33 000012.log`
(omitted)
`sst_dump --file=/tmp/db_noids/000004.sst --command=scan --show_properties`
`from [] to []`
`Process /tmp/db_noids/000004.sst`
`Sst file format: block-based`
`'key' seq:1, type:1 => val`
`Table Properties:`
(omitted)
`  creation time: 1592411576`
`  time stamp of earliest key: 3`
`  file creation time: 0`
`  Raw user collected properties`
(omitted)

`ls -al /tmp/db_ids`
`total 556`
`drwxr-xr-x 1 zitan users    354 Jun 17 09:46 .`
`drwxrwxrwt 1 root  root  320326 Jun 17 09:46 ..`
`-rw-r--r-- 1 zitan users    892 Jun 17 09:45 000004.sst`
`-rw-r--r-- 1 zitan users    892 Jun 17 09:45 000007.sst`
`-rw-r--r-- 1 zitan users    886 Jun 17 09:46 000010.sst`
`-rw-r--r-- 1 zitan users     24 Jun 17 09:46 000012.log`
(omitted)
`./sst_dump --file=/tmp/db_ids/000004.sst --command=scan --show_properties`
`Process /tmp/db_ids/000004.sst`
`Sst file format: block-based`
`from [] to []`
`'key' seq:1, type:1 => val`
`Table Properties:`
(omitted)
`  creation time: 1592412337`
`  time stamp of earliest key: 0`
`  file creation time: 0`
`  DB identity: e7c5d5f0-d400-4118-b3a8-cf82d813fd1d`
`  DB session identity: f132d07c-d281-4657-a656-3b412a79b280`
`  Raw user collected properties`
(omitted)

zhichao-cao:
You can also add the space overhead conclusion to the PR summary

pdillinger:
Great, for 64MB default target_file_size_base, that's one-to-two millionths additional space usage. :)

PR #6987, opened 2020-06-16
Update HISTORY after cherrypicking

riversand963:
Update HISTORY after cherry-picking the commit to 6.11

Test plan
make check


riversand963:
Thanks @ajkr. I will close this one and follow the convention.

PR #6988, opened 2020-06-16
add `CompactionFilter` to stress/crash tests

ajkr:
Added a `CompactionFilter` that is aware of the stress test's expected state. It only drops key versions that are already covered according to the expected state. It is incompatible with snapshots (same as all `CompactionFilter`s), so disables all snapshot-related features when used in the crash test.

Test Plan: running a minified blackbox crash test

```
$ TEST_TMPDIR=/dev/shm python tools/db_crashtest.py blackbox --max_key=1000000 -write_buffer_size=1048576 -max_bytes_for_level_base=4194304 -target_file_size_base=1048576 -value_size_mult=33 --interval=10 --duration=3600
```

### DIFF
--- a/db_stress_tool/db_stress_common.h
+++ b/db_stress_tool/db_stress_common.h
@@ -238,6 +238,7 @@ DECLARE_bool(sync_fault_injection);
 
 DECLARE_bool(best_efforts_recovery);
 DECLARE_bool(skip_verifydb);
+DECLARE_bool(enable_compaction_filter);
 
 const long KB = 1024;
 const int kRandomValueMaxFactor = 3;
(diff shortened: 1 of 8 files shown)


db_stress_tool/db_stress_common.h
```diff
@@ -439,19 +440,10 @@ extern inline bool GetIntVal(std::string big_endian_key, uint64_t* key_p) {
...
-  unsigned int pad = sizeof(uint64_t) - (size_key % sizeof(uint64_t));
-  if (pad < sizeof(uint64_t)) {
-    big_endian_key.append(pad, '\0');
-    size_key += pad;
-  }
-
   std::string little_endian_key;
   little_endian_key.resize(size_key);
-  for (size_t start = 0; start < size_key; start += sizeof(uint64_t)) {
+  for (size_t start = 0; start + sizeof(uint64_t) <= size_key;
```
anand1976, line 445: Nice. This takes advantage of the fact that there will be atleast one valid prefix in ```big_endian_key```.

db_stress_tool/db_stress_common.h
```diff
@@ -470,17 +462,41 @@ extern inline bool GetIntVal(std::string big_endian_key, uint64_t* key_p) {
     uint64_t pfx = prefixes[i];
     key += (pfx / key_gen_ctx.weights[i]) * key_gen_ctx.window +
            pfx % key_gen_ctx.weights[i];
+    if (i < prefixes.size() - 1) {
+      // The encoding writes a `key_gen_ctx.weights[i] - 1` that counts for
+      // `key_gen_ctx.weights[i]` when there are more prefixes to come. So we
+      // need to add back the one here as we're at a non-last prefix.
```
anand1976, line 468: Great catch!! Both "00 00 00 00 de ad be ef" and "00 00 00 00 de ad be ef 00 00 00 00 00 00 00 00" would have mapped to the same int val without this fix.

ajkr:
Thanks for the review!

ajkr:
https://app.circleci.com/pipelines/github/facebook/rocksdb/1772/workflows/372710b8-f73b-4d9b-a3c2-a4881a4ad660/jobs/2515/steps should have prevented the issue fixed in #7004. For some reason the command it ran (`make release -j32`) passed on CircleCI although it does fail when I run it locally...

PR #6989, opened 2020-06-16
Fix a bug of overwriting return code

riversand963:
In best-efforts recovery, an error that is not Corruption or IOError::kNotFound or IOError::kPathNotFound will be overwritten silently. Fix this by checking all non-ok cases and return early.

Test plan
make check


riversand963:
Thanks @ajkr for the review!

PR #6991, opened 2020-06-16
Add the wrapper classes for FileSystem APIs for IO tracing.

akankshamahajan15:
Summary: 1. Add the wrapper classes FileSystemTracingWrapper, FSSequentialFileTracingWrapper, FSRandomAccessFileTracingWrapper, FSWritableFileTracingWrapper, FSRandomRWFileTracingWrapper that forward the calls to underlying storage system and then pass the file operation information to IOTracer. IOTracer dumps the record in binary format for tracing.
2. Add the wrapper classes FileSystemPtr, FSSequentialFilePtr, FSRandomAccessFilePtr, FSWritableFilePtr and FSRandomRWFilePtr that overload operator-> and return ptr to underlying storage system or Tracing wrapper class based on enabling/disabling of IO tracing. These classes are added to bypass Tracing Wrapper classes when we disable tracing.
3. Add enums in trace.h that distinguish which options need to be added for different file operations(Read, close, write etc) as part of tracing record. 

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/env/file_system_tracer.h
+++ b/env/file_system_tracer.h
@@ -1,0 +1,319 @@
+#pragma once
+
+#include "rocksdb/file_system.h"
+#include "trace_replay/io_tracer.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// FileSystemTracingWrapper is a wrapper class above FileSystem that forwards
+// the call to the underlying storage system. It then invokes IOTracer to record
+// file operations and other contextual information in a binary format for
+// tracing. It overrides methods we are interested in tracing and extends
+// FileSystemWrapper, which forwards all methods that are not explicitly
+// overridden.
+class FileSystemTracingWrapper : public FileSystemWrapper {
+ public:
+  FileSystemTracingWrapper(std::shared_ptr<FileSystem> t,
+                           std::shared_ptr<IOTracer> io_tracer)
+      : FileSystemWrapper(t), io_tracer_(io_tracer) {}
+
+  ~FileSystemTracingWrapper() override {}
+
+  IOStatus NewWritableFile(const std::string& fname,
+                           const FileOptions& file_opts,
+                           std::unique_ptr<FSWritableFile>* result,
+                           IODebugContext* dbg) override;
+
+  IOStatus NewDirectory(const std::string& name, const IOOptions& io_opts,
+                        std::unique_ptr<FSDirectory>* result,
+                        IODebugContext* dbg) override;
+
+  IOStatus GetChildren(const std::string& dir, const IOOptions& io_opts,
+                       std::vector<std::string>* r,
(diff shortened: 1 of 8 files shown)


trace_replay/io_tracer.cc
```diff
@@ -0,0 +1,239 @@
...
+    default:
+      assert(false);
+  }
+  std::string encoded_trace;
+  TracerHelper::EncodeTrace(trace, &encoded_trace);
+  return trace_writer_->Write(encoded_trace);
+}
+
+Status IOTraceWriter::WriteHeader() {
+  Trace trace;
```
zhichao-cao, line 71: For the trace header, I suggest to include the following information such that when developer continues change the trace record format, we are able to do backward compatible.
`s << kTraceMagic << "\t"
    << "IO Trace Version: 0.1\t"
    << "RocksDB Version: " << kMajorVersion << "." << kMinorVersion << "\t"
    << "Format: Timestamp OpType Payload\n";`

zhichao-cao:
Thanks for the work! It will be very helpful to debug, investigate the performance, and cache some special IO situation.

trace_replay/io_tracer.cc
akankshamahajan15: @zhichao-cao , You mean to say I should store header format:  kTraceMagic, kMajorVersion, kMinorVersion in string format instead of binary format?

zhichao-cao: Just ignore my comments here. I thought we use string format for header but I checked the block_cache_tracer, it also uses the binary format. So just keep using binary format.

env/file_system_tracer.h
```diff
@@ -0,0 +1,319 @@
...
+#include "trace_replay/io_tracer.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// FileSystemTracingWrapper is a wrapper class above FileSystem that forwards
+// the call to the underlying storage system. It then invokes IOTracer to record
+// file operations and other contextual information in a binary format for
+// tracing. It overrides methods we are interested in tracing and extends
+// FileSystemWrapper, which forwards all methods that are not explicitly
+// overridden.
```
zhichao-cao, line 13: In order to use the IO tracer, application need to pass FileSystemTracingWrapper to the options.env?

akankshamahajan15: No they dont't. FileSystemPtr  overload operator -> and will check if tracing is enabled or not and return pointer to Filesystem or FileSystemTracingWrapper based on that. I will share the design doc with you.  Thank you for taking a look.

akankshamahajan15:
Because of merging issues moved to https://github.com/facebook/rocksdb/pull/7002

PR #5411, opened 2019-06-04
Make RocksDB secondary instance respect atomic groups in version edits.

HaoyuHuang:
With this commit, RocksDB secondary instance respects atomic groups in version edits.

Issue #6890, opened 2020-05-29
Flush may touch on deleted sst files


guozhangwang:
RocksDB is running on GlusterFS, and we encountered the following error:

```
org.rocksdb.RocksDBException: file is too short (2010 bytes) to be an sstable/tmp/kafka-streams/resilience-nwsl-mServiceAccounts-partyRoleAggregator-V1/0_19/rocksdb/partyRole/005595.sst\n\tat org.rocksdb.RocksDB.flush(Native Method)\n\tat org.rocksdb.RocksDB.flush(RocksDB.java:1760)\n\tat org.apache.kafka.streams.state.internals.RocksDBStore$SingleColumnFamilyAccessor.flush(RocksDBStore.java:515)\n\tat org.apache.kafka.streams.state.internals.RocksDBStore.flush(RocksDBStore.java:344)\n\t... 16 more\n","SOURCE_COMPONENT":"Aggregator"}
```

complaining about the sst file 5595. After looking at the logs we found that file 5595 is a quite short lived sst file:

* At 50-20 15:35:54 file 5595 is created.
* Then a minute later at 15:36:24 file 5595 and 5596 are compacted and as a result 5595 is deleted. 

```
2020/05/20-15:36:24.569817 7f66e67b1700 [DEBUG] [db/db_impl_files.cc:266] [JOB 470] Delete /tmp/kafka-streams/resilience-nwsl-mServiceAccounts-partyRoleAggregator-V1/0_6/rocksdb/partyRole/005595.sst type=2 #5595 – OK
2020/05/20-15:36:24.569827 7f66e67b1700 EVENT_LOG_v1 {"time_micros": 1589988984569824, "job": 470, "event": "table_file_deletion", "file_number": 5595}
```

Since then there’s no operations related on 5595 until the flush:

```
2020/05/21-08:22:53.530027 7f66e75dd700 [db/db_impl_write.cc:1363] [default] New memtable created with log file: #5350. Immutable memtables: 0.
2020/05/21-08:22:53.530088 7f66e4fae700 (Original Log Time 2020/05/21-08:22:53.530076) [db/db_impl_compaction_flush.cc:1399] Calling FlushMemTableToOutputFile with column family [default], flush slots available 1, compaction slots available 1, flush slots scheduled 1, compaction slots scheduled 1
2020/05/21-08:22:53.530091 7f66e4fae700 [db/flush_job.cc:302] [default] [JOB 485] Flushing memtable with next log file: 5350
[...]

PR #6996, opened 2020-06-17
Fail recovery when MANIFEST record checksum mismatch

riversand963:
#5411 refactored `VersionSet::Recover` but introduced a bug, explained as follows.
Before, once a checksum mismatch happens, `reporter` will set `s` to be non-ok. Therefore, Recover will stop processing the MANIFEST any further.
```
// Correct
// Inside Recover
LogReporter reporter;
reporter.status = &s;
log::Reader reader(..., reporter);
while (reader.ReadRecord() && s.ok()) {
...
}
```
The bug is that, the local variable `s` in `ReadAndRecover` won't be updated by `reporter` while reading the MANIFEST. It is possible that the reader sees a checksum mismatch in a record, but `ReadRecord` retries internally read and finds the next valid record. The mismatched record will be ignored and no error is reported.
```
// Incorrect
// Inside Recover
LogReporter reporter;
reporter.status = &s;
log::Reader reader(..., reporter);
s = ReadAndRecover(reader, ...);

// Inside ReadAndRecover
  Status s;  // Shadows the s in Recover.
  while (reader.ReadRecord() && s.ok()) {
   ...
  }
```
`LogReporter` can use a separate `log_read_status` to track the errors while reading the MANIFEST. RocksDB can process more MANIFEST entries only if `log_read_status.ok()`.
Test plan (devserver):
make check

### DIFF
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -27,12 +27,15 @@ VersionEditHandler::VersionEditHandler(
   assert(version_set_ != nullptr);
 }
 
-Status VersionEditHandler::Iterate(log::Reader& reader, std::string* db_id) {
+void VersionEditHandler::Iterate(log::Reader& reader, Status& log_read_status,
+                                 std::string* db_id) {
   Slice record;
   std::string scratch;
+  assert(log_read_status.ok());
   size_t recovered_edits = 0;
   Status s = Initialize();
-  while (reader.ReadRecord(&record, &scratch) && s.ok()) {
+  while (s.ok() && reader.ReadRecord(&record, &scratch) &&
+         log_read_status.ok()) {
     VersionEdit edit;
     s = edit.DecodeFrom(record);
     if (!s.ok()) {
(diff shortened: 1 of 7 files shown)


riversand963:
> LGTM!
> 
> Might be the cause of #6890.

Thanks @ajkr for the review!

db/version_edit_handler.cc
```diff
@@ -27,12 +27,15 @@ VersionEditHandler::VersionEditHandler(
   assert(version_set_ != nullptr);
 }
 
-Status VersionEditHandler::Iterate(log::Reader& reader, std::string* db_id) {
```
ajkr, line 30: I see we already have a parameter passed by non-const reference here, but I'd guess it's one of very few in the RocksDB codebase. Maybe we should prevent that pattern from spreading and pass pointers for consistency?

riversand963: hmm. I checked style guide https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs. The way I see it is that both `reader` and `log_read_status` are kind of input/output parameters, and they are both non-nullable. Therefore, reference seems to conform to the guide. Let me know if I am missing anything here.

riversand963: Actually, `log_read_status` is output parameter. Will update this.

ajkr: Yeah I checked and noticed that too. Maybe my experience with this is from FB's linter or obsolete -- can't really remember. Anyways, leave it up to you.

PR #6997, opened 2020-06-17
Add a new option for BackupEngine to store table files under shared_checksum using DB session id in the backup filenames

gg814:
Summary:
`BackupableDBOptions::new_naming_for_backup_files` is added. This option is false by default. When it is true, backup table filenames under directory shared_checksum are of the form `<file_number>_<crc32c>_<db_session_id>.sst`.

Note that when this option is true, it comes into effect only when both `share_files_with_checksum` and `share_table_files` are true.

Three new test cases are added.

Test plan:
Pased make check.


include/rocksdb/utilities/backupable_db.h
```diff
@@ -116,6 +119,12 @@ struct BackupableDBOptions {
   // Default: INT_MAX
   int max_valid_backups_to_open;
 
+  // If true, backup SST filenames consist of file_number, crc32c, db_session_id
+  // if false, backup SST filenames consist of file_number, crc32c, file_size
+  //
+  // Default: false
+  bool new_naming_for_backup_files;
```
pdillinger, line 126: I think the default should be the new scheme, as there's no forward/backward compatibility issue within RocksDB. There's a chance of people opting out for a few reasons (external dependency on naming scheme; longer names bad; slight performance regression due to reading SST file property), so I think we should have the option available.

Maybe "share_files_with_checksum_unsafe_legacy_naming". A long name might discourage people from using it. ;)

include/rocksdb/utilities/backupable_db.h
```diff
@@ -91,7 +91,10 @@ struct BackupableDBOptions {
   // backups can come from different databases, hence a sst is not uniquely
   // identifed by its name, but by the triple (file name, crc32c, file length)
   // Default: false
-  // Note: this is an experimental option, and you'll need to set it manually
+  // Note: If this option is set to true, we recommend setting
+  // new_naming_for_backup_files to true as well. Otherwise, there is a
+  // non-negligible chance of filename collision when sharing tables in
+  // shared_checksum among several DBs.
```
pdillinger, line 97: See below. The point of this enhancement is to make this option robust and no longer experimental.

utilities/backupable/backupable_db.cc
```diff
@@ -150,6 +162,11 @@ class BackupEngineImpl : public BackupEngine {
     const std::string filename;
     const uint64_t size;
     const uint32_t checksum_value;
+    // DB identities
+    // db_id is obtained for potential usage in the furture but not used
```
pdillinger, line 166: future

utilities/backupable/backupable_db.cc
```diff
@@ -1381,6 +1423,21 @@ Status BackupEngineImpl::CopyOrCreateFile(
...
   }
+  if (s.ok() && UseNewNaming()) {
+    // When copying SST files and using db_session_id in the name,
+    // try to get DB identities
+    // Note that when CopyOrCreateFile() is called while restoring, this block
+    // of code is not executed as for now we do not use session_id to verify
+    // the restored file
+    if (!src.empty()) {
+      // copying
+      bool isSst = (src.length() > 4 && src.rfind(".sst") == src.length() - 4);
```
pdillinger, line 1434: Could we instead pass in a FileType and check for kTableFile?

utilities/backupable/backupable_db.cc
```diff
@@ -1381,6 +1423,21 @@ Status BackupEngineImpl::CopyOrCreateFile(
...
+      if (isSst) {
+        // SST file
+        s = GetFileDbIdentities(src_env, src, db_id, db_session_id);
```
pdillinger, line 1437: We need to support existing SST files without db session identifiers. I recommend they continue to use the old naming scheme. (Benefit: no duplicating existing SST files on incremental backup after upgrade to this version.)

Technically, we should even allow really old SST files that are missing table properties.

utilities/backupable/backupable_db.cc
```diff
@@ -281,14 +298,16 @@ class BackupEngineImpl : public BackupEngine {
...
+  inline std::string GetSharedFileWithChecksum(
+      const std::string& file, const uint32_t checksum_value,
+      const uint64_t file_size, const std::string& db_session_id) const {
     assert(file.size() == 0 || file[0] != '/');
     std::string file_copy = file;
-    return file_copy.insert(file_copy.find_last_of('.'),
-                            "_" + ROCKSDB_NAMESPACE::ToString(checksum_value) +
-                                "_" + ROCKSDB_NAMESPACE::ToString(file_size));
+    const std::string suffix =
+        UseNewNaming() ? db_session_id : ROCKSDB_NAMESPACE::ToString(file_size);
```
pdillinger, line 307: Just because DBImpl session IDs are safe to use in file names does not mean that will hold for all DB implementations. We need to do something safe.

include/rocksdb/utilities/backupable_db.h
gg814: Let me set the default to using the new scheme :)

utilities/backupable/backupable_db.cc
gg814: We could do that. Some preparation work needs to be done in `BackupEngineImpl::Initialize` if passing in a FileType or do the check directly in `BackupEngineImpl::CopyOrCreateFile` without passing in FileType. 

In either case, we will use `ParseFileName` to determine the FileType but I think it just does a bit more than the check here, which also obtains the file number (not necessary here). So l would prefer leaving it as it is.

utilities/backupable/backupable_db.cc
gg814: Right, I shouldn't assume all the old tables have been backed up using the old naming scheme and then switch to the new naming scheme.

utilities/backupable/backupable_db.cc
gg814: Right, also same as the above. Will fall back to file size when the id is empty.

utilities/backupable/backupable_db.cc
pdillinger: Another option then is to have a shared function for determining the file type and calling that both places. Then it's clear that if checking the file extension is a hack, then it's at least a canonical hack. :)

pdillinger: and a consistent hack. :)

utilities/backupable/backupable_db.cc
```diff
@@ -765,7 +808,7 @@ Status BackupEngineImpl::Initialize() {
             work_item.src_env, work_item.dst_env, work_item.src_env_options,
             work_item.sync, work_item.rate_limiter, &result.size,
             &result.checksum_value, work_item.size_limit,
```
ajkr, line 810: I wonder why this function doesn't return a whole `CopyOrCreateResult`, rather than return one of its components and take the other components as output arguments. Might be worthwhile to refactor it in a later PR.

utilities/backupable/backupable_db.cc
```diff
@@ -1381,6 +1432,22 @@ Status BackupEngineImpl::CopyOrCreateFile(
   if (s.ok()) {
     s = dest_writer->Close();
   }
+  if (s.ok() && UseNewNaming()) {
+    // When copying SST files and using db_session_id in the name,
+    // try to get DB identities
+    // Note that when CopyOrCreateFile() is called while restoring, this block
+    // of code is not executed as for now we do not use session_id to verify
```
ajkr, line 1439: Sorry I didn't understand. Isn't `src` still nonempty in the restore case? I thought it's only empty in case we are creating a file, like in the CURRENT file case.

pdillinger:
I think this change did actually break the Java build: https://travis-ci.org/github/facebook/rocksdb/jobs/700987696

include/rocksdb/utilities/backupable_db.h
```diff
@@ -88,11 +88,17 @@ struct BackupableDBOptions {
   std::shared_ptr<RateLimiter> restore_rate_limiter{nullptr};
 
   // Only used if share_table_files is set to true. If true, will consider that
-  // backups can come from different databases, hence a sst is not uniquely
-  // identifed by its name, but by the triple (file name, crc32c, file length)
-  // Default: false
-  // Note: this is an experimental option, and you'll need to set it manually
+  // backups can come from different databases, hence an sst is not uniquely
+  // identifed by its name, but by the triple
+  // (file name, crc32c, db session id or file length)
```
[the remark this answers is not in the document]
gg814: Actually, we had some internal discussions of this task and it seems that users prefer some information in the file name such as its checksum value.

When using customized checksum functions, we may expose the corresponding value.

utilities/backupable/backupable_db.cc
gg814: Good question! I don't know what the author had in mind when writing this function. But I guess one reason may be that `CopyOrCreateResult` calls some other functions that return a `Status`, which is easier for `CopyOrCreateResult` to return directly.

utilities/backupable/backupable_db.cc
gg814: Right, the comment is not properly written :P. I just wanted to stress the fact that the ids are not used for any purposes in restoring the backups. (Will rewrite it properly :)

gg814:
> I think this change did actually break the Java build: https://travis-ci.org/github/facebook/rocksdb/jobs/700987696

Looking into it...

include/rocksdb/utilities/backupable_db.h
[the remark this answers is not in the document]
pdillinger: Additionally, I see this as a case of minimizing "single points of failure". If for some reason we don't get the entropy we think from db session id, such some custom DB implementation or another bug, checksum is a pretty good backstop. And DB session id means we aren't usually reliant on checksum being unique.

utilities/backupable/backupable_db.cc
```diff
@@ -281,14 +298,20 @@ class BackupEngineImpl : public BackupEngine {
...
+  inline std::string GetSharedFileWithChecksum(
+      const std::string& file, const uint32_t checksum_value,
+      const uint64_t file_size, const std::string& db_session_id) const {
     assert(file.size() == 0 || file[0] != '/');
     std::string file_copy = file;
-    return file_copy.insert(file_copy.find_last_of('.'),
-                            "_" + ROCKSDB_NAMESPACE::ToString(checksum_value) +
-                                "_" + ROCKSDB_NAMESPACE::ToString(file_size));
+    const std::string suffix = UseNewNaming() && !db_session_id.empty()
+                                   ? db_session_id
```
pdillinger, line 310: I'm still concerned we aren't fully guaranteeing that the contents of db_session_id is legal/appropriate for a file name. ('/' illegal, ' ' not appropriate)  Perhaps this is just a matter of saying in the contract of DB::GetDbSessionId that the characters are usable in a file name. That way, someone writing an alternate implementation will be aware of the expectation.

And we'll deal with this problem again with user checksums, which can be arbitrary byte data in a string.

gg814: Good point! I think we could also add some kind of sanitizer for the backup file name later. (I think the session id and db id generated by RocksDB consist of only hexadecimal digits and "-")

pdillinger:
@gg814 and I found the Java problem to be depending on sst_dump_tool.cc but that not being included in LIB_SOURCES. We are working on cleaning that up without simply moving whole tool to lib.

HISTORY.md
```diff
@@ -15,6 +15,7 @@
...
 * Added experimental option BlockBasedTableOptions::optimize_filters_for_memory for reducing allocated memory size of Bloom filters (~10% savings with Jemalloc) while preserving the same general accuracy. To have an effect, the option requires format_version=5 and malloc_usable_size. Enabling this option is forward and backward compatible with existing format_version=5.
+* `BackupableDBOptions::new_naming_for_backup_files` is added. This option is true by default. When it is true, backup table filenames are of the form `<file_number>_<crc32c>_<db_session_id>.sst` as opposed to `<file_number>_<crc32c>_<file_size>.sst`. When there is no `db_session_id` available in the table file, we use `file_size` as a fallback. Note that when this option is true, it comes into effect only when both `share_files_with_checksum` and `share_table_files` are true.
```
ajkr, line 18: I am also in favor of a name indicating it is only for legacy compatibility. Another benefit of that naming style is it'll be less surprising if we deprecate it in the future. Ideally we will find nobody's relying on the legacy naming convention and then remove the option in the next major release (7.0).

Maybe `legacy_sst_naming_no_session_id` or something.

utilities/backupable/backupable_db.cc
```diff
@@ -1551,6 +1652,48 @@ Status BackupEngineImpl::CalculateChecksum(const std::string& src, Env* src_env,
...
 
+Status BackupEngineImpl::GetFileDbIdentities(Env* src_env,
+                                             const std::string& file_path,
+                                             std::string* db_id,
+                                             std::string* db_session_id) {
+  // Prepare the full_path of file_path under src_env for SstFileDumper
+  std::string full_path;
+  src_env->GetAbsolutePath(file_path, &full_path);
+
+  SstFileDumper sst_reader(Options(), full_path,
```
ajkr, line 1663: Does using `Options()` here work generally? I'd think it uses default `env`, `table_factory`, etc. which will be an issue if backing up from a non-default environment, backing up a non-block based table, etc.

HISTORY.md
gg814: Thanks for the suggestion! Actually, @pdillinger has a similar suggestion with consideration of potential extension of the naming scheme in the future. I will probably create a follow-up PR for it soon (`enum option share_files_with_checksum_naming = {kChecksumAndFileSize, kChecksumAndDbSessionId}`). So we may leave it as it is for the moment :)

utilities/backupable/backupable_db.cc
ajkr: The tests attempt to catch this bug by using `ChrootEnv` for the source/destination `Env`s so, if we accidentally use the wrong `Env`, there'll be an error (probably missing file). Here, you discovered an interesting way to use the wrong `Env` while satisfying the tests -- get the absolute path using the right `ChrootEnv` and then access it with the default `Env`.

gg814: This is the same as how sst_dump reads a table file. I think sst_dump first accepts a default option and then gradually fine-tunes the option (to adapt for a specific table type, etc.) I may misunderstand this so please correct me if I'm wrong :P

ajkr: Yes you are correct. We also need to do the same to obtain the correct env, table type, etc. We can use something like `rocksdb::DB::GetOptions()` (I'm not sure yet what to do about column family-specific options as we don't know the column family here...).

gg814:
> I think the only remaining question I have is, why do we need to use `SstFileDumper` and, as a result, also move it out of the tool-specific code? Can we use `ReadTableProperties()` from "table/meta_blocks.h"?

Actually, `SstFileDumper` does access the table properties from `ReadTableProperties()` in "table/meta_blocks.h". I guess we could also do it directly without `SstFileDumper` but since `SstFileDumper` has already done the preparation work needed for us to use `ReadTableProperties()` from "table/meta_blocks.h". I think it makes sense to utilize this encapsulation provided by `SstFileDumper` rather than repeating the work.

gg814:
> > I think the only remaining question I have is, why do we need to use `SstFileDumper` and, as a result, also move it out of the tool-specific code? Can we use `ReadTableProperties()` from "table/meta_blocks.h"?
> 
> Actually, `SstFileReader` looks like an even better already existing API for your use case. It handles the setup of file reader for you, and also abstracts the details specific to table type (unlike `ReadTableProperties()`).

I did consider using `SstFileReader`. The thing is that its constructor needs an `Option`, which is column family dependent. We could use `DB::GetOption`, however, it requires us to know the specific column family (as you mentioned :)

It is after considering all the possibilities here that I made the choice to use `SstFileDumper` :P

HISTORY.md
gg814: (I promise I'll create a follow-up soon :P)

ajkr:
> > > I think the only remaining question I have is, why do we need to use `SstFileDumper` and, as a result, also move it out of the tool-specific code? Can we use `ReadTableProperties()` from "table/meta_blocks.h"?
> > 
> > 
> > Actually, `SstFileReader` looks like an even better already existing API for your use case. It handles the setup of file reader for you, and also abstracts the details specific to table type (unlike `ReadTableProperties()`).
> 
> I did consider using `SstFileReader`. The thing is that its constructor needs an `Option`, which is column family dependent. We could use `DB::GetOption`, however, it requires us to know the specific column family (as you mentioned :)
> 
> It is after considering all the possibilities here that I made the choice to use `SstFileDumper` :P

Got it. But as it turns out `SstFileDumper` also requires an `Options` to construct it and thus encounters the same problem, can we reconsider using the already available `SstFileReader`?

gg814:
> > > > I think the only remaining question I have is, why do we need to use `SstFileDumper` and, as a result, also move it out of the tool-specific code? Can we use `ReadTableProperties()` from "table/meta_blocks.h"?
> > > 
> > > 
> > > Actually, `SstFileReader` looks like an even better already existing API for your use case. It handles the setup of file reader for you, and also abstracts the details specific to table type (unlike `ReadTableProperties()`).
> > 
> > 
> > I did consider using `SstFileReader`. The thing is that its constructor needs an `Option`, which is column family dependent. We could use `DB::GetOption`, however, it requires us to know the specific column family (as you mentioned :)
> > It is after considering all the possibilities here that I made the choice to use `SstFileDumper` :P
> 
> Got it. But as it turns out `SstFileDumper` also requires an `Options` to construct it and thus encounters the same problem, can we reconsider using the already available `SstFileReader`?

The reason I chose `SstFileDumper` over `SstFileReader` is that it seems to me `SstFileDumper` can better handle older table types and that we can use it off-the-shelf. We could make `SstFileReader` to behave more like `SstFileDumper` but I resisted to do so because I felt it may be unnecessary work. What do you think? :)

ajkr:
> > > > > I think the only remaining question I have is, why do we need to use `SstFileDumper` and, as a result, also move it out of the tool-specific code? Can we use `ReadTableProperties()` from "table/meta_blocks.h"?
> > > > 
> > > > 
> > > > Actually, `SstFileReader` looks like an even better already existing API for your use case. It handles the setup of file reader for you, and also abstracts the details specific to table type (unlike `ReadTableProperties()`).
> > > 
> > > 
> > > I did consider using `SstFileReader`. The thing is that its constructor needs an `Option`, which is column family dependent. We could use `DB::GetOption`, however, it requires us to know the specific column family (as you mentioned :)
> > > It is after considering all the possibilities here that I made the choice to use `SstFileDumper` :P
> > 
> > 
> > Got it. But as it turns out `SstFileDumper` also requires an `Options` to construct it and thus encounters the same problem, can we reconsider using the already available `SstFileReader`?
> 
> The reason I chose `SstFileDumper` over `SstFileReader` is that it seems to me `SstFileDumper` can better handle older table types and that we can use it off-the-shelf. We could make `SstFileReader` to behave more like `SstFileDumper` but I resisted to do so because I felt it may be unnecessary work. What do you think? :)

Understood. Can the `SstFileDumper` API be changed to take a `Env*` rather than an `Options`? That seems to be the only part of `Options` that is necessary to set in order to construct a `SstFileDumper`. Then it's clear we are passing everything that's needed when we construct it from a place that lacks access to many options (particularly column family ones).

gg814:
> > > > > > I think the only remaining question I have is, why do we need to use `SstFileDumper` and, as a result, also move it out of the tool-specific code? Can we use `ReadTableProperties()` from "table/meta_blocks.h"?
> > > > > 
> > > > > 
> > > > > Actually, `SstFileReader` looks like an even better already existing API for your use case. It handles the setup of file reader for you, and also abstracts the details specific to table type (unlike `ReadTableProperties()`).
> > > > 
> > > > 
> > > > I did consider using `SstFileReader`. The thing is that its constructor needs an `Option`, which is column family dependent. We could use `DB::GetOption`, however, it requires us to know the specific column family (as you mentioned :)
> > > > It is after considering all the possibilities here that I made the choice to use `SstFileDumper` :P
> > > 
> > > 
> > > Got it. But as it turns out `SstFileDumper` also requires an `Options` to construct it and thus encounters the same problem, can we reconsider using the already available `SstFileReader`?
> > 
> > 
> > The reason I chose `SstFileDumper` over `SstFileReader` is that it seems to me `SstFileDumper` can better handle older table types and that we can use it off-the-shelf. We could make `SstFileReader` to behave more like `SstFileDumper` but I resisted to do so because I felt it may be unnecessary work. What do you think? :)
> 
> Understood. Can the `SstFileDumper` API be changed to take a `Env*` rather than an `Options`? 

I don't think we could simply pass in `Env*`. Based on my understanding of `SstFileDumper`, the constructor initializes a table reader, where `ImmutableCFOptions` and `MutableCFOptions` are necessary. Moreover, these options may be needed for other member functions in `SstFileDumper` for purposes other than reading table properties.

ajkr:
> > > > > > > I think the only remaining question I have is, why do we need to use `SstFileDumper` and, as a result, also move it out of the tool-specific code? Can we use `ReadTableProperties()` from "table/meta_blocks.h"?
> > > > > > 
> > > > > > 
> > > > > > Actually, `SstFileReader` looks like an even better already existing API for your use case. It handles the setup of file reader for you, and also abstracts the details specific to table type (unlike `ReadTableProperties()`).
> > > > > 
> > > > > 
> > > > > I did consider using `SstFileReader`. The thing is that its constructor needs an `Option`, which is column family dependent. We could use `DB::GetOption`, however, it requires us to know the specific column family (as you mentioned :)
> > > > > It is after considering all the possibilities here that I made the choice to use `SstFileDumper` :P
> > > > 
> > > > 
> > > > Got it. But as it turns out `SstFileDumper` also requires an `Options` to construct it and thus encounters the same problem, can we reconsider using the already available `SstFileReader`?
> > > 
> > > 
> > > The reason I chose `SstFileDumper` over `SstFileReader` is that it seems to me `SstFileDumper` can better handle older table types and that we can use it off-the-shelf. We could make `SstFileReader` to behave more like `SstFileDumper` but I resisted to do so because I felt it may be unnecessary work. What do you think? :)
> > 
> > 
> > Understood. Can the `SstFileDumper` API be changed to take a `Env*` rather than an `Options`?
> 
> I don't think we could simply pass in `Env*`. Based on my understanding of `SstFileDumper`, the constructor initializes a table reader, where `ImmutableCFOptions` and `MutableCFOptions` are necessary. Moreover, these options may be needed for other member functions in `SstFileDumper` for purposes other than reading table properties.

Agreed, the existing code makes it too complicated to do much else besides pass an `Options` with as much info as we think is necessary for the methods we plan to use. It feels a bit fragile as we're constructing a `SstFileDumper` on which only certain methods can work, and in the future the methods that work today may stop working. However, I don't have any proposal to avoid this problem that can be done reasonably quickly. Even if we turn to the lower-level APIs like `ReadTableProperties()`, they still require we provide things like `ImmutableCFOptions`, which we don't have.

gg814:
> > > > > > > > I think the only remaining question I have is, why do we need to use `SstFileDumper` and, as a result, also move it out of the tool-specific code? Can we use `ReadTableProperties()` from "table/meta_blocks.h"?
> > > > > > > 
> > > > > > > 
> > > > > > > Actually, `SstFileReader` looks like an even better already existing API for your use case. It handles the setup of file reader for you, and also abstracts the details specific to table type (unlike `ReadTableProperties()`).
> > > > > > 
> > > > > > 
> > > > > > I did consider using `SstFileReader`. The thing is that its constructor needs an `Option`, which is column family dependent. We could use `DB::GetOption`, however, it requires us to know the specific column family (as you mentioned :)
> > > > > > It is after considering all the possibilities here that I made the choice to use `SstFileDumper` :P
> > > > > 
> > > > > 
> > > > > Got it. But as it turns out `SstFileDumper` also requires an `Options` to construct it and thus encounters the same problem, can we reconsider using the already available `SstFileReader`?
> > > > 
> > > > 
> > > > The reason I chose `SstFileDumper` over `SstFileReader` is that it seems to me `SstFileDumper` can better handle older table types and that we can use it off-the-shelf. We could make `SstFileReader` to behave more like `SstFileDumper` but I resisted to do so because I felt it may be unnecessary work. What do you think? :)
> > > 
> > > 
> > > Understood. Can the `SstFileDumper` API be changed to take a `Env*` rather than an `Options`?
> > 
> > 
> > I don't think we could simply pass in `Env*`. Based on my understanding of `SstFileDumper`, the constructor initializes a table reader, where `ImmutableCFOptions` and `MutableCFOptions` are necessary. Moreover, these options may be needed for other member functions in `SstFileDumper` for purposes other than reading table properties.
> 
> Agreed, the existing code makes it too complicated to do much else besides pass an `Options` with as much info as we think is necessary for the methods we plan to use. It feels a bit fragile as we're constructing a `SstFileDumper` on which only certain methods can work, and in the future the methods that work today may stop working. However, I don't have any proposal to avoid this problem that can be done reasonably quickly. Even if we turn to the lower-level APIs like `ReadTableProperties()`, they still require we provide things like `ImmutableCFOptions`, which we don't have.

Yes, I have the same feeling and it bothered me quite a bit. I think RocksDB should have some kind of unified entrance or logic for external access of DB in the future.

gg814:
Thanks for your careful review! It helped me learn :)

I will implement the two unresolved suggestions (option name for the new naming scheme and setting the proper `env` for `Options`) in my follow-up.

PR #7002, opened 2020-06-18
Add FileSystem wrapper classes for IO tracing.

akankshamahajan15:
Summary: 1. Add the wrapper classes FileSystemTracingWrapper, FSSequentialFileTracingWrapper, FSRandomAccessFileTracingWrapper, FSWritableFileTracingWrapper, FSRandomRWFileTracingWrapper that forward the calls to underlying storage system and then pass the file operation information to IOTracer. IOTracer dumps the record in binary format for tracing.
2. Add the wrapper classes FileSystemPtr, FSSequentialFilePtr, FSRandomAccessFilePtr, FSWritableFilePtr and FSRandomRWFilePtr that overload operator-> and return ptr to underlying storage system or Tracing wrapper class based on enabling/disabling of IO tracing. These classes are added to bypass Tracing Wrapper classes when we disable tracing.
3. Add enums in trace.h that distinguish which options need to be added for different file operations(Read, close, write etc) as part of tracing record.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/env/file_system_tracer.cc
+++ b/env/file_system_tracer.cc
@@ -1,0 +1,319 @@
+// Copyright (c) 2019-present, Facebook, Inc.  All rights reserved.
+// This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#include "env/file_system_tracer.h"
+
+#include "rocksdb/env.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+IOStatus FileSystemTracingWrapper::NewWritableFile(
+    const std::string& fname, const FileOptions& file_opts,
+    std::unique_ptr<FSWritableFile>* result, IODebugContext* dbg) {
+  StopWatchNano timer(Env::Default());
+  timer.Start();
+  IOStatus s = target()->NewWritableFile(fname, file_opts, result, dbg);
+  uint64_t elapsed = timer.ElapsedNanos();
+  IOTraceRecord io_record(elapsed, TraceType::kIOFileName, __func__, elapsed,
+                          s.ToString(), fname);
+  io_tracer_->WriteIOOp(io_record);
+  return s;
+}
+
+IOStatus FileSystemTracingWrapper::NewDirectory(
+    const std::string& name, const IOOptions& io_opts,
+    std::unique_ptr<FSDirectory>* result, IODebugContext* dbg) {
+  StopWatchNano timer(Env::Default());
+  timer.Start();
+  IOStatus s = target()->NewDirectory(name, io_opts, result, dbg);
+  uint64_t elapsed = timer.ElapsedNanos();
+  IOTraceRecord io_record(env_->NowNanos(), TraceType::kIOFileName, __func__,
+                          elapsed, s.ToString(), name);
+  io_tracer_->WriteIOOp(io_record);
+  return s;
+}
+
+IOStatus FileSystemTracingWrapper::GetChildren(const std::string& dir,
+                                               const IOOptions& io_opts,
+                                               std::vector<std::string>* r,
+                                               IODebugContext* dbg) {
+  StopWatchNano timer(Env::Default());
+  timer.Start();
+  IOStatus s = target()->GetChildren(dir, io_opts, r, dbg);
+  uint64_t elapsed = timer.ElapsedNanos();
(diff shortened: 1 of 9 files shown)


env/file_system_tracer.cc
```diff
@@ -0,0 +1,253 @@
...
+  io_tracer_->WriteIOOp(io_record);
+  return s;
+}
+
+IOStatus FSRandomAccessFileTracingWrapper::Read(uint64_t offset, size_t n,
+                                                const IOOptions& options,
+                                                Slice* result, char* scratch,
+                                                IODebugContext* dbg) const {
+  IOStatus s = target()->Read(offset, n, options, result, scratch, dbg);
+  IOTraceRecord io_record(Timestamp(), TraceType::kIOOffsetAndLen, "Read",
```
zhichao-cao, line 137: When the API like Read, Append, are we able to know which file is accessed by this request?

trace_replay/trace_replay.h
```diff
@@ -46,6 +46,14 @@ enum TraceType : char {
   kBlockTraceDataBlock = 9,
   kBlockTraceUncompressionDictBlock = 10,
   kBlockTraceRangeDeletionBlock = 11,
+  // IO Trace related types based on options that will be added in trace file.
+  kIOGeneral = 12,
+  kIOFileName = 13,
```
zhichao-cao, line 51: I'm thinking we can put all the queries into two categories: 1) file related (e.g., NewWritableFile, Sync, Close) and 2) data related (e.g., Read, Write, Append, Truncate). For all the queries, type, timestamp, filename are the standard contents in the record. The rest of the payload is depends on the query types. Otherwise, when we replay or analyze the trace in the further, we may not be able to use the trace records in a precise way.

akankshamahajan15: Yes Make sense. I was thinking in terms of that but then in IOtracer I need to check what options to write and what not because even same Functions (like GetFileSize) in different wrappers have different options (One passes filename and one doesn't). I will check once again if I can make it consistent and pass missing options that way it would be more scalable.

zhichao-cao: Oh, I see. I did realize that the options make it more complex.

env/file_system_tracer.cc
akankshamahajan15: For some cases we can. I am not sure about all because its being used at multiple instances. My next PR that i am working on is to cache FileSystemPtr and other Ptrs instead of storing FileSystem and other pointers. So while making those changes i will have better idea where can I pass the file name as part of IODebugContext. I will add the filename in the next PR.

zhichao-cao: I see, that will be great.

akankshamahajan15:
Addressed comments

env/file_system_tracer.cc
```diff
@@ -0,0 +1,306 @@
...
+uint64_t Timestamp() {
+  return duration_cast<microseconds>(system_clock::now().time_since_epoch())
+      .count();
+}
+}  // namespace
+
+IOStatus FileSystemTracingWrapper::NewWritableFile(
+    const std::string& fname, const FileOptions& file_opts,
+    std::unique_ptr<FSWritableFile>* result, IODebugContext* dbg) {
+  uint64_t start_time = Timestamp();
```
anand1976, line 22: For consistency sake, we can use ```StopWatch```, like here https://github.com/facebook/rocksdb/blob/master/db/db_impl/db_impl_write.cc#L1404. What do you think?

akankshamahajan15: ```StopWatch``` requires Env pointer and in ```file_system_tracer.cc``` I don't have Env. 
It also needs statistics pointer for ```StopWatch```.

I can use ```StopWatchNano``` also as it requires ```Env``` not ```Statistics```.  Should I create Env object with ```Env::Default()``` in ```file_system_tracer.cc```?

akankshamahajan15:
Resolved Conflicts

env/file_system_tracer.cc
anand1976: Yes, ```StopWatchNano``` with ```Env::Default()``` should be okay. I'm not sure if using ```StopWatch/StopWatchNano``` will make it more portable, but rest of the codebase uses those, so I think its better to be consistent.

akankshamahajan15:
Addressed comments

PR #7005, opened 2020-06-19
Fix unread variable in db_stress

zhichao-cao:
remove the variable.
Pass make asan_check, db_stress


zhichao-cao:
> clang-analyze failure.

clang-analyze failure, I just be used to run asan_check each time:)

PR #7007, opened 2020-06-19
Fix double define in IO_tracer

zhichao-cao:
Fix the following error

"./trace_replay/io_tracer.h:20:20: error: redefinition of ‘const unsigned int rocksdb::{anonymous}::kCharSize’
 const unsigned int kCharSize = 1;
                    ^~~~~~~~~
In file included from unity.cc:177:
trace_replay/block_cache_tracer.cc:22:20: note: ‘const unsigned int rocksdb::{anonymous}::kCharSize’ previously defined here
 const unsigned int kCharSize = 1;"

### DIFF
--- a/trace_replay/io_tracer.cc
+++ b/trace_replay/io_tracer.cc
@@ -18,6 +18,10 @@
 
 namespace ROCKSDB_NAMESPACE {
 
+namespace {
+const unsigned int kCharSize = 1;
+}  // namespace
+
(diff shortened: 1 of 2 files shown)


zhichao-cao:
> Thanks for correcting this one. I never got any kind of error in any of the builds.

Thanks for the review. I guess the internal run will use different compile flags, that's why in our devserver, we did not have such errors.

PR #7008, opened 2020-06-19
Remove racially charged terms "whitelist" and "blacklist"

pdillinger:
Summary: We don't need them.

Test Plan: "make check" and ensure "make crash_test" starts

### DIFF
--- a/db_stress_tool/db_stress_gflags.cc
+++ b/db_stress_tool/db_stress_gflags.cc
@@ -414,10 +414,10 @@ static const bool FLAGS_kill_random_test_dummy __attribute__((__unused__)) =
     RegisterFlagValidator(&FLAGS_kill_random_test, &ValidateInt32Positive);
 extern int rocksdb_kill_odds;
 
-DEFINE_string(kill_prefix_blacklist, "",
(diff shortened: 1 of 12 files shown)


db_stress_tool/db_stress_gflags.cc
```diff
@@ -414,10 +414,10 @@ static const bool FLAGS_kill_random_test_dummy __attribute__((__unused__)) =
     RegisterFlagValidator(&FLAGS_kill_random_test, &ValidateInt32Positive);
 extern int rocksdb_kill_odds;
 
-DEFINE_string(kill_prefix_blacklist, "",
+DEFINE_string(kill_exclude_prefixes, "",
```
mrambacher, line 417: Unfortunately, this one is effectively in the public API as it is an option to the tool.  I do not know what the policy is on preserving command line options or if this requires an update to the public API documentation (HISTORY.md).

This option shows up via db_stress -help currently...

pdillinger: db_stress is a development tool, not a customer tool, so subject to change. Help is autogenerated.

PR #7009, opened 2020-06-22
freebsd: malloc_usable_size check malloc_np.h

grooverdan:
Per https://www.unix.com/man-page/freebsd/3/malloc_usable_size/
malloc_usable_size is in malloc_np.h as its a non-standard API.

Without patch it just fails to detect from ./CMakeFiles/CMakeError.log

In file included from /home/dan/build-rocksdb/CMakeFiles/CMakeTmp/CheckSymbolExists.cxx:2:
/usr/include/malloc.h:3:2: error: "<malloc.h> has been replaced by <stdlib.h>"
 ^
/home/dan/build-rocksdb/CMakeFiles/CMakeTmp/CheckSymbolExists.cxx:8:19: error: use of undeclared identifier 'malloc_usable_size'
  return ((int*)(&malloc_usable_size))[argc];
                  ^
2 errors generated.


grooverdan:
Missing `malloc_usable_size` on kFreeBSD would still be a regression so updated as suggest and noted it in the commit message.

PR #7010, opened 2020-06-22
filelock_test: add freebsd headers for waitpid

grooverdan:
Per manual https://www.unix.com/man-page/FreeBSD/2/waitpid


grooverdan:
Is the FB internal warning contain something that need correcting?

grooverdan:
Thank you.

PR #7011, opened 2020-06-22
gflags: freebsd include path + links

grooverdan:
The include path from find_package(gflags) needed to be included to
compile.

Because gflags got included in THIRDPARTY_LIBS as a PRIVATE library
to ROCKSDB_{SHARED|STATIC}_LIB, its functions aren't accessible to
the all the tools an utilities that use gflags directly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,6 +107,7 @@ else()
     endif()
   endif()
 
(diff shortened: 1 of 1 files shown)


grooverdan:
Thanks

PR #7013, opened 2020-06-22
Replace tracked_keys with a new LockTracker interface in TransactionDB

cheng-chang:
We're going to support more locking protocols such as range lock in transaction.

However, in current design, `TransactionBase` has a member `tracked_keys` which assumes that point lock (lock a single key) is used, and is used in snapshot checking (isolation protocol). When using range lock, we may use read committed instead of snapshot checking as the isolation protocol.

The most significant usage scenarios of `tracked_keys` are:
1. pessimistic transaction uses it to track the locked keys, and unlock these keys when commit or rollback.
2. optimistic transaction does not lock keys upfront, it only tracks the lock intentions in tracked_keys, and do write conflict checking when commit.
3. each `SavePoint` tracks the keys that are locked since the `SavePoint`, `RollbackToSavePoint` or `PopSavePoint` relies on both the tracked keys in `SavePoint`s and `tracked_keys`.

Based on these scenarios, if we can abstract out a `LockTracker` interface to hold a set of tracked locks (can be keys or key ranges), and have methods that can be composed together to implement the scenarios, then `tracked_keys` can be an internal data structure of one implementation of `LockTracker`. See `utilities/transactions/lock/lock_tracker.h` for the detailed interface design, and `utilities/transactions/lock/point_lock_tracker.cc` for the implementation. 

In the future, a `RangeLockTracker` can be implemented to track range locks without affecting other components.

After this PR, a clean interface for lock manager should be possible, and then ideally, we can have pluggable locking protocols.

Test Plan:
Run `transaction_test` and `optimistic_transaction_test`.


cheng-chang:
I plan to add a unit test for `PointLockTracker` in a follow up PR, in this PR, let's focus on the design of the interface.

utilities/transactions/optimistic_transaction.cc
```diff
@@ -97,9 +97,21 @@ Status OptimisticTransaction::CommitWithParallelValidate() {
...
   std::vector<std::unique_lock<std::mutex>> lks;
-  for (auto& cfit : GetTrackedKeys()) {
-    for (auto& keyit : cfit.second) {
-      lk_idxes.insert(fastrange64(GetSliceNPHash64(keyit.first), space));
+  std::unique_ptr<LockTracker::ColumnFamilyIterator> cf_it(
+      tracked_locks_->GetColumnFamilyIterator());
+  if (!cf_it) {
+    return Status::InvalidArgument(
+        "Lock tracker cannot iterate through column families, "
+        "must use a lock track that supports point lock");
```
lth, line 105: "lock tracker"

utilities/transactions/transaction_lock_mgr.cc
```diff
@@ -643,26 +643,27 @@ void TransactionLockMgr::UnLock(PessimisticTransaction* txn,
...
-                                const TransactionKeyMap* key_map, Env* env) {
-  for (auto& key_map_iter : *key_map) {
-    uint32_t column_family_id = key_map_iter.first;
-    auto& keys = key_map_iter.second;
-
-    std::shared_ptr<LockMap> lock_map_ptr = GetLockMap(column_family_id);
+                                const LockTracker& tracker, Env* env) {
+  std::unique_ptr<LockTracker::ColumnFamilyIterator> cf_it(
+      tracker.GetColumnFamilyIterator());
+  if (!cf_it) return;
```
lth, line 649: Maybe there should be an assertion.

cheng-chang:
good point, tracked keys are used in snapshot checking, but not necessarily assumed that snapshot checking must be used.

utilities/transactions/lock/lock_tracker.h
```diff
@@ -0,0 +1,193 @@
...
+   public:
+    virtual ~ColumnFamilyIterator() {}
+
+    // Whether there are remaining column families.
+    virtual bool HasNext() const = 0;
+
+    // Gets next column family id.
+    //
+    // If HasNext is false, calling this method has undefined behavior.
+    virtual ColumnFamilyId Next() = 0;
```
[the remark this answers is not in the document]
cheng-chang: the implementation might need to modify internal data, such as increasing an index.

utilities/transactions/lock/lock_tracker.h
```diff
@@ -0,0 +1,193 @@
...
+   public:
+    virtual ~KeyIterator() {}
+
+    // Whether there are remaining keys.
+    virtual bool HasNext() const = 0;
+
+    // Gets the next key.
+    //
+    // If HasNext is false, calling this method has undefined behavior.
+    virtual const std::string& Next() = 0;
```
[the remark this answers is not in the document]
cheng-chang: the implementation might need to modify internal data, such as increasing an index.

utilities/transactions/lock/lock_tracker.h
```diff
@@ -0,0 +1,193 @@
...
+
+#include <memory>
+
+#include "rocksdb/rocksdb_namespace.h"
+#include "rocksdb/status.h"
+#include "rocksdb/types.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+using ColumnFamilyId = uint32_t;
```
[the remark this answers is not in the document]
cheng-chang: I created a task to track this.

utilities/transactions/transaction_lock_mgr.cc
cheng-chang: The design was that if point lock is not used, then `GetColumnFamilyIterator()` might return `nullptr`.
But after second thought, I forced the API to never return `nullptr`, if there is no column family to iterate, just return an empty iterator.

utilities/transactions/lock/point_lock_tracker.cc
```diff
@@ -0,0 +1,269 @@
...
+
+ private:
+  const TrackedKeyInfos& key_infos_;
+  TrackedKeyInfos::const_iterator it_;
+};
+
+LockTracker::KeyIterator* PointLockTracker::GetKeyIterator(
+    ColumnFamilyId column_family_id) const {
+  if (tracked_keys_.find(column_family_id) == tracked_keys_.end()) {
+    return new EmptyKeysIterator();
```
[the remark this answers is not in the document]
cheng-chang: that situation is rare I think. 
The defense here is to ensure that the returned iterator works even when the provided cf_id does not exist.

utilities/transactions/lock/lock_tracker.h
```diff
@@ -0,0 +1,193 @@
...
+  // lock.
+  //
+  // The second returned bool will be true only if the first bool is true and
+  // the tracked lock is removed from the tracker completely, sometimes, only
+  // the lock is still tracked because previously the same lock is tracked
+  // multiple times.
+  //
+  // If this method is not supported, leave it as a no-op and
+  // returns {false, false}.
+  virtual std::pair<bool, bool> Untrack(
```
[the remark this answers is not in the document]
cheng-chang: replaced with `UntrackStatus`.

utilities/transactions/lock/point_lock_tracker.cc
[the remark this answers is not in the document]
lth: OK, in these cases, I wonder if an assertion would be useful, so that readers aren't left wondering if this codepath is ever exercised.

utilities/transactions/transaction_lock_mgr.cc
lth: I see, I still feel like we should add a debug assertion somewhere, since this is something that's not supposed to happen.

utilities/transactions/lock/point_lock_tracker.cc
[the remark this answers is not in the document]
cheng-chang: updated the API contract to enforce column_family_id to exist, and changed to assertion here.

utilities/transactions/transaction_lock_mgr.cc
cheng-chang: already added `assert(cf_it != nullptr)`

utilities/transactions/pessimistic_transaction.cc
```diff
@@ -548,28 +555,9 @@ Status PessimisticTransaction::TryLock(ColumnFamilyHandle* column_family,
...
-      if (!iter->second.exclusive && exclusive) {
-        lock_upgrade = true;
-      }
-      previously_locked = true;
-      tracked_at_seq = iter->second.seq;
-    }
-  }
+  PointLockStatus status = tracked_locks_->GetPointLockStatus(cfh_id, key_str);
+  bool previously_locked = status.locked;
+  bool lock_upgrade = exclusive && !status.exclusive;
```
spetrunia, line 560: Does the value of `lock_upgrade` have any meaning if `status.locked==false`?  Common sense shows that it shouldn't, but below one can see  this code:
```c++
        if (!previously_locked) {
          // Unlock key we just locked
          if (lock_upgrade) {
```
that is, is analyzed when `previously_locked==FALSE`... That is, `status.locked==false`...

utilities/transactions/lock/lock_tracker.cc
```diff
@@ -0,0 +1,17 @@
...
+// COPYING file in the root directory) and Apache 2.0 License
+// (found in the LICENSE.Apache file in the root directory).
+
+#include "utilities/transactions/lock/lock_tracker.h"
+
+#include "utilities/transactions/lock/point_lock_tracker.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+LockTracker* NewLockTracker() {
```
spetrunia, line 12: The function will need some context to know what kind of lock tracker to create  (AFAIU there's no global variable pointing which kind of lock manager is used). 
When trying to implement range locking for this, I ended up with this: https://gist.github.com/spetrunia/66c41c8a2c41d5f3f4bae27d505181fb .  Here, the Lock Manager has a function that returns a factory object that creates LockTracker objects that are suitable for use with this Lock Manager. Does this make sense?

utilities/transactions/write_unprepared_txn.h
```diff
@@ -212,7 +212,7 @@ class WriteUnpreparedTxn : public WritePreparedTxn {
   friend class WriteUnpreparedTxnDB;
 
   const std::map<SequenceNumber, size_t>& GetUnpreparedSequenceNumbers();
-  Status WriteRollbackKeys(const TransactionKeyMap& tracked_keys,
+  Status WriteRollbackKeys(const LockTracker& tracked_keys,
                            WriteBatchWithIndex* rollback_batch,
                            ReadCallback* callback, const ReadOptions& roptions);
```
spetrunia, line 217: Do I understand this code correctly: it walks the tracked_keys to get a set of rowkeys that were modified by the transaction.

I'm wondering if/how this should work with a Range-based Lock Tracker (RBLT)?

Should RBLT be able to return the set of modified rowkeys? (What if it returns a set of ranges that cover all the rowkeys instead? I assume that won't work?)

(I'll need to verify this, but I suspect that RangeLocking doesn't work with WriteUnprepared in the current Range-Locking code... when range_locking=1, `tracked_keys` is empty even if the transaction has made modifications...)

spetrunia: Found `WriteUnpreparedTxn::untracked_keys_` ... this looks like it could be used as a workaround...  (it would be a lousy one:  Range Locking is capable of "Lock Escalation" where adjacent locks are merged together and so a big transaction can be run without consuming O(#size-of-changes) RAM... if we just put all modified row keys into untracked_keys_ we'll get the O(#size-of-changes) memory consumption back)

utilities/transactions/lock/lock_tracker.h
```diff
@@ -0,0 +1,193 @@
...
+  // current tracker.
+  // The tracked locks in the specified tracker must be a subset of those
+  // tracked by the current tracker.
+  virtual LockTracker* GetTrackedLocksSinceSavePoint(
+      const LockTracker& /*save_point_tracker*/) const = 0;
+
+  // Gets lock related information of the key.
+  //
+  // If point lock is not supported, always returns LockStatus with
+  // locked=false.
```
spetrunia, line 135: Is this really the case? 
Suppose there is LockTracker which supports ranges and contains a range that covers point P .  Then, GetPointLockStatus(P) should return locked=false?
(I myself don't have a ready answer to this question. I'm getting to this but I'm not there yet)

utilities/transactions/write_unprepared_txn.h
spetrunia: @lth , Herman mentioned you could enlighten us here (on the questions in the first comment)?

utilities/transactions/lock/lock_tracker.h
```diff
@@ -0,0 +1,193 @@
...
+   public:
+    virtual ~KeyIterator() {}
+
+    // Whether there are remaining keys.
+    virtual bool HasNext() const = 0;
+
+    // Gets the next key.
+    //
+    // If HasNext is false, calling this method has undefined behavior.
+    virtual const std::string& Next() = 0;
```
ajkr, line 176: Is `TrackedKeyInfo` a generic structure that will be used by all imaginable implementations? If so, can it return `pair<const std::string&, const TrackedKeyInfo&>` (or, a separate iterator that returns `const TrackedKeyInfo&`s)? The benefit of providing capability to iterate over `TrackedKeyInfo` in the parent class, even if only exposed with protected visibility, is we can use that in the derived classes' `Merge()`, `Subtract()`, etc. Then we can remove the requirement that the user provides a specific type as we can access the tracked keys through a common interface rather than through the derived class's instance variable.

utilities/transactions/lock/point_lock_tracker.h
```diff
@@ -0,0 +1,80 @@
...
+  PointLockTracker() = default;
+
+  PointLockTracker(const PointLockTracker&) = delete;
+  PointLockTracker& operator=(const PointLockTracker&) = delete;
+
+  void Track(const PointLockRequest& lock_request) override;
+
+  UntrackStatus Untrack(const PointLockRequest& lock_request) override;
+
+  void Track(const RangeLockRequest& /*lock_request*/) override {}
```
ajkr, line 52: Maybe we should have a `IsRangeLockSupported()` function so clients can know whether this has any effect.

utilities/transactions/lock/lock_tracker.h
ajkr: Agreed, I think conflict checking should use a function that checks for both range locks and point locks. If other parts of the code require a function that specifically returns coverage of point locks only, perhaps we can have two separate functions.

utilities/transactions/write_unprepared_txn.h
lth: I think it helps to give some context on the goals of write unprepared. The goal of write unprepared is to reduce memory usage by large transactions. Today there are 3 ways memory would grow with size of transaction:
1. write batch
2. tracked keys
3. lock map

The solution to #1 was to flush the write batch into the db periodically and this is done. We still don't have an implemented solution for #2/#3, but the idea is that we could approximate many point locks by merging into a single range lock and save on memory for #2/#3. 

So we would use tracked_keys for a few purposes in write unprepared:
1. to get the list of keys to rollback 
2. to get the list of keys to unlock

I'm guessing the question is how #1 would work in context of range locks. I think there were two ways we were planning to go with 1. reading the list of keys to rollback from the WAL or 2. doing 2 snapshot reads (one at latest, one at the seqno to rollback to) in the tracked ranges to determine the delta in changed keys, and then rolling back the modified keys.

Right now, I think reusing tracked keys would be cleaner than reading from WAL, but this remains an open question. I think this would depend on the performance penalty of using range locking.

spetrunia:
Ok, I've found another issue. Strictly speaking, it is not in this patch, but since this patch is trying "abstract away" the difference between Point and Range-based lock managers, here it is:

Current PointLockManager can provide an efficient implementation of

```
  GetPointLockStatus(cfh_id, key_str);
```
but the RangeLockManager cannot. This is because it stores the acquired locks in a growable array. The array is unordered and will contain overlapping locks if the transaction has requested the same lock multiple times.
(I guess this is assumed to be rare, and in the worst case this causes a duplicate lock release call when the transaction is finalized).

I'm not yet sure which way this should be solved:
- Should I make RangeLockManager store transaction's locks in a data structure that allows for searches? This will add some CPU overhead.
- Or, should the code in PessimisticTransaction::TryLock() be made to avoid (non-debug) calls to GetPointLockStatus...

lth:
I've wondered if there was any value in keeping point locks searchable (since it consumes noticeable cpu in write heavy workloads). It seems to optimize for the case where the same key is locked multiple times. I don't think we need it if we have recursive locks.

This would also simplify our rollback to savepoint algorithm, as we could just unlock everything in the batch we're dropping.

utilities/transactions/pessimistic_transaction.cc
cheng-chang: good catch. I updated logic of `lock_upgrade` and the below block.

utilities/transactions/lock/lock_tracker.cc
cheng-chang: Yes, it makes sense.
During prototyping, originally there is a method like `bool SupportsPointLock()`, `bool SupportsRangeLock()`.
But I thought the lock manager interface can hold a LockTracker, like what you did. But I didn't do this in the current PR because I planned to do this in a future PR for new lock manager interface.

utilities/transactions/lock/lock_tracker.h
cheng-chang: This function returns true only when the key is point locked.
If we need the semantics of a key is covered by a range lock, we can have another function.

utilities/transactions/lock/lock_tracker.h
cheng-chang: Currently, `TrackedKeyInfo` is an implementation detail of point lock tracker.
If we find it to be useful in other lock tracker implementations in the future, we can add another iterator interface to return TrackedKeyInfos.

utilities/transactions/lock/lock_tracker.h
cheng-chang: Added `assert(IsPointLockSupported());` in the implementation.

utilities/transactions/lock/point_lock_tracker.h
cheng-chang: added `IsPointLockSupported()` and `IsRangeLockSupported()`, although only `IsPointLockSupported()` used in an assertion, future lock tracker implementations may find these methods useful.

utilities/transactions/lock/lock_tracker.cc
cheng-chang: Added `IsPointLockSupported()` and `IsRangeLockSupported()`, hope they can be useful in range lock tracker.

cheng-chang:
> I've wondered if there was any value in keeping point locks searchable (since it consumes noticeable cpu in write heavy workloads). It seems to optimize for the case where the same key is locked multiple times. I don't think we need it if we have recursive locks.
> 
> This would also simplify our rollback to savepoint algorithm, as we could just unlock everything in the batch we're dropping.

@lth Good idea. For this PR to make progress, let's focus on the interface design in this PR, we can have further PRs for optimizations.

cheng-chang:
> It makes sense to encapsulate the lock tracking logic in a separate class. It also makes sense to me to make an interface to enable different lock tracking structures/algorithms. I'm not sure I understood the plan to implement different types of locks in different subclasses, though. For range and point locking, do they need to be implemented separately while using a common interface? It looks like the common interface then has to include functions for locking both points and ranges, so each implementation can cover only a subset of the interface.

Yes, there will be a lock manager interface,  @spetrunia 's range lock PR https://github.com/facebook/rocksdb/pull/5041 works on this. Some lock manager implementations may only cover a subset of methods, and that's expected.

> 
> couple more questions:
> 
> * is there a plan to change `TransactionLockMgr` to support ranges?

Yes, @spetrunia 's range lock PR https://github.com/facebook/rocksdb/pull/5041 works on this.

> * do you expect a lot of overlapping ranges? Range delete has some code for fast lookup in such situations. It may be reusable.

The current range lock implementation in https://github.com/facebook/rocksdb/pull/5041 borrows implementation from TokuDB. The implementation has something overlap with existing RocksDB utilities, such as some posix utilities, and maybe some utilities related to ranges. I worry less about these overlap for now as long as the range lock implementation is in a standalone directory and no functions are exported outside of the directory.

cheng-chang:
> Ok, I've found another issue. Strictly speaking, it is not in this patch, but since this patch is trying "abstract away" the difference between Point and Range-based lock managers, here it is:
> 
> Current PointLockManager can provide an efficient implementation of
> 
> ```
>   GetPointLockStatus(cfh_id, key_str);
> ```
> 
> but the RangeLockManager cannot. This is because it stores the acquired locks in a growable array. The array is unordered and will contain overlapping locks if the transaction has requested the same lock multiple times.
> (I guess this is assumed to be rare, and in the worst case this causes a duplicate lock release call when the transaction is finalized).
> 
> I'm not yet sure which way this should be solved:
> 
> * Should I make RangeLockManager store transaction's locks in a data structure that allows for searches? This will add some CPU overhead.
> * Or, should the code in PessimisticTransaction::TryLock() be made to avoid (non-debug) calls to GetPointLockStatus...
> * (EDIT) perhaps option 3 is to make RangeLockManager::GetPointLockStatus always return NotTracked (will get back on this)

@spetrunia Can we determine whether to call `GetPointLockStatus` in `TryLock` by checking the new method `IsPointLockSupported`?

cheng-chang:
@spetrunia @lth @ajkr Thanks for the review!
@spetrunia I don't think we can fully "abstract away" the differences between point lock and range lock trackers in this interface. Implementations such as range lock may still need to know whether point lock or range lock is supported, so I added back the methods `IsPointLockSupported()` and `IsRangeLockSupported()`.

Based on current reviews and @spetrunia 's experiment with the current interface, it looks like the interface works (although there are definitely rooms to improve), if you all think it's the right direction to go, I hope we can get this in ASAP to make progress on range lock. Thanks!

spetrunia:
> @spetrunia Can we determine whether to call GetPointLockStatus in TryLock by checking the new method IsPointLockSupported?

I'm checking this

spetrunia:
@cheng-chang 
> @spetrunia Can we determine whether to call GetPointLockStatus in TryLock by checking the new method IsPointLockSupported?

OK, the answer is Yes. I was able to use that call and unify the code in PessimisticTransaction::TryLock.
(in case somebody's interested: https://github.com/spetrunia/rocksdb/blob/spetrunia-range-locking-rocksdb-pr2-locktrack/utilities/transactions/pessimistic_transaction.cc#L565)


The remaining question is about the code in WriteUnpreparedTxn::WriteRollbackKeys().

With Range Locking, that code is not functional. WriteUnprepared is [ab]using lock tracker to get the set of modified keys (in order to write undo commands into the WAL). Range Locking's LockTracker cannot provide this.

Fortunately, WriteUnprepare has a (temporary) mechanism to also undo changes to the keys that were not noted in the LockTracker. It is `untracked_keys_`, https://github.com/facebook/rocksdb/commit/dc23c125c30f9b865ed9663b5d9d6d6ae6548403.

Now, let's look at the new code in WriteUnpreparedTxn::WriteRollbackKeys().

It has calls to GetColumnFamilyIterator() and GetKeyIterator(), which are hard to implement meaningfully for RangeLockTracker, without resorting to ugly things like "pretend there are no keys".  I think the easiest solution is to wrap the iteration code in WriteUnpreparedTxn::WriteRollbackKeys() in `if (tracker.IsPointLockSupported()) {  }`. Should this be done in the scope of this PR or range locking PR?

Other than this, this PR looks good.

cheng-chang:
> @cheng-chang
> 
> > @spetrunia Can we determine whether to call GetPointLockStatus in TryLock by checking the new method IsPointLockSupported?
> 
> OK, the answer is Yes. I was able to use that call and unify the code in PessimisticTransaction::TryLock.
> (in case somebody's interested: https://github.com/spetrunia/rocksdb/blob/spetrunia-range-locking-rocksdb-pr2-locktrack/utilities/transactions/pessimistic_transaction.cc#L565)
> 
> The remaining question is about the code in WriteUnpreparedTxn::WriteRollbackKeys().
> 
> With Range Locking, that code is not functional. WriteUnprepared is [ab]using lock tracker to get the set of modified keys (in order to write undo commands into the WAL). Range Locking's LockTracker cannot provide this.
> 
> Fortunately, WriteUnprepare has a (temporary) mechanism to also undo changes to the keys that were not noted in the LockTracker. It is `untracked_keys_`, [dc23c12](https://github.com/facebook/rocksdb/commit/dc23c125c30f9b865ed9663b5d9d6d6ae6548403).
> 
> Now, let's look at the new code in WriteUnpreparedTxn::WriteRollbackKeys().
> 
> It has calls to GetColumnFamilyIterator() and GetKeyIterator(), which are hard to implement meaningfully for RangeLockTracker, without resorting to ugly things like "pretend there are no keys". I think the easiest solution is to wrap the iteration code in WriteUnpreparedTxn::WriteRollbackKeys() in `if (tracker.IsPointLockSupported()) { }`. Should this be done in the scope of this PR or range locking PR?
> 
> Other than this, this PR looks good.

@spetrunia I wrapped the WriteRollbackKeys in `if (tracker.IsPointLockSupported()) { }` in this PR. PTAL.

utilities/transactions/write_unprepared_txn.cc
```diff
@@ -666,13 +668,21 @@ Status WriteUnpreparedTxn::WriteRollbackKeys(
...
   };
 
-  for (const auto& cfkey : tracked_keys) {
-    const auto cfid = cfkey.first;
-    const auto& keys = cfkey.second;
-    for (const auto& pair : keys) {
-      auto s = WriteRollbackKey(pair.first, cfid);
-      if (!s.ok()) {
-        return s;
+  if (lock_tracker.IsPointLockSupported()) {
```
lth, line 671: If both range lock and write unprepared is used together, today it doesn't work right? Is it better to check that the options aren't used together, and assert here that we're using point locks?

cheng-chang: @lth I changed this to an assertion. @spetrunia When write unprepared supports range lock in the range lock PR, feel free to remove this assertion.

utilities/transactions/pessimistic_transaction.cc
```diff
@@ -614,15 +604,13 @@ Status PessimisticTransaction::TryLock(ColumnFamilyHandle* column_family,
...
-          if (lock_upgrade) {
-            s = txn_db_impl_->TryLock(this, cfh_id, key_str,
-                                      false /* exclusive */);
-            assert(s.ok());
-          } else {
-            txn_db_impl_->UnLock(this, cfh_id, key.ToString());
-          }
+        // Unlock key we just locked
+        if (lock_upgrade) {
+          s = txn_db_impl_->TryLock(this, cfh_id, key_str,
```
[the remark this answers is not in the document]
cheng-chang: `TryLock` can downgrade the lock. The exclusiveness is achieved by a field in `LockInfo`, see https://github.com/facebook/rocksdb/blob/master/utilities/transactions/transaction_lock_mgr.cc#L536

utilities/transactions/write_unprepared_txn.cc
spetrunia: @lth, AFAIU:  RangeLocking+WriteUnprepared didn't work together until this commit (by you) https://github.com/facebook/rocksdb/commit/dc23c125c30f9b865ed9663b5d9d6d6ae6548403 fixed this. Now, they should be able to work together, although this doesn't have any test coverage atm.

PR #7014, opened 2020-06-22
`BackupEngine::VerifyBackup` verifies checksum by default

gg814:
Summary:
A parameter `verify_with_checksum` is added to `BackupEngine::VerifyBackup`, which is true by default. So now `BackupEngine::VerifyBackup` verifies backup files with checksum AND file size by default. When `verify_with_checksum` is false, `BackupEngine::VerifyBackup` only compares file sizes to verify backup files.

Also add a test for the case when corruption does not change the file size.

Test plan:
Passed make check

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,7 +13,7 @@
 ### Public API Change
 * `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the LOG file on the line starting with "DB Session ID:".
 * `DB::OpenForReadOnly()` now returns `Status::NotFound` when the specified DB directory does not exist. Previously the error returned depended on the underlying `Env`. This change is available in all 6.11 releases as well.
-* A parameter `verify_with_checksum` is added to `BackupEngine::VerifyBackup`, which is true by default. So now `BackupEngine::VerifyBackup` verifies backup files with checksum and file size by default. When `verify_with_checksum` is false, `BackupEngine::VerifyBackup` only compares file sizes to verify backup files.
+* A parameter `verify_with_checksum` is added to `BackupEngine::VerifyBackup`, which is true by default. So now `BackupEngine::VerifyBackup` verifies checksums and file sizes of backup files by default. Pass `false` for `verify_with_checksum` to maintain the previous behavior and performance of `BackupEngine::VerifyBackup`, by only verifying sizes of backup files.
 
 ### New Features
 * DB identity (`db_id`) and DB session identity (`db_session_id`) are added to table properties and stored in SST files. SST files generated from SstFileWriter and Repairer have DB identity “SST Writer” and “DB Repairer”, respectively. Their DB session IDs are generated in the same way as `DB::GetDbSessionId`. The session ID for SstFileWriter (resp., Repairer) resets every time `SstFileWriter::Open` (resp., `Repairer::Run`) is called.
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,6 +13,7 @@
 ### Public API Change
 * `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the LOG file on the line starting with "DB Session ID:".
 * `DB::OpenForReadOnly()` now returns `Status::NotFound` when the specified DB directory does not exist. Previously the error returned depended on the underlying `Env`. This change is available in all 6.11 releases as well.
+* A parameter `verify_with_checksum` is added to `BackupEngine::VerifyBackup`, which is true by default. So now `BackupEngine::VerifyBackup` verifies backup files with checksum by default. When `verify_with_checksum` is false, `BackupEngine::VerifyBackup` only compares file sizes to verify backup files.
 
 ### New Features
 * DB identity (`db_id`) and DB session identity (`db_session_id`) are added to table properties and stored in SST files. SST files generated from SstFileWriter and Repairer have DB identity “SST Writer” and “DB Repairer”, respectively. Their DB session IDs are generated in the same way as `DB::GetDbSessionId`. The session ID for SstFileWriter (resp., Repairer) resets every time `SstFileWriter::Open` (resp., `Repairer::Run`) is called.
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,7 +13,7 @@
 ### Public API Change
 * `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the LOG file on the line starting with "DB Session ID:".
 * `DB::OpenForReadOnly()` now returns `Status::NotFound` when the specified DB directory does not exist. Previously the error returned depended on the underlying `Env`. This change is available in all 6.11 releases as well.
-* A parameter `verify_with_checksum` is added to `BackupEngine::VerifyBackup`, which is true by default. So now `BackupEngine::VerifyBackup` verifies backup files with checksum by default. When `verify_with_checksum` is false, `BackupEngine::VerifyBackup` only compares file sizes to verify backup files.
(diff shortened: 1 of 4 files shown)


include/rocksdb/utilities/backupable_db.h
```diff
@@ -396,9 +403,12 @@ class BackupEngine {
   }
 
   // checks that each file exists and that the size of the file matches our
-  // expectations. it does not check file checksum.
+  // expectations if with_checksum is false
+  // inspects whether the file checksum matches our expectation if
+  // with_checksum is true
   // Returns Status::OK() if all checks are good
-  virtual Status VerifyBackup(BackupID backup_id) = 0;
```
zhichao-cao, line 401: @pdillinger I'm not sure, is possible some users developed their own backup engine which is inherited from class BackupEngine?

zhichao-cao:
Thanks for the work! Left some comments after going through the first round.

include/rocksdb/utilities/backupable_db.h
```diff
@@ -274,16 +274,23 @@ class BackupEngineReadOnly {
...
+  //
   // If this BackupEngine created the backup, it compares the files' current
-  // sizes against the number of bytes written to them during creation.
+  // sizes (resp., current checksum) against the number of bytes written to
+  // them (resp., the checksum calculated) during creation.
   // Otherwise, it compares the files' current sizes against their sizes when
   // the BackupEngine was opened.
   //
   // Returns Status::OK() if all checks are good
-  virtual Status VerifyBackup(BackupID backup_id) = 0;
```
zhichao-cao, line 286: Maybe better to use the full name `verify_with_checksum`? It would be more readable in the code.

HISTORY.md
```diff
@@ -8,6 +8,7 @@
 
 ### Public API Change
 * `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the LOG file on the line starting with "DB Session ID:".
+* A parameter `with_checksum` is added to `BackupEngine::VerifyBackup`, which is true by default. So now `BackupEngine::VerifyBackup` verifies backup files with checksum by default. When `with_checksum` is false, `BackupEngine::VerifyBackup` only compares file sizes to verify backup files.
```
zhichao-cao, line 11: One question is, if user enable `with_checksum`, we do not need to verify the file size anymore?

utilities/backupable/backupable_db_test.cc
```diff
@@ -376,6 +376,27 @@ class FileManager : public EnvWrapper {
  public:
   explicit FileManager(Env* t) : EnvWrapper(t), rnd_(5) {}
 
+  Status GetRandomFileInDir(const std::string& dir, std::string* fname,
+                            uint64_t* fsize) {
+    std::vector<FileAttributes> children;
+    GetChildrenFileAttributes(dir, &children);
+    if (children.size() <= 2) {  // . and ..
+      return Status::NotFound("");
```
zhichao-cao, line 384: Better to put `dir` in the Status message.

HISTORY.md
gg814: Yes, the code logic does not verify file sizes if `with_checksum` is enabled, since when the sizes mismatch, checksum verification is very likely to fail.

pdillinger: If size is available, I think it's worth checking, especially if it's going to be checked on restore. It's an inexpensive check and 32-bit checksum collision is not inconceivable.

include/rocksdb/utilities/backupable_db.h
pdillinger: Ha. Good question. IMHO, that's rare enough that any affected users can modify their code. We have this listed under "public API change" so that seems like fair warning. :)

utilities/backupable/backupable_db.cc
```diff
@@ -1274,20 +1277,37 @@ Status BackupEngineImpl::VerifyBackup(BackupID backup_id) {
...
+    if (verify_with_checksum) {
+      // verify file checksum
+      uint32_t checksum_value = 0;
+      ROCKS_LOG_INFO(options_.info_log, "Verifying %s checksum...\n",
+                     abs_path.c_str());
+      CalculateChecksum(abs_path, backup_env_, EnvOptions(), 0 /* size_limit */,
+                        &checksum_value);
+      if (file_info->checksum_value != checksum_value) {
+        return Status::Corruption("File corrupted: Checksum mismatch for " +
+                                  abs_path);
```
pdillinger, line 1309: Let's add the expected and computed checksum values to the corruption message. That can help a lot for diagnostics.

utilities/backupable/backupable_db_test.cc
```diff
@@ -1050,6 +1070,71 @@ TEST_F(BackupableDBTest, CorruptionsTest) {
...
+    // get a random file in /private/1
+    ASSERT_OK(file_manager_->GetRandomFileInDir(backupdir_ + "/private/1",
+                                                &file_to_corrupt, &file_size));
+    // corrupt the file by replacing its content by file_size random bytes
+    ASSERT_OK(file_manager_->CorruptFile(file_to_corrupt, file_size));
+  }
+  // file sizes match
+  ASSERT_OK(backup_engine_->VerifyBackup(1, false));
+  // file checksums mismatch
+  ASSERT_NOK(backup_engine_->VerifyBackup(1, true));
```
pdillinger, line 1102: Let's add one case with default second argument

    ASSERT_NOK(backup_engine_->VerifyBackup(1));

Just to be sure that doesn't change accidentally.

HISTORY.md
```diff
@@ -11,6 +11,7 @@
 
 ### Public API Change
 * `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the LOG file on the line starting with "DB Session ID:".
+* A parameter `verify_with_checksum` is added to `BackupEngine::VerifyBackup`, which is true by default. So now `BackupEngine::VerifyBackup` verifies backup files with checksum and file size by default. When `verify_with_checksum` is false, `BackupEngine::VerifyBackup` only compares file sizes to verify backup files.
```
pdillinger, line 14: > verifies backup files with checksum and file size by default

Some English engineering:

_verifies checksums and file sizes of backup files by default_

For the last sentence, I think we should call out potential performance regression:

_Pass `false` for `verify_with_checksum` to maintain the previous behavior and performance of `BackupEngine::VerifyBackup`, by only verifying sizes of backup files._

include/rocksdb/utilities/backupable_db.h
```diff
@@ -274,16 +274,23 @@ class BackupEngineReadOnly {
     return RestoreDBFromLatestBackup(options, db_dir, wal_dir);
   }
 
+  // If verify_with_checksum is false, this function
```
zhichao-cao, line 277: Maybe it is better to put the description of "True" first.

gg814: Let me move the default case up :)

zhichao-cao:
Good job! I do not have further comments. How about @pdillinger ?

include/rocksdb/utilities/backupable_db.h
```diff
@@ -293,16 +293,23 @@ class BackupEngineReadOnly {
...
-  // the BackupEngine was opened.
+  // sizes (and current checksum) against the number of bytes written to
+  // them (and the checksum calculated) during creation.
+  // Otherwise, it compares the files' current sizes (and checksums) against
+  // their sizes (and checksums) when the BackupEngine was opened.
   //
   // Returns Status::OK() if all checks are good
-  virtual Status VerifyBackup(BackupID backup_id) = 0;
+  virtual Status VerifyBackup(BackupID backup_id,
+                              const bool& verify_with_checksum = true) = 0;
```
pdillinger, line 312: I just noticed `const bool &`. Should just use `bool`

utilities/backupable/backupable_db.cc
```diff
@@ -1316,20 +1319,40 @@ Status BackupEngineImpl::VerifyBackup(BackupID backup_id) {
...
   for (const auto& file_info : backup->GetFiles()) {
     const auto abs_path = GetAbsolutePath(file_info->filename);
+    // check existence of the file
     if (curr_abs_path_to_size.find(abs_path) == curr_abs_path_to_size.end()) {
       return Status::NotFound("File missing: " + abs_path);
     }
+    // verify file size
     if (file_info->size != curr_abs_path_to_size[abs_path]) {
-      return Status::Corruption("File corrupted: " + abs_path);
+      return Status::Corruption("File corrupted: File size mismatch for " +
```
pdillinger, line 1339: Let's also report "expected" and "found" file sizes in the message. :)

include/rocksdb/utilities/backupable_db.h
gg814: Just learnt this: [passing primitive type as constant reference](https://stackoverflow.com/questions/11932614/in-c-is-it-bad-to-pass-a-const-bool-by-reference) (linked here in case anyone is interested)

PR #7015, opened 2020-06-22
BackupEngine verifies table file checksums on creating new backups

gg814:
Summary:
When table file checksums are enabled and stored in the DB manifest by using the RocksDB default crc32c checksum function, BackupEngine will calculate the crc32c checksum of the file to be copied and compare the calculated result with the one stored in the DB manifest before copying the file to the backup directory.

After copying to the backup directory, BackupEngine will verify the checksum of the copied file with the one calculated before copying. This helps detect some rare corruption events such as bit-flips during the copying process.

No verification with checksums in DB manifest will be performed if the table file checksum function is not the RocksDB default crc32c checksum function.

In addition, If `share_table_files` and `share_files_with_checksum` are true, BackupEngine will compare the checksums computed before and after copying of the table files.

Corresponding tests are added.

Test plan:
Passed make check

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,9 @@
 ### Behavior Changes
 * Best-efforts recovery ignores CURRENT file completely. If CURRENT file is missing during recovery, best-efforts recovery still proceeds with MANIFEST file(s).
 * In best-efforts recovery, an error that is not Corruption or IOError::kNotFound or IOError::kPathNotFound will be overwritten silently. Fix this by checking all non-ok cases and return early.
+* When `file_checksum_gen_factory` is set to `GetFileChecksumGenCrc32cFactory()`, BackupEngine will compare the crc32c checksums of table files computed when creating a backup to the expected checksums stored in the DB manifest, and will fail `CreateNewBackup()` on mismatch (corruption). If the `file_checksum_gen_factory` is not set or set to any other customized factory, there is no checksum verification to detect if SST files in a DB are corrupt when read, copied, and independently checksummed by BackupEngine.
+
+### Bug fixes
 * Compressed block cache was automatically disabled with read-only DBs by mistake. Now it is fixed: compressed block cache will be in effective with read-only DB too.
 * Fix a bug of wrong iterator result if another thread finishes an update and a DB flush between two statement.
 * Disable file deletion after MANIFEST write/sync failure until db re-open or Resume() so that subsequent re-open will not see MANIFEST referencing deleted SSTs.
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -393,6 +393,9 @@ class DBImpl : public DB {
   virtual void GetLiveFilesMetaData(
       std::vector<LiveFileMetaData>* metadata) override;
 
+  virtual Status GetLiveFilesChecksumInfo(
+      FileChecksumList* checksum_list) override;
+
   // Obtains the meta data of the specified column family of the DB.
   // Status::NotFound() will be returned if the current DB does not have
   // any column family match the specified name.
--- a/include/rocksdb/utilities/backupable_db.h
+++ b/include/rocksdb/utilities/backupable_db.h
@@ -24,6 +24,11 @@
 
 namespace ROCKSDB_NAMESPACE {
 
+// The default DB file checksum function name.
+constexpr char kDbFileChecksumFuncName[] = "FileChecksumCrc32c";
+// The default BackupEngine file checksum function name.
+constexpr char kBackupFileChecksumFuncName[] = "crc32c";
+
 // BackupTableNameOption describes possible naming schemes for backup
 // table file names when the table files are stored in the shared_checksum
 // directory (i.e., both share_table_files and share_files_with_checksum
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -335,6 +335,14 @@ class BackupEngineImpl : public BackupEngine {
   inline bool IsSstFile(const std::string& fname) const {
     return fname.length() > 4 && fname.rfind(".sst") == fname.length() - 4;
   }
+  inline std::string ChecksumInt32ToStr(const uint32_t& checksum_int) {
+    std::string checksum_str;
+    PutFixed32(&checksum_str, EndianSwapValue(checksum_int));
+    return checksum_str;
+  }
+  inline uint32_t ChecksumStrToInt32(const std::string& checksum_str) {
+    return EndianSwapValue(DecodeFixed32(checksum_str.c_str()));
+  }
 
   // If size_limit == 0, there is no size limit, copy everything.
   //
@@ -382,6 +390,9 @@ class BackupEngineImpl : public BackupEngine {
     uint64_t size_limit;
     std::promise<CopyOrCreateResult> result;
     std::function<void()> progress_callback;
+    bool verify_checksum_after_work;
+    std::string src_checksum_func_name;
+    std::string src_checksum_str;
 
     CopyOrCreateWorkItem()
         : src_path(""),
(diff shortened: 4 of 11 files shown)


HISTORY.md
```diff
@@ -3,6 +3,9 @@
 ### Behavior Changes
 * Best-efforts recovery ignores CURRENT file completely. If CURRENT file is missing during recovery, best-efforts recovery still proceeds with MANIFEST file(s).
 * In best-efforts recovery, an error that is not Corruption or IOError::kNotFound or IOError::kPathNotFound will be overwritten silently. Fix this by checking all non-ok cases and return early.
+* When `file_checksum_gen_factory` is set to `GetFileChecksumGenCrc32cFactory()`, BackupEngine will calculate the crc32c checksum of the file to be backed up and compare the resulting value with the one stored in the DB manifest before copying the file to the backup directory. In addition, after copying to the backup directory, BackupEngine will double-check the checksum of the copied file with the one calculated before copying. This helps detect some rare corruption events such as bit-flips during the copying process.
```
zhichao-cao, line 6: We may need to point out the case that if the file_checksum_gen_factory is not set or set to other customized factory.

db/db_impl/db_impl.h
```diff
@@ -393,6 +393,9 @@ class DBImpl : public DB {
   virtual void GetLiveFilesMetaData(
       std::vector<LiveFileMetaData>* metadata) override;
 
+  virtual Status GetLiveFilesChecksumInfo(
```
pdillinger, line 396: This same information is available through GetLiveFilesMetaData, right? Are we adding this function for simplicity in getting the information? Or better performance?

I also think it's worth mentioning that this function (and others) are typically called with file deletions disabled, to avoid TOCTOU bugs / race conditions in cases where the information is expected to be current even after the call has completed.

gg814: Yes, `GetLiveFilesMetaData` gives us `MetaData` but it has more than what we need. I think using `GetLiveFilesChecksumInfo` is more efficient as it provides exactly what we want. (Thanks for pointing out these functions are typically called with file deletions disabled :)

utilities/backupable/backupable_db.cc
```diff
@@ -1494,15 +1562,47 @@ Status BackupEngineImpl::AddBackupFileWorkItem(
...
+  // before copying. Note that only table files may have a known checksum name
+  // passed in.
+  //
+  // If no default checksum function name is passed in, we will calculate the
+  // checksum *before* copying in two cases (we always calcuate checksums when
+  // copying or creating for any type files):
+  // a) share_files_with_checksum is true and file type is table;
+  // b) share_table_files is true and the file exists already.
+  if (kBackupFileChecksumFuncName == src_checksum_func_name) {
     s = CalculateChecksum(src_dir + fname, db_env_, src_env_options, size_limit,
```
pdillinger, line 1578: Oh interesting. I just realized that we are already reading file to compute checksum before copying. (Sigh) And am I reading right that if we have to copy the file after computing that checksum, we ignore the first checksum and re-compute the checksum on copying? I think this is what you are referring to in the first HISTORY entry, but it's not clear to me.

HISTORY.md
```diff
@@ -6,6 +6,10 @@
 ### Behavior Changes
 * Best-efforts recovery ignores CURRENT file completely. If CURRENT file is missing during recovery, best-efforts recovery still proceeds with MANIFEST file(s).
 * In best-efforts recovery, an error that is not Corruption or IOError::kNotFound or IOError::kPathNotFound will be overwritten silently. Fix this by checking all non-ok cases and return early.
+* If `share_table_files` and `share_files_with_checksum` are true, BackupEngine will compare the checksums computed before and after copying of the table files.
```
pdillinger, line 9: It's not re-reading the file after copying to get the "after copying" checksum, right? It's really comparing checksum before copying to checksum during copying, right?

utilities/backupable/backupable_db.cc
gg814: Right, I figured since we have the checksum before copying available, it may make sense to compare the checksum after copying against it without incurring additional costs.

HISTORY.md
gg814: That's right. No re-reading here. It is comparing the checksum before copying to the checksum obtained right after copying is finished.

gg814:
Rebase and some small changes are made.

HISTORY.md
```diff
@@ -6,6 +6,10 @@
...
+* When `file_checksum_gen_factory` is set to `GetFileChecksumGenCrc32cFactory()`, BackupEngine will calculate the crc32c checksum of the table file to be backed up and compare the resulting value with the one stored in the DB manifest before copying the file to the backup directory. In addition, after copying the table file to the backup directory, BackupEngine will double-check the checksum of the copied file with the one calculated before copying. If the `file_checksum_gen_factory` is not set or set to any other customized factory, checksum verification with DB manifest before copying will not be performed.
```
pdillinger, line 10: Related to above, I suggest this:

* When `file_checksum_gen_factory` is set to `GetFileChecksumGenCrc32cFactory()`, BackupEngine will compare the crc32c checksums of table files computed when creating a backup to the expected checksums stored in the DB manifest, and will fail CreateNewBackup on mismatch (corruption). If the `file_checksum_gen_factory` is not set or set to any other customized factory, there is no checksum verification to detect if SST files in a DB are corrupt when read, copied, and independently checksummed by BackupEngine.

zhichao-cao:
It looks good to me and I would defer to @pdillinger to approve

include/rocksdb/utilities/backupable_db.h
```diff
@@ -24,6 +24,11 @@
 
 namespace ROCKSDB_NAMESPACE {
 
+// The default DB file checksum function name.
+constexpr char kDbFileChecksumFuncName[] = "FileChecksumCrc32c";
```
zhichao-cao, line 28: Should we add "Default" to its name?

include/rocksdb/utilities/backupable_db.h
```diff
@@ -24,6 +24,11 @@
...
+// The default BackupEngine file checksum function name.
+constexpr char kBackupFileChecksumFuncName[] = "crc32c";
```
zhichao-cao, line 30: Also here

pdillinger: These will almost certainly be refactored soon, so NBD to me. :)

gg814:
Thanks! Updated HISTORY.md (and made some small changes).

utilities/backupable/backupable_db.cc
ajkr: Should we make a task to remove the first checksum calculation some day? I was also under the impression that we only checksum during the streaming copy. Also it looks like this PR introduces another case where checksum would be computed prior to copy: when `kDbFileChecksumFuncName == src_checksum_func_name`. Is there a need for it or is it just there for consistency (and thus can be removed together with the other cases of computing checksum prior to copy)?

gg814: I think some users (in an internal discussion) prefer validating the checksum in db manifest prior to copy. (Maybe we could make the validation optional.)

pdillinger: > I think some users (in an internal discussion) prefer validating the checksum in db manifest prior to copy. (Maybe we could make the validation optional.)

I believe it's a flawed approach (1) to (a) read file to validate checksum against db manifest and then (b) re-read file to copy (and maybe then re-validate checksum). I argue this is much worse than (2) read for copy and validate checksum at the end of copy. Not only do you have extra I/O reads, you inject an opportunity for a transient bit flip (CPU/mem/IO) to cause havoc. The only extra cost is writing a useless copy in the case of corruption detected. If your probability of corruption is high enough to worry about that, you have serious problems.

Increasing the surface area for transient corruption without improving power to detect permanent corruption does not improve data integrity.

PR #7017, opened 2020-06-23
Fix the memory leak in Env_basic_test

zhichao-cao:
Fix the memory leak broken asan and other test introduced by #6830 

test plan: pass asan_check


matthewvon:
That was my code that leaked.  My apologies.

zhichao-cao:
> That was my code that leaked. My apologies. Thank you for the fix.

A minor fix, thanks for your contribution!

PR #7018, opened 2020-06-23
Move kNoExpiration to blob_db.h

ltamasi:
Summary:
The constant `kNoExpiration` is currently defined in an
internal/implementation header (`blob_log_format.h`); the patch moves it
to the public header `blob_db.h` so it is accessible to users.

Test Plan:
`make check`

### DIFF
--- a/db/blob/blob_log_format.h
+++ b/db/blob/blob_log_format.h
@@ -9,7 +9,6 @@
 
 #ifndef ROCKSDB_LITE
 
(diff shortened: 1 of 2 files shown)


ajkr:
@ltamasi - I've backported this to 6.11.

ltamasi:
> @ltamasi - I've backported this to 6.11.

Thanks @ajkr !

PR #7020, opened 2020-06-23
Encryption with Change Management

matthewvon:
Encryption at rest is often an enterprise requirement.  This PR provides an implementation that uses OpenSSLs AES 256 CTL encryption.  

The code is written in such a way that the programmer can easily switch to a different library and/or encryption cipher.  And the switch can happen later without requiring the end user to dump their entire database then reload under the new library / cipher.  The encryption library is intentionally loaded at runtime, not compiled into rocksdb.  This allows for situations such as HeartBleed where users want a quick update of security code without wait for a new product release.  They simply load the new library and restart the database.

The code also allows for encryption keys to change.  There must be a "current key" for writing and reading.  There can also be previous keys for reading older .sst files.  This implies that the end user can migrate from older keys either by simply waiting for compactions to naturally replace older keys, or by forcing a compaction of all files.  The latter might take time, but the database is still operational.


matthewvon:
I am staring at this error and it really looks like a compiler bug:

https://travis-ci.org/github/facebook/rocksdb/jobs/701431752

Open to making any fixes that I am missing.

matthewvon:
@ajkr @mrambacher @yiwu-arbug  As promised, here is the second encryption PR that exploits inheritance from #6830.  Mark has already pointed out that part of the library_loader is already present Env.  I will rewrite to use it Wednesday.

include/rocksdb/env_encrypt2.h
```diff
@@ -0,0 +1,471 @@
...
+  }
+
+  virtual BlockAccessCipherStream* CreateCipherStream2(uint8_t code_version,
+                                                       uint8_t nonce[]) {
+    return new AESBlockAccessCipherStream(key_, code_version, nonce);
+  }
+
+  bool Valid() const { return valid_; };
+  const Sha1Description_t& key_desc() const { return key_desc_; };
+  const AesCtrKey_t& key() const { return key_; };
```
pdillinger, line 193: We hit this all the time with gcc 4.8, which is why we added the build. Ctor param shadows this member function.

matthewvon: @pdillinger Thank you for taking time to explain.  I can easily adapt the constructor.

include/rocksdb/env_encrypt2.h
```diff
@@ -0,0 +1,471 @@
...
+
+    // Look for encryption marker
+    EncryptMarker_t marker;
+    Slice marker_slice;
+    status = f->Read(0, sizeof(marker), &marker_slice, marker);
+    if (status.ok()) {
+      if (sizeof(marker) == marker_slice.size() &&
+          marker_slice.starts_with(Marker)) {
+        // code_version currently unused
+        uint8_t code_version = (uint8_t)marker_slice[7];
```
yiwu-arbug, line 346: can we verify `code_version` is indeed 0, or return error?

env/env_encrypt2.cc
```diff
@@ -0,0 +1,558 @@
...
+    crypto_.RAND_poll();
+  }
+}
+
+EncryptedEnv2::EncryptedEnv2(Env* base_env)
+    : EnvWrapper(base_env), valid_(false) {}
+
+void EncryptedEnv2::SetKeys(EncryptedEnv2::ReadKeys_t encrypt_read,
+                            EncryptedEnv2::WriteKey_t encrypt_write) {
+  encrypt_read_ = encrypt_read;
```
yiwu-arbug, line 224: what if there's concurrent access to the keys when we call `SetKeys`? Do we need a mutex?

matthewvon: This routine is currently called only by Encrypt2::Default().  There is not even a unit test that calls it.  How about making it "protected:" instead.  If we put a mutex, that implies every access to the keys would also need to lock the same mutex.  That is more overhead than I would really like for an extremely unlikely use case.

include/rocksdb/env_encrypt2.h
matthewvon: this is complete and passing tests now.

util/library_loader.h
```diff
@@ -0,0 +1,162 @@
...
+#include <string>
+
+#include "rocksdb/rocksdb_namespace.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// Base class / interface
+//  expectation is to derive one class for unux and one for Windows
+//
+class LibraryLoader {
```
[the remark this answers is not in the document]
matthewvon: I see two ways to achieve static link.  First is a conditional compile.  Second is to exploit PosixEnv::LoadLibrary()'s ability to pull symbols from either external library or running executable.  I prefer the latter.  It creates one code path for both scenarios.  It would also allow the user to ship with a "current" version and then override at runtime if an update version comes out with a critical fix.  The feature would require an option setting.  If you like this idea, would you suggest which options structure is most appropriate.

matthewvon:
@yiwu-arbug Per your item one about 16 versus 32 byte IV.  The NIST standard and samples use 16 byte IV AES CTR 256.  The code is currently verified against that standard.  I have two sources that say 16 byte/128bit is proper IV for AES 256:

- https://android.googlesource.com/platform/external/boringssl/+/f4e4272/src/crypto/cipher/test/cipher_test.txt
- https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf

The encrypt/decrypt samples from the second document are part of env_encrypt2_test.cc.

Do you have a reference that I can review?  There is nothing to prevent a coder from implementing a different cipher with different IV size.  In fact, the purpose of env_encrypt2's design is to enable and encourage those changes and differences.

include/rocksdb/env_encrypt2.h
```diff
@@ -0,0 +1,471 @@
...
+  bool IsValid() const { return valid; }
+};
+
+struct AesCtrKey_t {
+  uint8_t key[EVP_MAX_KEY_LENGTH];
+  bool valid;
+
+  AesCtrKey_t() : valid(false) { memset(key, 0, EVP_MAX_KEY_LENGTH); }
+
+  AesCtrKey_t(const uint8_t* Key, size_t KeyLen) : valid(false) {
```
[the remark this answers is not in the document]
matthewvon: done

include/rocksdb/env_encrypt2.h
```diff
@@ -0,0 +1,471 @@
...
+
+  EncryptedEnv2(Env* base_env, ReadKeys_t encrypt_read,
+                WriteKey_t encrypt_write);
+
+  void SetKeys(ReadKeys_t encrypt_read, WriteKey_t encrypt_write);
+
+  bool IsWriteEncrypted() const { return nullptr != encrypt_write_.second; }
+
+  // NewSequentialFile opens a file for sequential reading.
+  virtual Status NewSequentialFile(const std::string& fname,
```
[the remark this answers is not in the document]
matthewvon: done.  not just at this point, but everywhere I had virtual on an override line.

include/rocksdb/env_encrypt2.h
```diff
@@ -0,0 +1,471 @@
...
+#include "util/library_loader.h"
+#include "util/random.h"
+
+#endif
+
+namespace ROCKSDB_NAMESPACE {
+
+#ifndef ROCKSDB_LITE
+
+struct Sha1Description_t {
```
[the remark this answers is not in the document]
matthewvon: done

matthewvon:
@yiwu-arbug Can you point me to source code or such for item three's update in place?

util/library_loader.h
```diff
@@ -0,0 +1,162 @@
...
+                                       const unsigned char *key,
+                                       const unsigned char *iv);
+  using EVP_aes_256_ctr_t = const EVP_CIPHER *(*)(void);
+  using EVP_EncryptUpdate_t = int (*)(EVP_CIPHER_CTX *ctx, unsigned char *out,
+                                      int *outl, const unsigned char *in,
+                                      int inl);
+
+  EVP_CIPHER_CTX *EVP_CIPHER_CTX_new(void) const { return cipher_new_(); };
+
+  void EVP_CIPHER_CTX_free(EVP_CIPHER_CTX *ctx) { cipher_free_(ctx); }
```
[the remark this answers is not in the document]
matthewvon: Agreed.  And it ticked me off.  The code already compensates.  Please see:

https://github.com/facebook/rocksdb/blob/09d5080dc80648bcc4e2520cc5fb72d842961e6e/util/library_loader.cc#L96

The code simply maps the ENV_CIPHER_CTX_destroy into the cipher_free_ pointer since parameters are same.

util/library_loader.h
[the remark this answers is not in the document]
matthewvon: might still need a conditional compile to make sure the symbols exist in the executable.  I am guessing that a link against .a instead of .so would not pull in the symbols/functions needed unless there is "code calling them".  any thoughts here are welcome

matthewvon:
@mrambacher I have converted code to PosixEnv::LoadLibrary().  That revealed a bug in env_posix.cc where constants used for library suffixes within std::string might not have initialized by the time LoadLibrary is called.  Switched those three constants to static const char * so that compiler initializes, not loader.

env/env_encrypt2.cc
matthewvon: I implemented the move to "protected" and added comment concerning not thread safe.

include/rocksdb/env_encrypt2.h
matthewvon: done

env/env_encrypt2.cc
yiwu-arbug: The use case is online key rotation: caller want to replace the write key without interrupting service.

matthewvon: ok, but that implies changing on file open boundaries ... so more code would be require to save current write key within an open writable file.  I would prefer to leave that as an extension, a new PR, unless you say that is necessary for acceptance of this PR.  Doing that would reduce the number of times the mutex is hit compared to my previous concern:  once per file open versus every read/write operation.  And if some user code accidentally removes a key that is needed, system will have read errors ... so more code needed to make sure there is a union of current read keys with new keys (including forcing the current write key into read list when write is changed).

matthewvon:
Looking for help again, this time with:

https://travis-ci.org/github/facebook/rocksdb/jobs/702118891

This test appears to see if a Status object was created but its value never tested.  I see the creation and set on line 18 of util/library_loader.cc and the stat.ok() check made on line 19.  My understanding is that order of operation for && should force stat.ok() to test before nullptr != ... the latter test is likely null.  Any thoughts?

matthewvon:
@yiwu-arbug About item 2: The biggest performance hit is likely not the 3 calls you mention, but the EVP_CIPHER_CTX_new() and _free() calls.  There exists an EVP_CIPHER_CTX_reset() that would allow one Read or Write operation to reuse its context for all of its buffer.  From that point on, the 3 calls you mention appear to be required because the CTR encryption requires the key to be rebuilt for each block using a unique IV.  (I think I did not code the reuse because I had a type checking issue unless I made a subtle change to the original BlockAccessCipherStream to carry void * parameters/handles.  Default rocksdb build disables rtti.)  I have yet to find code or examples suggesting that the IV is somehow consistently incremented automatically within openssl's CTR code.  I am open to correction on this issue.

util/library_loader.h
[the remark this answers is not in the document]
yiwu-arbug: Not familiar with dylib loading, but it remind me Jemalloc allow to add prefixes to its methods so to not conflict with, say, libc. But if OpenSSL don't support that, maybe we just conditional compile to choose whether static link to dynamic link?

util/library_loader.h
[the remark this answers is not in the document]
yiwu-arbug: OpenSSL 1.0.2 doesn't seems to provide `EVP_CIPHER_CTX_destroy` as well. With 1.0.2 the ctx can only be created on stack. https://www.openssl.org/docs/man1.0.2/man3/EVP_EncryptInit_ex.html

include/rocksdb/env_encrypt2.h
yiwu-arbug: Same for `ReadSeqEncryptionPrefix`.

util/library_loader.h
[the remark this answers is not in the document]
matthewvon: openssl does not directly support prefixes with dylib loading.  I can make both static+dynamic or dynamic only work via compile time flag.  If static is not correctly identified as missing at compile time, the compile fails.  A couple of recent test passes are suggesting the dynamic load might be having dependency problems in your test infrastructure (I have added some debug code to figure this out.  Might have to make test for openssl more exhaustive in build_detect_platform).  I am continuing to work this issue.

util/library_loader.h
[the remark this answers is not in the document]
matthewvon: Ok.  That would actually simplify the current code.  Using thread_local to get around recoding a bunch of stuff.  Keeps one context per thread ... which equates to one context per read or write call.

include/rocksdb/env_encrypt2.h
matthewvon: oops ... will fix.

env/env_encrypt2.cc
yiwu-arbug: sounds good.

matthewvon: Ok to leave as extension in future PR, or Ok as to design and must implement now?

env/env_encrypt2.cc
```diff
@@ -0,0 +1,572 @@
...
+  //  loading of UnixLibCrypto before other routines start
+  //  using the encryption code.
+  static EncryptedEnv2 default_env(Env::Default());
+  return &default_env;
+}
+
+Env* EncryptedEnv2::Default(EncryptedEnv2::ReadKeys encrypt_read,
+                            EncryptedEnv2::WriteKey encrypt_write) {
+  EncryptedEnv2* default_env = (EncryptedEnv2*)Default();
+  default_env->SetKeys(encrypt_read, encrypt_write);
```
yiwu-arbug, line 564: It looks like a bad pattern to modify internal data for a singleton.. But I understand `Options` only hold raw pointer of `Env`. Shall we add a `std::shared_ptr<Env> env_owned` field to `Options` so that the DB can manage the lifetime of a non-default `Env`?

@pdillinger what do you think from rocksdb team perspective?

yiwu-arbug:
> @yiwu-arbug Per your item one about 16 versus 32 byte IV. The NIST standard and samples use 16 byte IV AES CTR 256. The code is currently verified against that standard. I have two sources that say 16 byte/128bit is proper IV for AES 256:
> 
> * https://android.googlesource.com/platform/external/boringssl/+/f4e4272/src/crypto/cipher/test/cipher_test.txt
> * https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf
> 
> The encrypt/decrypt samples from the second document are part of env_encrypt2_test.cc.
> 
> Do you have a reference that I can review? There is nothing to prevent a coder from implementing a different cipher with different IV size. In fact, the purpose of env_encrypt2's design is to enable and encourage those changes and differences.

IV for AES-CTR is indeed 16 bytes. What I mean is `EncryptedEnv` use 8 bytes nonce and 8 bytes counter to construct IV. When the 8 bytes counter overflows, it doesn't overflow into the upper 8 bytes nonce. And OpenSSL `EVP_aes_256_ctr` treat IV as a 16 bytes integer. When the lower 8 bytes overflow, the upper 8 bytes increase by 1. The difference make `EncryptedEnv` not compatible with OpenSSL.

matthewvon:
> > @yiwu-arbug Per your item one about 16 versus 32 byte IV. The NIST standard and samples use 16 byte IV AES CTR 256. The code is currently verified against that standard. I have two sources that say 16 byte/128bit is proper IV for AES 256:
> > 
> > * https://android.googlesource.com/platform/external/boringssl/+/f4e4272/src/crypto/cipher/test/cipher_test.txt
> > * https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf
> > 
> > The encrypt/decrypt samples from the second document are part of env_encrypt2_test.cc.
> > Do you have a reference that I can review? There is nothing to prevent a coder from implementing a different cipher with different IV size. In fact, the purpose of env_encrypt2's design is to enable and encourage those changes and differences.
> 
> IV for AES-CTR is indeed 16 bytes. What I mean is `EncryptedEnv` use 8 bytes nonce and 8 bytes counter to construct IV. When the 8 bytes counter overflows, it doesn't overflow into the upper 8 bytes nonce. And OpenSSL `EVP_aes_256_ctr` treat IV as a 16 bytes integer. When the lower 8 bytes overflow, the upper 8 bytes increase by 1. The difference make `EncryptedEnv` not compatible with OpenSSL.

I understand now.  Easy to fix.

matthewvon:
All:  from recent appveyor/pr failure:

Microsoft Visual Studio 2019 Version 16.5.5.
858Copyright (C) Microsoft Corp. All rights reserved.
859
860The license for Visual Studio has expired.
861
862The evaluation period for this product has ended.
863Command exited with code 1
8647z a build-failed.zip %APPVEYOR_BUILD_FOLDER%\build\ && appveyor PushArtifact build-failed.zip
865
866

see:  https://ci.appveyor.com/project/Facebook/rocksdb/builds/33765433/job/6chf3tpw9ey2le2l

include/rocksdb/env_encrypt2.h
```diff
@@ -0,0 +1,473 @@
...
+#include "util/library_loader.h"
+#include "util/random.h"
+
+#endif
+
+namespace ROCKSDB_NAMESPACE {
+
+#ifndef ROCKSDB_LITE
+
+struct Sha1Description {
```
mrambacher, line 31: I think this adds way more to the public API than is necessary.  I think most of this code can be moved into the CC file and not in the public API.

If you did the move, could you then methods to add/set new keys to the base EncryptedEnv that return NotSupported?  

As far as the public portion of the API, it seems like the only change is adding the changing of the keys.  Is there a reason to expose this as a Sha1Description versus something that is managed internally?  Or have a more abstract structure that you can use (like potentially a key name, nonce, IV, ...)?

More comments to come later...

matthewvon: @mrambacher Not saying I am correct ... I was trying to make more things available to future developers after having to fight, then restructure, the original env_encryption.cc/.h.  Maybe the compromise is to move these structures into an implementation .h within the env directory so they can be reused / extended easily ... while still not cluttering the public API files?  It feels wrong to me to bury them in the .cc.  That makes them useless to the next coder.

yiwu-arbug:
> @yiwu-arbug About item 2: The biggest performance hit is likely not the 3 calls you mention, but the EVP_CIPHER_CTX_new() and _free() calls. There exists an EVP_CIPHER_CTX_reset() that would allow one Read or Write operation to reuse its context for all of its buffer. From that point on, the 3 calls you mention appear to be required because the CTR encryption requires the key to be rebuilt for each block using a unique IV. (I think I did not code the reuse because I had a type checking issue unless I made a subtle change to the original BlockAccessCipherStream to carry void * parameters/handles. Default rocksdb build disables rtti.) I have yet to find code or examples suggesting that the IV is somehow consistently incremented automatically within openssl's CTR code. I am open to correction on this issue.

Not OpenSSL expert and don't know how the IV get updated in the `ctx` struct as well. Since I don't have test result I don't have strong opinion for this item.

matthewvon:
> > @yiwu-arbug About item 2: The biggest performance hit is likely not the 3 calls you mention, but the EVP_CIPHER_CTX_new() and _free() calls. There exists an EVP_CIPHER_CTX_reset() that would allow one Read or Write operation to reuse its context for all of its buffer. From that point on, the 3 calls you mention appear to be required because the CTR encryption requires the key to be rebuilt for each block using a unique IV. (I think I did not code the reuse because I had a type checking issue unless I made a subtle change to the original BlockAccessCipherStream to carry void * parameters/handles. Default rocksdb build disables rtti.) I have yet to find code or examples suggesting that the IV is somehow consistently incremented automatically within openssl's CTR code. I am open to correction on this issue.
> 
> Not OpenSSL expert and don't know how the IV get updated in the `ctx` struct as well. Since I don't have test result I don't have strong opinion for this item.

Actually you have already inspired a potentially better solution by making me aware of differences in 1.0.2 openssl.  It never occurred to me to create the EVP_CIPHER_CTX on the stack since all examples I found were calling EVP_CIPHER_CTX_new().  Let me recode it on the stack and get back to you.  I greatly prefer stack over heap.

env/env_encrypt2.cc
```diff
@@ -0,0 +1,572 @@
...
+                                               char* prefix,
+                                               size_t prefixLength) {
+  Status s;
+  if (EncryptedEnv2::crypto_.IsValid()) {
+    if (sizeof(Prefix0) <= prefixLength) {
+      int ret_val;
+
+      Prefix0* pf = {(Prefix0*)prefix};
+      memcpy(pf->key_description_, key_desc_.desc, sizeof(key_desc_.desc));
+      ret_val = EncryptedEnv2::crypto_.RAND_bytes(
```
yiwu-arbug, line 177: If we are going to fix item 1, shall we generate a full 16 bytes initial IV?

env/env_encrypt2.cc
```diff
@@ -0,0 +1,572 @@
...
+  } else {
+    s = Status::NotSupported("RAND_bytes() from libcrypto not available.");
+  }
+
+  return s;
+}
+
+// Returns an Env that encrypts data when stored on disk and decrypts data when
+// read from disk.
+Env* NewEncryptedEnv2(Env* base_env, EncryptedEnv2::ReadKeys encrypt_read,
```
yiwu-arbug, line 195: Shall we call it `EncryptedEnvV2` instead of `EncryptedEnv2`?

include/rocksdb/env_encrypt2.h
yiwu-arbug, line 31: Extremely unlikely, but do we need to handle if two keys have same Sha1 hash?

matthewvon: In my application code, the second key gets rejected and user has to rename.  Yes, there could be additional code to try a known sample to decrypt and decide which key ... but then that gives the bad guys a known sample too.  I recommend we let this slide.

yiwu-arbug: agree. seems to me it is the application responsibility to check it when providing the keys. just asking to confirm.

yiwu-arbug:
> @yiwu-arbug Can you point me to source code or such for item three's update in place?
> 
> update: I believe this confirms your "update in place" for software AES implementation.
> 
> https://github.com/openssl/openssl/blob/master/crypto/aes/aes_core.c
> 
> Searching now for confirmation when hardware AES used (I know it is likely true, but feel obligated to verify).
> 
> update 2: it looks to me that the memmove within EncryptedWritableFile::Append() is related to getting data for encryption aligned with offsets within files. Once the addresses and offsets are aligned on proper boundaries, encryption/decryption is an "update in place". The shifting is about alignment.
> 
> update 3: ok, I understand now how to eliminate the need for alignment of buffers and offsets for CTR mode. I do not know what other derived code might be depending upon the alignment. Therefore, I am going to copy those routines from EncryptedEnv into EnctypedEnv2 for modification. This should also simplify reuse of the EVP_CIPHER_CTX object.
> 
> update 4: hold the phone. the memmove is because we are handed a const buffer from the caller. we do not know the impact of encrypt in place to the caller. so the memmove is to create a copy that we then encrypt in place. the current EncryptionEnv is clean. The alignment stuff is purely a side benefit ... and likely required for use_direct_io.

👌

env/env_encrypt2.cc
matthewvon: I have zero hesitations in matching openssl equivalent.  Any chance you can point to openssl source code so that I can copy and therefore reduce the number of cycles here?

yiwu-arbug: This is not to match openssl, as IVs are provided by caller. But since IV (used as both of nonce and counter) is just a big 128-bit integer, I feel it more natural to rand-gen full 16 bytes IV.

yiwu-arbug: The 128-bit IV thing you can refer to this: https://github.com/openssl/openssl/blob/master/crypto/modes/ctr128.c#L26

util/library_loader.h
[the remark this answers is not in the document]
matthewvon: I have pulled the source code to 1.0.1u.  It has functions _new, _free, and _cleanup.  I have pulled the source code to 1.0.2.u.  It has the functions _new, _free, and _cleanup.  Going to strongly suggest a documentation problem on their website.  My code is using _new, _free, and _reset.  ENV_CIPHER_CTX_reset() is a name change for _cleanup.  Source code uses the two equivalently inside _free and _init.  I believe the real action item here is to create code that will pull _cleanup as an alternative entry point to _reset.  

I was happily looking at maybe using a stack based copy of ENV_CIPHER_CTX.  There is no destructor and there are internal allocations that need clean up.  _new & _free might be safer in the long run, though calling _cleanup/_reset when done with stack struct accomplishes the same.  Looking at other alternatives too based upon review of ctrl128.c reference you posted elsewhere in this PR.

matthewvon:
CI team:  Windows having GitHub connection problems:

https://app.circleci.com/pipelines/github/facebook/rocksdb/2012/workflows/c1f455b9-0315-40c2-af69-363f752e5598/jobs/4273/steps

env/env_encrypt2.cc
matthewvon: Done.

env/env_encrypt2.cc
matthewvon: This is done.  However, the new support routine  BigEndianAdd128() has a carry problem beyond the lower 8 bytes.  I am addressing this bug plus adding unit tests for this routine.

include/rocksdb/env_encrypt2.h
matthewvon: @mrambacher I have moved template implementations to .cc file.  Added env_encrypt2_impl.h for one class that is not needed for public API, but might be reusable in other implementations.

env/env_encrypt2.cc
matthewvon: All access to encryption keys within the object are surrounded by a read/write lock.  It was going to be difficult to put the lock within the class due to RWMutex being hidden within port (i.e. not public).  I cheated and made it a local global.  It is not going to be hit often even if multiple databases are active.

matthewvon:
@yiwu-arbug New code is available for review.  EncryptedWritableFileV2::Append() demonstrations encryption in place with one EVP_Encrypt_Update() call for entire buffer size.

I am updating the other write operations now and will add unit tests for them.  I believe today's check-ins address all issues raised.  These final write routines will finish the work (if not today, this weekend).

matthewvon:
@yiwu-arbug "All checks have passed" from CI.  I feel code complete.  It occurs to me that I have not provided a static build with OpenSSL (libcrypto).  Is that an absolute requirement?

yiwu-arbug:
> @yiwu-arbug "All checks have passed" from CI. I feel code complete. It occurs to me that I have not provided a static build with OpenSSL (libcrypto). Is that an absolute requirement?

I think the static build option can be added later, although I still prefer to have it.

util/library_loader.cc
```diff
@@ -54,9 +54,9 @@ size_t LibraryLoader::GetEntryPoints(std::map<std::string, void*>& functions) {
 
 UnixLibCrypto::UnixLibCrypto() : LibraryLoader(crypto_lib_name_) {
   if (is_valid_) {
-    // size of map minus two since _new/_create and _free/_destroy
-    //  only resolve one of the two.
-    is_valid_ = ((functions_.size() - 2) == GetEntryPoints(functions_));
+    // size of map minus three since _new/_create, _free/_destroy
```
yiwu-arbug, line 55: which version of OpenSSL does `_create` and `_destroy` comes from? sorry I didn't find the reference.

yiwu-arbug:
I don't have further questions, but I didn't read the latest changes carefully either. Can anyone from the rocksdb team have further review? Thanks. @ajkr @pdillinger

matthewvon:
1. I see the asan test bug.  Will work it Thursday.
2. There are two bugs with EncryptedRandomRWFileV2 that are mostly due to a lack of unit test.  I have the fixes and will create a unit test
3. There is a subtle performance optimization if I use openssl slightly differently.  Will work it Thursday.

matthewvon:
Assuming tests succeed, PR is ready.  Recent openssl usage changes provided a minimum of 12% performance increase.  EncryptedRandomRWFileV2 corrected and ready to rock.

table/block_based/block_based_table_reader.cc
```diff
@@ -542,9 +542,9 @@ Status GetGlobalSequenceNumber(const TableProperties& table_properties,
       std::array<char, 200> msg_buf;
       snprintf(
           msg_buf.data(), msg_buf.max_size(),
-          "An external sst file with version %u have global seqno property "
-          "with value %s, while largest seqno in the file is %llu",
-          version, seqno_pos->second.c_str(),
+          "An external sst file with version %u has global seqno property "
+          "with value %llu, while largest seqno in the file is %llu",
+          version, static_cast<unsigned long long>(global_seqno),
```
matthewvon: This was:

Corruption: An external sst file with version 2 have global seqno property with value ³CΕ	m, while largest seqno in the file is 15351

With above change:

Corruption: An external sst file with version 2 have global seqno property with value 8117678604447195114, while largest seqno in the file is 15351709

(Underlying issue corrected in env_encrypt2.cc.)

matthewvon:
@mrambacher @yiwu-arbug @pdillinger Anything I can do to generate interest in this PR?

matthewvon:
FYI:  new CI ran tonight because I addressed a src.mk conflict.  Best I can tell, all 5 failures are not related to this PR.

matthewvon:
@mrambacher @yiwu-arbug @pdillinger Code once again passes all checks.

util/library_loader_test.cc
```diff
@@ -0,0 +1,87 @@
...
+
+  ASSERT_TRUE(works.IsValid());
+  ASSERT_FALSE(fails.IsValid());
+
+  double (*floor)(double);
+
+  floor = (double (*)(double))works.GetEntryPoint("floor");
+  ASSERT_TRUE(nullptr != floor);
+  ASSERT_TRUE(2.0 == (*floor)(2.2));
+}
```
mrambacher, line 39: I like this test.  Can we move it into env_test and use it more regularly?

util/library_loader_test.cc
```diff
@@ -0,0 +1,87 @@
...
+#include "util/library_loader.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+#ifdef OS_MACOSX
+static const char* LIB_M_NAME = "libm";
+static const char* LIB_BAD_NAME = "libbubbagump.dylib";
+static const char* LIB_SSL_NAME = "ssl";
+#else
+static const char* LIB_M_NAME = "libm.so.6";
```
mrambacher, line 20: Why not just libm or m?

env/env_encrypt2.cc
```diff
@@ -0,0 +1,989 @@
...
+    if (!aes_context) {
+      aes_context = std::unique_ptr<EVP_CIPHER_CTX, void (*)(EVP_CIPHER_CTX*)>(
+          crypto_shared->EVP_CIPHER_CTX_new(),
+          crypto_shared->EVP_CIPHER_CTX_free_ptr());
+    }
+
+    memcpy(iv.bytes, nonce_, AES_BLOCK_SIZE);
+    BigEndianAdd128(iv.bytes, block_index);
+
+    ret_val = crypto_shared->EVP_EncryptInit_ex(
```
mrambacher, line 209: This might be my lack of understanding of libcrypto, but shouldn't Decrypt need to call EVP_Decrypt?

env/env_encrypt2.cc
```diff
@@ -0,0 +1,989 @@
...
+        aes_context.get(), crypto_shared->EVP_aes_256_ctr(), nullptr, key_.key,
```
mrambacher, line 210: Should the type of cipher be encoded in the header so that it could be 256 or something else?

env/env_encrypt2_impl.h
```diff
@@ -0,0 +1,81 @@
...
+typedef char EncryptMarker[8];
+static EncryptMarker kEncryptMarker = "Encrypt";
+
+// long term:  code_version could be used in a switch statement or factory
+// prefix version 0 is 12 byte sha1 description hash, 128 bit (16 byte)
+// nounce (assumed to be packed/byte aligned)
+typedef struct {
+  uint8_t key_description_[EVP_MAX_MD_SIZE];  // max md is 64
+  uint8_t nonce_[AES_BLOCK_SIZE];             // block size is 16
+} PrefixVersion0;
```
mrambacher, line 38: It seems like the header should have:
- The Marker that says it is encrypted
- The version of the header
- Maybe the type of CIPHER to use?
- The type of the description (sha1 vs md5 vs ???)
- The "description" (probably should called hash_description or something as it is not a text description)
- The nonce

env/env_encrypt2.cc
```diff
@@ -0,0 +1,989 @@
...
+  if (status.ok()) {
+    std::shared_ptr<const CTREncryptionProviderV2> provider;
+    std::unique_ptr<BlockAccessCipherStream> stream;
+    status = ReadRandEncryptionPrefix<RandomAccessFile>(underlying.get(),
+                                                        provider, stream);
+
+    if (status.ok()) {
+      if (provider) {
+        (*result) =
+            std::unique_ptr<RandomAccessFile>(new EncryptedRandomAccessFile(
```
mrambacher, line 699: Shouldn't this be RandomAccessFile2?

include/rocksdb/env_encrypt2.h
```diff
@@ -0,0 +1,321 @@
...
+                          size_t prefix_length)
+      : EncryptedWritableFile(std::move(f), std::move(s), prefix_length) {}
+
+  Status Append(const Slice& data) override;
+
+  Status PositionedAppend(const Slice& data, uint64_t offset) override;
+
+  // Indicates the upper layers if the current WritableFile implementation
+  // uses direct IO.
+  bool use_direct_io() const override { return false; };
```
mrambacher, line 174: Is there a reason direct IO does not work for these files?

build_tools/build_detect_platform
```diff
@@ -774,6 +786,7 @@ echo "JEMALLOC_LIB=$JEMALLOC_LIB" >> "$OUTPUT"
 echo "ROCKSDB_MAJOR=$ROCKSDB_MAJOR" >> "$OUTPUT"
 echo "ROCKSDB_MINOR=$ROCKSDB_MINOR" >> "$OUTPUT"
 echo "ROCKSDB_PATCH=$ROCKSDB_PATCH" >> "$OUTPUT"
+echo "ROCKSDB_OPENSSL_AES_CTR=$ROCKSDB_OPENSSL_AES_CTR" >> "$OUTPUT"
```
mrambacher, line 789: I do not think this line is necessary.  There is nothing set in this file to this variable, and nothing in the Makefile refers to it

env/env_encrypt2.cc
matthewvon: CTR mode is an exception to the rule.  Decrypt and Encrypt are the same.  Why?  Do not know.  Just following examples and seeing good data on the output side.

env/env_encrypt2.cc
```diff
@@ -0,0 +1,989 @@
...
+      --carry;
+    }
+  }  // for
+}
+
+//
+// AES_BLOCK_SIZE assumed to be 16
+//
+typedef union {
+  uint64_t nums[2];
```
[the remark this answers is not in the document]
matthewvon: This related to older implementation.  You are correct.  I will remove it.

env/env_encrypt2_impl.h
matthewvon: This is a design scope issue.  Nothing you suggest is wrong.  But I do not need it.  I put that '0' as code version at the end of the 'Encrypt' header to let anyone redefine what the next bytes contain.  'Tis why I call this branch "changeable encryption".  Someone could decide, 9 months into actual usage of this code, that any one or more of your variables are important and make code to change.  My explicit goal is to not anticipate future needs, but to create wiggle room to accommodate all of them.  AND, create a smooth transition path that does NOT require exporting and re-importing terabytes of data for an encryption design change.

build_tools/build_detect_platform
matthewvon: This sets the environment variable that make passes to the compilers.

env/env_encrypt2.cc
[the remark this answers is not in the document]
matthewvon: removed.  source file is now named env_openssl.cc ... which is likely why "Outdated" appears next to env/env_encrypt2.cc name above.

env/env_encrypt2.cc
```diff
@@ -0,0 +1,989 @@
...
+
+  provider.reset();  // nullptr for provider implies "no encryption"
+  stream.release();
+
+  // Look for encryption marker
+  EncryptMarker marker;
+  Slice marker_slice;
+  status = f->Read(0, sizeof(marker), &marker_slice, marker);
+  if (status.ok()) {
+    if (sizeof(marker) == marker_slice.size() &&
```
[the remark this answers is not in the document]
matthewvon: This is expected:  unencrypted file detected therefore provider is NOT set, but file is not in error.  Stated otherwise, file did not create an error and happens to not be encrypted.  This allows transition to and from encryption.

env/env_encrypt2.cc
```diff
@@ -0,0 +1,989 @@
...
+        ret_val = crypto_shared->EVP_EncryptUpdate(
+            aes_context.get(), (unsigned char*)data, &out_len,
+            (unsigned char*)data, (int)data_size);
+
+        if (1 != ret_val || (int)data_size != out_len) {
+          status = Status::InvalidArgument("EVP_EncryptUpdate failed: ",
+                                           (int)data_size == out_len
+                                               ? "bad return value"
+                                               : "output length short");
+        }
```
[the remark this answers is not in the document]
matthewvon: I reorganized the block.  The EncryptFinal_ex section is called only if there is positive test of 

        if (1 == ret_val && (int)data_size == out_len)

The above status = Status::InvalidArgument is then within the else clause.

env/env_encrypt2.cc
```diff
@@ -0,0 +1,989 @@
...
+
+      memcpy(iv.bytes, nonce_, AES_BLOCK_SIZE);
+      BigEndianAdd128(iv.bytes, block_index);
+
+      ret_val = crypto_shared->EVP_EncryptInit_ex(
+          aes_context.get(), crypto_shared->EVP_aes_256_ctr(), nullptr,
+          key_.key, iv.bytes);
+      if (1 == ret_val) {
+        out_len = 0;
+        ret_val = crypto_shared->EVP_EncryptUpdate(
```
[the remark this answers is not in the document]
matthewvon: Added error detail to match EVP_Encrypt Update failed message.

env/env_encrypt2.cc
matthewvon: The entire point to the environment is to make flexible for others to extend, not to necessarily try to anticipate all variants on this code.  I would prefer to leave that extension to the person that needs it.

env/env_encrypt2.cc
matthewvon: There is no EncryptedRandomAccessFileV2.  EncryptedRandomAccessFile class could be used as is.  Only EncryptedWriteableFile and EncryptedRandRWFile needed adjustment.

include/rocksdb/env_encrypt2.h
matthewvon: Alignment.  These two types of files have a different memory alignment and I lacked the time and experience to verify the impact.

build_tools/build_detect_platform
matthewvon: removed it and rebuild.  seems to all still work.  checking in without line.

env/env_encrypt2.cc
```diff
@@ -0,0 +1,989 @@
...
+#include "util/aligned_buffer.h"
+#include "util/coding.h"
+#include "util/mutexlock.h"
+#include "util/random.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+static port::RWMutex key_lock;
+
+static std::shared_ptr<UnixLibCrypto> crypto_shared;
```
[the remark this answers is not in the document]
matthewvon: GetCrypto implemented and uses the std::call_once construct.

matthewvon:
As of this moment, all checks have again passed.  At least two of the recent failures were infrastructure related ... did not check 178cf31.

util/library_loader.cc
```diff
@@ -0,0 +1,96 @@
...
+    cipher_reset_ = (EVP_CIPHER_CTX_reset_t)functions_["EVP_CIPHER_CTX_reset"];
+    if (nullptr == cipher_reset_) {
+      cipher_reset_ =
+          (EVP_CIPHER_CTX_reset_t)functions_["EVP_CIPHER_CTX_cleanup"];
+    }
+    cipher_free_ = (EVP_CIPHER_CTX_free_t)functions_["EVP_CIPHER_CTX_free"];
+    encrypt_init_ = (EVP_EncryptInit_ex_t)functions_["EVP_EncryptInit_ex"];
+    aes_256_ctr_ = (EVP_aes_256_ctr_t)functions_["EVP_aes_256_ctr"];
+    encrypt_update_ = (EVP_EncryptUpdate_t)functions_["EVP_EncryptUpdate"];
+    encrypt_final_ = (EVP_EncryptFinal_ex_t)functions_["EVP_EncryptFinal_ex"];
```
mrambacher, line 89: Are any of these allowed to be NULL?  Is there an advantage to doing it as LoadSymbols versus a direct call (LoadSymbol) with a better check?

util/library_loader.cc
```diff
@@ -0,0 +1,96 @@
...
+        func.second = tmp_ptr;
+      }
+    }
+  }
+
+  return num_found;
+}
+
+UnixLibCrypto::UnixLibCrypto() : LibraryLoader(crypto_lib_name_) {
+  if (is_valid_) {
```
mrambacher, line 56: Is there a reason to keep this in a separate file rather than in the env_openssl file?

mrambacher (commented):
This code should also be made to work with CMake and on Windows (if possible).

env/env_encryption.cc
```diff
@@ -60,6 +61,10 @@ Status EncryptionProvider::CreateFromString(
   }
   if (id == kCTRProviderName) {
     result->reset(new CTREncryptionProvider());
+#ifdef ROCKSDB_OPENSSL_AES_CTR
+  } else if (id == OpenSSLEncryptionProvider::kName()) {
+    status = NewOpenSSLEncryptionProvider(result);
+#endif
```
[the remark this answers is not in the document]
matthewvon: The CircleCI would not let this file into play without me running clang-format.  I ran it and CircleCI got happy.

PR #7021, opened 2020-06-23
Expose KeyMayExist in the C API

DarkEld3r:


db/c.cc
```diff
@@ -996,6 +996,53 @@ void rocksdb_multi_get_cf(
...
+  std::string time;
+  if (timestamp) {
+    time.assign(timestamp, timestamp_len);
+  }
+  bool found = false;
+  const bool result = db->rep->KeyMayExist(options->rep, Slice(key, key_len),
+                                           &tmp, timestamp ? &time : nullptr,
+                                           value_found ? &found : nullptr);
+  *val_len = tmp.size();
+  *value = CopyString(tmp);
```
pdillinger, line 1015: Though this function does not return a Status, our convention is not to obligate the caller to free anything (except Status message) unless Status is OK. I think the best analogy here is only to obligate the caller to free `*value` if `*value_found`. Any motivation for this way?

db/c.cc
```diff
@@ -996,6 +996,53 @@ void rocksdb_multi_get_cf(
...
+  std::string time;
+  if (timestamp) {
+    time.assign(timestamp, timestamp_len);
+  }
+  bool found = false;
+  const bool result = db->rep->KeyMayExist(
+      options->rep, column_family->rep, Slice(key, key_len), &tmp,
+      timestamp ? &time : nullptr, value_found ? &found : nullptr);
+  *val_len = tmp.size();
+  *value = CopyString(tmp);
```
pdillinger, line 1039: same

include/rocksdb/c.h
```diff
@@ -320,6 +320,17 @@ extern ROCKSDB_LIBRARY_API void rocksdb_multi_get_cf(
     const size_t* keys_list_sizes, char** values_list,
     size_t* values_list_sizes, char** errs);
 
+extern ROCKSDB_LIBRARY_API unsigned int rocksdb_key_may_exist(
```
pdillinger, line 323: And add a comment about exactly when caller is expected to free `*value`. Otherwise, should be a self-explanatory wrapper.

db/c.cc
DarkEld3r: I just mirrored the C++ API, but I agree that your proposition makes the C interface more ergonomic. I have updated the pull request.

PR #7022, opened 2020-06-23
Update Flush policy in PartitionedIndexBuilder on switching from user-key to internal-key mode

akankshamahajan15:
Summary: When format_version is high enough to support user-key and there are index entries for same user key that spans multiple data blocks then it changes from user-key mode to internal-key mode. But the flush policy is not reset to point to Block Builder of internal-keys. After this switch, no entries are added to user key index partition result, thus it never triggers flushing the block.

Fix: After adding the entry in sub_builder_index_, if there is a switch from user-key to internal-key, then flush policy is updated to point to Block Builder of internal-keys index partition.

Test Plan: 1. make check -j64
           2. Added one unit test case

Reviewers:

Subscribers:

Tasks: T68809728

Tags:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -22,6 +22,7 @@
 
 ### Bug Fixes
 * Fail recovery and report once hitting a physical log record checksum mismatch, while reading MANIFEST. RocksDB should not continue processing the MANIFEST any further.
+* Fix a bug when index_type == kTwoLevelIndexSearch in PartitionedIndexBuilder to update FlushPolicy to point to internal key partitioner when it changes from user-key mode to internal-key mode in index partition.
 
 ## 6.11 (6/12/2020)
 ### Bug Fixes
--- a/db/db_test2.cc
+++ b/db/db_test2.cc
@@ -144,6 +144,38 @@ INSTANTIATE_TEST_CASE_P(TestReadOnlyWithCompressedCache,
                         TestReadOnlyWithCompressedCache,
                         ::testing::Combine(::testing::Values(-1, 100),
                                            ::testing::Bool()));
+
+class PartitionedIndexTestListener : public EventListener {
+ public:
+  void OnFlushCompleted(DB* /*db*/, const FlushJobInfo& info) override {
+    ASSERT_GT(info.table_properties.index_partitions, 1);
+    ASSERT_EQ(info.table_properties.index_key_is_user_key, 0);
+  }
+};
+
+TEST_F(DBTest2, PartitionedIndexUserToInternalKey) {
+  BlockBasedTableOptions table_options;
+  Options options = CurrentOptions();
+  table_options.index_type = BlockBasedTableOptions::kTwoLevelIndexSearch;
+  PartitionedIndexTestListener* listener = new PartitionedIndexTestListener();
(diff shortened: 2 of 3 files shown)


db/db_test2.cc
```diff
@@ -144,6 +144,38 @@ INSTANTIATE_TEST_CASE_P(TestReadOnlyWithCompressedCache,
...
+
+class PartitionedIndexTestListener : public EventListener {
+ public:
+  void OnFlushCompleted(DB* /*db*/, const FlushJobInfo& info) override {
+    ASSERT_GT(info.table_properties.index_partitions, 1);
+    ASSERT_EQ(info.table_properties.index_key_is_user_key, 0);
+  }
+};
+
+TEST_F(DBTest2, PartitionedIndexUserToInternalKey) {
```
ajkr, line 156: nice test!

table/block_based/index_builder.cc
```diff
@@ -162,7 +167,12 @@ void PartitionedIndexBuilder::AddIndexEntry(
                                       first_key_in_next_block, block_handle);
     sub_index_last_key_ = std::string(*last_key_in_current_block);
     if (sub_index_builder_->seperator_is_key_plus_seq_) {
-      // then we need to apply it to all sub-index builders
+      // then we need to apply it to all sub-index builders and reset
+      // flush_policy to point to Block Builder of sub_index_builder_ that store
+      // internal keys.
+      flush_policy_.reset(FlushBlockBySizePolicyFactory::NewFlushBlockPolicy(
```
ajkr, line 173: This is an allocation per index entry which might be nontrivial. I think the easy way to avoid these extra allocations is to reset the flush policy only when `!seperator_is_key_plus_seq_`.

akankshamahajan15: Initially I added the check but then number of partitions got changed (it was changed from 11 to 2). So I couldn't understand why the value got changed by adding additional check ```!seperator_is_key_plus_seq_```.

ajkr: I see you added it back. Does it behave as expected now (produce partitions around `metadata_block_size` bytes) or is there still an unexpectedly small number of partitions?

akankshamahajan15: Oh my bad. The check  at line 132 is fine so i added that but by adding check here (line 170) partitions are still 2 (I forgot to remove this check while uploading the diff). I am still not sure why this is the case. I have added logs to look further for debugging.

ajkr: discussed offline, seems it is producing the expected number of partitions now.

HISTORY.md
```diff
@@ -22,6 +22,7 @@
 
 ### Bug Fixes
 * Fail recovery and report once hitting a physical log record checksum mismatch, while reading MANIFEST. RocksDB should not continue processing the MANIFEST any further.
+* Fix a bug in PartitionedIndexBuilder to update FlushPolicy to point to internal key partitioner when it changes from user-key mode to internal-key mode in index partition.
```
ajkr, line 25: For release note it's nice to describe in terms of affected user APIs / options, so people can quickly check whether they're affected (in this case the affected option is `index_type == kTwoLevelIndexSearch`)

PR #7023, opened 2020-06-24
Update HISTORY.md to include the Public API Change for DB::OpenForReadonly introduced earlier

gg814:
Summary:
`DB::OpenForReadOnly()` now returns `Status::NotFound` when the specified DB directory does not exist. Previously the error returned depended on the underlying `Env`.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -11,7 +11,7 @@
 
 ### Public API Change
 * `DB::GetDbSessionId(std::string& session_id)` is added. `session_id` stores a unique identifier that gets reset every time the DB is opened. This DB session ID should be unique among all open DB instances on all hosts, and should be unique among re-openings of the same or other DBs. This identifier is recorded in the LOG file on the line starting with "DB Session ID:".
-* `DB::OpenForReadOnly()` now returns `Status::NotFound` when the specified DB directory does not exist. Previously the error returned depended on the underlying `Env`.
+* `DB::OpenForReadOnly()` now returns `Status::NotFound` when the specified DB directory does not exist. Previously the error returned depended on the underlying `Env`. This change is available in all 6.11 releases as well.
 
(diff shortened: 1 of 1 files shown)


gg814:
> Needs to go in the 6.11 section, right?

I think @ajkr wanted me to include it in Unreleased?

ajkr:
> > Needs to go in the 6.11 section, right?
> 
> I think @ajkr wanted me to include it in Unreleased?

Right, that was my suggestion. I think it's simpler if the HISTORY.md is append only (i.e., only the unreleased section can be mutated). It'd be slightly confusing if the section titled "6.11 (6/12/2020)" looked different on 6/23/2020 vs 6/24/2020, or on master branch vs. a release branch. On the release branch I put these in 6.11.1 - https://github.com/facebook/rocksdb/blob/6.11.fb/HISTORY.md.

ajkr:
> > > Needs to go in the 6.11 section, right?
> > 
> > 
> > I think @ajkr wanted me to include it in Unreleased?
> 
> Right, that was my suggestion. I think it's simpler if the HISTORY.md is append only (i.e., only the unreleased section can be mutated). It'd be slightly confusing if the section titled "6.11 (6/12/2020)" looked different on 6/23/2020 vs 6/24/2020, or on master branch vs. a release branch. On the release branch I put these in 6.11.1 - https://github.com/facebook/rocksdb/blob/6.11.fb/HISTORY.md. So users of a 6.11 release will be aware of this change already.

To clarify, by "confusing" I am referring specifically to confusion that can result in a user missing a release note. Both ways are confusing, but I believe this way it's impossible to miss no matter the user's release note reading technique :P.

Peter suggested mentioning it's available since 6.11 in the release note. How about appending "This change is available in all 6.11 releases as well." ?

PR #7032, opened 2020-06-25
Generalize BackupEngine naming option for share_files_with_checksum SSTs and revert BackupEngine::VerifyBackup to check only file sizes by default

gg814:
Summary:
`bool BackupableDBOptions::new_naming_for_backup_files` is updated to `BackupTableNameOption BackupableDBOptions::share_files_with_checksum_naming`, where `BackupTableNameOption` is an `enum` type with two enumerators `kChecksumAndFileSize` and `kChecksumAndFileSize`. This opens up possibilities of extenting the current naming scheme for backup table files. By default, `BackupTableNameOption BackupableDBOptions::share_files_with_checksum_naming` is set to `kChecksumAndDbSessionId`.

Revert `BackupEngine::VerifyBackup` to only check file sizes by default.

Also fix the construction of the `SstFileDumper` in `GetFileDbIdentities` by setting a proper `Env` of the `Options` passed in the constructor.

Test plan:
make check


HISTORY.md
```diff
@@ -17,6 +17,7 @@
...
+* `bool BackupableDBOptions::new_naming_for_backup_files` is updated to `BackupTableNameOption BackupableDBOptions::share_files_with_checksum_naming`, where `BackupTableNameOption` is an `enum` type with two enumerators `kChecksumAndFileSize` and `kChecksumAndFileSize`. This opens up possibilities of extenting the current naming scheme for backup table files. By default, `BackupTableNameOption BackupableDBOptions::share_files_with_checksum_naming` is set to `kChecksumAndDbSessionId`. As before, when there is no `db_session_id` available in the table file, we use `file_size` as a fallback, and `kChecksumAndDbSessionId` comes into effect only when both `share_files_with_checksum` and `share_table_files` are true.
```
ajkr, line 20: Although I'm usually the one to argue for keeping release notes append only, I think in this case -- where we landed a new API that we expected to change very shortly afterwards, and then changed it -- we can just update the existing note in place :P. Otherwise it feels like the user gets burdened with extra complexity due to our workflow.

include/rocksdb/utilities/backupable_db.h
```diff
@@ -131,7 +140,7 @@ struct BackupableDBOptions {
   // and share_table_files are true. In the cases of old table files where no
```
ajkr, line 140: On line 137 the default should be `kChecksumAndDbSessionId`.

gg814:
Thanks for the comments! I've updated the PR accordingly :)

include/rocksdb/utilities/backupable_db.h
```diff
@@ -122,16 +131,19 @@ struct BackupableDBOptions {
   // Default: INT_MAX
   int max_valid_backups_to_open;
 
-  // If true, backup SST filenames consist of file_number, crc32c, db_session_id
-  // if false, backup SST filenames consist of file_number, crc32c, file_size
+  // Naming option for share_files_with_checksum table files
+  // If it is set to kChecksumAndDbSessionId,
```
pdillinger, line 135: I think these details should be comments on the enumerators.

We should also mention (again) that the naming scheme is not an issue for version upgrade or backup directory compatibility (for reasons I mentioned above).

gg814:
Thanks for the comments. HISTORY.md and code comments are updated accordingly.

include/rocksdb/utilities/backupable_db.h
```diff
@@ -122,16 +135,23 @@ struct BackupableDBOptions {
   // Default: INT_MAX
   int max_valid_backups_to_open;
 
-  // If true, backup SST filenames consist of file_number, crc32c, db_session_id
-  // if false, backup SST filenames consist of file_number, crc32c, file_size
+  // Naming option for share_files_with_checksum table files.
+  // The default behavior for this option fixes the backup file name collision
```
ajkr, line 139: In this and the next paragraph, it feels like we're focusing on the fix/change. But for people reading the code (rather than a PR), it's probably not necessary or easy for them to know about a past change, especially as more time passes. Maybe we can describe it in a more time-independent way with statements like "`kChecksumAndFileSize` is susceptible to collision", "`kChecksumAndDbSessionId` is immune to collision", "Modifying this option cannot introduce a downgrade compatibility issue", etc.

gg814: Good point :)

PR #7034, opened 2020-06-25
Fix data race to VersionSet::io_status_

riversand963:
Summary:
After #6949 , VersionSet::io_status_ can be concurrently accessed by multiple
threads without lock, causing tsan test to fail. For example, a bg flush thread
resets io_status_ before calling LogAndApply(), while another thread already in
the process of LogAndApply() reads io_status_. This is a bug.

We do not have to reset io_status_ each time we call LogAndApply(). io_status_
is part of the state of VersionSet, and it indicates the outcome of preceding
MANIFEST/CURRENT files IO operations. Its value should be updated only when:

1. MANIFEST/CURRENT files IO fail for the first time.
2. MANIFEST/CURRENT files IO succeed as part of recovering from a prior
   failure without process restart, e.g. calling Resume().

Test Plan (devserver):
COMPILE_WITH_TSAN=1 make check
COMPILE_WITH_TSAN=1 make db_test2
./db_test2 --gtest_filter=DBTest2.CompactionStall

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -3878,10 +3878,6 @@ Status VersionSet::ProcessManifestWrites(
   }
 #endif  // NDEBUG
 
-  uint64_t new_manifest_file_size = 0;
-  Status s;
-  IOStatus io_s;
-
   assert(pending_manifest_file_number_ == 0);
   if (!descriptor_log_ ||
       manifest_file_size_ > db_options_->max_manifest_file_size) {
@@ -3911,6 +3907,9 @@ Status VersionSet::ProcessManifestWrites(
     }
   }
 
+  uint64_t new_manifest_file_size = 0;
+  Status s;
+  IOStatus io_s;
   {
     FileOptions opt_file_opts = fs_->OptimizeForManifestWrite(file_options_);
     mu->Unlock();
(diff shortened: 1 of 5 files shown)


zhichao-cao:
Just double check, if the version->io_status_ is not ok, resume will be called. During this process, if ProcessManifestWrites is successful, io_status_ will be change back to IOStatus::OK() (the else case test io_status_.IsIOError()). Therefore, we do not need the reset anymore.

riversand963:
> Just double check, if the version->io_status_ is not ok, resume will be called. During this process, if ProcessManifestWrites is successful, io_status_ will be change back to IOStatus::OK() (the else case test io_status_.IsIOError()). Therefore, we do not need the reset anymore.

That's correct.

riversand963:
hmm, I think there are more data races to `VersionSet::io_status_`. Let me update.

riversand963:
@zhichao-cao Thanks for the previous review. Could you take another look?

db/version_set.cc
```diff
@@ -4044,6 +4039,14 @@ Status VersionSet::ProcessManifestWrites(
     mu->Lock();
   }
 
+  if (!io_s.ok()) {
```
zhichao-cao, line 4042: So the general fix logic is, io_status_ is only updated here. In other places, we use local variable to keep the status in e.g., io_s.

riversand963: Yes, access to io_status_ should be protected by the db_mutex.

zhichao-cao: Yeah, I notice that WriteCurrentStateToManifest is not protected by the db_mutex

zhichao-cao:
Thanks for the fix. I think it is good to go now.

PR #7043, opened 2020-06-29
Skip unnecessary allocation for mmap reads under 5000 bytes

ajkr:
With mmap enabled on an uncompressed file, we were previously always doing a heap allocation to obtain the scratch buffer for `RandomAccessFileReader::Read()`. However, that allocation was unnecessary as the underlying file reader returned a pointer into its mapped memory, not the provided scratch buffer. This PR makes passes the `BlockFetcher`'s inline buffer as the scratch buffer if the data block is small enough (less than `kDefaultStackBufferSize` bytes, currently 5000). Ideally we would not pass a scratch buffer at all for an mmap read; however, the `RandomAccessFile::Read()` API guarantees such a buffer is provided, and non-standard implementations may be relying on it even when `Options::allow_mmap_reads == true`. In that case, this PR still works but introduces an extra copy from the inline buffer to a heap buffer.

### DIFF
--- a/table/block_fetcher.cc
+++ b/table/block_fetcher.cc
@@ -99,10 +99,28 @@ inline bool BlockFetcher::TryGetCompressedBlockFromPersistentCache() {
 
 inline void BlockFetcher::PrepareBufferForBlockFromFile() {
   // cache miss read from device
-  if (do_uncompress_ &&
+  if ((do_uncompress_ || ioptions_.allow_mmap_reads) &&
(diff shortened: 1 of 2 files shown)


ajkr:
Thanks for the review, @cheng-chang! I wonder if in the long run we should introduce a separate function from `RandomAccessFile::Read()` that doesn't guarantee a scratch buffer is provided.

PR #7044, opened 2020-06-29
Reduce `env_->GetChildren()` calls in DBImpl::Recover()

roghnin:
Summary:
There currently exist multiple `GetChildren()` calls in `DBImpl::Recover()`, which can be expensive in cases of distributed file systems.
This pull request try to call `DBImpl::Recover()` of each necessary directory only _once_ and reuse the results in the places of repeated calls in current code.

Test Plan:
Run `make check` and use the default test suite. The modified code should be semantically identical to the current code. As a proof of this solution, we may optionally deploy the system onto a (real or simulated) distributed system and expect reduced latency caused by manifest fetching.


db/db_impl/db_impl_open.cc
```diff
@@ -368,7 +370,7 @@ Status DBImpl::Recover(
     if (!s.ok()) {
       return s;
     }
-
+
```
riversand963, line 373: White spaces in empty line should be removed. You can run `make format`.

db/db_impl/db_impl_open.cc
```diff
@@ -356,6 +356,8 @@ Status DBImpl::Recover(
 
   bool is_new_db = false;
   assert(db_lock_ == nullptr);
+  std::vector<std::string> dbname_children;
+  Status dbname_children_s = env_->GetChildren(dbname_, &dbname_children);
```
riversand963, line 360: If `best_efforts_recovery == false`, we may not have to call `GetChildren()` on `dbname`.

riversand963: I think we should not call `GetChildren()` here if `best_efforts_recovery == false`.

db/db_impl/db_impl_open.cc
```diff
@@ -532,68 +532,76 @@ Status DBImpl::Recover(
     // Note that prev_log_number() is no longer used, but we pay
     // attention to it in case we are recovering a database
     // produced by an older version of rocksdb.
-    std::vector<std::string> filenames;
+    std::vector<std::string>* filenames_p = nullptr;
+    std::vector<std::string> wal_dir_files;
     if (!immutable_db_options_.best_efforts_recovery) {
-      s = env_->GetChildren(immutable_db_options_.wal_dir, &filenames);
```
riversand963, line 537: It's OK to keep this.

db/version_set.cc
```diff
@@ -4720,7 +4720,9 @@ Status VersionSet::Recover(
 namespace {
 class ManifestPicker {
  public:
-  explicit ManifestPicker(const std::string& dbname, FileSystem* fs);
+  explicit ManifestPicker(const std::string& dbname,
+                          std::vector<std::string>* dbname_children,
```
riversand963, line 4724: `dbname_children` can be of type `const std::vector<std::string>&`.

db/version_set.cc
```diff
@@ -4720,7 +4720,9 @@ Status VersionSet::Recover(
...
+                          Status dbname_children_s);
```
riversand963, line 4725: const Status&

db/version_set.cc
```diff
@@ -4729,25 +4731,31 @@ class ManifestPicker {
 
  private:
   const std::string& dbname_;
-  FileSystem* const fs_;
   // MANIFEST file names(s)
   std::vector<std::string> manifest_files_;
   std::vector<std::string>::const_iterator manifest_file_iter_;
+  std::vector<std::string>* dbname_children_p;
```
riversand963, line 4737: The `_p` is not needed, while we end member variable of *class* with `_`.

db/db_impl/db_impl_open.cc
roghnin: > I think we should not call `GetChildren()` here if `best_efforts_recovery == false`.

Note that at the end of `Recover()`, the result of `GetChildren()` will always be required if `read_only == true`, regardless of the value of `best_efforts_recovery`.

I think the only case where `Recover()` need not be called is `read_only == false && best_efforts_recovery == false`. We can still avoid calling it in this case, though.

riversand963: Yeah, `read_only == false && best_efforts_recovery == false` is the common case.

db/db_impl/db_impl_open.cc
roghnin: `make format` reports an error in `clang-format-diff.py`:
`TypeError: initial_value must be str or None, not bytes`.
Do I need to pass in any arguments?

db/version_set.cc
```diff
@@ -4729,25 +4731,32 @@ class ManifestPicker {
...
 };
 
-ManifestPicker::ManifestPicker(const std::string& dbname, FileSystem* fs)
-    : dbname_(dbname), fs_(fs) {}
+ManifestPicker::ManifestPicker(const std::string& dbname,
+                               const std::vector<std::string>& dbname_children,
+                               const Status& dbname_children_s)
+    : dbname_(dbname),
+      dbname_children_(dbname_children),
+      status_(dbname_children_s) {}
```
siying, line 4746: Does it need to be named `dbname_children_s` which has little relationship with class ManifestPicker itself?

Also, consider to read the naming section of Google C++ Style: https://google.github.io/styleguide/cppguide.html#Naming and try to think of a better name.

roghnin: Thanks for the advice. I'll for sure use a better name if we decide to keep this piece of code in the end.

The biggest issue is that the result of `getChildren()` in `DBImpl::Recover()`, which is passed in here, is supposed to be identical to the result of `fs_->getChildren()` in the original code. But it wasn't.

So I'm going to revert this change in my next commit, and hope we can roll it back once we figured out how to use existing `getChildren()` results from `Recover()`.

riversand963: The reason for the difference you mentioned above is because `NewDB()` may be called if RocksDB decides to create a new database. In `NewDB()`, a `MANIFEST-000001` will be created. You may consider modifying this function so that the MANIFEST file name can be returned.

db/db_impl/db_impl_open.cc
```diff
@@ -356,6 +356,11 @@ Status DBImpl::Recover(
 
   bool is_new_db = false;
   assert(db_lock_ == nullptr);
+  std::vector<std::string> dbname_children;
+  Status dbname_children_s;
+  if (read_only || immutable_db_options_.best_efforts_recovery) {
+    dbname_children_s = env_->GetChildren(dbname_, &dbname_children);
```
riversand963, line 362: I don't think we need a separate `Status dbname_children_s` variable here. You can just define a variable `s` inside the `if` block.

db/db_impl/db_impl_open.cc
```diff
@@ -532,68 +534,69 @@ Status DBImpl::Recover(
     // Note that prev_log_number() is no longer used, but we pay
     // attention to it in case we are recovering a database
     // produced by an older version of rocksdb.
-    std::vector<std::string> filenames;
+    std::vector<std::string> wal_dir_children;
```
riversand963, line 537: How about `files_in_wal_dir`?

db/db_impl/db_impl_open.cc
```diff
@@ -532,68 +534,69 @@ Status DBImpl::Recover(
...
-    } else if (!s.ok()) {
-      return s;
-    }
+      s = env_->GetChildren(immutable_db_options_.wal_dir, &wal_dir_children);
+      if (s.IsNotFound()) {
+        return Status::InvalidArgument("wal_dir not found",
+                                       immutable_db_options_.wal_dir);
+      } else if (!s.ok()) {
+        return s;
+      }
```
riversand963, line 545: The following code blocks does not have to be moved into the big `if (!immutable_db_options_.best_efforts_recovery)` block.

db/db_impl/db_impl_open.cc
```diff
@@ -532,68 +534,69 @@ Status DBImpl::Recover(
...
+        if (ParseFileName(wal_dir_children[i], &number, &type) &&
+            type == kLogFile) {
+          if (is_new_db) {
+            return Status::Corruption(
+                "While creating a new Db, wal_dir contains "
+                "existing log file: ",
+                wal_dir_children[i]);
+          } else {
+            logs.push_back(number);
+          }
```
riversand963, line 560: Ditto

db/db_impl/db_impl_open.cc
```diff
@@ -532,68 +534,69 @@ Status DBImpl::Recover(
...
+          for (auto& log : logs) {
+            std::string fname = LogFileName(immutable_db_options_.wal_dir, log);
+            uint64_t bytes;
+            s = env_->GetFileSize(fname, &bytes);
+            if (s.ok()) {
+              if (bytes > 0) {
+                return Status::Corruption(
+                    "error_if_data_exists_in_logs is set but there are data "
+                    " in log files.");
+              }
```
riversand963, line 579: Ditto.

db/db_impl/db_impl_open.cc
```diff
@@ -532,68 +534,69 @@ Status DBImpl::Recover(
...
+                            &corrupted_log_found);
+        if (corrupted_log_found && recovered_seq != nullptr) {
+          *recovered_seq = next_sequence;
+        }
+        if (!s.ok()) {
+          // Clear memtables if recovery failed
+          for (auto cfd : *versions_->GetColumnFamilySet()) {
+            cfd->CreateNewMemtable(*cfd->GetLatestMutableCFOptions(),
+                                   kMaxSequenceNumber);
+          }
```
riversand963, line 599: Ditto.

db/db_impl/db_impl_open.cc
roghnin: This `dbname_children_s` will later be used at least once at the end of the function when `read_only == true`, and may get passed in to TryRecover(). Having `s` in the if block may lose this status.

db/db_impl/db_impl_open.cc
roghnin: Sure.

db/version_set.cc
roghnin: So this means that we can simply use the (known) manifest file name returned by (updated) `NewDB()` and create a result, instead of running `GetChildren()` again?

riversand963: Yes if `NewDB()` is called.

db/db_impl/db_impl_open.cc
riversand963: I think we can delay calling `GetChildren()` while still being able to cache. How about the following:
1. best_efforts_recovery is true. Call GetChildren() once when we determine whether to call `NewDB()`. Use the cached result to compute the set of MANIFEST files which we pass to `ManifestPicker()` when calling `TryRecover`. if `read_only` happens to be true, we can also reuse this result when we determine the `options_file_number_`. Although I tend to think `best_efforts_recovery` should be incompatible with read only mode since it will delete SST files. Maybe you can add a PR to disallow the combination of opening a `DBImplReadOnly` with `best_efforts_recovery == true`.
2. best_effort_recovery is false. The first place we call `GetChildren()` should be where we try to find the WAL files. Later, if `read_only && wal_dir == db_name_`, then we can reuse the result.

riversand963: > This `dbname_children_s` will later be used at least once at the end of the function when `read_only == true`, and may get passed in to TryRecover(). Having `s` in the if block may lose this status.

Is it possible just to use a single `Status s` in this function?

roghnin: > I think we can delay calling `GetChildren()` while still being able to cache. How about the following:
> 
> 1. best_efforts_recovery is true. Call GetChildren() once when we determine whether to call `NewDB()`. Use the cached result to compute the set of MANIFEST files which we pass to `ManifestPicker()` when calling `TryRecover`. if `read_only` happens to be true, we can also reuse this result when we determine the `options_file_number_`. Although I tend to think `best_efforts_recovery` should be incompatible with read only mode since it will delete SST files. Maybe you can add a PR to disallow the combination of opening a `DBImplReadOnly` with `best_efforts_recovery == true`.
> 2. best_effort_recovery is false. The first place we call `GetChildren()` should be where we try to find the WAL files. Later, if `read_only && wal_dir == db_name_`, then we can reuse the result.

I'll try to address this in my next commit. I think we may be able to do this, although the code may look a bit messy.

roghnin: > > This `dbname_children_s` will later be used at least once at the end of the function when `read_only == true`, and may get passed in to TryRecover(). Having `s` in the if block may lose this status.
> 
> Is it possible just to use a single `Status s` in this function?

It's tempting, but I don't think so. For example, when `read_only == false`, the first call to `getChildren()` (originally) don't care about returned status and `s` is given pre-programmed values.

db/db_impl/db_impl_open.cc
```diff
@@ -379,10 +385,9 @@ Status DBImpl::Recover(
       s = env_->FileExists(current_fname);
     } else {
       s = Status::NotFound();
-      std::vector<std::string> files;
-      // No need to check return value
-      env_->GetChildren(dbname_, &files);
-      for (const std::string& file : files) {
+      files_status = env_->GetChildren(dbname_, &db_file_names);
```
riversand963, line 388: How about
```
Status io_s = env_->GetChildren(dbname_, &db_file_names);
if (!io_s.ok()) {
  s = io_s;
  db_file_names.clear();
}
```
Basically, `io_s` can be a local var in this code block.

db/db_impl/db_impl_open.cc
```diff
@@ -445,8 +451,11 @@ Status DBImpl::Recover(
   if (!immutable_db_options_.best_efforts_recovery) {
     s = versions_->Recover(column_families, read_only, &db_id_);
   } else {
+    if (read_only) {  // TODO: get rid of this case?
+      files_status = env_->GetChildren(dbname_, &db_file_names);
+    }
     s = versions_->TryRecover(column_families, read_only, &db_id_,
-                              &missing_table_file);
+                              &missing_table_file, db_file_names, files_status);
```
riversand963, line 458: If `files_status` is non-ok, we do not call this. In fact, we should have already returned at line 422. Therefore, removing `files_status`.

db/db_impl/db_impl_open.cc
```diff
@@ -445,8 +451,11 @@ Status DBImpl::Recover(
   if (!immutable_db_options_.best_efforts_recovery) {
     s = versions_->Recover(column_families, read_only, &db_id_);
   } else {
+    if (read_only) {  // TODO: get rid of this case?
```
riversand963, line 454: Move the comment to a new line. This is not a simple note, and more text may be expected, thus moving to new line so that adding new text is easier.

db/db_impl/db_impl_open.cc
```diff
@@ -533,8 +542,15 @@ Status DBImpl::Recover(
     // attention to it in case we are recovering a database
     // produced by an older version of rocksdb.
     std::vector<std::string> filenames;
+    std::vector<std::string>* files_in_wal_dir = &filenames;
     if (!immutable_db_options_.best_efforts_recovery) {
-      s = env_->GetChildren(immutable_db_options_.wal_dir, &filenames);
+      if (immutable_db_options_.wal_dir == dbname_) {
```
riversand963, line 547: At this point, we do not need to distinguish between `wal_dir == dbname_` and `wal_dir != dbname_`. We can just do
```
s = env_->GetChildren(wal_dir, &filenames);
```

db/db_impl/db_impl_open.cc
```diff
@@ -606,21 +623,24 @@ Status DBImpl::Recover(
     // updated to versions_->NewFileNumber() in RenameTempFileToOptionsFile.
     std::vector<std::string> file_names;
     if (s.ok()) {
-      s = env_->GetChildren(GetName(), &file_names);
+      if (dbname_ != immutable_db_options_.wal_dir) {
+        // GetChildren() on dbname_ was NOT called above.
+        files_status = env_->GetChildren(dbname_, &db_file_names);
```
riversand963, line 628:
```
s = env_->GetChildren(dbname_, &db_file_names);
```

db/db_impl/db_impl_open.cc
```diff
@@ -606,21 +623,24 @@ Status DBImpl::Recover(
     // updated to versions_->NewFileNumber() in RenameTempFileToOptionsFile.
     std::vector<std::string> file_names;
     if (s.ok()) {
-      s = env_->GetChildren(GetName(), &file_names);
+      if (dbname_ != immutable_db_options_.wal_dir) {
```
riversand963, line 626: This `if` can be combined with the `if` above.

db/db_impl/db_impl_open.cc
```diff
@@ -606,21 +623,24 @@ Status DBImpl::Recover(
...
+        // GetChildren() on dbname_ was NOT called above.
+        files_status = env_->GetChildren(dbname_, &db_file_names);
+      }
+      s = files_status;
```
riversand963, line 630: This can be removed.

db/db_impl/db_impl_open.cc
```diff
@@ -606,21 +623,24 @@ Status DBImpl::Recover(
...
     }
     if (s.ok()) {
       uint64_t number = 0;
       uint64_t options_file_number = 0;
       FileType type;
-      for (const auto& fname : file_names) {
+      for (const auto& fname : db_file_names) {
```
riversand963, line 636: The `file_names` can be removed.

db/version_set.cc
```diff
@@ -4729,7 +4729,9 @@ Status VersionSet::Recover(
 namespace {
 class ManifestPicker {
  public:
-  explicit ManifestPicker(const std::string& dbname, FileSystem* fs);
+  explicit ManifestPicker(const std::string& dbname,
+                          const std::vector<std::string>& file_names,
+                          const Status& files_status);
```
riversand963, line 4734: `files_status` can be removed. If it's non-ok, we shouldn't have reached here.

db/version_set.cc
```diff
@@ -4738,25 +4740,26 @@ class ManifestPicker {
 
  private:
   const std::string& dbname_;
-  FileSystem* const fs_;
   // MANIFEST file names(s)
   std::vector<std::string> manifest_files_;
   std::vector<std::string>::const_iterator manifest_file_iter_;
+  const std::vector<std::string>& file_names_;
```
riversand963, line 4746: `file_names_` as a member variable, can be removed, since populating `manifest_files_` is light weight and can now be done in constructor. We didn't do it in the previous implementation because we avoid doing IO in constructors.

db/db_impl/db_impl_open.cc
```diff
@@ -356,6 +360,8 @@ Status DBImpl::Recover(
 
   bool is_new_db = false;
   assert(db_lock_ == nullptr);
+  std::vector<std::string> db_file_names;
```
riversand963, line 363: `db_file` is ambiguous. We can just use `files_in_dbname` for clarity.

db/db_impl/db_impl_open.cc
```diff
@@ -356,6 +360,8 @@ Status DBImpl::Recover(
...
+  Status files_status;
```
riversand963, line 364: `files_status` can be removed.

db/db_impl/db_impl_open.cc
riversand963: We can avoid this to make code cleaner.
Pass both `fs` and `db_file_names` to `ManifestPicker`. If `db_file_names` is empty, then use `fs` to list directory.

db/db_impl/db_impl_open.cc
```diff
@@ -533,8 +542,15 @@ Status DBImpl::Recover(
     // attention to it in case we are recovering a database
     // produced by an older version of rocksdb.
     std::vector<std::string> filenames;
+    std::vector<std::string>* files_in_wal_dir = &filenames;
```
riversand963, line 545: `files_in_wal_dir` is not needed.

db/version_set.cc
roghnin: I lean towards leaving the `manifest_files_` population in `SeekToFirstManifest()`, since if `file_names_` is empty, we'll still need to do IO in this process.

db/db_impl/db_impl_open.cc
roghnin: Note that `filenames` is in an inner scope invisible to the `if(read_only)` scope at the bottom. If we don't distinguish these two, then we'll have to move this variable to an outer scope in order to make it reusable.

db/db_impl/db_impl.h
```diff
@@ -852,8 +852,8 @@ class DBImpl : public DB {
   InstrumentedMutex* mutex() const { return &mutex_; }
 
   // Initialize a brand new DB. The DB directory is expected to be empty before
-  // calling it.
-  Status NewDB();
+  // calling it. Optionally return new manifest file name.
+  Status NewDB(std::vector<std::string>* new_filenames = nullptr);
```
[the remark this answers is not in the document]
roghnin: There are other usages of NewDB() (e.g. repair.cc:188)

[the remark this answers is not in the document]
riversand963: Yeah, this is the only other case.

db/db_impl/db_impl_open.cc
riversand963: I think moving it to an outer scope sounds fine.

db/db_impl/db_impl_open.cc
roghnin: I may not be able to do this naturally. See comment in new commit.

db/version_set.cc
riversand963: We can still do this without having `file_names_` as member variable.
In constructor, you pass `file_names` with which you can populate the `manifest_files_`.
- If `manifest_files_` becomes not-empty afterwards, no need to populate it in `SeekToFirstManifest`.
- If `manifest_files_` is still empty when we reach `SeekToFirstManifest`, then we do IO with a local variable `files` obtained by calling `fs->GetChildren()`.

riversand963:
Another pass.

db/db_impl/db_impl_open.cc
```diff
@@ -614,14 +614,16 @@ Status DBImpl::Recover(
...
+    if (s.ok()) {
+      if (dbname_ != immutable_db_options_.wal_dir){
+        // GetChildren() on dbname_ was NOT called above.
+        s = env_->GetChildren(dbname_, &files_in_dbname);
+        filenames = &files_in_dbname;
+      } else {
+        s = Status::OK();
+        filenames = &files_in_wal_dir;
+      }
+    }
```
roghnin: I don't feel so comfortable here... any thoughts?

riversand963: What is the concern?

roghnin: It seems that it just moved the pointer juggling from `wal_dir`-related section to here :) anyways I'm OK with it as long as it looks good to you.

db/db_impl/db_impl.h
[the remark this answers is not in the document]
roghnin: > Yeah, this is the only other case.

I feel hesitant to modify the other use case, as the "nullptr" as argument makes less sense there. Please let me know if that's actually better.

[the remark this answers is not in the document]
riversand963: Declaring a function with default parameter has one advantage: it makes you write less code. This comes at the cost of less readability: when you see `NewDB()`, you need to check its declaration to figure out what the signature is.
In this case, the benefit of less code is minimal since it is used only twice in the entire codebase. The cons outweigh the pros.
Passing a nullptr means we do not need return value. It still makes sense: sometimes we need a out-arg as return value, sometimes we do not.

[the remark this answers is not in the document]
riversand963: Plus, writing it as `NewDB(/*new_filenames=*/nullptr)` is preferred. We have been following this convention to further improve readability.

db/db_impl/db_impl_open.cc
riversand963: When we call `GetChildren()` on `wal_dir` for the first time, we know it must be empty since `best_efforts_recovery` must be false, and we must not have called `GetChildren(dbname)` at all. Therefore, there is no need to check whether `dbname_ == wal_dir`. Even if they are equal, we need to call `GetChildren(wal_dir)`.

In the final `read_only` if block, we do not have to call `GetChildren()` on `dbname` again if
- best_efforts_recovery is true, or
- best_efforts_recovery is false and dbname_ is equal to wal_dir.

roghnin: Gotcha. Thanks a lot!

roghnin: > * best_efforts_recovery is true, or

This is assuming we called `getChildren()` at `TryRecover()`, right? If that's the case, we need some way to cache the `getChildren()` result in `ManifestPicker`. I think we need to roll back to the older version where `files_in_dbname` is a member in `ManifestPicker` so that we can have the result returned.

db/db_impl/db_impl.h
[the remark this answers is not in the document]
roghnin: Super clear! Thanks!

db/db_impl/db_impl_open.cc
riversand963: We can let `TryRecover()` return the result of `GetChildren()`. For example (without considering error handling for simplicity). Still no need to add member variable `file_names_` to `ManifestPicker`.
```
TryRecovery(......, bool read_only, ......, std::vector<std::string>& files_in_dbname) {
  if (read_only) {
    assert(files_in_dbname.empty());
    fs_->GetChildren(&files_in_dbname);
  }
  ManifestPicker(..., files_in_dbname);
  ...
}
```

riversand963: Or just do a `GetChildren()` if `best_efforts_recovery && read_only`.

db/db_impl/db_impl_open.cc
```diff
@@ -356,6 +360,8 @@ Status DBImpl::Recover(
 
   bool is_new_db = false;
   assert(db_lock_ == nullptr);
+  std::vector<std::string> files_in_dbname;
+  // Status files_status;
```
riversand963, line 364: Remove commented code.

db/db_impl/db_impl_open.cc
riversand963: Here,
```
if best_efforts_recovery == true
  files_in_dbname can be used.
else if `dbname_ == wal_dir`
  files_in_wal_dir can be used.
else
  need to call GetChildren(dbname_).
```

riversand963: When comparing path, you may need to call `NormalizePath()`. Otherwise, `/path/to/file` will be treated as different from `/path/to//file`.

db/version_set.cc
```diff
@@ -4743,27 +4745,28 @@ class ManifestPicker {
...
   std::vector<std::string>::const_iterator manifest_file_iter_;
   Status status_;
+
+  void PopulateManifestFiles(std::vector<std::string>& all_files);
 };
 
-ManifestPicker::ManifestPicker(const std::string& dbname, FileSystem* fs)
-    : dbname_(dbname), fs_(fs) {}
+ManifestPicker::ManifestPicker(const std::string& dbname,
+                               std::vector<std::string>& files_in_dbname,
```
riversand963, line 4753: `const std::vector<std::string>& files_in_dbname`.

db/version_set.cc
```diff
@@ -4743,27 +4745,28 @@ class ManifestPicker {
...
-    uint64_t file_num = 0;
-    FileType file_type;
-    bool parse_ok = ParseFileName(fname, &file_num, &file_type);
-    if (parse_ok && file_type == kDescriptorFile) {
-      manifest_files_.push_back(fname);
-    }
+  if (!status_.ok()) {
+    return;
   }
+  PopulateManifestFiles(files_in_dbname);
```
riversand963, line 4768: This is necessary only if `manifest_files_.empty()` in line 4761.

db/version_set.cc
```diff
@@ -4812,12 +4815,25 @@ std::string ManifestPicker::GetNextManifest(uint64_t* number,
   }
   return ret;
 }
+
+void ManifestPicker::PopulateManifestFiles(
+    std::vector<std::string>& all_files) {
```
riversand963, line 4820: Argument can be const ref.

db/db_impl/db_impl_open.cc
```diff
@@ -293,6 +293,10 @@ Status DBImpl::NewDB() {
   if (s.ok()) {
     // Make "CURRENT" file that points to the new manifest file.
     s = SetCurrentFile(fs_.get(), dbname_, 1, directories_.GetDbDir());
+    if (new_filenames) {
+      new_filenames->emplace_back(
+          manifest.substr(manifest.find_last_of("/\\") + 1));
```
riversand963, line 298: Nit: consider using kFilePathSeparator defined in filename.h.

db/db_impl/db_impl_open.cc
```diff
@@ -544,15 +551,16 @@ Status DBImpl::Recover(
     }
 
     std::vector<uint64_t> logs;
-    for (size_t i = 0; i < filenames.size(); i++) {
+    for (size_t i = 0; i < files_in_wal_dir.size(); i++) {
```
riversand963, line 554: Nit: `for (const auto& log_file : files_in_wal_dir)`

db/version_set.cc
```diff
@@ -4729,41 +4729,36 @@ Status VersionSet::Recover(
...
-  Status s = fs_->GetChildren(dbname_, IOOptions(), &children, /*dbg=*/nullptr);
-  if (!s.ok()) {
-    status_ = s;
-    return;
-  }
-  for (const auto& fname : children) {
+ManifestPicker::ManifestPicker(const std::string& dbname,
+                               const std::vector<std::string>& files_in_dbname,
+                               FileSystem* fs)
+    : dbname_(dbname), fs_(fs) {
```
riversand963, line 4750: `fs_` can be removed.

db/version_set.cc
```diff
@@ -4816,10 +4810,15 @@ std::string ManifestPicker::GetNextManifest(uint64_t* number,
...
-  Status s = manifest_picker.status();
+    std::string* db_id, bool* has_missing_table_file,
+    std::vector<std::string>& files_in_dbname) {
+  Status s;
+  if (read_only) {
+    assert(files_in_dbname.empty());
+    s = fs_->GetChildren(dbname_, IOOptions(), &files_in_dbname,
+                         /*dbg=*/nullptr);
+  }
+  ManifestPicker manifest_picker(dbname_, files_in_dbname, fs_);
```
riversand963, line 4821: After second thought, let's move this to line 448 of db_impl_open.cc

db/db_impl/db_impl_open.cc
riversand963: Actually, we can leave it as is for now.

db/version_set.cc
roghnin: I assume it's within the scope of `if(best_efforts_recovery)`? LIne 448 is in an outer scope in the current commit.

db/db_impl/db_impl_open.cc
roghnin: I'm not sure if I need to call it on dbname_. I just called on each of them.

PR #7045, opened 2020-06-29
Fix status message size assert

zhichao-cao:
In status.cc, the assert is `assert(sizeof(msgs) > index)`; msgs is a const char* array, sizeof(msgs) is the array size*char* size, which will make the assert pass all the time. Change it to sizeof(msgs)/sizeof(char*) > index.

test plan: pass make check

### DIFF
--- a/util/status.cc
+++ b/util/status.cc
@@ -134,7 +134,7 @@ std::string Status::ToString() const {
   std::string result(type);
   if (subcode_ != kNone) {
     uint32_t index = static_cast<int32_t>(subcode_);
-    assert(sizeof(msgs) > index);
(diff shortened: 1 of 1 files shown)


util/status.cc
```diff
@@ -134,7 +134,7 @@ std::string Status::ToString() const {
   std::string result(type);
   if (subcode_ != kNone) {
     uint32_t index = static_cast<int32_t>(subcode_);
-    assert(sizeof(msgs) > index);
+    assert(sizeof(msgs) / sizeof(char*) > index);
```
riversand963, line 137: Nit: we can do so without explicitly specifying type
```
assert(sizeof(msgs) / sizeof(msgs[0]) > index);
```

PR #7055, opened 2020-06-30
extend listener callback functions to more file I/O operations

roghnin:
Summary:
Currently, `EventListener` in listner.h only have callback functions for file read and write. One may favor extended callback functions for more file I/O operations like flush, sync and close. This PR tries to add those interface and have them called when appropriate throughout the code base.

Test Plan:
Write an experimental listener with those new callback functions with log output in them; run experiments and check logs to see those functions are actually called.
Default test suits `make check` should also be included.


file/writable_file_writer.cc
```diff
@@ -187,7 +187,20 @@ IOStatus WritableFileWriter::Flush() {
     }
   }
 
-  s = writable_file_->Flush(IOOptions(), nullptr);
+  {
+    IOSTATS_TIMER_GUARD(write_nanos);
```
riversand963, line 191: Are we measuring write here?

file/writable_file_writer.cc
```diff
@@ -187,7 +187,20 @@ IOStatus WritableFileWriter::Flush() {
...
+    TEST_SYNC_POINT("WritableFileWriter::Flush:BeforeAppend");
```
riversand963, line 192: Test sync points must have unique names. This is duplicate of the sync point in line 400 inside `WriteDirect`.

file/writable_file_writer.cc
```diff
@@ -187,7 +187,20 @@ IOStatus WritableFileWriter::Flush() {
...
+    FileOperationInfo::TimePoint start_ts;
+    if (ShouldNotifyListeners()) {
+      start_ts = std::chrono::system_clock::now();
+    }
+    s = writable_file_->Flush(IOOptions(), nullptr);
+    if (ShouldNotifyListeners()) {
+      auto finish_ts = std::chrono::system_clock::now();
+      NotifyOnFileFlushFinish(next_write_offset_, buf_.CurrentSize(), start_ts,
```
riversand963, line 200: `buf_.CurrentSize()` may not be the size of the IO. For `Flush()`, we do not have to report it now if it's hard to get the actual size.

file/writable_file_writer.cc
roghnin: It seems I misunderstood the meaning of "write" here. Considering the entries in IOStatsContext (iostats_context.h), I may need a new entry for flush?

riversand963: We have not been measuring it, so I guess just removing the `IOSTATS` line here is good for this PR.

roghnin: Copy.

file/writable_file_writer.cc
```diff
@@ -275,11 +291,23 @@ IOStatus WritableFileWriter::SyncInternal(bool use_fsync) {
...
   auto prev_perf_level = GetPerfLevel();
   IOSTATS_CPU_TIMER_GUARD(cpu_write_nanos, env_);
+#ifndef ROCKSDB_LITE
+  FileOperationInfo::TimePoint start_ts;
+  if (ShouldNotifyListeners()) {
+    start_ts = std::chrono::system_clock::now();
+  }
+#endif
   if (use_fsync) {
     s = writable_file_->Fsync(IOOptions(), nullptr);
```
roghnin, line 301: Should I treat `Fsync()` and `Sync()` here separately? I didn't because `Sync()` calls `Fsync()` by default underneath.

file/writable_file_writer.cc
```diff
@@ -187,7 +187,22 @@ IOStatus WritableFileWriter::Flush() {
     }
   }
 
-  s = writable_file_->Flush(IOOptions(), nullptr);
+  {
+    TEST_SYNC_POINT("WritableFileWriter::Flush:0");
```
riversand963, line 191: If not used, the sync point should be removed.

file/writable_file_writer.cc
riversand963: Actually it depends on the underlying FileSystem. In `PosixWritableFile`, `Fsync` calls `fsync()` while `Sync` calls `fdata_sync()`, as in env/io_posix.cc.
I am OK with your current approach, maybe you can add a member in `FileInfo` to denote the type of the operation.
I am also OK if you focus on `Flush` in this PR.

include/rocksdb/listener.h
```diff
@@ -459,7 +459,27 @@ class EventListener {
   // operation finishes.
   virtual void OnFileWriteFinish(const FileOperationInfo& /* info */) {}
 
-  // If true, the OnFileReadFinish and OnFileWriteFinish will be called. If
+  // A callback function for RocksDB which will be called whenever a file flush
+  // operation finishes.
+  virtual void onFileFlushFinish(const FileOperationInfo& /* info */) {}
```
riversand963, line 464: To be consistent, the first letter 'o' should be capital.

file/writable_file_writer.cc
```diff
@@ -139,17 +139,59 @@ IOStatus WritableFileWriter::Close() {
   // In direct I/O mode we write whole pages so
   // we need to let the file know where data ends.
   if (use_direct_io()) {
-    interim = writable_file_->Truncate(filesize_, IOOptions(), nullptr);
+    {
+#ifndef ROCKSDB_LITE
+      FileOperationInfo::TimePoint start_ts;
+      if (ShouldNotifyListeners()) {
+        start_ts == std::chrono::system_clock::now();
```
riversand963, line 146: s/==/=/

file/writable_file_writer.cc
```diff
@@ -139,17 +139,59 @@ IOStatus WritableFileWriter::Close() {
...
+      }
+#endif
+    }
     if (interim.ok()) {
-      interim = writable_file_->Fsync(IOOptions(), nullptr);
+      {
+#ifndef ROCKSDB_LITE
+        FileOperationInfo::TimePoint start_ts;
+        if (ShouldNotifyListeners()) {
+          start_ts == std::chrono::system_clock::now();
```
riversand963, line 162: s/==/=/

riversand963:
I think an update in HISTORY.md is needed, and Travis complain about typos.

file/writable_file_writer.cc
roghnin: Oops. Nice catch. Thanks.

HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # Rocksdb Change Log
 ## Unreleased
+### Public API Change
+* File I/O listener is extended to cover more I/O operations. Now class `EventListener` in listener.h contains new callback functions: `OnFileFlushFinish()`, `OnFileSyncFinish()`, `OnFileRangeSyncFinish()`, `OnFileTruncateFinish()`, and ``OnFileCloseFinish()``.
+
```
riversand963, line 5: This can be combined with the `Public API Change` section.

roghnin: Ah I saw there are two of them below, so I thought it was one-header-per-PR basis. Updated.

file/writable_file_writer.cc
roghnin: Is it better to have just one callback function with different types passed in?

db/listener_test.cc
```diff
@@ -1025,11 +1073,23 @@ TEST_F(EventListenerTest, OnFileOperationTest) {
...
 
   Reopen(options);
   ASSERT_GE(listener->file_reads_.load(), listener->file_reads_success_.load());
   ASSERT_GT(listener->file_reads_.load(), 0);
+  ASSERT_GE(listener->file_closes_.load(),
+            listener->file_closes_success_.load());
+  ASSERT_GT(listener->file_closes_.load(), 0);
+  ASSERT_GE(listener->file_syncs_.load(), listener->file_syncs_success_.load());
+  ASSERT_GT(listener->file_syncs_.load(), 0);
+  // TODO: add the following tests only if (WriteableFile::use_direct_io() ==
```
riversand963, line 1089: Try setting `use_direct_io_for_flush_and_compaction` to `true` when opening the db.

roghnin: Worked!

include/rocksdb/listener.h
```diff
@@ -148,10 +148,22 @@ struct TableFileDeletionInfo {
...
+  kSync,
+  kFsync,
+  kRangeSync
+};
+
 struct FileOperationInfo {
   using TimePoint = std::chrono::time_point<std::chrono::system_clock,
                                             std::chrono::nanoseconds>;
 
+  FileOperationType type;
```
riversand963, line 166: Since we are on this. Would it be possible to update the signature of `FileOperationInfo(...)` so that it takes additional arguments `status` and `type`?

roghnin: done.

db/listener_test.cc
```diff
@@ -1087,9 +1086,11 @@ TEST_F(EventListenerTest, OnFileOperationTest) {
   ASSERT_GT(listener->file_closes_.load(), 0);
   ASSERT_GE(listener->file_syncs_.load(), listener->file_syncs_success_.load());
   ASSERT_GT(listener->file_syncs_.load(), 0);
-  ASSERT_GE(listener->file_truncates_.load(),
+  if (true == options.use_direct_io_for_flush_and_compaction){
```
riversand963: Will this ever be true?

riversand963: The default value is `false`, and the case of it being `true` is never tested. Consequently `OnFileTruncateFinish` is not tested. Can we test it?

roghnin: > The default value is `false`, and the case of it being `true` is never tested. Consequently `OnFileTruncateFinish` is not tested. Can we test it?

I'm not sure I can set it to `true` without having segfault in the test in Phabricator. Maybe I'll try testing this function without direct I/O.

mrambacher:
The OnFileOperationTest fails regularly on MacOS, as described in #7133 .

PR #7060, opened 2020-07-01
riscv64 need -latomic as library

### LINKED ISSUE #7051: RocksDB build fails on riscv64
### Expected behavior

Build is OK

### Actual behavior

Build fail with:
```
/usr/bin/ld: librocksdblib.a(memtable.cc.o): in function `.L0 ':
./builddir/storage/rocksdb/./storage/rocksdb/rocksdb/util/dynamic_bloom.h:177:
undefined reference to `__atomic_fetch_or_1'
/usr/bin/ld: librocksdblib.a(memtable.cc.o): in function `.LVL1731':
./builddir/storage/rocksdb/./storage/rocksdb/rocksdb/util/dynamic_bloom.h:179:
undefined reference to `__atomic_fetch_or_1'
/usr/bin/ld: librocksdblib.a(memtable.cc.o): in function `SaveValue':
./builddir/storage/rocksdb/./storage/rocksdb/rocksdb/db/memtable.cc:596:
undefined reference to `__atomic_compare_exchange_1'
/usr/bin/ld: librocksdblib.a(memtable.cc.o): in function `.L0 ':
/usr/include/c++/8/bits/atomic_base.h:434: undefined reference to
`__atomic_compare_exchange_1'
/usr/bin/ld: /usr/include/c++/8/bits/atomic_base.h:434: undefined
reference to `__atomic_compare_exchange_1'
/usr/bin/ld: /usr/include/c++/8/bits/atomic_base.h:434: undefined
reference to `__atomic_compare_exchange_1'
/usr/bin/ld: /usr/include/c++/8/bits/atomic_base.h:434: undefined
reference to `__atomic_compare_exchange_1'
/usr/bin/ld: librocksdblib.a(memtable.cc.o):/usr/include/c++/8/bits/atomic_base.h:434:
more undefined references to `__atomic_compare_exchange_1' follow
collect2: error: ld returned 1 exit status
```
See full log at: https://buildd.debian.org/status/fetch.php?pkg=mariadb-10.3&arch=riscv64&ver=1%3A10.3.16-1&stamp=1561225015&raw=0

Build RocksDB on riscv64 arch.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=933151
https://jira.mariadb.org/browse/MDEV-23051

grooverdan:
note may not be the best way to add a library, and maybe it needs to be more general than just Linux.

riscv64 architecture specificly chosen to include atomics per https://www.sifive.com/blog/all-aboard-part-1-compiler-args

Compile error per bug reports:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=933151
https://jira.mariadb.org/browse/MDEV-23051

Per Debian bug -latomic is needed which is consistent with
build_detect_platform code which include atomic (though only for
clang).

Tested with https://toolchains.bootlin.com/releases_riscv64.html
( riscv64--glibc--bleeding-edge-2020.02-2)

cmake -DCMAKE_TOOLCHAIN_FILE=../rocks-riscv64-toolchain.cmake -DPORTABLE=TRUE ../rocksdb

rocks-riscv64-toolchain.cmake:
set(CMAKE_SYSTEM_NAME Linux)
set(CMAKE_SYSTEM_PROCESSOR riscv64)

set(tools /home/dan/repos/riscv64--glibc--bleeding-edge-2020.02-2)
set(CMAKE_C_COMPILER ${tools}/bin/riscv64-buildroot-linux-gnu-gcc)
set(CMAKE_CXX_COMPILER ${tools}/bin/riscv64-buildroot-linux-gnu-g++)

set(CMAKE_C_FLAGS "-march=rv64iafd")
set(CMAKE_CXX_FLAGS "-march=rv64iafd")
set(WITH_GFLAGS OFF)

set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)

closes #7051

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,6 +252,11 @@ if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
   endif(HAS_ARMV8_CRC)
 endif(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64|AARCH64")
 
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL "riscv64")
+  string(APPEND CMAKE_C_STANDARD_LIBRARIES " -latomic")
+  string(APPEND CMAKE_CXX_STANDARD_LIBRARIES " -latomic")
+endif()
+
 option(PORTABLE "build a portable binary" OFF)
 option(FORCE_SSE42 "force building with SSE4.2, even when PORTABLE=ON" OFF)
 if(PORTABLE)
--- a/Makefile
+++ b/Makefile
@@ -134,6 +134,10 @@ OPT += -momit-leaf-frame-pointer
 endif
 endif
 
+ifneq ($(findstring riscv64, $(MACHINE)),)
+LDFLAGS += -latomic
+endif
+
 ifeq (,$(shell $(CXX) -fsyntax-only -maltivec -xc /dev/null 2>&1))
 CXXFLAGS += -DHAS_ALTIVEC
 CFLAGS += -DHAS_ALTIVEC


xnox:
It is odd that with all projects that use CMake this is needed. Given that toolchain, specifically gcc, has spec file to add `-latomic` on riscv64. Does CMake not use `gcc` to call linker?! I wonder if this should be fixed in CMake upstream, instead of forcing every project add this to support riscv64.

See https://github.com/riscv/riscv-gcc/commit/2c4857d0981501b7c50bbf228de9e287611f8ae5

xnox:
Opened https://gitlab.kitware.com/cmake/cmake/-/issues/20895

But yes, please merge this pull request. It is a correct thing to do, until CMake is fixed up.

cpaelzer:
Just FYI @gaearon I have put this patch on top of mariadb-10.3 and still run into the same [build issues](https://launchpad.net/~ci-train-ppa-service/+archive/ubuntu/4128/+build/19532693).
Here is the link to the[ log file](https://launchpadlibrarian.net/486689974/buildlog_ubuntu-groovy-riscv64.mariadb-10.3_1%3A10.3.22-1ubuntu1~ppa1_BUILDING.txt.gz).

To me it still seems to be the same issue:

```
...
cd /<<PKGBUILDDIR>>/builddir/storage/rocksdb && /usr/bin/cmake -E cmake_link_script CMakeFiles/rocksdb_aux_lib.dir/link.txt --verbose=1
/usr/bin/ar qc librocksdb_aux_lib.a  CMakeFiles/rocksdb_aux_lib.dir/rdb_cf_options.cc.o CMakeFiles/rocksdb_aux_lib.dir/event_listener.cc.o CMakeFiles/rocksdb_aux_lib.dir/rdb_perf_context.cc.
/usr/bin/ld: librocksdblib.a(memtable.cc.o): in function `rocksdb::SpinMutex::try_lock()':
./builddir/storage/rocksdb/./storage/rocksdb/rocksdb/util/mutexlock.h:106: undefined reference to `__atomic_compare_exchange_1'
/usr/bin/ld: librocksdblib.a(memtable.cc.o): in function `void rocksdb::DynamicBloom::AddHash<rocksdb::DynamicBloom::AddHashConcurrently(unsigned int)::{lambda(std::atomic<unsigned char>*, u
./builddir/storage/rocksdb/./storage/rocksdb/rocksdb/util/dynamic_bloom.h:181: undefined reference to `__atomic_fetch_or_1'
/usr/bin/ld: ./builddir/storage/rocksdb/./storage/rocksdb/rocksdb/util/dynamic_bloom.h:182: undefined reference to `__atomic_fetch_or_1'
/usr/bin/ld: librocksdblib.a(memtable.cc.o): in function `rocksdb::MemTable::Add(unsigned long, rocksdb::ValueType, rocksdb::Slice const&, rocksdb::Slice const&, bool, rocksdb::MemTablePostP
./builddir/storage/rocksdb/./storage/rocksdb/rocksdb/db/memtable.cc:583: undefined reference to `__atomic_fetch_or_1'
/usr/bin/ld: ./builddir/storage/rocksdb/./storage/rocksdb/rocksdb/db/memtable.cc:583: undefined reference to `__atomic_fetch_or_1'
/usr/bin/ld: librocksdblib.a(memtable.cc.o): in function `std::__atomic_base<bool>::compare_exchange_weak(bool&, bool, std::memory_order, std::memory_order)':
/usr/include/c++/9/bits/atomic_base.h:457: undefined reference to `__atomic_compare_exchange_1'
/usr/bin/ld: /usr/include/c++/9/bits/atomic_base.h:457: undefined reference to `__atomic_compare_exchange_1'
/usr/bin/ld: /usr/include/c++/9/bits/atomic_base.h:457: undefined reference to `__atomic_compare_exchange_1'
/usr/bin/ld: /usr/include/c++/9/bits/atomic_base.h:457: undefined reference to `__atomic_compare_exchange_1'
/usr/bin/ld: /usr/include/c++/9/bits/atomic_base.h:457: undefined reference to `__atomic_compare_exchange_1'
/usr/bin/ld: librocksdblib.a(memtable.cc.o):/usr/include/c++/9/bits/atomic_base.h:457: more undefined references to `__atomic_compare_exchange_1' follow 
collect2: error: ld returned 1 exit status 
```

gaearon:
Did you mean to mention someone else? I’m not familiar with this project.

xnox:
I am taking alternative approach of trying to fix default CMake toolchain / platofrm files for Linux, to pass `-latomic` on RISCV. This is not without a precedent, this is also done for some Android NDK levels by CMake.

Net result, is that I expect for rocksdb, opendht to "just build" unmodified on RISCV. Afterall, that's how it should be, when one is using stock C atomics.h which compiler / build-system are supposed to "just provide" without doing anything on any architecture.

Patch to cmake being tested in Groovy is 

```
Description: Append -latomic on RISCV, as some atomics need libatomic support.
Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/cmake/+bug/1876814
Author: Dimitri John Ledkov <xnox@ubuntu.com>


Index: cmake-3.16.3/Modules/Platform/Linux.cmake
===================================================================
--- cmake-3.16.3.orig/Modules/Platform/Linux.cmake
+++ cmake-3.16.3/Modules/Platform/Linux.cmake
@@ -57,3 +57,15 @@ if(NOT CMAKE_CROSSCOMPILING AND EXISTS "
   set_property(GLOBAL PROPERTY FIND_LIBRARY_USE_LIB32_PATHS FALSE)
   set_property(GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS FALSE)
 endif()
+
+# RISCV needs -latomic
+# https://gitlab.kitware.com/cmake/cmake/-/issues/20895
+#
+# Ideally I'd use CMAKE_C_COMPILER_ARCHITECTURE_ID, but that is only
+# available on Win platforms. Not sure why though, cause every
+# compiler on Linux has a default architecture it compiles for too,
+# when no flags are passed to ask the compiler to behave differently.
+if(CMAKE_SYSTEM_PROCESSOR STREQUAL "riscv64")
+  string(APPEND CMAKE_C_STANDARD_LIBRARIES " -latomic")
+  string(APPEND CMAKE_CXX_STANDARD_LIBRARIES " -latomic")
+endif()
```

grooverdan:
redid this with CMAKE_C{,XX}_STANDARD_LIBRARIES per upstream.

MariaDB's build failure is because they need the same fix at a higher level too. They are linking their rdb_perf_context.cc against the static rocksdb library. This evidently doesn't link the libraries into the static library. Or it could possible because I didn't include 'CMAKE_STATIC_LINKER_FLAGS' before. The STANDARD_LIBRARIES seems like a better approach, but since mariadb use atomics elsewhere in they code, they'll need the same fix outside of rocksdb.

cpaelzer:
> Did you mean to mention someone else? I’m not familiar with this project.

Yeah sorry @gaearon I meant to ping @grooverdan

adamretter:
@grooverdan Is this only required for CMake, or so we also need to add something similar to the Make based build?

grooverdan:
Sorry took me a while to get back to this and work out the test incarnation to make a cross compile with Make. I got a libnuma fail because it wasn't in the chroot however the latomic got added per command line in commit. I've also assumed that on native riscv64 uname -m will contain riscv64. If for some reason this is incorrectly please accept apologies and show what uname -m returns.

mrambacher:
@grooverdan I recently proposed #8183 that should address this problem.  Can you confirm if that fix would also address this issue?  I do not have access to a riscv64 machine...

grooverdan:
Hey @mrambacher, I don't have access to hardware either hence the cross compile attempt the beginning of this which I think you should be able to use as a test.

Natively there might be a lauchpad way - https://launchpad.net/ubuntu/+source/rocksdb . Alternately you might be able to use https://packages.debian.org/source/buster/rocksdb to create your own https://salsa.debian.org/public build.

adamretter:
@mrambacher @grooverdan if we don't have access to hardware and this doesn't break anything, then I would say we go with it :-)

grooverdan:
what could possibly go wrong(TM) ? :smile:

adamretter:
@mrambacher Are you happy to merge this one?

mrambacher:
This should be fixed by #8183.  We can re-open and re-investigate it if it still happens.

PR #7063, opened 2020-07-01
Fix db_id and db_session_id nullptr warning by clang analyzer

gg814:
Summary: 
GetFileDbIdentities requires either db_id non-null or db_session_id non-null.
Passing nullptr for db_id or db_session_id in CopyOrCreateFile indicates the caller does not want to obtain the value for db_id or db_session_id.

Test plan:
USE_CLANG=1 make analyze
backupable_db_test

### DIFF
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -1469,7 +1469,7 @@ Status BackupEngineImpl::CopyOrCreateFile(
     // the restored file
     if (!src.empty()) {
       // copying
-      if (IsSstFile(src) && db_session_id != nullptr) {
+      if (IsSstFile(src) && (db_id != nullptr || db_session_id != nullptr)) {
         // SST file
         // Ignore the returned status
         // In the failed cases, db_id and db_session_id will be empty
@@ -1688,6 +1688,7 @@ Status BackupEngineImpl::GetFileDbIdentities(Env* src_env,
                                              const std::string& file_path,
                                              std::string* db_id,
                                              std::string* db_session_id) {
+  assert(db_id != nullptr || db_session_id != nullptr);
   // // Prepare the full_path of file_path under src_env for SstFileDumper
   std::string full_path;
   src_env->GetAbsolutePath(file_path, &full_path);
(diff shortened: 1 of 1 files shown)


utilities/backupable/backupable_db.cc
```diff
@@ -1716,8 +1716,12 @@ Status BackupEngineImpl::GetFileDbIdentities(Env* src_env,
   }
 
   if (table_properties != nullptr) {
-    db_id->assign(table_properties->db_id);
+    assert(db_session_id != nullptr);
```
pdillinger, line 1719: Function preconditions are best checked at the start of the function body, especially for readability.

If we are "future proofing" this code, I think a better precondition is `db_id != null || db_session_id != null`, which could also be checked in CopyOrCreateFile before calling this.

If we are not future proofing, then both functions can expect non-null for both parameters, simplifying the code.

utilities/backupable/backupable_db.cc
```diff
@@ -1373,6 +1373,7 @@ Status BackupEngineImpl::CopyOrCreateFile(
     uint64_t size_limit, std::function<void()> progress_callback,
     std::string* db_id, std::string* db_session_id) {
   assert(src.empty() != contents.empty());
+  assert(db_id != nullptr || db_session_id != nullptr);
```
pdillinger, line 1376: Sometimes we have explicit behaviors for release builds for cases in which an assertion would fail in a debug build, but I don't think that's needed here.

(a) This assertions says "a precondition of this function is that at least one is not null".

(b) The declaration of this function has default parameters with both null, so says "it's ok to pass both null."

(c) The new 'if' condition below implies the condition needs to be checked, so must not be a precondition of calling this function.

Which way do you want to fix so that everything agrees?

gg814: I'd choose (b). When null is passed in for db_id  or db_session in CopyOrCreateFile, it indicates the caller does not want to obtain the value of db_id or db_session_id. 

GetFileDbIdentities can be called whenever db_id or db_session_id is not null.

I will the remove the assertions. Before calling GetFileDbIdentities in CopyOrCreateFile, the condition that checks if one of the two ids is not null will be kept.

What do you think?

pdillinger: I meant (a), (b), and (c) as pieces of evidence, not options for fixing. ;)

I support removing this assertion but keeping the one in GetFileDbIdentities, because that function doesn't make sense if both are null.

PR #7069, opened 2020-07-02
add property handle for getting latest sequence number at specified level

DorianZheng:
@yiwu-arbug PTAL


DorianZheng:
Sorry, I will format the code and add some test case tomorrow.

PR #7070, opened 2020-07-02
Revert "Whole DBTest to skip fsync (#7049)"

pdillinger:
This reverts commit 4f1534bdb028b5e37f2f282433906ae493c54163.

This commit caused failures and deadlocks in
MultiThreadedDBTest.MultiThreaded/69 and others.


pdillinger:
Format check fails because this is a pure revert.

pdillinger:
Even though the failure is intermittent, I got a very clear signal on bisection.

PR #7074, opened 2020-07-03
Revert "Directly use unit test tempalte buck (#6926)"

ajkr:
This reverts commit 2e7070b1947fc36a1fd9922f558f8a8eeaba899b. It is
necessary because `cpp_unittest` is not compatible with "db/c_test.c".

Test Plan: built/ran test with buck.

### DIFF
--- a/TARGETS
+++ b/TARGETS
@@ -78,6 +78,7 @@ ROCKSDB_PREPROCESSOR_FLAGS = [
     # Directories with files for #include
     "-I" + REPO_PATH + "include/",
     "-I" + REPO_PATH,
+    "-I" + REPO_PATH + "third-party/gtest-1.8.1/fused-src/",
 ]
 
 ROCKSDB_ARCH_PREPROCESSOR_FLAGS = {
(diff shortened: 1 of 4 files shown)


ajkr:
I'm still unsure about revert vs. introduce `cpp_binary` and `custom_unittest` targets for "db/c_test.c". I guess this isn't worthy of a weekend patch release so we can discuss later.

ajkr:
> I'm still unsure about revert vs. introduce `cpp_binary` and `custom_unittest` targets for "db/c_test.c". I guess this isn't worthy of a weekend patch release so we can discuss later.

Discussed with @siying offline. Will change to the other approach (introduce `cpp_binary` and `custom_unittest` targets for "db/c_test.c").

ajkr:
> > I'm still unsure about revert vs. introduce `cpp_binary` and `custom_unittest` targets for "db/c_test.c". I guess this isn't worthy of a weekend patch release so we can discuss later.
> 
> Discussed with @siying offline. Will change to the other approach (introduce `cpp_binary` and `custom_unittest` targets for "db/c_test.c").

Fixed by e0d0b49577fc221f5027e287dc444098c99c93e6.

PR #7080, opened 2020-07-03
Fixed Factory construct just for calling .Name()

rockeet:


options/options_helper.cc
```diff
@@ -939,7 +939,7 @@ Status GetTableFactoryFromMap(
     }
     table_factory->reset(new BlockBasedTableFactory(bbt_opt));
     return s;
-  } else if (factory_name == PlainTableFactory().Name()) {
+  } else if (factory_name == "PlainTable") {
```
ajkr, line 942: Can we have a named constant here too?

rockeet: > Can we have a named constant here too?

No, there is no such a constant defined for PlainTable.

ajkr: How about define such a constant for "PlainTable" first, then?

rockeet: > How about define such a constant for "PlainTable" first, then?

OK, done!

PR #7081, opened 2020-07-03
Add enum reflection support and reflect for some enum define

rockeet:
## With enum reflection, we can convert enum to/from string
For example:
```c++
ROCKSDB_ENUM_PLAIN(CompactionStyle, char,
  kCompactionStyleLevel = 0x0,
  kCompactionStyleUniversal = 0x1,
  kCompactionStyleFIFO = 0x2,
  kCompactionStyleNone = 0x3 // comma(,) can not be present here
);
assert(enum_name(kCompactionStyleUniversal) == "kCompactionStyleUniversal");
assert(enum_name(CompactionStyle(100)).size() == 0);
CompactionStyle cs= kCompactionStyleLevel;
assert(enum_value("kCompactionStyleUniversal", &cs) && cs == kCompactionStyleUniversal);
assert(!enum_value("bad", &cs) && cs == kCompactionStyleUniversal); // cs is not changed
```

## There are 4 macros to define a enum with reflection
```c++
// plain old enum defined in a namespace(not in a class/struct)
ROCKSDB_ENUM_PLAIN(EnumType, IntRep, e1 = 1, e2 = 2);
// this generates:
enum EnumType : IntRep { e1 = 1, e2 = 2 };
// enum reflection supporting code ...
// ...
// the supporting code makes template function enum_name and
// enum_value works for this EnumType

// other three macros are similar with some difference:

// enum class defined in a namespace(not in a class/struct)
ROCKSDB_ENUM_CLASS(EnumType, IntRep, Enum values ...);

// plain old enum defined in a class/struct(not in a namespace)
ROCKSDB_ENUM_PLAIN_INCLASS(EnumType, IntRep, Enum values ...);

// enum class defined in a class/struct (not in a a namespace)
ROCKSDB_ENUM_CLASS_INCLASS(EnumType, IntRep, Enum values ...);

```


rockeet:
I have added /Zc::preprocessor to fix `build windows fail`.

rockeet:
> Can you please provide more comments on the utility of this feature and why it is important? I thought for awhile about doing something similar during work as part of #6423 but had trouble justifying it. The string version of the enums generated here are not the ones used by the Options utilities, which makes it harder for me to figure out when and where this feature will be used.

Enum reflection does not `repeat the code`, with `ROCKSDB_ENUM_XXX`:
1. We do not need to manually define and implement enum to/from string functions.
2. If we change enum definition(add/remove enum values), we just need to change one place, not three or more places.

I have read #6423 as you mentioned, it is a shinning example of why enum reflection is useful.

In #6423, the enum's string form is the same as the enum's name, the leading `'k'` of enum names are preserved. I don't know why **"The string version of the enums generated here are not the ones used by the Options utilities"**, and event the leading `'k'` should be deleted from the string form, it is just a trivial work to write another `enum_name/enum_value` function.

rockeet:
CircleCI: build-windows: the visual c++ version is 14.22.27905 too old, it did not implement standard conform preprocessor.

The min version which implemented standard conform preprocessor is `Visual Studio 2019 version 16.6`, which internal version number(C macro `_MSC_VER` or CMake `MSVC_VERSION` is 1926).

@mrambacher @siying could you upgrade or use newer Visual Studio 2019 version in CircleCI?

adamretter:
@rockeet I think we need to stay compatible with VS2015

PR #7085, opened 2020-07-06
BackupEngine supports custom file checksums

gg814:
Summary:
A new option `std::shared_ptr<FileChecksumGenFactory> backup_checksum_gen_factory` is added to `BackupableDBOptions`. This allows custom checksum functions to be used for creating, verifying, or restoring backups.

Tests are added.

Test plan:
Passed make check


pdillinger:
I have thought some more about this approach. Schema changes are scary, so what about this approach:

Setting backup_checksum_gen_factory (maybe rename to file_checksum_gen_factory) in BackupableDBOptions means that we essentially treat the backed-up DB manifest as an extension of the backup manifest, and use both crc32(c) checksums and the custom checksums to validate backups.

    Event          | crc32c action | custom checksum action
    -------------------------------------------------------
    Create backup  | compute       | validate
    Verify backup  | validate      | validate
    Restore backup | validate      | validate

This does not enable performance improvements like atomic copy / hard link for creating backups, but that might be rarely used anyway so that we can verify db manifest checksums on copy.

Thoughts?

gg814:
Sounds good. In your proposed approach, if I understand correctly, you advocate users to adopt the same custom checksum function used in DB for backup engine. In this way, we have a meaningful extra guard of data integrity (crc32c and custom checksum method in DB).

To extend this a little bit, we could always compute the default crc32c checksum for backup. Moreover, if there is custom checksum function (different from the one used in DB), we would also use it to compute another checksum. Both checksums (crc32c and backup engine custom checksum) will be used for validation. This also provides an extra protection for data corruption. But an obvious drawback is the additional computation costs of the custom checksum.

pdillinger:
If we are actually copying files, rather than taking advantage of an atomic copy or hard link, I think CPU is relatively cheap here compared to I/O. Backups will often be on a remote filesystem, for example.

While the independent crc32c checksum does in theory improve data integrity, in my mind the primary motivation is forward compatibility (old RocksDB version restoring backups taken with a newer version). Besides, the weakest link has the biggest impact on overall data integrity. (Related to the cache-friendly Bloom filter, https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter#the-math)

> Moreover, if there is custom checksum function (different from the one used in DB), we would also use it to compute another checksum.

The problem is that (I believe) we would need a backup schema change (break forward compatibility) to store checksum information separate from what's already in the backup manifest schema and from the DB manifest file being backed up. Even if we added the information in an extra file, IIRC older version would see that as a "garbage" file and delete it. That's actually not terrible as it's like an automatic "downgrade", but if we go that far, we should probably do the schema change to reap the benefits.

I expect most users will want to use the same file checksum for backups as for DB, so I wouldn't expect it to be a big limitation. If Zhichao or anyone believes differently, speak up. :)

gg814:
> The problem is that (I believe) we would need a backup schema change (break forward compatibility) to store checksum information separate from what's already in the backup manifest schema and from the DB manifest file being backed up. 

I think we could store the custom checksum information together with the default crc32c information in the backup meta/manifest.

Currently, it is stored as
`<file1> <crc32(literal string)> <crc32c_value>`
we could extend it to
`<file1> <crc32(literal string)> <crc32c_value> <custom(literal string)> <checksum_name>:<checksum_value>`

The issue for this that I can think of now is that for forward compatibility, the file checksum info string above can be rather long with many custom checksum names and values that are no longer used.

pdillinger:
Extra info on the line would be reported as corruption by an older version: https://github.com/facebook/rocksdb/blob/master/utilities/backupable/backupable_db.cc#L2135

But actually, the error message for using different/unknown checksum is better than I remember: https://github.com/facebook/rocksdb/blob/master/utilities/backupable/backupable_db.cc#L2144

That makes me more comfortable with this change as it is (broken forward compatibility on custom checksum; use custom checksum instead of crc32c) because attempting to restore backup with an older version will at least give a somewhat useful message "Unknown checksum type ..."

I could foresee someone needing a tool to "downgrade" a backup (manifest) though. Since ldb has backup and restore options already, this seems like a logical place. And actually, can ldb even operate on backups using custom checksum yet? (without a custom build of ldb, would need dynamic linking, right?) CC @mrambacher

gg814:
> Extra info on the line would be reported as corruption by an older version: https://github.com/facebook/rocksdb/blob/master/utilities/backupable/backupable_db.cc#L2135

If one opens a new version meta file with an old version backup engine then yes. With a new backup engine, it should be ok. In this PR:

HISTORY.md
```diff
@@ -22,6 +22,7 @@
...
+* A new option `std::shared_ptr<FileChecksumGenFactory> backup_checksum_gen_factory` is added to `BackupableDBOptions`. The default value for this option is `nullptr`. If this option is null, the default backup engine checksum function (crc32c) will be used for creating, verifying, or restoring backups. If it is not null, a custom checksum function will be used. When creating a new backup with non-null `backup_checksum_gen_factory`, the checksum function to be used is the one returned by `backup_checksum_gen_factory->CreateFileChecksumGenerator(FileChecksumGenContext)` with `FileChecksumGenContext::requested_function_name` being empty. When verifying or restoring a backup that was created using a custom checksum function, `Status::NotSupported` will be returned if no checksum factory is provided, and `Status::NotFound` will be returned if the name of the custom checksum function cannot be found in the checksum factory passed in by `backup_checksum_gen_factory`.
```
zhichao-cao, line 25: Can you make this sentence more clear "When creating a new backup with non-null `backup_checksum_gen_factory`, the checksum function to be used is the one returned by `backup_checksum_gen_factory->CreateFileChecksumGenerator(FileChecksumGenContext)` with `FileChecksumGenContext::requested_function_name` being empty." It is hard to follow.

include/rocksdb/file_checksum.h
```diff
@@ -25,6 +25,8 @@ constexpr char kUnknownFileChecksumFuncName[] = "Unknown";
 
 struct FileChecksumGenContext {
   std::string file_name;
+  // The name of the requested checksum generator
+  std::string requested_function_name;
```
zhichao-cao, line 29: Would be better to use: requested_checksum_func_name.

utilities/backupable/backupable_db.cc
```diff
@@ -354,14 +411,15 @@ class BackupEngineImpl : public BackupEngine {
...
+      uint32_t* checksum_value = nullptr, uint64_t size_limit = 0,
       std::function<void()> progress_callback = []() {},
       std::string* db_id = nullptr, std::string* db_session_id = nullptr);
 
-  Status CalculateChecksum(const std::string& src, Env* src_env,
-                           const EnvOptions& src_env_options,
-                           uint64_t size_limit, uint32_t* checksum_value);
+  Status CalculateChecksum(
+      const std::string& src, Env* src_env, const EnvOptions& src_env_options,
+      uint64_t size_limit, uint32_t* checksum_value,
```
zhichao-cao, line 421: I have one question here. If the FileChecksumGenerator provided by user does not provide the uint32_t type checksum results (e.g., SHA1 or MD5), is there any issues here?

gg814: Good question, I think that would be problematic. It seems better to use `string` for storing any checksum result (default or custom). In that case, one more thing to worry about is that the custom checksum result may contain non-printable content or characters not allowed to appear in the filename.

gg814: If we decide to always use the backup engine default checksum and treat the custom checksum as an extra protection, we may simply use the default checksum in the filename so that we do not need to worry about non-printable content or disallowed characters. (perhaps still need to take care of the case that some raw bytes may be converted to newline characters when storing checksum results in backup meta files?)

pdillinger:
> That makes me more comfortable with this change as it is (broken forward compatibility on custom checksum; use custom checksum instead of crc32c)

Actually, I think that the "emergency" ability to restore from backup with ldb, with no custom code in ldb, is important. It doesn't scare *me* too much if this restore skips checksum verification, because checksums should be verified at read time in the DB, but I suspect that users (esp users inside FB) will not want to be cornered into skipping checksum verification in any step. (I think they'll be OK with skipping a second layer of checksum verification in a step--we don't even have to make that obvious to the user--but a copying step with no checksum verification is scary for users.)

So that is more weight in favor of keeping the existing schema (backup manifest always includes crc32c checksums only) and extend the checking/verification when a custom checksum is configured in BackupEngine and DB (checksums in DB manifest).

gg814:
[03152e0](https://github.com/facebook/rocksdb/pull/7085/commits/03152e0d8749c26b29ddc45b1c34082feda6fa91) uses the backuped db MANIFEST as an extension of the backup meta. Specifically, this means when creating, verifying, or restoring backups, in addition to its default crc32c, backup engine will try to use the custom checksums stored in the backuped db MANIFEST, and it will use them if the corresponding checksum functions are available in the custom checksum factory passed in to backup engine.

This additional checksum computation does not incur additional I/O but additional CPU usage.

If custom checksums are available, the checksum value in the table filenames will be the custom checksum value.

The checksum values will be changed to hex encoded later as in #7110.

Something to be decided:
* > if there is custom checksum function (different from the one used in DB), we would also use it to compute another checksum.
* if `verify_with_checksum` is `true`, `VerifyBackup` only compares crc32c but not the custom checksums? (and add another option for verifying with custom checksums?)

HISTORY.md
```diff
@@ -22,6 +22,7 @@
...
+* A new option `std::shared_ptr<FileChecksumGenFactory> file_checksum_gen_factory` is added to `BackupableDBOptions`. The default value for this option is `nullptr`. If this option is null, the default backup engine checksum function (crc32c) will be used for creating, verifying, or restoring backups. If it is not null and is set to the DB custom checksum factory, the custom checksum function used in DB will also be used for creating, verifying, or restoring backups, in addition to the default checksum function (crc32c). If it is not null and is set to a custom checksum factory different than the DB custom checksum factory (which may be null), BackupEngine will still use the default checksum function (crc32c).
```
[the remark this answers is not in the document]
gg814: Good point! I had in mind the case where backup and db might have different custom checksum factories so didn't think of the possibility of reducing the factory option to simply a flag. But in fact, only `CreateNewBackup` has access to the db pointer (and thus `DBOptions`). `VerifyBackup` and `RestoreDBFromBackup` do not have access to `DBOptions`.

gg814:
Rebase and some small changes (use hex-encoded checksums, update comments, refactor helper functions, etc). Should be ready for another look :)

zhichao-cao:
> Rebase and some small changes (use hex-encoded checksums, update comments, refactor helper functions, etc). Should be ready for another look :)

Thanks for the updates and dressing the comments, I will take a look today.

include/rocksdb/utilities/backupable_db.h
```diff
@@ -165,6 +165,33 @@ struct BackupableDBOptions {
   // db_session_id as a fallback.
   BackupTableNameOption share_files_with_checksum_naming;
 
+  // Option for custom checksum functions.
```
zhichao-cao, line 168: Nit: "customized" instead of "custom"?

include/rocksdb/utilities/backupable_db.h
```diff
@@ -165,6 +165,33 @@ struct BackupableDBOptions {
...
+  // the checksum function.
+  //
+  // When it is not nullptr, this option comes into effect only if DB has a
+  // custom checksum factory and this option is set to the same factory.
+  //
+  // Regardless of what is set for this option, BackupEngine will always use
+  // its default crc32c checksum for creating, verifying, or restoring backups.
+  //
+  // If this option is set to the DB checksum factory, BackupEngine will try
+  // to find the requested checksum function and use it to create, verify, or
```
zhichao-cao, line 179: Better to explain what is the "requested checksum function" here. It is unclear what's the relationship between this option, DB checksum factory and requested checksum function.

include/rocksdb/utilities/backupable_db.h
```diff
@@ -25,9 +25,9 @@
 namespace ROCKSDB_NAMESPACE {
 
 // The default DB file checksum function name.
-constexpr char kDbFileChecksumFuncName[] = "FileChecksumCrc32c";
+constexpr char kDefaultDbFileChecksumFuncName[] = "FileChecksumCrc32c";
```
zhichao-cao, line 28: Would be better to move the `kDefaultDBFileChecksumFuncName` to include/rocskdb/file_checksum.h.

utilities/backupable/backupable_db.cc
```diff
@@ -60,6 +61,17 @@ inline std::string ChecksumInt32ToHex(const uint32_t& checksum_value) {
   PutFixed32(&checksum_str, EndianSwapValue(checksum_value));
   return ChecksumStrToHex(checksum_str);
 }
+inline bool IsSameChecksumFunc(const std::string& dst_checksum_func_name,
```
zhichao-cao, line 64: Add a comment here to explain the equal condition.

utilities/backupable/backupable_db.cc
```diff
@@ -181,6 +197,8 @@ class BackupEngineImpl : public BackupEngine {
     const std::string filename;
     const uint64_t size;
     const std::string checksum_hex;
+    const std::string custom_checksum_hex;
+    const std::string checksum_func_name;
```
zhichao-cao, line 201: Do you think we need to use custom_checksum_func_name?

utilities/backupable/backupable_db.cc
```diff
@@ -358,6 +376,53 @@ class BackupEngineImpl : public BackupEngine {
     return GetBackupMetaDir() + "/" + (tmp ? "." : "") +
            ROCKSDB_NAMESPACE::ToString(backup_id) + (tmp ? ".tmp" : "");
   }
+  inline bool HasCustomChecksumFunc() const {
```
zhichao-cao, line 379: Better name to `HasCustomChecksumGenFactory`

utilities/backupable/backupable_db.cc
```diff
@@ -358,6 +376,53 @@ class BackupEngineImpl : public BackupEngine {
...
+    return options_.file_checksum_gen_factory != nullptr;
+  }
+  // Returns nullptr if file_checksum_gen_factory is not set or
+  // file_checksum_gen_factory is not able to create a generator with
+  // name being requested_checksum_func_name
+  inline std::unique_ptr<FileChecksumGenerator> GetCustomChecksumFunc(
```
zhichao-cao, line 385: Better name to `GetCustomChecksumGenerator`

utilities/backupable/backupable_db.cc
```diff
@@ -358,6 +376,53 @@ class BackupEngineImpl : public BackupEngine {
...
+    if (checksum_factory == nullptr) {
+      return nullptr;
+    } else {
+      FileChecksumGenContext gen_context;
+      gen_context.requested_checksum_func_name = requested_checksum_func_name;
+      return checksum_factory->CreateFileChecksumGenerator(gen_context);
+    }
+  }
+  // Set the checksum generator by the requested checksum function name
+  inline Status SetChecksumFunc(
```
zhichao-cao, line 398: `SetchecksumGenerator`

utilities/backupable/backupable_db.cc
```diff
@@ -1384,7 +1525,11 @@ Status BackupEngineImpl::RestoreDBFromBackup(const RestoreOptions& options,
       s = item_status;
       break;
     } else if (item.checksum_hex != result.checksum_hex) {
-      s = Status::Corruption("Checksum check failed");
+      // Compare crc32c checksums (especially for non-table files)
+      std::string checksum_info("Expected checksum is " + item.checksum_hex +
```
zhichao-cao, line 1529: Better to point out it is the default crc32c checksum.

utilities/backupable/backupable_db.cc
```diff
@@ -1441,27 +1596,74 @@ Status BackupEngineImpl::VerifyBackup(BackupID backup_id,
...
       // verify file checksum
-      std::string checksum_hex;
+      // try setting checksum_func
+      std::unique_ptr<FileChecksumGenerator> checksum_func;
+      std::string src_checksum_func_name = kUnknownFileChecksumFuncName;
+      std::string src_checksum_str = kUnknownFileChecksum;
+      std::string src_checksum_hex;
+      if (IsSstFile(file_info->filename)) {
+        // Extract table file number
+        const std::string& file = file_info->filename;
```
zhichao-cao, line 1606: Can we move the file number extraction logic to a independent function?

utilities/backupable/backupable_db.cc
```diff
@@ -1441,27 +1596,74 @@ Status BackupEngineImpl::VerifyBackup(BackupID backup_id,
...
       if (file_info->checksum_hex != checksum_hex) {
         std::string checksum_info(
             "Expected checksum is " + file_info->checksum_hex +
             " while computed checksum is " + checksum_hex);
-        return Status::Corruption("File corrupted: Checksum mismatch for " +
+        return Status::Corruption("File corrupted: crc32c mismatch for " +
                                   abs_path + ": " + checksum_info);
       }
+      if (checksum_func != nullptr && src_checksum_hex != custom_checksum_hex) {
+        std::string checksum_info("Expected checksum is " + src_checksum_hex +
```
zhichao-cao, line 1649: Also point out the custom checksum verification failed in the message.

utilities/backupable/backupable_db_test.cc
```diff
@@ -766,6 +856,139 @@ class BackupableDBTestWithParam : public BackupableDBTest,
   }
 };
 
+TEST_F(BackupableDBTest, DbAndBackupSameCustomChecksum) {
```
zhichao-cao, line 859: Did you add the testing case to cover the backward compatible?

include/rocksdb/utilities/backupable_db.h
```diff
@@ -165,6 +165,33 @@ struct BackupableDBOptions {
...
+  // custom checksum factory and this option is set to the same factory.
+  //
+  // Regardless of what is set for this option, BackupEngine will always use
+  // its default crc32c checksum for creating, verifying, or restoring backups.
+  //
+  // If this option is set to the DB checksum factory, BackupEngine will try
+  // to find the requested checksum function and use it to create, verify, or
+  // restore backups, in addition to the default crc32c checksum function.
+  // When the DB checksum factory is set to GetFileChecksumGenCrc32cFactory(),
+  // we recommend not setting this option to the same factory to avoid
```
ajkr, line 182: Sorry I didn't quite understand this restriction. What would it take to avoid extra work when a user configures both DB and backup custom checksum function factories with `GetFileChecksumGenCrc32cFactory()`?

include/rocksdb/utilities/backupable_db.h
gg814: I'd prefer "custom". As English is not our first language, maybe Peter or Andrew can help us resolve the conflict? XD

include/rocksdb/utilities/backupable_db.h
gg814: My considerations were as follows. (none of them are critical, just writing down my thoughts, and let me know yours :P)

* The default crc32c function of the backup engine is not exactly the same as `GetFileChecksumGenCrc32cFactory()`. The default backup crc32c is little-endian while `GetFileChecksumGenCrc32cFactory()` is big-endian. (But actually, this consideration is no longer relevant after I changed the backup crc32c value to be a hex-encoded string where I also convert it to a big-endian string. Should I change it back to a little-endian hex-encoded string?)

* We may use one of the two functions but it is not clear which one is better. Should we use the default backup crc32c function? Or should we assume that by setting the option to `GetFileChecksumGenCrc32cFactory()`, users want to overwrite the backup default crc32c function (with a slightly different implementation)?

* When this option is enabled, it is supposed to compute two checksums. So if for whatever reasons users want to compute another crc32c similar to the default by hacking this option, maybe we should allow it.

* The codes look cleaner and it may be easier to maintain. With some extra logic and checking in the codes, we can choose to use one of the two functions. But if for whatever reasons the implementation of `GetFileChecksumGenCrc32cFactory()` changes (such as changing endianness, adding bit-mask at the start and/or at the end), we will not need to change too much in the backup engine. (Maybe we should unify the crc32c functions for db and backup in the future?)

utilities/backupable/backupable_db.cc
gg814: `checksum_func_name` always gives priority to custom checksum function names. That is, when a custom checksum function is used, `checksum_func_name` is always the name of the custom checksum function. Otherwise, it is the name of the backup default. Since there is only one backup default and if it is used, `checksum_func` will be null, we do not need another variable for the name here.

A related question might be: do we need `checksum_func_name`?
I think could remove it (although I prefer keeping it) because it mainly serves as an in-memory sanity-check (or for the purpose of debugging) since we do not write `checksum_func_name` to the backup meta file. If the checking of `checksum_func_name` failed, something seriously wrong should have happened. 

This also reminds me another question I had in mind. Currently, when `LoadFromFile()` is called, backup engine only reads the backup meta files as before and it does not read the backed up db manifests. (Because this will make the initialization of this version's backup engine slower, which might not be a good idea.) Backup engine will read the corresponding db manifest for a specific backup id when it is needed for verification/restoration.

If we choose to read the backed up db manifests when opening the backup engine, then the usage of `checksum_func_name` will be more than in-memory sanity-check. It may help detect corruption of shared table files that have been backed up.

utilities/backupable/backupable_db_test.cc
gg814: Which backward compatible case you meant? If you meant no custom checksum factory for the backup, I think the existing tests (excluding those added by this PR) already cover it :P

gg814:
Thanks Zhichao and Andrew for the review! I've gone through your comments once and will update the PR shortly. :)

utilities/backupable/backupable_db.cc
gg814: Actually, I think custom checksum failure information is in the message :) (`src_checksum_func_name` is the custom checksum function name.)

utilities/backupable/backupable_db.cc
zhichao-cao: Agree, better to keep the `checksum_func_name` for now. Not a high overhead.

utilities/backupable/backupable_db.cc
zhichao-cao: I see, thanks.

ajkr:
> LGTM, thanks for the work! I do not have further comments so I will approve it. See if @ajkr has further comments?

I have no further comments, please go ahead :).

include/rocksdb/utilities/backupable_db.h
```diff
@@ -165,6 +163,34 @@ struct BackupableDBOptions {
...
+  // unnecessary additional computation of crc32c since the backup engine
+  // default checksum function is also crc32c.
+  //
+  // Note: If share_files_with_checksum and share_table_files are true,
+  // the <checksum> appeared in the table filenames will be the custom checksum
+  // value if db session ids are available (namely, table file naming options
+  // is kOptionalChecksumAndDbSessionId and the db session ids obtained from
+  // the table files are nonempty).
+  //
+  // Default: nullptr
```
pdillinger, line 191: I think it's worth adding another note here that we do not require the same setting to file_checksum_gen_factory during restore as was set during backup (each can do the extra checksum verification independently vs. db manifest), but we strongly recommend setting it to the same as the DB file checksum function for all BackupEngine interactions when practical.

include/rocksdb/file_checksum.h
```diff
@@ -22,9 +22,13 @@ namespace ROCKSDB_NAMESPACE {
 constexpr char kUnknownFileChecksum[] = "";
 // The unknown sst file checksum function name.
 constexpr char kUnknownFileChecksumFuncName[] = "Unknown";
+// The default DB file checksum function name.
```
pdillinger, line 25: What do you mean default? This seems confusing to me.

include/rocksdb/file_checksum.h
```diff
@@ -22,9 +22,13 @@ namespace ROCKSDB_NAMESPACE {
...
+constexpr char kDefaultDbFileChecksumFuncName[] = "FileChecksumCrc32c";
 
 struct FileChecksumGenContext {
   std::string file_name;
+  // The name of the requested checksum generator
```
pdillinger, line 30: So that a single factory can create generators for a suite of different functions, like for migrating from one function to another? That seems like interesting new functionality worth mentioning here and in HISTORY.md.

Should a factory have a default function that it returns when this string is empty?

include/rocksdb/file_checksum.h
```diff
@@ -50,6 +54,7 @@ class FileChecksumGenerator {
   virtual std::string GetChecksum() const = 0;
 
   // Returns a name that identifies the current file checksum function.
+  // The name should not contain ':'.
```
pdillinger, line 57: Why?

utilities/backupable/backupable_db_test.cc
```diff
@@ -766,6 +856,139 @@ class BackupableDBTestWithParam : public BackupableDBTest,
...
+    for (int j = 0; j <= i; ++j) {
+      AssertBackupConsistency(j + 1, 0, keys_iteration * (j + 1),
+                              keys_iteration * (j + 2));
+    }
+
+    // delete old data
+    DestroyDB(dbname_, options_);
+  }
+}
+
```
pdillinger, line 991: I don't see failure tests: that failure is reported as expected when crc32c matches but custom checksum doesn't match. Maybe we can hack together a specific way to corrupt a file while preserving its crc32c, to make that "easy" to test?

HISTORY.md
```diff
@@ -27,6 +27,7 @@
...
+* A new option `std::shared_ptr<FileChecksumGenFactory> file_checksum_gen_factory` is added to `BackupableDBOptions`. The default value for this option is `nullptr`. If this option is null, the default backup engine checksum function (crc32c) will be used for creating, verifying, or restoring backups. If it is not null and is set to the DB custom checksum factory, the custom checksum function used in DB will also be used for creating, verifying, or restoring backups, in addition to the default checksum function (crc32c). If it is not null and is set to a custom checksum factory different than the DB custom checksum factory (which may be null), BackupEngine will use only the default checksum function (crc32c) for creating, verifying, or restoring backups.
```
pdillinger, line 30: Now that I see we are supporting different checksum functions from same factory, I'm more concerned about the case of users backup checksum factory quietly not supporting the functions used in the DB. Proposal: it should be an error (invalid argument) if BackupableDBOptions::file_checksum_gen_factory is set and we try to back up a DB with custom checksum not supported by it. If someone needs to backup a DB for which they don't have access to (at least one) custom checksum, I think the best remediation is to require them to unset file_checksum_gen_factory on backup, to acknowledge degraded data integrity checking, rather than very quietly degrading data integrity checking.

gg814: Right, I actually implemented this in an earlier draft but with different error status (`NotSupported()` for a different factory and `NotFound()` for a different generator). The code is still there but the error status is suppressed. So it would be relatively easy to do it in this PR or later :)

include/rocksdb/file_checksum.h
gg814: Yeah, I feel the same. It is more precisely the "default" checksum function provided by the DB if one would like to use file checksum. Didn't figure out a better name and felt this name is "symmetric" to the one used in backup. I think I should update the inline code comment.

include/rocksdb/file_checksum.h
gg814: This is leftover from an earlier draft :P (will remove it)

include/rocksdb/utilities/backupable_db.h
gg814: > I would be OK with removing the special treatment (if we haven't already?)

The treatment is actually not removed but generalized: whenever the `file_checksum_gen_factory` in DB is not null, we will use the data to check hand-off checksums (when backup engine has the same factory and generator).

And yes, I think we should have the story that "setting this option is good" :) I'll remove this confusing inline comment.

utilities/backupable/backupable_db_test.cc
gg814: Right, for arbitrary files it is not difficult to produce such a case. But it may not be easy to hack SST files. Do you have any ideas? :)

pdillinger: Alternate / easier proposal: write a custom checksum function for testing purposes that is not a pure function, but checks a flag on whether to produce a different result value than it normally would.

utilities/backupable/backupable_db_test.cc
```diff
@@ -38,6 +38,141 @@ namespace ROCKSDB_NAMESPACE {
...
 
+class DummyFileChecksumGen : public FileChecksumGenerator {
+ public:
+  explicit DummyFileChecksumGen(const FileChecksumGenContext& /*context*/) {
+    checksum_ = 0;
+  }
+
+  void Update(const char* data, size_t n) override {
+    Random32 rand32(n > 0 ? static_cast<uint32_t>(data[0]) : 0);
+    checksum_ = rand32.Uniform(std::numeric_limits<uint32_t>::max());
```
pdillinger, line 49: This is really weird. Is the goal to get a fresh random value or a seeded pseudorandom value? It seems we're getting a seeded pseudorandom value that depends on how the data is broken up in calls to Update()! And there would be 1/2^32 chance of match anyway. (Inconsistent inconsistency?)

utilities/backupable/backupable_db_test.cc
```diff
@@ -766,6 +901,163 @@ class BackupableDBTestWithParam : public BackupableDBTest,
...
+    ASSERT_OK(backup_engine_->CreateNewBackup(db_.get()));
+    ASSERT_OK(backup_engine_->VerifyBackup(1, false));
+    ASSERT_OK(backup_engine_->VerifyBackup(1, true));
+    CloseDBAndBackupEngine();
+    AssertBackupConsistency(1, 0, keys_iteration, keys_iteration + 1);
+    // delete old data
+    DestroyDB(dbname_, options_);
+  }
+
+  // Mimic a checksum mismatch for custom checksum function by using a dummy
```
pdillinger, line 951: We also need a test showing that such a mismatch in custom checksum can fail Restore also, not just creating new backup.

include/rocksdb/file_checksum.h
pdillinger: I still don't believe "provided by the DB", but I do believe it's provided by the RocksDB API. You have to call GetFileChecksumGenCrc32cFactory() and pass that to DB options for it to be used, right? Can we say "standard" instead of "default"?

include/rocksdb/file_checksum.h
pdillinger: Either here or at CreateFileChecksumGenerator, we have to say something about expectations around
(a) backward compatibility, including
(b) if requested_checksum_func_name is empty
(c) if requested_checksum_func_name is not a recognized function that can be provided. Should requested_checksum_func_name return null unique_ptr?

cc @zhichao-cao

utilities/backupable/backupable_db_test.cc
gg814: Setting a fix seed should work here...

utilities/backupable/backupable_db_test.cc
gg814: Then I'll have to use a sync point :P (since random checksums would not allow one to create a backup let alone verify/restore backups...)

utilities/backupable/backupable_db_test.cc
```diff
@@ -766,6 +908,183 @@ class BackupableDBTestWithParam : public BackupableDBTest,
...
+    AssertBackupConsistency(i - 1, 0, keys_iteration * (i - 1),
+                            keys_iteration * i);
+    AssertBackupConsistency(i, 0, keys_iteration * i, keys_iteration * (i + 1));
+    AssertBackupConsistency(i + 1, 0, keys_iteration * (i + 1),
+                            keys_iteration * (i + 2));
+
+    // 4) no custom checksums
+    ++i;
+    OpenDBAndBackupEngine(false /* destroy_old_data */, false /* dummy */,
+                          sopt);
```
pdillinger, line 1071: I don't think this tests what it's intended to. When I add this here it fails:

    ASSERT_FALSE(backupable_options_->file_checksum_gen_factory);

include/rocksdb/file_checksum.h
pdillinger: Still TODO

utilities/backupable/backupable_db_test.cc
gg814: Indeed, this is a silly mistake by me :P

include/rocksdb/file_checksum.h
zhichao-cao: > So that a single factory can create generators for a suite of different functions, like for migrating from one function to another? That seems like interesting new functionality worth mentioning here and in HISTORY.md.
> 
> Should a factory have a default function that it returns when this string is empty?

I think Zippy is trying the similar things. For the same Factory, they use SHA1 as default, but also support crc32c. But I'm not sure how they achieve the switch.

zhichao-cao: > Either here or at CreateFileChecksumGenerator, we have to say something about expectations around
> (a) backward compatibility, including
> (b) if requested_checksum_func_name is empty
> (c) if requested_checksum_func_name is not a recognized function that can be provided. Should requested_checksum_func_name return null unique_ptr?
> 
> cc @zhichao-cao

My question here is, should we set a concrete contract with the user that provide the implementation of ChecksumGenFactory. If so, then, we need to set clear rules to regulate the behavior of CreateFileChecksumGenerator 1) if requested_checksum_func_name is empty, 2) if requested_checksum_func_name does not match. Otherwise, user can define the behavior of their own Factory.

PR #7089, opened 2020-07-07
[WIP] Track WAL in MANIFEST

cheng-chang:
This PR adds WAL lifecycle related information to VersionEdit.
When a WAL is closed because the corresponding memtable is flushed, the WAL's log number and size are tracked in VersionEdit. When a WAL is archived or deleted, the event is also tracked in VersionEdit.

An alternative design is to track WAL's file path, instead of the log number. This PR prefers to track log number to save disk space occupied by MANIFEST.

Another design choice is to just track added WAL and deleted WAL, for archived WAL, treat is as deleting the old WAL and adding a new WAL in the archive directory. This PR prefers to introduce a specific entry for archived WAL, because it only needs to track the log number in one entry instead of two entries (delete and add), which saves disk space of MANIFEST.

Follow up PRs will:
1. apply the WAL related VersionEdits to MANIFEST.
2. create WAL related VersionEdits in relevant code paths of WAL's lifecycle.
3. compute and track checksum of the WALs.

Test Plan:
A set of new unit tests are added to version_edit_test.cc.

### DIFF
--- a/db/version_edit.cc
+++ b/db/version_edit.cc
@@ -83,13 +83,13 @@ enum NewFileCustomTag : uint32_t {
 
 }  // anonymous namespace
 
-void AddedWal::EncodeTo(std::string* dst) const {
+void WalMetadata::EncodeTo(std::string* dst) const {
   PutVarint64(dst, number_);
   PutVarint64(dst, bytes_);
 }
 
-Status AddedWal::DecodeFrom(Slice* src) {
-  constexpr char class_name[] = "AddedWal";
+Status WalMetadata::DecodeFrom(Slice* src) {
+  constexpr char class_name[] = "WalMetadata";
   if (!GetVarint64(src, &number_)) {
     return Status::Corruption(class_name, "Error decoding WAL log number");
   }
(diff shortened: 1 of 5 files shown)


ltamasi:
Just wanted to clarify what our goal is here... If we want to make sure that WAL files don't go missing from the filesystem, I think we should register them in the MANIFEST when they are first opened (i.e. upon DB open or CF flush) as opposed to when they are closed. (Of course, no size/checksum information is available at that time, so we would need to write a second record with that information upon close.) Also note we already have some information in the MANIFEST about which WAL files can be deleted (see `log_number_` and friends in `VersionEdit`).

db/version_edit.h
```diff
@@ -241,6 +241,57 @@ struct LevelFilesBrief {
   }
 };
 
+using WalNumber = uint64_t;
+
+// Information of a new WAL tracked by VersionEdit.
+class AddedWal {
+ public:
+  AddedWal() = default;
```
ltamasi: Note that this leaves `number_` and `bytes_` essentially uninitialized; consider using in-class initializers. (Affects `DeletedWal` too.)

db/version_edit.h
```diff
@@ -241,6 +241,57 @@ struct LevelFilesBrief {
...
+  Status DecodeFrom(Slice* src);
+
+  std::string DebugString() const;
+
+ private:
+  WalNumber number_;
+  uint64_t bytes_;
+};
+
+JSONWriter& operator<<(JSONWriter& jw, const AddedWal& wal);
```
ltamasi: We should probably declare the other `operator<<` too here (also affects the other class).

db/version_edit.h
```diff
@@ -413,6 +464,26 @@ class VersionEdit {
   bool IsInAtomicGroup() const { return is_in_atomic_group_; }
   uint32_t GetRemainingEntries() const { return remaining_entries_; }
 
+  // Add a completed WAL.
+  void AddWal(const AddedWal& wal) { new_wals_.push_back(wal); }
+
+  // Get a list of non-archived WALs in the order of AddWal.
+  std::vector<AddedWal> GetNewWals() { return new_wals_; }
```
ltamasi: This takes a copy; return a const-ref instead? Also, the method itself could be marked `const`. (Affects the other two getters too.)

db/version_edit_test.cc
```diff
@@ -309,6 +309,109 @@ TEST_F(VersionEditTest, BlobFileAdditionAndGarbage) {
...
+}
+
+TEST_F(VersionEditTest, AddedWal) {
+  constexpr uint64_t kLogNumber = 10;
+  constexpr uint64_t kBytes = 100;
+
+  VersionEdit edit;
+  edit.AddWal(AddedWal(kLogNumber, kBytes));
+
+  auto wals = edit.GetNewWals();
```
ltamasi, line 339: `const auto& wals` ? Also, `const AddedWal& wal` below. (Affects the other two test cases as well.)

db/version_edit.cc
```diff
@@ -302,6 +378,22 @@ bool VersionEdit::EncodeTo(std::string* dst) const {
     PutVarint32(dst, kInAtomicGroup);
     PutVarint32(dst, remaining_entries_);
   }
+
+  for (const AddedWal& wal : new_wals_) {
+    PutVarint32(dst, kNewWal);
+    wal.EncodeTo(dst);
+  }
+
+  for (WalNumber log_number : archived_wals_) {
```
ltamasi, line 351: Assuming we want to track these events, we could consider introducing a dedicated class `ArchivedWal` that would encapsulate the related logic.

db/version_edit.h
```diff
@@ -241,6 +241,57 @@ struct LevelFilesBrief {
...
+};
+
+JSONWriter& operator<<(JSONWriter& jw, const AddedWal& wal);
+
+// Information of a deleted WAL.
+class DeletedWal {
+ public:
+  DeletedWal() = default;
+
+  DeletedWal(WalNumber number, bool archived)
```
ltamasi, line 276: Again, this is contingent on what we want to track but `archived` could be inferred from the presence of `kArchivedWal`, right?

cheng-chang:
Closing the PR for now since it's still WIP.

PR #7096, opened 2020-07-07
Update Flush policy in PartitionedIndexBuilder on switching from user-key to internal-key mode

akankshamahajan15:
Summary: When format_version is high enough to support user-key and
there are index entries for same user key that spans multiple data
blocks then it changes from user-key mode to internal-key mode. But the
flush policy is not reset to point to Block Builder of internal-keys.
After this switch, no entries are added to user key index partition
result, thus it never triggers flushing the block.

Fix: 1. After adding the entry in sub_builder_index_, if there is a switch
from user-key to internal-key, then flush policy is updated to point to
Block Builder of internal-keys index partition.
2. Set sub_builder_index_->seperator_is_key_plus_seq_ = true if
seperator_is_key_plus_seq_  is set to true so that subsequent partitions
can also use internal key mode.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/table/block_based/index_builder.cc
+++ b/table/block_based/index_builder.cc
@@ -104,6 +104,15 @@ void PartitionedIndexBuilder::MakeNewSubIndexBuilder() {
       comparator_, table_opt_.index_block_restart_interval,
       table_opt_.format_version, use_value_delta_encoding_,
       table_opt_.index_shortening, /* include_first_key */ false);
+
+  // Set sub_index_builder_->seperator_is_key_plus_seq_ to true if
+  // seperator_is_key_plus_seq_ is true (internal-key mode) (set to false by
+  // default on Creation) so that flush policy can point to
+  // sub_index_builder_->index_block_builder_
+  if (seperator_is_key_plus_seq_) {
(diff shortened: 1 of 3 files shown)


table/block_based/index_builder.cc
```diff
@@ -104,6 +104,9 @@ void PartitionedIndexBuilder::MakeNewSubIndexBuilder() {
       comparator_, table_opt_.index_block_restart_interval,
       table_opt_.format_version, use_value_delta_encoding_,
       table_opt_.index_shortening, /* include_first_key */ false);
+  if (seperator_is_key_plus_seq_) {
+    sub_index_builder_->seperator_is_key_plus_seq_ = false;
```
ajkr, line 108: Why do we flip this back to false after `seperator_is_key_plus_seq_` becomes true?

akankshamahajan15: I was checking for something. Kindly ignore this change. It was in draft state initially. I will update it for review once i complete the changes.

akankshamahajan15:
All testcases have passed. LMK if you think this logic have any side effects

PR #7097, opened 2020-07-07
More Makefile Cleanup

mrambacher:
Cleans up some of the dependencies on test code in the Makefile while building tools:
- Moves the test::RandomString, DBBaseTest::RandomString into Random
- Moves the test::RandomHumanReadableString into Random
- Moves the DestroyDir method into file_utils
- Moves the SetupSyncPointsToMockDirectIO into sync_point.
- Moves the FaultInjection Env and FS classes under env

These changes allow all of the tools to build without dependencies on test_util, thereby simplifying the build dependencies.  By moving the FaultInjection code, the dependency in db_stress on different libraries for debug vs release was eliminated.

Tested both release and debug builds via Make and CMake for both static and shared libraries.

More work remains to clean up how the tools are built and remove some unnecessary dependencies.  There is also more work that should be done to get the Makefile and CMake to align in their builds -- what is in the libraries and the sizes of the executables are different.

### DIFF
--- a/test_util/sync_point.h
+++ b/test_util/sync_point.h
@@ -124,6 +124,9 @@ class SyncPoint {
   Data*  impl_;
 };
 
+// Sets up sync points to mock direct IO instead of actually issuing direct IO
+// to the file system.
+void SetupSyncPointsToMockDirectIO();
 }  // namespace ROCKSDB_NAMESPACE
 
 // Use TEST_SYNC_POINT to specify sync points inside code base.
--- a/utilities/fault_injection_fs.cc
+++ b/utilities/fault_injection_fs.cc
@@ -14,11 +14,12 @@
 // FileSystem related operations, by specify the "IOStatus Error", a specific
 // error can be returned when file system is not activated.
 
-#include "test_util/fault_injection_test_fs.h"
+#include "utilities/fault_injection_fs.h"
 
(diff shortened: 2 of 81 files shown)


utilities/fault_injection_fs.cc
```diff
@@ -501,8 +505,7 @@ IOStatus FaultInjectionTestFS::InjectError(ErrorOperation op,
               // The randomly generated string could be identical to the
               // original one, so retry
               do {
-                str = DBTestBase::RandomString(&ctx->rand,
-                                    static_cast<int>(len));
+                str = ctx->rand.RandomString(static_cast<int>(len));
```
pdillinger, line 508: I was a little concerned about 'git blame's ability to track this rename with level of change, but I checked out, collapsed to one commit, and the git blame works fine. :)

test_util/sync_point.h
```diff
@@ -124,6 +124,9 @@ class SyncPoint {
   Data*  impl_;
 };
 
+// Sets up sync points to mock direct IO instead of actually issuing direct IO
+// to the file system.
+void SetupSyncPointsToMockDirectIO();
```
pdillinger, line 129: A little weird having this here, but I agree is less bad than other options. :)

PR #7100, opened 2020-07-07
Implement IOTracing 

akankshamahajan15:
Summary: As part of IOTracing project, this PR contains following changes.
1. Changed the code to cache "FileSystemPtr" pointer(wrapper class that returns file system pointer based on tracing enabled) instead of "FileSystem" pointer.
2. Add IOTrace in Options.
3. "FileSystemPtr" pointer is created in db_impl.cc and Options.env->GetFileSystem() and Options.io_tracer are passed as arguments.
4. "ImmutableDBOptions.fs" is used in the code to get the "FileSystemPtr" pointer.
5. In "SequentialFileReader" class, "FSSequentialFilePtr" (wrapper class) ptr is created and that pointer is used to call underlying file system based on tracing enabled/disabled.
6. For "FSRandomAccessFile" and  "FSWritableFile" as well, their Reader/Writer classes create a wrapper class pointer and io_tracer pointer is passed to the Reader/Writer classes for creation of wrapper class pointers.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -616,6 +616,7 @@ set(SOURCES
         env/env_encryption.cc
         env/env_hdfs.cc
         env/file_system.cc
+        env/file_system_tracer.cc
         env/mock_env.cc
         file/delete_scheduler.cc
         file/file_prefetch_buffer.cc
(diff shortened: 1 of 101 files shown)


akankshamahajan15:
Creating new smaller PRs for better review and changing the implementation as there doesn't need to do a lot of changes and this design is way too complicated than it needs to be.

PR #7104, opened 2020-07-08
Fix GetFileDbIdentities

gg814:
Summary:
Although PR #7032 fixes the construction of the `SstFileDumper` in `GetFileDbIdentities` by setting a proper `Env` of the `Options` passed in the constructor, the file path was not corrected accordingly. This actually disables backup engine to use db session ids in the file names since the `db_session_id` is always empty.

Now it is fixed by setting the correct path in the construction of `SstFileDumper`. Furthermore, to preserve the Direct IO property that backup engine already has, parameter `EnvOptions` is added to `GetFileDbIdentities` and `SstFileDumper`.

The `BackupUsingDirectIO` test is updated accordingly.

Test plan:
backupable_db_test and some manual tests.

### DIFF
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -364,8 +364,9 @@ class BackupEngineImpl : public BackupEngine {
                            uint64_t size_limit, uint32_t* checksum_value);
 
   // Obtain db_id and db_session_id from the table properties of file_path
-  Status GetFileDbIdentities(Env* src_env, const std::string& file_path,
-                             std::string* db_id, std::string* db_session_id);
+  Status GetFileDbIdentities(Env* src_env, const EnvOptions& src_env_options,
+                             const std::string& file_path, std::string* db_id,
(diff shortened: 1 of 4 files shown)


utilities/backupable/backupable_db.cc
```diff
@@ -1798,20 +1802,20 @@ Status BackupEngineImpl::CalculateChecksum(const std::string& src, Env* src_env,
...
 
 Status BackupEngineImpl::GetFileDbIdentities(Env* src_env,
+                                             const EnvOptions& src_env_options,
                                              const std::string& file_path,
                                              std::string* db_id,
                                              std::string* db_session_id) {
   assert(db_id != nullptr || db_session_id != nullptr);
-  // // Prepare the full_path of file_path under src_env for SstFileDumper
-  std::string full_path;
-  src_env->GetAbsolutePath(file_path, &full_path);
```
ajkr, line 1807: Oh right, I also forgot we should remove this after fixing the src `Env`.

PR #7105, opened 2020-07-08
Parameterize a few tests in DBWALTest

riversand963:
Summary:
As title. The goal is to shorten the execution time of several tests
when they are combined together in a single TEST_F.

Test Plan:
make db_wal_test
./db_wal_test

### DIFF
--- a/db/db_wal_test.cc
+++ b/db/db_wal_test.cc
@@ -1041,15 +1041,32 @@ class RecoveryTestHelper {
   }
 };
 
-class DBWALTestWithParams : public DBWALTestBase,
-                            public ::testing::WithParamInterface<
-                                std::tuple<bool, int, int, WALRecoveryMode>> {
+class DBWALTestWithParams
+    : public DBWALTestBase,
+      public ::testing::WithParamInterface<std::tuple<bool, int, int>> {
  public:
   DBWALTestWithParams() : DBWALTestBase("/db_wal_test_with_params") {}
 };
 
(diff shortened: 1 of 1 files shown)


db/db_wal_test.cc
```diff
@@ -1035,76 +1041,93 @@ class RecoveryTestHelper {
...
-        } else {
-          options.wal_recovery_mode =
-              WALRecoveryMode::kTolerateCorruptedTailRecords;
-          ASSERT_NOK(TryReopen(options));
-        }
-      }
-    }
+TEST_P(DBWALTestWithParams, kTolerateCorruptedTailRecords) {
+  WALRecoveryMode recovery_mode = std::get<3>(GetParam());
+  if (recovery_mode != WALRecoveryMode::kTolerateCorruptedTailRecords) {
```
ltamasi, line 1070: I see that the first few test cases really only operate in one recovery mode; we could consider parameterizing this test in two different ways (using two different `DBWALTestBase` subclasses as fixture), one taking `std::tuple<bool, int, int>` as parameter and one taking `std::tuple<bool, int, int, WALRecoveryMode>`. (The latter would be used for `RecoverFromCorruptedWALWithoutFlush`.)

riversand963: SG. Will update accordingly. This can reduce the number of tests.

riversand963:
Test failure in FormatLatest/ColumnFamilyTest.IteratorCloseWALFile1/0 on ARM64 is not related to this PR.

PR #7115, opened 2020-07-11
db_wal_test.cc: ASSERT_LT -> ASSERT_LE

rockeet:
this `ASSERT_LT` always fail with equal value, so I think it should be `ASSERT_LE`.

### DIFF
--- a/db/db_wal_test.cc
+++ b/db/db_wal_test.cc
@@ -1585,7 +1585,7 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
   ASSERT_EQ(1, log_files_after.size());
   ASSERT_LT(log_files_after[0]->SizeFileBytes(), 1 * kKB);
   // The preallocated space should be truncated.
-  ASSERT_LT(GetAllocatedFileSize(dbname_ + file_before->PathName()),
+  ASSERT_LE(GetAllocatedFileSize(dbname_ + file_before->PathName()),
             preallocated_size);
 }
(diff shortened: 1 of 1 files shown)


db/db_wal_test.cc
```diff
@@ -1585,7 +1585,7 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
   ASSERT_EQ(1, log_files_after.size());
   ASSERT_LT(log_files_after[0]->SizeFileBytes(), 1 * kKB);
   // The preallocated space should be truncated.
-  ASSERT_LT(GetAllocatedFileSize(dbname_ + file_before->PathName()),
+  ASSERT_LE(GetAllocatedFileSize(dbname_ + file_before->PathName()),
```
ajkr, line 1588: Hm, I think if they're equal, that implies the preallocated space was not truncated. We should probably investigate more why it's not being truncated. If you have more details on platform and whether strace shows `ftruncate()` being called, it could help us investigate.

For example on my machine:
```
$ strace -ye ftruncate,fallocate ./db_wal_test --gtest_filter=DBWALTest.TruncateLastLogAfterRecoverWithoutFlush
...
fallocate(8</tmp/rocksdbtest-9383/db_wal_test_13666978242616246148/000003.log>, FALLOC_FL_KEEP_SIZE, 0, 18427904) = 0
...
ftruncate(9</tmp/rocksdbtest-9383/db_wal_test_13666978242616246148/000003.log>, 27) = 0
...
```

PR #7117, opened 2020-07-11
Memtable refactory to support general memtable

rockeet:
Now MemTableRep use prefix len encoded key-value, this works good for rocksdb builtin MemTables, but it limit the design space for more general MemTable implementation, such as a patricia trie implementation, which does not store the key content, but restore the key content during searching/iterating.

This PR refactory the MemTableRep interface with `Slice key, value`, which is more general. Though `Key` is not required to be hard wired as prefix-len encoding, the `Value` keep prefix-len encoding to simplify `Update` implementaion.


rockeet:
@siying Could this PR be merged?

PR #7141, opened 2020-07-16
Remove time out testing cases in error_handler_fs_test 

zhichao-cao:
Remove the 3 testing cases that cause the time out in linux build by #6765 . Will fix them later.

Test plan: make asan_check, buck run


zhichao-cao:
> Can the tests be marked as `DISABLED_` and left in the codebase? https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#temporarily-disabling-tests

It might be a better idea. I will do as suggested.

PR #7142, opened 2020-07-16
IO Tracer Parsing tool

akankshamahajan15:
Summary: Implement a parsing tool io_tracer_parser that takes input_file (binary file) with command line argument --input_file and output_file with --output_file and dumps the IO trace records in output_file in human readable form.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/tools/io_tracer_parser_tool.cc
+++ b/tools/io_tracer_parser_tool.cc
@@ -1,0 +1,147 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//    This source code is licensed under both the GPLv2 (found in the
+//    COPYING file in the root directory) and Apache 2.0 License
+//    (found in the LICENSE.Apache file in the root directory).
+
+#ifndef ROCKSDB_LITE
+
+#include "rocksdb/io_tracer_parser_tool.h"
+
+#include <cinttypes>
+#include <cstdio>
+#include <iomanip>
+#include <memory>
+#include <sstream>
+
+#include "port/lang.h"
+#include "rocksdb/env.h"
+#include "trace_replay/io_tracer.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+IOTraceRecordParser::IOTraceRecordParser(const std::string& input_file,
+                                         const std::string& output_file)
+    : input_file_(input_file), output_file_(output_file) {}
+
+IOTraceRecordParser::~IOTraceRecordParser() {
+  if (trace_writer_) {
+    trace_writer_->Flush();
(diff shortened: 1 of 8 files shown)


tools/io_tracer_parser_tool.cc
```diff
@@ -0,0 +1,108 @@
...
+
+  status = NewFileTraceReader(env, EnvOptions(), input_file_, &trace_reader);
+  if (!status.ok()) {
+    fprintf(stderr, "%s: %s\n", input_file_.c_str(), status.ToString().c_str());
+    return 1;
+  }
+  reader.reset(new IOTraceReader(std::move(trace_reader)));
+
+  IOTraceHeader header;
+  status = reader->ReadHeader(&header);
```
mrambacher, line 67: Is there nothing from the header that is meant to be printed?

tools/io_tracer_parser_tool.cc
```diff
@@ -0,0 +1,108 @@
...
+
+namespace ROCKSDB_NAMESPACE {
+
+IOTraceRecordParser::IOTraceRecordParser(std::string input_file)
+    : input_file_(input_file) {}
+
+void IOTraceRecordParser::PrintHumanReadableIOTraceRecord(
+    const IOTraceRecord& record) {
+  fprintf(stdout,
+          "Access Time : %" PRIu64 ", File Operation: %s, Latency: %" PRIu64
```
zhichao-cao, line 26: Should we give an option to decide whether the output file is for human read or for program processing? If we out put a file in this format, might not be convenient for other applications to process.

akankshamahajan15: This is for human read. I am working on providing an output_file option. I am not sure about the program processing.

tools/io_tracer_parser_tool.cc
```diff
@@ -0,0 +1,108 @@
...
+#include "rocksdb/env.h"
+#include "trace_replay/io_tracer.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+IOTraceRecordParser::IOTraceRecordParser(std::string input_file)
+    : input_file_(input_file) {}
+
+void IOTraceRecordParser::PrintHumanReadableIOTraceRecord(
+    const IOTraceRecord& record) {
```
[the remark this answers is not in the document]
akankshamahajan15: Yes, I am working on providing an output_file option as well

akankshamahajan15:
Moved over https://github.com/facebook/rocksdb/pull/7143 that contain all commits related to IOTracing

PR #7143, opened 2020-07-16
Io tracer parser tool

akankshamahajan15:
Summary: Implement a parsing tool io_tracer_parser that takes input_file (binary file) with command line argument --input_file and output_file with --output_file and dumps the IO trace records in output_file in human readable form.

Screen Shot 2020-07-16 at 1 53 33 PM

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/file/writable_file_writer.h
+++ b/file/writable_file_writer.h
@@ -112,7 +112,8 @@ class WritableFileWriter {
   bool ShouldNotifyListeners() const { return !listeners_.empty(); }
   void UpdateFileChecksum(const Slice& data);
 
-  std::unique_ptr<FSWritableFile> writable_file_;
+  std::unique_ptr<FSWritableFilePtr> writable_file_ptr_;
+  FSWritableFile* writable_file_;
   std::string file_name_;
   Env* env_;
   AlignedBuffer buf_;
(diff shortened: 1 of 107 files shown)


tools/io_tracer_parser_tool.cc
```diff
@@ -0,0 +1,147 @@
...
+    }
+    PrintHumanReadableIOTraceRecord(record);
+  }
+  return 0;
+}
+
+int IOTracerParserTool::run(int argc, char const* const* argv) {
+  std::string input_file;
+  std::string output_file;
+  for (int i = 1; i < argc; i++) {
```
zhichao-cao, line 119: It would be much easier to use gflags to process the input arguments. And also we can add the more options in the future.

file/writable_file_writer.h
```diff
@@ -138,11 +139,11 @@ class WritableFileWriter {
  public:
   WritableFileWriter(
       std::unique_ptr<FSWritableFile>&& file, const std::string& _file_name,
-      const FileOptions& options, Env* env = nullptr,
-      Statistics* stats = nullptr,
+      const FileOptions& options, std::shared_ptr<IOTracer> io_tracer,
```
mrambacher, line 142: Rather than passing the IOTracer here, would it be better to add the IOTrace to the Env/FileOptions?  This would eliminate the need for the Ptr classes, as you could simply return the proper "Writer"

akankshamahajan15:
Creating new smaller PRs for better review and changing the implementation as there doesn't need to do a lot of changes and this design is way too complicated than it needs to be.

PR #7144, opened 2020-07-17
Add cmake-mingw in circle-build

akankshamahajan15:
Summary: Add cmake-mignw in circle-build

Test Plan: watch circle cmake-mingw build

Reviewers:

Subscribers:

Tasks: T70028805

Tags:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -354,7 +354,7 @@ jobs:
   build-examples:
     machine:
       image: ubuntu-1604:202007-01
-    resource_class: medium
+    resource_class: large
     steps:
       - pre-steps
       - install-gflags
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -364,6 +364,29 @@ jobs:
...
+          command: |
+            sudo apt-get install snapd && sudo snap install cmake --beta --classic
+            export PATH=/snap/bin:$PATH
+            sudo apt-get install -y openjdk-8-jdk
+            export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
+            export PATH=$JAVA_HOME/bin:$PATH
+            echo "JAVA_HOME=${JAVA_HOME}"
+            which java && java -version
+            which javac && javac -version
+            mkdir build && cd build && cmake -DJNI=1 -DWITH_GFLAGS=OFF .. -DCMAKE_C_COMPILER=x86_64-w64-mingw32-gcc -DCMAKE_CXX_COMPILER=x86_64-w64-mingw32-g++ -DCMAKE_SYSTEM_NAME=Windows && make -j4 rocksdb rocksdbjni
```
jay-zhuang, line 387: do we need to run test?

.circleci/config.yml
```diff
@@ -354,7 +354,7 @@ jobs:
   build-examples:
     machine:
       image: ubuntu-1604:202007-01
-    resource_class: medium
+    resource_class: large
```
jay-zhuang, line 357: is medium too slow?

jay-zhuang (approved):
LGTM. Thanks.

.circleci/config.yml
akankshamahajan15: yes it is. Thats why i changed it to large

.circleci/config.yml
akankshamahajan15: In travis, these tests were run so added the same command here as well.

PR #7145, opened 2020-07-17
stagger first DumpMallocStats after opening DB

ajkr:
Previously when running `db_bench` with large value for `num_multi_dbs` and enabled `Options::dump_malloc_stats`, we would see most CPU spent in jemalloc locking. After this PR that no longer shows up at the top of the profile.

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -685,9 +685,18 @@ void DBImpl::StartTimedTasks() {
     stats_dump_period_sec = mutable_db_options_.stats_dump_period_sec;
     if (stats_dump_period_sec > 0) {
       if (!thread_dump_stats_) {
+        // In case of many `DB::Open()` in rapid succession we can have all
+        // threads dumping at once, which causes severe lock contention in
+        // jemalloc. Ensure successive `DB::Open()`s are staggered by at least
+        // one second in the common case.
+        static unsigned int stats_dump_threads_started = 0;
         thread_dump_stats_.reset(new ROCKSDB_NAMESPACE::RepeatableThread(
             [this]() { DBImpl::DumpStats(); }, "dump_st", env_,
-            static_cast<uint64_t>(stats_dump_period_sec) * kMicrosInSecond));
+            static_cast<uint64_t>(stats_dump_period_sec) * kMicrosInSecond,
+            static_cast<uint64_t>(stats_dump_threads_started %
+                                  stats_dump_period_sec) *
+                kMicrosInSecond));
+        ++stats_dump_threads_started;
       }
     }
     stats_persist_period_sec = mutable_db_options_.stats_persist_period_sec;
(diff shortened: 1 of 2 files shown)


db/db_impl/db_impl.cc
```diff
@@ -685,9 +685,18 @@ void DBImpl::StartTimedTasks() {
...
+        // In case of many `DB::Open()` in rapid succession we can have all
+        // threads dumping at once, which causes severe lock contention in
+        // jemalloc. Ensure successive `DB::Open()`s are staggered by at least
+        // one second in the common case.
+        static unsigned int stats_dump_threads_started = 0;
         thread_dump_stats_.reset(new ROCKSDB_NAMESPACE::RepeatableThread(
             [this]() { DBImpl::DumpStats(); }, "dump_st", env_,
-            static_cast<uint64_t>(stats_dump_period_sec) * kMicrosInSecond));
+            static_cast<uint64_t>(stats_dump_period_sec) * kMicrosInSecond,
+            static_cast<uint64_t>(stats_dump_threads_started %
```
riversand963, line 696: It may be rare case:
since `stats_dump_threads_started` has process life scope, using an `unsigned int` as a counter sounds a little scary to me. How about make it an `uint64_t`?

db/db_impl/db_impl.cc
```diff
@@ -1083,10 +1092,20 @@ Status DBImpl::SetDBOptions(
           mutex_.Lock();
         }
         if (new_options.stats_dump_period_sec > 0) {
+          // In case many DBs have `stats_dump_period_sec` enabled in rapid
+          // succession, we can have all threads dumping at once, which causes
+          // severe lock contention in jemalloc. Ensure successive enabling of
+          // `stats_dump_period_sec` are staggered by at least one second in the
+          // common case.
+          static unsigned int stats_dump_threads_started = 0;
```
riversand963, line 1100: Ditto.

db/db_impl/db_impl.cc
ajkr: Sure. I don't think unsigned int overflow is dangerous in this situation, but we can avoid it anyways since it's confusing.

riversand963:
> > > Thanks @ajkr for the fix.
> > 
> > 
> > Also, @jay-zhuang is working on letting multiple dbs share the same thread for stats dumping and persisting. Maybe this issue will go away once that's done.
> 
> Thanks, I was not aware of that work. Indeed it should go away (will people have to enable it?).

Currently, we enable stats dumping and persisting by default with an interval of 600 seconds, assuming users by default want this feature. If this assumption still holds, I don't think user need to explicitly re-enable Jay's WIP feature. If the assumption changes and we change the default values of `stats_dump_period_sec` and/or `stats_persist_period_sec` back to 0, then Jay's feature need to be enabled explicitly.

db/db_impl/db_impl.cc
mrambacher, line 1100: I have two questions:
1.  Is it really intended to have a different "stats_dump_threads_started" count for SetDBOptions than from StartTimedTasks?  Should the value be shared and static to the file, rather than to each method.

2.  In the case of SetDBOptions, should this "stated" delay always be there, or only there if the task was not scheduled previously?

ajkr: 1. I suppose they can be independent, as I'm not worried about the case where enabling a stats dump thread via DB open and enabling it by `SetOptions()` are called in rapid succession.
2. I think always is fine. The users of `SetDBOptions()` often apply a dynamic option change to all their DBs in rapid succession, so it seems important to stagger every time we start a stats dump thread.

PR #7148, opened 2020-07-18
Using Read/Write Lock guard in all corresponding places

toktarev:
The problem in many places is that RocksDB uses Read/Write locks in non safe mode.

This PR protects corresponding places with corresponding guards.

### DIFF
--- a/db/memtable.cc
+++ b/db/memtable.cc
@@ -13,6 +13,7 @@
 #include <array>
 #include <limits>
 #include <memory>
+
 #include "db/dbformat.h"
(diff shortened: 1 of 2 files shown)


toktarev:
@ajkr may I have any chance for review ?

cculianu:
Lock guards should always be used in C++ code that may throw.  I forget if rocksdb is compiled with exceptions (I think it may not be?). Regardless lack of use of lock guards in modern C++ is in general a code smell... so this PR or something like it should be considered!

PR #7149, opened 2020-07-20
minimize BlockIter comparator scope

ajkr:
PR #6944 transitioned `BlockIter` from using `Comparator*` to using
concrete `UserComparatorWrapper` and `InternalKeyComparator`. However,
adding them as instance variables to `BlockIter` was not optimal.
Bloating `BlockIter` caused the `ArenaWrappedDBIter`'s arena allocator to do more heap
allocations (in certain cases) which harmed performance of `DB::NewIterator()`. This PR
pushes down the concrete comparator objects to the point of usage, which
forces them to be on the stack. As a result, the `BlockIter` is back to
its original size prior to #6944 (actually a bit smaller since there
were two `Comparator*` before).

Test Plan: verified our internal `DB::NewIterator()`-heavy regression
test no longer reports meaningful regression.

### DIFF
--- a/table/block_based/block.h
+++ b/table/block_based/block.h
@@ -168,7 +168,7 @@ class Block {
 
   BlockBasedTableOptions::DataBlockIndexType IndexType() const;
 
-  // ucmp is a raw (i.e., not wrapped by `UserComparatorWrapper`) user key
+  // raw_ucmp is a raw (i.e., not wrapped by `UserComparatorWrapper`) user key
   // comparator.
   //
   // If iter is null, return new Iterator
@@ -187,13 +187,13 @@ class Block {
   // NOTE: for the hash based lookup, if a key prefix doesn't match any key,
   // the iterator will simply be set as "invalid", rather than returning
   // the key that is just pass the target key.
-  DataBlockIter* NewDataIterator(const Comparator* ucmp,
+  DataBlockIter* NewDataIterator(const Comparator* raw_ucmp,
                                  SequenceNumber global_seqno,
                                  DataBlockIter* iter = nullptr,
                                  Statistics* stats = nullptr,
                                  bool block_contents_pinned = false);
 
-  // ucmp is a raw (i.e., not wrapped by `UserComparatorWrapper`) user key
+  // raw_ucmp is a raw (i.e., not wrapped by `UserComparatorWrapper`) user key
   // comparator.
   //
   // key_includes_seq, default true, means that the keys are in internal key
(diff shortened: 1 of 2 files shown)


table/block_based/block.h
```diff
@@ -419,15 +421,16 @@ class BlockIter : public InternalIteratorBase<TValue> {
...
-      return ucmp_wrapper_.Compare(raw_key_.GetUserKey(), other);
+      return ucmp().Compare(raw_key_.GetUserKey(), other);
     } else if (global_seqno_ == kDisableGlobalSequenceNumber) {
-      return icmp_.Compare(raw_key_.GetInternalKey(), other);
+      return icmp().Compare(raw_key_.GetInternalKey(), other);
     }
-    return icmp_.Compare(raw_key_.GetInternalKey(), global_seqno_, other,
-                         kDisableGlobalSequenceNumber);
+    return icmp().Compare(raw_key_.GetInternalKey(), global_seqno_, other,
+                          kDisableGlobalSequenceNumber);
```
XinyuZeng, line 429: Hi @ajkr , I found that after this PR, each time when we call a CompareCurrentKey operation a new comparator will be initiated and constructed. I did some basic profiling and found that this may be a bottleneck. I was wondering what is the reason for this change, and would it harm performance? Thanks!

XinyuZeng: Thanks for your reply. Could you explain more? From what I see, DataBlockIterator still calls CompareCurrentKey in BlockIter and either icmp() or ucmp() will allocate a new comparator.

XinyuZeng: @mrambacher Thanks. I get that the "raw comparator" will be saved as raw_ucmp_ and passed as pointer to become UserComparatorWrapper when calling either icmp() or ucmp(), so this step has no allocation cost. However, calling icmp() or ucmp() will create a new Comparator instance, which is a subclass of Customizable. My concern is that the construction and destruction of such instances in every comparison operation may have a cost, as showed in some of my profiling.

mrambacher: @XinyuZeng I was suggesting something different.  Store the "Internal Comparator" in the iterator (the icmp) instead of the raw comparator.  The icmp now returns this without creating anything.  The ucmp returns the "user_comparator" from the InternalKeyComparator -- again without creating anything.  The size of the object does not change (just storing a different pointer) and there is no overhead of creating objects on every call.

If you are not comfortable attempting this change, I can try to look into it more later in the week.

XinyuZeng: @mrambacher Yes that is exactly what I am thinking about----I thought you were saying such logic already exists in code yesterday. I can attemp it this weekend and come back to you if I encounter any challenge since I did not commit to rocksdb before. Many thanks for your explaination!

mrambacher:
@XinyuZeng In block.h, make InitializeBase a "protected" function.  Otherwise keep it the same.  In DataBlockIter, make the constructor and Initialize take an InternalKeyComparator (and then fix up the remaining logic as mentioned earlier).

Please submit this as a new PR.  If you can, please report on the performance delta as well.  I will look for it and review it when you submit it.

Thanks for your effort!

Issue #7133, opened 2020-07-15
EventListenerTest.OnFileOperationTest fails on MacOS


mrambacher:
The recently added OnFileOperationTest fails on MacOS.  I do not know if this is related to simulated sleeps or something else, but it consistently fails.

[ RUN      ] EventListenerTest.OnFileOperationTest
db/listener_test.cc:1057: Failure
Expected: (duration.count()) > (0), actual: 0 vs 0

### Steps to reproduce the behavior

make dbg
./listener_test

mrambacher:
I did some further investigation into this test failure.  The OnFileOperationTest fails on MacOS sometimes when checking the result of the Flush operation.  For Posix File Systems, the flush operation is a no-op (does nothing and just returns OK), which will take very little time.

FWIW, When I added debugging, I noticed that if I use "steady_clock" instead of "system_clock", I no longer see duration of 0.

This leads me to a few questions:
- Should we be using system or steady clock for these calculations?
- Should the Listener struct be storing the "duration" or the start/stop time?

riversand963:
Thanks @mrambacher for the investigation. It looks like these tests do not really care about what time it is. It is "duration" that is more meaningful.
@roghnin can you follow up on this?

mrambacher:
In my testing on MacOS, duration can be zero for the Flush operation when using system_clock.

roghnin:
> In my testing on MacOS, duration can be zero for the Flush operation when using system_clock.

This makes sense to me, but I'm having a hard time trying to reproduce this on my machine. Do you happen to have the call stack of the failed assertion?

roghnin:
> In my testing on MacOS, duration can be zero for the Flush operation when using system_clock.

If I got it correctly, the root problem is the behavior of the clock itself. Having the listener store "durations" won't fix the problem, as durations are calculated using timepoints.

I think we need to investigate more about `FileOperationInfo::TimePoint`: it's widely used in RocksDB to keep track of time, and the start and end timestamps are got not just for testing purposes. Maybe we should switch to `steady_clock` in all those cases as it fits better in the role of "stopwatch".

mrambacher:
> > In my testing on MacOS, duration can be zero for the Flush operation when using system_clock.
> 
> If I got it correctly, the root problem is the behavior of the clock itself. Having the listener store "durations" won't fix the problem, as durations are calculated using timepoints. Are you worried that having the readings of `steady_clock` stored in Listener doesn't make much sense?

[...]

PR #7153, opened 2020-07-21
Use steady_clock instead of system_clock in FileOperationInfo::TimePoint

roghnin:
Summary:
Issue #7133 reported that using `system_clock` in `FileOperationInfo::TimePoint` causes the duration of file flush operation (which can be a noop on MacOS in some scenarios) appears to be 0 and fail an assertion in listener_test. Using `steady_clock` supposedly fixed the problem.
`steady_clock` actually fits better into the use cases of `FileOperationInfo::TimePoint` as all usages care about durations but not wall clock time.

Test Plan:
make check.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 ### Public API Change
 * Encryption file classes now exposed for inheritance in env_encryption.h
 * File I/O listener is extended to cover more I/O operations. Now class `EventListener` in listener.h contains new callback functions: `OnFileFlushFinish()`, `OnFileSyncFinish()`, `OnFileRangeSyncFinish()`, `OnFileTruncateFinish()`, and ``OnFileCloseFinish()``.
+* `FileOperationInfo` now reports `duration` measured by `std::chrono::steady_clock` and `start_ts` measured by `std::chrono::system_clock` instead of start and finish timestamps measured by `system_clock`. Note that `system_clock` is called before `steady_clock` in program order at operation starts.
 
 ### Behavior Changes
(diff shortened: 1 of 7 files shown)


roghnin:
> I can confirm that these changes fix the problem I have been experiencing with the OnFileOperationTest.
> 
> The only question I have is if it would be better to store only the duration in the FileOperationInfo callback. Is the user going to care the time at which the operation happened as opposed to how long the operation took to complete?

Well this works for me at least. I'll need to get a third pair of eyes on this as I'm not an expert on interface design. Thanks for the suggestion.

riversand963:
> I can confirm that these changes fix the problem I have been experiencing with the OnFileOperationTest.
> 
> The only question I have is if it would be better to store only the duration in the FileOperationInfo callback. Is the user going to care the time at which the operation happened as opposed to how long the operation took to complete?

System time can be interesting to users if they want to plot a time-series of file access latencies. I think it helps it we know when a latency increase happens.

roghnin:
> > I can confirm that these changes fix the problem I have been experiencing with the OnFileOperationTest.
> > The only question I have is if it would be better to store only the duration in the FileOperationInfo callback. Is the user going to care the time at which the operation happened as opposed to how long the operation took to complete?
> 
> System time can be interesting to users if they want to plot a time-series of file access latencies. I think it helps it we know when a latency increase happens.

We can revert the last commit if we decide recording timestamps is better than just durations.

riversand963:
Can you update HISTORY.md, too? @roghnin

roghnin:
> Can you update HISTORY.md, too? @roghnin

Of course.

PR #7159, opened 2020-07-22
Make max_subcompactions dynamically changeable

jay-zhuang:
Make `max-subcompactions` dynamically changeable by passing the `DBOption` to Compaction.


options/db_options.h
```diff
@@ -116,6 +115,7 @@ struct MutableDBOptions {
   bool strict_bytes_per_sync;
   size_t compaction_readahead_size;
   int max_background_flushes;
+  uint32_t max_subcompactions;
```
mrambacher, line 118: Nit: Can we keep the order of options in the Mutable/Immutable options the same as that in the public DBOptions?  Otherwise it is hard to tell if all of them are covered and where.

PR #7162, opened 2020-07-22
Move CompressionType to its own header file

ltamasi:
Summary:
The patch moves `CompressionType` to its own header file and makes sure
all other public headers include this new header directly, as opposed to
relying on transitive includes or forward declarations.

Test Plan:
`make check`

### DIFF
--- a/include/rocksdb/advanced_options.h
+++ b/include/rocksdb/advanced_options.h
@@ -10,6 +10,7 @@
 
 #include <memory>
 
+#include "rocksdb/compression_type.h"
 #include "rocksdb/memtablerep.h"
 #include "rocksdb/universal_compaction.h"
 
@@ -17,7 +18,6 @@ namespace ROCKSDB_NAMESPACE {
 
 class Slice;
 class SliceTransform;
-enum CompressionType : unsigned char;
 class TablePropertiesCollectorFactory;
 class TableFactory;
 struct Options;
--- a/include/rocksdb/compression_type.h
+++ b/include/rocksdb/compression_type.h
@@ -1,0 +1,40 @@
+// Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#pragma once
+
+#include "rocksdb/rocksdb_namespace.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// DB contents are stored in a set of blocks, each of which holds a
+// sequence of key,value pairs.  Each block may be compressed before
+// being stored in a file.  The following enum describes which
+// compression method (if any) is used to compress a block.
+
+enum CompressionType : unsigned char {
+  // NOTE: do not change the values of existing entries, as these are
+  // part of the persistent format on disk.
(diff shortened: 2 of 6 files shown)


ltamasi:
> There are a bunch of other files (not in the public API) that use CompressionType that are included through a transitive property. Should those be updated as well?

It depends. Not all transitive includes are evil; it is perfectly fine to rely on, say, `foo.h` including `bar.h` when `Foo` exposes a `Bar` in its public interface, which would be the case with `ColumnFamilyOptions` exposing a `CompressionType` member. What's not OK is relying on `foo.h` including `bar.h` when `Bar` is just an implementation detail of `Foo` (or included for an unrelated reason).

Now, the flip side of the above is that when `ColumnFamilyOptions` exposes a `CompressionType` member, it is actually good practice to make the header containing `ColumnFamilyOptions` self-contained by (in this case) including `compression_type.h`, even if one could technically get away with a forward declaration in the header. So we're actually dealing with two different guideline violations here: `advanced_options.h`, `listener.h`, and `leveldb_options.h` are not self-contained, while `convenience.h` relies on a `rocksdb/db.h` including `rocksdb/options.h` (which seems to be done for the purposes of defining `ColumnFamilyDescriptor`).

mrambacher:
Personally, I prefer the forward declaration to the inclusion of files.  I also prefer that each file be more-or-less self-contained, which rocksdb is certainly not.  There are a bunch of places that get access to some underlying class/struct from some distantly included header file.  I also hate code in header files when it forces the inclusion of files that would not otherwise be necessary.  , tRant off.

I was more thinking about files like util/compression.h, which includes options solely for CompressionType (I think), table/format.h, db/blob/blob_index.h, table/block_based/block_based_table_builder.h and maybe flush_job.h.

ltamasi:
> Personally, I prefer the forward declaration to the inclusion of files. I also prefer that each file be more-or-less self-contained, which rocksdb is certainly not. There are a bunch of places that get access to some underlying class/struct from some distantly included header file. I also hate code in header files when it forces the inclusion of files that would not otherwise be necessary. , tRant off.
> 

Sure, forward declarations are a great tool for reducing compile dependencies. In this specific case though, I think they do more harm than good: anybody who actually wants to assign a (valid) value to a `CompressionType` or compare it with any enumerator value needs access to the full definition.

> I was more thinking about files like util/compression.h, which includes options solely for CompressionType (I think), table/format.h, db/blob/blob_index.h, table/block_based/block_based_table_builder.h and maybe flush_job.h.

I'll take a look at those. In general, I do agree that `#includes` in RocksDB are a mess.

ltamasi:
@mrambacher You accidentally edited my previous comment instead of replying so I deleted it to avoid confusion.

I do agree with including `compression_type.h` as opposed to `options.h` wherever possible. I'll try and take a look where we can do that.

mrambacher:
I also wonder if the declaration of CompressionTypeSupported and CompressionTypeToString should be moved into the compression_type.h header.

ltamasi:
> I also wonder if the declaration of CompressionTypeSupported and CompressionTypeToString should be moved into the compression_type.h header.

I'd rather not tack anything else onto this PR.

ltamasi:
P.S. Regarding the use of `#include`s and forward declarations, we might look into automated tools rather than trying to optimize it by hand.

PR #7164, opened 2020-07-22
Define WAL related classes to be used in VersionEdit and VersionSet

cheng-chang:
`WalAddition`, `WalDeletion` are defined in `wal_version.h` and used in `VersionEdit`.
`WalAddition` is used to represent events of creating a new WAL (no size, just log number), or closing a WAL (with size).
`WalDeletion` is used to represent events of deleting or archiving a WAL, it means the WAL is no longer alive (won't be replayed during recovery).

`WalSet` is the set of alive WALs kept in `VersionSet`.

1. Why use `WalDeletion` instead of relying on `MinLogNumber` to identify outdated WALs

On recovery, we can compute `MinLogNumber()` based on the log numbers kept in MANIFEST, any log with number < MinLogNumber can be ignored. So it seems that we don't need to persist `WalDeletion` to MANIFEST, since we can ignore the WALs based on MinLogNumber.

But the `MinLogNumber()` is actually a lower bound, it does not exactly mean that logs starting from MinLogNumber must exist. This is because in a corner case, when a column family is empty and never flushed, its log number is set to the largest log number, but not persisted in MANIFEST. So let's say there are 2 column families, when creating the DB, the first WAL has log number 1, so it's persisted to MANIFEST for both column families. Then CF 0 is empty and never flushed, CF 1 is updated and flushed, so a new WAL with log number 2 is created and persisted to MANIFEST for CF 1. But CF 0's log number in MANIFEST is still 1. So on recovery, MinLogNumber is 1, but since log 1 only contains data for CF 1, and CF 1 is flushed, log 1 might have already been deleted from disk.

We can make `MinLogNumber()` be the exactly minimum log number that must exist, by persisting the most recent log number for empty column families that are not flushed. But if there are N such column families, then every time a new WAL is created, we need to add N records to MANIFEST.

In current design, a record is persisted to MANIFEST only when WAL is created, closed, or deleted/archived, so the number of WAL related records are bounded to 3x number of WALs.

2. Why keep `WalSet` in `VersionSet` instead of applying the `VersionEdit`s to `VersionStorageInfo`

`VersionEdit`s are originally designed to track the addition and deletion of SST files. The SST files are related to column families, each column family has a list of `Version`s, and each `Version` keeps the set of active SST files in `VersionStorageInfo`.

But WALs are a concept of DB, they are not bounded to specific column families. So logically it does not make sense to store WALs in a column family's `Version`s. 
Also, `Version`'s purpose is to keep reference to SST / blob files, so that they are not deleted until there is no version referencing them. But a WAL is deleted regardless of version references.
So we keep the WALs in `VersionSet`  for the purpose of writing out the DB state's snapshot when creating new MANIFESTs.

Test Plan:
make version_edit_test && ./version_edit_test
make wal_edit_test && ./wal_edit_test


db/version_edit.h
```diff
@@ -374,10 +376,31 @@ class VersionEdit {
     return blob_file_garbages_;
   }
 
+  // Add a WAL (either just created or closed).
+  void AddWal(WalNumber number, const WalMetadata& metadata = WalMetadata()) {
```
ltamasi, line 380: You could consider taking `WalMetadata` by value here and `std::move`'ing it in the method body (same below).

db/version_edit_test.cc
```diff
@@ -309,6 +309,102 @@ TEST_F(VersionEditTest, BlobFileAdditionAndGarbage) {
...
+TEST_F(VersionEditTest, AddWalEncodeDecode) {
+  VersionEdit edit;
+  for (uint64_t log_number = 1; log_number <= 20; log_number++) {
+    edit.AddWal(log_number, WalMetadata(rand() % 100));
+  }
+  TestEncodeDecode(edit);
+}
+
+TEST_F(VersionEditTest, AddWalDebug) {
+  const int n = 2;
```
ltamasi, line 321: Minor: this could be

```
constexpr size_t n = 2;
constexpr std::array<uint64_t, n> kLogNumbers{10, 20};
constexpr std::array<uint64_t, n> kBytes{100, 200};
```

(same below for deletes)

db/version_edit_test.cc
```diff
@@ -309,6 +309,102 @@ TEST_F(VersionEditTest, BlobFileAdditionAndGarbage) {
...
+  const std::vector<uint64_t> kLogNumbers{10, 20};
+  const std::vector<uint64_t> kBytes{100, 200};
```
ltamasi, line 323: `kBytes` is a bit misleading, how about `kSizeInBytes` ?

db/wal_version.cc
```diff
@@ -0,0 +1,126 @@
...
+}
+
+void WalSet::AddWals(const WalAdditions& wals) {
+  for (const WalAddition& wal : wals) {
+    AddWal(wal);
+  }
+}
+
+void WalSet::DeleteWal(const WalDeletion& wal) {
+  wals_.erase(wal.GetLogNumber());
```
ltamasi, line 115: This (and `AddWal` above) is where the consistency checks could come into the picture; i.e. we could consider making sure that the WAL was in the set in the first place.

db/wal_version.h
```diff
@@ -0,0 +1,129 @@
...
+ private:
+  // The size of WAL is unknown, used when the WAL is not closed yet.
+  constexpr static uint64_t kUnknownWalSize = 0;
+
+  // Size of a closed WAL in bytes.
+  uint64_t size_bytes_ = kUnknownWalSize;
+};
+
+// Records the event of adding a WAL in VersionEdit.
+class WalAddition {
```
ltamasi, line 48: Question: what's the rationale behind the `WalAddition` / `WalMetadata` split? To me, the file number is just another piece of metadata. Are you concerned about the file number being stored twice (both as the key and as a part of the value) in `WalSet`? (I personally don't think that's a big deal...)

db/wal_version.h
```diff
@@ -0,0 +1,129 @@
...
+ public:
+  WalAddition() : number_(0), metadata_() {}
+
+  explicit WalAddition(WalNumber number) : number_(number), metadata_() {}
+
+  explicit WalAddition(WalNumber number, const WalMetadata& meta)
```
ltamasi, line 54: You could also consider taking `meta` by value here and moving it into `metadata_`.

db/wal_version_test.cc
```diff
@@ -0,0 +1,56 @@
...
+#include "test_util/testharness.h"
+#include "test_util/testutil.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+TEST(WalSet, AddDeleteReset) {
+  WalSet wals;
+  ASSERT_TRUE(wals.GetWals().empty());
+
+  for (WalNumber log_number = 0; log_number < 10; log_number++) {
```
ltamasi, line 19: We should probably number the WALs starting from 1 here since 0 is supposed to be an invalid file number.

db/wal_version.h
```diff
@@ -0,0 +1,129 @@
...
+  void DeleteWal(const WalDeletion& wal);
+  void DeleteWals(const WalDeletions& wals);
+
+  // Resets the internal state.
+  void Reset();
+
+  const std::map<WalNumber, WalMetadata>& GetWals() const { return wals_; }
+
+ private:
+  std::map<WalNumber, WalMetadata> wals_;
```
ltamasi, line 126: We could consider using `unordered_map` unless we need this to be sorted.

CMakeLists.txt
```diff
@@ -607,6 +607,7 @@ set(SOURCES
         db/version_edit_handler.cc
         db/version_set.cc
         db/wal_manager.cc
+        db/wal_version.cc
```
riversand963, line 610: @cheng-chang  As you pointed out in the PR's description, a WAL addition or deletion is db-wise operation, not a mutation to a specific `Version`. Therefore, I wonder whether `wal_version` is the most accurate term.

db/wal_version.h
cheng-chang: need this to be sorted so that we can check the WALs in increasing order of log number.

db/version_edit.h
cheng-chang: there will be a copy anyway, what's the benefit then?

db/version_edit.cc
```diff
@@ -55,6 +55,8 @@ enum Tag : uint32_t {
   kDbId,
   kBlobFileAddition,
   kBlobFileGarbage,
+  kWalAddition,
+  kWalDeletion,
```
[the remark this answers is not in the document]
cheng-chang: Yes, they will be logged during `PurgeObsoleteFiles`.
The consideration here is that, even if a log has been obsolete due to flush, if `PurgeObsoleteFiles` hasn't deleted/archived the log, the log should still be expected to exist on disk. During recovery checking, we check against the on-disk log files.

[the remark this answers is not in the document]
cheng-chang: The idea is that the version edits related to WAL track the state of the on-disk logs, instead of the logical active WALs. The purpose of the tracked on-disk WAL state is to check the on-disk logs during recovery.

db/version_edit.cc
```diff
@@ -55,6 +55,8 @@ enum Tag : uint32_t {
   kDbId,
   kBlobFileAddition,
   kBlobFileGarbage,
+  kWalAddition,
```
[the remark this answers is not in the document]
cheng-chang: The WAL related edits are designed to track on on-disk WAL states. So WalAddition means there is a WAL on-disk, the WalAddition may have optional size / checksum, it's not related to logical WAL operations.

On consistency checks, the WAL related edits are designed to verify on-disk WALs on recovery, the logical checks such as must be created before being closed are not the design purpose.

CMakeLists.txt
cheng-chang: `wal_manifest` is more appropriate, but the `manifest` concept is named `version` in the codebase, so I prefer to keep it as `wal_version` for now.

db/version_edit.h
```diff
@@ -374,10 +376,31 @@ class VersionEdit {
...
+    wal_additions_.emplace_back(number, metadata);
+  }
+
+  // Retrieve all the added WALs.
+  const WalAdditions& GetWalAdditions() const { return wal_additions_; }
+
+  // VersionEdits with WAL addition shouldn't contain other types of edits.
+  bool IsWalAddition() const { return !wal_additions_.empty(); }
```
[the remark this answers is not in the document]
cheng-chang: As the comment points out, there shouldn't be WalAddition and WalDeletion in the same VersionEdit simultaneously. So if wal_additions_ is not empty, then IsWalAddition.

db/wal_version.h
cheng-chang: During recovery check, we need to iterate through `wals_` in increasing order of log number.

db/wal_version.cc
cheng-chang: I don't want to pay the cost of checking during runtime, I added an assert here.

db/wal_version.h
cheng-chang: `WalMetadata` is the metadata of a WAL file, it's like file stat. Two different log files' metadata may be equal even if their log numbers are different.
`WalAddition` is the event of adding a WAL on disk, currently it's just a log number + `WalMetadata`, but we may also add other information such as timestamp (when the event happened).

cheng-chang:
@ltamasi @riversand963 Thanks for review, PTAL.

CMakeLists.txt
riversand963: How about `wal_mutation` or `wal_addition` and `wal_deletion` or something like that, avoiding `version`. I'll leave this up to you.

db/version_edit.h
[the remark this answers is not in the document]
riversand963: I see. Maybe add an assertion for that.

db/wal_version.cc
riversand963: The problem with assert is that it may be a no-op if `NDEBUG`. It is still helpful to verify consistency.

CMakeLists.txt
cheng-chang: I renamed it to wal_edit

db/wal_version.cc
cheng-chang: I now think the consistency check is valuable in runtime, especially when replaying the MANIFEST.
So I added runtime consistency checks.

db/version_edit.h
```diff
@@ -374,10 +376,45 @@ class VersionEdit {
...
+  // Retrieve all the deleted WALs.
+  const WalDeletions& GetWalDeletions() const { return wal_deletions_; }
+
+  // VersionEdits with WAL deletion shouldn't contain other types of edits.
+  bool IsWalDeletion() const {
+#ifndef NDEBUG
+    if (!wal_deletions_.empty()) {
+      assert(NumEntries() == wal_deletions_.size());
+    }
+#endif
```
ltamasi, line 409: This (and the one above) could be simplified to `assert(wal_deletions_.empty() || NumEntries() == wal_deletions_.size());`. Then again, I'm wondering whether we should do this in the first place, as opposed to say, making the `WalDeletion` part of the `VersionEdit` of the flush job that makes the WAL file in question obsolete. (Note: this ties back to the question above about what we want to track.)

db/wal_edit.h
```diff
@@ -0,0 +1,134 @@
...
+};
+
+// Records the event of adding a WAL in VersionEdit.
+class WalAddition {
+ public:
+  WalAddition() : number_(0), metadata_() {}
+
+  explicit WalAddition(WalNumber number) : number_(number), metadata_() {}
+
+  explicit WalAddition(WalNumber number, WalMetadata meta)
```
ltamasi, line 54: Nit: no need for this one to be explicit since it can't be called with a single argument.

db/wal_edit_test.cc
```diff
@@ -0,0 +1,127 @@
...
+  wals.Reset();
+  ASSERT_TRUE(wals.GetWals().empty());
+}
+
+TEST(WalSet, Overwrite) {
+  constexpr WalNumber kNumber = 100;
+  constexpr uint64_t kBytes = 200;
+  WalSet wals;
+  wals.AddWal(WalAddition(kNumber));
+  ASSERT_TRUE(!wals.GetWals().at(kNumber).HasSize());
```
ltamasi, line 52: Nit: you could use `ASSERT_FALSE` and ditch the negation here.

db/wal_edit.cc
```diff
@@ -0,0 +1,184 @@
...
+  return os;
+}
+
+std::string WalDeletion::DebugString() const {
+  std::ostringstream oss;
+  oss << *this;
+  return oss.str();
+}
+
+Status WalSet::AddWal(const WalAddition& wal) {
```
ltamasi, line 119: Minor: since `wals_` is an `std::map`, the logic here could be optimized using `lower_bound` and the variant of `insert` that takes a hint.

db/wal_edit.cc
```diff
@@ -0,0 +1,184 @@
...
+  for (const WalAddition& wal : wals) {
+    s = AddWal(wal);
+    if (!s.ok()) {
+      break;
+    }
+  }
+  return s;
+}
+
+Status WalSet::DeleteWal(const WalDeletion& wal) {
```
ltamasi, line 155: This could also be optimized by storing the iterator to eliminate repeated lookups for the same key.

db/version_edit_test.cc
```diff
@@ -309,6 +309,107 @@ TEST_F(VersionEditTest, BlobFileAdditionAndGarbage) {
   TestEncodeDecode(edit);
 }
 
+TEST_F(VersionEditTest, AddWalEncodeDecode) {
```
ltamasi, line 312: We should probably also add unit tests for the decoding failure cases.

db/version_edit.cc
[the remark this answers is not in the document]
cheng-chang: Tracking the logically obsolete WAL files is the same as tracking the `MinLogNumber` I mentioned in the PR summary. This `MinLogNumber` is already tracked in MANIFEST during every flush. But for unflushed empty column families, their min log number is always the first log number, so `MinLogNumber` is only a lower bound of the logs that should be kept alive.

I prefer viewing the WAL as the source of truth for logical data including the in-memory unflushed data, and MANIFEST as  the source of truth for on-disk files. For example, MANIFEST tracks SST files. In this PR, we make it track WAL files. In future PRs, maybe we let it track option files.

db/wal_version.h
cheng-chang: going to make this unordered_map, then sort the log numbers during recovery.

db/wal_edit.cc
cheng-chang: changed to unordered_map

db/wal_edit.cc
cheng-chang: changed to unordered_map

db/wal_version.h
cheng-chang: also want to write the logs in increasing order when creating manifest snapshot.

db/version_edit.cc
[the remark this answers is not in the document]
ltamasi: > Tracking the logically obsolete WAL files is the same as tracking the `MinLogNumber` I mentioned in the PR summary. This `MinLogNumber` is already tracked in MANIFEST during every flush. But for unflushed empty column families, their min log number is always the first log number, so `MinLogNumber` is only a lower bound of the logs that should be kept alive.
> 

Hm, I don't think they're the same thing. I do understand the corner case with empty column families. My point is that we already have a way (used in `FindObsoleteFiles`) of identifying the actual set of obsolete WAL files, as opposed to just a soft lower bound, right? IIUC, your plan is to log a dedicated `VersionEdit` (with `DeleteWal`) at that point. What I'm suggesting is moving this up in time a bit and piggybacking that same `DeleteWal` on the `VersionEdit` that is logged for the (last) flush that makes the WAL file obsolete. Please let me know if I'm missing something.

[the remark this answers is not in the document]
cheng-chang: Are you suggesting piggybacking the WalDeletion in https://github.com/facebook/rocksdb/blob/master/db/flush_job.cc#L187 ?
Yes, since db mutex is held when entering this method, we can delete WALs in WalSet whose log number < GetNextLogNumber. One complexity here is that in 2pc mode, this is not the logic for determining min log number, see https://github.com/facebook/rocksdb/blob/master/db/db_impl/db_impl_files.cc#L25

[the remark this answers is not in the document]
ltamasi: Yes, making the `DeleteWal`s a part of the `FlushJob`'s `VersionEdit` is what I was thinking. Of course, we would need to make sure that both the non-2PC and the 2PC case are handled consistently with what `FindObsoleteFiles` does.

db/wal_edit.cc
```diff
@@ -0,0 +1,175 @@
...
+    }
+    it->second = wal.GetMetadata();
+  } else {
+    // The WAL must not exist beforehand.
+    if (it != wals_.end() && it->first == wal.GetLogNumber()) {
+      std::stringstream ss;
+      ss << "WAL " << wal.GetLogNumber() << " is created more than once";
+      return Status::Corruption("WalSet", ss.str());
+    }
+    wals_[wal.GetLogNumber()] = wal.GetMetadata();
```
ltamasi, line 129: This is where you could use the hint form of `insert` or `emplace_hint`.

db/wal_edit.cc
```diff
@@ -0,0 +1,175 @@
...
+    s = AddWal(wal);
+    if (!s.ok()) {
+      break;
+    }
+  }
+  return s;
+}
+
+Status WalSet::DeleteWal(const WalDeletion& wal) {
+  auto it = wals_.lower_bound(wal.GetLogNumber());
```
ltamasi, line 146: `find` is sufficient here, i.e. no need for `lower_bound` since we're not inserting.

db/version_edit.cc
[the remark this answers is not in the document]
cheng-chang: Since this PR does not implement wal deletion anyway, we can continue this discussion in a future PR dedicated to wal deletion. I'll try to make logical wal deletions in flush jobs. In this PR, I removed assumption WalDeletion cannot be part of other version edits. See the last commit.

db/wal_edit_test.cc
```diff
@@ -0,0 +1,127 @@
...
+  constexpr WalNumber kNonExistingNumber = 100;
+  WalSet wals;
+  Status s = wals.DeleteWal(WalDeletion(kNonExistingNumber));
+  ASSERT_TRUE(s.IsCorruption());
+  ASSERT_TRUE(s.ToString().find("WAL 100 must exist before deletion") !=
+              std::string::npos);
+}
+
+TEST(WalSet, DeleteNonClosedWal) {
+  constexpr WalNumber kNonExistingNumber = 100;
```
ltamasi, line 112: Nit: `kNonExistingNumber` is not the best name in this specific test case.

PR #7165, opened 2020-07-22
Compaction Read/Write Stats by Compaction Type

akabcenell:
Summary: Adds compaction statistics (total bytes read and written) for compactions that occur for delete-triggered, periodic, and TTL compaction reasons.

Test Plan: TTL and periodic can be checked by runnning db_bench with the options activated:

/db_bench --benchmarks="fillrandom,stats" --statistics --num=10000000 -base_background_compactions=16 -periodic_compaction_seconds=1
./db_bench --benchmarks="fillrandom,stats" --statistics --num=10000000 -base_background_compactions=16 -fifo_compaction_ttl=1

Setting the time to one second causes non-zero bytes read/written for those compaction reasons. Disabling them or setting them to times longer than the test run length causes the stats to return to zero as expected.

Delete-triggered compaction counting is tested in DBTablePropertiesTest.DeletionTriggeredCompactionMarking


akabcenell:
@ajkr we should also add a Java expert to confirm the Enum numbering conventions

db/compaction/compaction_job.cc
```diff
@@ -1466,10 +1466,24 @@ Status CompactionJob::InstallCompactionResults(
 
 void CompactionJob::RecordCompactionIOStats() {
   RecordTick(stats_, COMPACT_READ_BYTES, IOSTATS(bytes_read));
+  RecordTick(stats_, COMPACT_WRITE_BYTES, IOSTATS(bytes_written));
+  CompactionReason compaction_reason =
+      compact_->compaction->compaction_reason();
+  if (compaction_reason == CompactionReason::kFilesMarkedForCompaction) {
+    RecordTick(stats_, COMPACT_READ_BYTES_DELETE_TRIGGERED,
+               IOSTATS(bytes_read));
+    RecordTick(stats_, COMPACT_WRITE_BYTES_DELETE_TRIGGERED,
```
ajkr, line 1475: Can we call these `COMPACT_{READ,WRITE}_BYTES_MARKED`? I know we all have been calling it delete-triggered compaction, but actually a deletion-based trigger is just one of the pluggable ways that files can be marked for compaction.

monitoring/statistics.cc
```diff
@@ -105,6 +105,14 @@ const std::vector<std::pair<Tickers, std::string>> TickersNameMap = {
     {COMPACT_READ_BYTES, "rocksdb.compact.read.bytes"},
     {COMPACT_WRITE_BYTES, "rocksdb.compact.write.bytes"},
     {FLUSH_WRITE_BYTES, "rocksdb.flush.write.bytes"},
+    {COMPACT_READ_BYTES_DELETE_TRIGGERED,
+     "rocksdb.compact.read.delete_triggered.bytes"},
```
ajkr, line 109: Same, how about `rocksdb.compact.{read,write}.marked.bytes`

PR #7166, opened 2020-07-23
Avoid converting MERGES to PUTS when allow_ingest_behind is true

### LINKED ISSUE #6490: prevent MERGE->PUT conversion when `allow_ingest_behind=true`

### Expected behavior

Compaction must not convert MERGEs to PUTs when `allow_ingest_behind = true`. A converted PUT can incorrectly cover a future base value or merge ingested-behind into the bottom level.

### Actual behavior

I do not see anything preventing it. For example, see how the seqnum zeroing optimization is explicitly disabled when `allow_ingest_behind = true`: https://github.com/facebook/rocksdb/blob/e171a219d592b1bda9b654db126f04d4d834a815/db/compaction/compaction_iterator.cc#L672.

We should first write a test case to verify, though.

hiepd:
Summary:
- Closes #6490
- Currently MERGEs are converted to PUTs at bottom or compaction has reached the beginning of the key, this can wrongly cover a PUT future base case.

Test Plan:
- Automated: `make all check`
- Manual: With `allow_ingest_behind = true`, add Merge operations to a key then run compaction. Then run ingesting external files to make sure the base case is probably compacted with existing Merges.

### DIFF
--- a/db/compaction/compaction_iterator.cc
+++ b/db/compaction/compaction_iterator.cc
@@ -86,8 +86,10 @@ CompactionIterator::CompactionIterator(
       info_log_(info_log) {
   assert(compaction_filter_ == nullptr || compaction_ != nullptr);
   assert(snapshots_ != nullptr);
-  bottommost_level_ =
-      compaction_ == nullptr ? false : compaction_->bottommost_level();
+  bottommost_level_ = compaction_ == nullptr
+                          ? false
+                          : compaction_->bottommost_level() &&
+                                !compaction_->allow_ingest_behind();
   if (compaction_ != nullptr) {
     level_ptrs_ = std::vector<size_t>(compaction_->number_levels(), 0);
   }
--- a/db/compaction/compaction_iterator_test.cc
+++ b/db/compaction/compaction_iterator_test.cc
@@ -169,13 +169,15 @@ class FakeCompaction : public CompactionIterator::CompactionProxy {
   Slice GetLargestUserKey() const override {
     return "\xff\xff\xff\xff\xff\xff\xff\xff\xff";
   }
-  bool allow_ingest_behind() const override { return false; }
+  bool allow_ingest_behind() const override { return is_allow_ingest_behind; }
 
   bool preserve_deletes() const override { return false; }
 
   bool key_not_exists_beyond_output_level = false;
 
   bool is_bottommost_level = false;
+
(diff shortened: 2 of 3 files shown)


db/compaction/compaction_iterator.cc
```diff
@@ -592,8 +592,11 @@ void CompactionIterator::NextFromInput() {
       // have hit (A)
       // We encapsulate the merge related state machine in a different
       // object to minimize change to the existing flow.
-      Status s = merge_helper_->MergeUntil(input_, range_del_agg_,
-                                           prev_snapshot, bottommost_level_);
+      auto allow_ingest_behind =
```
hiepd, line 595: @ajkr I'm trying to access `allow_ingest_behind` options from `compaction proxy` here in `compaction iterator` but it's not always present. Do you think this is a viable approach? Thanks!

ajkr: It looks reasonable as the proxy should be non-nullptr whenever the `CompactionIterator` is used for compaction. `CompactionIterator` is also used for flush, in which case proxy can be nullptr, but MERGE->PUT conversion doesn't happen there.

Although, the new code is a bit confusing when it has a variable `allow_ingest_behind == false` during a flush even though the DB may be configured with `allow_ingest_behind == true`.

How about folding this logic into the computation for `bottommost_level_`? It is reasonable to say the `CompactionIterator` is not targeting the bottommost level whenever the DB has `allow_ingest_behind` is set, since we have no way to guarantee something isn't ingested under the compaction output file later.

hiepd:
@ajkr Thanks for the review! I have changed to the approach you suggested.

db/compaction/compaction_iterator_test.cc
```diff
@@ -968,6 +985,45 @@ TEST_F(CompactionIteratorWithSnapshotCheckerTest, CompactionFilter_FullMerge) {
       compaction_filter.get());
 }
 
+// Tests how CompactionIterator work together with AllowIngestBehind.
+class CompactionIteratorWithAllowIngestBehindTest
+    : public CompactionIteratorTest {
+ public:
+  bool AllowIngestBehind() const override { return true; }
```
ajkr, line 992: It seems more conventional to add an `allow_ingest_behind` parameter to `RunTest()` and `InitIterators()`. But we can take this either way.

PR #7168, opened 2020-07-23
export stats_persist_period_sec

developerfred:
fixed
 - https://github.com/rust-rocksdb/rust-rocksdb/issues/447 
 -  https://github.com/rust-rocksdb/rust-rocksdb/pull/448

### DIFF
--- a/db/c_test.c
+++ b/db/c_test.c
@@ -1672,6 +1672,9 @@ int main(int argc, char** argv) {
     rocksdb_options_set_stats_dump_period_sec(o, 18);
     CheckCondition(18 == rocksdb_options_get_stats_dump_period_sec(o));
 
+    rocksdb_options_set_stats_persist_period_sec(o, 5);
+    CheckCondition(5 == rocksdb_options_get_stats_persist_period_sec(o));
+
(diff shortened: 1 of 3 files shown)


DarkEld3r:
You should add an implementation for these functions in the `db/c.cc` file and probably tests in the `db/c_test.c` file.

developerfred:
> You should add an implementation for these functions in the `db/c.cc` file and probably tests in the `db/c_test.c` file.

@DarkEld3r Thanks I will do!

db/c_test.c
```diff
@@ -1825,6 +1828,7 @@ int main(int argc, char** argv) {
     CheckCondition(1 == rocksdb_options_get_is_fd_close_on_exec(copy));
     CheckCondition(1 == rocksdb_options_get_skip_log_error_on_recovery(copy));
     CheckCondition(18 == rocksdb_options_get_stats_dump_period_sec(copy));
+    CheckCondition(600 == rocksdb_options_get_stats_persist_period_sec(copy));
```
DarkEld3r, line 1831: I suppose there should be 5 here, not 600.

developerfred: done

PR #7172, opened 2020-07-24
Remove redundant ROCKSDB_LITE check

jay-zhuang:
It's already inside of a `#ifdef ROCKSDB_LITE` block.

### DIFF
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -1284,7 +1284,6 @@ Status StressTest::TestBackupRestore(
   return s;
 }
 
-#ifndef ROCKSDB_LITE
 Status StressTest::TestApproximateSize(
     ThreadState* thread, uint64_t iteration,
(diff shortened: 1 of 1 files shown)


jay-zhuang:
> Thanks! (Btw, the comments at [#L1202](https://github.com/facebook/rocksdb/blob/master/db_stress_tool/db_stress_test_base.cc#L1202) and [#L1538](https://github.com/facebook/rocksdb/blob/master/db_stress_tool/db_stress_test_base.cc#L1538) are also confusing :P)

The comments are mapping to the name of the `#if` check, like:
```
#ifdef LITE
...
#else // LITE
...
#endif // LITE
```
another example:
https://github.com/facebook/rocksdb/blob/00de699096d88af0a796cc8e1c6c306bc031c1fb/db/db_impl/db_impl_secondary.cc#L656-L672

gg814:
Ah, I see. Now I learned it, thanks! (although I still feel it's confusing >_<)

PR #7175, opened 2020-07-27
Makefile allow ingest user's extra/plugin source files

rockeet:
User may need to ingest some extra/plugin source files to rocksdb, with this pull request, this work becomes trivially:
```bash
# for example build shared_lib
env LIB_SOURCES="my_extra.cc my_plugin.cc" make shared_lib
```


ajkr:
Sorry this got automatically closed. We do support this now. See https://github.com/facebook/rocksdb/tree/main/plugin.

rockeet:
> Sorry this got automatically closed. We do support this now. See https://github.com/facebook/rocksdb/tree/main/plugin.

Very Thanks!

PR #7178, opened 2020-07-27
ReadOptions.iter_start_ts should support tombstones

riversand963:
Summary: as title.
When ReadOptions.iter_start_ts is not nullptr, DBIter::key() should
return internal keys including value type.

Test Plan:
make check

### DIFF
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -243,8 +243,8 @@ TEST_F(DBBasicTestWithTimestamp, SimpleForwardIterate) {
 }
 
 TEST_F(DBBasicTestWithTimestamp, SimpleForwardIterateLowerTsBound) {
-  const int kNumKeysPerFile = 128;
-  const uint64_t kMaxKey = 1024;
+  constexpr int kNumKeysPerFile = 128;
+  constexpr uint64_t kMaxKey = 1024;
   Options options = CurrentOptions();
   options.env = env_;
   options.create_if_missing = true;
@@ -314,6 +314,7 @@ TEST_F(DBBasicTestWithTimestamp, SimpleForwardIterateLowerTsBound) {
     for (it->Seek(Key1(0)), key = 0; it->Valid(); it->Next(), ++count, ++key) {
       CheckIterEntry(it.get(), Key1(key), kTypeDeletionWithTimestamp, Slice(),
                      write_ts);
+      // Skip key@ts=3 and land on tombstone key@ts=5
       it->Next();
     }
     ASSERT_EQ(kMaxKey + 1, count);
@@ -374,9 +375,11 @@ TEST_F(DBBasicTestWithTimestamp, ForwardIterateStartSeqnum) {
     SequenceNumber expected_seq = start_seqs[i] + (kMaxKey - kMinKey) + 1;
     uint64_t key = kMinKey;
     for (iter->Seek(Key1(kMinKey)); iter->Valid(); iter->Next()) {
-      CheckIterEntry(iter.get(), Key1(key), expected_seq,
-                     (key % 2) ? kTypeValue : kTypeDeletionWithTimestamp,
-                     "value" + std::to_string(i + 1), write_ts_list[i + 1]);
+      CheckIterEntry(
+          iter.get(), Key1(key), expected_seq,
+          (key % 2) ? kTypeValue : kTypeDeletionWithTimestamp,
+          (key % 2) ? "value" + std::to_string(i + 1) : std::string(),
+          write_ts_list[i + 1]);
       ++key;
       --expected_seq;
     }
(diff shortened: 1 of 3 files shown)


starsiriusa:
InternalKey hasn't been exposed so far. ParseInternalKey() and ParsedInternalKey are not available for application to consume.
Or there is already interface for it?

riversand963:
> InternalKey hasn't been exposed so far. ParseInternalKey() and ParsedInternalKey are not available for application to consume.
> Or there is already interface for it?

The C++ types `InternalKey` or `ParsedInternalKey` are not exposed. Neither is `ParseInternalKey()` function. However, RocksDB has been able to return the internal key as a `Slice` back to application. For example, if `ReadOptions.iter_start_seqnum == true`, then internal keys will be returned as `Slice`. The motivation is that some applications do want to list all versions of data including tombstones.

starsiriusa:
I guess any application that is taking advantage of iter_start_seqnum_ is also assuming knowledge of internal key format. Not sure if it was intentional, but would be nice if support exists in the interface without internal structure being exposed.

riversand963:
@starsiriusa , actually in include/rocksdb/types.h, we have `ParseFullKey(const Slice& internal_key, FullKey* fkey)` exposed. `FullKey` can be found in rocksdb/types.h.
So the internal key is somewhat exposed via `FullKey`, though `rocksdb/types.h` may not be super up-to-date.

hliu18:
Yes ParseFullKey() does the job.

hliu18:
Maybe add a new type for GetEntryType(kTypeDeletionWithTimestamp)?

riversand963:
> Maybe add a new type for GetEntryType(kTypeDeletionWithTimestamp)?

SG. Will do tmr.

db/db_with_timestamp_basic_test.cc
```diff
@@ -296,34 +348,60 @@ TEST_F(DBBasicTestWithTimestamp, ForwardIterateStartSeqnum) {
...
+    read_opts.iter_start_seqnum = start_seqs[i] + 1;
     std::unique_ptr<Iterator> iter(db_->NewIterator(read_opts));
-    SequenceNumber expected_seq = start_seqs[i] + 1;
+    SequenceNumber expected_seq = start_seqs[i] + (kMaxKey - kMinKey) + 1;
     uint64_t key = kMinKey;
     for (iter->Seek(Key1(kMinKey)); iter->Valid(); iter->Next()) {
-      CheckIterEntry(iter.get(), Key1(key), expected_seq, kTypeValue,
+      CheckIterEntry(iter.get(), Key1(key), expected_seq,
+                     (key % 2) ? kTypeValue : kTypeDeletionWithTimestamp,
+                     "value" + std::to_string(i + 1), write_ts_list[i + 1]);
```
ltamasi, line 379: Minor: we could avoid passing `expected_value` if it's a deletion.

db/db_with_timestamp_basic_test.cc
```diff
@@ -296,34 +348,60 @@ TEST_F(DBBasicTestWithTimestamp, ForwardIterateStartSeqnum) {
...
+  }
+  // Scan with both read_opts.iter_start_seqnum and read_opts.iter_start_ts set.
+  std::vector<std::string> read_ts_lb_list;
+  for (int t = 0; t < kNumTimestamps - 1; ++t) {
+    read_ts_lb_list.push_back(Timestamp(2 * t, /*do not care*/ 17));
+  }
+  for (size_t i = 0; i < read_ts_list.size(); ++i) {
+    Slice read_ts = read_ts_list[i];
+    Slice read_ts_lb = read_ts_lb_list[i];
+    (void)read_ts_lb;
```
ltamasi, line 392: Nit: probably left over/unnecessary line?

riversand963: Sorry about this. Will remove.

db/db_with_timestamp_basic_test.cc
```diff
@@ -255,17 +278,46 @@ TEST_F(DBBasicTestWithTimestamp, SimpleForwardIterateLowerTsBound) {
...
+    std::string read_timestamp_lb = Timestamp(2, 0);
+    Slice read_ts_lb = read_timestamp_lb;
+    read_opts.iter_start_ts = &read_ts_lb;
+    std::unique_ptr<Iterator> it(db_->NewIterator(read_opts));
+    int count = 0;
+    uint64_t key = 0;
+    for (it->Seek(Key1(0)), key = 0; it->Valid(); it->Next(), ++count, ++key) {
+      CheckIterEntry(it.get(), Key1(key), kTypeDeletionWithTimestamp, Slice(),
+                     write_ts);
+      it->Next();
```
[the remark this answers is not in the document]
riversand963: Correct.

db/db_with_timestamp_basic_test.cc
riversand963: Do you suggest adding an overloaded version of `CheckIterEntry` that takes one fewer argument?

ltamasi: That would also work but what I had in mind originally was to avoid creating the value string for deletes (since it's unused by the callee), i.e. something along the lines of `(key % 2) ? ("value" + std::to_string(i + 1)) : std::string()`.

riversand963: SG. Will update. Thanks!

PR #7180, opened 2020-07-28
[IOTracing PR1] Store FileSystemPtr object that contains FileSystem ptr

akankshamahajan15:
Summary: As part of the IOTracing project, this PR
    1. Caches "FileSystemPtr" object(wrapper class that returns file system pointer based on tracing enabled) instead of "FileSystem" pointer.
    2. FileSystemPtr object is created using FileSystem pointer and IOTracer
    pointer.
    3. IOTracer shared_ptr is created in DBImpl and it is passed to different classes through constructor.
    4. When tracing is enabled through DB::StartIOTrace, FileSystemPtr
    returns FileSystemTracingWrapper pointer for tracing purpose and when
    it is disabled underlying FileSystem pointer is returned.

Test Plan:  make check -j64

    Reviewers:

    Subscribers:

    Tasks:

    Tags:

    Tags:

### DIFF
--- a/db/compaction/compaction_job.cc
+++ b/db/compaction/compaction_job.cc
@@ -328,8 +328,9 @@ CompactionJob::CompactionJob(
     const SnapshotChecker* snapshot_checker, std::shared_ptr<Cache> table_cache,
     EventLogger* event_logger, bool paranoid_file_checks, bool measure_io_stats,
     const std::string& dbname, CompactionJobStats* compaction_job_stats,
-    Env::Priority thread_pri, const std::atomic<bool>* manual_compaction_paused,
-    const std::string& db_id, const std::string& db_session_id)
+    Env::Priority thread_pri, const std::shared_ptr<IOTracer>& io_tracer,
+    const std::atomic<bool>* manual_compaction_paused, const std::string& db_id,
+    const std::string& db_session_id)
     : job_id_(job_id),
       compact_(new CompactionState(compaction)),
       compaction_job_stats_(compaction_job_stats),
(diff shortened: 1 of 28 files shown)


akankshamahajan15:
Passed IOTracer as argument in existing constructor instead of writing a new one as list of arguments was long in the constructors and BlockCacheTrace was already passed as nullptr in many test cases so I passed IOTracer=nullptr as well in those testcases.

db/compaction/compaction_job.cc
```diff
@@ -1564,7 +1565,8 @@ Status CompactionJob::OpenCompactionOutputFile(
                            &syncpoint_arg);
 #endif
   Status s;
-  IOStatus io_s = NewWritableFile(fs_, fname, &writable_file, file_options_);
+  IOStatus io_s = NewWritableFile(db_options_.fs.get(), fname, &writable_file,
```
anand1976, line 1568: Shouldn't we pass the right pointer from ```fs_``` instead of ```db_options_.fs```? Otherwise, this won't get traced.

env/file_system_tracer.h
```diff
@@ -80,6 +79,9 @@ class FileSystemPtr {
     }
   }
 
+  /* Returns the underlying File System pointer */
+  FileSystem* get() const { return fs_.get(); }
```
anand1976, line 83: This should be conditional on whether tracing is on or off. If its on, it should return ```fs_tracer_.get()```.

db/compaction/compaction_job.cc
akankshamahajan15: Right now, this is not getting traced.  But since its a one line function to trace FileSystem::NewWritableFile so I will pass the FileSystemPtr pointer.

akankshamahajan15:
Addressed Comments

PR #7181, opened 2020-07-28
Makefile support for link-time optimization

ajkr:
`USE_LTO=1` in `make` commands now enables LTO. The archiver (`ar`) needed
to change in this PR to use a wrapper that enables the LTO plugin.

Test Plan:

build a few ways
```
$ make clean && USE_LTO=1 make -j48 db_bench
$ make clean && USE_CLANG=1 USE_LTO=1 make -j48 db_bench
$ make clean && ROCKSDB_NO_FBCODE=1 USE_LTO=1 make -j48 db_bench
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,6 +192,11 @@ ifeq ($(filter -DROCKSDB_ASSERT_STATUS_CHECKED,$(OPT)),)
 endif
 endif
 
+ifeq ($(USE_LTO), 1)
+	CXXFLAGS += -flto
+	LDFLAGS += -flto -fuse-linker-plugin
+endif
+
 $(warning Warning: Compiling in debug mode. Don't use the resulting binary in production)
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -195,6 +195,12 @@ endif
 $(warning Warning: Compiling in debug mode. Don't use the resulting binary in production)
 endif
 
+# `USE_LTO=1` enables link-time optimizations. Among other things, this enables
+# more devirtualization opportunities and inlining across translation units.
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -195,6 +195,11 @@ endif
 $(warning Warning: Compiling in debug mode. Don't use the resulting binary in production)
 endif
 
+ifeq ($(USE_LTO), 1)
```
mrambacher, line 198: Please add a comment on what USE_LTO means.

Also, should something similar be put into the CMakeLists.txt?  It seems like the two build tools should support similar options as much as possible

ajkr: > Also, should something similar be put into the CMakeLists.txt? It seems like the two build tools should support similar options as much as possible

Agreed. I looked into it a bit and found cmake provides a standardized way to build with LTO. Something like:

```
$ cmake .. -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=True -DCMAKE_POLICY_DEFAULT_CMP0069=NEW -DCMAKE_BUILD_TYPE=Release
```

The `-DCMAKE_POLICY_DEFAULT_CMP0069=NEW` could be elided if we bump the value in `cmake_minimum_required()` to 3.9. Also on my machine the link step currently fails. These require some more investigation, so I'll probably leave this PR's scope as adding support in Makefile only for now.

ajkr: > Please add a comment on what USE_LTO means.

Done.

mrambacher: Can you please capture what you found in an issue for CMake so that we do not lose it.

ajkr: Sure, here it is: #7186.

PR #7189, opened 2020-07-28
Add stress tests for verifying checksums of SST files

gg814:
Summary:
The new stress tests inject random bit-flips in the files and try to detect the corruption.

Note: The tests are likely to hit the `assert()` in the codes since the corruption is random. So these tests are better compiled with `DEBUG_LEVEL=0`.

Test plan:
`DEBUG_LEVEL=0 make db_stress`
`./db_stress -inject_corruption_and_verify_checksum=1`
`./db_stress -inject_corruption_and_verify_checksum_one_in=1000`
`./db_stress -checksum_type=kxxHash64 -inject_corruption_and_verify_checksum_one_in=5000 -inject_corruption_and_verify_checksum=1`
`./db_stress -checksum_type=kxxHash -inject_corruption_and_verify_checksum_one_in=5000 -inject_corruption_and_verify_checksum=1`


pdillinger:
> Note: The corruption injected by the tests are likely to cause the assert() in the codebase to fail. So these tests are better compiled with DEBUG_LEVEL=0.

Ugh. This is unfortunate. We generally run the stress/crash test with DEBUG_LEVEL=1. If anyone else has ideas on how to resolve this, speak up!

An alternative idea I had for verifying completeness of VerifyChecksum was to instrument the code (ifndef NDEBUG) such that it would track which offset ranges within the file it had verified block checksums for, and fail assertion if there are any gaps, before the footer (currently unchecked). As long as db_stress sometimes called VerifyChecksum, such an assertion should check for completeness.

pdillinger:
cc @anand1976 if he has ideas

gg814:
Adding some context: if `DEBUG_LEVEL` is not zero, the test may hit the following `assert()`:
```
db_stress: table/block_based/data_block_hash_index.cc:79: void rocksdb::DataBlockHashIndex::Initialize(const char*, uint16_t, uint16_t*): Assertion `num_buckets_ > 0' failed.
```
if `DEBUG_LEVEL=0`, then the above case will be reported as checksum mismatch in the property block (and thus, the corruption is detected).

anand1976:
@gg814 Is the ```assert()``` hit due to the data block checksum not being verified in some cases? Did you also hit the assert when running the crash tests using ```db_crashtest.py```? I noticed that ```db_crashtest.py``` forces ```ReadOptions::verify_checksum``` to ```true```, whereas ```db_stress``` defaults it to ```false```.

gg814:
> @gg814 Is the `assert()` hit due to the data block checksum not being verified in some cases? 

@anand1976 In the case I mentioned above, it is hit because the injected bit-flips happen to be in somewhere in the property block and also by coincident it results in failing the corresponding `assert()` that uses the corrupted information.

I remember I also saw other cases where the `assert()` failed in some other parts of the SST file because of the injected bit-flips.

I also manually tested the cases where `assert()` failed by using `sst_dump` (not `./sst_dump`) which compiled using `DEBUG_LEVEL=0`, and `sst_dump` successfully reported the corruption.

> Did you also hit the assert when running the crash tests using `db_crashtest.py`? 

I didn't run `make crash_test` with nonzero `DEBUG_LEVEL`. 
(Actually, I ran `DEBUG_LEVEL=0 make crash_test` with the new stress test option and it has been running for at least 3 days in my dev server with 24 cores... All the blackbox tests are passed and one whitebox test is passed but not sure when it is going to finish the whitebox tests... >_<)

pdillinger:
> An alternative idea I had for verifying completeness of VerifyChecksum ...

The big hole in this proposed check is that it does not verify that a Corruption status in the details of VerifyChecksum actually makes its way to the result of VerifyChecksum. A way we could extend my proposal to check that--without running afoul of assertions with DEBUG_LEVEL>0--is to track not only the ranges of blocks verified, but the positions of all the embedded checksums in the file, and corrupt a random one of those and re-check VerifyChecksum, which should report corruption. (Or corrupt each one in succession, rather than a single random one. This would make failures much easier to reproduce.) (This approach also works around the potential problem for some algorithms of a small chance of checksum input corruptions not modifying the checksum output.)

PR #7190, opened 2020-07-28
[IOTracing PR2]  Store FSSequentialFilePtr object in SequenceFileReader 

akankshamahajan15:
Summary: This diff contains following changes:
    1. Replace `FSSequentialFile` pointer with `FSSequentialFilePtr` object that wraps `FSSequentialFile` pointer in `SequenceFileReader`.

Objective: If tracing is enabled, `FSSequentialFilePtr` returns `FSSequentialFileTracingWrapper` pointer that includes all necessary information in `IORecord` and calls underlying FileSystem and invokes `IOTracer` to dump that record in a binary file. If tracing is disabled then, underlying `FileSystem` pointer is returned directly. `FSSequentialFilePtr` wrapper class is added to bypass the `FSSequentialFileTracingWrapper` when tracing is disabled.
   
Test Plan: make check -j64

    Reviewers:

    Subscribers:

    Tasks:

    Tags:

### DIFF
--- a/env/file_system_tracer.h
+++ b/env/file_system_tracer.h
@@ -131,26 +131,33 @@ class FSSequentialFileTracingWrapper : public FSSequentialFileWrapper {
 // FSSequentialFileTracingWrapper when tracing is disabled.
 class FSSequentialFilePtr {
  public:
-  FSSequentialFilePtr(FSSequentialFile* fs, std::shared_ptr<IOTracer> io_tracer)
-      : fs_(fs), io_tracer_(io_tracer) {
-    fs_tracer_ = new FSSequentialFileTracingWrapper(fs_, io_tracer_);
+  FSSequentialFilePtr() {}
+  FSSequentialFilePtr(std::unique_ptr<FSSequentialFile>&& fs,
+                      const std::shared_ptr<IOTracer>& io_tracer)
+      : fs_(std::move(fs)), io_tracer_(io_tracer) {
+    fs_tracer_.reset(new FSSequentialFileTracingWrapper(fs_.get(), io_tracer_));
   }
-
-  explicit FSSequentialFilePtr(FSSequentialFile* fs)
-      : fs_(fs), io_tracer_(nullptr), fs_tracer_(nullptr) {}
 
   FSSequentialFile* operator->() const {
     if (io_tracer_ && io_tracer_->is_tracing_enabled()) {
-      return fs_tracer_;
+      return fs_tracer_.get();
     } else {
-      return fs_;
+      return fs_.get();
+    }
+  }
+
+  FSSequentialFile* get() const {
+    if (io_tracer_ && io_tracer_->is_tracing_enabled()) {
+      return fs_tracer_.get();
+    } else {
+      return fs_.get();
     }
   }
 
  private:
-  FSSequentialFile* fs_;
+  std::unique_ptr<FSSequentialFile> fs_;
   std::shared_ptr<IOTracer> io_tracer_;
-  FSSequentialFileTracingWrapper* fs_tracer_;
+  std::unique_ptr<FSSequentialFileTracingWrapper> fs_tracer_;
 };
 
 // FSRandomAccessFileTracingWrapper is a wrapper class above FSRandomAccessFile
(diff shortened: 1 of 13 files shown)


env/file_system_tracer.h
```diff
@@ -131,26 +131,30 @@ class FSSequentialFileTracingWrapper : public FSSequentialFileWrapper {
 // FSSequentialFileTracingWrapper when tracing is disabled.
 class FSSequentialFilePtr {
```
mrambacher, line 132: Can you explain why we need this class at all?  Why not have a the caller decide whether to create a "simple" FSSequentialFile or one that is a FSSequentialFileTracingWrapper?  

Changing to using this class (instead of the underlying FSSequentialFile) creates an additional layer of wrapping and causes a bunch of methods to be changed for no material gain.

env/file_system_tracer.h
```diff
@@ -83,6 +85,9 @@ class FileSystemPtr {
     }
   }
 
+  /* Returns the underlying File System pointer */
+  std::shared_ptr<FileSystem> GetFileSystem() const { return fs_; }
```
mrambacher, line 89: this should return a const std::shared_ptr<FileSystem> &

env/file_system_tracer.h
```diff
@@ -72,9 +72,11 @@ class FileSystemPtr {
         fs_tracer_(
             std::make_shared<FileSystemTracingWrapper>(fs_, io_tracer_)) {}
 
-  explicit FileSystemPtr(std::shared_ptr<FileSystem> fs)
```
mrambacher, line 75: This class has the same questions.  Why do we need it?  Can we change the DB to store either the "raw" FileSystem" or a "TracedFileSystem" based on what the input IO settings would be?

env/file_system_tracer.h
akankshamahajan15: When user enables IOTracing then  we call FSSequentialFileTracingWrapper. It creates trace log and call appropriate function to dump that trace log and underlying file system. But in the middle if user wants to disable the tracing. We have no way of finding if tracing is disabled or not in FileSystem level because its pointer has been created to point to FSSequentialFileTracingWrapper. So even if the tracing is disabled it does extra work of creating trace log and passing additional context as part of trace log(like thread id, or other info  not passed yet).  So we want to bypass that.

mrambacher: What does "in the middle" mean?  Once an object is created as traced or not, you are always still going the same direction.

This changeset is very large to solve a very small, localized problem.  

It is much simpler to to create the right class (if given an IOTracer, create a TracingWrapper, if not create a non-wrapped one).  Inside the TracingWrapper, you can decide whether or not to trace based on the current IOTracer settings.  Inside of the TracingWrapper you can handle all of the cases required (like when to create a trace log, etc) based on the settings.  The log may be created when the Wrapper is first created or it may be created later if the settings changed (if that is allowed).  You should be able to have the TracingWrapper be almost no additional code or work if tracing is disabled.

adamretter: > What does "in the middle" mean?

@akankshamahajan15 I would also like to understand what you mean by that please.

akankshamahajan15: Sorry for not being clear in explaining. What I meant was after few DB operations if user want to end the tracing and perform remaining operations, and later on wants to enable the tracing again for few operations.                                        I added a wrapper that overloaded operator -> because in FSSequentialFile, FSWritableFileetc, adding an extra wrapper (virtual functions) and passing calls from TracingWrapper to FileSystem might be costly (tracing is not enabled) . So in order to avoid extra cost when tracing is not enabled and bypass calls from tracing wrapper to underlying File system when tracing is disabled, this new class(FSSequentialFilePtr) is added.

mrambacher: I am still not sure I understand how having this class helps this situation at all.  In order to enable or use tracing, the user must have supplied the io_tracer.  If one is provided and tracing is on, give them back a wrapped class.  Otherwise give them back a raw class.  Or for finer grain, give them a wrapped class if a tracer is provided and for every call of that wrapped class decide whether or not to do the tracing.

Changing all of the APIs to take a different object rather than hiding the complexities inside of a single method seems like a lot of extra complexity and code changes and against some of the basic OO principles.

anand1976: The ```FSSequentialFile``` pointer is stored inside ```SequenceFileReader```, so we need to pass ```io_tracer_``` to the constructor in order to allocate the tracing wrapper and store it there.

PR #7192, opened 2020-07-28
[IOTracing PR3] Store FSRandomAccessPtr object in RandomAccessFileReader

akankshamahajan15:
Summary: Replace FSRandomAccessFile pointer with FSRandomAccessFilePtr
    object in RandomAccessFileReader.
    This new object wraps FSRandomAccessFile pointer.

    Objective: If tracing is enabled, FSRandomAccessFile Ptr returns
    FSRandomAccessFileTracingWrapper pointer that includes all necessary
    information in IORecord and calls underlying FileSystem and invokes
    IOTracer to dump that record in a binary file. If tracing is disabled
    then, underlying FileSystem pointer is returned directly.
    FSRandomAccessFilePtr wrapper class is added to bypass the FSRandomAccessFileWrapper when
    tracing is disabled.

    Test Plan: make check -j64

    Reviewers:

    Subscribers:

    Tasks:

    Tags:

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -1284,8 +1284,8 @@ Status Version::GetTableProperties(std::shared_ptr<const TableProperties>* tp,
   // pass the magic number check in the footer.
   std::unique_ptr<RandomAccessFileReader> file_reader(
       new RandomAccessFileReader(
-          std::move(file), file_name, nullptr /* env */, nullptr /* stats */,
-          0 /* hist_type */, nullptr /* file_read_hist */,
+          std::move(file), file_name, nullptr /* env */, nullptr /* IOTracer */,
+          nullptr /* stats */, 0 /* hist_type */, nullptr /* file_read_hist */,
           nullptr /* rate_limiter */, ioptions->listeners));
   s = ReadTableProperties(
       file_reader.get(), file_meta->fd.GetFileSize(),
@@ -3618,9 +3618,10 @@ VersionSet::VersionSet(const std::string& dbname,
                        WriteController* write_controller,
                        BlockCacheTracer* const block_cache_tracer,
                        const std::shared_ptr<IOTracer>& io_tracer)
-    : column_family_set_(new ColumnFamilySet(
-          dbname, _db_options, storage_options, table_cache,
-          write_buffer_manager, write_controller, block_cache_tracer)),
+    : column_family_set_(
+          new ColumnFamilySet(dbname, _db_options, storage_options, table_cache,
+                              write_buffer_manager, write_controller,
+                              block_cache_tracer, io_tracer)),
       env_(_db_options->env),
       fs_(_db_options->fs, io_tracer),
       dbname_(dbname),
@@ -3660,9 +3661,9 @@ void VersionSet::Reset() {
     Cache* table_cache = column_family_set_->get_table_cache();
     WriteBufferManager* wbm = column_family_set_->write_buffer_manager();
     WriteController* wc = column_family_set_->write_controller();
-    column_family_set_.reset(new ColumnFamilySet(dbname_, db_options_,
-                                                 file_options_, table_cache,
-                                                 wbm, wc, block_cache_tracer_));
+    column_family_set_.reset(
+        new ColumnFamilySet(dbname_, db_options_, file_options_, table_cache,
+                            wbm, wc, block_cache_tracer_, io_tracer_));
   }
   db_id_.clear();
   next_file_number_.store(2);
--- a/env/file_system_tracer.h
+++ b/env/file_system_tracer.h
@@ -200,27 +200,35 @@ class FSRandomAccessFileTracingWrapper : public FSRandomAccessFileWrapper {
 // FSRandomAccessFileTracingWrapper when tracing is disabled.
 class FSRandomAccessFilePtr {
  public:
-  FSRandomAccessFilePtr(FSRandomAccessFile* fs,
-                        std::shared_ptr<IOTracer> io_tracer)
-      : fs_(fs),
-        io_tracer_(io_tracer),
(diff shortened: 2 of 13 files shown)


db/version_set.cc
```diff
@@ -4881,7 +4882,8 @@ Status VersionSet::ListColumnFamilies(std::vector<std::string>* column_families,
     if (!s.ok()) {
       return s;
   }
-  file_reader.reset(new SequentialFileReader(std::move(file), manifest_path));
+  file_reader.reset(new SequentialFileReader(std::move(file), manifest_path,
```
zhichao-cao, line 4885: What's the reason to pass `nullptr` here when creating the new `SequentialFileReader`?

file/file_util.h
```diff
@@ -13,13 +13,15 @@
 #include "rocksdb/sst_file_writer.h"
 #include "rocksdb/status.h"
 #include "rocksdb/types.h"
+#include "trace_replay/io_tracer.h"
 
 namespace ROCKSDB_NAMESPACE {
 // use_fsync maps to options.use_fsync, which determines the way that
 // the file is synced after copying.
 extern IOStatus CopyFile(FileSystem* fs, const std::string& source,
```
zhichao-cao, line 21: We may also need to update the interface of `GenerateOneFileChecksum` in L36

db/version_set.cc
akankshamahajan15: In some places we don't have access to ```IOTracer```. ```IOTracer``` object is created in ```DBImpl``` and then passed to different classes. ```VersionSet::ListColumnFamilies```  is static function so it doesn't have access to IOTracer. I will remove the nullptr here as  ```SequentialFileReader``` accepts default ```nullptr``` for IOTracer argument.

file/file_util.h
akankshamahajan15: I will check that interface. Thanks!

db/version_set.cc
zhichao-cao: Thanks for the explanation.

env/file_system_tracer.h
```diff
@@ -193,27 +200,35 @@ class FSRandomAccessFileTracingWrapper : public FSRandomAccessFileWrapper {
...
+      return fs_.get();
     }
   }
 
  private:
-  FSRandomAccessFile* fs_;
+  std::unique_ptr<FSRandomAccessFile> fs_;
   std::shared_ptr<IOTracer> io_tracer_;
-  FSRandomAccessFileTracingWrapper* fs_tracer_;
+  std::unique_ptr<FSRandomAccessFileTracingWrapper> fs_tracer_;
```
anand1976, line 231: Does it really need to be a ```unique_ptr```, or can it just be an embedded object? The constructor is allocating it no matter whether ```io_tracer``` is passed or not, so we could avoid the extra malloc call.

env/file_system_tracer.h
```diff
@@ -131,26 +131,33 @@ class FSSequentialFileTracingWrapper : public FSSequentialFileWrapper {
...
+      return fs_.get();
     }
   }
 
  private:
-  FSSequentialFile* fs_;
+  std::unique_ptr<FSSequentialFile> fs_;
   std::shared_ptr<IOTracer> io_tracer_;
-  FSSequentialFileTracingWrapper* fs_tracer_;
+  std::unique_ptr<FSSequentialFileTracingWrapper> fs_tracer_;
```
anand1976, line 160: Does it need to be a ```unique_ptr```?

file/random_access_file_reader.h
```diff
@@ -100,8 +102,13 @@ class RandomAccessFileReader {
 #endif
   }
 
-  RandomAccessFileReader(RandomAccessFileReader&& o) ROCKSDB_NOEXCEPT {
-    *this = std::move(o);
+  RandomAccessFileReader(RandomAccessFileReader&& o) ROCKSDB_NOEXCEPT
+      : file_(std::move(o.file_)) {
```
anand1976, line 106: What's the reason for ```std::move``` of individual member variables instead of ```std::move(o)```? Also, what about ```file_name_``` and ```listeners_```?

env/file_system_tracer.h
```diff
@@ -131,24 +131,24 @@ class FSSequentialFileTracingWrapper : public FSSequentialFileWrapper {
...
-    fs_tracer_.reset(new FSSequentialFileTracingWrapper(fs_.get(), io_tracer_));
-  }
+      : fs_(std::move(fs)),
+        io_tracer_(io_tracer),
+        fs_tracer_(fs_.get(), io_tracer_) {}
 
   FSSequentialFile* operator->() const {
     if (io_tracer_ && io_tracer_->is_tracing_enabled()) {
-      return fs_tracer_.get();
+      return const_cast<FSSequentialFileTracingWrapper*>(&fs_tracer_);
```
anand1976, line 143: Is ```const_cast``` necessary, since ```FSSequentialFileTracingWrapper``` is a descendant of ```FSSequentialFile```?

akankshamahajan15: It is giving error without const cast because function is defined const. So I had to add the type cast

file/random_access_file_reader.h
akankshamahajan15: In std::move(o), it expects  FSRandomAccessFilePtr to have default constructor, which  in turns expect FSRandomAccessFileTracingWrapper(since its a object) to have default Constructor but FSRandomAccessFileTracingWrapper is inherited from FSRandomAccessFileWrapper and it expects pointer to file system (No default constructor). So I added it to do member by member move. (similar to move operator). But i wanted to remove these move functions as operator= also doesn't have ```file_name_``` and ```listeners_``` .

akankshamahajan15:
Updated ```RandomAccessFileRandom``` to use implicit move constructor and move operator.

PR #7193, opened 2020-07-29
[IOTracing PR4] Store FSWritableFilePtr object in WritableFileWriter

akankshamahajan15:
Summary: Replace FSWritableFile pointer with FSWritableFilePtr
    object in WritableFileWriter.
    This new object wraps FSWritableFile pointer.

    Objective: If tracing is enabled, FSWritableFile Ptr returns
    FSWritableFileTracingWrapper pointer that includes all necessary
    information in IORecord and calls underlying FileSystem and invokes
    IOTracer to dump that record in a binary file. If tracing is disabled
    then, underlying FileSystem pointer is returned directly.
    FSWritableFilePtr wrapper class is added to bypass the
    FSWritableFileWrapper when
    tracing is disabled.

    Test Plan: make check -j64

    Reviewers:

    Subscribers:

    Tasks:

    Tags:


db/compaction/compaction_job.cc
```diff
@@ -1542,7 +1542,8 @@ Status CompactionJob::OpenCompactionOutputFile(
                            &syncpoint_arg);
 #endif
   Status s;
-  IOStatus io_s = NewWritableFile(fs_, fname, &writable_file, file_options_);
```
zhichao-cao, line 1545: Is it possible that we change the fs_ pointer instead of using db_options_.fs.GetFileSystem().get() directly. It might cause potential that using different fs_ in the same compaction job.

db/flush_job.cc
```diff
@@ -400,10 +400,10 @@ Status FlushJob::WriteLevel0Table() {
           output_compression_, mutable_cf_options_.sample_for_compression,
           mutable_cf_options_.compression_opts,
           mutable_cf_options_.paranoid_file_checks, cfd_->internal_stats(),
-          TableFileCreationReason::kFlush, &io_s, event_logger_,
-          job_context_->job_id, Env::IO_HIGH, &table_properties_, 0 /* level */,
-          creation_time, oldest_key_time, write_hint, current_time, db_id_,
-          db_session_id_);
+          TableFileCreationReason::kFlush, &io_s, nullptr /*IOTracer*/,
```
anand1976, line 403: Can we cache io_tracer in the constructor and pass it here? Otherwise flush IO calls will not get traced.

db/version_set.cc
```diff
@@ -1284,8 +1284,8 @@ Status Version::GetTableProperties(std::shared_ptr<const TableProperties>* tp,
   // pass the magic number check in the footer.
   std::unique_ptr<RandomAccessFileReader> file_reader(
       new RandomAccessFileReader(
-          std::move(file), file_name, nullptr /* env */, nullptr /* stats */,
-          0 /* hist_type */, nullptr /* file_read_hist */,
+          std::move(file), file_name, nullptr /* env */, nullptr /* IOTracer */,
```
anand1976, line 1287: I thought io_tracer is passed as an argument to the constructor. Can we use that instead of ```nullptr```?

env/file_system_tracer.h
```diff
@@ -265,24 +270,35 @@ class FSWritableFileTracingWrapper : public FSWritableFileWrapper {
...
-      : fs_(fs),
+  FSWritableFilePtr(std::unique_ptr<FSWritableFile>&& fs,
+                    const std::shared_ptr<IOTracer>& io_tracer)
+      : fs_(std::move(fs)),
         io_tracer_(io_tracer),
-        fs_tracer_(new FSWritableFileTracingWrapper(fs_, io_tracer_)) {}
-
-  explicit FSWritableFilePtr(FSWritableFile* fs)
-      : fs_(fs), io_tracer_(nullptr), fs_tracer_(nullptr) {}
+        fs_tracer_(new FSWritableFileTracingWrapper(fs_.get(), io_tracer_)) {}
```
anand1976, line 277: Can we make ```fs_tracer_``` of type ```FSWritableFileTracingWrapper``` and avoid the extra allocation?

akankshamahajan15: Originally, ```WritableFileWriter::Close``` checks if ```FSWritableFile``` ptr is  then return else does some work and close the file and set ```FSWritableFile``` ptr to null. (WritableFileWriter::Close can be called twice). So I kept ```FSWritableFileTracingWrapper``` pointer here so that I can reset it to nullptr in ```WritableFileWriter::Close```  and can check if its null or not by ``` if(writable_file_.get() != nullptr)```

db/flush_job_test.cc
```diff
@@ -466,14 +466,14 @@ TEST_F(FlushJobTest, Snapshots) {
...
-                     &event_logger, true, true /* sync_output_directory */,
-                     true /* write_manifest */, Env::Priority::USER);
+  FlushJob flush_job(
+      dbname_, versions_->GetColumnFamilySet()->GetDefault(), db_options_,
+      *cfd->GetLatestMutableCFOptions(), nullptr /* memtable_id */,
+      env_options_, versions_.get(), &mutex_, &shutting_down_, snapshots,
+      kMaxSequenceNumber, snapshot_checker, &job_context, nullptr, nullptr,
+      nullptr, kNoCompression, db_options_.statistics.get(), &event_logger,
+      true, true /* sync_output_directory */, true /* write_manifest */,
+      Env::Priority::USER, nullptr /*IOTracer*/);
```
zhichao-cao, line 476: One question, for IOtracer here, we pass nullptr. Is this a temporal change? In later on PR, it will be removed?

db/version_edit_handler.h
```diff
@@ -36,7 +36,7 @@ class VersionEditHandler {
       bool read_only,
       const std::vector<ColumnFamilyDescriptor>& column_families,
       VersionSet* version_set, bool track_missing_files,
-      bool ignore_missing_files);
+      bool ignore_missing_files, const std::shared_ptr<IOTracer>& io_tracer);
```
zhichao-cao, line 39: Since io_tracer is a shared_ptr type, is it necessary to use reference?

db/flush_job_test.cc
akankshamahajan15: Right now, nullptr is passed as IOTracer where it can't be accessed. It's created in ```DBImpl``` and then passed to all constructors. We can make changes in future  to create separate IOTracer pointer where it can't be passed from ```DBImpl```.

zhichao-cao: I see, thanks for the explanation!

file/writable_file_writer.cc
```diff
@@ -197,7 +197,7 @@ IOStatus WritableFileWriter::Close() {
     s = interim;
   }
 
-  writable_file_.reset();
+  writable_file_.Reset();
```
anand1976, line 200: Nit: Can we call it ```reset``` to make it consistent with ```get```?

anand1976 (approved):
LGTM

db/version_edit_handler.h
akankshamahajan15: I used reference to avoid copy as this pointer will not be deleted till DBImpl is destroyed or someone call DB::EndIOTrace().

PR #7195, opened 2020-07-29
Add a new EntryType for deletion with timestamp

riversand963:
Add `kEntryDeleteWithTimestamp` to `EntryType` which is a public API.

Test plan:
make check

### DIFF
--- a/include/rocksdb/types.h
+++ b/include/rocksdb/types.h
@@ -18,14 +18,15 @@ typedef uint64_t SequenceNumber;
 const SequenceNumber kMinUnCommittedSeq = 1;  // 0 is always committed
 
 // User-oriented representation of internal key types.
+// Ordering of this enum entries should not change.
 enum EntryType {
   kEntryPut,
   kEntryDelete,
-  kEntryDeleteWithTimestamp,
   kEntrySingleDelete,
   kEntryMerge,
   kEntryRangeDeletion,
   kEntryBlobIndex,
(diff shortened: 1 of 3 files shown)


riversand963:
Will update HISTORY.md later.
IMO, I think `EntryType` should use safer enum types supported in C++, or at least assign values explicitly, but for now I'll stick to the current convention.

riversand963:
cc @hliu18

hliu18:
LGTM

include/rocksdb/types.h
```diff
@@ -21,6 +21,7 @@ const SequenceNumber kMinUnCommittedSeq = 1;  // 0 is always committed
 enum EntryType {
   kEntryPut,
   kEntryDelete,
+  kEntryDeleteWithTimestamp,
```
ajkr, line 24: enum order changes kind of scare me. For example, someone might store or transfer some keys/values along with the EntryType. In that case they likely encode EntryType as an int. So upgrading, or transferring to a host running an older/newer version will decode that int as a different EntryType from what was intended.

We could comment on this enum that the ordering is subject to change.

riversand963: It scares me too. Imo, int-based enums should always have explicit value assignments. I looked at previous PRs changing this enum, and they also changed the order...

riversand963: okay, let me not change the ordering of enum.

PR #7198, opened 2020-07-29
[IOTracing PR5] Use FSRandomRWFilePtr Object to call underlying file system.

akankshamahajan15:
Summary: Replace FSRandomRWFile pointer with FSRandomRWFilePtr object in the rocksdb internal code.
This new object wraps FSRandomRWFile pointer.

Objective: If tracing is enabled, FSRandomRWFile object returns FSRandomRWFileTracingWrapper pointer that includes all necessary information in IORecord and calls underlying FileSystem and invokes IOTracer to dump that record in a binary file. If tracing is disabled then, underlying FileSystem pointer is returned directly.
FSRandomRWFilePtr wrapper class is added to bypass the FSRandomRWFileWrapper when
tracing is disabled.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:


zhichao-cao (approved):
LGTM, thanks!

Issue #7197, opened 2020-07-29
double-lock in WriteImplWALOnly


byteink:
In WriteImplWALOnly, mutex_ is locked twice, cause deadlock.

First time:
https://github.com/facebook/rocksdb/blob/d9d190742c0a8a3d5ff0b0867383bd38c1d0f5f0/db/db_impl/db_impl_write.cc#L689

Then the second one in the call of WriteStatusCheck:
https://github.com/facebook/rocksdb/blob/d9d190742c0a8a3d5ff0b0867383bd38c1d0f5f0/db/db_impl/db_impl_write.cc#L838

zhichao-cao:
Hi @byteink , you are right, will create a PR to fix it. Thanks for pointing out it!

PR #7199, opened 2020-07-29
Fix the potential deadlock in WriteImplWALOnly and UnorderedWriteMemtable

zhichao-cao:
Pointed out by #7197 , there is a double lock in WriteImplWALOnly.
Also find another deadlock in UnorderedWriteMemtable. Move the check after switch_all_.notify_all().

Test plan: pass make check

### DIFF
--- a/db/db_impl/db_impl_write.cc
+++ b/db/db_impl/db_impl_write.cc
@@ -689,7 +689,7 @@ Status DBImpl::WriteImplWALOnly(
       InstrumentedMutexLock l(&mutex_);
       bool need_log_sync = false;
       status = PreprocessWrite(write_options, &need_log_sync, &write_context);
-      WriteStatusCheck(status);
+      WriteStatusCheckOnLocked(status);
     }
     if (!status.ok()) {
       WriteThread::WriteGroup write_group;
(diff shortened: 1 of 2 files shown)


db/db_impl/db_impl_write.cc
```diff
@@ -686,9 +686,10 @@ Status DBImpl::WriteImplWALOnly(
     // TODO(myabandeh): Make preliminary checks thread-safe so we could do them
     // without paying the cost of obtaining the mutex.
     if (status.ok()) {
-      InstrumentedMutexLock l(&mutex_);
+      mutex_.Lock();
       bool need_log_sync = false;
       status = PreprocessWrite(write_options, &need_log_sync, &write_context);
+      mutex_.Unlock();
```
anand1976, line 692: I'm worried if this might expose some race conditions. For example, if PreprocessWrite() returns an error status due to bg_error_, but it gets cleared before the call to WriteStatusCheck(). Maybe we can define a WriteStatusCheckLocked() and make WriteStatusCheck() a wrapper around it. PreprocessWrite() can directly call WriteStatusCheckLocked().

zhichao-cao: I see, maybe WriteStatusCheckLocked() is a better solution for this.

db/db_impl/db_impl_write.cc
```diff
@@ -830,6 +830,17 @@ Status DBImpl::WriteImplWALOnly(
...
+  mutex_.AssertHeld();
+  if (immutable_db_options_.paranoid_checks && !status.ok() &&
+      !status.IsBusy() && !status.IsIncomplete()) {
+    error_handler_.SetBGError(status, BackgroundErrorReason::kWriteCallback);
+  }
+}
+
 void DBImpl::WriteStatusCheck(const Status& status) {
   // Is setting bg_error_ enough here?  This will at least stop
   // compaction and fail any further writes.
```
anand1976, line 846: Minor - can we acquire mutex here and call ```WriteStatusCheckOnLocked``` so the code is not duplicated?

zhichao-cao: @anand1976 Updated and also addressing the deadlock in UnorderedWriteMemtable.

byteink: I worry about this fix maybe cause performance problems. I mean, we must to acquire the mutex even if the if condition is not met. the condition is ```immutable_db_options_.paranoid_checks && !status.ok() && !status.IsBusy() && !status.IsIncomplete()```

anand1976: That's a good point. It can cause performance regression for unordered writes. I'm ok with reverting this function back to the way it was.

PR #7202, opened 2020-07-30
Allow optimization level to be set in Makefile

ajkr:
`-O3` is already adopted widely, so we should make it easier to configure
for development/open source. This PR adds an `OPTIMIZE_LEVEL` variable
that users can set to override the `-O` flag chosen in the Makefile.

Test Plan: built a few different ways and verified correct value is passed for `-O` flag

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,8 @@ else
 	OPTIMIZE_LEVEL ?= -Os
 endif
 endif
+# `OPTIMIZE_LEVEL` is empty when the user does not set it and `DEBUG_LEVEL=2`.
+# In that case, the compiler default (`-O0` for gcc and clang) will be used.
 OPT += $(OPTIMIZE_LEVEL)
 
 # compile with -O2 if debug level is not 2
--- a/Makefile
+++ b/Makefile
@@ -131,11 +131,12 @@ endif
 # Figure out optimize level.
 ifneq ($(DEBUG_LEVEL), 2)
 ifeq ($(LITE), 0)
-	OPT += -O2
+	OPTIMIZE_LEVEL ?= -O2
 else
-	OPT += -Os
+	OPTIMIZE_LEVEL ?= -Os
 endif
 endif
+OPT += $(OPTIMIZE_LEVEL)
 
 # compile with -O2 if debug level is not 2
 ifneq ($(DEBUG_LEVEL), 2)


Makefile
```diff
@@ -131,11 +131,12 @@ endif
 # Figure out optimize level.
 ifneq ($(DEBUG_LEVEL), 2)
```
mrambacher, line 132: Are you allowing/expecting OPTIMIZE_LEVEL to be set when DEBUG_LEVEL=2?  This current code allows it.  

This might read cleaner as:
ifneq($(DEBUG_LEVEL,2)
ifneq ($OPTIMIZE_LEVEL,)
OPT += $(OPTIMIZE_LEVEL)
elif ($LITE)
else

ajkr: I think it's conceptually simpler to allow `OPTIMIZE_LEVEL` (a more specific option) to override the level implied by `DEBUG_LEVEL` (a more general option). That doesn't mean all combinations will work, but if somebody specifies something like `OPTIMIZE_LEVEL=-O2`, I'd rather let it fail than sanitize it to something else.

ajkr:
> What about equivalent changes for CMake? We should try to keep the two systems from diverging as much as possible...

cmake users can already build with `-O3` like this:

```
$ cmake <rocksdb dir> -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS_RELEASE=-O3
```

Makefile
mrambacher: I am fine with falling through, but was wondering if that should be the intent or if -O0 should be set for debug (or if no optimization level should be set).  Please just add a comment as to the intent so it becomes clear.

Also, do you expect that the DEBUG_LEVEL=1 and 0 should use the same optimization level or not?

mrambacher:
> > What about equivalent changes for CMake? We should try to keep the two systems from diverging as much as possible...
> 
> cmake users can already build with `-O3` like this:
> 
> ```
> $ cmake <rocksdb dir> -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS_RELEASE=-O3
> ```

But in the case of LITE mode it is hard-coded to do -Os in the CMake build, so the user cannot really override it, can they?

ajkr:
> > > What about equivalent changes for CMake? We should try to keep the two systems from diverging as much as possible...
> > 
> > 
> > cmake users can already build with `-O3` like this:
> > ```
> > $ cmake <rocksdb dir> -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS_RELEASE=-O3
> > ```
> 
> But in the case of LITE mode it is hard-coded to do -Os in the CMake build, so the user cannot really override it, can they?

Right. I verified that does not result in a consistent behavior. Actually even if we omit the `-DCMAKE_CXX_FLAGS_RELEASE=-O3` in the command above, it still adds `-O2` by default, which results in building with `-Os` and `-O2` -- only the latter can take effect. So the way `-Os` is specified in CMakeLists.txt is incorrect.

Makefile
ajkr: > I am fine with falling through, but was wondering if that should be the intent or if -O0 should be set for debug (or if no optimization level should be set).

I am ok with leaving it blank and allowing the compiler default (`-O0` for gcc and clang) to be used. 

> Also, do you expect that the DEBUG_LEVEL=1 and 0 should use the same optimization level or not?

It seems ok. I usually think of `DEBUG_LEVEL=1` as enabling assertions (i.e., not specifying `-DNDEBUG`), while allowing optimizations.

ajkr:
> > > > What about equivalent changes for CMake? We should try to keep the two systems from diverging as much as possible...
> > > 
> > > 
> > > cmake users can already build with `-O3` like this:
> > > ```
> > > $ cmake <rocksdb dir> -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS_RELEASE=-O3
> > > ```
> > 
> > 
> > But in the case of LITE mode it is hard-coded to do -Os in the CMake build, so the user cannot really override it, can they?
> 
> Right. I verified that does not result in a consistent behavior. Actually even if we omit the `-DCMAKE_CXX_FLAGS_RELEASE=-O3` in the command above, it still adds `-O2` by default, which results in building with `-Os` and `-O2` -- only the latter can take effect. So the way `-Os` is specified in CMakeLists.txt is incorrect.

Created issue #7213.

PR #7203, opened 2020-07-30
 Add support to start and end IOTracing through DB APIs

akankshamahajan15:
Summary: 1. Add support to start io tracing through DB::StartIOTrace(Env*, const TraceOptions&, std::unique_ptr<TraceWriter>&&) and end tracing through DB::EndIOTrace(). This doesn't trace DB::Open.

User side code:

//Open DB
DB::Open(options, dbname, &db);

/* Start tracing */
db->StartIOTrace(env, trace_opt, std::move(trace_writer));

/* Perform Operations */

/*End tracing*/
db->EndIOTrace();

2. Fix the build errors for Windows.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -55,6 +55,7 @@
 #include "rocksdb/write_buffer_manager.h"
 #include "table/scoped_arena_iterator.h"
 #include "trace_replay/block_cache_tracer.h"
+#include "trace_replay/io_tracer.h"
 #include "trace_replay/trace_replay.h"
 #include "util/autovector.h"
 #include "util/hash.h"
@@ -444,6 +445,13 @@ class DBImpl : public DB {
 
   using DB::EndBlockCacheTrace;
   Status EndBlockCacheTrace() override;
+
+  using DB::StartIOTrace;
+  Status StartIOTrace(Env* env, const TraceOptions& options,
+                      std::unique_ptr<TraceWriter>&& trace_writer) override;
+
+  using DB::EndIOTrace;
+  Status EndIOTrace() override;
 
   using DB::GetPropertiesOfAllTables;
   virtual Status GetPropertiesOfAllTables(
@@ -997,6 +1005,7 @@ class DBImpl : public DB {
   bool own_info_log_;
   const DBOptions initial_db_options_;
   Env* const env_;
+  std::shared_ptr<IOTracer> io_tracer_;
   std::shared_ptr<FileSystem> fs_;
   const ImmutableDBOptions immutable_db_options_;
   MutableDBOptions mutable_db_options_;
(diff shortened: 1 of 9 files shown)


db/db_impl/db_impl.h
```diff
@@ -437,6 +438,19 @@ class DBImpl : public DB {
...
 
+  // Need a static version that can be called before DB::Open.
+  // It is used by rocksdb internal only.
+  static Status StartIOTrace(Env* env, const TraceOptions& options,
+                             std::unique_ptr<TraceWriter>&& trace_writer);
+
+  static Status EndIOTrace();
+
+  static std::shared_ptr<IOTracer> GetIOTracer() {
+    static IOTracer io_tracer;
```
anand1976, line 449: This means ```io_tracer``` is process scope and not DB. How would a user enable/disable per DB?

include/rocksdb/db.h
```diff
@@ -1587,6 +1587,15 @@ class DB {
     return Status::NotSupported("EndTrace() is not implemented.");
   }
 
+  virtual Status StartIOTrace(Env* /*env*/, const TraceOptions& /*options*/,
```
mrambacher, line 1590: I think these methods also need to be added to the StackableDB

mrambacher, line 1590: You need to document the expected behavior of this methods and their arguments.

mrambacher, line 1590: Why not have the method take in the trace file name (rather than the trace writer) and have the trace writer created internally?  I think you have all of the other arguments and it would make for a cleaner interface.

akankshamahajan15: Other tracers like BlockCacheTrace and Tracer, all have start trace functions that take TraceWriter as argument. So I added this to make it consistent with other tracers.   So user set trace_writer through  NewFileTraceWriter(Env* env, const EnvOptions& env_options, const std::string& trace_filename, std::unique_ptr<TraceWriter>* trace_writer); and pass that to StartTrace/ StartBlockCacheTrace/ StartIOTrace.

akankshamahajan15: Thank you for pointing that out. I will add to StackableDB as well.

akankshamahajan15:
> Are there tests for these new methods?

There aren't any tests right now for new methods because this PR adds support. It doesn't trace anything yet. I will upload the remaining PRs that have FileSystem side changes because of which tracing will happen and then add the test cases. For IOTracer I have already added the tests that test if records are correctly traced (dumped) and read and once IOTracer::EndIOTrace is called no records are added later.

include/rocksdb/db.h
akankshamahajan15: I was thinking that once all changes are done (File System side),  I will add it to History.md, wiki and other required places.

anand1976: Maybe just add a comment saying they are experimental, for now.

PR #7207, opened 2020-07-30
feat: export Env::default()

developerfred:
I need this exported function to disable threads in rust.

https://github.com/rust-rocksdb/rust-rocksdb/pull/448#issuecomment-666628096


developerfred:
> Can you please better explain what you are trying to do here? Are you trying to get the default environment or something else? Is there a reason "rocksdb_create_default_env" does not work for your use case?
> 
> I am not sure I understand exactly what this code is doing or you hope to accomplish.

I need to disable the threads by default, currently the default that we consume returns an object you can learn more in this chat link
https://github.com/rust-rocksdb/rust-rocksdb/pull/448#issuecomment-666628096

Issue #6698, opened 2020-04-14
TimerTest.SingleScheduleRepeatedlyTest can hang


pdillinger:
I had a 'make check' hang in timer_test. I was able to reproduce twice with `while ./timer_test; do :; done` and running some other processes at the same time:

[==========] Running 4 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 4 tests from TimerTest
[ RUN      ] TimerTest.SingleScheduleOnceTest
[       OK ] TimerTest.SingleScheduleOnceTest (1 ms)
[ RUN      ] TimerTest.MultipleScheduleOnceTest
[       OK ] TimerTest.MultipleScheduleOnceTest (0 ms)
[ RUN      ] TimerTest.SingleScheduleRepeatedlyTest
1

sagar0:
Thanks @pdillinger , I'll look into this.

pdillinger:
Hung for me again locally

pdillinger:
Again, in TimerTest.MultipleScheduleRepeatedlyTest this time

pdillinger:
Hung for me again, and kept my 'make check' run from getting to the format check before pushing to PR.

sagar0:
Sorry about that. I'll be looking into this today. I'll disable this test as a short term mitigation if I can't fix it today.

sagar0:
I could not reproduce the issue on my dev server. 
```
svemuri@devbig187 ~/rocksdb (timer-disable-test) (timer-disable-test) $ TEST_TMPDIR=/dev/shm ~/gtest-parallel/gtest-parallel ./timer_test --gtest_filter=TimerTest.SingleScheduleRepeatedlyTest --repeat=1
[1/1] TimerTest.SingleScheduleRepeatedlyTest (4 ms)
svemuri@devbig187 ~/rocksdb (timer-disable-test) (timer-disable-test) $ TEST_TMPDIR=/dev/shm ~/gtest-parallel/gtest-parallel ./timer_test --gtest_filter=TimerTest.SingleScheduleRepeatedlyTest --repeat=10
[10/10] TimerTest.SingleScheduleRepeatedlyTest (4 ms)
svemuri@devbig187 ~/rocksdb (timer-disable-test) (timer-disable-test) $ TEST_TMPDIR=/dev/shm ~/gtest-parallel/gtest-parallel ./timer_test --gtest_filter=TimerTest.SingleScheduleRepeatedlyTest --repeat=100
[...]

PR #7208, opened 2020-07-31
Fix hang timer tests on macos

jay-zhuang:
And re-enable disabled tests.
The issue is caused by `CondVar.TimedWait()` doesn't use `MockTimeEnv`.

Issue: #6698

Test: `./timer_test --gtest_repeat=1000`

### DIFF
--- a/util/timer.h
+++ b/util/timer.h
@@ -13,7 +13,7 @@
 #include <utility>
 #include <vector>
 
-#include "port/port.h"
+#include "monitoring/instrumented_mutex.h"
 #include "rocksdb/env.h"
 #include "util/mutexlock.h"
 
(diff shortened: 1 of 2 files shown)


jay-zhuang:
> Thanks @jay-zhuang for the fix.
> Can you, in PR description, briefly talk about the reason for this fix on Mac OS?

On MacOS, `CondVar.TimedWait()` doesn't use the time from `MockTimeEnv`, instead it still uses the system time. I copied the solution from https://github.com/facebook/rocksdb/blob/52d59e0c93dfa3629cbcc2c2991a7009e3ec0595/monitoring/stats_history_test.cc#L46
It's not a perfect solution and only work for this case. I updated code comments and summary for that.

pdillinger:
See also #7101 which got caught up in my unexpected few weeks absence. This change is OK by me and I can fix conflicts when I'm able to return to coding work.

PR #7214, opened 2020-08-02
BaseDeltaIterator now adheres to ReadOptions iterate_upper_bound

adamretter:
Previously `iterate_upper_bound` set on a `BaseDeltaIterator` was not correctly respected.

There were no existing tests for this, so I added quite a lot of tests first.

As well as several fixes to `BaseDeltaIterator`, this introduces the functions `Iterator::ChecksLowerBound` and `Iterator::ChecksUpperBound` these can be used by a wrapping Iterator, such as `BaseDeltaIterator`, to prevent duplication the bounds check work. At present other wrapped iterators that do correct check bounds do not necessarily implement these, a future piece of work would be to have other bounds checking iterators (such as `ForwardIterator` and `ReverseIterator`) implement these.

Additionally, whilst this PR also has added much of the base work for correctly implementing `iterate_lower_bound` on `BaseDeltaIterator`, it's focus was that of the more commonly used `iterate_upper_bound`. The author makes no claims as to the correctness of `iterate_lower_bound`.

### DIFF
--- a/include/rocksdb/iterator.h
+++ b/include/rocksdb/iterator.h
@@ -115,6 +115,24 @@ class Iterator : public Cleanable {
     assert(false);
     return Slice();
   }
+
+  /**
+   * true if this iterator is already
+   * checking ReadOptions::iterate_lower_bound,
+   * false otherwise.
+   */
+  virtual bool has_lower_bound() const {
+    return false;
+  }
+
+  /**
+   * true if this iterator is already
+   * checking ReadOptions::iterate_upper_bound,
+   * false otherwise.
+   */
+  virtual bool has_upper_bound() const {
+    return false;
+  }
 };
 
 // Return an empty iterator (yields nothing).
--- a/include/rocksdb/iterator.h
+++ b/include/rocksdb/iterator.h
@@ -117,42 +117,22 @@ class Iterator : public Cleanable {
   }
 
   /**
-   * true if this iterator is already
-   * checking ReadOptions::iterate_lower_bound.
-   * false indicates that either the iterator does
-   * not check the lower bound, or that it does not
-   * report that it is performing the check.
-   */
-  virtual bool ChecksLowerBound() const = 0;
-
-  /**
    * Returns the lower bound if this iterator has a
    * ReadOptions::iterate_lower_bound set, else
    * returns nullptr.
    *
-   * Just because a lower bound is present, it does
-   * not mean that it is checked. This can however
-   * be determined by calling Iterator::ChecksLowerBound().
+   * If a lower bound is present, it should be assumed
+   * that the iterator adheres to it.
    */
   virtual const Slice* lower_bound() const = 0;
-
-  /**
-   * true if this iterator is already
-   * checking ReadOptions::iterate_upper_bound.
-   * false indicates that either the iterator does
-   * not check the upper bound, or that it does not
-   * report that it is performing the check.
-   */
-  virtual bool ChecksUpperBound() const = 0;
 
   /**
    * Returns the upper bound if this iterator has a
    * ReadOptions::iterate_upper_bound set, else
    * returns nullptr.
    *
-   * Just because an upper bound is present, it does
-   * not mean that it is checked. This can however
-   * be determined by calling Iterator::ChecksUpperBound().
+   * If an upper bound is present, it should be assumed
+   * that the iterator adheres to it.
    */
   virtual const Slice* upper_bound() const = 0;
 };
--- a/include/rocksdb/iterator.h
+++ b/include/rocksdb/iterator.h
@@ -123,7 +123,7 @@ class Iterator : public Cleanable {
    * not check the lower bound, or that it does not
    * report that it is performing the check.
    */
-  virtual bool ChecksLowerBound() const { return false; }
+  virtual bool ChecksLowerBound() const = 0;
 
   /**
    * Returns the lower bound if this iterator has a
@@ -134,7 +134,7 @@ class Iterator : public Cleanable {
    * not mean that it is checked. This can however
    * be determined by calling Iterator::ChecksLowerBound().
    */
-  virtual const Slice* lower_bound() const { return nullptr; }
+  virtual const Slice* lower_bound() const = 0;
 
   /**
    * true if this iterator is already
@@ -143,7 +143,7 @@ class Iterator : public Cleanable {
    * not check the upper bound, or that it does not
    * report that it is performing the check.
    */
-  virtual bool ChecksUpperBound() const { return false; }
+  virtual bool ChecksUpperBound() const = 0;
 
   /**
    * Returns the upper bound if this iterator has a
@@ -154,7 +154,7 @@ class Iterator : public Cleanable {
    * not mean that it is checked. This can however
    * be determined by calling Iterator::ChecksUpperBound().
    */
-  virtual const Slice* upper_bound() const { return nullptr; }
+  virtual const Slice* upper_bound() const = 0;
 };
 
 // Return an empty iterator (yields nothing).
--- a/include/rocksdb/iterator.h
+++ b/include/rocksdb/iterator.h
@@ -121,7 +121,7 @@ class Iterator : public Cleanable {
    * checking ReadOptions::iterate_lower_bound,
    * false otherwise.
    */
-  virtual bool check_lower_bound() const { return false; }
+  virtual bool ChecksLowerBound() const { return false; }
 
   /**
    * Returns the lower bound if this iterator has a
@@ -130,16 +130,16 @@ class Iterator : public Cleanable {
    *
    * Just because a lower bound is present, it does
    * not mean that it is checked. This can however
-   * be determined by calling Iterator::check_lower_bound().
+   * be determined by calling Iterator::ChecksLowerBound().
    */
-  virtual const Slice* iterate_lower_bound() const { return nullptr; }
+  virtual const Slice* lower_bound() const { return nullptr; }
 
   /**
    * true if this iterator is already
    * checking ReadOptions::iterate_upper_bound,
    * false otherwise.
    */
-  virtual bool check_upper_bound() const { return false; }
+  virtual bool ChecksUpperBound() const { return false; }
 
   /**
    * Returns the upper bound if this iterator has a
@@ -148,9 +148,9 @@ class Iterator : public Cleanable {
    *
    * Just because an upper bound is present, it does
    * not mean that it is checked. This can however
-   * be determined by calling Iterator::check_upper_bound().
+   * be determined by calling Iterator::ChecksUpperBound().
    */
-  virtual const Slice* iterate_upper_bound() const { return nullptr; }
+  virtual const Slice* upper_bound() const { return nullptr; }
 };
 
 // Return an empty iterator (yields nothing).
--- a/include/rocksdb/iterator.h
+++ b/include/rocksdb/iterator.h
@@ -121,14 +121,36 @@ class Iterator : public Cleanable {
    * checking ReadOptions::iterate_lower_bound,
    * false otherwise.
    */
-  virtual bool has_lower_bound() const { return false; }
+  virtual bool check_lower_bound() const { return false; }
+
+  /**
+   * Returns the lower bound if this iterator has a
+   * ReadOptions::iterate_lower_bound set, else
+   * returns nullptr.
+   *
+   * Just because a lower bound is present, it does
+   * not mean that it is checked. This can however
+   * be determined by calling Iterator::check_lower_bound().
+   */
+  virtual const Slice* iterate_lower_bound() const { return nullptr; }
 
   /**
    * true if this iterator is already
    * checking ReadOptions::iterate_upper_bound,
    * false otherwise.
    */
-  virtual bool has_upper_bound() const { return false; }
+  virtual bool check_upper_bound() const { return false; }
+
+  /**
+   * Returns the upper bound if this iterator has a
+   * ReadOptions::iterate_upper_bound set, else
+   * returns nullptr.
+   *
+   * Just because an upper bound is present, it does
+   * not mean that it is checked. This can however
+   * be determined by calling Iterator::check_upper_bound().
+   */
+  virtual const Slice* iterate_upper_bound() const { return nullptr; }
 };
 
 // Return an empty iterator (yields nothing).
--- a/include/rocksdb/iterator.h
+++ b/include/rocksdb/iterator.h
@@ -121,18 +121,14 @@ class Iterator : public Cleanable {
    * checking ReadOptions::iterate_lower_bound,
    * false otherwise.
    */
-  virtual bool has_lower_bound() const {
-    return false;
-  }
+  virtual bool has_lower_bound() const { return false; }
 
   /**
    * true if this iterator is already
    * checking ReadOptions::iterate_upper_bound,
    * false otherwise.
    */
-  virtual bool has_upper_bound() const {
-    return false;
-  }
+  virtual bool has_upper_bound() const { return false; }
 };
 
 // Return an empty iterator (yields nothing).
--- a/include/rocksdb/iterator.h
+++ b/include/rocksdb/iterator.h
@@ -47,7 +47,7 @@ class Iterator : public Cleanable {
   // Valid() after this call iff the source is not empty.
   virtual void SeekToLast() = 0;
 
-  // Position at the first key in the source that at or past target.
+  // Position at the first key in the source that is at or past target.
   // The iterator is Valid() after this call iff the source contains
   // an entry that comes at or past target.
   // All Seek*() methods clear any error status() that the iterator had prior to
@@ -56,7 +56,7 @@ class Iterator : public Cleanable {
   // Target does not contain timestamp.
   virtual void Seek(const Slice& target) = 0;
 
-  // Position at the last key in the source that at or before target.
+  // Position at the last key in the source that is at or before target.
   // The iterator is Valid() after this call iff the source contains
   // an entry that comes at or before target.
   // Target does not contain timestamp.
--- a/include/rocksdb/iterator.h
+++ b/include/rocksdb/iterator.h
@@ -118,8 +118,10 @@ class Iterator : public Cleanable {
 
   /**
    * true if this iterator is already
-   * checking ReadOptions::iterate_lower_bound,
-   * false otherwise.
+   * checking ReadOptions::iterate_lower_bound.
+   * false indicates that either the iterator does
+   * not check the lower bound, or that it does not
+   * report that it is performing the check.
    */
   virtual bool ChecksLowerBound() const { return false; }
 
@@ -136,8 +138,10 @@ class Iterator : public Cleanable {
 
   /**
    * true if this iterator is already
-   * checking ReadOptions::iterate_upper_bound,
-   * false otherwise.
+   * checking ReadOptions::iterate_upper_bound.
+   * false indicates that either the iterator does
+   * not check the upper bound, or that it does not
+   * report that it is performing the check.
    */
   virtual bool ChecksUpperBound() const { return false; }
 
--- a/utilities/write_batch_with_index/write_batch_with_index_test.cc
+++ b/utilities/write_batch_with_index/write_batch_with_index_test.cc
@@ -561,25 +561,25 @@ class KVIter : public Iterator {
   Slice value() const override { return iter_->second; }
   Status status() const override { return Status::OK(); }
 
-  bool check_lower_bound() const override {
+  bool ChecksLowerBound() const override {
     return read_options_ != nullptr &&
            read_options_->iterate_lower_bound != nullptr;
   }
 
-  const Slice* iterate_lower_bound() const override {
-    if (check_lower_bound()) {
+  const Slice* lower_bound() const override {
+    if (ChecksLowerBound()) {
       return read_options_->iterate_lower_bound;
     }
     return nullptr;
   }
 
-  bool check_upper_bound() const override {
+  bool ChecksUpperBound() const override {
     return read_options_ != nullptr &&
            read_options_->iterate_upper_bound != nullptr;
   }
 
-  const Slice* iterate_upper_bound() const override {
-    if (check_upper_bound()) {
+  const Slice* upper_bound() const override {
+    if (ChecksUpperBound()) {
       return read_options_->iterate_upper_bound;
     }
     return nullptr;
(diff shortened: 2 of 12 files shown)


adamretter:
@mrambacher Good shout! It looks like there are some issues with `seekToLast` as well :-( I will look into them...

include/rocksdb/utilities/write_batch_with_index.h
```diff
@@ -153,9 +153,9 @@ class WriteBatchWithIndex : public WriteBatchBase {
   // time.
   //
   // The returned iterator should be deleted by the caller.
-  WBWIIterator* NewIterator(ColumnFamilyHandle* column_family);
+  WBWIIterator* NewIterator(ColumnFamilyHandle* column_family, const ReadOptions* read_options = nullptr);
   // Create an iterator of the default column family.
-  WBWIIterator* NewIterator();
+  WBWIIterator* NewIterator(const ReadOptions* read_options = nullptr);
```
mrambacher, line 158: I am not certain on the coding rules or conventions, but does this need to be marked as "explicit"?

utilities/write_batch_with_index/write_batch_with_index_test.cc
```diff
@@ -1021,6 +1067,29 @@ TEST_F(WriteBatchWithIndexTest,
...
+  // iter->Prev();
+  // ASSERT_OK(iter->status());
+  // ASSERT_TRUE(iter->Valid());
+
+  // AssertIter(iter.get(), "k06", "v06");
+  // iter->Next();
+  // AssertIter(iter.get(), "k07", "k07");
+  // iter->Next();
+  // ASSERT_OK(iter->status());
+  // ASSERT_FALSE(iter->Valid()) << "Should have reached upper_bound";
```
mrambacher, line 1092: Why are these commented out?

include/rocksdb/utilities/write_batch_with_index.h
pdillinger: It's not an actual constructor nor user-defined conversion function, so no.

utilities/write_batch_with_index/write_batch_with_index_test.cc
adamretter: @mrambacher It was a work in progress.

Just to update you on progress - The approach I took initially in this PR was not strong enough, with further tests other problems emerged. I now have many more tests and just one that is now failing. As soon as I have that test passing, I will push my new approach to this PR...

adamretter:
@mrambacher @pdillinger Okay this is a completely reworked approach. Solving the issue turned out to be much more involved than I initially though as there were other hidden problems. The fix once known is relatively simple. This is my 5th attempt at fixing it, as I kept adding tests which would show additional problems.

Let me know your thoughts... Hopefully this is pretty much done now ;-)

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -52,28 +51,56 @@ class BaseDeltaIterator : public Iterator {
...
-    forward_ = false;
-    base_iterator_->SeekToLast();
-    delta_iterator_->SeekToLast();
+    progress_ = Progress::SEEK_TO_LAST;
+
+    // is there an upper bound constraint?
+    if (read_options_ != nullptr &&
+        read_options_->iterate_upper_bound != nullptr) {
+      // yes, and is base_iterator already constrained by an upper_bound?
+      if (!base_iterator_->has_upper_bound()) {
```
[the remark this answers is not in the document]
adamretter: Good point! I have now incorporated the idea that the base and delta iterators may have different (or the same) upper and lower bounds. I also added a new test which demonstrates this `TestIteraratorWithBaseUpperBoundOnBaseAndDifferentUpperBoundOnBatch`

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -240,21 +290,32 @@ class BaseDeltaIterator : public Iterator {
...
+    // upper_bounds_check already
+    return base_iterator_->Valid() &&
+           (base_iterator_->has_upper_bound()
+                ? true
+                : IsWithinBounds(base_iterator_->key()));
+  }
+  bool DeltaValid() const {
+    return delta_iterator_->Valid() &&
+           IsWithinBounds(delta_iterator_->Entry().key);
+  }
```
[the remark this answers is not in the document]
adamretter: Yes, that may be a possible optimisation...

In the first pass I wanted to focus on correctness with as little modification to the current approach as possible. As it is tricky to say the least! I would still like to see this PR merged before optimisations are considered. If there is no upper or lower bound set I don't think this PR will have much (if any) impact on current performance.

I suspect few people can have been using bounds with WBWI - otherwise like me, they would likely have noticed the bugs ;-)

I would prefer that later a subsequent PR could be created which would then add such an optimisation for when bounds are in play.

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -240,21 +290,32 @@ class BaseDeltaIterator : public Iterator {
...
     }
   }
-  bool BaseValid() const { return base_iterator_->Valid(); }
-  bool DeltaValid() const { return delta_iterator_->Valid(); }
+  bool BaseValid() const {
+    // NOTE: we don't need the bounds check on
+    // base_iterator if the base iterator has an
+    // upper_bounds_check already
+    return base_iterator_->Valid() &&
+           (base_iterator_->has_upper_bound()
```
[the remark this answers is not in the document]
adamretter: If `base_iterator_->has_upper_bound() == true`, then the base_iterator is declaring that it knows about and is responsible for managing its own upper bound. This fits well with the fix I just added where the base and delta iterators may have different upper bounds.

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -329,14 +392,47 @@ class BaseDeltaIterator : public Iterator {
...
+    TO_BE_DETERMINED = 0,
+
+    SEEK_TO_FIRST = 1,
+    SEEK = 2,
+    FORWARD = 3,
+
+    BACKWARD = 4,
+    SEEK_FOR_PREV = 5,
+    SEEK_TO_LAST = 6
+  };
```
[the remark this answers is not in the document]
adamretter: Good idea! I have now done so :-)

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -52,28 +51,56 @@ class BaseDeltaIterator : public Iterator {
...
     base_iterator_->Seek(k);
     delta_iterator_->Seek(k);
     UpdateCurrent();
   }
 
   void SeekForPrev(const Slice& k) override {
-    forward_ = false;
+    progress_ = Progress::SEEK_FOR_PREV;
     base_iterator_->SeekForPrev(k);
     delta_iterator_->SeekForPrev(k);
```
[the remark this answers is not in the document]
adamretter: According to `ReadOptions::iterate_upper_bound` is only in play for forward iterators. As far as I understand it, `SeekForPrev` results in a backwards iterator.

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -52,28 +51,56 @@ class BaseDeltaIterator : public Iterator {
...
+    // is there an upper bound constraint?
+    if (read_options_ != nullptr &&
+        read_options_->iterate_upper_bound != nullptr) {
+      // yes, and is base_iterator already constrained by an upper_bound?
+      if (!base_iterator_->has_upper_bound()) {
+        // no, so we have to seek it to before iterate_upper_bound
+        base_iterator_->Seek(*(read_options_->iterate_upper_bound));
+        if (base_iterator_->Valid()) {
+          base_iterator_->Prev();  // upper bound should be exclusive!
+        }
```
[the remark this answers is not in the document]
adamretter: You are right.. this does need changing... but I think the `Prev` has to be conditional on whether we found the upper bound.

I have now modified the code to:

```
        base_iterator_->Seek(*(base_upper_bound));
        if (base_iterator_->Valid()) {
          base_iterator_->Prev();  // upper bound should be exclusive!
        } else {
          // the base_upper_bound is beyond the base_iterator, so just SeekToLast()
          base_iterator_->SeekToLast();
        }
```

I have also fixed the same issue with the `delta_iterator`. To check this I also added the tests:
1. `TestIteraratorWithBaseNoSuchUpperBoundOnBaseAndBatch`
2. `TestIteraratorWithBaseOverUpperBoundOnBaseWithBaseConstraintAndBatch`
3. `TestIteraratorWithBaseOverUpperBoundOnBaseWithoutBaseConstraintAndBatch`

include/rocksdb/iterator.h
```diff
@@ -123,14 +123,36 @@ class Iterator : public Cleanable {
    * checking ReadOptions::iterate_lower_bound,
    * false otherwise.
    */
-  virtual bool has_lower_bound() const { return false; }
+  virtual bool check_lower_bound() const { return false; }
```
mrambacher, line 126: I am a little confused as to when/how these methods will be used.  They are new and the only Iterators that implement them are the test ones (even the BaseDelta does not implement it).  It seems like we might need a solution without these new methods, as all non-test solutions would effectively be without them.  Or am I missing something?

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -61,18 +61,17 @@ class BaseDeltaIterator : public Iterator {
...
-        read_options_->iterate_upper_bound != nullptr) {
+    const Slice* base_upper_bound = base_iterator_upper_bound();
+    if (base_upper_bound != nullptr) {
       // yes, and is base_iterator already constrained by an upper_bound?
-      if (!base_iterator_->has_upper_bound()) {
-        // no, so we have to seek it to before iterate_upper_bound
-        base_iterator_->Seek(*(read_options_->iterate_upper_bound));
+      if (!base_iterator_->check_upper_bound()) {
+        // no, so we have to seek it to before base_upper_bound
+        base_iterator_->Seek(*(base_upper_bound));
```
mrambacher, line 69: I think the case/code for base_iterator_ and delta_iterator_ should be the same.  Seek to the read_options->upper_bound.  If the iterator is not valid, SeekToLast, otherwise Prev.

I do not think the new methods are required

include/rocksdb/iterator.h
adamretter: I can certainly remove them. They are there to allow an optimisation where when you have an iterator wrapping an iterator (several places in Rocks) if your underlying iterator handles upper and lower bounds, you don't need to check them again.

I think it would be better to keep them, and expand the use of them. For example, `ForwardIterator` and `BackwardIterator` should likely implement them as they already handle upper and lower bounds checks respectively.

utilities/write_batch_with_index/write_batch_with_index.cc
adamretter: I am not sure I follow you here... the code for `base_iterator_` and `delta_iterator_` are the same apart from the case where the `base_iterator_` might already handle upper/lower bounds.
I did look at making `delta_iterator_` also handle the upper/lower bounds directly itself, but it was too difficult as it is a very different type of iterator which is really just concerned with the SkipList really.

I am fairly certain I have the logic correct as I kept adding tests to ensure this. However, of course it is indeed possible that I missed something... can you see a test case where the logic doesn't hold?

Perhaps it is easier to have a call to discuss?

adamretter:
> I think SeekToPrev is still broken if k is > upper bound.

I will have to check that one, I mostly focused on SeekToLast.

include/rocksdb/iterator.h
mrambacher: Should the BaseDeltaIterator implement these functions?  

Not to complicate your world (and I am in no way suggesting you fix this, perhaps document it) but if the BaseIterator had bounds on it but the ReadOptions did not, wouldn't you potentially get really strange results?  That sounds more like something that should be an assert somewhere...

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -52,28 +51,62 @@ class BaseDeltaIterator : public Iterator {
...
+          base_iterator_->SeekToLast();
+        }
+      } else {
+        // yes, so the base_iterator will take care of base_upper_bound
+        base_iterator_->SeekToLast();
+      }
+
+      // delta iterator does not itself support iterate_upper_bound,
+      // so we have to seek it to before iterate_upper_bound
+      delta_iterator_->Seek(*(read_options_->iterate_upper_bound));
```
mrambacher, line 84: I think this value could be nullptr, if the base iterator had an upper bound but the read_options did not.

include/rocksdb/iterator.h
adamretter: The ReadOptions is passed to the BaseIterator - so I am not quite sure how you mean that one could have bounds but not the other?

mrambacher: That is only true in the test code.  Outside of tests ,the "NewIteratorFromBase" takes in an Iterator, which becomes the base iterator.  In this case, the base iterator could have very different bounds than the DeltaIterator.

Or am I missing something?

I am not suggesting that should be legal BTW, only if it happens I do not know what the results would look like.

adamretter: Hmmm I will add some tests for that too...

adamretter: So this is already handled, and I just added the tests: `TestIteraratorWithBaseUpperBoundOnBaseNoUpperBoundOnBatch` `TestIteraratorWithBaseNoUpperBoundOnBaseUpperBoundOnBatch` for further evidence of this :-)

utilities/write_batch_with_index/write_batch_with_index.cc
adamretter: I have fixed that one now too

include/rocksdb/iterator.h
```diff
@@ -117,6 +117,42 @@ class Iterator : public Cleanable {
     assert(false);
     return Slice();
   }
+
+  /**
+   * true if this iterator is already
+   * checking ReadOptions::iterate_lower_bound,
+   * false otherwise.
+   */
+  virtual bool check_lower_bound() const { return false; }
```
pdillinger, line 126: `check_lower_bound` is not an existing or new stateful member, so why name it like an accessor with a verb name? How about `ChecksLowerBound` or `DoesCheckLowerBound`?

include/rocksdb/iterator.h
```diff
@@ -117,6 +117,42 @@ class Iterator : public Cleanable {
...
+  /**
+   * Returns the lower bound if this iterator has a
+   * ReadOptions::iterate_lower_bound set, else
+   * returns nullptr.
+   *
+   * Just because a lower bound is present, it does
+   * not mean that it is checked. This can however
+   * be determined by calling Iterator::check_lower_bound().
+   */
+  virtual const Slice* iterate_lower_bound() const { return nullptr; }
```
pdillinger, line 137: Also a verb, where the reader is supposed to understand from the use of lowercase_with_underscores that this is an accessor not an action. Since we're already in an iterator, can we just call it `lower_bound`?

include/rocksdb/iterator.h
```diff
@@ -117,6 +117,42 @@ class Iterator : public Cleanable {
...
+
+  /**
+   * true if this iterator is already
+   * checking ReadOptions::iterate_upper_bound,
+   * false otherwise.
+   */
+  virtual bool check_upper_bound() const { return false; }
```
pdillinger, line 144: see above

include/rocksdb/iterator.h
```diff
@@ -117,6 +117,42 @@ class Iterator : public Cleanable {
...
+  /**
+   * Returns the upper bound if this iterator has a
+   * ReadOptions::iterate_upper_bound set, else
+   * returns nullptr.
+   *
+   * Just because an upper bound is present, it does
+   * not mean that it is checked. This can however
+   * be determined by calling Iterator::check_upper_bound().
+   */
+  virtual const Slice* iterate_upper_bound() const { return nullptr; }
```
pdillinger, line 155: see above

adamretter:
@pdillinger I have renamed the members as you suggested. Thanks

include/rocksdb/iterator.h
```diff
@@ -117,6 +117,42 @@ class Iterator : public Cleanable {
     assert(false);
     return Slice();
   }
+
+  /**
+   * true if this iterator is already
+   * checking ReadOptions::iterate_lower_bound,
+   * false otherwise.
+   */
+  virtual bool ChecksLowerBound() const { return false; }
```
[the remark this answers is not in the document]
adamretter: > Still wow. Baking into a API a way to say "I have a bug and here's what it is. If you wrap me, can you please fix it?"

That's not what it does at all.

If an Iterator which is wrapped supports upper and lower bounds checks, it can inform the wrapping iterator so that the wrapping iterator need not worry about it. It won't cause any harm if they both make the checks. In the past the wrapping iterator had no way of knowing whether the underlying iterator was also making the checks, so it would always have to make the checks itself as well.

adamretter:
@pdillinger I believe I have added the comments and amended the PR description now as you requested.

include/rocksdb/iterator.h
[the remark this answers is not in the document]
pdillinger: I still don't understand the use case for exposing what your lower bound is without checking it. If that's a transitional state in migrating code, is it really necessary as part of a public API? Actually, should these functions be protected instead?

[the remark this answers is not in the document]
pdillinger: Nor do I think an iterator needs to be able to express "I would be checking bounds if I had one."

My contention is still that we only need optional upper and lower bounds. If an iterator is simply the UNION of several other iterators, its lower/upper bound would be the smallest/largest (resp.) bound of its constituents, where no bound is considered smaller/larger than any specific value. We don't have to do any addition checking to represent that UNION.

If an iterator is INTERSECTED with a range, we can ignore that additional constraint if our existing lower and upper bounds are within that range. (Performance optimization.) But if our lower/upper bound is outside of that range constraint, we have to use the lower and/or upper range bound as our refined lower/upper bound AND check that bound ourselves.

Perhaps it is helpful to think of the union and intersection roles of a wrapping iterator independently as I have described.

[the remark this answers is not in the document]
mrambacher: @pdillinger I think I agree with your sentiment here and personally am not a fan of these functions -- partially because they are not implemented by more of the Iterator hierarchy. 

Having said that, it seems like there should be a means to tell if an Iterator should check its bounds or assume the parent iterator can do it instead.  I am not sure if these functions are the proper ones to fulfill that contract.  I think the Iterator paradigm should have some sort of "Filter" to allow filtering of bounds (and prefixes).  This might allow some better optimizations as well.  But I am not sure that a "Filter" is in scope for this PR and think it would require more thought to implement properly.

Also, shouldn't the math you stated above be the largest lower bound and smallest upper?

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -172,6 +233,18 @@ class BaseDeltaIterator : public Iterator {
     return delta_iterator_->status();
   }
 
+  bool ChecksLowerBound() const override { return false; }
```
pdillinger, line 236: Do you mean "return true"? If so, I still question why we need the ChecksLower/UpperBound functions.

If you do mean "return false," can you explain?

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -329,14 +412,105 @@ class BaseDeltaIterator : public Iterator {
...
+  inline const Slice* base_iterator_lower_bound() const {
+    const Slice* lower_bound = base_iterator_->lower_bound();
+    if (lower_bound == nullptr && read_options_ != nullptr) {
+      return read_options_->iterate_lower_bound;
+    }
+    return lower_bound;
+  }
+
+  bool BaseIsWithinBounds() const {
+    const Slice* lower_bound = base_iterator_lower_bound();
```
pdillinger, line 448: Seems for efficiency we should check IsMovingBackward() before looking up lower bound. (Same for forward and upper bound.)

utilities/write_batch_with_index/write_batch_with_index.cc
adamretter: I do mean false. This PR is about implementing support for the upper_bound, much of the machinery is there for the lower_bound because it is common, but my focus was implementing for the upper_bound.

utilities/write_batch_with_index/write_batch_with_index.cc
adamretter: Thanks @pdillinger that's a good suggestion. I have now pushed an additional commit for that.

pdillinger:
From private discussion:

> (me) I'm still concerned about "polluting" the public Iterator API with implementation concerns.

> (Adam) ... I also just noticed that there is an interface InternalIteratorBase which has at least one method that looks very similar to what I am trying to do: UpperBoundCheckResult . However it does not extend Iterator, ...

It looks like @siying wrote that code pretty recently in #7200. Any help/advice Siying?

include/rocksdb/iterator.h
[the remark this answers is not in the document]
siying: I agree with @pdillinger and @mrambacher . The interface of iterator should be made simpler, rather than more complicated. The interface is already in the confusing part, so we should not make it even more confusing. We do have some quite tricky functions in InternalIterator for performance reasons. They are not ideal but having an internal implementation a little bit messier is relatively much more acceptable than make public interface complicated.

include/rocksdb/iterator.h
```diff
@@ -117,6 +117,46 @@ class Iterator : public Cleanable {
...
+  virtual const Slice* lower_bound() const = 0;
+
+  /**
+   * true if this iterator is already
+   * checking ReadOptions::iterate_upper_bound.
+   * false indicates that either the iterator does
+   * not check the upper bound, or that it does not
+   * report that it is performing the check.
+   */
+  virtual bool ChecksUpperBound() const = 0;
```
siying, line 148: Can you explain it? RocksDB's Iterator is always supposed to have upper bound checked. Did you see a bug somewhere?

adamretter: I didn't find that to be the case. There are iterators which don't check upper/lower bounds, which is what triggered this PR. When you wrap an iterator, if the wrapped iterator does not implement upper/lower bounds checks, you may need to know that, so that you can implement the checks in your wrapper iterator to get the behaviour you need.

cbi42:
Implemented in #11680

PR #7216, opened 2020-08-03
Fix Timer unable to schedule new added job

jay-zhuang:
And added test to reproduce the problem.

### DIFF
--- a/util/timer_test.cc
+++ b/util/timer_test.cc
@@ -239,14 +239,13 @@ TEST_F(TimerTest, MultipleScheduleRepeatedlyTest) {
 
 TEST_F(TimerTest, AddAfterStartTest) {
   const int kIterations = 5;
-  uint64_t time_counter = 0;
   InstrumentedMutex mutex;
   InstrumentedCondVar test_cv(&mutex);
 
-  // wait timer to running and then add a new task
-  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->LoadDependency(
+  // wait timer to run and then add a new job
+  SyncPoint::GetInstance()->LoadDependency(
       {{"Timer::Run::Waiting", "TimerTest:AddAfterStartTest:1"}});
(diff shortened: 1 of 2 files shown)


util/timer_test.cc
```diff
@@ -237,6 +237,49 @@ TEST_F(TimerTest, MultipleScheduleRepeatedlyTest) {
...
+    while (count < kIterations) {
+      time_counter += kSecond;
+      mock_env_->set_current_time(time_counter);
+      test_cv.TimedWait(time_counter);
+    }
+  }
+
+  ASSERT_TRUE(timer.Shutdown());
+
+  ASSERT_EQ(5, count);
```
riversand963, line 280: Nit: `ASSERT_EQ(kIterations, count)`

util/timer_test.cc
```diff
@@ -237,6 +237,49 @@ TEST_F(TimerTest, MultipleScheduleRepeatedlyTest) {
   ASSERT_EQ(count2, 5);
 }
 
+TEST_F(TimerTest, AddAfterStartTest) {
+  const int kIterations = 5;
+  uint64_t time_counter = 0;
```
riversand963, line 242: This line `uint64_t time_counter = 0` can be moved closer to where it is first used.

util/timer_test.cc
```diff
@@ -237,6 +237,49 @@ TEST_F(TimerTest, MultipleScheduleRepeatedlyTest) {
...
+  InstrumentedMutex mutex;
+  InstrumentedCondVar test_cv(&mutex);
+
+  // wait timer to running and then add a new task
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->LoadDependency(
+      {{"Timer::Run::Waiting", "TimerTest:AddAfterStartTest:1"}});
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->EnableProcessing();
```
riversand963, line 249: The preceding `ROCKSDB_NAMESPACE` is not necessary.

PR #7221, opened 2020-08-06
Always try to range sync after WriteBuffered

tabokie:
Signed-off-by: tabokie <xy.tao@outlook.com>

Fix a case where large write go through WriteBuffered without being range synced afterwards. Also simplify the logic of `WritableFileWriter::Append()`.


tabokie:
CI reports `Failed to check Status` for EnvLoggerTest.

yiwu-arbug:
> Fix a case where large write go through `WriteBuffered` without being range synced afterwards.

@ajkr do you mind take a look and double check if this is the case?

PR #7224, opened 2020-08-06
Fix delete triggered compaction for single level universal

anand1976:
Delete triggered compaction (DTC) for universal compaction style with ```num_levels = 1``` has been disabled for sometime due to a data correctness bug. This PR re-enables it with a bug fix. A file marked for compaction can be picked, along with all L0 files after it as the compaction input. We stop adding files to the input once we encounter a file already being compacted (the original bug failed to check the compaction status of the files).

Tests:
Add unit tests to ```compaction_picker_test.cc```

### DIFF
--- a/db/compaction/compaction_picker_universal.cc
+++ b/db/compaction/compaction_picker_universal.cc
@@ -859,7 +859,8 @@ Compaction* UniversalCompactionBuilder::PickDeleteTriggeredCompaction() {
       FileMetaData* f = vstorage_->LevelFiles(0)[loop];
       if (f->marked_for_compaction) {
         start_level_inputs.files.push_back(f);
-        start_index = static_cast<int>(loop);  // Consider this as the first candidate.
+        start_index =
+            static_cast<int>(loop);  // Consider this as the first candidate.
         break;
       }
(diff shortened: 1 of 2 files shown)


db/compaction/compaction_picker_universal.cc
```diff
@@ -841,34 +841,47 @@ Compaction* UniversalCompactionBuilder::PickDeleteTriggeredCompaction() {
...
-      if (f->marked_for_compaction) {
-        compact = true;
+    // Find the first file marked for compaction. Ignore the last file
+    for (size_t loop = 0; loop + 1 < sorted_runs_.size(); loop++) {
+      SortedRun* sr = &sorted_runs_[loop];
+      if (sr->being_compacted) {
+        continue;
       }
-      if (compact) {
+      FileMetaData* f = vstorage_->LevelFiles(0)[loop];
```
mrambacher, line 859: I am confused by this code.  Looking at how sorted_runs_ is built, it appears that the value for loop could be greater than the depth of LevelFiles[0].  Won't this run off the end of the LevelFiles(0)[] array?

PR #7227, opened 2020-08-07
Get() with timestamp should respect snapshot

riversand963:
If user-defined timestamp is enabled, current implementation can expose
newer data to queries even if an older sequence number is specified via
read_options.snapshot. This PR makes Get() respect sequence-number-based
snapshot.

Solution is simple. Besides using <ukey, ts, seq> to search the index for the key,
we also verify that the candidate result's seq is smaller than or equal to seq. This
requires passing a seq via `GetContext`, which results in the majority of code
change caused by this PR.

Also added a few unit tests to demonstrate standard visibility during point lookup
and range scan when timestamp and snapshot are both present.

Test plan (devserver):
```
make check
$./db_bench --benchmarks=fillseq,readrandom -cache_size=$[64*1024*1024]
```
Result
this PR: readrandom   :       4.827 micros/op 207180 ops/sec;   22.9 MB/s (1000000 of 1000000 found)
master:  readrandom   :       4.936 micros/op 202610 ops/sec;   22.4 MB/s (1000000 of 1000000 found)

### DIFF
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -601,7 +601,8 @@ class DataVisibilityTest : public DBBasicTestWithTimestampBase {
 //                               seq'=11
 //                               write finishes
 //         GetImpl(ts,seq)
-// It is OK to return <k, t1, s1> if ts>=t1 AND seq>=s1.
+// It is OK to return <k, t1, s1> if ts>=t1 AND seq>=s1. If ts>=1t1 but seq<s1,
+// the key should not be returned.
 TEST_F(DataVisibilityTest, PointLookupWithoutSnapshot1) {
   Options options = CurrentOptions();
   const size_t kTimestampSize = Timestamp(0, 0).size();
@@ -648,7 +649,8 @@ TEST_F(DataVisibilityTest, PointLookupWithoutSnapshot1) {
 //                               write finishes
 //                               Flush
 //         GetImpl(ts,seq)
-// It is OK to return <k, t1, s1> if ts>=t1 AND seq>=s1.
+// It is OK to return <k, t1, s1> if ts>=t1 AND seq>=s1. If ts>=t1 but seq<s1,
+// the key should not be returned.
 TEST_F(DataVisibilityTest, PointLookupWithoutSnapshot2) {
   Options options = CurrentOptions();
   const size_t kTimestampSize = Timestamp(0, 0).size();
@@ -809,7 +811,8 @@ TEST_F(DataVisibilityTest, PointLookupWithSnapshot2) {
 //                            seq'=11
 //                            write finishes
 //      scan(ts,seq)
-// <k, t1, s1> can be seen in scan as long as ts>=t1 AND seq>=s1.
+// <k, t1, s1> can be seen in scan as long as ts>=t1 AND seq>=s1. If ts>=t1 but
+// seq<s1, then the key should not be returned.
 TEST_F(DataVisibilityTest, RangeScanWithoutSnapshot) {
   Options options = CurrentOptions();
   const size_t kTimestampSize = Timestamp(0, 0).size();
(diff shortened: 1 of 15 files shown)


db/db_with_timestamp_basic_test.cc
```diff
@@ -588,6 +588,326 @@ TEST_F(DBBasicTestWithTimestamp, CompactDeletionWithTimestampMarkerToBottom) {
...
+
+// Application specifies timestamp but not snapshot.
+//           reader              writer
+//                               ts'=90
+//           ts=100
+//           seq=10
+//                               seq'=11
+//                               write finishes
+//         GetImpl(ts,seq)
+// It is OK to return <k, t1, s1> if ts>=t1 AND seq>=s1.
```
ltamasi, line 604: I feel this could be made clearer by wording it differently, e.g. by emphasizing that the key-value should *not* be returned because of the (implicit) snapshot taken at seq 10. (Similarly for other test cases below.)

db/db_with_timestamp_basic_test.cc
```diff
@@ -588,6 +588,326 @@ TEST_F(DBBasicTestWithTimestamp, CompactDeletionWithTimestampMarkerToBottom) {
...
+  Iterator* it = db_->NewIterator(read_opts);
+  ASSERT_NE(nullptr, it);
+  writer_thread.join();
+  int count = 0;
+  for (it->SeekToFirst(); it->Valid(); it->Next(), ++count) {
+    ASSERT_OK(it->status());
+    ASSERT_EQ("key" + std::to_string(count), it->key());
+    ASSERT_EQ("value" + std::to_string(count), it->value());
+  }
+  ASSERT_EQ(0, count);
```
ltamasi, line 851: We could just call `SeekToFirst`, `ASSERT_FALSE(it->Valid())` and call it a day here, right?

db/db_with_timestamp_basic_test.cc
```diff
@@ -588,6 +588,326 @@ TEST_F(DBBasicTestWithTimestamp, CompactDeletionWithTimestampMarkerToBottom) {
...
+  Iterator* it = db_->NewIterator(read_opts);
+  ASSERT_NE(nullptr, it);
+  it->Seek("key0");
+  ASSERT_FALSE(it->Valid());
+
+  delete it;
+  db_->ReleaseSnapshot(snap);
+  Close();
+}
+
```
ltamasi, line 910: We could consider adding a "positive" test case/cases that makes sure KVs are returned if they do satisfy both the timestamp and the sequence number condition.

db/memtable.cc
```diff
@@ -669,6 +669,13 @@ static bool SaveValue(void* arg, const char* entry) {
     if (!s->CheckCallback(seq)) {
       return true;  // to continue to the next seq
     }
+    if (ts_sz > 0) {
+      assert(s->key);
+      SequenceNumber seq_ub = s->key->sequence(/*type=*/nullptr);
+      if (seq_ub < seq) {
+        return true;  // to continue to the next seq
+      }
+    }
```
ltamasi, line 678: I'm wondering if this functionality could be implemented using the existing `ReadCallback` interface? It seems to be just the thing we need, and it would also eliminate a whole bunch of code changes in `GetContext`'s clients (and also in `LookupKey` above).

db/db_with_timestamp_basic_test.cc
riversand963: I added a sentence to emphasize this case here and in other tests.

db/db_with_timestamp_basic_test.cc
riversand963: Yes.

db/db_with_timestamp_basic_test.cc
riversand963: I think all other tests in this file are positive tests. The purpose of this PR is to give some "negative" cases.

db/memtable.cc
riversand963: Good idea. I followed the suggestion and made the PR shorter. The only issue is that we need to pass a read callback to `MultiGet`, which I think can be done in another PR.

db/memtable.cc
```diff
@@ -669,6 +669,12 @@ static bool SaveValue(void* arg, const char* entry) {
     if (!s->CheckCallback(seq)) {
       return true;  // to continue to the next seq
     }
+    if (ts_sz > 0) {
+      assert(s->key);
+      if (s->callback_ && !s->callback_->IsVisible(seq)) {
+        return true;  // to continue to the next seq
+      }
+    }
```
ltamasi, line 677: I think you can remove this as well since `s->CheckCallback` above does the same thing.

riversand963: Right, I did the same for `GetContext::SaveValue`, but forgot this. Will update.

PR #7228, opened 2020-08-08
Timer should run scheduled function without mutex

riversand963:
Summary:
Timer (defined in timer.h) schedules and runs user-specified fuctions
regularly. Current implementation holds the mutex while running user
function, which will lead to contention and waiting.
To fix, Timer::Run releases mutex before running user function, and
re-acquires it afterwards.
This fix will impact how we can cancel a task. If the task is running,
it is not holding the mutex. The thread calling Cancel() should wait
until the current task finishes.

Test Plan (devserver):
make check

### DIFF
--- a/util/timer.h
+++ b/util/timer.h
@@ -118,6 +118,7 @@ class Timer {
       }
 
       FunctionInfo* current_fn = heap_.top();
+      assert(current_fn);
 
       if (!current_fn->IsValid()) {
         heap_.pop();
@@ -140,7 +141,8 @@ class Timer {
         // So current_fn is still a valid ptr.
         heap_.pop();
 
-        if (current_fn->repeat_every_us > 0) {
+        // current_fn may be cancelled already.
+        if (current_fn->IsValid() && current_fn->repeat_every_us > 0) {
           current_fn->next_run_time_us = env_->NowMicros() +
               current_fn->repeat_every_us;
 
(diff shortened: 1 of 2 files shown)


util/timer.h
```diff
@@ -156,6 +158,8 @@ class Timer {
     }
 
     while (!heap_.empty()) {
+      const FunctionInfo& top = *(heap_.top());
+      WaitForTaskCompleteIfNecessary(top);
       heap_.pop();
```
jay-zhuang, line 163: There's a race condition during shutdown that it may `pop` a running function. As the `heap_` may change during `cond_var_.wait()` inside of `WaitForTaskCompleteIfNecessary()`. The `top` here may already be `pop`ed out and `heap_.top()` is the second function in the queue, which may be running. Here is a test to reproduce it: https://github.com/jay-zhuang/rocksdb/commit/e46414512a78bf94fdbdde04906c8ff5722a361b

riversand963: Good catch. Will fix.

util/timer.h
```diff
@@ -151,14 +155,34 @@ class Timer {
...
+#ifndef NDEBUG
+      const FunctionInfo* func_info = heap_.top();
+      assert(func_info);
+      assert(!func_info->IsExecuting());
+#endif  // !NDEBUG
       heap_.pop();
     }
-
     map_.clear();
   }
```
jay-zhuang, line 187: Not sure if it's a valid usecase: what if there're 2 threads trying to shutdown at the same time? For example, 1st thread try to shutdown, which may wait in `WaitForTaskCompleteIfNecessary()`, then 2nd thread try to call `shutdown` too, which stops in `WaitForTaskCompleteIfNecessary()` too if the function is still running. One of them will awake first and clean up all functions. Then the other will check invalid `fn.executing` when it's awaken eventually.

riversand963: Will update.

util/timer.h
```diff
@@ -128,16 +141,22 @@ class Timer {
...
 
         // Remove the work from the heap once it is done executing.
         // Note that we are just removing the pointer from the heap. Its
         // memory is still managed in the map (as it holds a unique ptr).
         // So current_fn is still a valid ptr.
         heap_.pop();
 
-        if (current_fn->repeat_every_us > 0) {
+        // current_fn may be cancelled already.
+        if (current_fn->IsValid() && current_fn->repeat_every_us > 0) {
```
jay-zhuang, line 159: should we add check for running, like:
```suggestion
        if (running_ && current_fn->IsValid() && current_fn->repeat_every_us > 0) {
```

riversand963: It's a good practice to check due to the preceding mutex release. In this case, if `current_fn->IsValid()` is true, then `running_` must still be true. This is because `running_` is set to `false` and all `fn->valid` are set to `false` in `Shutdown` before releasing the mutex.
I can add this.

PR #7229, opened 2020-08-10
Add CircleCI for tests on non-shm

riversand963:
As title.

Test plan:
Watch for CircleCI results.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -261,6 +261,13 @@ jobs:
     machine:
       image: ubuntu-1604:201903-01
     resource_class: 2xlarge
+    parameters:
+      start_test:
+        default: ""
+        type: string
+      end_test:
+        default: ""
+        type: string
     steps:
       - checkout # checkout the code in the project directory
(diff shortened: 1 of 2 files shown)


riversand963:
> Oh wow, we plan to run all tests on /tmp? Is it necessary? I don't have objection to get some quick start though. Another question is whether 40 minutes are short enough?
> 
> It's good to go anyway. We can always improve later.

Thanks @siying for the review.
It's not necessary to run all tests on /tmp, and I need to specify which tests to run. This is not trivial in two folds.
1. Determine which tests are not necessary. Perhaps tests that run completely in DRAM.
2. Specify the set of tests in Makefile and CircleCI config. We currently use `SUBSET`, `ROCKSDBTESTS_START` and `ROCKSDBTESTS_END` to pick some tests listed in `src.mk`. The selected tests may be randomly distributed in this list, thus using these macros may be tedious.

I can shift the load between non-shm1 and non-shm2 so that their execution time is more even.

PR #7230, opened 2020-08-10
Add circleci tests with no compression

riversand963:
As title.

Test plan
Check the result of CircleCI tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -257,6 +257,27 @@ jobs:
           command: |
             OPT=-DTRAVIS V=1 make -j4 static_lib && cd examples && make -j4 | ../.circleci/cat_ignore_eagain
 
+  build-no-compression:
+    machine:
+      image: ubuntu-1604:201903-01
+    resource_class: 2xlarge
+    steps:
+      - checkout
+      - run: pyenv global 3.5.2
+      - run: sudo apt-get update -y && sudo apt-get install -y libgflags-dev
+      - run:
+          name: "Build and run unit tests with no compression"
+          command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -257,6 +257,27 @@ jobs:
...
+      - run: pyenv global 3.5.2
+      - run: sudo apt-get update -y && sudo apt-get install -y libgflags-dev
+      - run:
+          name: "Build and run unit tests with no compression"
+          command: |
+            export TEST_TMPDIR=/dev/shm/rocksdb
+            rm -rf /dev/shm/rocksdb
+            mkdir /dev/shm/rocksdb
+            make clean
+            cat build_tools/fbcode_config.sh | grep -iv dzstd | grep -iv dzlib | grep -iv dlz4 | grep -iv dsnappy | grep -iv dbzip2 > .tmp.fbcode_config.sh
```
pdillinger, line 275: There are make flags like ROCKSDB_DISABLE_BZIP to avoid all this ugliness. (Actually they have to be set as environment variables not make flags, since they aren't exported in the Makefile. :( )

.circleci/config.yml
```diff
@@ -257,6 +257,27 @@ jobs:
...
+            mv .tmp.fbcode_config.sh build_tools/fbcode_config.sh
+            cat Makefile | grep -v tools/ldb_test.py > .tmp.Makefile
```
pdillinger, line 277: Do you know why this test is excluded? Can't we make the test work on all valid build configurations?

riversand963:
Closing this one due to #9850

PR #7236, opened 2020-08-10
Track WAL in MANIFEST: add method to check WAL consistency

cheng-chang:
Add a method `CheckWals` in `WalSet` to check the logs on disk. See `CheckWals`'s comments.
This method will be used to check consistency of WALs during DB recovery.

Test Plan:
a set of tests are added to wal_edit_test.cc.


db/wal_edit_test.cc
```diff
@@ -118,6 +119,111 @@ TEST(WalSet, DeleteNonClosedWal) {
               std::string::npos);
 }
 
+class WalSetTest : public testing::Test {
+ public:
+  void SetUp() override {
+    env_ = Env::Default();
```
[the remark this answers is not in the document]
cheng-chang: I tried to extend DBTestBase directly.

cheng-chang:
@riversand963 PTAL

db/wal_edit.cc
```diff
@@ -172,4 +172,41 @@ Status WalSet::DeleteWals(const WalDeletions& wals) {
...
+      break;
+    }
+
+    if (wal_meta.HasSize()) {
+      uint64_t log_file_size = 0;
+      s = env->GetFileSize(logs_on_disk.at(log_number), &log_file_size);
+      if (!s.ok()) {
+        break;
+      }
+      if (wal_meta.GetSizeInBytes() != log_file_size) {
```
riversand963, line 198: Per our discussion offline, I think this can be relaxed. The on-disk WAL after a crash can be larger than what was recorded in MANIFEST.
Also, depending on when you write the first `WalEdit` for a WAL file, we should take different actions.
- If you write and sync MANIFEST when you first created a WAL file. Then it is possible for a WAL to be recorded in MANIFEST but not on disk.
- If you write and sync MANIFEST when you sync the WAL, then it's different.

riversand963 (changes requested):

db/wal_edit.cc
cheng-chang: Added a "synced" field to WalMetadata, if the WAL or WAL directory is not synced, then we skip checking the WAL.

riversand963: Is the addition of "synced" field pushed?

cheng-chang: added syncing related information to WalMetadata: https://github.com/facebook/rocksdb/pull/7414

cheng-chang:
Will update the PR after https://github.com/facebook/rocksdb/pull/7414 is merged.

db/wal_edit.h
```diff
@@ -150,6 +151,16 @@ class WalSet {
 
   const std::map<WalNumber, WalMetadata>& GetWals() const { return wals_; }
 
+  // Checks whether there are missing or corrupted WALs.
+  // Returns Status::OK if there is no missing nor corrupted WAL,
+  // otherwise returns Status::Corruption.
```
riversand963, line 156: The return value can be that of `GetFileSize()`, not just `Status::Corruption()`.

riversand963 (approved):
LGTM. Thanks @cheng-chang for the PR!

PR #7238, opened 2020-08-11
MOD: allow to use hard link when creating the backup

git-hulk:
hi, does it make sense to allow us to use the hard link when copy file when the backup and DB dir in the same file system? we use the backup to replication in [https://github.com/bitleak/kvrocks](https://github.com/bitleak/kvrocks) and it may cause high io usage when creating the backup.

### DIFF
--- a/env/env.cc
+++ b/env/env.cc
@@ -397,6 +397,7 @@ void AssignEnvOptions(EnvOptions* env_options, const DBOptions& options) {
       options.writable_file_max_buffer_size;
   env_options->allow_fallocate = options.allow_fallocate;
   env_options->strict_bytes_per_sync = options.strict_bytes_per_sync;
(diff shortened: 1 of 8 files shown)


git-hulk:
@siying can you take time to review this PR

pdillinger:
We'll have to consider this carefully in relation to other recent and ongoing changes to BackupEngine.

Without a compatibility-breaking schema change to backups, we need to know the crc32c checksum of all backed up files for the backup manifest. Unless crc32c happens to be the file checksum function configured for the DB, we need to read the whole files to compute those checksums. Saving the writes and storage on copy with hard links is good, though.

A change like this needs good unit test coverage.

git-hulk:
@pdillinger, I'm sorry for didn't take a look at  the other parts in copy file which has calculated the file checksum, close the PR

PR #7239, opened 2020-08-11
Travis: fail fast on install failures

pdillinger:
Summary: A recent build continued with confusing results after failing
to "snap install cmake" so ensure failures in installs are fatal. Also
upgrade snapd before "snap install" to hopefully avoid this error:

    error: cannot perform the following tasks:
    - Mount snap "cmake" (513) (snap "cmake" assumes unsupported features: command-chain (try to update snapd and refresh the core snap))

Test Plan: watch Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -239,7 +239,7 @@ install:
       CC=gcc-4.8 && CXX=g++-4.8;
     fi
   - if [[ "${JOB_NAME}" == cmake* ]] && [ "${TRAVIS_OS_NAME}" == linux ]; then
-      sudo snap install cmake --beta --classic || exit $?;
+      sudo apt-get install snapd && sudo snap install cmake --beta --classic || exit $?;
       export PATH=/snap/bin:$PATH;
     fi
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -224,33 +224,33 @@ install:
       PATH=$PATH:/usr/local/opt/ccache/libexec;
     fi
   - if [ "${JOB_NAME}" == cmake-gcc8 ]; then
-      sudo apt-get install -y g++-8;
+      sudo apt-get install -y g++-8 || exit $?;
       CC=gcc-8 && CXX=g++-8;
     fi
   - if [ "${JOB_NAME}" == cmake-gcc9 ] || [ "${JOB_NAME}" == cmake-gcc9-c++20 ]; then
-      sudo apt-get install -y g++-9;
+      sudo apt-get install -y g++-9 || exit $?;
       CC=gcc-9 && CXX=g++-9;
     fi
   - if [ "${JOB_NAME}" == cmake-mingw ]; then
-      sudo apt-get install -y mingw-w64 ;
+      sudo apt-get install -y mingw-w64 || exit $?;
     fi
   - if [ "${JOB_NAME}" == make-gcc4.8 ]; then
-      sudo apt-get install -y g++-4.8 ;
+      sudo apt-get install -y g++-4.8 || exit $?;
       CC=gcc-4.8 && CXX=g++-4.8;
     fi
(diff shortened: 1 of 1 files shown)


pdillinger:
For reference: https://travis-ci.org/github/facebook/rocksdb/jobs/716643910

adamretter:
@pdillinger I am not sure this will help. If `sudo snap install cmake --beta --classic` exited with a non-zero code I think the build would have stopped, if it exited with a zero code, then I don't think adding `exit $?` will help you, as you will just get the same code... or did I miss something subtle?

pdillinger:
Updated Travis link (I'm getting a bad one from GitHub): https://travis-ci.org/github/facebook/rocksdb/builds/716990002

It looks like two builds are failing the snap install.

The subtle thing is this:

```
$ bash -c 'cat doesnotexist; echo next_step'; echo $?
cat: doesnotexist: No such file or directory
next_step
0
$ bash -c 'cat doesnotexist || exit $?; echo next_step'; echo $?
cat: doesnotexist: No such file or directory
1
$
```

Unless using `set -e` or an operator like && or ||, the return code of a command will be ignored if followed by another command.

adamretter:
> Unless using set -e 

Of course! Apologies, I am so used to always running scripts that include `set -e`

pdillinger:
@adamretter Can you look at fixing the two builds that are completely failing the snap install? Feel free to create your own PR that includes this change.

adamretter:
@pdillinger Yes, but I probably won't get to it until the weekend

pdillinger:
@craigscott-crascit Is there a missing "snap refresh core" step we weren't told about or something? Our linux builds here are failing with

    error: cannot perform the following tasks:
    - Mount snap "cmake" (513) (snap "cmake" assumes unsupported features: command-chain (try to update snapd and refresh the core snap))

as in https://travis-ci.org/github/facebook/rocksdb/jobs/716990011

craigscott-crascit:
> @craigscott-crascit Is there a missing "snap refresh core" step we weren't told about or something? Our linux builds here are failing with
> 
> ```
> error: cannot perform the following tasks:
> - Mount snap "cmake" (513) (snap "cmake" assumes unsupported features: command-chain (try to update snapd and refresh the core snap))
> ```
> 
> as in https://travis-ci.org/github/facebook/rocksdb/jobs/716990011

That's the first I've heard of this problem. After a quick bit of searching, it does seem likely that the core snap is not recent enough and needs to be refreshed. [This issue](https://forum.snapcraft.io/t/better-guidance-for-missing-assumes-features/504/3) seems to describe the problem reasonably well. Based on that, I would expect a `sudo snap refresh core` might fix it, but you'd have to try that and see.

pdillinger:
> [This issue](https://forum.snapcraft.io/t/better-guidance-for-missing-assumes-features/504/3) seems to describe the problem reasonably well. Based on that, I would expect a `sudo snap refresh core` might fix it, but you'd have to try that and see.

I think we have a winner with `sudo apt-get install snapd` to update snapd before running snap. Perhaps this subsumes `snap refresh core`. (?) Thanks.

adamretter:
@pdillinger Another option might be to place `snaps` under addons. I think (but am not certain) that there might then be an update option. The travis syntax for snaps does not have much documentation, but perhaps the following might work:

```
addons:
  snaps:
    update: true
    packages:
      - cmake
         confinement:
           - classic
           - beta
```

pdillinger:
Anything we add to Travis configuration now is better if it's more easily migrated to other CI systems.

pdillinger:
> @pdillinger Okay understood. So the preference is to script things, rather than use the YAML config options?

At least not spend extra time trying to use Travis-specific configs. (Approval?)

adamretter (approved):

PR #7240, opened 2020-08-11
AES-CTR Encryption for RocksDB

mulugetam:
This PR introduces a high performant and lightweight AES-CTR encryption for RocksDB. It extends from the existing encryption provider interface and therefore requires no code changes in the interface. The internal AES-CTR implementation is based on the highly-optimized open source cryptographic library, [IPP-Crypto](https://github.com/intel/ipp-crypto), that is portable on x86-64. It also introduces new options in db_bench (encrypt_db, block_cipher, block_cipher_key_size_in_bytes) to help measure its performance impact. 

Measurements on fillseq, readrandom, and readrandomwriterandom workloads indicate a performance impact <2% compared to the non-encrypted case and within run-to-run variance.

Example use: examples/ipp_aes_ctr_example.cc.


env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,131 @@
...
+#ifndef ROCKSDB_LITE
+
+IppAESCTRCipherStream::IppAESCTRCipherStream(IppsAESSpec* aes_ctx,
+                                             const char* init_vector)
+    : aes_ctx_(aes_ctx) {
+#if defined(HAVE_SSE42) || defined(HAVE_SSE2)
+  init_vector_ = _mm_loadu_si128((__m128i*)init_vector);
+#else
+  init_vector_ = init_vector;
+#endif
```
mrambacher, line 28: From the code, it is not clear to me what the lifetime of "init_vector" is and who "owns/deletes" it.  Can you clarify that via some comments somewhere please?

Also, should this be "const IppsAESSpec* aes_ctx"?

include/rocksdb/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,123 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
```
mrambacher, line 1: Is there a reason this header needs to be in the public API any longer?

include/rocksdb/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,123 @@
...
+#ifdef IPPCP
+
+#if !defined(ROCKSDB_LITE)
+
+#include <wmmintrin.h>
+
+#include <string>
+
+#include "env_encryption.h"
+#include "ippcp.h"
```
mrambacher, line 18: Is this a system library?  Should it be <ippcp.h>

include/rocksdb/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,123 @@
...
+
+#ifdef IPPCP
+
+#if !defined(ROCKSDB_LITE)
+
+#include <wmmintrin.h>
+
+#include <string>
+
+#include "env_encryption.h"
```
mrambacher, line 17: rocksdb/env_encryption.h

tools/db_bench_tool.cc
```diff
@@ -7119,6 +7131,41 @@ int db_bench_tool(int argc, char** argv) {
     FLAGS_env = new ROCKSDB_NAMESPACE::HdfsEnv(FLAGS_hdfs);
   }
 
+#ifdef IPPCP
```
mrambacher, line 7134: Can this be done via EncryptionProvider::CreateFromString rather than here?  It would make it easier to use the provider in different scenarios

include/rocksdb/ipp_aes_ctr_provider.h
mrambacher, line 1: Does this still need to be in the public API?

include/rocksdb/ipp_aes_ctr_provider.h
mrambacher, line 18: Is this a system header file?

include/rocksdb/ipp_aes_ctr_provider.h
mrambacher, line 17: include rocksb/env_encryption.h

include/rocksdb/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,123 @@
...
+  virtual rocksdb::Status CreateNewPrefix(const std::string& fname,
+                                          char* prefix,
+                                          size_t prefixLength) override;
+
+  virtual rocksdb::Status CreateCipherStream(
+      const std::string& fname, const rocksdb::EnvOptions& options,
+      rocksdb::Slice& prefix,
+      std::unique_ptr<rocksdb::BlockAccessCipherStream>* result) override;
+
+  virtual ~IppAESCTRProvider() {
```
mrambacher, line 109: Can we move the implementation to the C++ file instead?  Then we might not need to include some of the other header files here.

examples/ipp_aes_ctr_example.cc
```diff
@@ -0,0 +1,74 @@
...
+
+  DB* db;
+  Options options;
+  options.create_if_missing = true;
+
+  uint8_t key[] = {0xa6, 0xd2, 0xae, 0x28, 0x16, 0x15, 0x7e, 0x2b,
+                   0x3c, 0x4f, 0xcf, 0x09, 0x88, 0x15, 0xf7, 0xab};
+  std::unique_ptr<EncryptionProvider> provider(
+      new IppAESCTRProvider((char*)key, KeySize::AES_128));
+  options.env = NewEncryptedEnv(Env::Default(), provider.get());
```
mrambacher, line 34: Provider::CreateFromString

env/ipp_aes_ctr_provider_test.cc
```diff
@@ -0,0 +1,172 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
```
mrambacher, line 1: Can these tests be built into the env_encryption_test as a different parameterized provider now?

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,131 @@
...
+  }
+  IppAESCTRCipherStream cs(aes_ctx_, prefix);
+  return cs.Encrypt(0, prefix + IppAESCTRCipherStream::kBlockSize,
+                    prefixLength - IppAESCTRCipherStream::kBlockSize);
+}
+
+Status IppAESCTRProvider::CreateCipherStream(
+    const std::string& /*fname*/, const EnvOptions& /*options*/, Slice& prefix,
+    std::unique_ptr<BlockAccessCipherStream>* result) {
+  assert(prefix.size() >= IppAESCTRCipherStream::kBlockSize);
```
mrambacher, line 118: assert(result != nullptr)

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,131 @@
...
+  *result = std::unique_ptr<BlockAccessCipherStream>(
+      new IppAESCTRCipherStream(aes_ctx_, prefix.data()));
```
mrambacher, line 120: result->reset(new ...)

mrambacher (commented):
Please create a factory method for creating the EncryptionProvider that will validate the key and compilation switches (LITE, ICPP, etc) are set appropriately.

I think some of the code/logic needs to be updated to account for recent changes to the EncryptedEnv and related classes.  These changes should make it easier to plug in your code.
I cannot comment on whether or not the encryption logic is accurate/appropriate.

jay-zhuang:
Currently several tests are failing for ENCRYPTED_ENV:
```
$ ENCRYPTED_ENV=1 make -j32 check
```
it would be great if we could fix them first, also add `IppAESCTRProvider` as provider:
https://github.com/facebook/rocksdb/blob/a4a4a2dabdb9fc8dbd8567abaa50042de84a44f4/db/db_test_util.cc#L79-L82

mulugetam:
@mrambacher, @jay-zhuang: thanks! Will update the PR based on your comments and the new changes made to the API.

mrambacher:
I have cleaned up/fixed many of the tests that fail for MEM_ENV and ENCRYPTED_ENV.  I will be filing a PR next week with this change.  After it is accepted, I suggest you re-run your code to be sure it passes with ENCRYPTED_ENV set to use your provider.

mrambacher:
> I have cleaned up/fixed many of the tests that fail for MEM_ENV and ENCRYPTED_ENV. I will be filing a PR next week with this change. After it is accepted, I suggest you re-run your code to be sure it passes with ENCRYPTED_ENV set to use your provider.

The ENCRYPTED_ENV tests all pass now.  You should be able to set up your provider to run against them.

mulugetam:
The ENCRYPTED_ENV tests all pass for the IPP_AES provider.

tools/db_bench_tool.cc
```diff
@@ -7340,6 +7351,42 @@ int db_bench_tool(int argc, char** argv) {
...
   }
 
+#ifndef ROCKSDB_LITE
+  // Check for the encryption flag (and create an encrypted environment)
+  Random rnd(301);
+  std::shared_ptr<EncryptionProvider> provider;
+  if (FLAGS_encrypt_db && !strcasecmp(FLAGS_block_cipher.c_str(), "IPP_AES")) {
+    std::string key(FLAGS_block_cipher_key_size_in_bytes, ' ');
+    for (int i = 0; i < FLAGS_block_cipher_key_size_in_bytes; ++i)
+      key[i] = rnd.Uniform(256) & 0xFF;
```
mrambacher, line 7361: By using a random key like this, can I still run db_bench multiple times over the same data set successfully?  If so, is there a reason not to make the random key the "test" key and just use TEST_Initialize?

tools/db_bench_tool.cc
```diff
@@ -597,6 +598,16 @@ DEFINE_bool(show_table_properties, false,
...
 DEFINE_string(db, "", "Use the db with the following name.");
 
+#ifndef ROCKSDB_LITE
+// Read encryption flags
+DEFINE_bool(encrypt_db, false, "If true, the database will be encrypted.");
+
+DEFINE_string(block_cipher, "IPP_AES", "The name of the block cipher.");
+
+DEFINE_int32(block_cipher_key_size_in_bytes, 32,
+             "The key size (in bytes) of the block cipher.");
```
mrambacher, line 608: Should there be a flag that is the key?

examples/ipp_aes_ctr_example.cc
```diff
@@ -0,0 +1,64 @@
...
+int main() {
+  setbuf(stdout, NULL);
+
+  DB* db;
+  Options options;
+  options.create_if_missing = true;
+
+  std::shared_ptr<EncryptionProvider> provider;
+  std::string provider_str = "IPP_AES:256:a6d2ae2816157e2b3c4fcf098815f7xb";
+  Status status = EncryptionProvider::CreateFromString(ConfigOptions(),
```
mrambacher, line 29: Add Cipher?

env/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+ private:
+  IppsAESSpec* aes_ctx_;
+  __m128i init_vector_;
+};
+
+#endif  // IPPCP
+
+class IppAESCTRProvider : public CTREncryptionProvider {
```
mrambacher, line 58: Why are you extending CTREncryptionProvider rather than just EncryptionProvider?

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,174 @@
...
+Status IppAESCTRCipherStream::Decrypt(uint64_t fileOffset, char* data,
+                                      size_t dataSize) {
+  return Encrypt(fileOffset, data, dataSize);
+}
+
+#endif  // IPPCP
+
+Status IppAESCTRProvider::CreateProvider(const std::string& id,
+                                         IppAESCTRProvider** provider) {
+#if defined(IPPCP) && (defined(HAVE_SSE42) || defined(HAVE_SSE2))
```
mrambacher, line 75: This is probably better as a std::shared_ptr<> on std::unique_ptr<>

env/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,89 @@
...
+ public:
+  static Status CreateProvider(const std::string& id,
+                               IppAESCTRProvider** provider);
+
+#if defined(IPPCP) && (defined(HAVE_SSE42) || defined(HAVE_SSE2))
+
+  virtual size_t GetPrefixLength() const override { return 4096; }
+
```
mrambacher, line 66: Can you make this a constant please

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,174 @@
...
+  // expected format: IPP_AES:<128|192|256>:key
+  if (!std::regex_match(id, std::regex("IPP_AES:(128|192|256):.+"))) {
+    return Status::NotSupported("Invalid provider ", id);
+  }
```
mrambacher, line 79: Maybe use StringSplit here instead?

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,174 @@
...
+
+  // key size starts at index 8, key starts at index 12
+  size_t key_size_bits = std::stoi(id.substr(8, 3), nullptr);
```
mrambacher, line 82: This seems dangerous to me.  The StringSplit should help with this

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,174 @@
...
+  std::string key = id.substr(12);
+  if (key_size_bits != key.size() * 8) {
+    return Status::NotSupported("Key size mismatch in ", id);
+  }
```
mrambacher, line 86: Why does the key have to be specified here rather than via AddCipher

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,174 @@
...
+    case 128:
+      ptr_ipp->key_size_ = KeySize::AES_128;
+      break;
+    case 192:
+      ptr_ipp->key_size_ = KeySize::AES_192;
+      break;
+    case 256:
+      ptr_ipp->key_size_ = KeySize::AES_256;
+      break;
+  }
```
mrambacher, line 100: Add a default case and return an error?

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,174 @@
...
+
+  // key size starts at index 8, key starts at index 12
+  size_t key_size_bits = std::stoi(id.substr(8, 3), nullptr);
+  std::string key = id.substr(12);
+  if (key_size_bits != key.size() * 8) {
+    return Status::NotSupported("Key size mismatch in ", id);
+  }
+
+  IppAESCTRProvider* ptr_ipp = new IppAESCTRProvider();
+  ptr_ipp->ctx_size_ = 0;
```
mrambacher, line 89: Why is the ctx_size_ not initialized in the constructor?

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,174 @@
...
+      ptr_ipp->key_size_ = KeySize::AES_192;
+      break;
+    case 256:
+      ptr_ipp->key_size_ = KeySize::AES_256;
+      break;
+  }
+
+  // get size for context
+  IppStatus ipp_status = ippsAESGetSize(&ptr_ipp->ctx_size_);
+  assert(ipp_status == ippStsNoErr);
```
mrambacher, line 104: If this statement were to fail and asserts were off, wouldn't the status be lost?

env/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,89 @@
...
+  virtual Status CreateCipherStream(
+      const std::string& fname, const EnvOptions& options, Slice& prefix,
+      std::unique_ptr<BlockAccessCipherStream>* result) override;
+
+  virtual ~IppAESCTRProvider();
+
+ private:
+  IppAESCTRProvider() {}
+  IppAESCTRProvider(const IppAESCTRProvider&) = delete;
```
mrambacher, line 78: At least initialize the member variables to nullptr/0 here...

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,174 @@
...
+
+  Ipp8u ctr_block[kBlockSize];
+
+  __m128i counter = _mm_add_epi64(init_vector_, _mm_cvtsi64_si128(index));
+  Ipp8u* ptr_counter = (Ipp8u*)&counter;
+  for (size_t i = 0; i < kBlockSize; ++i)
+    ctr_block[i] = ptr_counter[kBlockSize - 1 - i];
+
+  IppStatus ipp_status = ippStsNoErr;
+  if (!offset) {
```
mrambacher, line 42: offset == 0

env/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,89 @@
...
+  virtual size_t BlockSize() override { return kBlockSize; }
+
+ protected:
+  virtual void AllocateScratch(std::string&) override {}
+  virtual Status EncryptBlock(uint64_t, char*, char*) override {
+    return Status::OK();
+  }
+  virtual Status DecryptBlock(uint64_t, char*, char*) override {
+    return Status::OK();
+  }
```
mrambacher, line 49: Are these ever called?  Should they assert/return something else?

env/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,89 @@
...
+};
+
+#endif  // IPPCP
+
+class IppAESCTRProvider : public CTREncryptionProvider {
+ public:
+  static Status CreateProvider(const std::string& id,
+                               IppAESCTRProvider** provider);
+
+#if defined(IPPCP) && (defined(HAVE_SSE42) || defined(HAVE_SSE2))
```
mrambacher, line 63: Should this #ifdef be around everything or just the variables?  Seems like weird compilation rules otherwise...

tools/db_bench_tool.cc
```diff
@@ -7340,6 +7351,42 @@ int db_bench_tool(int argc, char** argv) {
     FLAGS_env = new ROCKSDB_NAMESPACE::HdfsEnv(FLAGS_hdfs);
   }
 
+#ifndef ROCKSDB_LITE
+  // Check for the encryption flag (and create an encrypted environment)
+  Random rnd(301);
+  std::shared_ptr<EncryptionProvider> provider;
+  if (FLAGS_encrypt_db && !strcasecmp(FLAGS_block_cipher.c_str(), "IPP_AES")) {
```
mrambacher, line 7358: If you need the IPP_AES here, can you make it a constant somewhere that is accessible by multiple places?  Perhaps EncryptionProvider::kIPPAESName()?

tools/db_bench_tool.cc
```diff
@@ -7340,6 +7351,42 @@ int db_bench_tool(int argc, char** argv) {
...
+        break;
+      default:
+        fprintf(stderr,
+                "key size for the AES block cipher must be 16, 24, or "
+                "32 bytes.\n");
+        exit(1);
+    }
+    provider_str += ":" + key;
+    EncryptionProvider::CreateFromString(ConfigOptions(), provider_str,
+                                         &provider);
```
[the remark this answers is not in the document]
mulugetam: Will add the key as a flag and remove this.

tools/db_bench_tool.cc
mulugetam: Same. Will add the key as a flag and remove this.

tools/db_bench_tool.cc
mulugetam: Yes.

tools/db_bench_tool.cc
mulugetam: Sure.

env/ipp_aes_ctr_provider.h
mulugetam: I wanted to avoid putting the #ifdef in EncryptionProvider::CreateFromString in env/env_encryption.cc. Will do that.

env/ipp_aes_ctr_provider.cc
mulugetam: Sure.

env/ipp_aes_ctr_provider.h
mrambacher: I don't want the #ifdef in EncryptionProvider either.  But you still can define this class, just without the values that reference IPPCP headers.  Make them conditional but leave the class as valid perhaps?

env/ipp_aes_ctr_provider.cc
mulugetam: Sure.

env/ipp_aes_ctr_provider.h
mulugetam: Sure.

env/ipp_aes_ctr_provider.cc
mulugetam: Sure. Didn't know we have that function.

env/ipp_aes_ctr_provider.cc
mulugetam: Sure.

env/ipp_aes_ctr_provider.cc
mulugetam: The check is made earlier -- one of the three cases will be taken.

env/ipp_aes_ctr_provider.cc
mulugetam: Should be.

env/ipp_aes_ctr_provider.h
mulugetam: Will do.

env/ipp_aes_ctr_provider.cc
mulugetam: Correct. Will add a check.

env/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,89 @@
...
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#pragma once
+
+#if !defined(ROCKSDB_LITE)
+
+#ifdef IPPCP
+
```
mrambacher, line 12: Can you add a comment in here about the IPPCP provider in general?  What it does, where to get it, etc.  A header comment for the file and/or class would be helpful.

mulugetam:
@mrambacher I have tried to address the comments, could you please review?

db/db_test_util.cc
```diff
@@ -77,6 +77,10 @@ DBTestBase::DBTestBase(const std::string path, bool env_do_fsync)
     Status s = EncryptionProvider::CreateFromString(
         ConfigOptions(), std::string("test://") + getenv("ENCRYPTED_ENV"),
         &provider);
+    if (!strcmp(getenv("ENCRYPTED_ENV"),
```
mrambacher, line 80: Can't this be done in your provider via TEST_Initialize()?

env/env_basic_test.cc
```diff
@@ -117,9 +117,19 @@ INSTANTIATE_TEST_CASE_P(EncryptedEnv, EnvMoreTestWithParam,
...
 
 #if defined(IPPCP) && (defined(HAVE_SSE42) || defined(HAVE_SSE2))
+static Env* NewTestIPPEncryptedEnv(Env* base, const std::string& provider_id) {
+  std::shared_ptr<EncryptionProvider> provider;
+  EXPECT_OK(EncryptionProvider::CreateFromString(ConfigOptions(), provider_id,
+                                                 &provider));
+  EXPECT_OK(
+      provider->AddCipher("", "a6d2ae2816157e2b3c4fcf098815f7xb", 32, false));
+  std::unique_ptr<Env> encrypted(NewEncryptedEnv(base, provider));
+  return new NormalizingEnvWrapper(std::move(encrypted));
```
mrambacher, line 127: Again, you should be able to do this via TEST_Initialize

env/env_encryption.cc
```diff
@@ -61,10 +59,8 @@ Status EncryptionProvider::CreateFromString(
   }
   if (id == kCTRProviderName) {
     result->reset(new CTREncryptionProvider());
-  } else if (StartsWith(id, kIPPAESProviderName)) {
-    IppAESCTRProvider* ipp_provider = nullptr;
-    status = IppAESCTRProvider::CreateProvider(id, &ipp_provider);
-    result->reset(ipp_provider);
+  } else if (id == kIPPAESProviderName) {
```
mrambacher, line 62: The way I have done this for other classes is to do the following in your provider class:
static const char *kName() { return "IPP_AES"; }
const char *Name() const override ( return kName(); }

Then this method can use StartsWith(id, IppAESCTRProvider::kName())

env/ipp_aes_ctr_provider.cc
```diff
@@ -8,136 +8,86 @@
...
+  return Status::OK();
 }
 
-#endif  // IPPCP
-
-Status IppAESCTRProvider::CreateProvider(const std::string& id,
-                                         IppAESCTRProvider** provider) {
+Status IppAESCTRProvider::AddCipher(const std::string& /*descriptor*/,
+                                    const char* cipher, size_t len,
+                                    bool /*for_write*/) {
```
mrambacher, line 33: You either need:
- Support more than one encryption key
- Prevent the user from calling this method if there is already an encryption key (return an error)
- If this method is called and there already is a key, delete the old key before you insert the new one

tools/db_bench_tool.cc
```diff
@@ -7353,33 +7356,21 @@ int db_bench_tool(int argc, char** argv) {
...
-                "32 bytes.\n");
-        exit(1);
+  if (FLAGS_encrypt_db &&
+      !strcmp(FLAGS_block_cipher.c_str(),
+              EncryptionProviderNames::kIPPAESProviderName)) {
+    if (FLAGS_block_cipher_key.size() != FLAGS_block_cipher_key_size_in_bytes) {
+      fprintf(stderr,
+              "The size of key doesn't match with "
+              "block_cipher_key_size_in_bytes.\n");
+      exit(1);
```
mrambacher, line 7367: Is this already done inside of the EncryptionProvider?  I wonder if you need it here or not?

tools/db_bench_tool.cc
```diff
@@ -7353,33 +7356,21 @@ int db_bench_tool(int argc, char** argv) {
...
     }
-    provider_str += ":" + key;
-    EncryptionProvider::CreateFromString(ConfigOptions(), provider_str,
-                                         &provider);
+    EncryptionProvider::CreateFromString(
```
mrambacher, line 7369: Check the status.  If not ok(), fprint and return exit

tools/db_bench_tool.cc
```diff
@@ -7353,33 +7356,21 @@ int db_bench_tool(int argc, char** argv) {
...
+        ConfigOptions(), EncryptionProviderNames::kIPPAESProviderName,
```
mrambacher, line 7370: Why not pass in "FLAGS_block_cipher" instead?  Then won't it work for any provider?

tools/db_bench_tool.cc
```diff
@@ -7353,33 +7356,21 @@ int db_bench_tool(int argc, char** argv) {
...
-        provider_str += ":256";
-        break;
-      default:
-        fprintf(stderr,
-                "key size for the AES block cipher must be 16, 24, or "
-                "32 bytes.\n");
-        exit(1);
+  if (FLAGS_encrypt_db &&
+      !strcmp(FLAGS_block_cipher.c_str(),
+              EncryptionProviderNames::kIPPAESProviderName)) {
```
mrambacher, line 7362: Is there a reason to check this here?  Won't EncryptionProvider::CreateFromString do the right thing?

tools/db_bench_tool.cc
```diff
@@ -7353,33 +7356,21 @@ int db_bench_tool(int argc, char** argv) {
...
     }
-    provider_str += ":" + key;
-    EncryptionProvider::CreateFromString(ConfigOptions(), provider_str,
-                                         &provider);
+    EncryptionProvider::CreateFromString(
+        ConfigOptions(), EncryptionProviderNames::kIPPAESProviderName,
+        &provider);
+    provider->AddCipher("", FLAGS_block_cipher_key.c_str(),
+                        FLAGS_block_cipher_key_size_in_bytes, false);
     FLAGS_env = NewEncryptedEnv(Env::Default(), provider);
```
mrambacher, line 7374: Is FLAGS_env already valid?  Could it be used here instead of Env::Default()?

mrambacher (commented):
I think this is much closer.  Please look if overriding EncryptionProvider::TEST_Initialize would allow you to set the cipher without having code for your provider in more places.

mulugetam:
@mrambacher I have tried to address your comments. Thanks again!

tools/db_bench_tool.cc
mulugetam: FLAGS_env cannot be used here instead of Env::Default().

db/db_test_util.cc
```diff
@@ -77,7 +77,10 @@ DBTestBase::DBTestBase(const std::string path, bool env_do_fsync)
     Status s = EncryptionProvider::CreateFromString(
         ConfigOptions(), std::string("test://") + getenv("ENCRYPTED_ENV"),
         &provider);
-    encrypted_env_ = NewEncryptedEnv(mem_env_ ? mem_env_ : base_env, provider);
+    if (s.ok()) {
```
mrambacher, line 80: This should probably be an EXPECT_OK(s)

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,204 @@
...
+    ctr_block[i] = ptr_counter[kBlockSize - 1 - i];
+
+  IppStatus ipp_status = ippStsNoErr;
+  if (offset == 0) {
+    ipp_status = ippsAESEncryptCTR((Ipp8u*)(data), (Ipp8u*)data, dataSize,
+                                   aes_ctx_, ctr_block, kCounterLen);
+  } else {
+    Ipp8u zero_block[kBlockSize]{0};
+    ipp_status = ippsAESEncryptCTR(zero_block, zero_block, kBlockSize, aes_ctx_,
+                                   ctr_block, kCounterLen);
```
mrambacher, line 177: This status is never checked and overwritten a few lines later.

env/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,109 @@
...
+// IPPCP is lightweight cryptography library that is highly-optimized for
+// various Intel CPUs.
+//
+// We use it here to provide an AES-128/192/256 encryption with a CTR mode of
+// operation.
+//
+// Download URL: https://github.com/intel/ipp-crypto.
+//
+#include <emmintrin.h>
+#include <ippcp.h>
```
mrambacher, line 23: By using a forward declaration of IppsAESSpec and moving IppAESCTRCipherStream into the C++ file, you could do away with the ifdefs and system includes here.  Might make things a little cleaner...

env/ipp_aes_ctr_provider.h
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+ private:
+  IppsAESSpec* aes_ctx_;
+  __m128i init_vector_;
+};
+
+#endif  // IPPCP
+
+class IppAESCTRProvider : public EncryptionProvider {
```
mrambacher, line 67: Please add a block comment as to what this provider does

examples/ipp_aes_ctr_example.cc
```diff
@@ -0,0 +1,70 @@
...
+
+using namespace ROCKSDB_NAMESPACE;
+
+std::string kDBPath = "/tmp/ipp_aes_example";
+
+int main() {
+  setbuf(stdout, NULL);
+
+#if defined(IPPCP) && (defined(HAVE_SSE42) || defined(HAVE_SSE2))
+  DB* db;
```
mrambacher, line 24: You probably do not need this ifdef and can use the return from EncryptoinProvider::CreateFromString?

mrambacher (commented):
Looks good!  Few mainly minor comments

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,204 @@
...
+    return Status::NotSupported("Multiple encryption keys not supported.");
+  }
+
+  // AES supports key sizes of only 16, 24, or 32 bytes
+  if (len != 16 && len != 24 && len != 32) {
+    return Status::NotSupported("Invalid key size in provider.");
+  }
+
+  size_t key_size = std::string(cipher).size();
+  if (key_size != len) {
```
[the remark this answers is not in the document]
mulugetam: May be I should remove the check and let it fail later. A char* and a len are passed in as arguments. How do we validate that the char*, the cipher, has a size of len?

examples/ipp_aes_ctr_example.cc
mulugetam: The ifdef is added to avoid a fail on the unit tests on systems with no ipp-crypto - e.g. the CircleCI checks would fail on the assert of the status returned from EncryptionProvider::CreateFromString (line 31).

env/env_encryption.cc
```diff
@@ -62,6 +62,8 @@ Status EncryptionProvider::CreateFromString(
     result->reset(new CTREncryptionProvider());
   } else if (is_test) {
     result->reset(new CTREncryptionProvider());
+  } else if (id == IppAESCTRProvider::kName()) {
+    status = IppAESCTRProvider::CreateProvider(id, result);
```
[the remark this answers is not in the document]
mulugetam: Fixed.

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,204 @@
...
+    return Status::NotSupported("Invalid provider ", id);
+  }
+#if defined(IPPCP) && (defined(HAVE_SSE42) || defined(HAVE_SSE2))
+  IppAESCTRProvider* ptr_ipp = new IppAESCTRProvider();
+  provider->reset(ptr_ipp);
+  return Status::OK();
+#else
+  (void)id;
+  (void)provider;
+  return Status::Aborted("ipp-crypto library not found and requires SSE2+.");
```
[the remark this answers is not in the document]
mulugetam: Fixed.

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,204 @@
...
+#if defined(IPPCP) && (defined(HAVE_SSE42) || defined(HAVE_SSE2))
+  // We currently don't support more than one encryption key
+  if (aes_ctx_ != nullptr) {
+    return Status::NotSupported("Multiple encryption keys not supported.");
+  }
+
+  // AES supports key sizes of only 16, 24, or 32 bytes
+  if (len != 16 && len != 24 && len != 32) {
+    return Status::NotSupported("Invalid key size in provider.");
+  }
```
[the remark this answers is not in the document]
mulugetam: Fixed.

env/ipp_aes_ctr_provider.cc
```diff
@@ -0,0 +1,204 @@
...
+
+  size_t key_size = std::string(cipher).size();
+  if (key_size != len) {
+    return Status::NotSupported("Key size mismatch in provider.");
```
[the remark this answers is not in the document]
mulugetam: I have removed this check entirely because I found no good way to do it without forbidding an embedded null character in the cipher which I don't want to do.

db/db_test_util.cc
mulugetam: Fixed.

env/ipp_aes_ctr_provider.cc
mulugetam: Fixed.

env/ipp_aes_ctr_provider.h
mulugetam: Fixed.

env/ipp_aes_ctr_provider.h
mulugetam: Fixed.

mulugetam:
The latest version of this patch is available here: https://github.com/facebook/rocksdb/pull/8943.

PR #7243, opened 2020-08-12
Fix Java test for uint64add merge operator

shirolimit:
The PR fixes a Java test for Merge operator `uint64add`.

The current implementation uses wrong byte order for long serialization, but fails to catch this error because the merge sum is lower than `256`.

The PR makes this test case more representative (i.e. it fails with wrong byte order) and changes the byte order to little endian.

Some background: RocksDB uses LittleEndian byte order for integer serialization across all platforms. `MergeTest` uses `ByteBuffer` that defaults to BigEndian byte order.

This test case might probably be used as a sample of `MergeOperator` usage in Java.

### DIFF
--- a/java/src/test/java/org/rocksdb/MergeTest.java
+++ b/java/src/test/java/org/rocksdb/MergeTest.java
@@ -5,17 +5,17 @@
 
 package org.rocksdb;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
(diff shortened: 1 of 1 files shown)


shirolimit:
Hmm, it seems that one of the Travis CI jobs has timed out. I don't think that it is relevant to the PR.
What should I do about it?

adamretter:
@shirolimit It happens from time to time. I have restarted it...

shirolimit:
@adamretter got it, thank you. Looks like all checks have passed

Issue #4452, opened 2018-10-04
Missing pkg-config file


ael-code:
If I compile through the provided makefile, all the optional compression libraries found on my system are kicked in as dependencies. When I later try to link against the generated rocksDB shared library I need to link with all those compression libraries, otherwise I'll get:
```Compression type <CompressionLibrary> is not linked with the binary```

While compiling my software I don't have an easy way to know which compression libraries have been enabled and required by the rocksDB shared  library.
If a pkg-config file would have been provided for the shared library, it would be easy to understand which compression library I need to link with.

yiwu-arbug:
@ael-code We are not familiar with pkg-config. Are you interested in send us a pull request for it? Thanks.

ajkr:
The values of the options @anand1976 mentioned will tell you which compression libraries need to be linked.

ael-code:
> The values of the options @anand1976 mentioned will tell you which compression libraries need to be linked.

That's true. But if rocksDB have been installed on my system by another person or through the package manager I don't have any clue about which configuration option have been used.

ajkr:
Those are runtime options that the user provides to RocksDB every time a DB is opened. They are not set during RocksDB shared library installation.

ajkr:
I agree we should have a pkg-config file, though. I just don't see that it needs to require compression libraries.

PR #7244, opened 2020-08-12
Generate and install a pkg-config file

jgoerzen:
pkg-config files are quite useful for communicating to users of a
library how to compile against them. This commit generates and installs
a pkg-config file that can be used for both static and dynamic builds
against the RocksDB library. This should make life easier for developers
of client programs, language bindings, etc.

Example usage:

```
g++ `pkg-config --cflags rocksdb` -o simple_example simple_example.cc `pkg-config --libs rocksdb`

g++ `pkg-config --cflags --static rocksdb` -static \
   -o simple_example simple_example.cc `pkg-config --libs --static rocksdb`
```

The commit also adds the generated file to .gitignore, to the uninstall
target, and to clean.

No additional dependencies are added to RocksDB itself, and this does
not make RocksDB use pkg-config as part of its build process.

Resolves https://github.com/facebook/rocksdb/issues/4452

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 make_config.mk
+rocksdb.pc
 
 *.a
 *.arc
--- a/Makefile
+++ b/Makefile
@@ -265,7 +265,7 @@ dummy := $(shell (export ROCKSDB_ROOT="$(CURDIR)"; \
 include make_config.mk
 
 export JAVAC_ARGS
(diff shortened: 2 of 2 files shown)


jgoerzen:
Hi @ajkr, This is working through legal process right now (I contributed this as part of my work at Fastly).  It's in the works and I would expect it to be signed early next week.  Thanks!

PR #7252, opened 2020-08-13
[wip] Revert "Disable recycle_log_file_num when it is incompatible with recovery mode (#6351)"

ajkr:
We don't know what the reverted PR solves but suspect it had something to do with 2PC. However, that PR disabled WAL recycling in general scenarios not specific to 2PC, which caused confusion. If we figure out the 2PC problem, it needs to be solved in a more targeted way.


db/db_impl/db_impl_open.cc
```diff
@@ -92,11 +92,12 @@ DBOptions SanitizeOptions(const std::string& dbname, const DBOptions& src) {
...
        result.wal_recovery_mode == WALRecoveryMode::kAbsoluteConsistency)) {
-    // kPointInTimeRecovery is inconsistent with recycle log file feature since
-    // we define the "end" of the log as the first corrupt record we encounter.
+    // kPointInTimeRecovery is indistinguishable from
+    // kTolerateCorruptedTailRecords in recycle mode since we define
+    // the "end" of the log as the first corrupt record we encounter.
     // kAbsoluteConsistency doesn't make sense because even a clean
     // shutdown leaves old junk at the end of the log file.
-    result.recycle_log_file_num = 0;
+    result.wal_recovery_mode = WALRecoveryMode::kTolerateCorruptedTailRecords;
```
ajkr, line 100: The existing code here makes no sense to me either, tbh, but I'd rather incrementally improve on what we had before.

ajkr:
Experimented with it a bit. This behavior doesn't look good:

1. Create two unflushed WALs for default CF. Both have one key per WAL in default CF.
```
In [1]: import rocksdb

In [2]: db = rocksdb.DB('./test-db', rocksdb.Options(create_if_missing=True, wal_recovery_mode='kPointInTimeRecovery', recycle_log_file_num=2))

In [3]: db.create_column_family('a')
Out[3]: 'a'

In [4]: db.put(b'ok', b'ok', 'a')

In [5]: db.put(b'key1', b'wal1')

In [6]: db.flush('a')

In [7]: db.put(b'key2', b'wal2')

In [8]: del db
```

2. Manually corrupt `"key1": "wal1"` (i.e., second entry in first WAL).

3. Reopen the DB. Observe recovery permitted a hole, which is not a consistent point in time.

```
In [3]: db = rocksdb.DB('./test-db', rocksdb.Options(create_if_missing=True, wal_recovery_mode='kPointInTimeRecovery', recycle_log_file_num=2), column_families=['default', 'a'])

In [4]: db.get(b'key1')

In [5]: db.get(b'key2')
Out[5]: b'wal2'
```

ajkr:
Well, the exact same problem still happens with `kTolerateCorruptedTailRecords`. Arguably that's worse as it's supposed to give a stronger recovery guarantee. Anyways, the plan is now to enable it in `kPointInTimeRecovery` and disable it in `kTolerateCorruptedTailRecords`. To do this, we first need to make a checksum mismatch error actually stop the WAL replay, even if newer WALs are present.

ajkr:
No longer WIP (at least not by me). For the short term fix for the problem above, https://github.com/facebook/rocksdb/pull/7271 disabled the feature more broadly.

PR #7254, opened 2020-08-13
Fix flaky BackupableDBTest.CustomChecksumTransition

gg814:
Summary:
The flaky test in the title is caused by two problems. First, there is a bug in the BackupEngine that results in skipping computing the default crc32 checksum when `share_table_files` is enabled and the table is already backed up. Second, when `RestoreDBFromBackup` fails and the backup was being restored to the DB directory, it is likely that `RestoreDBFromBackup` has cleaned up the DB directory before it fails, and therefore, files in old backups may collide with files to be backed up if `share_files_with_checksum` is not enabled.

New tests that cover the above problems are added.

Test plan:
`./backupable_db_test`

### DIFF
--- a/utilities/backupable/backupable_db_test.cc
+++ b/utilities/backupable/backupable_db_test.cc
@@ -1053,12 +1053,23 @@ TEST_F(BackupableDBTest, CustomChecksumTransition) {
                           sopt);
     ASSERT_OK(backup_engine_->CreateNewBackup(db_.get()));
     CloseBackupEngine();
+
+    ++i;
     // Say, we accidentally change the factory
     backupable_options_->file_checksum_gen_factory = hash32_factory;
     OpenBackupEngine();
-    ASSERT_NOK(backup_engine_->VerifyBackup(i + 2, true));
-    ASSERT_NOK(backup_engine_->RestoreDBFromBackup(i + 2, dbname_, dbname_));
-    ASSERT_OK(backup_engine_->DeleteBackup(i + 2));
+    // Unable to verify the latest backup.
+    ASSERT_NOK(backup_engine_->VerifyBackup(i + 1, true));
+    // Unable to restore the latest backup.
+    ASSERT_NOK(backup_engine_->RestoreDBFromBackup(i + 1, dbname_, dbname_));
+    CloseBackupEngine();
+    // Reset the factory to the same as the one used in the db.
+    backupable_options_->file_checksum_gen_factory = hash_factory;
+    OpenBackupEngine();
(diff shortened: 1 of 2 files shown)


pdillinger:
> First, there is a bug in the BackupEngine that results in skipping computing the default crc32 checksum when `share_table_files` is enabled and the table is already backed up.

Does the bug exist before #7085?

utilities/backupable/backupable_db_test.cc
```diff
@@ -1106,6 +1117,92 @@ TEST_F(BackupableDBTest, CustomChecksumTransition) {
...
+    FillDB(db_.get(), 0, keys_iteration);
+    ASSERT_OK(backup_engine_->CreateNewBackup(db_.get()));
+    CloseBackupEngine();
+    // Say, we accidentally change the factory
+    backupable_options_->file_checksum_gen_factory = hash32_factory;
+    OpenBackupEngine();
+    // Note that when backup engine is unable to restore, it may have cleaned up
+    // the db directory. If the db directory has been cleaned up, it is
+    // sensitive to file collision.
+    ASSERT_NOK(backup_engine_->RestoreDBFromBackup(2, dbname_, dbname_));
```
pdillinger, line 1180: Do we really need a failed restore here to set up these conditions?

gg814:
> Does the bug exist before #7085?

The bug is only introduced in the custom checksum PR.

utilities/backupable/backupable_db_test.cc
gg814: It is not a necessary condition; just one way to reproduce the file collision problem for sharing table without checksum. (For example, we may deliberately clean up the DB to reproduce the problem.)

pdillinger: Let's do something much more standard like delete (destroy) the DB. (I assume that's what you mean by "clean.")

pdillinger:
Failure in BackupableDBTest.ShareTableFilesWithChecksumsNewNaming: https://travis-ci.org/github/facebook/rocksdb/jobs/717977855

pdillinger:
> Failure in BackupableDBTest.ShareTableFilesWithChecksumsNewNaming: https://travis-ci.org/github/facebook/rocksdb/jobs/717977855

I don't have definitive data, but I can believe this is pre-existing and can be fixed in another PR.

pdillinger:
Actually, here's a similar pre-existing failure in BackupableDBTest.ShareTableFilesWithChecksumsNewNamingTransition: https://travis-ci.org/github/facebook/rocksdb/jobs/717143578

That's pretty definitive that it's a separate issue

PR #7256, opened 2020-08-14
Track WAL in MANIFEST: persist WALs to and recover WALs from MANIFEST

cheng-chang:
This PR makes it able to `LogAndApply` `VersionEdit`s related to WALs, and also be able to `Recover` from MANIFEST with WAL related `VersionEdit`s.

The `VersionEdit`s related to WAL are treated similarly as those related to column family operations, they are not applied to versions, but can be in a commit group. Mixing WAL related `VersionEdit`s with other types of edits will make logic in `ProcessManifestWrite` more complicated, so `VersionEdit`s related to WAL can either be WAL additions or deletions, like column family add and drop.

Test Plan:

a set of unit tests are added in `version_set_test.cc`


db/version_set.cc
```diff
@@ -3898,6 +3927,14 @@ Status VersionSet::ProcessManifestWrites(
       assert(curr_state.find(cfd->GetID()) == curr_state.end());
       curr_state[cfd->GetID()] = {cfd->GetLogNumber()};
     }
+
+    for (const auto& wal : wals_.GetWals()) {
+      WalNumber number = wal.first;
+      const WalMetadata& meta = wal.second;
+      VersionEdit* edit = new VersionEdit();
```
riversand963, line 3934: Can we avoid using `new` while holding mutex?

db/version_set.cc
```diff
@@ -4230,6 +4287,20 @@ Status VersionSet::LogAndApply(
                                new_cf_options);
 }
 
+void VersionSet::LogAndApplyWalHelper(VersionEdit* edit) {
+  assert(edit->IsWalManipulation());
+
+  edit->SetPrevLogNumber(prev_log_number_);
+  edit->SetNextFile(next_file_number_.load());
+  // The log might have data that is not visible to memtbale and hence have not
+  // updated the last_sequence_ yet. It is also possible that the log has is
```
riversand963, line 4296: "the log has is", typo?

riversand963: Is this comment about `two_write_queues_`?

db/version_set.cc
```diff
@@ -4474,6 +4545,26 @@ Status VersionSet::ReadAndRecover(
...
       break;
     }
+    if (edit.IsWalManipulation()) {
+      if (edit.IsWalAddition()) {
+        for (const WalAddition& wal : edit.GetWalAdditions()) {
+          if (wal.GetMetadata().HasSize()) {
+            // Create WAL before closing.
+            wals_.AddWal(WalAddition(wal.GetLogNumber()));
+          }
+          wals_.AddWal(wal);
```
riversand963, line 4555: Do we check the return value of `AddWal`?

db/version_set.cc
```diff
@@ -5281,6 +5372,7 @@ void VersionSet::MarkMinLogNumberToKeep2PC(uint64_t number) {
 
 Status VersionSet::WriteCurrentStateToManifest(
     const std::unordered_map<uint32_t, MutableCFState>& curr_state,
+    const std::vector<std::unique_ptr<VersionEdit>>& wal_additions,
```
riversand963, line 5375: I think here we should record `VersionSet::wals_`.

db/version_set.cc
cheng-chang: I copied from `LogAndApplyCFHelper`, deleted this duplicated comment.
Yes, it explains why the last sequence is related to two write queues.

cheng-chang:
@riversand963 Thanks for the review! PTAL.

db/version_set.cc
```diff
@@ -3745,12 +3748,27 @@ Status VersionSet::ProcessManifestWrites(
     // No group commits for column family add or drop
     LogAndApplyCFHelper(first_writer.edit_list.front());
     batch_edits.push_back(first_writer.edit_list.front());
+  } else if (first_writer.edit_list.front()->IsWalManipulation()) {
+    // WAL addition/deletion won't appear together with other kinds of edits,
```
riversand963, line 3752: Why cannot one writer perform MANIFEST write for another writer's WAL manipulations?

db/version_set.h
```diff
@@ -1262,7 +1262,19 @@ class VersionSet {
   Status VerifyFileMetadata(const std::string& fpath,
                             const FileMetaData& meta) const;
 
-  WalSet wals_;
+  // Note on synchronization:
+  // when opening a DB,
+  // 1. the DB recovery thread will read and write wals_.
```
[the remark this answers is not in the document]
cheng-chang: On recovery, the main db thread will read the records from manifest and update `wals_`.

After recovery, whenever we `LogAndApply` a WAL related edit, we'll update `wals_`, when `WriteCurrentStateToManifest`, we'll read `wals_`.

db/version_set.cc
cheng-chang: There is no reason that WAL edits cannot be grouped.

But since WAL edits are not treated differently from other kinds of edits (such as not applied to versions, and needs to update `wals_`), handling them specially like column family edits makes the logic easier to understand.

riversand963: I feel that additional logic to (unnecessarily) separate the processing of wal-related edits make the `ProcessManifestWrites` more complicated than before. Can we do something like
```
for (const auto& edit : edits) {
  if (wal-related) {
    LogAndApplyWalManipulation();
  } else {
    LogAndApplyHelper();
  }
}
```

db/version_set.h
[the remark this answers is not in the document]
riversand963: I think either `Recover` or `LogAndApply` will for some period of time hold the db mutex. Can we access `wals_` while holding the db mutex?

db/version_set.cc
cheng-chang: the tricky part is to handle `version` related logic differently for WAL edits, there is no need to update versions for WAL edits.

cheng-chang: updated the logic, PTAL

db/version_set.cc
```diff
@@ -5329,6 +5402,28 @@ Status VersionSet::WriteCurrentStateToManifest(
     }
   }
 
+  {
+    // Save WALs.
+    if (!wals_.load()->GetWals().empty()) {
+      std::unique_ptr<VersionEdit> wal_additions(new VersionEdit());
+      for (const auto& wal : wals_.load()->GetWals()) {
```
riversand963, line 5409: Not sure if I follow the intent of using atomic variable here.
`wals_` is `std::atomic<WalSet*>`. It means the pointer is atomic, but not the object pointed to by the pointer.
This piece of code is outside db mutex, and the `WalSet` object needs to be protected. What we do with other VersionStorage information is to copy them.

riversand963: Will your future changes modify `wals_` dynamically, i.e. let the pointer point to different `WalSet` objects? If so, how do you reason about the correctness here?

db/version_set.cc
```diff
@@ -3825,13 +3838,17 @@ Status VersionSet::ProcessManifestWrites(
         } else if (group_start != std::numeric_limits<size_t>::max()) {
           group_start = std::numeric_limits<size_t>::max();
         }
-        Status s = LogAndApplyHelper(last_writer->cfd, builder, e, mu);
-        if (!s.ok()) {
-          // free up the allocated memory
-          for (auto v : versions) {
-            delete v;
+        if (e->IsWalManipulation()) {
+          LogAndApplyWalHelper(e);
```
riversand963, line 3842: If you let `LogAndApplyWalHelper` return `Status`, then the error handling logic can be shared, saving some indentation.

db/version_set.h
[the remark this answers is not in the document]
cheng-chang: we could do, but it's unnecessary, as explained in https://github.com/facebook/rocksdb/pull/7256/files#diff-6822534358e613a82db5431a19a6036fR5387.

I'll try to describe the synchronization behavior for `wals_` more clearly in the comment.

db/version_set.cc
cheng-chang: I thought about the synchronization again, I should leverage release-acquire ordering of the atomic operation (as described in `Release-Acquire ordering` section in https://en.cppreference.com/w/cpp/atomic/memory_order). 

No, only the `ProcessManifestWrite` will update `wals_` after recovery.

db/version_set.cc
cheng-chang: I realized that we can get rid of `LogAndApplyWalHelper`!

db/version_set.cc
cheng-chang: Actually, there is no need to use atomic operation to establish the happens-before relationship between the recover thread and the LogAndApply thread, I updated the comment of wals_ to explain the reason behind this.

cheng-chang:
@riversand963 thanks for comments, PTAL.

db/version_set.h
```diff
@@ -1262,6 +1262,20 @@ class VersionSet {
...
 
+  // Note on synchronization:
+  // When opening a DB,
+  // 1. the DB recovery thread will read and write wals_.
+  // After recovery,
+  // 2. only the LogAndApply thread will read and write wals_.
+  //
+  // Before entering DB recovery, DB mutex is acquired,
+  // before LogAndApply, DB mutex is also acquired,
+  // so writes to wals_ by the DB recovery thread happens before
```
riversand963, line 1273: Note that both recovery and `LogAndApply` can release db mutex in the middle.

I think the real reason is that
- no user thread will call `LogAndApply` concurrently with the thread calling `Recover()` because application cannot write to a `db` before the database is opened and `db` pointer is returned. The thread calling `Recover()` also calls `LogAndApply()`.
- threads calling `LogAndApply` after recovery succeeds and db is opened. These threads are serialized via the `manifest_writers_` queue.

I think it is very likely that some APIs will need read access to the `wals_` which can happen concurrently with `LogAndApply`. For example, `DB::GetSortedWalFiles()` may want to use `wals_` to verify the WAL files. If so, some sort of synchronization will be necessary.

cheng-chang: During recovery, since the read and write are in the same thread, there is no need to synchronize on wals_.

I updated the comment to be more explicit again.

cheng-chang: Yes, during recovery, `GetSortedWalFiles` may need to access `wals_` concurrently with `LogAndApply`.

For `GetSortedWalFiles`, since it's in the same recovery thread, the `wals_` written during manifest recovery is visible.

For `LogAndApply`, since DB mutex is always acquired before `ProcessManifestWrite`, even if it happens concurrently with the recovery thread, as long as it happens after `VersionSet::Recover`, the mutex establishes a happens before relationship between the recovery thread and the write thread, so `wals_` is still visible to the write thread.

riversand963: > Yes, during recovery, `GetSortedWalFiles` may need to access `wals_` concurrently with `LogAndApply`.
No, `GetSortedWalFiles` cannot happen concurrently with `Recover()`. When a thread is executing `Recover()`, no `db` ptr has been returned to the application yet. How can the application call `DB::GetSortedWal()` without a db ptr?
`GetSortedWals` is a `DB`-level API and can happen concurrently with methods that call `LogAndApply` except `Recover()`.
> 
> For `GetSortedWalFiles`, since it's in the same recovery thread, the `wals_` written during manifest recovery is visible.
> 
It's not necessarily in the same thread. It's an API, and application can call it anywhere as long as the application has the `db` ptr after recovery.

> For `LogAndApply`, since DB mutex is always acquired before `ProcessManifestWrite`, even if it happens concurrently with the recovery thread, as long as it happens after `VersionSet::Recover`, the mutex establishes a happens before relationship between the recovery thread and the write thread, so `wals_` is still visible to the write thread.

Threads performing db recovery during db open cannot happen concurrently with application write thread.

cheng-chang: I thought the `GetSortedWalFiles` you mentioned refers to getting the sorted WALs in `DB::Recover`.
`WalSet::CheckWals` is only called in `DB::Recover` after recovering from MANIFEST and before recovering from WAL.
`GetSortedWalFiles` public API won't verify the tracked WALs, the WAL verification only happens during `Recovery`.

cheng-chang: `Threads performing db recovery during db open cannot happen concurrently with application write thread.`

Yes, here I'm referring to the `LogAndApply`s called during `Recover`.

db/memtable_list.cc
```diff
@@ -726,6 +741,15 @@ Status InstallMemtableAtomicFlushResults(
     assert(0 == num_entries);
   }
 
+  std::unique_ptr<VersionEdit> wal_deletions;
```
cheng-chang, line 744: @riversand963  `InstallMemtableAtomicFlushResults` didn't handle 2pc mode, is 2pc not supported for atomic flush or did I miss anything?

db/version_set.h
```diff
@@ -1262,6 +1284,7 @@ class VersionSet {
   Status VerifyFileMetadata(const std::string& fpath,
                             const FileMetaData& meta) const;
 
+  // Protected by DB mutex.
```
cheng-chang, line 1287: @riversand963 I decided to protect it by db mutex for simplicity.

cheng-chang:
@riversand963  Current status: the implementation is ready for review, I'm adding more end-to-end tests.

db/version_set.cc
```diff
@@ -4321,9 +4369,7 @@ Status VersionSet::LogAndApplyHelper(ColumnFamilyData* cfd,
   edit->SetLastSequence(db_options_->two_write_queues ? last_allocated_sequence_
                                                       : last_sequence_);
 
-  Status s = builder->Apply(edit);
-
-  return s;
+  return builder ? builder->Apply(edit) : Status::OK();
```
riversand963, line 4372: Maybe add an assertion or comment to explain why `Status::OK()` is returned when builder is nullptr.

db/wal_edit.h
```diff
@@ -98,30 +89,36 @@ JSONWriter& operator<<(JSONWriter& jw, const WalAddition& wal);
...
   Status DecodeFrom(Slice* src);
 
   std::string DebugString() const;
 
+  void Reset() { number_ = EMPTY_; }
+
+  bool IsEmpty() const { return number_ == EMPTY_; }
+
  private:
+  static constexpr WalNumber EMPTY_ = 0;
```
riversand963, line 114: Nit: kEmpty according to naming convention.

riversand963 (commented):
Thanks @cheng-chang for the PR! Left a few more comments.

db/version_edit.h
```diff
@@ -453,27 +453,32 @@ class VersionEdit {
...
 
-  bool HasWalDeletion() const { return !wal_deletions_.empty(); }
+  bool IsWalManipulation() const { return IsWalAddition() || IsWalDeletion(); }
 
   // Number of edits
   size_t NumEntries() const {
     return new_files_.size() + deleted_files_.size() +
            blob_file_additions_.size() + blob_file_garbages_.size() +
-           wal_additions_.size() + wal_deletions_.size();
+           wal_additions_.size() + (wal_deletion_.GetLogNumber() != 0);
```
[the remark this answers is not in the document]
cheng-chang: this is outdated

db/version_edit_handler.cc
```diff
@@ -82,6 +82,16 @@ void VersionEditHandler::Iterate(log::Reader& reader, Status* log_read_status,
...
+  if (s.ok()) {
+    // Consider the case:
+    // 1. DB is opened with tracking WAL in MANIFEST is enabled,
+    // 2. then DB is reopened with tracking WAL disabled,
+    // 3. then DB is reopened with tracking WAL enabled.
+    // In step 2, the deleted WALs are not tracked in MANIFEST,
+    // so in step 3, during recovery, we should delete the WALs.
+    s = version_set_->wals_.DeleteWalsBefore(
+        WalDeletion(version_set_->MinLogNumberToKeep()));
+  }
```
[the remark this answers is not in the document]
cheng-chang: this is outdated

db/version_set.cc
```diff
@@ -4011,6 +4026,18 @@ Status VersionSet::ProcessManifestWrites(
...
+          s = wals_.AddWals(e->GetWalAdditions());
+          if (!s.ok()) {
+            break;
+          }
+        } else if (e->IsWalDeletion()) {
+          s = wals_.DeleteWals(e->GetWalDeletions());
+          if (!s.ok()) {
+            break;
+          }
+        }
```
riversand963, line 4039: I think this block of applying `e` to `wals_` should occur after writing and syncing MANIFEST succeeds.

db/version_set.cc
riversand963: Sorry for not being specific.
I think under current circumstances, the following will be clearer.
```
// The builder can be nullptr only if edit is wal operation,
// and we return Status::OK() in this case.
assert(builder || edit->IsWalOperation());
```

db/version_set.cc
cheng-chang: `AddWal` and `DeleteWal` will do some internal consistency check for the edit.
Even if `AddWal`/`DeleteWal` succeeds, and writing to MANIFEST fails, since it only modifis the in-memory state, it should be fine.
What's your concern here?

riversand963: > `AddWal` and `DeleteWal` will do some internal consistency check for the edit.
> Even if `AddWal`/`DeleteWal` succeeds, and writing to MANIFEST fails, since it only modifis the in-memory state, it should be fine.
> What's your concern here?

You are accessing `wals_` without dbmutex here, which violates the contract.
Consequently, another thread can access `wals_` concurrently before this thread writes to MANIFEST. Should writing to MANIFEST fails, the reader of `wals_` may have already done something by mistake using the unpersisted `wals_`. We should avoid this because it makes recovery hard to reason.
Also note the function's name is "log and apply".

cheng-chang: Make sense. The check of `AddWal` and `DeleteWal` are more for consistency checking during recovery, if the checks fail here, it's a programming error. So I'll update `wals_` after MANIFEST write.

cheng-chang:
@riversand963 thanks for comments, PTAL

db/db_impl/db_impl_open.cc
```diff
@@ -582,21 +582,41 @@ Status DBImpl::Recover(
...
         }
       }
     }
 
-    if (logs.size() > 0) {
+    if (immutable_db_options_.track_and_verify_wals_in_manifest) {
+      // Verify WALs in MANIFEST.
+      s = versions_->GetWalSet().CheckWals(env_, log_files);
+      if (!s.ok()) {
+        return Status::Corruption("Failed to verify WALs tracked in MANIFEST",
```
riversand963, line 595: `CheckWals(...)` can return `IOError` if `GetFileSize` encounters an `IOError`. Maybe just propagate the IO error to application.

db/db_impl/db_impl_open.cc
```diff
@@ -582,21 +582,41 @@ Status DBImpl::Recover(
...
+        WalNumber number = wal.first;
+        edit.DeleteWal(number);
+      }
+      s = versions_->LogAndApplyToDefaultColumnFamily(&edit, &mutex_);
+      if (!s.ok()) {
+        return s;
+      }
+    }
+
+    if (log_files.size() > 0) {
```
riversand963, line 611: Nit: maybe use `!empty()`.

db/db_impl/db_impl_open.cc
```diff
@@ -582,21 +582,41 @@ Status DBImpl::Recover(
...
-    if (logs.size() > 0) {
+    if (immutable_db_options_.track_and_verify_wals_in_manifest) {
+      // Verify WALs in MANIFEST.
+      s = versions_->GetWalSet().CheckWals(env_, log_files);
+      if (!s.ok()) {
+        return Status::Corruption("Failed to verify WALs tracked in MANIFEST",
+                                  s.ToString());
+      }
+    } else if (!versions_->GetWalSet().GetWals().empty()) {
+      // Tracking is disabled, clear previously tracked WALs from MANIFEST.
```
riversand963, line 599: A little more explanation will be helpful: e.g. to prevent future recovery with track_wal=true to fail.

db/version_set.cc
```diff
@@ -3831,16 +3832,27 @@ Status VersionSet::ProcessManifestWrites(
...
-        builder = builder_guards.back()->version_builder();
-      }
-      assert(builder != nullptr);  // make checker happy
+        bool is_all_wal_maniputations = true;
+        for (const auto& e : last_writer->edit_list) {
+          if (!e->IsWalManipulation()) {
+            is_all_wal_maniputations = false;
+            break;
+          }
+        }
```
riversand963, line 3841: I wonder whether this computation can be moved to the constructor of `ManifestWriter`, and stored in a new member of `ManifestWriter`.

db/wal_edit.cc
```diff
@@ -117,31 +107,23 @@ std::string WalDeletion::DebugString() const {
...
+    std::stringstream ss;
+    ss << "WAL " << wal.GetLogNumber() << " is created more than once";
+    return Status::Corruption("WalSet", ss.str());
   }
   // If the WAL has synced size, it must >= the previous size.
-  if (existing && it->second.HasSyncedSize() &&
-      (!wal.GetMetadata().HasSyncedSize() ||
-       wal.GetMetadata().GetSyncedSizeInBytes() <
-           it->second.GetSyncedSizeInBytes())) {
+  if (wal.GetMetadata().HasSyncedSize() && existing &&
```
riversand963, line 116: If we recycle wal file numbers, is it possible that we sync the oldest wal for a smaller size?
cc @ajkr

db/version_set.cc
```diff
@@ -5366,6 +5419,22 @@ Status VersionSet::WriteCurrentStateToManifest(
     }
   }
 
+  // Save WALs.
+  if (!wal_additions.GetWalAdditions().empty()) {
+    TEST_SYNC_POINT_CALLBACK(
+        "VersionSet::WriteCurrentStateToManifest:SaveWal",
+        reinterpret_cast<void*>(const_cast<VersionEdit*>(&wal_additions)));
```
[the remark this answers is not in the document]
cheng-chang: compile fails with `const VersionEdit* cannot convert to void*`, so needs to at least do the const cast, removed the reinterpret cast.

db/wal_edit.cc
cheng-chang: Assume that WAL 10 is recycled, when it becomes obsolete after flushing, it will be removed from MANIFEST. So afterwards, even if log number 10 is reused, `WalSet` shouldn't contain information for log number 10.

cheng-chang:
@riversand963 thanks for comments! PTAL.

db/db_impl/db_impl_open.cc
```diff
@@ -582,21 +582,40 @@ Status DBImpl::Recover(
...
+      s = versions_->GetWalSet().CheckWals(env_, wal_files);
+    } else if (!versions_->GetWalSet().GetWals().empty()) {
+      // Tracking is disabled, clear previously tracked WALs from MANIFEST,
+      // otherwise, in the future, if WAL tracking is enabled again,
+      // since the WALs deleted when WAL tracking is disabled are not persisted
+      // into MANIFEST, WAL check may fail.
+      VersionEdit edit;
+      for (const auto& wal : versions_->GetWalSet().GetWals()) {
+        WalNumber number = wal.first;
+        edit.DeleteWal(number);
```
riversand963, line 602: In `WalSet`, it's possible to mark wals with smaller file number as deleted, as shown in `WalSet::DeleteWalsBefore()`.
When we write MANIFEST, can we log the operation as "mark all wals smaller than a specific file number as deleted"?

db/version_set.cc
```diff
@@ -4302,7 +4343,6 @@ Status VersionSet::LogAndApplyHelper(ColumnFamilyData* cfd,
   (void)cfd;
 #endif
   mu->AssertHeld();
-  assert(!edit->IsColumnFamilyManipulation());
```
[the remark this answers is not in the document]
cheng-chang: deleted by accident

db/db_impl/db_impl_open.cc
cheng-chang: this is what I'm thinking of doing, it depends on how we are going to write the deleted WALs into MANIFEST. If we write the deleted WAL edits in PurgeObsoleteFiles, then we need to delete individual WALs. If we piggyback the deleted WAL edits during flushing, then we can use `DeleteWalsBefore`. 
I prefer to making this decision in next PR where we decide where to delete the WALs. In my current prototype, I'm going to use `DeleteWalsBefore`, so the `DeleteWal` API might be removed in next PR.

riversand963 (approved):

PR #7257, opened 2020-08-14
Track WAL in MANIFEST: track WAL creation

cheng-chang:
When creating a new WAL, track the WAL's log number in MANIFEST.

Test Plan:
will add tests to DBTest after merging https://github.com/facebook/rocksdb/pull/7256.

### DIFF
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -1388,7 +1388,8 @@ Status DB::Open(const DBOptions& db_options, const std::string& dbname,
                       !kSeqPerBatch, kBatchPerTxn);
 }
 
-IOStatus DBImpl::CreateWAL(uint64_t log_file_num, uint64_t recycle_log_number,
+IOStatus DBImpl::CreateWAL(bool is_db_mutex_locked, uint64_t log_file_num,
+                           uint64_t recycle_log_number,
                            size_t preallocate_block_size,
                            log::Writer** new_log) {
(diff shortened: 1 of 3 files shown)


db/db_impl/db_impl_open.cc
```diff
@@ -1426,6 +1427,29 @@ IOStatus DBImpl::CreateWAL(uint64_t log_file_num, uint64_t recycle_log_number,
     *new_log = new log::Writer(std::move(file_writer), log_file_num,
                                immutable_db_options_.recycle_log_file_num > 0,
                                immutable_db_options_.manual_wal_flush);
+    {
+      // Track WAL in MANIFEST.
+      if (!is_db_mutex_locked) {
+        mutex_.Lock();
+      }
+
+      VersionEdit edit;
```
riversand963, line 1436: If WAL or WAL dir is not synced, what do we do with this edit?

cheng-chang:
@riversand963 Thanks for review. I'm going to create one PR containing all the logic for tracking WAL's creation, syncing, closing, and obsoletion. Hope that would make it easier to review.

db/db_impl/db_impl_open.cc
cheng-chang: The edit is written to manifest, but on recovery, we'll know that the wal is not synced, so we'll skip the check. We track the WAL creation here because we want to track the whole lifecycle of the WAL, instead of just tracking the syncing event.

PR #7260, opened 2020-08-14
Re-enable param tests for backup engine

gg814:
Summary:
The param tests did not take any effect previously. This PR re-enables it.

Test plan:
Some manual tests and `./backupable_db_test`.

### DIFF
--- a/utilities/backupable/backupable_db_test.cc
+++ b/utilities/backupable/backupable_db_test.cc
@@ -747,7 +747,8 @@ class BackupableDBTest : public testing::Test {
   }
 
   void OpenDBAndBackupEngine(bool destroy_old_data = false, bool dummy = false,
-                             ShareOption shared_option = kShareNoChecksum) {
+                             ShareOption shared_option = kShareNoChecksum,
+                             bool share_options_param_test = false) {
     // reset all the defaults
     test_backup_env_->SetLimitWrittenFiles(1000000);
     test_db_env_->SetLimitWrittenFiles(1000000);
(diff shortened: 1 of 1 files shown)


utilities/backupable/backupable_db_test.cc
```diff
@@ -763,14 +764,21 @@ class BackupableDBTest : public testing::Test {
     db_.reset(db);
     backupable_options_->destroy_old_data = destroy_old_data;
     backupable_options_->share_table_files = shared_option != kNoShare;
-    backupable_options_->share_files_with_checksum =
-        shared_option == kShareWithChecksum;
+    if (!share_options_param_test) {
```
pdillinger, line 767: I don't understand why we don't always overwrite the share_files_with_checksum setting

gg814: For `BackupableDBTestWithParam`, the share_files_with_checksum setting is initialized at the beginning of the test. I think we should not overwrite it by default as it may conflict with the original purpose of the test. (But one can still overwrite it on purpose by calling `OpenDBAndBackupEngine` in a param test with `share_options_param_test` being false.

pdillinger: OK, BackupableDBTestWithParam extends BackupableDBTest, so we should use virtual function and override in some way to make this cleaner than remembering to call a different function in any BackupableDBTestWithParam

pdillinger:
> Thank you for your pull request. We require contributors to sign our Contributor License Agreement, and yours needs attention.

Code was written while original CLA was in effect.

PR #7267, opened 2020-08-17
Eliminates a no-op compaction upon snapshot release when disabling auto compactions

Connor1996:
#### Summary:
After releasing a snapshot, it checks whether it is suitable to trigger bottom compactions. 
When disabling auto compactions, it may still schedule compaction when releasing a snapshot. Whereas no compaction job will be actually handled, so the state of LSM is not changed and compaction will be triggered again and again every time releasing a snapshot.

Too frequent compactions lead to high CPU usage and high db_mutex lock contention which affects foreground write duration finally.

#### Test Plan:
- make check 
- manual test

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 ### Bug fixes
 * Fix a performance regression introduced in 6.4 that makes a upper bound check for every Next() even if keys are within a data block that is within the upper bound.
 * Fix a possible corruption to the LSM state (overlapping files within a level) when a `CompactRange()` for refitting levels (`CompactRangeOptions::change_level == true`) and another manual compaction are executed in parallel.
+* Fix possible endless compactions scheduled when options.disable-auto-compactions = true.
 
 ### New Features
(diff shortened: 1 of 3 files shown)


Connor1996:
@ajkr PTAL

Connor1996:
> Hm, actually upon looking further I'm not able to repro the problem using the provided test (without the fix). The test causes one CFD to be enqueued for compaction unnecessarily. But no compaction gets picked due to
> 
> https://github.com/facebook/rocksdb/blob/ce4192375de7215b9bca5ec545de6a60e66b7026/db/db_impl/db_impl_compaction_flush.cc#L2630
> 
> . Then I see one occurrence of "Compaction nothing to do" in the LOG, and then nothing further gets scheduled.

Yes, this unit test is just to ensure that the compaction shouldn't be scheduled when disabling auto compaction. For the case, it has to get and release snapshot persistently which is typical for scan workload and causes a lot of `Compaction nothing to do`. But it is hard for a unit test to do that, so I only check it by manual.

Do you have some suggestions?

Connor1996:
Thanks for your advice. PTAL again @ajkr

ajkr:
> LGTM. It looks like a lot of HISTORY.md changes related to the markdown.

Regarding the markdown changes, if you did intend to include them, would you mind separating them into another PR? Also, we've recently decided to adopt ISO 8601 date format (YYYY-MM-DD). So if you do decide to format all of HISTORY.md, it'd be nice to also change the dates (e.g., "6/12/2020" -> "2020-06-12").

Connor1996:
@ajkr Not intend to do that, it's just changed automatically by markdown-lint. Already revert it.

PR #7270, opened 2020-08-17
fix doc about kTolerateCorruptedTailRecords recovery

ajkr:
- Made it clear only one record in the tail is allowed to have a problem
- Added detail about the valid use case instead of calling it legacy behavior

### DIFF
--- a/include/rocksdb/options.h
+++ b/include/rocksdb/options.h
@@ -303,8 +303,18 @@ struct ColumnFamilyOptions : public AdvancedColumnFamilyOptions {
 
 enum class WALRecoveryMode : char {
   // Original levelDB recovery
-  // We tolerate incomplete record in trailing data on all logs
-  // Use case : This is legacy behavior
+  //
+  // We tolerate one incomplete record caused by crash in the trailing data of
+  // any log. Zeroed bytes from preallocation are also tolerated in the trailing
+  // data of any log.
+  //
+  // Use case : Applications for which updates, once applied, must not be rolled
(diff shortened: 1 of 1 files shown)


include/rocksdb/options.h
```diff
@@ -303,8 +303,23 @@ struct ColumnFamilyOptions : public AdvancedColumnFamilyOptions {
...
+  // Use case: Applications for which updates, once applied, must not be rolled
+  // back even after a crash-recovery. By default RocksDB guarantees this as
+  // long as `WritableFile::Append()` writes are durable. In case the user needs
+  // the guarantee in more situations (e.g., when `WritableFile::Append()`
+  // writes to page cache, but the user desires this guarantee in face of power-
+  // loss crash-recovery), RocksDB offers various mechanisms to additionally
+  // invoke `WritableFile::Sync()` in order to strengthen the guarantee.
+  //
+  // This differs from `kPointInTimeRecovery` in that, in case a corruption is
+  // detected during recovery, this mode will refuse to open the DB. Whereas,
```
riversand963, line 320: Just to clarify, `kTolerateCorruptedTailRecords` will refuse to open DB if the corruption is not the trailing record?

ajkr: Yes, and additionally if it's not able to tell if the corruption is the trailing record. For example, an incomplete write to a preallocated file would lead to a checksum mismatch because the record contains a bunch of valid bytes followed by a bunch of zeroes.

PR #7275, opened 2020-08-18
Track WAL in MANIFEST: add option track_and_verify_wals_in_manifest

cheng-chang:
This option determines whether WALs will be tracked in MANIFEST and verified on recovery.

Test Plan:
db_options_test
options_test


options/db_options.h
```diff
@@ -22,6 +22,7 @@ struct ImmutableDBOptions {
   bool create_missing_column_families;
   bool error_if_exists;
   bool paranoid_checks;
+  bool check_wal;
```
riversand963, line 25: NIt: per our team discussion, maybe it's better to name it `should_check_wal`? cc @pdillinger

pdillinger: On further reflection (re naming in #7214), I am more critical of re-using verb member variable names in classes (as accessors) than having them in structs to begin with. This struct already has a lot of verb names. ¯\_(ツ)_/¯

include/rocksdb/options.h
```diff
@@ -374,6 +374,14 @@ struct DBOptions {
   // Default: true
   bool paranoid_checks = true;
 
+  // If true, check on-disk WALs against WAL information stored in MANIFEST
+  // during recovery.
+  // It checks whether there are missing or corrupted WALs.
+  // No matter whether this is true or false, the WAL information are always
+  // tracked by MANIFEST.
+  // Default: false
+  bool check_wal = false;
```
pdillinger, line 383: Why wouldn't this be true by default if we always pay the ongoing cost of tracking, and the only check is during recovery?

When / why would someone want to use false?

include/rocksdb/options.h
```diff
@@ -374,6 +374,14 @@ struct DBOptions {
   // Default: true
   bool paranoid_checks = true;
 
+  // If true, check on-disk WALs against WAL information stored in MANIFEST
+  // during recovery.
```
pdillinger, line 378: "check" is not very descriptive to me. How about something like "If true, fail recovery if there is an inconsistency between WAL information in MANIFEST and actual WALs on disk"?

What sort of corruption might this catch that's not otherwise caught? (Don't we always want to report corruption?)

pdillinger: Also, do we quietly ignore extra/untracked WALs or fail in that case?

options/db_options.h
pdillinger: But I do think it's worth thinking about a better name. `verify_wals`? `verify_wals_with_manifest`?

cheng-chang: I think `check_wal` delivers the meaning, there are other "checks", such as "paranoid_checks", "check_sst_file_size", etc.

include/rocksdb/options.h
cheng-chang: If the user does not even sync WALs, then the user might skip checking the WALs since there is no guarantee of WAL's existence or completeness on recovery.

include/rocksdb/options.h
cheng-chang: missing or corrupted WAL will be reported

cheng-chang:
@riversand963 actually, I think Peter's suggestion makes sense. Since we always track WAL in MANIFEST, it doesn't make too much sense to disable checking WALs during recovery. This PR may be unnecessary. What's your thought?

riversand963:
> @riversand963 actually, I think Peter's suggestion makes sense. Since we always track WAL in MANIFEST, it doesn't make too much sense to disable checking WALs during recovery. This PR may be unnecessary. What's your thought?

It sounds a good idea to always check WAL.
Another question, would some users want to disable WAL tracking sometimes for performance considerations? If so, the wal verification logic should keep this in mind.

cheng-chang:
After offline discussion, we'll keep this `check_wal` option, if it's false, then WALs are not tracked in MANIFEST and won't be checked on recovery.

options/db_options.h
cheng-chang: @riversand963 @pdillinger  does `track_and_verify_wals` sound good to you?

include/rocksdb/options.h
pdillinger: It's still not clear to me how extra/untracked WALs will be handled. Can a DB switch between using track_and_verify_wals_in_manifest=true and false?

include/rocksdb/options.h
```diff
@@ -390,6 +390,18 @@ struct DBOptions {
...
+  //
+  // If a WAL is tracked in MANIFEST but is missing from disk on recovery,
+  // or the size of the tracked WAL is larger than the WAL's on-disk size,
+  // an error is reported and recovery is aborted.
+  //
+  // If a WAL is not tracked in MANIFEST, then no verification will happen
+  // during recovery.
+  //
+  // Default: false
+  bool track_and_verify_wals_in_manifest = false;
```
pdillinger, line 403: If this option is not currently connected to anything (we don't have diff stacks unfortunately), please drop in something like

    // FIXME(cheng): This option is part of a work in progress and does not yet work

in case through some mixup or unexpected absence from work, a release only has some of your changes

pdillinger (approved):
Even though the new name is kind of long, I think it's clearer and therefore better.

I don't know this area well enough to endorse or reject the overall strategy, but my previous concerns are sufficiently addressed. Some additional comments / clarification requested.

include/rocksdb/options.h
cheng-chang: Yes, switches between true and false can happen. The verification will only verify those tracked WALs.

PR #7277, opened 2020-08-18
Fix a timer_test deadlock

jay-zhuang:
There's a potential deadlock caused by MockTimeEnv time value get to a large number, which causes TimedWait() wait forever. The test misuses the microseconds as seconds, making it more likely to happen.

### DIFF
--- a/test_util/mock_time_env.h
+++ b/test_util/mock_time_env.h
@@ -44,10 +44,10 @@ class MockTimeEnv : public EnvWrapper {
   // TODO: this is a workaround for the different behavior on different platform
   // for timedwait timeout. Ideally timedwait API should be moved to env.
   // details: PR #7101.
-  void InstallTimeWaitCallbackForMACOS() {
-#if defined(OS_MACOSX) && !defined(NDEBUG)
+  void InstallTimedWaitFixCallback() {
     SyncPoint::GetInstance()->DisableProcessing();
     SyncPoint::GetInstance()->ClearAllCallBacks();
+#if defined(OS_MACOSX) && !defined(NDEBUG)
     // This is an alternate way (vs. SpecialEnv) of dealing with the fact
     // that on some platforms, pthread_cond_timedwait does not appear to
     // release the lock for other threads to operate if the deadline time
@@ -61,8 +61,8 @@ class MockTimeEnv : public EnvWrapper {
             *reinterpret_cast<uint64_t*>(arg) = this->RealNowMicros() + 1000;
           }
         });
-    SyncPoint::GetInstance()->EnableProcessing();
 #endif  // OS_MACOSX && !NDEBUG
+    SyncPoint::GetInstance()->EnableProcessing();
   }
 
  private:
--- a/test_util/mock_time_env.h
+++ b/test_util/mock_time_env.h
@@ -41,6 +41,30 @@ class MockTimeEnv : public EnvWrapper {
     current_time_ = time;
   }
 
+  // TODO: this is a workaround for the different behavior on different platform
+  // for timedwait timeout. Ideally timedwait API should be moved to env.
+  // details: PR #7101.
+  void InstallTimeWaitCallbackForMACOS() {
+#if defined(OS_MACOSX) && !defined(NDEBUG)
+    SyncPoint::GetInstance()->DisableProcessing();
+    SyncPoint::GetInstance()->ClearAllCallBacks();
+    // This is an alternate way (vs. SpecialEnv) of dealing with the fact
+    // that on some platforms, pthread_cond_timedwait does not appear to
+    // release the lock for other threads to operate if the deadline time
+    // is already passed. (TimedWait calls are currently a bad abstraction
+    // because the deadline parameter is usually computed from Env time,
+    // but is interpreted in real clock time.)
+    SyncPoint::GetInstance()->SetCallBack(
+        "InstrumentedCondVar::TimedWaitInternal", [&](void* arg) {
+          uint64_t time_us = *reinterpret_cast<uint64_t*>(arg);
+          if (time_us < this->RealNowMicros()) {
+            *reinterpret_cast<uint64_t*>(arg) = this->RealNowMicros() + 1000;
+          }
+        });
+    SyncPoint::GetInstance()->EnableProcessing();
+#endif  // OS_MACOSX && !NDEBUG
+  }
+
  private:
   std::atomic<uint64_t> current_time_{0};
 };
--- a/util/timer.h
+++ b/util/timer.h
@@ -147,8 +147,9 @@ class Timer {
   void TEST_WaitForRun(std::function<void()> callback = nullptr) {
     InstrumentedMutexLock l(&mutex_);
     // It act as a spin lock
-    while (executing_task_ || (!heap_.empty() &&
-           heap_.top()->next_run_time_us <= env_->NowMicros())) {
+    while (executing_task_ ||
+           (!heap_.empty() &&
+            heap_.top()->next_run_time_us <= env_->NowMicros())) {
       cond_var_.TimedWait(env_->NowMicros() + 1000);
     }
     if (callback != nullptr) {
@@ -157,8 +158,9 @@ class Timer {
     cond_var_.SignalAll();
     do {
       cond_var_.TimedWait(env_->NowMicros() + 1000);
-    } while (executing_task_ || (!heap_.empty() &&
-             heap_.top()->next_run_time_us <= env_->NowMicros()));
+    } while (
+        executing_task_ ||
+        (!heap_.empty() && heap_.top()->next_run_time_us <= env_->NowMicros()));
   }
 
   size_t TEST_GetPendingTaskNum() const {
--- a/util/timer.h
+++ b/util/timer.h
@@ -138,8 +138,15 @@ class Timer {
   }
 
 #ifndef NDEBUG
+  // Wait until Timer starting waiting, call the optional callback, then wait
+  // for Timer waiting again.
+  // Tests can provide a custom env object to mock time, and use the callback
+  // here to bump current time and trigger Timer. See timer_test for example.
+  //
+  // Note: only support one caller of this method.
   void TEST_WaitForRun(std::function<void()> callback = nullptr) {
     InstrumentedMutexLock l(&mutex_);
+    // It act as a spin lock
     while (!heap_.empty() &&
            heap_.top()->next_run_time_us <= env_->NowMicros()) {
       cond_var_.TimedWait(env_->NowMicros() + 1000);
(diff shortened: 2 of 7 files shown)


pdillinger:
Btw, I did not intend for SafeMockTimeEnv to be migrated out of a specific test file because tests should own setting up, enabling, and tearing down sync points. Since SafeMockTimeEnv depends on sync points, it breaks that abstraction and will lead someone to confusion or worse in developing a test using sync points. From #7101:

    Intended follow-up:
    
    Remove TimedWait from the public API of InstrumentedCondVar, and only
    make that accessible through Env by passing in an InstrumentedCondVar and
    a deadline. Then the Env implementations mocking time can fix this problem
    without using sync points. (Test infrastructure using sync points interferes
    with individual tests' control over sync points.)

util/timer_test.cc
```diff
@@ -11,126 +11,55 @@ namespace ROCKSDB_NAMESPACE {
...
-  {
-    InstrumentedMutexLock l(&mutex);
-    while(count < kIterations) {
-      time_counter += kSecond;
-      mock_env_->set_current_time(time_counter);
-      test_cv.TimedWait(time_counter);
-    }
-  }
+  mock_time_sec += kInitDelaySec;
+  timer.TEST_WaitForRun([&] { mock_env_->set_current_time(mock_time_sec); });
```
pdillinger, line 34: I don't know what TEST_WaitForRun is supposed to do, and it seems deeply flawed outside of a mock time environment. Is it only intended for a mock time environment?

Wouldn't it be much clearer to assert_eq(0, count) immediately before this line and assert_eq(1, count) immediately after? (I think that's the intention here.)

jay-zhuang:
> Btw, I did not intend for SafeMockTimeEnv to be migrated out of a specific test file because tests should own setting up, enabling, and tearing down sync points. Since SafeMockTimeEnv depends on sync points, it breaks that abstraction and will lead someone to confusion or worse in developing a test using sync points. From #7101:
> 
> ```
> Intended follow-up:
> 
> Remove TimedWait from the public API of InstrumentedCondVar, and only
> make that accessible through Env by passing in an InstrumentedCondVar and
> a deadline. Then the Env implementations mocking time can fix this problem
> without using sync points. (Test infrastructure using sync points interferes
> with individual tests' control over sync points.)
> ```

Make sense. I changed it to a function in MockTimeEnv to install the callback. Which needs to be called explicitly from the unittest.
For the follow-up, I think we should have a separate diff moving timedwait API from condVar to env.

util/timer_test.cc
jay-zhuang: The function is the same as https://github.com/facebook/rocksdb/blob/fdf882ded218344c136c97daf76dfb59e4bc155f/util/repeatable_thread.h#L64 which is a similar thread with timedwait. I added a comment on that.

And added assert_eq(0, count) for all tests.

util/timer_test.cc
```diff
@@ -15,122 +15,56 @@ class TimerTest : public testing::Test {
...
+
+  void SetUp() override { mock_env_->InstallTimeWaitCallbackForMACOS(); }
 };
 
 TEST_F(TimerTest, SingleScheduleOnceTest) {
-  const int kIterations = 1;
-  uint64_t time_counter = 0;
-  mock_env_->set_current_time(0);
-
-  InstrumentedMutex mutex;
```
riversand963, line 49: This line seems suspicious to me (although you have removed it).
If we are going to use `MockTimeEnv`, shouldn't we specify the env when creating this `mutex`?

test_util/mock_time_env.h
```diff
@@ -41,6 +41,30 @@ class MockTimeEnv : public EnvWrapper {
     current_time_ = time;
   }
 
+  // TODO: this is a workaround for the different behavior on different platform
+  // for timedwait timeout. Ideally timedwait API should be moved to env.
+  // details: PR #7101.
+  void InstallTimeWaitCallbackForMACOS() {
+#if defined(OS_MACOSX) && !defined(NDEBUG)
```
pdillinger, line 48: Nice compromise. :) Two concerns:

That this function only needs to fix something on MacOS is an implementation detail. (Perhaps `InstallTimeWaitFixCallback`?)

Related to that, we could easily see different test behavior by platform due to this function having no effect on SyncPoints on most platforms and resetting them on MacOS. As much as possible, behavior should be consistent across platforms, with everything except SetCallBack outside of the ifdef (as I did in SafeMockTimeEnv).

util/timer.h
```diff
@@ -138,8 +138,15 @@ class Timer {
   }
 
 #ifndef NDEBUG
+  // Wait until Timer starting waiting, call the optional callback, then wait
+  // for Timer waiting again.
+  // Tests can provide a custom env object to mock time, and use the callback
+  // here to bump current time and trigger Timer. See timer_test for example.
+  //
+  // Note: only support one caller of this method.
   void TEST_WaitForRun(std::function<void()> callback = nullptr) {
```
pdillinger, line 147: Since Run() now releases the lock while executing the registered timer function, don't we have a race condition where this function thinks the timer is waiting but in fact it's executing the registered timer function with the mutex unlocked?

jay-zhuang:
> > There's a potential deadlock between `start the timer thread` and `start waiting`. If we start waiting first before timer thread is started, they will wait for each forever and cause timer_test hang.
> > Refactor the test to use `TEST_WaitForRun()` to control the timer running.
> 
> Can you elaborate on the `wait for each other` part?

Seems the problem is actually the test is using microsecond as second: https://github.com/facebook/rocksdb/blob/6ac1d25fd09b407b475174748c39bf6816fb2049/util/timer_test.cc#L69-L73

Which could cause the value (https://github.com/facebook/rocksdb/blob/69760b4d05a6e96bdf1175c72e4ac8356deacfde/util/timer.h#L21) very high, like `3296000002000000`, which cause the timer keep waiting here https://github.com/facebook/rocksdb/blob/69760b4d05a6e96bdf1175c72e4ac8356deacfde/util/timer.h#L219:

util/timer.h
jay-zhuang: Good point. I added `executing_task_` check here.

test_util/mock_time_env.h
jay-zhuang: Make sense. Updated.

util/timer_test.cc
```diff
@@ -15,122 +15,56 @@ class TimerTest : public testing::Test {
...
-      mock_env_->set_current_time(time_counter);
-      test_cv.TimedWait(time_counter);
-    }
-  }
+  mock_time_sec += kInitDelaySec;
+  timer.TEST_WaitForRun([&] { mock_env_->set_current_time(mock_time_sec); });
 
   ASSERT_TRUE(timer.Shutdown());
-
   ASSERT_EQ(1, count);
```
pdillinger, line 40: Since we intend for something stronger to hold, that count is 1 before Shutdown, it would be clearer to me reading it if the assertion came before shutdown (as you've done in other tests, thanks!)

jay-zhuang:
> My concerns so far are addressed. Thanks! Ping me if you'd like me to review the remaining details instead of (or in addition to) Yanqin. :)

Yes, please. Thanks.

util/timer_test.cc
```diff
@@ -15,122 +15,56 @@ class TimerTest : public testing::Test {
...
-  {
-    InstrumentedMutexLock l(&mutex);
-    while(count < kIterations) {
-      time_counter += kSecond;
-      mock_env_->set_current_time(time_counter);
-      test_cv.TimedWait(time_counter);
-    }
-  }
+  mock_time_sec += kInitDelaySec;
+  timer.TEST_WaitForRun([&] { mock_env_->set_current_time(mock_time_sec); });
```
pdillinger, line 37: It seems like it would be simpler vs. the `mock_time_sec` and `set_current_time` if MockTimeEnv implemented SleepForMicroseconds as SpecialEnv does, and use that in the TEST_WaitForRun calls instead. Then it's easier to line up the timer parameters with the mock time elapse parameters. I think everything would then use microseconds.

util/timer_test.cc
```diff
@@ -176,106 +104,76 @@ TEST_F(TimerTest, SingleScheduleRepeatedlyTest) {
...
-  {
-    InstrumentedMutexLock l(&mutex1);
-    while(count1 < kIterations1) {
-      time_counter += kSecond;
-      mock_env_->set_current_time(time_counter);
-      test_cv1.TimedWait(time_counter);
-    }
+  for (; count1 < kIterations; mock_time_sec++) {
+    timer.TEST_WaitForRun([&] { mock_env_->set_current_time(mock_time_sec); });
+    ASSERT_EQ((mock_time_sec + 2) / 2, count1);
```
pdillinger, line 130: I believe the provenance of 2 in `/ 2` is `kRepeatSec`, so `/ kRepeatSec` would be more clear.

util/timer_test.cc
```diff
@@ -15,122 +15,56 @@ class TimerTest : public testing::Test {
...
-  timer.Add(
-      [&] {
-        InstrumentedMutexLock l(&mutex2);
-        count2 += 5;
-        if (count2 >= kIterations) {
-          test_cv2.SignalAll();
-        }
-      },
-      "fn_sch_test2", 3 * kSecond, 0);
+  timer.Add([&] { count2 += 5; }, "fn_sch_test2", kInitDelay2Sec * kSecond, 0);
```
riversand963, line 54: Is `5` picked randomly?

util/timer_test.cc
riversand963: +1

util/timer_test.cc
jay-zhuang: Good idea. I think we could create a follow up diff which worth another review.

util/timer_test.cc
jay-zhuang: Yes, I just switched it `count2++`, which is consistent with other tests.

jay-zhuang:
> Better state than it was. Take these suggestions as you wish.
> 
> Critiques of timer and its tests beyond this change:
> 
> * When adding a new timer, it's not clear whether repeat_every_us is the interval between starting times of calling the function or between ending time of the last call and starting time of the next call. Comments should clarify that. (I believe it's the latter.) And ideally there would be a test that can distinguish the difference (didn't see one; shouldn't be hard).
> * Are we sufficiently testing the test? If we accidentally grant too much time in the test--or the timer has a bug wherein it triggers before the requested time--I believe many of these tests still pass. (At least the "Multiple" tests can detect some cases of this problem.) It would be nice if, perhaps at least once per test such as very end, we ensure the timer does not trigger when granted just less than the expected trigger time.

Thanks @pdillinger . I added comment and an unittest for `repeat_every_us`, which could be delayed if the function running for long timer.
And expanded a test for long repeat interval, which should not be executed during the test (until long time later).

PR #7279, opened 2020-08-18
Changes to EncryptedEnv public API

mrambacher:
Cleaned up the public API to use the EncryptedEnv.  This change will allow providers to be developed and added to the system easier in the future.  It will also allow better integration in the future with the OPTIONS file.

- The internal classes were moved out of the public API into an internal "env_encryption_ctr.h" header.  Short-cut constructors were added to provide the original API functionality.
- The APIs to the constructors were changed to take shared_ptr, rather than raw pointers or references to allow better memory management and alternative implementations.
- CreateFromString methods were added to allow future expansion to other provider and cipher implementations through a standard API.

Additionally, there was a code duplication in the NewXXXFile methods.  This common code was moved under a templatized function.

A first-pass at structuring the code was made to potentially allow multiple EncryptionProviders in a single EncryptedEnv.  The idea was that different providers may use different cipher keys or different versions/algorithms.  The EncryptedEnv should have some means of picking different providers based on information.  The groundwork was started for this (the use of the provider_ member variable was localized) but the work has not been completed.


env/env_encryption_ctr.h
```diff
@@ -0,0 +1,136 @@
...
+// Note: This is a sample implementation of BlockCipher,
+// it is NOT considered safe and should NOT be used in production.
+class ROT13BlockCipher : public BlockCipher {
+ private:
+  size_t blockSize_;
+
+ public:
+  ROT13BlockCipher(size_t blockSize) : blockSize_(blockSize) {}
+  virtual ~ROT13BlockCipher(){};
+  virtual const char* Name() const override;
```
matthewvon, line 25: Somebody once yelled at me for using both "virtual" and "override".  Might have been in a RocksDB PR, maybe not.  Suggest asking somebody if they care.

env/env_encryption_ctr.h
```diff
@@ -0,0 +1,136 @@
...
+  Status AddCipher(const char* /*cipher*/, size_t /*len*/,
+                   bool /*for_write*/) override;
+
+ protected:
+  Status TEST_Initialize() override;
+
+  // PopulateSecretPrefixPart initializes the data into a new prefix block
+  // that will be encrypted. This function will store the data in plain text.
+  // It will be encrypted later (before written to disk).
+  // Returns the amount of space (starting from the start of the prefix)
```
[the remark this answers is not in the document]
mrambacher: This code/comment was not changed at all but was a direct copy from the existing code.  I have no idea what the comment is supposed to mean.  In fact, the default implementation does nothing.

env/env_encryption_ctr.h
mrambacher: Just moved the files, but fixed it.  I

env/env_encryption.cc
```diff
@@ -361,9 +413,152 @@ Status EncryptedRandomRWFile::Close() { return file_->Close(); }
 
 // EncryptedEnv implements an Env wrapper that adds encryption to files stored
 // on disk.
-class EncryptedEnv : public EnvWrapper {
+class EncryptedEnvImpl : public EnvWrapper {
+  Status GetWritableProvider(const std::string& /*fname*/,
+                             EncryptionProvider** result) {
```
matthewvon, line 418: seems like a "protocol violation".  Should this be a std::shared_ptr<EncryptionProvider> & result ?

env/env_encryption.cc
```diff
@@ -361,9 +413,152 @@ Status EncryptedRandomRWFile::Close() { return file_->Close(); }
...
+      return Status::NotFound("No Provider specified");
+    }
+  }
+
+  template <class TypeFile>
+  Status CreateWritableCipherStream(
+      const std::string& fname, const std::unique_ptr<TypeFile>& underlying,
+      const EnvOptions& options, size_t* prefix_length,
+      std::unique_ptr<BlockAccessCipherStream>* stream) {
+    EncryptionProvider* provider = nullptr;
```
matthewvon, line 444: assume this comment everywhere ... should it be std::shared_ptr<EncryptionProvider> ?

env/env_encryption.cc
```diff
@@ -361,9 +413,152 @@ Status EncryptedRandomRWFile::Close() { return file_->Close(); }
...
+    // Read prefix (if needed)
+    AlignedBuffer buffer;
+    Slice prefix;
+    *prefix_length = provider_->GetPrefixLength();
+    if (*prefix_length > 0) {
+      // Read prefix
+      buffer.Alignment(underlying->GetRequiredBufferAlignment());
+      buffer.AllocateNewBuffer(*prefix_length);
+      Status status =
+          underlying->Read(*prefix_length, &prefix, buffer.BufferStart());
```
matthewvon, line 527: if this were a PositionedRead, the file offset does not move.  Then failure of CreateCipherStream because the file is not encrypted "just works".  If CreateCipherStream succeeds, the underlying->Skip(*prefix_length) to get offset aligned with next read point.

matthewvon: Define a specific status code for CreateCipherStream to return if not encrypted ... either a new kNotEncrypted or maybe kAborted (pick one, I do not care).  Then we need a BlockAccessCipherStream specifically for not encrypted files ... that is what every provider returns as a not-me equivalent.

matthewvon: *prefix_length gets set to zero in that case ... fixing that part of the equation.

matthewvon: status = CreateSequentialCipherStream(fname, underlying, options,
                                          &prefix_length, &stream);
    if (status.ok()) {
      result->reset(new EncryptedSequentialFile(
          std::move(underlying), std::move(stream), prefix_length));
    }
    return status;

becomes

    status = CreateRandomReadCipherStream(fname, underlying, options,
                                          &prefix_length, &stream);
    if (status.ok()) {
      underlying->Skip(prefix_length);
      result->reset(new EncryptedRandomAccessFile(
          std::move(underlying), std::move(stream), prefix_length));
    } else if (status.IsNotEncrypted() {
        status = status.OK();
    }

env/env_encryption.cc
mrambacher: It was meant to be a raw pointer and not a shared (or other pointer).  First, the returned value is not "owned" by the caller.  It is a de-reference of an object within this class.  And the callers only need access to the underlying pointer, and should not care how it is memory managed.

Second, the raw pointer would allow the EnvEncryption to store the underlying provider via different means (raw/shared/unique pointer) without changing this interface/method or any of the callers.  

I will add a comment to outline what is returned (and why).  If you feel strongly that it should be a std::shared_ptr *, I can change it, but would prefer to leave it "as written"

zhichao-cao (approved):
LGTM, thanks!

PR #7280, opened 2020-08-18
Add initial set of options for integrated blob write path

ltamasi:
Test Plan:
`make check`

### DIFF
--- a/include/rocksdb/advanced_options.h
+++ b/include/rocksdb/advanced_options.h
@@ -721,7 +721,7 @@ struct AdvancedColumnFamilyOptions {
   bool enable_blob_files = false;
   uint64_t min_blob_size = 0;
   uint64_t blob_file_size = 1 << 18;
-  CompressionType blob_compression = kNoCompression;
+  CompressionType blob_compression_type = kNoCompression;
 
   // Create ColumnFamilyOptions with default values for all fields
   AdvancedColumnFamilyOptions();
--- a/include/rocksdb/advanced_options.h
+++ b/include/rocksdb/advanced_options.h
@@ -749,7 +749,7 @@ struct AdvancedColumnFamilyOptions {
   // Default: 256 MB
   //
   // Dynamically changeable through the SetOptions() API
-  uint64_t blob_file_size = 1 << 18;
+  uint64_t blob_file_size = 1 << 28;
 
   // UNDER CONSTRUCTION -- DO NOT USE
   // The compression algorithm to use for large values stored in blob files.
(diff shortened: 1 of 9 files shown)


include/rocksdb/advanced_options.h
```diff
@@ -717,6 +717,50 @@ struct AdvancedColumnFamilyOptions {
   // data is left uncompressed (unless compression is also requested).
   uint64_t sample_for_compression = 0;
 
+  // UNDER CONSTRUCTION -- DO NOT USE
```
riversand963, line 720: Maybe return non-ok status in `ValidateOptions()` if application sets these options to non-default value?

include/rocksdb/advanced_options.h
```diff
@@ -717,6 +717,50 @@ struct AdvancedColumnFamilyOptions {
...
+  bool enable_blob_files = false;
+
+  // UNDER CONSTRUCTION -- DO NOT USE
+  // The size of the smallest value to be stored separately in a blob file.
+  // Values which have an uncompressed size smaller than this threshold are
+  // stored alongside the keys in SST files in the usual fashion. Note that
+  // enable_blob_files has to be set in order for this option to have any
+  // effect.
+  //
+  // Default: 0
```
riversand963, line 739: Does `0` mean it is disabled or all values should go to blob files?

include/rocksdb/advanced_options.h
```diff
@@ -717,6 +717,50 @@ struct AdvancedColumnFamilyOptions {
...
+
+  // UNDER CONSTRUCTION -- DO NOT USE
+  // The size limit for blob files. When writing blob files, a new file is
+  // opened once this limit is reached. Note that enable_blob_files has to be
+  // set in order for this option to have any effect.
+  //
+  // Default: 256 MB
+  //
+  // Dynamically changeable through the SetOptions() API
+  uint64_t blob_file_size = 1 << 28;
```
riversand963, line 752: Nit: just to be safe, `1ULL` or `static_cast<uint64_t>(1)`. For a shift of 28 bits, this is ok, but if somebody changes the default here to `1 << 32` by mistake, then it's going to be a problem.

include/rocksdb/advanced_options.h
ltamasi: Hm, I think if we were going to do that, it would be an extremely short-lived check: in the next step, I'm planning to start using these options and setting them in tests. So I do need them to be configurable, they're just not ready for end users yet.

include/rocksdb/advanced_options.h
ltamasi: The latter; will point it out in the comment.

include/rocksdb/advanced_options.h
ltamasi: Nice catch, thanks!

PR #7282, opened 2020-08-18
Track WAL in MANIFEST: minor udpates

cheng-chang:
The updates resolve comments left from https://github.com/facebook/rocksdb/pull/7164.

Test Plan:
wal_edit_test

### DIFF
--- a/db/wal_edit.cc
+++ b/db/wal_edit.cc
@@ -126,7 +126,7 @@ Status WalSet::AddWal(const WalAddition& wal) {
       ss << "WAL " << wal.GetLogNumber() << " is created more than once";
       return Status::Corruption("WalSet", ss.str());
     }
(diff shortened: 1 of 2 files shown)


db/wal_edit.cc
```diff
@@ -143,7 +143,7 @@ Status WalSet::AddWals(const WalAdditions& wals) {
 }
 
 Status WalSet::DeleteWal(const WalDeletion& wal) {
-  auto it = wals_.lower_bound(wal.GetLogNumber());
+  auto it = wals_.find(wal.GetLogNumber());
   // The WAL must exist and has been closed.
   if (it == wals_.end() || it->first != wal.GetLogNumber()) {
```
ltamasi, line 148: You can also remove `it->first != wal.GetLogNumber()` from here (since it's only necessary with `lower_bound`).

PR #7286, opened 2020-08-19
Fixed missing space in DeleteScheduler::BackgroundEmptyTrash log message

freewilll:
Example of a faulty log message:
Rate limiting is enabled with penalty 18203625after deleting file ...


David132639 (approved):
Looks good 👍

PR #7287, opened 2020-08-19
Resume writes early when snapshot consistency is not required

huachaohuang:
The current implementation blocks writes during LogAndApply(), which can cause
very high latency up to seconds. This commit resume writes early when snapshot
consistency is not required.

I'm not sure if it is a good idea to reuse the `snapshot_consistency` option, but it seems reasonable.


db/db_impl/db_impl.cc
```diff
@@ -4417,6 +4419,19 @@ Status DBImpl::IngestExternalFiles(
         versions_->SetLastPublishedSequence(last_seqno + consumed_seqno_count);
         versions_->SetLastSequence(last_seqno + consumed_seqno_count);
       }
+    }
+
+    // The following LogAndApply() can cause very high latency up to seconds. If
```
siying, line 4424: Can you explain why LogAndApply() can cause high latency? Are we doing more than fsync()?

ajkr: Also note - the `FlushMemTable()` on line 4385 blocks writes while it runs, and we often see a multi-second latency come from there.

huachaohuang: @siying The "manifest.file.sync.micros" from RocksDB shows that it can take one or two seconds sometimes.

This is how the statistics look like across hundreds of instances.
屏幕快照 2020-09-09 10 12 30

huachaohuang: @ajkr Yes, that's a problem too. But I have figured out a way to work around that in our scenario. https://github.com/tikv/rust-rocksdb/blob/master/librocksdb_sys/crocksdb/c.cc#L4241

siying: @huachaohuang I trust what you see. What I didn't understand is why it takes so long. Is it that one fsync() takes more than one second, or we are doing more than one for this bulkloading path? If it is the latter case, maybe we should fix the root cause.

huachaohuang: @siying I tried to add more statistics around `SyncManifest()` and it did take quite long sometimes. And I think "manifest.file.sync.micros" also showed similar results. Maybe a better disk can reduce the latency, but that's not up to me :(

siying: @huachaohuang is it reproducible? If it is reproducible, maybe strace fsync and match the bulkload calls to fsync calls issued for the command. We will discuss internally about this feature request in the mean time.

PR #7290, opened 2020-08-20
[RFC] Allow block cache to do memcpy for insertion

riversand963:
Summary:
In PutDataBlockToCache, RocksDB tries to insert a Block or BlockContents
object to (uncompressed) block cache. After insertion succeeds, RocksDB
currently assumes that the original object is added to the cache without
memcpy. This is true now. However, Cache interface does not explicitly forbid
Cache doing memcpy internally. This is sometimes necessary if the underlying
Cache implementation manages multiple heterogeneouos memories/storage
different from the memory in which the original Block/BlockContents'
payload reside. In this case, the address of the new buffer after memcpy
will be returned. The new buffer will be managed by the Cache
internally, and RocksDB block based table does not own it.

Test Plan (devserver):
make check


table/block_based/block_based_table_reader.cc
```diff
@@ -1319,8 +1319,22 @@ Status BlockBasedTable::PutDataBlockToCache(
...
+      cached_block->SetCachedValue(reinterpret_cast<TBlocklike*>(ptr_in_cache),
+                                   block_cache, cache_handle);
+      // If ptr_in_cache is equal to block_holder.get(), it means the block has
+      // been inserted into the cache without memcpy. Therefore, the block
+      // cache starts to take the responsibility of managing the underlying
+      // memory, and block_holder needs to release ownership.
+      // If ptr_in_cache is not equal to block_holder.get(), it means
+      // block_cache copies the block to some internally managed buffer. When
+      // block_holder goes out of scope, its underlying memory should be
+      // free'ed, while cached_block will point to the new buffer.
```
ltamasi, line 1332: Thanks for the patch @riversand963 ! I have some questions/points to discuss:

1) I'm wondering if we could avoid copying altogether by allocating `BlockContents` and friends using the appropriate allocator? This could probably be done by generalizing `CacheAllocationPtr` (which is already used for the actual raw bytes of the block).
2) If we *do* need to copy, one thing to watch out for is that `BlockContents` / `Block` are not trivially copyable, so `memcpy` is not really appropriate for them. Instead, what we would probably want is to use placement `new` to move construct a new `BlockContents` in the "special" memory pool so that the ownership of `allocation` is correctly transferred.
3) As for the API, we could also consider introducing a dedicated `Insert` variant that is allowed to copy. (I'm afraid there might be code out there that relies on the current behavior, so changing the convention around the existing `Insert` might be risky.)

riversand963: > Thanks for the patch @riversand963 ! I have some questions/points to discuss:
> 
> 1. I'm wondering if we could avoid copying altogether by allocating `BlockContents` and friends using the appropriate allocator? This could probably be done by generalizing `CacheAllocationPtr` (which is already used for the actual raw bytes of the block).

This may be hard in some cases due to implementation complexity and disruption to existing code. Take persistent memory for an example. Existing `MemoryAllocator` interface does not work well with persistent memory when we try to utilize persistence and recoverability. There is currently no consensus on what the interface should be like, and @roghnin knows better than I do. Furthermore, persistent memory programming libraries, e.g. Intel's libpmemobj sometimes use their own pointer types instead of `void*`. We cannot insert such a pointer to a `Cache`. The idea of `CacheAllocationPtr` is reasonable, but it means different Cache/Allocator combinations may have different `CacheAllocationPtr`.

> 2. If we _do_ need to copy, one thing to watch out for is that `BlockContents` / `Block` are not trivially copyable, so `memcpy` is not really appropriate for them. Instead, what we would probably want is to use placement `new` to move construct a new `BlockContents` in the "special" memory pool so that the ownership of `allocation` is correctly transferred.

Yeah, when we do copy, we will be aware of the type, and will for sure do a deep copy with memory allocation in the separate memory. Move can be hard, due to the reasons mentioned above.

> 3. As for the API, we could also consider introducing a dedicated `Insert` variant that is allowed to copy. (I'm afraid there might be code out there that relies on the current behavior, so changing the convention around the existing `Insert` might be risky.)

Thanks @ltamasi for the feedback, I have tried to address your concerns in the quoted text above.

ltamasi: Thanks for the clarification @riversand963 , makes sense. We might want to remove the references to `memcpy` from the comment and the PR title to avoid any confusion.

One more question: are we primarily looking at using this for the uncompressed block cache? There's another place that inserts into that in `GetDataBlockFromCache` (when the block is not in the uncompressed cache but *is* in the compressed cache).

riversand963: Good catch @ltamasi and thanks for the feedback. We need to look at other call sites of `Insert` into both compressed and uncompressed cache. This PR is a start and RFC.

PR #7296, opened 2020-08-21
Bug Fix for memtables not trimmed down.

akankshamahajan15:
Summary: When a memtable is trimmed in MemTableListVersion, the memtable
is only added to delete list if it is
the last reference. However it is not the last reference as it is held
by the super version. But the super version would not be switched if the
delete list is empty. So the memtable is never destroyed and memory
usage increases beyond write_buffer_size +
max_write_buffer_size_to_maintain.

Test Plan: 1.  ./db_bench -benchmarks=randomtransaction
-optimistic_transaction_db=1 -statistics -stats_interval_seconds=1
-duration=90 -num=500000 --max_write_buffer_size_to_maintain=16000000
--transaction_set_snapshot

Reviewers:

Subscribers:

Tasks:

Tags:


akankshamahajan15:
Planning to add test unit as well in separate PR

akankshamahajan15:
> Do you plan to add a unit test?

Yes, I am planning to add it in a separate PR if that is fine or I can hold this PR for the test unit as well

PR #7300, opened 2020-08-21
Fix for Regression failure

akankshamahajan15:
Summary: RocksDb regression commands are exiting with error
/usr/bin/ar: creating
librocksdb.a
/usr/bin/ld: ./cache/cache.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC

Bug: It tries to link the static code into a shared lib.
Fix: Added make clean before building shared_lib

Test Plan: make clean
           make -j$(nproc) static_lib
           make -j$(nproc) shared_lib

Reviewers:

Subscribers:

Tasks:

Tags:


pdillinger:
I think set -e is a good idea, but I'm happy to make the change when any breakage would happen on my watch. 😉

PR #7304, opened 2020-08-24
Make examples work on Windows

alucarded:
Quick fixes to examples to make it easier to get familiar with RocksDB for Windows users:

- Set proper temporary directory path on Windows for all examples (with C++17 we should start using std::filesystem)
- Fixed typo and got rid of warnings treated as errors in multi_processes_example.cc
- Get number of available cores on Windows in c_simple_example.c
- Add command to remove DB directory for Windows in compaction_filter_example.cc (print error, but carry on with example upon error, because error code is returned if there is no such directory on Windows)

### DIFF
--- a/examples/c_simple_example.c
+++ b/examples/c_simple_example.c
@@ -10,18 +10,35 @@
 
 #include "rocksdb/c.h"
 
+#if defined(OS_WIN)
+#include <Windows.h>
+#else
 #include <unistd.h>  // sysconf() - get CPU count
+#endif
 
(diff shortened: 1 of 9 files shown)


alucarded:
I have just started working with RocksDB on Windows and figured out that all the examples are not working, so I fixed them.
Can anyone review and merge if everything looks good? It may help others.

@ajkr @pdillinger @akankshamahajan15 @siying I am mentioning you since you are the most active developers/reviewers recently. Please help with getting this merged! Thank you!

PR #7306, opened 2020-08-24
Add a blob file builder class that can be used in background jobs

ltamasi:
Summary:
The patch adds a class called `BlobFileBuilder` that can be used to build
and cut blob files in background jobs (flushes/compactions). The class
enforces a value size threshold (`min_blob_size`; smaller blobs will be inlined
in the LSM tree itself), and supports specifying a blob file size limit (`blob_file_size`),
as well as compression (`blob_compression_type`) and checksums for blob files.
It also keeps track of the generated blob files and their associated `BlobFileAddition`
metadata, which can be applied as part of the background job's `VersionEdit`.

Test Plan:
`make check`


db/blob/blob_file_builder_test.cc
```diff
@@ -0,0 +1,574 @@
...
+ public:
+  uint64_t operator()() { return ++next_file_number_; }
+
+ private:
+  uint64_t next_file_number_ = 1;
+};
+
+class BlobFileBuilderTest : public testing::Test {
+ protected:
+  BlobFileBuilderTest() : env_(Env::Default()), fs_(&env_) {}
```
riversand963, line 43: Nit: maybe we can use `Env::LoadEnv` and a env var to create the right type of `Env`, similar to `db_basic_test.cc`. This will allow us to test blob db on other FS.

db/blob/blob_file_builder_test.cc
```diff
@@ -0,0 +1,574 @@
...
+  ASSERT_TRUE(builder.Add(key, value, &blob_index).IsCorruption());
+
+  SyncPoint::GetInstance()->DisableProcessing();
+  SyncPoint::GetInstance()->ClearAllCallBacks();
+}
+
+TEST_F(BlobFileBuilderTest, Checksum) {
+  // Build a blob file with checksum
+
+  class DummyFileChecksumGenerator : public FileChecksumGenerator {
```
riversand963, line 423: Any particular reason we define this class only in this function? Will it ever be re-used?

db/blob/blob_file_builder.cc
```diff
@@ -0,0 +1,291 @@
...
+  CompressionOptions opts;
+  CompressionContext context(blob_compression_type_);
+  constexpr uint64_t sample_for_compression = 0;
+
+  CompressionInfo info(opts, context, CompressionDict::GetEmptyDict(),
+                       blob_compression_type_, sample_for_compression);
+
+  constexpr uint32_t compression_format_version = 2;
+
+  if (!CompressData(*blob, info, compression_format_version, compressed_blob)) {
```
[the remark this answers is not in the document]
ltamasi: Yes, I can definitely see us implementing sampling etc. (as in `BlockBasedTable`) in a later phase.

db/blob/blob_file_builder_test.cc
ltamasi: Hm, `env_` is really a `MockEnv` here so no real file operations should be involved... I'm wondering what (if anything) `MockEnv` uses its target for though?

db/blob/blob_file_builder_test.cc
ltamasi: I don't expect to add more checksum related test cases at this point but we can always move it if needed.

db/blob/blob_file_builder_test.cc
ltamasi: Will rename the member variable to `mock_env_` to make it clear that it's not a real `Env`.

PR #7310, opened 2020-08-25
Map retryable IO error during Flush without WAL to soft error and no switch memtable during resume

zhichao-cao:
In the current implementation, any retryable IO error happens during Flush is mapped to a hard error. In this case, DB is stopped and write is stalled unless the background error is cleaned. In this PR, if WAL is DISABLED, the retryable IO error during FLush is mapped to a soft error. Such that, the memtable can continue receive the writes. At the same time, if auto resume is triggered, SwtichMemtable will not be called during Flush when resuming the DB to avoid to many small memtables. Testing cases are added.

Test plan: adding new unit test, pass make check.

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -247,7 +247,6 @@ Status DBImpl::FlushMemTableToOutputFile(
         error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush);
       } else {
         // If the WAL is empty, we use different error reason
-        printf("regular flush\n");
         error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlushNoWAL);
       }
     } else {
@@ -1651,6 +1650,7 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
       return s;
     }
   }
+
   FlushRequest flush_req;
   {
     WriteContext context;
@@ -1666,8 +1666,8 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
     }
     WaitForPendingWrites();
 
-    if ((!cfd->mem()->IsEmpty() || !cached_recoverable_state_empty_.load())
-          && flush_reason != FlushReason::kFlushNoSwitchMemtable) {
+    if ((!cfd->mem()->IsEmpty() || !cached_recoverable_state_empty_.load()) &&
+        flush_reason != FlushReason::kFlushNoSwitchMemtable) {
       s = SwitchMemtable(cfd, &context);
     }
     if (s.ok()) {
@@ -1676,8 +1676,8 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
         flush_memtable_id = cfd->imm()->GetLatestMemTableID();
         flush_req.emplace_back(cfd, flush_memtable_id);
       }
-      if (immutable_db_options_.persist_stats_to_disk
-          && flush_reason != FlushReason::kFlushNoSwitchMemtable) {
+      if (immutable_db_options_.persist_stats_to_disk &&
+          flush_reason != FlushReason::kFlushNoSwitchMemtable) {
         ColumnFamilyData* cfd_stats =
             versions_->GetColumnFamilySet()->GetColumnFamily(
                 kPersistentStatsColumnFamilyName);
@@ -1706,7 +1706,6 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
         }
       }
     }
-
     if (s.ok() && !flush_req.empty()) {
       for (auto& elem : flush_req) {
         ColumnFamilyData* loop_cfd = elem.first;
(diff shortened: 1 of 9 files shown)


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -217,10 +222,14 @@ Status DBImpl::FlushMemTableToOutputFile(
...
-                            ? BackgroundErrorReason::kFlush
-                            : BackgroundErrorReason::kManifestWrite;
-      error_handler_.SetBGError(io_s, err_reason);
+      if (!versions_->io_status().ok()) {
+        error_handler_.SetBGError(io_s, BackgroundErrorReason::kManifestWrite);
+      } else if (total_log_size_ > 0) {
+        error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush);
+      } else {
+        // If the WAL is empty, we use different error reason
+        error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlushNoWAL);
```
anand1976, line 231: To be on the safe side, can we also check if ```SyncClosedLogs``` failed on line 184, and if so, set reason to ```kFlush```?

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1613,7 +1627,8 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
     }
     WaitForPendingWrites();
 
-    if (!cfd->mem()->IsEmpty() || !cached_recoverable_state_empty_.load()) {
+    if ((!cfd->mem()->IsEmpty() || !cached_recoverable_state_empty_.load()) &&
+        flush_reason != FlushReason::kFlushNoSwitchMemtable) {
```
anand1976, line 1631: Should we assert here that ```cfd->imm()->NumNotFlushed() > 0``` here if ```flush_reason == FlushReason::kFlushNoSwitchMemtable```?

include/rocksdb/listener.h
```diff
@@ -115,6 +115,7 @@ enum class FlushReason : int {
   kAutoCompaction = 0x09,
   kManualFlush = 0x0a,
   kErrorRecovery = 0xb,
+  kFlushNoSwitchMemtable = 0xc,
```
anand1976, line 118: The name might be a bit confusing as this is also during error recovery. Maybe we can call it something like ```kErrorRecoveryRetryFlush```?

db/error_handler.h
```diff
@@ -59,9 +71,11 @@ class ErrorHandler {
     bool IsBGWorkStopped() {
       return !bg_error_.ok() &&
              (bg_error_.severity() >= Status::Severity::kHardError ||
-              !auto_recovery_);
+              !auto_recovery_ || soft_error_no_bg_work_);
     }
 
+    bool IsSoftErrorNoBGWork() { return soft_error_no_bg_work_; }
```
anand1976, line 77: IIUC, this is only to prevent a manual flush thread from waiting indefinitely while recovery is going on. Can we just use ```IsBGWorkStopped```?

db/db_impl/db_impl_compaction_flush.cc
zhichao-cao: Sure. Yeah, that part is missed.

zhichao-cao: > To be on the safe side, can we also check if `SyncClosedLogs` failed on line 184, and if so, set reason to `kFlush`?

Sure. Yeah, that part is missed.

db/db_impl/db_impl_compaction_flush.cc
zhichao-cao: > Should we assert here that `cfd->imm()->NumNotFlushed() > 0` here if `flush_reason == FlushReason::kFlushNoSwitchMemtable`?

If there is no new KV-pair being inserted since last switch. Do we still generate an empty immutable Memtable? If so, it is better to assert `cfd->imm()->NumNotFlushed() > 0` here if `flush_reason == FlushReason::kFlushNoSwitchMemtable`

db/error_handler.h
zhichao-cao: Oh, yes. once `!bg_error_.ok() && soft_error_no_bg_work_`, `IsBGWorkStopped` will return true. It is the same as just check `soft_error_no_bg_work_` since `bg_error_` must be **non_ok**.

include/rocksdb/listener.h
zhichao-cao: Sure, but we may be better to indicate that SwitchMemtable will not be called in this case at some place. Where do you think is appropriate to write the comment?

db/db_impl/db_impl_compaction_flush.cc
anand1976: I don't think an empty immutable memtable is generated. The assert I'm suggesting is to just make sure there is something to do, i.e if ```SwitchMemtable``` is not called, then atleast there should be 1 unflushed immutable memtable. That should usually be the case since its a flush error that would have caused the ```kFlushNoSwitchMemtable``` reason.

include/rocksdb/listener.h
anand1976: The comment can be here.

db/error_handler.cc
```diff
@@ -389,16 +442,25 @@ Status ErrorHandler::ClearBGError() {
...
+    // the bg work should follow this tag.
+    soft_error_no_bg_work_ = false;
+
+    // In manual resume, the flush reason is kErrorRecovery.
+    recover_context_.flush_reason = FlushReason::kErrorRecovery;
   }
 
-  if (bg_error_.severity() == Status::Severity::kSoftError) {
+  if (bg_error_.severity() == Status::Severity::kSoftError &&
+      no_bg_work_original_flag == false) {
```
anand1976, line 463: Can ```recover_context_.flush_reason``` be used here? It feels a little non-intuitive to rely on ```no_bg_work_original_flag```, since that's mainly to determine how other BG work should behave.

anand1976 (approved):
LGTM. I've added a few minor comments, but nothing that's a showstopper.

PR #7311, opened 2020-08-25
Add unit test for max_write_buffer_size_to_maintain

akankshamahajan15:
Summary: Add a unit test case to check memory usage when
max_write_buffer_size_to_maintain is set if flushed immutable memtables are
trimmed timely or not.

Test Plan: Compared the results with before bug fix.

Reviewers:

Subscribers:

Tasks: T72035179

Tags:


db/db_test.cc
```diff
@@ -6677,6 +6677,49 @@ TEST_F(DBTest, CreationTimeOfOldestFile) {
...
+  options.max_write_buffer_size_to_maintain = 1000000;
+  options.write_buffer_size = 16000000;
+  Reopen(options);
+  Random rnd(301);
+  bool memory_limit_increasing = false;
+  uint64_t size_all_mem_table = 0;
+  uint64_t cur_active_mem = 0;
+
+  for (int i = 0; i < 5000; i++) {
+    std::string value = rnd.RandomString(100000);
```
ltamasi, line 6691: I'm wondering if we could scale this value size down (as well as `max_write_buffer_size_to_maintain` and `write_buffer_size`) by say, a factor of 1000 to make the test more efficient?

db/db_test.cc
```diff
@@ -6677,6 +6677,49 @@ TEST_F(DBTest, CreationTimeOfOldestFile) {
...
+    // memory_limit_increasing is set to false.
+    if ((cur_active_mem >=
+         static_cast<uint64_t>(options.max_write_buffer_size_to_maintain)) &&
+        (size_all_mem_table > options.max_write_buffer_size_to_maintain +
+                                  options.write_buffer_size)) {
+      if (memory_limit_increasing) {
+        ASSERT_TRUE(false);
+      } else {
+        memory_limit_increasing = true;
+      }
```
ltamasi, line 6716: This could be simplified:

```
ASSERT_FALSE(memory_limit_exceeded);
memory_limit_exceeded = true;
```

db/db_test.cc
```diff
@@ -6677,6 +6677,49 @@ TEST_F(DBTest, CreationTimeOfOldestFile) {
   ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->DisableProcessing();
 }
 
+TEST_F(DBTest, MemoryUsageWithMaxWriteBufferSizeToMaintain) {
+  Options options = CurrentOptions();
+  options.max_write_buffer_size_to_maintain = 1000000;
+  options.write_buffer_size = 16000000;
+  Reopen(options);
+  Random rnd(301);
+  bool memory_limit_increasing = false;
```
ltamasi, line 6686: Maybe `memory_limit_exceeded` would be a better name for this one?

db/db_test.cc
akankshamahajan15: Scaled ```max_write_buffer_size_to_maintain```, ```write_buffer_size``` down by a factor of 100 (wasn't able to reproduce if reduce by 1000) and reduced the  total iteration to 1000.

PR #7314, opened 2020-08-26
Expose rocksdb_open_column_families_with_ttl C function

sezaru:
This PR creates `rocksdb_open_column_families_with_ttl` which allows C API users to open a DBWithTLL with column families.


sezaru:
I forgot that we need to use CreateColumnFamilyWithTtl instead of CreateColumnFamily function, so I added new code to expose this too via `rocksdb_create_column_family_with_ttl` function.

sezaru:
Hmm, seems like that windows build failed because of this error:
```
0curl: (7) Failed to connect to download.visualstudio.microsoft.com port 443: Timed out
Download of the VS 2017 installer failed
```

How can I run this build again without doing a new commit?

PR #7315, opened 2020-08-26
Add sst_file_dumper status check

jay-zhuang:
test plan:
`ASSERT_STATUS_CHECKED=1 make sst_dump_test && ./sst_dump_test`
And manually run `./sst_dump --file=*.sst` before and after the change.

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -3179,19 +3179,44 @@ Status BlockBasedTable::GetKVPairsFromDataBlocks(
   return Status::OK();
 }
 
+// It's an output stream helper to build string to WritableFile.
+// Which auto flushes to file when the object is destroyed (like RAII), the
+// flush error is ignored, unless manually call `flush()`.
+class WritableFileStringStreamHelper {
+ public:
+  WritableFileStringStreamHelper(WritableFile* writable_file)
+      : file_(writable_file) {}
+
+  ~WritableFileStringStreamHelper() { assert(flush()); }
+
+  bool flush() {
+    Status s = file_->Append(stream_buff_.str());
+    stream_buff_.clear();
+    return s.ok();
+  }
+
+  // Simple template just forward the input to stringstream buffer
+  template <typename T>
+  WritableFileStringStreamHelper& operator<<(const T& t) {
+    stream_buff_ << t;
+    return *this;
+  }
+
+ private:
+  WritableFile* file_;
+  std::ostringstream stream_buff_;
+};
+
 Status BlockBasedTable::DumpTable(WritableFile* out_file) {
+  WritableFileStringStreamHelper out_stream(out_file);
   // Output Footer
-  out_file->Append(
-      "Footer Details:\n"
-      "--------------------------------------\n"
-      "  ");
-  out_file->Append(rep_->footer.ToString().c_str());
-  out_file->Append("\n");
+  out_stream << "Footer Details:\n"
+                "--------------------------------------\n";
+  out_stream << "  " << rep_->footer.ToString() << "\n";
 
   // Output MetaIndex
-  out_file->Append(
-      "Metaindex Details:\n"
-      "--------------------------------------\n");
+  out_stream << "Metaindex Details:\n"
+                "--------------------------------------\n";
   std::unique_ptr<Block> metaindex;
   std::unique_ptr<InternalIterator> metaindex_iter;
   ReadOptions ro;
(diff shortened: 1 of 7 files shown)


table/block_based/block_based_table_reader.cc
```diff
@@ -3178,19 +3178,44 @@ Status BlockBasedTable::GetKVPairsFromDataBlocks(
...
+
+  bool flush() {
+    Status s = file_->Append(stream_buff_.str());
+    stream_buff_.clear();
+    return s.ok();
+  }
+
+  // Simple template just forward the input to stringstream buffer
+  template <typename T>
+  WritableFileStringStreamHelper& operator<<(const T& t) {
```
pdillinger, line 3199: I don't see any manual or automatic flushes. So doesn't this build an in-memory textual dump of the whole file before writing it out? That's not practical/safe for very large SST files.

jay-zhuang: Good point. I missed that. Added a flush threshold and a simple test for that.

table/block_based/block_based_table_reader.cc
```diff
@@ -3179,19 +3179,49 @@ Status BlockBasedTable::GetKVPairsFromDataBlocks(
...
+    stream_buff_ << t;
+    if (stream_buff_.tellp() > kBuffFlushThreshold) {
+      flush();
```
pdillinger, line 3204: maybe we should assert here also? (If we really need our own buffering ¯\\_(ツ)_/¯)

I don't think we currently have a good way of always evaluating something and asserting (debug only) on the result while avoiding unused variable error on release build.

table/sst_file_dumper.cc
```diff
@@ -178,7 +178,10 @@ Status SstFileDumper::DumpTable(const std::string& out_filename) {
   Env* env = options_.env;
   env->NewWritableFile(out_filename, &out_file, soptions_);
   Status s = table_reader_->DumpTable(out_file.get());
-  out_file->Close();
+  if (!s.ok()) {
+    return s;
```
pdillinger, line 182: Is it OK to return without closing the WritableFile? (I wish the API docs were any help 😡😡)

table/sst_file_dumper.cc
```diff
@@ -187,7 +190,11 @@ uint64_t SstFileDumper::CalculateCompressedTableSize(
     uint64_t* num_data_blocks) {
   std::unique_ptr<WritableFile> out_file;
   std::unique_ptr<Env> env(NewMemEnv(options_.env));
-  env->NewWritableFile(testFileName, &out_file, soptions_);
+  Status s = env->NewWritableFile(testFileName, &out_file, soptions_);
+  if (!s.ok()) {
+    fputs(s.ToString().c_str(), stderr);
+    exit(1);
```
pdillinger, line 196: I see this function already has some calls to exit, but this makes me nervous as we recently added a call from library code to SstFileDumper (in backupable_db.cc). Can we go ahead and refactor this to return status? 🙏 🙏

table/block_based/block_based_table_reader.cc
jay-zhuang: Good point. I removed the buffering. Now we have an adapter class to make `WritableFile` working with `ostream`.

table/block_based/block_based_table_reader.cc
```diff
@@ -3179,19 +3179,49 @@ Status BlockBasedTable::GetKVPairsFromDataBlocks(
...
 
+// It's an output stream helper to build string to WritableFile.
+// Which auto flushes to file when the object is destroyed (like RAII) or the
+// data size is bigger than `kBuffFlushThreshold`, the flush error is ignored.
+class WritableFileStringStreamHelper {
+ public:
+  WritableFileStringStreamHelper(WritableFile* writable_file)
+      : file_(writable_file) {}
+
+  ~WritableFileStringStreamHelper() { assert(flush()); }
```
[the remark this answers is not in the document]
jay-zhuang: Removed.

table/block_based/block_based_table_reader.cc
jay-zhuang: Removed.

table/sst_file_dumper.cc
jay-zhuang: Good point, I think we should try to close whether there's error or not.

table/sst_file_dumper.cc
jay-zhuang: Updated.

table/sst_file_dumper.cc
```diff
@@ -206,23 +213,25 @@ uint64_t SstFileDumper::CalculateCompressedTableSize(
...
+  *compressed_table_size = table_builder->FileSize();
   assert(num_data_blocks != nullptr);
   *num_data_blocks = table_builder->GetTableProperties().num_data_blocks;
-  env->DeleteFile(testFileName);
-  return size;
+  s = env->DeleteFile(testFileName);
+  if (!s.ok()) {
+    return s;
+  }
+  return Status::OK();
```
pdillinger, line 231: The shorter version (which I prefer but others might not) is `return env->DeleteFile(...);`  😄  That should be OK by the status check.

pdillinger (approved):
Thanks for adding Status to the recompression stuff. 😄

And I didn't see an explicit reply to this:

> Have you done a manual before vs. after diff on a dump file? That might be useful. :)

jay-zhuang:
> Thanks for adding Status to the recompression stuff. 😄
> 
> And I didn't see an explicit reply to this:
> 
> > Have you done a manual before vs. after diff on a dump file? That might be useful. :)

Yes, the output files are exactly same.

I did find a mismatch for `double`:
```
double d1 = 123;
std::cout << d1 << std::endl; // print 123
std::cout << ToString(d1) << std::endl; // print 123.000000
```
To keep it consistent, it's switched back to `ToString()`.

PR #7324, opened 2020-08-28
Log info about generated blob files in BlobFileBuilder

ltamasi:
Summary:
The patch adds a log message to `BlobFileBuilder` that is logged upon
generating a blob file, similarly to how we log the generation of table files
during flush and compaction. The log message contains the column family
name, job id, blob file number, and the number and total size of blobs in
the new file.

Test Plan:
Ran `make check` and checked the actual log messages using a custom `db_bench`.

### DIFF
--- a/db/blob/blob_file_builder.cc
+++ b/db/blob/blob_file_builder.cc
@@ -28,21 +28,23 @@ BlobFileBuilder::BlobFileBuilder(
     VersionSet* versions, Env* env, FileSystem* fs,
     const ImmutableCFOptions* immutable_cf_options,
     const MutableCFOptions* mutable_cf_options, const FileOptions* file_options,
-    int job_id, uint32_t column_family_id, std::string column_family_name,
-    Env::IOPriority io_priority, Env::WriteLifeTimeHint write_hint,
+    int job_id, uint32_t column_family_id,
+    const std::string& column_family_name, Env::IOPriority io_priority,
+    Env::WriteLifeTimeHint write_hint,
     std::vector<BlobFileAddition>* blob_file_additions)
     : BlobFileBuilder([versions]() { return versions->NewFileNumber(); }, env,
                       fs, immutable_cf_options, mutable_cf_options,
                       file_options, job_id, column_family_id,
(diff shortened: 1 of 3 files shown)


db/blob/blob_file_builder.h
```diff
@@ -70,7 +72,9 @@ class BlobFileBuilder {
   uint64_t blob_file_size_;
   CompressionType blob_compression_type_;
   const FileOptions* file_options_;
+  int job_id_;
   uint32_t column_family_id_;
+  std::string column_family_name_;
```
riversand963, line 77: Nit: we probably can use `const std::string&` here because the column family should outlive this builder.

db/blob/blob_file_builder.cc
```diff
@@ -268,6 +272,13 @@ Status BlobFileBuilder::CloseBlobFile() {
                                      std::move(checksum_method),
                                      std::move(checksum_value));
 
+  assert(immutable_cf_options_);
+  ROCKS_LOG_INFO(immutable_cf_options_->info_log,
```
riversand963, line 276: Do you think it will be better and clearer to call out the fact in the LOG that the blob file is not synced at this point?

db/blob/blob_file_builder.cc
```diff
@@ -27,20 +28,21 @@ BlobFileBuilder::BlobFileBuilder(
     VersionSet* versions, Env* env, FileSystem* fs,
     const ImmutableCFOptions* immutable_cf_options,
     const MutableCFOptions* mutable_cf_options, const FileOptions* file_options,
-    uint32_t column_family_id, Env::IOPriority io_priority,
-    Env::WriteLifeTimeHint write_hint,
+    int job_id, uint32_t column_family_id, std::string column_family_name,
```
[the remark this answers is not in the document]
ltamasi: Yeah, other than tests, I don't think there are cases where we would pass an rvalue. Let me change this.

db/blob/blob_file_builder.cc
ltamasi: It actually is synced by the time we get to this point (as a result of the `AppendFooter` call above).

PR #7331, opened 2020-08-31
Recompress blobs during GC if compression changed

jay-zhuang:
Summary: Recompress blobs if compression type is changed.

Test Plan: `make check`

### DIFF
--- a/utilities/blob_db/blob_compaction_filter.cc
+++ b/utilities/blob_db/blob_compaction_filter.cc
@@ -373,11 +373,14 @@ CompactionFilter::BlobDecision BlobIndexCompactionFilterGC::PrepareBlobOutput(
       const Status status =
           blob_db_impl->DecompressSlice(blob, compression_type, &blob);
       if (!status.ok()) {
+        gc_stats_.SetError();
         return BlobDecision::kCorruption;
       }
     }
     if (blob_db_impl->bdb_options_.compression != kNoCompression) {
-      blob.PinSelf(blob_db_impl->GetCompressedSlice(blob, &compression_output));
+      blob_db_impl->GetCompressedSlice(blob, &compression_output);
+      blob = PinnableSlice(&compression_output);
+      blob.PinSelf();
     }
   }
 
--- a/utilities/blob_db/blob_compaction_filter.cc
+++ b/utilities/blob_db/blob_compaction_filter.cc
@@ -360,9 +360,25 @@ CompactionFilter::BlobDecision BlobIndexCompactionFilterGC::PrepareBlobOutput(
 
   PinnableSlice blob;
   CompressionType compression_type = kNoCompression;
+  std::string compression_output;
   if (!ReadBlobFromOldFile(key, blob_index, &blob, false, &compression_type)) {
     gc_stats_.SetError();
     return BlobDecision::kIOError;
+  }
+
+  // If the compression_type is changed, re-compress it with the new compression
+  // type.
+  if (compression_type != blob_db_impl->bdb_options_.compression) {
+    if (compression_type != kNoCompression) {
+      const Status status =
+          blob_db_impl->DecompressSlice(blob, compression_type, &blob);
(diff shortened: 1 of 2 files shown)


utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -360,11 +360,27 @@ CompactionFilter::BlobDecision BlobIndexCompactionFilterGC::PrepareBlobOutput(
...
   }
 
+  // If the compression_type is changed, re-compress it with the new compression
+  // type.
+  if (compression_type != blob_db_impl->bdb_options_.compression) {
+    if (compression_type != kNoCompression) {
+      const Status status =
+          blob_db_impl->DecompressSlice(blob, compression_type, &blob);
+      if (!status.ok()) {
+        return BlobDecision::kCorruption;
```
ltamasi, line 376: We should call `gc_stats_.SetError()` here, like in the other failure cases (note: this is not needed if we make the change below).

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -360,11 +360,27 @@ CompactionFilter::BlobDecision BlobIndexCompactionFilterGC::PrepareBlobOutput(
...
+      }
+    }
```
ltamasi, line 378: `ReadBlobFromOldFile` already supports decompression; we could simply introduce a boolean `const bool compression_changed = (compression_type != blob_db_impl->bdb_options_.compression;` and pass it as `need_decompress` to avoid having to explicitly call `DecompressSlice` here.

utilities/blob_db/blob_compaction_filter.cc
```diff
@@ -360,11 +360,27 @@ CompactionFilter::BlobDecision BlobIndexCompactionFilterGC::PrepareBlobOutput(
...
+    if (blob_db_impl->bdb_options_.compression != kNoCompression) {
+      blob.PinSelf(blob_db_impl->GetCompressedSlice(blob, &compression_output));
```
ltamasi, line 380: `PinSelf` here copies the compression results into `blob`. I'm wondering if we could eliminate that? E.g. something like `blob = PinnableSlice(&compression_output);` after the `GetCompressedSlice` call should work.

utilities/blob_db/blob_db_test.cc
```diff
@@ -562,7 +562,160 @@ TEST_F(BlobDBTest, DecompressAfterReopen) {
...
+  bdb_options.compression = kSnappyCompression;
+  Reopen(bdb_options);
+  for (; data_idx < 400; data_idx++) {
+    PutRandom("put-key" + ToString(data_idx), &rnd, &data);
+  }
+  VerifyDB(data);
+
+  bdb_options.compression = kLZ4Compression;
+  Reopen(bdb_options);
+  for (; data_idx < 400; data_idx++) {
```
ltamasi, line 703: This was probably meant to be `500`.

utilities/blob_db/blob_compaction_filter.cc
ltamasi: Just checking: are you planning to rely on the `ReadBlobFromOldFile` call above to do the decompression?

utilities/blob_db/blob_compaction_filter.cc
jay-zhuang: Thanks @ltamasi for the review.
For this issue, the problem is the compression_type is also returning from `ReadBlobFromOldFile()`. (I guess it's combining the getting the compression_type and reading data in one call to reduce IO or mutex lock).
One option is to pass a `new_compression_type` into `ReadBlobFromOldFile()`, then only decompress if it's different from the old one. But that feels not a clean function.

ltamasi: Right, we have a chicken and egg problem because `compression_type` is set by `ReadBlobFromOldFile` :)

PR #7333, opened 2020-09-01
Add IO Tracer Parser

akankshamahajan15:
Summary: Implement a parsing tool io_tracer_parser that takes IO trace file (binary file) with command line argument --io_trace_file and output file with --output_file and dumps the IO trace records in outputfile in human readable form.

Also added unit test cases that generates IO trace records and calls io_tracer_parse to parse those records.

Test Plan: make check -j64,
                  Add unit test cases.

Reviewers:

Subscribers:

Tasks:

Tags:
Screen Shot 2020-09-10 at 11 14 21 AM

### DIFF
--- a/tools/io_tracer_parser_tool.cc
+++ b/tools/io_tracer_parser_tool.cc
@@ -127,12 +127,12 @@ int io_tracer_parser(int argc, char** argv) {
 
   if (FLAGS_io_trace_file.empty()) {
     fprintf(stderr, "IO Trace file path is empty\n");
-    exit(1);
+    return 1;
   }
 
   if (FLAGS_output_file.empty()) {
     fprintf(stderr, "Output file path is empty\n");
-    exit(1);
+    return 1;
   }
 
   IOTraceRecordParser io_tracer_parser(FLAGS_io_trace_file, FLAGS_output_file);
--- a/tools/io_tracer_parser_tool.cc
+++ b/tools/io_tracer_parser_tool.cc
@@ -1,0 +1,144 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//    This source code is licensed under both the GPLv2 (found in the
+//    COPYING file in the root directory) and Apache 2.0 License
+//    (found in the LICENSE.Apache file in the root directory).
+
+#ifndef ROCKSDB_LITE
+#ifdef GFLAGS
+#include "tools/io_tracer_parser_tool.h"
+
+#include <cinttypes>
+#include <cstdio>
+#include <iomanip>
+#include <memory>
+#include <sstream>
+
+#include "port/lang.h"
+#include "trace_replay/io_tracer.h"
+#include "util/gflags_compat.h"
+
+using GFLAGS_NAMESPACE::ParseCommandLineFlags;
+
+DEFINE_string(io_trace_file, "", "The IO trace file path.");
+DEFINE_string(
+    output_file, "",
+    "The output file path for storing human readable format of IO trace "
+    "records that are stored in IO trace file.");
+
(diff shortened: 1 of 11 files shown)


tools/io_tracer_parser_tool.cc
```diff
@@ -0,0 +1,144 @@
...
+#include "trace_replay/io_tracer.h"
+#include "util/gflags_compat.h"
+
+using GFLAGS_NAMESPACE::ParseCommandLineFlags;
+
+DEFINE_string(io_trace_file, "", "The IO trace file path.");
+DEFINE_string(
+    output_file, "",
+    "The output file path for storing human readable format of IO trace "
+    "records that are stored in IO trace file.");
```
zhichao-cao, line 26: Would be better to describe the format in the command line

zhichao-cao (approved):
LGTM, thanks!

tools/io_tracer_parser_tool.cc
```diff
@@ -0,0 +1,144 @@
...
+
+#include "port/lang.h"
+#include "trace_replay/io_tracer.h"
+#include "util/gflags_compat.h"
+
+using GFLAGS_NAMESPACE::ParseCommandLineFlags;
+
+DEFINE_string(io_trace_file, "", "The IO trace file path.");
+DEFINE_string(
+    output_file, "",
```
anand1976, line 24: Can we just write to stdout? The user can redirect to a file or pipe it as they see fit.

anand1976 (commented):
LGTM. Wondering if we need the output_file argument, or just write to stdout. It would make it a little easier to use.

tools/io_tracer_parser_tool.cc
akankshamahajan15: In command line using -help?

zhichao-cao: Yes, something like that.

akankshamahajan15: It does print the command line arguments when run 
 ```./io_tracer_parser --help ```

  Flags from tools/io_tracer_parser_tool.cc:   
-io_trace_file (The IO trace file path.) type: string default: ""

akankshamahajan15:
@anand1976  I am getting data race error in  below code for ```tracing_enabled```
```
Status IOTracer::StartIOTrace(Env* env, const TraceOptions& trace_options,
                              std::unique_ptr<TraceWriter>&& trace_writer) {
  InstrumentedMutexLock lock_guard(&trace_writer_mutex_);
  if (writer_.load()) {
    return Status::Busy();
  }
  trace_options_ = trace_options;
  writer_.store(new IOTraceWriter(env, trace_options, std::move(trace_writer)));
  tracing_enabled = true;
  return writer_.load()->WriteHeader();
}
```
at setting ```trace_enabled = true``` and 
then in ```operator->()```  in FileSystem classes when it check 
``` if (is_tracing_enabled())``` which return variable ```tracing_enabled```

We decided that ```writer_``` can continue to be an atomic variable, and ```WriteIOOp``` anyway checks ```writer_ ``` and acquires mutex, so the worst that can happen is a caller calls ```is_tracing_enabled()```, and then calls WriteIOOp, which would simply ignore it if tracing has been disabled in the meantime.

so is there a way to bypass/ignore this race warning in TSAN run for io_tracer_parser_test or should I exclude this test from TSAN?

anand1976:
@akankshamahajan15 Would marking the function with ```__attribute__((no_sanitize("thread")))``` work?

akankshamahajan15:
> @akankshamahajan15 Would marking the function with `__attribute__((no_sanitize("thread")))` work?

Yes this works. Thank you!!

anand1976 (approved):

PR #7334, opened 2020-09-01
Add options for forcing AVX and AVX2 instructions

Daniel-B-Smith:
This PR is set up to merge into master, but it would be great to get this into a patch release if possible.


pdillinger:
I'm curious about the motivation: Bloom filter performance? General performance?

Btw, I'm adding BMI and LZCNT in #7338, if you're interested in adding those to cmake as well (I don't know cmake very well).

Daniel-B-Smith:
The motivation is general performance. My benchmarks are too noisy for me to give a hard number, but they look positive. gcc can do a fair bit of auto-vectorization, so I expect this change to be positive.

Thanks for the heads up about bmi/lzcnt. I ended up adding them when AVX2 is enabled since they came out with the same generation of CPUs and that's already what MSVC does.

jay-zhuang:
I think the ideal solution is integrating cmake with https://github.com/facebook/rocksdb/blob/4b0a509a9126411c492cffed64d123e98b46d385/build_tools/build_detect_platform#L1
So cmake could have the same flags as make.

pdillinger:
> So cmake could have the same flags as make.

I don't think we want cmake bulid to depend on bash (because Windows) but you're right that we want to set the preprocessor macros appropriately in cmake (and buck!) builds. And minimize duplicate infrastructure where possible. I've made myself a note to revisit the issue. Thanks

PR #7336, opened 2020-09-01
remove `DeleteFile()` from public API and language bindings

ajkr:
At the risk of stating the obvious, `DeleteFile()` is not the proper level of abstraction for a key-value store. While it appears at first glance analogous to `AddFile()` (now called `IngestExternalFile()`), it is actually not. File ingestion adds a bunch of updates as the newest component in the LSM, and the outcome is fully deterministic. `DeleteFile()`, on the other hand, removes some subset of the oldest component in the LSM. The data in the oldest component is totally non-deterministic and depends on what compactions have happened. Besides that, the restrictions on what files can be deleted was entirely undocumented.


ajkr:
Created a tag, rocksdb-7.0, for changes like this that should be landed in the next major release.

riversand963:
@ajkr , I feel CompactFiles() are somewhat on the same boat.

ajkr:
Kept it per #8491

PR #7337, opened 2020-09-01
add warning on `DeleteFile()` API

ajkr:
Since we can't land #7336 until the next major release, added a strong warning against the `DeleteFile()` API in the meantime.


ajkr:
> LGTM. You probably should add the same comment to StackableDB.

Done. Thanks for the review!

PR #7342, opened 2020-09-03
Status check enforcement for error_handler_fs_test

zhichao-cao:
Summary: Added status check enforcement for error_test_fs_test

Test Plan: ASSERT_STATUS_CHECKED=1 make -j48 error_test_fs_test

### DIFF
--- a/db/compaction/compaction_iterator.cc
+++ b/db/compaction/compaction_iterator.cc
@@ -125,6 +125,7 @@ CompactionIterator::CompactionIterator(
 CompactionIterator::~CompactionIterator() {
   // input_ Iteartor lifetime is longer than pinned_iters_mgr_ lifetime
   input_->SetPinnedItersMgr(nullptr);
+  status_.PermitUncheckedError();
 }
 
 void CompactionIterator::ResetRecordCounts() {
--- a/db/compaction/compaction_iterator.cc
+++ b/db/compaction/compaction_iterator.cc
@@ -125,7 +125,6 @@ CompactionIterator::CompactionIterator(
 CompactionIterator::~CompactionIterator() {
   // input_ Iteartor lifetime is longer than pinned_iters_mgr_ lifetime
   input_->SetPinnedItersMgr(nullptr);
-  status_.PermitUncheckedError();
 }
 
 void CompactionIterator::ResetRecordCounts() {
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -2609,7 +2609,6 @@ Status DBImpl::CreateColumnFamilyImpl(const ColumnFamilyOptions& cf_options,
     NewThreadStatusCfInfo(
         static_cast_with_check<ColumnFamilyHandleImpl>(*handle)->cfd());
   }
-  persist_options_status.PermitUncheckedError();
   return s;
 }
 
@@ -4554,6 +4553,7 @@ Status DBImpl::IngestExternalFiles(
       // be pessimistic and try write to a new MANIFEST.
       // TODO: distinguish between MANIFEST write and CURRENT renaming
       const IOStatus& io_s = versions_->io_status();
+      // Should handle return error?
       error_handler_.SetBGError(io_s, BackgroundErrorReason::kManifestWrite)
           .PermitUncheckedError();
     }
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -407,7 +407,7 @@ Status DBImpl::ResumeImpl(DBRecoverContext context) {
     // during previous error handling.
     if (file_deletion_disabled) {
       // Always return ok
-      EnableFileDeletions(/*force=*/true);
+      s = EnableFileDeletions(/*force=*/true);
     }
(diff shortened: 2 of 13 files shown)


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3044,6 +3059,7 @@ Status DBImpl::BackgroundCompaction(bool* made_progress,
     m->in_progress = false;  // not being processed anymore
   }
   TEST_SYNC_POINT("DBImpl::BackgroundCompaction:Finish");
+  io_s.PermitUncheckedError();
```
mrambacher, line 3062: Can you move this line to earlier?  It seems like it belongs around 2943 (if status.ok() && ! io_s.ok()))

db/error_handler.cc
```diff
@@ -338,7 +338,7 @@ Status ErrorHandler::SetBGError(const IOStatus& bg_io_err,
   }
   if (BackgroundErrorReason::kManifestWrite == reason) {
     // Always returns ok
-    db_->DisableFileDeletionsWithLock();
+    db_->DisableFileDeletionsWithLock().PermitUncheckedError();
```
mrambacher, line 341: Please add a comment that this always returns OK so the status here is unimportant (or change the method to return a void...)

db/db_impl/db_impl_write.cc
```diff
@@ -848,7 +849,8 @@ void DBImpl::WriteStatusCheck(const Status& status) {
   if (immutable_db_options_.paranoid_checks && !status.ok() &&
       !status.IsBusy() && !status.IsIncomplete()) {
     mutex_.Lock();
-    error_handler_.SetBGError(status, BackgroundErrorReason::kWriteCallback);
+    error_handler_.SetBGError(status, BackgroundErrorReason::kWriteCallback)
```
mrambacher, line 852: 1.  There are other PRs changing these same things at the moment so prepare for merges.

2.  (Same comment on the other PRs):  Why not have this method return a void or create an alternate signature that returns a void, rather than ignoring the result everywhere?

db/db_impl/db_impl.cc
```diff
@@ -407,7 +407,7 @@ Status DBImpl::ResumeImpl(DBRecoverContext context) {
     // during previous error handling.
     if (file_deletion_disabled) {
       // Always return ok
-      EnableFileDeletions(/*force=*/true);
+      s = EnableFileDeletions(/*force=*/true);
```
mrambacher, line 410: Is the "Always return ok" still valid? If so, you might want to PermitUncheckedError here rather than status

db/db_impl/db_impl_compaction_flush.cc
zhichao-cao: Sure.

db/error_handler.cc
zhichao-cao: Sure

db/db_impl/db_impl_write.cc
zhichao-cao: Actually, in the future stage, we want to use the returned states from SetBGError and make the actions. For example, if the error is a retryable I/O error and satisfy some condition, SetBGError might map it to a hard error. When the compaction or flush logic check the returns, it will take actions such as rewrite the file, or fail the current flush. So we prefer to keep the current return.

db/db_impl/db_impl.cc
zhichao-cao: Sure.

db/compaction/compaction_iterator.cc
```diff
@@ -124,6 +124,7 @@ CompactionIterator::CompactionIterator(
 CompactionIterator::~CompactionIterator() {
   // input_ Iteartor lifetime is longer than pinned_iters_mgr_ lifetime
   input_->SetPinnedItersMgr(nullptr);
+  status_.PermitUncheckedError();
```
akankshamahajan15, line 127: Can you explain why you added ```PermitUncheckedError``` instead of checking the status through ```status()``` in the test?

zhichao-cao: If user need the status, they will call `const Status& status()` to get the status during iterator calling. So user might call it and might not. Also, check the status_ in `status()` returns the same status to the user. So I may prefer to add `PermitUncheckedError` to the destructor.

akankshamahajan15: Thank you for clarifying the doubt.

akankshamahajan15 (approved):
LGTM!

db/db_impl/db_impl.cc
```diff
@@ -407,7 +407,7 @@ Status DBImpl::ResumeImpl(DBRecoverContext context) {
     // during previous error handling.
     if (file_deletion_disabled) {
       // Always return ok
-      EnableFileDeletions(/*force=*/true);
+      EnableFileDeletions(/*force=*/true).PermitUncheckedError();
```
akankshamahajan15, line 410: Can we add assert or check status here in case if different status is returned  in future,  PermitUncheckedError might ignore it?

zhichao-cao: I will add an assert here, it always return OK.

PR #7343, opened 2020-09-03
Add file checksum to stress/crash test

pdillinger:
Summary: This change has the crash test randomly select from a few file
checksum implementations, or nullptr, for DB file_checksum_gen_factory.
For compatibility across runs on same DB, each non-null factory can
understand all the other functions, but the default changes.

Test Plan: 'make blackbox_crash_test' for a while, including with some
debug output to ensure code is being exercised.


pdillinger:
Will fix facebook lints before shipping

zhichao-cao (approved):
LGTM, thanks for adding the file checksum generator factory to stress/crash test!

PR #7344, opened 2020-09-03
Fix db_bench_tool_test.  Fixes 7341

mrambacher:
1.  Failed to compile because of use of FileSystem* instead of Env* to some methods;

2.  Failed to compile with addition of ConfigOptions to some methods

3.  Failed to run successfully because the database and/or db_bench would change some of the options, invalidating the comparison

4.  Failed to run successfully if Snappy was not available.

### DIFF
--- a/tools/db_bench_tool_test.cc
+++ b/tools/db_bench_tool_test.cc
@@ -8,6 +8,8 @@
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
 #include "rocksdb/db_bench_tool.h"
+
+#include "db/db_impl/db_impl.h"
 #include "options/options_parser.h"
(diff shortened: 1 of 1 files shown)


jay-zhuang:
Thanks @mrambacher for the fix. Seems the test is not included in the default build and test. I'm wondering if we should add it to `src.mk`.

jay-zhuang (approved):

mrambacher:
> Thanks @mrambacher for the fix. Seems the test is not included in the default build and test. I'm wondering if we should add it to `src.mk`.

There are a few tests that are not included in the src.mk.  env_mirror_test is another (and that one does not pass).  May I suggest another PR to address the src.mk

PR #7346, opened 2020-09-03
Fix wrong level args

cheng-chang:
The level args should be output level instead of input levels.

Test Plan:
make check

### DIFF
--- a/db/compaction/compaction_picker_universal.cc
+++ b/db/compaction/compaction_picker_universal.cc
@@ -1033,9 +1033,9 @@ Compaction* UniversalCompactionBuilder::PickCompactionToOldest(
       MaxFileSizeForLevel(mutable_cf_options_, output_level,
                           kCompactionStyleUniversal),
       LLONG_MAX, path_id,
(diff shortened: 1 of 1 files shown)


ajkr (commented):
LGTM, thanks. Can we verify bottom level compression now works with universal compaction?

ajkr (approved):

ajkr:
Sorry I forgot. This should have a release note since it's a user-visible bug. Here's the one I'm including on the 6.12 backport:

```
### Bug fixes
* Fixed a bug in universal compaction to bottom level where the compression settings for the first input level were used rather than the compression settings for the bottom level.
```

pdillinger:
> Sorry I forgot. This should have a release note since it's a user-visible bug. Here's the one I'm including on the 6.12 backport:

Going to add it to master, and 6.13?

ajkr:
@cheng-chang do you have time to fix up the rest of the release notes? Sorry but there are too many investigations I'm involved in lately.

PR #7348, opened 2020-09-03
Fix, enable, and enhance backup/restore in db_stress

pdillinger:
Summary: Although added to db_stress, testing of backup/restore
was never integrated into the crash test, originally concerned about
performance. I've enabled it now and to address the peformance concern,
testing backup/restore is always skipped once the db exceeds a certain
size threshold, default 100MB. This should provide sufficient
opportunity for testing BackupEngine without bogging down everything
else with heavier and heavier operations.

Also fixed backup/restore in db_stress by making sure PurgeOldbackups
can remove manifest files, which are normally kept around for db_stress.

Added more coverage of backup options, and up to three backups being
saved in one backup directory (in some cases).

Test Plan: ran 'make blackbox_crash_test' for a while, with heightened
probably of taking backups (1/10k). Also confirmed with some debug
output that the code is being covered, TestBackupRestore only takes
a few seconds to complete when triggered, and even at 1/10k and ~50MB
database, there's <,~ 1 thread testing backups at any time.

### DIFF
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -1271,9 +1271,9 @@ Status StressTest::TestBackupRestore(
         thread->rand.OneIn(2) /* verify_with_checksum */);
   }
   if (s.ok()) {
+    int count = static_cast<int>(backup_info.size());
     s = backup_engine->RestoreDBFromBackup(
-        RestoreOptions(),
-        backup_info[thread->rand.Uniform(backup_info.size())].backup_id,
+        RestoreOptions(), backup_info[thread->rand.Uniform(count)].backup_id,
         restore_dir /* db_dir */, restore_dir /* wal_dir */);
   }
   if (s.ok()) {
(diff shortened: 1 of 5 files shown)


db_stress_tool/db_stress_test_base.cc
```diff
@@ -1211,6 +1224,30 @@ Status StressTest::TestBackupRestore(
...
+      backup_opts.share_files_with_checksum = false;
+    } else {
+      backup_opts.share_files_with_checksum = true;
+      if (thread->rand.OneIn(2)) {
+        // old
+        backup_opts.share_files_with_checksum_naming = kChecksumAndFileSize;
+      } else {
+        // new
+        backup_opts.share_files_with_checksum_naming =
+            kOptionalChecksumAndDbSessionId;
```
ajkr, line 1247: Will all files have checksums in the MANIFEST since they're newly created?

ajkr (approved):
LGTM, thanks for finding a way to get this in crash test.

db_stress_tool/db_stress_test_base.cc
pdillinger: They will all have DB session ids, but not DB manifest checksums. We change the DB checksum settings on each restart. So we're exercising the within-release space pretty well (IMHO) but not so much the upgrade scenarios.

PR #7349, opened 2020-09-04
Fix the handling of the case when a blob file with a lower number gets added in VersionBuilder

ltamasi:
Summary:
When multiple background jobs are generating blob files in parallel, it is actually
possible for a blob file to be added with a file number that is lower than the
highest one in the base version. (This is a harmless race condition.) The patch
fixes the handling of this case and adds a unit test.

Test Plan:
`make check`


riversand963 (approved):
Thanks @ltamasi for the fix.

PR #7351, opened 2020-09-05
[cmake] tests need linked with third_party libs

jay-zhuang:
To fix the cmake build with third_party libs, like:
`mkdir build && cd build && cmake .. -DWITH_SNAPPY=1 && make`

Error:
```
Undefined symbols for architecture x86_64:
  "snappy::RawCompress(char const*, unsigned long, char*, unsigned long*)"
```


pdillinger (approved):
👍

PR #7357, opened 2020-09-07
Fix backup/restore in stress/crash test

pdillinger:
Summary: (1) Skip check on specific key if restoring an old backup
(small minority of cases) because it can fail in those cases. (2) Remove
an old assertion about number of column families and number of keys
passed in, which is broken by atomic flush (cf_consistency) test. Like
other code (for better or worse) assume a single key and iterate over
column families. (3) Apply mock_direct_io to NewSequentialFile so that
db_stress backup works on /dev/shm.

Also add more context to output in case of backup/restore db_stress
failure.

Also a minor fix to BackupEngine to report first failure status in
creating new backup.

Reverts "Disable backup/restore stress test (#7350)"

Test Plan: Using backup_one_in=10000,
"USE_CLANG=1 make crash_test_with_atomic_flush" for 30+ minutes
"USE_CLANG=1 make blackbox_crash_test" for 30+ minutes
And with use_direct_reads with TEST_TMPDIR=/dev/shm/rocksdb

### DIFF
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -1182,7 +1182,9 @@ Status BackupEngineImpl::CreateNewBackupWithMetadata(
       new_backup->SetSequenceNumber(sequence_number);
     }
   }
-  ROCKS_LOG_INFO(options_.info_log, "add files for backup done, wait finish.");
+  ROCKS_LOG_INFO(options_.info_log,
+                 "add files for backup done (%s), wait finish.",
+                 s.ok() ? "OK" : "not OK");
   Status item_status;
   for (auto& item : backup_items_to_finish) {
     item.result.wait();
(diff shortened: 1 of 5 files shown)


utilities/backupable/backupable_db.cc
```diff
@@ -1182,7 +1182,9 @@ Status BackupEngineImpl::CreateNewBackupWithMetadata(
       new_backup->SetSequenceNumber(sequence_number);
     }
   }
-  ROCKS_LOG_INFO(options_.info_log, "add files for backup done, wait finish.");
+  ROCKS_LOG_INFO(options_.info_log,
+                 "add files for backup done (%s), wait finish.",
+                 s.ok() ? "OK" : "not OK");
```
riversand963, line 1187: If `s` is not `ok`, can we print `s.ToString()` instead of "not OK"?

riversand963 (approved):
Thanks @pdillinger for the fix.

utilities/backupable/backupable_db.cc
pdillinger: It's printed below in "Backup failed ...". No need to print the details twice. I just find it very misleading to print this line when status is known not-OK. Adding OK vs. not OK can help to localize what reported the non-OK status.

PR #7358, opened 2020-09-08
Fix compile error for old gcc-4.8

jay-zhuang:
gcc-4.8 returns error when using the constructor. Not sure if it's a compiler bug/limitation or code issue:
```
table/block_based/block_based_table_reader.cc:3183:67: error: use of deleted function ‘rocksdb::WritableFileStringStreamAdapter::WritableFileStringStreamAdapter(rocksdb::WritableFileStringStreamAdapter&&)’
```
Workaround this by set the `WritableFile` separately.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,7 @@ jobs:
       - checkout
       - run: pyenv global 3.5.2
       - run: sudo apt-get update -y && sudo apt-get install gcc-4.8 g++-4.8 libgflags-dev
-      - run: CC=gcc-4.8 CXX=g++-4.8 V=1 SKIP_LINK=1 make -j32 all | .circleci/cat_ignore_eagain # Linking broken because libgflags compiled with newer ABI
+      - run: CC=gcc-4.8 CXX=g++-4.8 V=1 SKIP_LINK=1 make -j4 all | .circleci/cat_ignore_eagain # Linking broken because libgflags compiled with newer ABI
 
   build-windows:
     executor: windows-2xlarge
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,7 @@ jobs:
       - checkout
       - run: pyenv global 3.5.2
       - run: sudo apt-get update -y && sudo apt-get install gcc-4.8 g++-4.8 libgflags-dev
-      - run: CC=gcc-4.8 CXX=g++-4.8 V=1 SKIP_LINK=1 make -j32 all | .circleci/cat_ignore_eagain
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -165,6 +165,17 @@ jobs:
...
       - run: TEST_TMPDIR=/dev/shm && make V=1 -j16 unity_test | .circleci/cat_ignore_eagain
 
+  build-linux-gcc-4-8:
+    docker:
+      - image: ubuntu-1604:201903-01
+    resource_class: xlarge
+    steps:
+      - checkout
+      - run: pyenv global 3.5.2
+      - run: sudo apt-get install gcc-4.8 g++-4.8 && sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 50 && sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 50
```
pdillinger, line 175: Should not need both update-alternatives and setting CC and CXX below. Although it offers some protection against accidental use of plain 'gcc' and 'g++' commands in build, I think it's more misleading than protective.

table/block_based/block_based_table_reader.cc
```diff
@@ -3180,7 +3180,8 @@ Status BlockBasedTable::GetKVPairsFromDataBlocks(
 }
 
 Status BlockBasedTable::DumpTable(WritableFile* out_file) {
-  auto out_file_wrapper = WritableFileStringStreamAdapter(out_file);
```
pdillinger, line 3183: IIRC gcc 4.8 has an incomplete / non-final implementation of C++11. I suspect that this would work:
`WritableFileStringStreamAdapter out_file_wrapper(out_file);`

.circleci/config.yml
```diff
@@ -165,6 +165,17 @@ jobs:
...
+      - run: sudo apt-get update -y && sudo apt-get install -y libgflags-dev
+      - run: CC=gcc-4.8 CXX=g++-4.8 V=1 SKIP_LINK=1 make -j32 all | .circleci/cat_ignore_eagain
```
pdillinger, line 177: Please include this comment from the Travis configuration about why SKIP_LINK is used. It could take someone a while to reverse engineer that justification.

`Linking broken because libgflags compiled with newer ABI`

pdillinger (changes requested):

table/block_based/block_based_table_reader.cc
jay-zhuang: You're right, that works. Thanks.

pdillinger (approved):
👍 👍

PR #7361, opened 2020-09-09
More backup/restore stress test fixes

pdillinger:
Summary: (a) Missed a case in updating handling of rand_keys
(b) Only opening restored db with DB::Open so don't (yet)
attempt to open restored BlobDB or TransactionDB.

Test Plan: better than being broken

### DIFF
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -1313,7 +1313,8 @@ Status StressTest::TestBackupRestore(
   }
   DB* restored_db = nullptr;
   std::vector<ColumnFamilyHandle*> restored_cf_handles;
-  if (s.ok()) {
+  // Not yet implemented: opening restored BlobDB or TransactionDB
+  if (s.ok() && !FLAGS_use_txn && !FLAGS_use_blob_db) {
     Options restore_options(options_);
     restore_options.listeners.clear();
     std::vector<ColumnFamilyDescriptor> cf_descriptors;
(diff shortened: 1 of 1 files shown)


db_stress_tool/db_stress_test_base.cc
```diff
@@ -1214,6 +1217,17 @@ void StressTest::TestCompactFiles(ThreadState* /* thread */,
   std::terminate();
 }
 #else   // ROCKSDB_LITE
+void StressTest::ClearBackupRestoreData(ThreadState* thread) {
+  std::string backup_dir = FLAGS_db + "/.backup" + ToString(thread->tid);
+  std::string restore_dir = FLAGS_db + "/.restore" + ToString(thread->tid);
+  // Reasons for this:
+  // * Free up disk space ASAP after crash & restart
+  // * Attempting to restore from backup from prior invocation could fail
```
ajkr, line 1225: Just wondering, do you have an example of such an option?

ajkr (approved):
LGTM!

db_stress_tool/db_stress_test_base.cc
pdillinger: Here's the failure I'm trying to fix:

    Failure in DB::Open in backup/restore with: Not implemented: WritePrepared/WriteUnprepared txn tag when write_after_commit_ is enabled (in default WriteCommitted mode). If it is not due to corruption, the WAL must be emptied before changing the WritePolicy.

After private chat, my diagnosis is a bit wrong. We can't (necessarily) use vanilla DB::Open when blobdb or txndb are enabled. Revision coming

Issue #7340, opened 2020-09-02
[BUG]FlushMemTable races with each other and crashes under DEBUG mode


wolfkdy:
this issue is duplicated with https://github.com/facebook/rocksdb/issues/5126
I open another ticket as the previous one is opened one year ago and seems untracked now as no one contributed a unittest.
this issue won't be a problem in release mode as asserts are eliminated.
The two asserts won't be held.
```
Status DBImpl::FlushMemTableToOutputFile(
    ColumnFamilyData* cfd, const MutableCFOptions& mutable_cf_options,
    bool* made_progress, JobContext* job_context,
    SuperVersionContext* superversion_context, LogBuffer* log_buffer) {
  mutex_.AssertHeld();
  assert(cfd->imm()->NumNotFlushed() != 0);
  assert(cfd->imm()->IsFlushPending());

```

### Steps to reproduce the behavior
+ flush thread 1 pick column family 0，1，2,  3 then flush 0 (release the mutex lock). This is constructed by Writing to all cfs and issuing a switch wal request.
+ when thread 1 is flushing cf 0(with no mutex lock), user issue a Put and Flush to cf 1. let's assume flush thread 2 handles this request.
+ flush thread 2 flushes all the memtables of cf 1 and returns OK.
+ thread 1 is re-scheduled and tries to flush cf 1 and asserts.

these steps have been constributed into the unittest below, which can reproduce the two asserts.
The test works under 5.18.3, I guess it should also work under master branch.

```
TEST_F(DBFlushTest, ParallelFlush) {                                                                                
  Options options;                                                                                                  
  options.disable_auto_compactions = true;                                                                          
  options.max_background_flushes = 20;                                                                              
  options.env = env_;                                                                                               
  CreateAndReopenWithCF({"1", "2", "3"}, options);                                                                  
  FlushOptions flushOpt;                                                                                            
  flushOpt.wait = true;                                                                                             
  flushOpt.allow_write_stall=true;                                                                                  
                                                                                                                    
  bool triggered = false;                                                                                           
  SyncPoint::GetInstance()->SetCallBack(                                                                            
      "FlushJob::WriteLevel0Table", [&](void* /*arg*/) {                                                            
         if (triggered) return;                                                                                     
         triggered = true;                                                                                          
         ASSERT_OK(Put(static_cast<int>(1) /*cf*/, "key", "value"));                                                
         ASSERT_OK(dbfull()->Flush(flushOpt, handles_[1]));                                                         
[...]

PR #7362, opened 2020-09-09
Fix assertion failure in bg flush

riversand963:
#7340 reports and reproduces an assertion failure caused by a combination of the following:
- atomic flush is disabled.
- a column family can appear multiple times in the flush queue at the same time. This behavior was introduced in release 5.17.

Consequently, it is possible that two flushes race with each other. One bg flush thread flushes all memtables. The other thread calls `FlushMemTableToOutputFile()` afterwards, and hits the assertion error below.

```
  assert(cfd->imm()->NumNotFlushed() != 0);
  assert(cfd->imm()->IsFlushPending());
```

Fix this by reverting the behavior. In non-atomic-flush case, a column family can appear in the flush queue at most once at the same time.

Test plan:
make check
Also run stress test successfully for 10 times.
```
make crash_test
```


jay-zhuang:
Does this impact `atomic_flush = true`?

Seems the unittest provided in #7340 hang if `atomic_flush = true`.

riversand963:
> Does this impact `atomic_flush = true`?
> 
> Seems the unittest provided in #7340 hang if `atomic_flush = true`.

The changes in this PR should apply to non-atomic-flush.
If you set `atomic_flush` to true in the aforementioned unit test, the bg flush thread will be waiting on something that it is going to do later, thus the wait will never finish, as expected.

riversand963:
Requesting review from @ajkr since you recently reviewed these files and are familiar with flush code path (I think). This is not  very urgent, but getting rid of the assertion failure is also a nice thing. Let me know if you would like to review it. Thanks

zhichao-cao:
I spend some time and debugged the `assert(cfd->imm()->NumNotFlushed() > 0)` failure in FlushMemtable, that failure might not related to this one. The cause is, bg_cv_.SignalAll() in BackgroundCallFlush is after MaybeScheduleFlushOrCompaction() is called, the ResumeImpl() will get out of WaitForBackgroundWork() only when all the memtable is flushed by the regular scheduled flush (https://github.com/facebook/rocksdb/blob/master/db/db_impl/db_impl_compaction_flush.cc#L2495). In this case, when a new flush (FlushMemtable()) is scheduled by the ResumeImpl(), it will fail in https://github.com/facebook/rocksdb/blob/master/db/db_impl/db_impl_compaction_flush.cc#L1730 assertion.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -2256,19 +2269,36 @@ ColumnFamilyData* DBImpl::PickCompactionFromQueue(
...
-    cfd->SetFlushReason(flush_reason);
+  if (!immutable_db_options_.atomic_flush) {
+    // For the non-atomic flush case, we never schedule multiple column
+    // families in the same flush request.
+    assert(flush_req.size() == 1);
+    ColumnFamilyData* cfd = flush_req[0].first;
+    assert(cfd);
+    if (!cfd->queued_for_flush() && cfd->imm()->IsFlushPending()) {
+      cfd->Ref();
+      cfd->set_queued_for_flush(true);
```
anand1976, line 2284: Is there any harm in setting/checking this even for atomic flush? Would simplify the code a bit.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -2256,19 +2269,36 @@ ColumnFamilyData* DBImpl::PickCompactionFromQueue(
...
+  mutex_.AssertHeld();
   if (flush_req.empty()) {
     return;
   }
-  for (auto& iter : flush_req) {
-    ColumnFamilyData* cfd = iter.first;
-    cfd->Ref();
-    cfd->SetFlushReason(flush_reason);
+  if (!immutable_db_options_.atomic_flush) {
+    // For the non-atomic flush case, we never schedule multiple column
```
anand1976, line 2277: Is there a need to impose this restriction here? It seems like the whole code flow could be simpler if we simply loop through ```flush_req``` here and queue a column family if its not been queued before, skip it otherwise. Maybe I'm missing something.

db/db_impl/db_impl_compaction_flush.cc
riversand963: `queue_for_flush` is not used by atomic flush. In atomic flush, a column family can appear in multiple flush requests in the flush queue at the same time. Using a boolean is not useful because we want to clear this variable when a flush is about to be performed for this column family. There is not an easy way to determine when to clear it, unless we use an integer count (which can be done later in a separate PR). If we set the flag but not clear it, then the assertion for `queued_for_flush_` being false will fail in `~ColumnFamilyData()`.

db/db_impl/db_impl_compaction_flush.cc
riversand963: In the non-atomic flush case, we would like bg flush for different column families to run in different bg threads. This is actually the behavior before atomic flush was introduced. The advantage is that flush can finish more quickly to avoid write stalls when there are sufficient threads in the pool. This can be beneficial for the case of remote storage, too.

riversand963:
Thanks @anand1976 for the review

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1734,7 +1728,8 @@ Status DBImpl::FlushMemTable(ColumnFamilyData* cfd,
       if (cfd->imm()->NumNotFlushed() != 0 || !cfd->mem()->IsEmpty() ||
           !cached_recoverable_state_empty_.load()) {
         flush_memtable_id = cfd->imm()->GetLatestMemTableID();
-        flush_req.emplace_back(cfd, flush_memtable_id);
+        FlushRequest req{{cfd, flush_memtable_id}};
```
ajkr, line 1731: It's a bit hard for me to reason about restricting the flush to `flush_memtable_id` while at the same time allowing only one entry per CF in the flush queue. If there's a second manual flush to the same CF while the first is still queued, how do we know the second manual flush will make progress? It looks like `SchedulePendingFlush()` will skip scheduling the second one due to `cfd->queued_for_flush()`, but the one in the queue would not include the latest memtable. 

It looks like we are saved from the above worry because the memtable ID is ignored (?) in non-atomic flush, but still that feels a bit risky. I wonder if we should be more explicit here, passing the maximum possible value or something.

ajkr (approved):
LGTM. I think it is a good approach to have each CF appear at most once in the flush queue for non-atomic flush, where a flush will always pick all available memtables when the job starts. Although one of the code paths is a bit confusing as it appears manual flush in the non-atomic case is setting `BGFlushArg::max_memtable_id_` to something possibly smaller than what it actually ends up flushing (please correct me if needed). It also seems a good change to avoid `FlushRequest` containing multiple entries when it's not needed, i.e., outside of atomic flush.

db/db_impl/db_impl_compaction_flush.cc
riversand963: In non-atomic flush, `FlushJob::max_memtable_id_` is nullptr, thus `PickMemtables()` will pick all memtables. In atomic flush, it is `&flush_memtable_id`, thus atomic flush will flush memtables up to this. We make this difference when we construct `FlushJob` objects. It is a little misleading and I will change it.

riversand963:
Thanks @ajkr for the review!

db/flush_job.cc
```diff
@@ -80,22 +80,24 @@ const char* GetFlushReasonString (FlushReason flush_reason) {
...
-    FSDirectory* output_file_directory, CompressionType output_compression,
-    Statistics* stats, EventLogger* event_logger, bool measure_io_stats,
-    const bool sync_output_directory, const bool write_manifest,
-    Env::Priority thread_pri, const std::shared_ptr<IOTracer>& io_tracer,
-    const std::string& db_id, const std::string& db_session_id,
-    std::string full_history_ts_low)
+FlushJob::FlushJob(const std::string& dbname, ColumnFamilyData* cfd,
+                   const ImmutableDBOptions& db_options,
+                   const MutableCFOptions& mutable_cf_options,
+                   uint64_t max_memtable_id, const FileOptions& file_options,
```
riversand963, line 86: The only change here is from `const uint64_t* max_memtable_id` to `uint64_t max_memtable_id`. The rest results from auto-formatting.

PR #7365, opened 2020-09-09
Expose the start of the expiration range for TTL blob files through LiveFileMetaData

ltamasi:
Summary:
The patch adds support for exposing the start of the expiration range
for TTL blob files through the `GetLiveFilesMetaData` API. This can be
used for monitoring purposes, i.e. to make sure TTL blob files are
deleted in a timely manner. The patch also fixes a couple of uninitialized
variable issues.

Test Plan:
`make check`

### DIFF
--- a/include/rocksdb/metadata.h
+++ b/include/rocksdb/metadata.h
@@ -62,7 +62,9 @@ struct SstFileMetaData {
         being_compacted(false),
         num_entries(0),
         num_deletions(0),
-        oldest_blob_file_number(0) {}
+        oldest_blob_file_number(0),
+        oldest_ancester_time(0),
+        file_creation_time(0) {}
 
   SstFileMetaData(const std::string& _file_name, uint64_t _file_number,
                   const std::string& _path, size_t _size,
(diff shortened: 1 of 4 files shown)


include/rocksdb/metadata.h
```diff
@@ -117,6 +119,8 @@ struct SstFileMetaData {
   // oldest SST file that is the compaction ancester of this file.
   // The timestamp is provided Env::GetCurrentTime().
   // 0 if the information is not available.
+  //
+  // Note: for TTL blob files, it contains the start of the expiration range.
   uint64_t oldest_ancester_time;
```
pdillinger, line 124: Before this change, this field would generally hold a timestamp in the past, but now for blob files it will often be in the future?

Just an oddity. An alternative might not be better.

pdillinger (approved):
LGTM

jay-zhuang (approved):

include/rocksdb/metadata.h
ltamasi: Agreed. The field does serve a similar purpose though for both SSTs and TTL blob files.

PR #7366, opened 2020-09-10
Add Stats for MultiGet

akankshamahajan15:
Summary: Add following stats for MultiGet in Histogram to get more insight on MultiGet.
    1. Number of index and filter blocks read from file as part of MultiGet
    request per level.
    2. Number of data blocks read from file per level.
    3. Number of SST files loaded from file system per level.

Test Plan:

Reviewers:

Subscribers:

Tasks: T69090067

Tags:

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -1482,6 +1482,21 @@ Status BlockBasedTable::MaybeReadBlockAndLoadToCache(
         s = block_fetcher.ReadBlockContents();
         raw_block_comp_type = block_fetcher.get_compression_type();
         contents = &raw_block_contents;
+        if (get_context) {
+          switch (block_type) {
+            case BlockType::kIndex:
+              ++get_context->get_context_stats_.num_index_read;
+              break;
+            case BlockType::kFilter:
+              ++get_context->get_context_stats_.num_filter_read;
+              break;
+            case BlockType::kData:
+              ++get_context->get_context_stats_.num_data_read;
+              break;
+            default:
+              break;
+          }
+        }
       } else {
         raw_block_comp_type = contents->get_compression_type();
       }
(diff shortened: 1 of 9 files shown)


akankshamahajan15:
Working on writing the test case for issuing MultiGet without cache and print the stats.

anand1976 (commented):
This LGTM. We need to think about how to aggregate these stats per MultiGet. Maybe we can have avg # of files per level with > 0 data blocks read. Similar for index and filter blocks.

Another issue is the user may not enable per level perf context, in which case we will have no visibility.

akankshamahajan15:
> This LGTM. We need to think about how to aggregate these stats per MultiGet. Maybe we can have avg # of files per level with > 0 data blocks read. Similar for index and filter blocks.
> 
> Another issue is the user may not enable per level perf context, in which case we will have no visibility.

Ok. I will explore further how to add these in RecordTicks.

table/block_based/block_based_table_reader.cc
```diff
@@ -2430,6 +2461,10 @@ void BlockBasedTable::MultiGet(const ReadOptions& read_options,
       iiter_unique_ptr.reset(iiter);
     }
 
+    if (sst_file_range.begin()->get_context) {
+      ++(sst_file_range.begin()->get_context->get_context_stats_.num_sst_read);
```
anand1976, line 2465: This shouldn't be incremented unconditionally, instead only when blocks have to be read. Line 2589 might be a better place to increment it.

akankshamahajan15:
Addressed comments

anand1976:
The code LGTM. Can we add a test to verify the stats are being set as expected?

table/block_based/block_based_table_reader.cc
```diff
@@ -1889,6 +1904,22 @@ Status BlockBasedTable::RetrieveBlock(
         GetMemoryAllocator(rep_->table_options), for_compaction,
         rep_->blocks_definitely_zstd_compressed,
         rep_->table_options.filter_policy.get());
+
+    if (get_context) {
```
akankshamahajan15, line 1908: @anand1976  In ```TableCache::MultiGet```, we first get the ```TableReader``` and then call the ```BlockBasedTable::MultiGet```. In ```TableCahe::GetTableReader``` flow, it creates Index reader which call ```RetrieveBlock``` for index block type with ```get_context = nullptr```. So that won't be logged as of now. Is this ok for the experiment or that should also be recorded?

anand1976: @akankshamahajan15 Its ok to not record it. We don't plan to parallelize file open anytime soon. Even if we want to track it at some point, it should probably be a separate stat.

akankshamahajan15: Thanks for clarifying. I have unit test ready. I will upload it by EOD.

anand1976 (approved):
LGTM. Thanks!

PR #7367, opened 2020-09-10
Add more tests to ASSERT_STATUS_CHECKED

mrambacher:
db_options_test
options_file_test
auto_roll_logger_test
options_util_test
persistent_cache_test


db/compaction/compaction_job.cc
```diff
@@ -606,6 +606,10 @@ Status CompactionJob::Run() {
       status = state.status;
       io_s = state.io_status;
       break;
+    } else if (!state.io_status.ok()) {
```
zhichao-cao, line 609: If `state.io_status` is not ok, `state.status` must be not ok. So this else branch should never be executed.

db/compaction/compaction_job.cc
```diff
@@ -1079,7 +1080,7 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
     CompactionIterationStats range_del_out_stats;
     Status s = FinishCompactionOutputFile(status, sub_compact, &range_del_agg,
                                           &range_del_out_stats);
-    if (status.ok()) {
+    if (!s.ok() && status.ok()) {
```
zhichao-cao, line 1083: This check `!s.ok()` is unnecessary.

logging/auto_roll_logger.cc
```diff
@@ -45,8 +45,8 @@ AutoRollLogger::AutoRollLogger(Env* env, const std::string& dbname,
     RollLogFile();
   }
   GetExistingFiles();
-  ResetLogger();
-  if (status_.ok()) {
+  s = ResetLogger();
+  if (s.ok()) {
```
zhichao-cao, line 49: Should we also check `status_ `here?

db/compaction/compaction_job.cc
mrambacher: The check is necessary to "clear" the checked flag in s.  Otherwise, we may not clear it.

zhichao-cao (approved):
LGTM, thanks!

PR #7368, opened 2020-09-10
Disallow trivial move if BottommostLevelCompaction is kForce*

jay-zhuang:
If `BottommostLevelCompaction.kForce*` is set, compaction should avoid
trivial move and always compact the sst to the target size.

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -873,6 +873,7 @@ Status DBImpl::CompactRange(const CompactRangeOptions& options,
       int output_level;
       for (int level = first_overlapped_level; level <= max_overlapped_level;
            level++) {
+        bool disallow_trivial_move = false;
         // in case the compaction is universal or if we're compacting the
         // bottom-most level, the output level will be the same as input one.
         // level 0 can never be the bottommost level (i.e. if all files are in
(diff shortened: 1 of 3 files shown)


jay-zhuang:
Rebase to master. @ajkr please review when you have time.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -883,9 +884,19 @@ Status DBImpl::CompactRange(const CompactRangeOptions& options,
               level == 0) {
             output_level = ColumnFamilyData::kCompactToBaseLevel;
           }
+          // if it's a BottommostLevel compaction and `kForce*` compaction is
+          // set, disallow trivial move
+          if (level == max_overlapped_level &&
```
ajkr, line 889: The existing conditionals feel a bit too complicated. I wonder if we can decide whether to allow trivial move outside of them, just above the `RunManualCompaction()`. Something like, disable trivial move if `level == max_output_level && output_level != level && options.bottommost_level_compaction ...`.

ajkr (approved):
LGTM, thanks!

ajkr:
> LGTM, thanks!

Also can we add a "Bug Fixes" release note, as it does fix a user-visible contract violation (although quite a special case)?

db/db_impl/db_impl_compaction_flush.cc
ajkr: Feel free to land this as is. I don't think the solution proposed above really helps. Calling `RunManualCompaction()` outside the loop when `final_output_level > max_overlapped_level` would probably help, but we can revisit it later as it's better to have something that works.

jay-zhuang: I agree that the logic here is getting more complicated and worth refactoring. I'll commit this change first to catch the next release. And revisit this later.

PR #7374, opened 2020-09-10
Add a new IOStatus subcode to indicate that writes are fenced off

anand1976:
In a distributed file system, directory ownership is enforced by fencing
off the previous owner once they've been preempted by a new owner. This
PR adds a IOStatus subcode for ```StatusCode::IOError``` to indicate this.
Once this error is returned for a file write, the DB is put in read-only
mode and not allowed to resume in read-write mode.

Test Plan:
Add new unit tests in ```error_handler_fs_test```


db/error_handler_fs_test.cc
```diff
@@ -1962,6 +1962,191 @@ TEST_F(DBErrorHandlingFSTest, WALWriteRetryableErrorAutoRecover2) {
...
+  listener->EnableAutoRecovery(true);
+  DestroyAndReopen(options);
+
+  Put(Key(0), "val");
+  SyncPoint::GetInstance()->SetCallBack("FlushJob::Start", [&](void*) {
+    fault_fs->SetFilesystemActive(false, IOStatus::IOFenced("IO fenced"));
+  });
+  SyncPoint::GetInstance()->EnableProcessing();
+  s = Flush();
+  ASSERT_EQ(s.severity(), ROCKSDB_NAMESPACE::Status::Severity::kFatalError);
```
riversand963, line 1990: Should we also `ASSERT_EQ(fenced, s)` here?

db/error_handler_fs_test.cc
```diff
@@ -1962,6 +1962,191 @@ TEST_F(DBErrorHandlingFSTest, WALWriteRetryableErrorAutoRecover2) {
...
+    for (auto i = 0; i < 100; ++i) {
+      batch.Put(Key(i), rnd.RandomString(1024));
+    }
+
+    WriteOptions wopts;
+    wopts.sync = true;
+    s = dbfull()->Write(wopts, &batch);
+    ASSERT_TRUE(s.IsIOFenced());
+  }
+  Close();
```
riversand963, line 2144: Nit: this `Close` may not be necesssary.

db/db_impl/db_impl_write.cc
```diff
@@ -854,8 +854,9 @@ void DBImpl::WriteStatusCheck(const Status& status) {
 void DBImpl::IOStatusCheck(const IOStatus& io_status) {
   // Is setting bg_error_ enough here?  This will at least stop
   // compaction and fail any further writes.
-  if (immutable_db_options_.paranoid_checks && !io_status.ok() &&
-      !io_status.IsBusy() && !io_status.IsIncomplete()) {
+  if ((immutable_db_options_.paranoid_checks && !io_status.ok() &&
```
riversand963, line 857: We are not adding this check to `WriteStatusCheck` because we expect this to occur only for `io_status`? If so, can we add an assertion?

riversand963 (approved):
Thanks @anand1976 for the PR. Only a few minor comments.

db/db_impl/db_impl_write.cc
anand1976: Yes, any IO errors are supposed to be handled based on the ```IOStatus```.  If ```WriteStatusCheck``` is called for an IO error, ```SetBGError``` should have already been called for it. I'll add an assertion for it.

db/error_handler_fs_test.cc
anand1976: The ```Close``` is actually needed here in order to ensure the DB is closed before the ```Env``` is destroyed.

PR #7375, opened 2020-09-10
Assert valid linked list for write group

pdillinger:
Summary: We've seen some segfaults in db_write_test, with at least one
suggesting corruption of a write group linked list. Adding an assertion
to have this fail in a more specific way if that is the broken
invariant.

Test Plan: make check


mrambacher (approved):
LGTM.

Should a comparable check be put into EnterAsMemTableWriter?

anand1976 (approved):
LGTM

PR #7377, opened 2020-09-11
Update comments to remove out-of-date information

jay-zhuang:
`BlockBasedTableOptions` is no longger mutable. It's used by
`BlockBasedTableFactory` for `options.table_factory`, seems not easy to
make it dynamically changeable:
https://github.com/facebook/rocksdb/blob/b9bb59d49df264363d7250ceba59213706b3bb61/options/cf_options.cc#L511-L514

Test Plan:
```
dbfull()->SetOptions({{"block_based_table_factory", "{block_size=7k;}"}});
// get error: Invalid argument: Option not changeable: block_based_table_factory
```

### DIFF
--- a/include/rocksdb/table.h
+++ b/include/rocksdb/table.h
@@ -163,7 +163,7 @@ struct BlockBasedTableOptions {
   // Approximate size of user data packed per block.  Note that the
   // block size specified here corresponds to uncompressed data.  The
   // actual size of the unit read from disk may be smaller if
-  // compression is enabled.  This parameter can be changed dynamically.
+  // compression is enabled.
   size_t block_size = 4 * 1024;
 
   // This is used to close a block before it reaches the configured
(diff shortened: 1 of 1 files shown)


jay-zhuang:
Here is the question the user asked: https://stackoverflow.com/questions/63716516/how-to-dynamically-change-rocksdbs-block-size

mrambacher:
I agree this might not be easy to fix and do not believe it has worked for quite some time (at least as far back as 6.9, if ever).  One big issue is that the TableFactory is not a member of the MutableCFOptions structure, which means that no TableFactory options can be changed via this API.

@siying Is this something that should be fixed?  If so, we can open an issue and I will take a look at it (and this PR should be abandoned).  

FWIW, making options Mutable would be *much simpler* if we changed the Options hierarchy (e.g. struct DBOptions: public struct MutableDBOptions).

jay-zhuang:
> I agree this might not be easy to fix and do not believe it has worked for quite some time (at least as far back as 6.9, if ever). One big issue is that the TableFactory is not a member of the MutableCFOptions structure, which means that no TableFactory options can be changed via this API.
> 
> @siying Is this something that should be fixed? If so, we can open an issue and I will take a look at it (and this PR should be abandoned).
> 
> FWIW, making options Mutable would be _much simpler_ if we changed the Options hierarchy (e.g. struct DBOptions: public struct MutableDBOptions).

@mrambacher I think it would be great if we can make it dynamically changeable again. Would you please take a look see if it would be an easy fix?

jay-zhuang:
@mrambacher do you have plan to make `block_size` changeable? Or should we just update the doc that it's immutable?

PR #7379, opened 2020-09-13
Makefile support subset/individual valgrind tests

ajkr:
Introduced `valgrind_check_some`, which is analogous to the `check_some` target for non-valgrind tests. It simplifies the process for running a single valgrind test or subset of valgrind tests when trying to repro a failure.

I also added a `ROCKSDBTESTS_ONLY` parameter, which simplifies selecting a single test to run. Previously the user would have to use `ROCKSDBTESTS_START` and `ROCKSDBTESTS_END`, but it was difficult to determine the end variable since it is an exclusive endpoint and must match an actual test name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -629,13 +629,13 @@ endif
 
 	# Enable building all unit tests, but use check_some to run only tests
 	# known to pass ASC (ASSERT_STATUS_CHECKED)
-	SUBSET := $(TESTS_PASSING_ASC)
+	ROCKSDBTESTS_SUBSET ?= $(TESTS_PASSING_ASC)
 	# Alternate: only build unit tests known to pass ASC, and run them
 	# with make check
 	#TESTS := $(filter $(TESTS_PASSING_ASC),$(TESTS))
 	#PARALLEL_TEST := $(filter $(TESTS_PASSING_ASC),$(PARALLEL_TEST))
 else
-	SUBSET := $(TESTS)
+	ROCKSDBTESTS_SUBSET ?= $(TESTS)
 endif
 # Not necessarily well thought out or up-to-date, but matches old list
 TESTS_PLATFORM_DEPENDENT := \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -677,6 +677,10 @@ ifdef ROCKSDBTESTS_END
         SUBSET := $(shell echo $(SUBSET) | sed 's/$(ROCKSDBTESTS_END).*//')
 endif
 
+ifdef ROCKSDBTESTS_ONLY
+        SUBSET := $(filter $(ROCKSDBTESTS_ONLY), $(SUBSET))
```
pdillinger, line 681: To me it doesn't make sense for SUBSET and ROCKSDBTESTS_ONLY to be separate variables. As implemented, if I write

    make ROCKSDBTESTS_ONLY=does_not_exist_test check_some

it would trivially succeed. Without this change,

    $ make SUBSET=does_not_exist_test check_some
    ...
    make: *** No rule to make target `does_not_exist_test', needed by `check_some'.  Stop.

Which I think is a superior behavior. Note that for `SUBSET=does_not_exist_test make check_some` to work, the two alternate initial assignments to SUBSET would need to use `?=` instead of `:=`.

Renaming SUBSET to be RocksDB-specific is good by me.

pdillinger (approved):
Sure, though I would always use

    ROCKSDB_VALGRIND_RUN=1 DISABLE_JEMALLOC=1 make -j24 check
    ROCKSDB_VALGRIND_RUN=1 DISABLE_JEMALLOC=1 valgrind --leak-check=full ./some_test

(Not asking for another round of approvals re: my concern)

Makefile
mrambacher: Is there a better way to do some of this work rather than directly int he Makefile?  Can the test rule that does parallel tests be separated into its own script (for example) so that the script could be used by both check and check_some and  CMake tests and so on...

ajkr: >Which I think is a superior behavior. Note that for SUBSET=does_not_exist_test make check_some to work, the two alternate initial assignments to SUBSET would need to use ?= instead of :=.
>
> Renaming SUBSET to be RocksDB-specific is good by me.

Sounds good, I will make this change.

> Is there a better way to do some of this work rather than directly int he Makefile? Can the test rule that does parallel tests be separated into its own script (for example) so that the script could be used by both check and check_some and CMake tests and so on...

Probably. I was also recently wanting `check_some` to use the same parallelism as when the same tests are run under `check`. So I think what you're suggesting would be quite useful.

ajkr: > > Which I think is a superior behavior. Note that for SUBSET=does_not_exist_test make check_some to work, the two alternate initial assignments to SUBSET would need to use ?= instead of :=.
> > Renaming SUBSET to be RocksDB-specific is good by me.
> 
> Sounds good, I will make this change.

Done.

ajkr (commented):
Thanks for the reviews!

PR #7389, opened 2020-09-15
Add basic support for user-defined timestamp to db_bench

riversand963:
Update db_bench so that we can run it with user-defined timestamp.
Currently, only 64-bit timestamp is supported, while others are disabled by assertion.

Test plan:
./db_bench -benchmarks=fillseq,fillrandom,readrandom,readsequential,....., -user_timestamp_size=8

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -292,6 +292,9 @@ static bool ValidateUint32Range(const char* flagname, uint64_t value) {
 
 DEFINE_int32(key_size, 16, "size of each key");
 
+DEFINE_int32(user_timestamp_size, 0,
+             "number of bytes in a user-defined timestamp");
+
 DEFINE_int32(num_multi_db, 0,
              "Number of DBs used in the benchmark. 0 means single DB.");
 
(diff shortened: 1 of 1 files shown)


tools/db_bench_tool.cc
```diff
@@ -6246,15 +6445,24 @@ class Benchmark {
...
+    if (user_timestamp_size_ > 0) {
+      ts_guard.reset(new char[user_timestamp_size_]);
+    }
     // the number of iterations is the larger of read_ or write_
     while (!duration.Done(1)) {
       DB* db = SelectDB(thread);
       GenerateKeyFromInt(thread->rand.Next() % FLAGS_num, FLAGS_num, &key);
+      Slice ts;
+      if (user_timestamp_size_ > 0) {
+        ts = mock_app_clock_->Allocate(ts_guard.get());
```
ltamasi, line 6458: Just to confirm: we're intentionally reading the latest value unconditionally here and in the test cases below, right?

ltamasi (approved):
Thanks for the patch @riversand963!

tools/db_bench_tool.cc
```diff
@@ -1339,6 +1342,9 @@ DEFINE_bool(report_file_operations, false, "if report number of file "
             "operations");
 DEFINE_int32(readahead_size, 0, "Iterator readahead size");
 
+DEFINE_bool(read_with_latest_user_timestamp, true,
+            "Use latest timestamp for read");
```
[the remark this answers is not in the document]
riversand963: SG. Will update.

tools/db_bench_tool.cc
riversand963: Yes. This is due to the operation being read-modify-write. Therefore, we read the latest using the newest timestamp.

ltamasi: Yeah that's what I thought :) Thanks!

PR #7404, opened 2020-09-17
MultiGet() with timestamp should respect snapshot

jay-zhuang:
Similar to PR #7227, add read callback to filter out rows with with
higher sequence number.

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -1815,6 +1815,9 @@ std::vector<Status> DBImpl::MultiGet(
           read_options, nullptr, iter_deref_lambda, &multiget_cf_data,
           &consistent_seqnum);
 
+  TEST_SYNC_POINT("DBImpl::MultiGet:AfterGetSeqNum1");
+  TEST_SYNC_POINT("DBImpl::MultiGet:AfterGetSeqNum2");
+
   // Contain a list of merge operations if merge occurs.
   MergeContext merge_context;
 
@@ -1837,6 +1840,14 @@ std::vector<Status> DBImpl::MultiGet(
   size_t num_found = 0;
   size_t keys_read;
   uint64_t curr_value_size = 0;
+
+  GetWithTimestampReadCallback timestamp_read_callback(0);
+  ReadCallback* read_callback = nullptr;
+  if (read_options.timestamp->size() > 0) {
+    timestamp_read_callback.Refresh(consistent_seqnum);
+    read_callback = &timestamp_read_callback;
+  }
+
   for (keys_read = 0; keys_read < num_keys; ++keys_read) {
     merge_context.Clear();
     Status& s = stat_list[keys_read];
@@ -1857,12 +1868,14 @@ std::vector<Status> DBImpl::MultiGet(
     bool done = false;
     if (!skip_memtable) {
       if (super_version->mem->Get(lkey, value, timestamp, &s, &merge_context,
-                                  &max_covering_tombstone_seq, read_options)) {
+                                  &max_covering_tombstone_seq, read_options,
+                                  read_callback)) {
         done = true;
         RecordTick(stats_, MEMTABLE_HIT);
-      } else if (super_version->imm->Get(
-                     lkey, value, timestamp, &s, &merge_context,
-                     &max_covering_tombstone_seq, read_options)) {
+      } else if (super_version->imm->Get(lkey, value, timestamp, &s,
+                                         &merge_context,
+                                         &max_covering_tombstone_seq,
+                                         read_options, read_callback)) {
         done = true;
         RecordTick(stats_, MEMTABLE_HIT);
       }
@@ -1872,7 +1885,8 @@ std::vector<Status> DBImpl::MultiGet(
       PERF_TIMER_GUARD(get_from_output_files_time);
       super_version->current->Get(read_options, lkey, &pinnable_val, timestamp,
                                   &s, &merge_context,
-                                  &max_covering_tombstone_seq);
+                                  &max_covering_tombstone_seq, nullptr, nullptr,
+                                  nullptr, read_callback);
       value->assign(pinnable_val.data(), pinnable_val.size());
       RecordTick(stats_, MEMTABLE_MISS);
     }
@@ -2130,12 +2144,19 @@ void DBImpl::MultiGet(const ReadOptions& read_options, const size_t num_keys,
       read_options, nullptr, iter_deref_lambda, &multiget_cf_data,
       &consistent_seqnum);
 
+  GetWithTimestampReadCallback timestamp_read_callback(0);
+  ReadCallback* read_callback = nullptr;
+  if (read_options.timestamp->size() > 0) {
+    timestamp_read_callback.Refresh(consistent_seqnum);
+    read_callback = &timestamp_read_callback;
+  }
+
   Status s;
   auto cf_iter = multiget_cf_data.begin();
   for (; cf_iter != multiget_cf_data.end(); ++cf_iter) {
     s = MultiGetImpl(read_options, cf_iter->start, cf_iter->num_keys,
                      &sorted_keys, cf_iter->super_version, consistent_seqnum,
-                     nullptr, nullptr);
+                     read_callback, nullptr);
     if (!s.ok()) {
       break;
     }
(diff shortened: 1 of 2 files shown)


db/db_impl/db_impl.cc
```diff
@@ -1872,7 +1885,8 @@ std::vector<Status> DBImpl::MultiGet(
       PERF_TIMER_GUARD(get_from_output_files_time);
       super_version->current->Get(read_options, lkey, &pinnable_val, timestamp,
                                   &s, &merge_context,
-                                  &max_covering_tombstone_seq);
+                                  &max_covering_tombstone_seq, nullptr, nullptr,
```
riversand963, line 1888: Nit: it improves readability by doing sth. like `/*value_found=*/nullptr` for constants.

db/db_with_timestamp_basic_test.cc
```diff
@@ -590,8 +590,130 @@ TEST_F(DBBasicTestWithTimestamp, CompactDeletionWithTimestampMarkerToBottom) {
...
+    AssertVisibility(ts, seq, s8);
+  }
+
+  void VerifyDefaultCF(const Snapshot* snap = nullptr) {
+    for (int i = 0; i <= kTestDataSize; i++) {
+      VerifyDefaultCF(i, snap);
+    }
+  }
 };
+constexpr int DataVisibilityTest::kTestDataSize;
```
riversand963, line 716: Nit: is this needed?

db/db_with_timestamp_basic_test.cc
```diff
@@ -590,8 +590,130 @@ TEST_F(DBBasicTestWithTimestamp, CompactDeletionWithTimestampMarkerToBottom) {
...
+    std::string value;
+    int ts;
+    std::string timestamp;
+    SequenceNumber seq_num;
+  };
+
+  constexpr static int kTestDataSize = 3;
+  TestData test_data_[kTestDataSize];
+
+  void PutTestData(int index, ColumnFamilyHandle* cfh = nullptr) {
```
riversand963, line 616: Nit: I am doubtful about the benefit of default argument here. Explicitly specifying `cfh` is clearer.

db/db_with_timestamp_basic_test.cc
```diff
@@ -590,8 +590,130 @@ TEST_F(DBBasicTestWithTimestamp, CompactDeletionWithTimestampMarkerToBottom) {
...
+    WriteOptions write_opts;
+    Slice ts_slice = test_data_[index].timestamp;
+    write_opts.timestamp = &ts_slice;
+
+    if (cfh == nullptr) {
+      ASSERT_OK(
+          db_->Put(write_opts, test_data_[index].key, test_data_[index].value));
+      const Snapshot* snap = db_->GetSnapshot();
+      test_data_[index].seq_num = snap->GetSequenceNumber();
+      if (index != 0) {
```
riversand963, line 627: Nit: `if(index > 0)`?

db/db_with_timestamp_basic_test.cc
```diff
@@ -590,8 +590,130 @@ TEST_F(DBBasicTestWithTimestamp, CompactDeletionWithTimestampMarkerToBottom) {
...
+        ASSERT_GT(test_data_[index].seq_num, test_data_[index - 1].seq_num);
+      }
+      db_->ReleaseSnapshot(snap);
+    } else {
+      ASSERT_OK(db_->Put(write_opts, cfh, test_data_[index].key,
+                         test_data_[index].value));
+    }
+  }
+
+  void AssertVisibility(int ts, SequenceNumber seq, Status statuses[]) {
```
riversand963, line 637: Nit: I am afraid our linter will complain about the usage of `statuses`.

db/db_with_timestamp_basic_test.cc
```diff
@@ -590,8 +590,130 @@ TEST_F(DBBasicTestWithTimestamp, CompactDeletionWithTimestampMarkerToBottom) {
...
+    std::vector<ColumnFamilyHandle*> cfs(kTestDataSize, cfh);
+    SequenceNumber seq =
+        snap ? snap->GetSequenceNumber() : kMaxSequenceNumber - 1;
+
+    // There're several MultiGet interfaces with not exactly the same
+    // implementations, query data with all of them.
+    auto keys = GetKeys();
+    std::vector<std::string> values;
+    auto s1 = db_->MultiGet(read_opts, cfs, keys, &values);
+    AssertVisibility(ts, seq, s1.data());
```
riversand963, line 672: Why not just `AssertVisibility(ts, seq, s1)`? Same for below.

db/db_with_timestamp_basic_test.cc
```diff
@@ -906,6 +1028,153 @@ TEST_F(DataVisibilityTest, RangeScanWithSnapshot) {
...
+  const Snapshot* snap2 = db_->GetSnapshot();
+  PutTestData(2);
+  VerifyDefaultCF();
+  VerifyDefaultCF(snap0);
+  VerifyDefaultCF(snap1);
+  VerifyDefaultCF(snap2);
+
+  db_->ReleaseSnapshot(snap0);
+  db_->ReleaseSnapshot(snap1);
+  db_->ReleaseSnapshot(snap2);
```
riversand963, line 1063: We need to `Close()` here because `test_cmp` will go out of scope.

db/db_with_timestamp_basic_test.cc
```diff
@@ -906,6 +1028,153 @@ TEST_F(DataVisibilityTest, RangeScanWithSnapshot) {
...
+  Options options = CurrentOptions();
+  const size_t kTimestampSize = Timestamp(0, 0).size();
+  TestComparator test_cmp(kTimestampSize);
+  options.comparator = &test_cmp;
+  DestroyAndReopen(options);
+
+  ColumnFamilyHandle* second_cf;
+  ColumnFamilyOptions cf_opts;
+  cf_opts.comparator = &test_cmp;
+  ASSERT_OK(db_->CreateColumnFamily(cf_opts, "second", &second_cf));
```
riversand963, line 1123: For multi-cf test, we can use `DBTestBase::CreateAndReopen`.

riversand963 (commented):
Thanks @jay-zhuang for working on this. I left a few comments.

db/db_impl/db_impl.cc
jay-zhuang: I feel a modern IDE should help to do that.
For example, I'm using CLion, it's able to auto display arguments name if it's constant:
image

db/db_with_timestamp_basic_test.cc
jay-zhuang: As it's a static variable, it has to be declared somewhere.

db/db_with_timestamp_basic_test.cc
jay-zhuang: if `cfh` is not specified, then it writes to default cf, I think it's consistent with our other APIs. Otherwise, we have to change all
```
PutTestData(1);
```
to:
```
PutTestData(1, nullptr);
```

db/db_impl/db_impl.cc
riversand963: The assumption of a modern IDE feels too strong.

db/db_with_timestamp_basic_test.cc
riversand963: hmm, for constexpr pod type, it seems unnecessary (verified with g++ 11).
```
#include <cstdio>

class TestClass {
 public:
  constexpr static int kConst =1;
};

class TestClass1 {
 public:
  const static int kConst;
};

// uncomment me
// const int TestClass1::kConst = 1;

int main() {
  fprintf(stdout, "%d %d\n", TestClass::kConst, TestClass1::kConst);
  return 0;
}
```

db/db_with_timestamp_basic_test.cc
jay-zhuang: okay, I changed all array to std::vector.

db/db_with_timestamp_basic_test.cc
jay-zhuang: s1 is a `std::vector`, the argument was an array, now it's changed to std::vector too.

db/db_with_timestamp_basic_test.cc
jay-zhuang: Cool, that would be much simpler.

db/db_impl/db_impl.cc
jay-zhuang: Updated.

Even though I'm fully convinced that it's a better style, but I don't have strong option about that either.

riversand963 (approved):

PR #7406, opened 2020-09-17
build: make it compile with @mode/win

xavierd:
While rocksdb can compile on both macOS and Linux with Buck, it couldn't be
compiled on Windows. The only way to compile it on Windows was with the CMake
build.

To keep the multi-platform complexity low, I've simply included all the Windows
bits in the TARGETS file, and added large #if blocks when not on Windows, the
same was done on the posix specific files.

Test Plan:
On my devserver:
  buck test //rocksdb/...
On Windows:
  buck build @mode/win //rocksdb/src:rocksdb_lib

### DIFF
--- a/port/win/env_win.cc
+++ b/port/win/env_win.cc
@@ -7,6 +7,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
+#if defined(OS_WIN)
+
 #include "port/win/env_win.h"
 #include "port/win/win_thread.h"
 #include <algorithm>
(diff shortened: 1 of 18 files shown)


port/win/env_win.cc
```diff
@@ -7,6 +7,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
+#if defined(OS_WIN)
```
pdillinger, line 10: I personally prefer putting compilation conditions inside the source file as much as possible, to simplify build scripts. If others disagree, speak up!

pdillinger (approved):
LGTM :)

PR #7412, opened 2020-09-18
Allow mutex to be released in GetAggregatedIntProperty

riversand963:
Current implementation holds db mutex while calling
`GetAggregatedIntProperty()`. For property kEstimateTableReadersMem,
this can be expensive, especially if the number of table readers is
high.
We can release and re-acquire db mutex if
property_info.need_out_of_mutex is true.

Test plan:
make check
COMPILE_WITH_TSAN=1 make check
Also test internally on a shadow host. Used bpf to verify the
excessively long db mutex holding no longer exists when applications
call GetApproximateMemoryUsageByType().

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -3135,17 +3135,17 @@ bool DBImpl::GetIntPropertyInternal(ColumnFamilyData* cfd,
     }
   } else {
     SuperVersion* sv = nullptr;
-    if (!is_locked) {
-      sv = GetAndRefSuperVersion(cfd);
-    } else {
-      sv = cfd->GetSuperVersion();
+    if (is_locked) {
+      mutex_.Unlock();
     }
+    sv = GetAndRefSuperVersion(cfd);
 
     bool ret = cfd->internal_stats()->GetIntPropertyOutOfMutex(
         property_info, sv->current, value);
 
-    if (!is_locked) {
-      ReturnAndCleanupSuperVersion(cfd, sv);
+    ReturnAndCleanupSuperVersion(cfd, sv);
+    if (is_locked) {
+      mutex_.Lock();
     }
 
     return ret;
@@ -3182,6 +3182,7 @@ bool DBImpl::GetAggregatedIntProperty(const Slice& property,
   }
 
   uint64_t sum = 0;
+  bool ret = true;
   {
     // Needs mutex to protect the list of column families.
     InstrumentedMutexLock l(&mutex_);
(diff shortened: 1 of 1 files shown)


riversand963:
Travis test failure is not related.

db/db_impl/db_impl.cc
```diff
@@ -3135,17 +3135,17 @@ bool DBImpl::GetIntPropertyInternal(ColumnFamilyData* cfd,
...
+    sv = GetAndRefSuperVersion(cfd);
 
     bool ret = cfd->internal_stats()->GetIntPropertyOutOfMutex(
         property_info, sv->current, value);
 
-    if (!is_locked) {
-      ReturnAndCleanupSuperVersion(cfd, sv);
+    ReturnAndCleanupSuperVersion(cfd, sv);
+    if (is_locked) {
+      mutex_.Lock();
```
jay-zhuang, line 3148: Just a question: could lock and unlock for every GetIntProp call have performance issue (assuming most of the GetIntProp calls are fast)? Could the lock/unlock be done at upper level, like `GetAggregatedIntProperty()`?

db/db_impl/db_impl.cc
```diff
@@ -3182,6 +3182,7 @@ bool DBImpl::GetAggregatedIntProperty(const Slice& property,
   }
 
   uint64_t sum = 0;
+  bool ret = true;
   {
     // Needs mutex to protect the list of column families.
     InstrumentedMutexLock l(&mutex_);
```
jay-zhuang, line 3188: Seems mutex is hold to make sure column families list is not changed. Is it still true if we unlock inside of the call?

riversand963: The intent is not to prevent column family set from changing. The purpose is to use `mutex_` to protect the data structure. I think it's OK for the column family set to change in the meantime.
Given this, the usage of mutex is not necessary, according to https://github.com/facebook/rocksdb/blob/master/db/column_family.h#L616

db/db_impl/db_impl.cc
riversand963: Other props have `need_out_mutex` as `false`, and they do not cause release/acquire of mutex. For `kEstimateTableReadersMem`, its `need_out_mutex` is true, and the operation itself can be expensive if holding mutex all the time.

db/db_impl/db_impl.cc
jay-zhuang: sorry, do you mean the mutex here should be removed?

riversand963: Oh, I see the confusion. We should hold the mutex before iterating the column family set, but we can release and acquire, as follows.
```
mutex_.Lock();
for (auto* cfd : column_family_set_) {
  cfd->Ref();
  mutex_.Unlock();
  mutex_.Lock();
  cfd->UnrefAndTryDelete();
}
```

jay-zhuang (approved):

riversand963:

PR #7414, opened 2020-09-18
Track WAL in MANIFEST: update WalMetadata for WAL syncing

cheng-chang:
There are some tricky behaviors related to WAL sync:

- When creating a WAL, the WAL might not be synced, if the WAL directory is not synced, the WAL file's metadata may not even be synced to disk, so during recovery, when listing the WAL directory, the WAL may not even show up.
- During each DB::Write, the WriteOption can control whether the WAL should be synced, so a WAL previously not synced on creation can be synced during Write.

For each `SyncWAL`, we'll track the synced status and the current WAL size. Previously, we only track the WAL size on closing.
During recovery, we check that the on-disk WAL size is >= the last synced size.

So this PR introduces `synced_size` and `closed` to `WalMetadata` for the above design update.

Test Plan:
- updated wal_edit_test
- updated version_edit_test

### DIFF
--- a/db/wal_edit.h
+++ b/db/wal_edit.h
@@ -36,7 +36,7 @@ class WalMetadata {
 
   bool IsSynced() const { return synced_; }
 
-  void SetSynced(bool synced) { synced_ = synced; }
+  void SetSynced() { synced_ = true; }
 
   bool HasSize() const { return size_bytes_ != kUnknownWalSize; }
 
@@ -49,7 +49,7 @@ class WalMetadata {
   constexpr static uint64_t kUnknownWalSize = 0;
 
   // Whether the WAL and WAL dir are synced.
-  bool synced_ = true;
+  bool synced_ = false;
 
   // Size of a closed WAL in bytes.
   uint64_t size_bytes_ = kUnknownWalSize;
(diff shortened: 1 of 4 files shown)


riversand963:
> There are some tricky behaviors related to WAL sync:
> 
> * When creating a WAL, the WAL might not be synced, if the WAL directory is not synced, the WAL file's metadata may not even be synced to disk, so during recovery, when listing the WAL directory, the WAL may not even show up.
> * During each DB::Write, the WriteOption can control whether the WAL should be synced, so a WAL previously not synced on creation can be synced during Write.
> * On closing a WAL, if the WAL is not synced, then the size of the WAL may grow afterwards because the some data hasn't been synced to disk.

After closing WAL, how can the size of the WAL keep growing?

> 
> So this PR introduces `synced` and `closed` to `WalMetadata`, then during recovery, we can differentiate among different combinations of `synced` and `closed`.
> 
> Test Plan:
> 
> * updated wal_edit_test
> * updated version_edit_test

cheng-chang:
@riversand963 I mean the WAL is closed, but not synced. So if OS syncs data asynchronously, then the WAL size will grow.

db/wal_edit.h
```diff
@@ -30,6 +32,12 @@ class WalMetadata {
 
   explicit WalMetadata(uint64_t size_bytes) : size_bytes_(size_bytes) {}
 
+  bool IsClosed() const { return HasSize(); }
```
riversand963, line 35: Can you remind me why `IsClosed()` is equivalent to `HasSize()`?

cheng-chang: left inline comments

riversand963: If application calls `SyncWAL()` or writes with `WriteOptions.sync=true`, then the WAL will have part of it synced. Recovery should check the actual WAL file size against this.
Things can be more complex if we allow log file number recycling.

cheng-chang: We will only write size of the WAL after closing it.

riversand963 (approved):
LGTM. Thanks @cheng-chang for the PR

PR #7418, opened 2020-09-21
Possible fix to flaky db_write_test

pdillinger:
Summary: Make the test robust to spurious wakeups on condition variable,
and clear sync points to ensure no use-after-free.

Test Plan: repeated runs on updated test, watch CircleCI for recurrence


mrambacher (approved):
LGTM.  You might want to add a comment why you are waiting on 1 and 6 writers.  I understand it is the number of threads, but it might be helpful to just add a comment saying that.

PR #7420, opened 2020-09-22
Provide users with option to opt-in to get corrupt data in logs/messages

akankshamahajan15:
Summary:  Add a new Option "allow_data_in_errors". When it's set by users, it allows them to opt-in to get error messages containing corrupted keys/values. Corrupt keys, values will be logged in the messages, logs, status etc. that will help users with the useful information regarding affected data. 
By default value is set false to prevent users data to be exposed in the messages.

Test Plan: 1. make check -j64
           2. Add a new test case

Reviewers:

Subscribers:

Tasks: T70153225

Tags:


db/builder.cc
```diff
@@ -179,7 +179,7 @@ Status BuildTable(
         &snapshots, earliest_write_conflict_snapshot, snapshot_checker, env,
         ShouldReportDetailedTime(env, ioptions.statistics),
         true /* internal key corruption is not ok */, range_del_agg.get(),
-        blob_file_builder.get());
+        blob_file_builder.get(), ioptions.log_corruption_errors);
```
ajkr, line 182: I wonder if we should use a more generic name, such as `redact_data_from_errors` or (inversely) `allow_data_in_errors`. Then, in addition to determining whether data can appear in logged errors, it can also determine whether data appears in `Status`es returned via our user API.

db/compaction/compaction_iterator.cc
```diff
@@ -271,12 +272,20 @@ void CompactionIterator::NextFromInput() {
...
+
       // If `expect_valid_internal_key_` is false, return the corrupted key
       // and let the caller decide what to do with it.
       // TODO(noetzli): We should have a more elegant solution for this.
       if (expect_valid_internal_key_) {
-        assert(!"Corrupted internal key not expected.");
-        status_ = Status::Corruption("Corrupted internal key not expected.");
+        if (log_corruption_errors_) {
+          std::string msg("Corrupted internal key not expected.");
+          msg.append("Corrupt key: " + ikey_.user_key.ToString(/*hex=*/true) +
```
ajkr, line 282: Should start with a space (i.e., " Corrupt key").

db/compaction/compaction_iterator.cc
```diff
@@ -271,12 +272,20 @@ void CompactionIterator::NextFromInput() {
...
       iter_stats_.num_input_corrupt_records++;
+
       // If `expect_valid_internal_key_` is false, return the corrupted key
       // and let the caller decide what to do with it.
       // TODO(noetzli): We should have a more elegant solution for this.
       if (expect_valid_internal_key_) {
-        assert(!"Corrupted internal key not expected.");
-        status_ = Status::Corruption("Corrupted internal key not expected.");
+        if (log_corruption_errors_) {
+          std::string msg("Corrupted internal key not expected.");
```
ajkr, line 281: It's fine to log this part of the message even if the option is set to false since it doesn't contain user data.

db/corruption_test.cc
```diff
@@ -115,6 +115,7 @@ class CorruptionTest : public testing::Test {
       }
       //if ((i % 100) == 0) fprintf(stderr, "@ %d of %d\n", i, n);
       Slice key = Key(i, &key_space);
+      printf("%s\n", std::string(key.data(), key.size()).c_str());
```
ajkr, line 118: Can probably remove.

db/compaction/compaction_iterator.cc
ajkr: Also can include `ikey_.type` for completeness.

db/memtable.cc
```diff
@@ -778,11 +779,16 @@ static bool SaveValue(void* arg, const char* entry) {
         return true;
       }
       default: {
-        std::string msg("Unrecognized value type: " +
-                        std::to_string(static_cast<int>(type)) + ". ");
-        msg.append("User key: " + user_key_slice.ToString(/*hex=*/true) + ". ");
-        msg.append("seq: " + std::to_string(seq) + ".");
-        *(s->status) = Status::Corruption(msg.c_str());
+        if (s->log_corruption_errors) {
+          std::string msg("Corrupted value not expected.");
```
ajkr, line 783: Same, I think this part of the message is fine to log regardless of the option's value.

ajkr (changes requested):
Looks great! Just a few minor things.

ajkr:
Oh, also we should mention in HISTORY.md "Public API Change" section that this new option is introduced and the logs/statuses will start having data redacted by default.

db/compaction/compaction_iterator.cc
akankshamahajan15: I was confused that it will anyway assert  if option is not set, so will that status be of any use in that case?

akankshamahajan15:
Thank you for the review. Addressed comments.

db/compaction/compaction_iterator.cc
ajkr: I think yes because the `assert()`s are excluded from optimized builds. But it's good to delete this assertion anyways, as returning a `Status::Corruption` is the proper way to report corrupt data.

ajkr (approved):
Awesome, thanks!

PR #7421, opened 2020-09-22
Remove invalid assertion in compaction_picker_universal.cc

anand1976:
The assertion checks that there is no overlap in sequence numbers across levels in universal compaction. However, this assumption doesn't hold when there is a delete triggered compaction or a trivial move, as they operate on a subset of a level.

Tests -
make check

### DIFF
--- a/db/compaction/compaction_picker_universal.cc
+++ b/db/compaction/compaction_picker_universal.cc
@@ -461,7 +461,6 @@ Compaction* UniversalCompactionBuilder::PickCompaction() {
 
 // validate that all the chosen files of L0 are non overlapping in time
 #ifndef NDEBUG
-  SequenceNumber prev_smallest_seqno = 0U;
   bool is_first = true;
 
   size_t level_index = 0U;
@@ -471,7 +470,6 @@ Compaction* UniversalCompactionBuilder::PickCompaction() {
       if (is_first) {
         is_first = false;
       }
-      prev_smallest_seqno = f->fd.smallest_seqno;
     }
     level_index = 1U;
   }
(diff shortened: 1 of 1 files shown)


ajkr (approved):
LGTM.

db/compaction/compaction_picker_universal.cc
```diff
@@ -483,16 +481,7 @@ Compaction* UniversalCompactionBuilder::PickCompaction() {
...
         is_first = false;
-      } else if (prev_smallest_seqno > 0) {
-        // A level is considered as the bottommost level if there are
-        // no files in higher levels or if files in higher levels do
-        // not overlap with the files being compacted. Sequence numbers
-        // of files in bottommost level can be set to 0 to help
-        // compression. As a result, the following assert may not hold
-        // if the prev_smallest_seqno is 0.
-        assert(prev_smallest_seqno > largest_seqno);
       }
```
jay-zhuang, line 484: Or should we change it to:

```
assert(largest_seqno == 0 || prev_smallest_seqno > largest_seqno);
```

anand1976: ```prev_smallest_seqno > largest_seqno``` wouldn't hold even for non-bottommost levels if there's a delete triggered compaction.

jay-zhuang: sorry, just a question, what is `delete triggered compaction`? is it `CompactionReason.kTtl`: 
https://github.com/facebook/rocksdb/blob/c268628c254add17efb35aa8244bd07d3a90c6ec/include/rocksdb/listener.h#L92-L93

ajkr: > delete triggered compaction

Formally, delete triggered compaction (DTC) is using this feature (https://github.com/facebook/rocksdb/blob/510c66fb75196de9412acf83637a63312e2e7d13/include/rocksdb/utilities/table_properties_collectors.h#L15-L19) to trigger a compaction.

But we also sometimes say DTC when referring to the underlying mechanism that supports it. That underlying mechanism "marks" files for compaction when they're generated and includes the mark in the file's manifest entry. The compaction picking algorithm can pick marked files for compaction regardless of everything else (e.g., whether the level score is large enough to need compaction).

PR #7424, opened 2020-09-22
Fix valgrind issues with configurable_test

mrambacher:
Valgrind was reporting a problem with the configurable_test in some GTEST code.  This problem was caused by using a std::function as a GTEST parameter.  This change changes the test to use a string as a function parameter (backed by a map) and fixes the valgrind issue.


pdillinger:
As discussed through private channels, this is not exactly elegant, but it uses human readable data for gtest parameters, which is at minimum useful for --gtest_list_tests and apparently avoids valgrind failures like this.

I verified this fixes the valgrind failure in configurable_test. Just fix merge issue on HISTORY.md.

pdillinger (approved):
👍

PR #7425, opened 2020-09-22
Add event listeners to RocksJava

alucarded:
Allows adding event listeners in RocksJava.

* Adds listeners getter and setter in `Options` and `DBOptions` classes.
* Adds `EventListener` Java interface and base class for implementing custom event listener callbacks - `AbstractEventListener`, which has an underlying native callback class implementing C++ `EventListener` class.
* `AbstractEventListener` class has mechanism for selectively enabling its callback methods in order to prevent invoking Java method if it is not implemented. This decreases performance cost in case only subset of event listener callback methods is needed - the JNI code for remaining "no-op" callbacks is not executed.
* The code is covered by unit tests in `EventListenerTest.java`, there are also tests added for setting/getting listeners field in `OptionsTest.java` and `DBOptionsTest.java`.


alucarded:
As regards fix to RTTI issue (avoid dynamic_cast) added in [the commit](https://github.com/facebook/rocksdb/pull/7425/commits/995a1c57782c443597a20cd078fbbb60c67e2e22):

We cannot static_cast from derived to base class pointer
and then reinterpret_cast from base to derived pointer.
See experiments such as: https://stackoverflow.com/a/60203479
https://hacksoflife.blogspot.com/2007/02/c-objects-part-3-multiple-inheritance.html
Docs:
https://timsong-cpp.github.io/cppwp/n4659/basic.compound#4
https://timsong-cpp.github.io/cppwp/n4659/class#7

Even if we redesign EventListenerJniCallback to have single inheritance (JniCallback would be a member class) we will not get standard-layout class (because of virtual methods in EventListener) and thus EventListener and EventListenerJniCallback would not be pointer-interconvertible.

adamretter (approved):
We have been through several rounds of internal review with @alucarded already here at Evolved Binary. I am happy this PR is now in a good shape.

Unless anyone has some suggestions about the RTTI issue for avoiding `dynamic_cast`, then I think it is good to merge.

@pdillinger @siying any comments?

pdillinger:
> Unless anyone has some suggestions about the RTTI issue for avoiding `dynamic_cast`, then I think it is good to merge.

If you're considering reinterpret_cast, I assume you know it's an EventListenerJniCallback object. If the jlong came from an EventListener*, wouldn't you want to reinterpret_cast from jlong back to EventListener* and static_cast from EventListener* to EventListenerJniCallback*?

alucarded:
@pdillinger It all makes sense now... For some reason I assumed I cannot downcast in this case, which is actually not the case. I am making the change.

java/src/main/java/org/rocksdb/EventListener.java
```diff
@@ -0,0 +1,332 @@
...
+   *
+   * Note that the this function must be implemented in a way such that
+   * it should not run for an extended period of time before the function
+   * returns. Otherwise, RocksDB may be blocked.
+   *
+   * @param db the database
+   * @param flushJobInfo the flush job info, contains data copied from
+   *     respective native structure.
+   */
+  void onFlushCompleted(final RocksDB db, final FlushJobInfo flushJobInfo);
```
darionyaphet, line 53: Great Work Do you mind move `onFlushBegin` before `onFlushCompleted ` ?

java/src/main/java/org/rocksdb/RocksObject.java
```diff
@@ -5,6 +5,8 @@
 
 package org.rocksdb;
 
+import java.util.Objects;
```
darionyaphet, line 8: why import this package?

pdillinger: or just say "unused import"

PR #7426, opened 2020-09-22
Fix/minimize mock_time_env.h dependencies

pdillinger:
Summary: (a) own copy of kMicrosInSecond
(b) out-of-line sync point code

Test Plan: FB internal

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -731,6 +731,7 @@ set(SOURCES
         table/table_factory.cc
         table/table_properties.cc
         table/two_level_iterator.cc
+        test_util/mock_time_env.cc
         test_util/sync_point.cc
         test_util/sync_point_impl.cc
         test_util/testutil.cc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -731,7 +731,6 @@ set(SOURCES
         table/table_factory.cc
         table/table_properties.cc
         table/two_level_iterator.cc
-        test_util/mock_time_env.cc
         test_util/sync_point.cc
         test_util/sync_point_impl.cc
         test_util/testutil.cc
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -731,6 +731,7 @@ set(SOURCES
         table/table_factory.cc
         table/table_properties.cc
         table/two_level_iterator.cc
+        test_util/mock_time_env.cc
```
mrambacher, line 734: Shouldn't this be in TEST_LIB_SOURCES and not LIB_SOURCES?

src.mk
```diff
@@ -178,6 +178,7 @@ LIB_SOURCES =                                                   \
   table/table_factory.cc                                        \
   table/table_properties.cc                                     \
   table/two_level_iterator.cc                                   \
+  test_util/mock_time_env.cc                                    \
```
mrambacher, line 181: TEST_LIB_SOURCES?

test_util/mock_time_env.h
```diff
@@ -62,29 +64,11 @@ class MockTimeEnv : public EnvWrapper {
   // TODO: this is a workaround for the different behavior on different platform
   // for timedwait timeout. Ideally timedwait API should be moved to env.
   // details: PR #7101.
-  void InstallTimedWaitFixCallback() {
```
mrambacher, line 65: Should the methods that use std::numeric_limits also be moved or should <limits> be included?

mrambacher (approved):
LGTM, with the exception of LIB_SOURCES

jay-zhuang (approved):

jay-zhuang (approved):
Thanks for the fix.

test_util/mock_time_env.h
pdillinger: env.h -> limits is sufficient

CMakeLists.txt
pdillinger: I was just copying from sync_point.cc. I'll try it and see if I have difficulty working out the public as well as internal dependencies.

jay-zhuang: Maybe `TESTUTIL_SOURCE` for cmake.

PR #7427, opened 2020-09-22
Add few unit test cases in ASSERT_STATUS_CHECKED build

akankshamahajan15:
Summary: Fix few test cases and add them in ASSERT_STATUS_CHECKED build.

Test Plan: 1.  ASSERT_STATUS_CHECKED=1 make -j48 check,
                 2. travis build for ASSERT_STATUS_CHECKED,
                 3. Without ASSERT_STATUS_CHECKED:  make check -j64, CircleCI build and travis build

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -1102,10 +1102,10 @@ Status ColumnFamilyData::RangesOverlapWithMemtables(
       super_version->mem->NewRangeTombstoneIterator(read_opts, read_seq);
   range_del_agg.AddTombstones(
       std::unique_ptr<FragmentedRangeTombstoneIterator>(active_range_del_iter));
-  super_version->imm->AddRangeTombstoneIterators(read_opts, nullptr /* arena */,
-                                                 &range_del_agg);
-
   Status status;
+  status = super_version->imm->AddRangeTombstoneIterators(
+      read_opts, nullptr /* arena */, &range_del_agg);
+  assert(status.ok());
   for (size_t i = 0; i < ranges.size() && status.ok() && !*overlap; ++i) {
     auto* vstorage = super_version->current->storage_info();
     auto* ucmp = vstorage->InternalComparator()->user_comparator();
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -1105,7 +1105,9 @@ Status ColumnFamilyData::RangesOverlapWithMemtables(
   Status status;
   status = super_version->imm->AddRangeTombstoneIterators(
       read_opts, nullptr /* arena */, &range_del_agg);
+  // AddRangeTombstoneIterators always return Status::OK.
   assert(status.ok());
+
   for (size_t i = 0; i < ranges.size() && status.ok() && !*overlap; ++i) {
     auto* vstorage = super_version->current->storage_info();
     auto* ucmp = vstorage->InternalComparator()->user_comparator();
--- a/db/compaction/compaction_job.cc
+++ b/db/compaction/compaction_job.cc
@@ -338,7 +338,6 @@ CompactionJob::CompactionJob(
 CompactionJob::~CompactionJob() {
   assert(compact_ == nullptr);
   ThreadStatusUtil::ResetThreadStatus();
-  io_status_.PermitUncheckedError();
 }
 
 void CompactionJob::ReportStartedCompaction(Compaction* compaction) {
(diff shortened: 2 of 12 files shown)


db/compaction/compaction_job.cc
```diff
@@ -1024,6 +1025,8 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
       RecordDroppedKeys(range_del_out_stats,
                         &sub_compact->compaction_job_stats);
     }
+    // Status is checked in while loop.
```
mrambacher, line 1028: Since I agree that the status is checked in the while loop, I am wondering why these lines are necessary?

db/compaction/compaction_job.cc
```diff
@@ -1640,6 +1643,7 @@ void CompactionJob::CleanupCompaction() {
         TableCache::Evict(table_cache_.get(), out.meta.fd.GetNumber());
       }
     }
+    sub_compact.io_status.PermitUncheckedError();
```
mrambacher, line 1646: Is there a reason not to do this in the SubCompactionState destructor instead?  If not, please add a comment why this is okay here (because the sub_compact is about to be deleted anyway...)

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1259,6 +1261,7 @@ void DBImpl::NotifyOnCompactionBegin(ColumnFamilyData* cfd, Compaction* c,
     BuildCompactionJobInfo(cfd, c, st, job_stats, job_id, current, &info);
     for (auto listener : immutable_db_options_.listeners) {
       listener->OnCompactionBegin(this, info);
+      info.status.PermitUncheckedError();
```
mrambacher, line 1264: This should probably be outside of the for loop

db/merge_helper.cc
```diff
@@ -140,15 +140,17 @@ Status MergeHelper::MergeUntil(InternalIterator* iter,
   ParsedInternalKey orig_ikey;
   bool succ = ParseInternalKey(original_key, &orig_ikey);
   assert(succ);
+  Status s;
```
mrambacher, line 143: Are these changes necessary?

db/column_family.cc
```diff
@@ -1102,10 +1102,10 @@ Status ColumnFamilyData::RangesOverlapWithMemtables(
       super_version->mem->NewRangeTombstoneIterator(read_opts, read_seq);
   range_del_agg.AddTombstones(
       std::unique_ptr<FragmentedRangeTombstoneIterator>(active_range_del_iter));
-  super_version->imm->AddRangeTombstoneIterators(read_opts, nullptr /* arena */,
-                                                 &range_del_agg);
-
   Status status;
+  status = super_version->imm->AddRangeTombstoneIterators(
```
mrambacher, line 1106: Since AddRangeTombstoneIterators always returns OK, maybe just change it to return a void instead?

mrambacher (commented):
I cannot tell from your test plan, but you also need to run the tests without ASSERT_STATUS_CHECKED set, to verify the tests pass that way as well.

db/column_family.cc
akankshamahajan15: Yes, I wasn't sure if I should  change the API's (return type) for this change.

db/merge_helper.cc
akankshamahajan15: It gives error because Status s is defined but its status is not checked in the flow where direct status is returned  instead of s. So to make it consistent I assigned the error messages to s and returned s.

akankshamahajan15:
> I cannot tell from your test plan, but you also need to run the tests without ASSERT_STATUS_CHECKED set, to verify the tests pass that way as well.

Thank you for pointing that out. I updated the test plan.

db/compaction/compaction_job.cc
akankshamahajan15: I checked again, it's not needed so removed it.

akankshamahajan15:
build-windows-4 failure not related to this change.

db/column_family.cc
mrambacher: Can you add a comment or PR that says the method always returns OK so it can be addressed sometime later?

mrambacher (approved):
If it passes the tests, it LGTM.  The other comments I had are fairly minor.

PR #7431, opened 2020-09-23
Add and fix clang -Wshift-sign-overflow

pdillinger:
Summary: This option is apparently used by some teams within Facebook
(internal ref T75998621)

Test Plan: USE_CLANG=1 make check before (fails) and after


jay-zhuang (approved):
👍

PR #7433, opened 2020-09-24
[circleci] Remove a gtest-parallel workaround

jay-zhuang:
As the issue is fixed in official repo:
https://github.com/google/gtest-parallel/pull/79

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -336,7 +336,7 @@ jobs:
             sed -i 's/[[:space:]]*$//; s/ / \.\//g; s/.*/.\/&/' /tmp/test_list
             cat /tmp/test_list
             export TEST_TMPDIR=/tmp/rocksdb_test_tmp
-            /usr/bin/python ../gtest-parallel/gtest-parallel $(</tmp/test_list) --output_dir=/tmp | cat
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -337,7 +336,7 @@ jobs:
             sed -i 's/[[:space:]]*$//; s/ / \.\//g; s/.*/.\/&/' /tmp/test_list
             cat /tmp/test_list
             export TEST_TMPDIR=/tmp/rocksdb_test_tmp
-            /usr/bin/python ../gtest-parallel/gtest-parallel $(</tmp/test_list) --output_dir=/tmp --print_task_status_continuously
+            /usr/bin/python ../gtest-parallel/gtest-parallel $(</tmp/test_list) --output_dir=/tmp | cat
```
ajkr, line 339: This doesn't matter, just for my understanding -- do we lose color support by tricking the python program into thinking it's not a tty?

ajkr (approved):
LGTM! I figured out the reason for `| cat` by reading the linked issue, but we could add a comment above where it's used for convenience.

.circleci/config.yml
jay-zhuang: I'll add comment on that.
We won't lose color support with `cat`, it's just to output job status continuously.
Without that, it won't print anything while running on circleci output, because the gtest_parallel script try to re-use the same line for status update, and on web-ui it doesn't print anything.

PR #7434, opened 2020-09-24
Introduce a helper method UncompressData

ltamasi:
Summary:
The patch introduces a helper method in `util/compression.h` called `UncompressData`
that dispatches calls to the correct uncompression method based on type, and changes
`UncompressBlockContentsForCompressionType` and `Benchmark::Uncompress` in
`db_bench` so they are implemented in terms of the new method. This eliminates
some code duplication. (`Benchmark::Compress` is also updated to use the previously
introduced `CompressData` helper.)

In addition, the patch brings the implementation of `Snappy_Uncompress` into sync with
the other uncompression methods by making the method compute the buffer size and allocate
the buffer itself. Finally, the patch eliminates some potentially risky back-and-forth conversions
between various unsigned and signed integer types by exposing the size of the allocated buffer
as a `size_t` instead of an `int`.

Test Plan:
`make check`
`./db_bench -benchmarks=compress,uncompress --compression_type ...`

### DIFF
--- a/port/win/xpress_win.cc
+++ b/port/win/xpress_win.cc
@@ -129,10 +129,9 @@ bool Compress(const char* input, size_t length, std::string* output) {
 }
 
 char* Decompress(const char* input_data, size_t input_length,
-  int* decompress_size) {
-
+                 size_t* uncompressed_size) {
   assert(input_data != nullptr);
-  assert(decompress_size != nullptr);
+  assert(uncompressed_size != nullptr);
 
   if (input_length == 0) {
     return nullptr;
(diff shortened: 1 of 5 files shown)


mrambacher (commented):
While I support what the general idea here of cleaning up and centralizing the Compressor interfaces, I wonder if the approach taken by @lucagiac81 in #6717 is not a better approach.  If we skip the portions of "pluggable compression" proposed in #6717, can we reach a consensus on the interface for Compression and Decompression and move the various implementations that are "built in" into this model?  If so, it should be possible to later add in support for pluggable compression algorithms and both PRs will be satisfied (and the code cleaner and easier to understand).

lucagiac81:
The first commit in #6717 cleans up the compression code as well. It takes it one step further by defining a Compressor interface and turning the existing algorithms into Compressor subclasses. It also introduces a CompressorRegistry to keep a list of available compressors in one place. The interface and registry make it easier to support new compressors, including dynamically loaded ones (covered by the other commits in the same PR).
To turn snappy into a Compressor, I made a similar change and moved buffer size computation and allocation into SnappyCompressor::Uncompress.

Following @mrambacher's proposal, we can make the first commit of #6717 a separate PR, leaving support for custom/dynamically loaded compressors for a subsequent PR. If you prefer to merge #7434 first, it's still easy to update #6717 to use the new UncompressData and modified Snappy_Uncompress functions.

riversand963 (approved):
LGTM.

ltamasi:
Thanks @lucagiac81 . Seems to me there isn't any significant conflict between the two PRs, so I'm going to go ahead and merge this cleanup.

PR #7435, opened 2020-09-24
EnableFileDeletions only read field while holding mutex

pdillinger:
Summary: Possible fix for a TSAN issue reported in EnableFileDeletions.
disable_delete_obsolete_files_ should only be accessed holding the db
mutex, but for logging it was being accessed outside holding the mutex,
now fixed.

Test Plan: existing tests, watch for recurrence

### DIFF
--- a/db/db_impl/db_impl_files.cc
+++ b/db/db_impl/db_impl_files.cc
@@ -58,7 +58,7 @@ Status DBImpl::EnableFileDeletions(bool force) {
   // Job id == 0 means that this is not our background process, but rather
   // user thread
   JobContext job_context(0);
-  int counter;
+  int saved_counter;  // initialize on all paths
   {
     InstrumentedMutexLock l(&mutex_);
     if (force) {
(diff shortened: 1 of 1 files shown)


db/db_impl/db_impl_files.cc
```diff
@@ -58,7 +58,7 @@ Status DBImpl::EnableFileDeletions(bool force) {
   // Job id == 0 means that this is not our background process, but rather
   // user thread
   JobContext job_context(0);
-  bool file_deletion_enabled = false;
+  int counter;
```
mrambacher, line 61: I know it does not matter, but can you either initialize this to 0 or put a comment that says it is initialized in the lock block.  My natural reaction was to think it was not guaranteed to be initialized.

mrambacher (approved):
Took a bit to figure out the case you were actually fixing.  Once I did, LGTM :-)

Zenix27 (approved):

db/db_impl/db_impl_files.cc
pdillinger: Because we compile with warnings as errors, not initializing it enforces that it be set on all paths (leading to use). I'll add a comment.

ltamasi (approved):

PR #7436, opened 2020-09-25
Genericize and clean up FastRange

pdillinger:
Summary: A generic algorithm in progress depends on a templatized
version of fastrange, so this change generalizes it and renames
it to fit our style guidelines, FastRange32, FastRange64, and now
FastRangeGeneric.

Test Plan: added a few more test cases

### DIFF
--- a/util/fastrange.h
+++ b/util/fastrange.h
@@ -1,0 +1,99 @@
+//  Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+// fastrange/FastRange: A faster alternative to % for mapping a hash value
+// to an arbitrary range. See https://github.com/lemire/fastrange
+//
+// Generally recommended are FastRange32 for mapping results of 32-bit
+// hash functions and FastRange64 for mapping results of 64-bit hash
+// functions. FastRange is less forgiving than % if the input hashes are
+// not well distributed over the full range of the type (32 or 64 bits).
+//
+// Also included is a templated implementation FastRangeGeneric for use
+// in generic algorithms, but not otherwise recommended because of
+// potential ambiguity. Unlike with %, it is critical to use the right
+// FastRange variant for the output size of your hash function.
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <type_traits>
+
+#ifdef TEST_UINT128_COMPAT
+#undef HAVE_UINT128_EXTENSION
+#endif
+
+namespace ROCKSDB_NAMESPACE {
+
+namespace detail {
+
+// Using a class template to support partial specialization
+template <typename Hash, typename Range>
+struct FastRangeGenericImpl {
+  // only reach this on no supported specialization
+};
+
+template <typename Range>
+struct FastRangeGenericImpl<uint32_t, Range> {
+  static inline Range Fn(uint32_t hash, Range range) {
+    static_assert(std::is_unsigned<Range>::value, "must be unsigned");
+    static_assert(sizeof(Range) <= sizeof(uint32_t),
+                  "cannot be larger than hash (32 bits)");
+
+    uint64_t product = uint64_t{range} * hash;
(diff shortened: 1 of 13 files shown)


util/math.h
```diff
@@ -6,11 +6,13 @@
...
 
 #include <assert.h>
-#include <stdint.h>
 #ifdef _MSC_VER
 #include <intrin.h>
 #endif
 
+#include <cstdint>
+#include <type_traits>
+
```
jay-zhuang, line 15: Seems these includes are not needed.

util/fastrange.h
```diff
@@ -0,0 +1,99 @@
...
+    tmp >>= 32;
+    tmp += uint64_t{range64 & 0xffffFFFF} * uint64_t{hash >> 32};
+    // Avoid overflow: first add lower 32 of tmp2, and later upper 32
+    uint64_t tmp2 = uint64_t{range64 >> 32} * uint64_t{hash & 0xffffFFFF};
+    tmp += static_cast<uint32_t>(tmp2);
+    tmp >>= 32;
+    tmp += (tmp2 >> 32);
+    tmp += uint64_t{range64 >> 32} * uint64_t{hash >> 32};
+    return static_cast<Range>(tmp);
+#endif
```
jay-zhuang, line 77: Why not just call https://github.com/facebook/rocksdb/blob/c4d8838a2b54cb364ee3f05d86c8f980ac59bbc1/util/math128.h#L163

util/fastrange.h
```diff
@@ -0,0 +1,99 @@
...
+
+// The most popular / convenient / recommended variants:
+// size_t is standard for mapping to things in memory
+inline size_t FastRange64(uint64_t hash, size_t range) {
+  return FastRangeGeneric(hash, range);
+}
+
+inline uint32_t FastRange32(uint32_t hash, uint32_t range) {
+  return FastRangeGeneric(hash, range);
+}
```
jay-zhuang, line 97: Then why not just remove them and use `FastRangeGeneric` everywhere?

jay-zhuang (approved):
LGTM. Just a few minor comments.

util/fastrange.h
pdillinger: I want fastrange to be available with a pretty "lean" header because I want it to be widely available in the app without really thinking about it. There is many places where you want to map a hash value to an arbitrary range.

math128 seems pretty special-purpose to me, so I don't want to pull it in all over the app just for fastrange.

I generally frown upon code duplication myself, but this code should require no maintenance. It expresses what we want to a broad set of compilers.

util/fastrange.h
pdillinger: Good question, and I'll add a comment about that.

If the hash is only 32 bits, it might be put in a type like unsigned long, which may be 32 or 64 bits, without worrying much about it. Besides, with %, it doesn't really matter if you overshoot the size of your type. But fastrange gives extremely bad answers if you accidentally feed 32-bit hash values to a 64-bit fastrange. And a mistake like this might not be noticed in unit tests because good hashing is generally only needed for good performance, not correctness. Plus it could be a platform-dependent mistake (though probably appearing on Linux when it might occur anywhere).

Example:

    // Passes
    EXPECT_EQ(FastRangeGeneric(static_cast<unsigned>(0x80000000), uint16_t{12468}),
              uint16_t{6234});
    // Fails on Linux
    EXPECT_EQ(FastRangeGeneric(static_cast<unsigned long>(0x80000000), uint16_t{12468}),
              uint16_t{6234});

Because of this danger, I strongly advise writing "32" or "64" for the variant appropriate for the hash size, and only use the generic version where it's really needed.

util/fastrange.h
pdillinger: I thought about the two sharing a "multiply two 64-bit to upper and lower 64 bit outputs" function, but that's asking more of the inliner and optimizer to get rid of unnecessary code where we take the address of variables, etc. Call me paranoid, but at least on key platforms, fastrange has to be inner-loop single(ish)-instruction fast.

util/math.h
pdillinger: I just assume some compiler will want them:

https://en.cppreference.com/w/cpp/types/is_integral
https://en.cppreference.com/w/cpp/types/integer

PR #7438, opened 2020-09-26
On ARM platform, use yield op to relax CPU. See issue 7376

wolfkdy:
see https://github.com/facebook/rocksdb/issues/7376.
The `wfe` op on ARM platform is not suitable to relax CPU. Use `yield` op.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,9 @@
 ### Bug Fixes
 * Use thread-safe `strerror_r()` to get error messages.
 
+### Performance Improvements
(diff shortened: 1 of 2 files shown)


riversand963 (approved):

riversand963 (approved):
Per https://github.com/facebook/folly/blob/master/folly/portability/Asm.h#L42.

adamretter (commented):
@jay-zhuang On paper this looks fine. I am not super familiar with the arm instruction set. Do you think it would be worth me compiling and testing this first on our arm64 machine?

PR #7442, opened 2020-09-27
fix dummy collector's name

darionyaphet:


ajkr (changes requested):
Some tests are failing. I think "table/table_test.cc" line 1502 has an assertion with the old naming.

darionyaphet:
> Some tests are failing. I think "table/table_test.cc" line 1502 has an assertion with the old naming.

@ajkr  thanks & fixed

ajkr (approved):
LGTM, thanks!

PR #7443, opened 2020-09-28
Range Locking: Allow different LockManagers, add Range Lock definitions

spetrunia:
This PR has two commits:
1.  Modify the code to allow different Lock Managers (of any kind) to be used.  It is implied that a LockManager uses its own custom LockTracker.
2.  Add definitions for Range Locking (class Endpoint and GetRangeLock() function.

@cheng-chang, is this what you've had in mind (should the PR have both item 1 and item 2?)

### DIFF
--- a/include/rocksdb/utilities/transaction.h
+++ b/include/rocksdb/utilities/transaction.h
@@ -24,9 +24,81 @@ using TransactionName = std::string;
 
 using TransactionID = uint64_t;
 
-// An endpoint for a range of keys.
+/*
+  class Endpoint allows to define prefix ranges.
+
+  Prefix ranges are introduced below.
+
+  == Basic Ranges ==
+  Let's start from basic ranges. Key Comparator defines ordering of rowkeys.
+  Then, one can specify finite closed ranges by just providing rowkeys of their
+  endpoints:
+
+    lower_endpoint <= X <= upper_endpoint
+
+  However our goal is to provide a richer set of endpoints. Read on.
+
+  == Lexicographic ordering ==
+  A lexicographic (or dictionary) ordering satisfies these criteria: If there
+  are two keys in form
+    key_a = {prefix_a, suffix_a}
+    key_b = {prefix_b, suffix_b}
+  and
+    prefix_a < prefix_b
+  then
+    key_a < key_b.
+
+  == Prefix ranges ==
+  With lexicographic ordering, one may want to define ranges in form
+
+     "prefix is $PREFIX"
+
+  which translates to a range in form
+
+    {$PREFIX, -infinity} < X < {$PREFIX, +infinity}
+
+  where -infinity will compare less than any possible suffix, and +infinity
+  will compare as greater than any possible suffix.
+
+  class Endpoint allows to define these kind of rangtes.
+
+  == Notes ==
+  BytewiseComparator and ReverseBytewiseComparator produce lexicographic
+  ordering.
+
+  The row comparison function is able to compare key prefixes. If the data
+  domain includes keys A and B, then the comparison function is able to compare
+  equal-length prefixes:
+
+    min_len= min(byte_length(A), byte_length(B));
+    cmp(Slice(A, min_len), Slice(B, min_len));  // this call is valid
+
+  == Other options ==
+  As far as MyRocks is concerned, the alternative to prefix ranges would be to
+  support both open (non-inclusive) and closed (inclusive) range endpoints.
+*/
+
 class Endpoint {
-  // TODO
+ public:
+  Slice slice;
+
+  /*
+    true  : the key has a "+infinity" suffix. A suffix that would compare as
+            greater than any other suffix
+    false : otherwise
+  */
+  bool inf_suffix;
+
+  explicit Endpoint(const Slice& slice_arg, bool inf_suffix_arg = false)
+      : slice(slice_arg), inf_suffix(inf_suffix_arg) {}
+
+  explicit Endpoint(const char* s, bool inf_suffix_arg = false)
+      : slice(s), inf_suffix(inf_suffix_arg) {}
+
+  Endpoint(const char* s, size_t size, bool inf_suffix_arg = false)
+      : slice(s, size), inf_suffix(inf_suffix_arg) {}
+
+  Endpoint() : inf_suffix(false) {}
 };
 
 // Provides notification to the caller of SetSnapshotOnNextOperation when
@@ -280,6 +352,12 @@ class Transaction {
       pinnable_val->PinSelf();
       return s;
     }
+  }
+
+  // Get a range lock on [start_endpoint; end_endpoint].
+  virtual Status GetRangeLock(ColumnFamilyHandle*, const Endpoint&,
+                              const Endpoint&) {
+    return Status::NotSupported();
   }
 
   virtual Status GetForUpdate(const ReadOptions& options, const Slice& key,
--- a/include/rocksdb/utilities/transaction_db.h
+++ b/include/rocksdb/utilities/transaction_db.h
@@ -31,6 +31,98 @@ enum TxnDBWritePolicy {
 
 const uint32_t kInitialMaxDeadlocks = 5;
 
+class LockManager;
+struct RangeLockInfo;
+
+// A lock manager handle
+// The workflow is as follows:
+//  * Use a factory method (like NewRangeLockManager()) to create a lock
+//    manager and get its handle.
+//  * A Handle for a particular kind of lock manager will have extra
+//    methods and parameters to control the lock manager
+//  * Pass the handle to RocksDB in TransactionDBOptions::lock_mgr_handle. It
+//    will be used to perform locking.
+class LockManagerHandle {
+ public:
+  // PessimisticTransactionDB will call this to get the Lock Manager it's going
+  // to use.
+  virtual LockManager* getLockManager() = 0;
+
+  virtual ~LockManagerHandle() {}
+};
+
+// Same as class Endpoint, but use std::string to manage the buffer allocation
+struct EndpointWithString {
+  std::string slice;
+  bool inf_suffix;
+};
+
+struct RangeDeadlockInfo {
+  TransactionID m_txn_id;
+  uint32_t m_cf_id;
+  bool m_exclusive;
+
+  EndpointWithString m_start;
+  EndpointWithString m_end;
+};
+
+struct RangeDeadlockPath {
+  std::vector<RangeDeadlockInfo> path;
+  bool limit_exceeded;
+  int64_t deadlock_time;
+
+  explicit RangeDeadlockPath(std::vector<RangeDeadlockInfo> path_entry,
+                             const int64_t& dl_time)
+      : path(path_entry), limit_exceeded(false), deadlock_time(dl_time) {}
+
+  // empty path, limit exceeded constructor and default constructor
+  explicit RangeDeadlockPath(const int64_t& dl_time = 0, bool limit = false)
+      : path(0), limit_exceeded(limit), deadlock_time(dl_time) {}
+
+  bool empty() { return path.empty() && !limit_exceeded; }
+};
+
+// A handle to control RangeLockManager (Range-based lock manager) from outside
+// RocksDB
+class RangeLockManagerHandle : public LockManagerHandle {
+ public:
+  // Total amount of lock memory to use (per column family)
+  virtual int SetMaxLockMemory(size_t max_lock_memory) = 0;
+  virtual size_t GetMaxLockMemory() = 0;
+
+  using RangeLockStatus =
+      std::unordered_multimap<ColumnFamilyId, RangeLockInfo>;
+
+  virtual RangeLockStatus GetRangeLockStatusData() = 0;
+
+  class Counters {
+   public:
+    // Number of times lock escalation was triggered (for all column families)
+    uint64_t escalation_count;
+
+    // How much memory is currently used for locks (total for all column
+    // families)
+    uint64_t current_lock_memory;
+  };
+
+  // Get the current counter values
+  virtual Counters GetStatus() = 0;
+
+  // Functions for range-based Deadlock reporting.
+  virtual std::vector<RangeDeadlockPath> GetRangeDeadlockInfoBuffer() = 0;
+  virtual void SetRangeDeadlockInfoBufferSize(uint32_t target_size) = 0;
+
+  virtual ~RangeLockManagerHandle() {}
+};
+
+// A factory function to create a Range Lock Manager. The created object should
+// be:
+//  1. Passed in TransactionDBOptions::lock_mgr_handle to open the database in
+//     range-locking mode
+//  2. Used to control the lock manager when the DB is already open.
+RangeLockManagerHandle* NewRangeLockManager(
+    std::shared_ptr<TransactionDBMutexFactory> mutex_factory);
+
 struct TransactionDBOptions {
   // Specifies the maximum number of keys that can be locked at the same time
   // per column family.
@@ -91,6 +183,10 @@ struct TransactionDBOptions {
   // logic in myrocks. This hack of simply not rolling back merge operands works
   // for the special way that myrocks uses this operands.
   bool rollback_merge_operands = false;
+
+  // nullptr means use default lock manager.
+  // Other value means the user provides a custom lock manager.
+  std::shared_ptr<LockManagerHandle> lock_mgr_handle;
 
   // If true, the TransactionDB implementation might skip concurrency control
   // unless it is overridden by TransactionOptions or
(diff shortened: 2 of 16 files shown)


include/rocksdb/utilities/transaction_db.h
```diff
@@ -31,6 +31,49 @@ enum TxnDBWritePolicy {
 
 const uint32_t kInitialMaxDeadlocks = 5;
 
+class BaseLockMgr;
```
cheng-chang, line 34: nit: can we use "Manager" consistently?

utilities/transactions/lock/lock_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+  void Resize(uint32_t target_size);
+  std::vector<DeadlockPath> PrepareBuffer();
+};
+
+//
+// Base class for Point and Range-based lock manager.
+//
+class BaseLockMgr {
+ public:
+  virtual LockTrackerFactory* getLockTrackerFactory() = 0;
```
cheng-chang, line 46: Is a `LockTrackerFactory` `new`ed each time when calling `getLockTrackerFactory`? If so, can we call it `NewLockTrackerFactory` and leave a note indicating that the caller owns the returned factory?

cheng-chang: Seems like a static instance is returned each time, then can we comment that the caller does not own the returned object?

cheng-chang: nit: `Get` for consistency?

utilities/transactions/lock/lock_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+                      const LockTracker& tracker, Env* env) = 0;
+  virtual void UnLock(PessimisticTransaction* txn, uint32_t column_family_id,
+                      const std::string& key, Env* env) = 0;
+
+  // Resize the deadlock info buffer
+  virtual void Resize(uint32_t) = 0;
+  virtual std::vector<DeadlockPath> GetDeadlockInfoBuffer() = 0;
+
+  // TransactionDB will call this at start
+  virtual void init(TransactionDB*){};
```
cheng-chang, line 63: nit: capitalize `Init` for consistency?

utilities/transactions/lock/point/point_lock_tracker.h
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
```
cheng-chang, line 1: I assume this file is only moved from `lock/lock_tracker.h` without modification?

utilities/transactions/lock/point/point_lock_tracker.cc
```diff
@@ -0,0 +1,272 @@
+// Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
```
cheng-chang, line 1: I assume this file is only moved from `lock/lock_tracker.cc` without modification?

utilities/transactions/lock/lock_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+// Base class for Point and Range-based lock manager.
+//
+class BaseLockMgr {
+ public:
+  virtual LockTrackerFactory* getLockTrackerFactory() = 0;
+
+  virtual void AddColumnFamily(const ColumnFamilyHandle* cfh) = 0;
+  virtual void RemoveColumnFamily(const ColumnFamilyHandle* cfh) = 0;
+
+  virtual Status TryLock(PessimisticTransaction* txn, uint32_t column_family_id,
```
cheng-chang, line 51: can we consistently use the `ColumnFamilyId` typedef instead of `uint32_t`? also applies to other occurrences of cf id.

utilities/transactions/lock/point/point_lock_tracker.h
```diff
@@ -0,0 +1,93 @@
...
+
+ private:
+  TrackedKeys tracked_keys_;
+};
+
+class PointLockTrackerFactory : public LockTrackerFactory {
+ public:
+  LockTracker* Create() const override { return new PointLockTracker; }
+
+  static PointLockTrackerFactory instance;
```
cheng-chang, line 89: can we make this private and have a public `Get` method instead?

include/rocksdb/utilities/transaction_db.h
```diff
@@ -31,6 +31,49 @@ enum TxnDBWritePolicy {
...
+  virtual ~LockManagerHandle(){};
+};
+
+// A handle to control RangeLockMgr (Range-based lock manager) from outside
+// RocksDB
+class RangeLockMgrHandle : public LockManagerHandle {
+ public:
+  virtual int set_max_lock_memory(size_t max_lock_memory) = 0;
+
+  class Counters {
```
cheng-chang, line 59: this feels like specific to certain range lock implementations?

utilities/transactions/lock/point/point_lock_mgr.h
```diff
@@ -17,41 +17,28 @@
...
   bool m_exclusive;
   std::string m_waiting_key;
 };
 
-class Slice;
 class PessimisticTransactionDB;
 
-class TransactionLockMgr {
+// Point lock manager
+class TransactionLockMgr : public BaseLockMgr {
```
cheng-chang, line 41: can we rename it to `PointLockManager` to be consistent with namings like `PointLockTracker`?

utilities/transactions/pessimistic_transaction.h
```diff
@@ -116,6 +116,12 @@ class PessimisticTransaction : public TransactionBaseImpl {
 
   int64_t GetDeadlockDetectDepth() const { return deadlock_detect_depth_; }
 
+  virtual Status GetRangeLock(ColumnFamilyHandle*, const Endpoint&,
+                              const Endpoint&) override {
+    // This is a placeholder for Range Locking implementation
+    return Status::NotSupported();
```
cheng-chang, line 122: Ideally, there will be a range lock manager interface, and the implementation here will be based on that interface. 

In current design, will this method's implementation be bounded to a specific implementation of range lock?

utilities/transactions/pessimistic_transaction_db.cc
```diff
@@ -30,17 +31,33 @@ PessimisticTransactionDB::PessimisticTransactionDB(
...
   assert(db_impl_ != nullptr);
   info_log_ = db_impl_->GetDBOptions().info_log;
 }
 
+void PessimisticTransactionDB::init_lock_manager() {
+  if (txn_db_options_.lock_mgr_handle) {
+    // A custom lock manager was provided in options
+    auto mgr = txn_db_options_.lock_mgr_handle->getLockManager();
+    lock_mgr_ =
+        std::shared_ptr<BaseLockMgr>(txn_db_options_.lock_mgr_handle, mgr);
```
cheng-chang, line 45: would `reset(mgr)` work here?

cheng-chang (commented):
Thanks @spetrunia for this PR. 

The current `RangeLockManagerHandle` looks too general to me. Although we have `BaseLockMgr` interface, we don't have an interface for `RangeLockMgr`, and the `LockRange` method in `Transaction` will depend on the implementation of range lock.

I'm thinking can we have a general `RangeLockManager` interface, and implement `Transaction::LockRange` with that interface. Then in https://github.com/facebook/rocksdb/pull/7506, we just implement the interface with toku's range tree?

cheng-chang:
I tried to abstract out the LockManager interface at https://github.com/facebook/rocksdb/pull/7532, but that's just the first step, when implementing the range lock, updates to the interface may be needed.

include/rocksdb/utilities/transaction_db.h
```diff
@@ -31,6 +31,49 @@ enum TxnDBWritePolicy {
...
+   public:
+    uint64_t escalation_count;
+    uint64_t current_lock_memory;
```
cheng-chang, line 62: Are the concept of lock memory and escalation count are specific for the range lock implementation based on range tree?

utilities/transactions/lock/lock_mgr.h
```diff
@@ -0,0 +1,71 @@
...
+      : paths_buffer_(n_latest_dlocks), buffer_idx_(0) {}
+  void AddNewPath(DeadlockPath path);
+  void Resize(uint32_t target_size);
+  std::vector<DeadlockPath> PrepareBuffer();
+};
+
+//
+// Base class for Point and Range-based lock manager.
+//
+class BaseLockMgr {
```
cheng-chang, line 44: This interface feels like for point lock only? E.g. there is no API like `TryLockRange`?

include/rocksdb/utilities/transaction_db.h
spetrunia: `current_lock_memory` is the total amount of memory currently used for lock data structures. It is applicable even to Point Lock Manager (although the Point Lock Manager doesn't report the number currently). 

for `escalation_count`... it may depend. One of the goals of the Range-based Lock Manager is that it can naturally support  Lock Escalation, so I guess the answer is more "yes" than "no".   One can imagine a range-based Lock Manager that doesn't support lock escalation but it won't be as useful.

utilities/transactions/lock/lock_mgr.h
spetrunia: Yes. 

Yes. The reasons for this is that class `PessimisticTransactionDB` has several places where it "knows" which kind of lock manager it is using and so it calls to `RangeLockMgr` : 

```c++
void PessimisticTransactionDB::init_lock_manager() {
    range_lock_mgr_ = lock_mgr_->getRangeLockMgr();
```

```c++
Status PessimisticTransactionDB::TryRangeLock(PessimisticTransaction* txn,
                                              uint32_t cfh_id,
                                              const Endpoint& start_endp,
                                              const Endpoint& end_endp) {
  if (range_lock_mgr_) {
    return range_lock_mgr_->TryRangeLock(txn, cfh_id, start_endp, end_endp,
                                         /*exclusive=*/true);
  } else
    return Status::NotSupported();
}
```

```c++
void PessimisticTransactionDB::UnLock(PessimisticTransaction* txn,
                                      const LockTracker& keys,
                                      bool all_keys_hint) {
  if (all_keys_hint && range_lock_mgr_) {
    range_lock_mgr_->UnLockAll(txn, GetEnv());
    return;
  }
  lock_mgr_->UnLock(txn, keys, GetEnv());
}
```

Should this code be changed so that the calls are made through the `BaseLockMgr`  pointer and all difference is hidden in the lock manager implementation?

spetrunia:
> I'm thinking can we have a general `RangeLockManager` interface, and implement `Transaction::LockRange` with that interface. Then in #7506, we just implement the interface with toku's range tree?

This looks like something I could try doing. 
Which PR should this change go into?

spetrunia:
> The current `RangeLockManagerHandle` looks too general to me.

The idea of this object was that it is visible to the RocksDB's user, so it was intentionally very opaque.

The user is supposed to be able to:
* Create a lock manager with `NewRangeLockManager`, and just get an opaque handle 
* Pass the Handle to RocksDB in `TransactionDBOptions::lock_mgr_handle` .
* When the DB is running, use the handle for accessing the user-visible knobs of the lock manager - call `set_max_lock_memory()` or get counters with `GetStatus()`.

`RangeLockMgrHandle` achieves all of the above.


If we get rid of it,  what alternative solution should we use?
(I'm looking at how other Lock Manager-related data is exposed to the user...  There are pure virtual methods defined in TransactionDB()  (e.g. `GetDeadlockInfoBuffer`).  Then there is an implementation in PessimisticTransactionDB, which passes the call to the Lock Manager.   Is this the pattern to follow?)

cheng-chang:
@spetrunia https://github.com/facebook/rocksdb/pull/7532 is merged. Can you merge this PR with https://github.com/facebook/rocksdb/pull/7506, and ideally put the new files under the `transactions/lock/range/range_tree` directory?

utilities/transactions/lock/lock_manager.cc
```diff
@@ -11,11 +11,17 @@
...
-                            const TransactionDBOptions& opt) {
+std::shared_ptr<LockManager> NewLockManager(PessimisticTransactionDB* db,
+                                            const TransactionDBOptions& opt) {
   assert(db);
-  // TODO: determine the lock manager implementation based on configuration.
-  return new PointLockManager(db, opt);
+  if (opt.lock_mgr_handle) {
+    // A custom lock manager was provided in options
+    auto mgr = opt.lock_mgr_handle->getLockManager();
+    return std::shared_ptr<LockManager>(opt.lock_mgr_handle, mgr);
```
cheng-chang, line 20: `mgr` can be retrieved from `lock_mgr_handle`, why need to pass both as parameters?

cheng-chang (approved):
@spetrunia Thanks for the update! I only have one question left.

include/rocksdb/utilities/transaction.h
```diff
@@ -24,9 +24,81 @@ using TransactionName = std::string;
...
+  Slice slice;
+
+  /*
+    true  : the key has a "+infinity" suffix. A suffix that would compare as
+            greater than any other suffix
+    false : otherwise
+  */
+  bool inf_suffix;
+
+  Endpoint(const Slice& slice_arg, bool inf_suffix_arg = false)
```
cheng-chang, line 92: `explicit`?

include/rocksdb/utilities/transaction.h
```diff
@@ -24,9 +24,81 @@ using TransactionName = std::string;
...
+      : slice(slice_arg), inf_suffix(inf_suffix_arg) {}
+
+  Endpoint(const char* s, bool inf_suffix_arg = false)
```
cheng-chang, line 95: `explicit`?

include/rocksdb/utilities/transaction_db.h
```diff
@@ -31,6 +31,98 @@ enum TxnDBWritePolicy {
...
+  };
+
+  // Get the current counter values
+  virtual Counters GetStatus() = 0;
+
+  // Functions for range-based Deadlock reporting.
+  virtual std::vector<RangeDeadlockPath> GetRangeDeadlockInfoBuffer() = 0;
+  virtual void SetRangeDeadlockInfoBufferSize(uint32_t target_size) = 0;
+
+  virtual ~RangeLockManagerHandle(){};
```
cheng-chang, line 115: remove `;`?

utilities/transactions/lock/point/point_lock_manager_test.cc
```diff
@@ -5,80 +5,12 @@
...
-
-#include "file/file_util.h"
-#include "port/port.h"
-#include "port/stack_trace.h"
-#include "rocksdb/utilities/transaction_db.h"
-#include "test_util/testharness.h"
-#include "test_util/testutil.h"
-#include "utilities/transactions/pessimistic_transaction_db.h"
-#include "utilities/transactions/transaction_db_mutex_impl.h"
+#include "point_lock_manager_test.h"
```
cheng-chang, line 8: the include path needs to be relative to rocksdb's root directory such as `utilities/transactions/lock/point/point_lock_manager_test.h`.

cheng-chang:
@spetrunia Thanks for the update, it mostly looks good to me except some minor comments. The description of the PR seems outdated, can you update it? After merging this PR, can you also update https://github.com/facebook/rocksdb/pull/7506 by merging the master branch?

include/rocksdb/utilities/transaction_db.h
spetrunia: WIll do

include/rocksdb/utilities/transaction.h
spetrunia: Will do.

include/rocksdb/utilities/transaction.h
spetrunia: Will do.

utilities/transactions/lock/point/point_lock_manager_test.cc
spetrunia: Will do.

utilities/transactions/lock/lock_manager.cc
spetrunia: > mgr can be retrieved from lock_mgr_handle, why need to pass both as parameters?

This is because of the following:

`opt.lock_mgr_handle` is defined as:
```
  std::shared_ptr<LockManagerHandle> lock_mgr_handle;
```
while this function's return value is:
```
  std::shared_ptr<LockManager>
```

Classes `LockManager` and `LockManagerHandle` do not inherit from one another (one needs to call getLockManager() to get one from the other).

And apparently, std::shared_ptr has a special constructor template for such cases, called "aliasing constructor":

```
return std::shared_ptr<LockManager>(opt.lock_mgr_handle, mgr);
```
this produces a shared pointer looks like a `LockManager*` when dereferenced, but its free-on-delete part shares the pointer to `LockManagerHandle` so it can share with `opt.lock_mgr_handle`. 

(I'm sorry if I'm re-stating the obvious here... It's the first time I saw/used this constructor).

cheng-chang:
The format check fails, can you run `make format`? I think we are good to go after formatting.

PR #7445, opened 2020-09-28
Add flush_job_test to the list of ASSERT_STATUS_CHECKED tests

zhichao-cao:
test plan: pass ASSERT_STATUS_CHECKED=1 make -j48 flush_job_test

### DIFF
--- a/db/flush_job.cc
+++ b/db/flush_job.cc
@@ -130,6 +130,7 @@ FlushJob::FlushJob(
 }
 
 FlushJob::~FlushJob() {
+  io_status_.PermitUncheckedError();
   ThreadStatusUtil::ResetThreadStatus();
 }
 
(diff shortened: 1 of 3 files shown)


db/flush_job.cc
```diff
@@ -130,6 +130,7 @@ FlushJob::FlushJob(
 }
 
 FlushJob::~FlushJob() {
+  io_status_.PermitUncheckedError();
```
akankshamahajan15, line 133: Can we add assert ```io_status()```  is ok in flush_job_test.cc instead of ```PermitUncheckedError()```?

mrambacher: I actually think how it is coded is correct and it should not be an assert.  If one simply creates a FlushJob and does nothing with it, then the assert would be triggered.

mrambacher (approved):
LGTM.

db/flush_job.cc
zhichao-cao: @akankshamahajan15 I agree with what Mark said.

PR #7446, opened 2020-09-28
Enable force_consistency_checks by default

pdillinger:
Summary: This has been running in production on some key workloads, so
we believe it to be safe and extremely low cost. Nevertheless, I've
added code to ensure that "force_consistency_checks" is mentioned in
any corruption reports so that people know how to disable in case of
false positive corruption reports.

Test Plan: make check, CI, temporary debug print new message with
./version_builder_test


ajkr (approved):
LGTM!

mrambacher (approved):
LGTM.

Does this change apply to only new column families?  The old value will be stored in the Options file; which value will take precedence?

I would also recommend runing some tests to verify that the change in options behavior is backwards compatible.

pdillinger:
> Does this change apply to only new column families? The old value will be stored in the Options file; which value will take precedence?

I guess maybe. Not going to worry about that too much, except to mention in HISTORY.md

> I would also recommend runing some tests to verify that the change in options behavior is backwards compatible.

We have a continuously triggered script for that, tools/check_format_compatible.sh, and it would have been an existing bug for anyone changing the option.

PR #7447, opened 2020-09-28
Support injecting read errors for RandomAccessFile when using FaultInjectionTestEnv

ltamasi:
Summary:
The patch adds support for injecting errors when reading from `RandomAccessFile`
using `FaultInjectionTestEnv`. (This functionality was curiously missing
w/r/t `RandomAccessFile`, even though it was implemented for `RandomRWFile`.)
The patch also fixes up a test case in `blob_db_test` which uses `FaultInjectionTestEnv`
but has so far relied on reads from `RandomAccessFile`s succeeding even after
deactivating the filesystem.

Test Plan:
`make check`

### DIFF
--- a/utilities/fault_injection_env.cc
+++ b/utilities/fault_injection_env.cc
@@ -116,11 +116,40 @@ TestRandomAccessFile::TestRandomAccessFile(
 
 Status TestRandomAccessFile::Read(uint64_t offset, size_t n, Slice* result,
                                   char* scratch) const {
+  assert(env_);
   if (!env_->IsFilesystemActive()) {
     return env_->GetError();
   }
 
+  assert(target_);
   return target_->Read(offset, n, result, scratch);
+}
+
+Status TestRandomAccessFile::Prefetch(uint64_t offset, size_t n) {
+  assert(env_);
+  if (!env_->IsFilesystemActive()) {
+    return env_->GetError();
+  }
+
+  assert(target_);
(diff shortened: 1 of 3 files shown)


utilities/fault_injection_env.cc
```diff
@@ -107,6 +107,51 @@ Status TestDirectory::Fsync() {
...
+}
+
+Status TestRandomAccessFile::MultiRead(ReadRequest* reqs, size_t num_reqs) {
+  assert(env_);
+  if (!env_->IsFilesystemActive()) {
+    const Status s = env_->GetError();
+
+    assert(reqs);
+    for (size_t i = 0; i < num_reqs; ++i) {
+      reqs[i].status = s;
```
zhichao-cao, line 145: Maybe better assert the size of `reqs` is always larger than `i`

ltamasi: It's not really possible to do that with this interface: given only a pointer, there's no way of getting the "real" size of `reqs`. We have to rely on what the caller passes in `num_reqs`.

zhichao-cao: > It's not really possible to do that with this interface: given only a pointer, there's no way of getting the "real" size of `reqs`. We have to rely on what the caller passes in `num_reqs`.

Oh, you are right.

zhichao-cao (approved):

PR #7451, opened 2020-09-28
Add is_full_compaction to CompactionJobStats, cleanup

pdillinger:
Summary: This exposes to the listener interface whether a compaction was
full or not. Also cleaned up API comment for CompactionJobInfo::stats,
which is not of a nullable type. And since CompactionJob is always
created with non-null CompactionJobStats, removed conditionals on it
being nullptr and instead assert non-null.

TODO later: update C and Java interfaces

Test Plan: updated existing unit tests to check new field, make check


db/compaction/compaction_job_stats_test.cc
```diff
@@ -571,7 +572,7 @@ CompactionJobStats NewManualCompactionJobStats(
     uint64_t num_input_records, size_t key_size, size_t value_size,
     size_t num_output_files, uint64_t num_output_records,
     double compression_ratio, uint64_t num_records_replaced,
-    bool is_manual = true) {
+    bool is_full = false, bool is_manual = true) {
```
ltamasi, line 575: Bit of a risky move but...

db/compaction/compaction_job_stats_test.cc
```diff
@@ -981,25 +983,20 @@ TEST_P(CompactionJobStatsTest, UniversalCompactionTest) {
...
-            1.0, 0, false));
+    // based on whether it is a full compaction.
+    uint64_t smallest_key = is_full ? key_base : key_base * (num_flushes - 1);
+
+    stats_checker->AddExpectedStats(NewManualCompactionJobStats(
+        Key(smallest_key, 10),
+        Key(smallest_key + key_base * num_input_units - key_interval, 10),
+        num_input_units, num_input_units > 2 ? num_input_units / 2 : 0,
+        num_keys_per_table * num_input_units, kKeySize, kValueSize,
+        num_input_units, num_keys_per_table * num_input_units, 1.0, 0, is_full,
```
ltamasi, line 998: ... this is the only place where `is_manual` is passed I presume?

util/compaction_job_stats_impl.cc
```diff
@@ -20,6 +20,7 @@ void CompactionJobStats::Reset() {
   num_output_records = 0;
   num_output_files = 0;
 
+  is_full_compaction = 0;
```
ltamasi, line 23: Would be nicer to initialize these two using `false` (or else the linter might complain).

ltamasi (approved):

db/compaction/compaction_job_stats_test.cc
pdillinger: I felt especially comfortable with the change since it's a single-file function. Somewhere else already put "is_full" before "is_manual" so I tried to keep that consistent. :)

PR #7453, opened 2020-09-29
Enforce status check for corruption_test

riversand963:
Enforce status check for corruption_test.

Test plan:
```
ASSERT_STATUS_CHECKED=1 make corruption_test
./corruption_test
```

### DIFF
--- a/db/compaction/compaction_job.cc
+++ b/db/compaction/compaction_job.cc
@@ -943,7 +943,6 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
   std::string last_key_for_partitioner;
 
   while (status.ok() && !cfd->IsDropped() && c_iter->Valid()) {
-    assert(c_iter->status().ok());
     // Invariant: c_iter.status() is guaranteed to be OK if c_iter->Valid()
     // returns true.
     const Slice& key = c_iter->key();
@@ -1074,8 +1073,6 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
   }
   if (status.ok()) {
     status = c_iter->status();
-  } else {
-    c_iter->status().PermitUncheckedError();
   }
 
   if (status.ok() && sub_compact->builder == nullptr &&
--- a/db/compaction/compaction_job.cc
+++ b/db/compaction/compaction_job.cc
@@ -943,6 +943,7 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
   std::string last_key_for_partitioner;
 
   while (status.ok() && !cfd->IsDropped() && c_iter->Valid()) {
+    assert(c_iter->status().ok());
     // Invariant: c_iter.status() is guaranteed to be OK if c_iter->Valid()
     // returns true.
     const Slice& key = c_iter->key();
@@ -1073,6 +1074,8 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
   }
   if (status.ok()) {
     status = c_iter->status();
+  } else {
+    c_iter->status().PermitUncheckedError();
   }
 
   if (status.ok() && sub_compact->builder == nullptr &&
--- a/db/corruption_test.cc
+++ b/db/corruption_test.cc
@@ -56,7 +56,8 @@ class CorruptionTest : public testing::Test {
     options_.wal_recovery_mode = WALRecoveryMode::kTolerateCorruptedTailRecords;
     options_.env = &env_;
     dbname_ = test::PerThreadDBPath("corruption_test");
-    DestroyDB(dbname_, options_);
+    Status s = DestroyDB(dbname_, options_);
+    EXPECT_OK(s);
 
     db_ = nullptr;
     options_.create_if_missing = true;
@@ -111,12 +112,12 @@ class CorruptionTest : public testing::Test {
     for (int i = 0; i < n; i++) {
       if (flush_every != 0 && i != 0 && i % flush_every == 0) {
         DBImpl* dbi = static_cast_with_check<DBImpl>(db_);
-        dbi->TEST_FlushMemTable();
+        ASSERT_OK(dbi->TEST_FlushMemTable());
       }
       //if ((i % 100) == 0) fprintf(stderr, "@ %d of %d\n", i, n);
       Slice key = Key(i, &key_space);
       batch.Clear();
-      batch.Put(key, Value(i, &value_space));
+      ASSERT_OK(batch.Put(key, Value(i, &value_space)));
       ASSERT_OK(db_->Write(WriteOptions(), &batch));
     }
   }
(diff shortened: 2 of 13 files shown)


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -2982,9 +2983,10 @@ Status DBImpl::BackgroundCompaction(bool* made_progress,
       auto err_reason = versions_->io_status().ok()
                             ? BackgroundErrorReason::kCompaction
                             : BackgroundErrorReason::kManifestWrite;
-      error_handler_.SetBGError(io_s, err_reason);
+      error_handler_.SetBGError(io_s, err_reason).PermitUncheckedError();
```
mrambacher, line 2986: If (almost) everywhere that calls SetBGError just ignores the error, maybe the method should be changed to return a void?  Or there should be a second signature that returns a void?

db/table_cache.cc
```diff
@@ -171,6 +171,7 @@ Status TableCache::FindTable(const ReadOptions& ro,
 
   if (*handle == nullptr) {
     if (no_io) {  // Don't do IO and return a not-found status
+      s.PermitUncheckedError();
```
mrambacher, line 174: Can you fix the comment above about "Not Found',  It returns Incomplete"

db/corruption_test.cc
```diff
@@ -151,6 +153,7 @@ class CorruptionTest : public testing::Test {
         correct++;
       }
     }
+    iter->status().PermitUncheckedError();
```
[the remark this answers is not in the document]
riversand963: I thought `PermitUncheckedError()` is for the purpose of temporarily letting some of the tests pass. If I put it in destructor of `CompactionIterator`, then it seems to mean callers of `CompactionIterator` does not have to check its `status_`. By letting caller determine whether and where to check `status_`, we can better enforce the checking. Let me know if I have misunderstood the intention.

db/db_impl/db_impl_compaction_flush.cc
riversand963: Good point. cc @anand1976

db/corruption_test.cc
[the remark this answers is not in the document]
mrambacher: I think there is generally an issue with Status as member variables.  If I create a CompactionIterator and never do anything with it, I will get an error about the status not being checked.  This seems wrong to me as I never did anything with the iterator. 

In any case, my guess is someone else is going to hit another case where the CompactionIterator status is not checked as well...

db/table_cache.cc
riversand963: Done.

db/corruption_test.cc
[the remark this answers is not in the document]
riversand963: In the case you mentioned, I think it is easy to just call `PermitUncheckedError()` for the member `status`. This seems a little more, but not the worst thing. It will be good if this can be simplified.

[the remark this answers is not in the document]
riversand963: Things can be more tedious if the class has multiple status members...

db/version_set.cc
```diff
@@ -4137,7 +4137,8 @@ Status VersionSet::ProcessManifestWrites(
                      "\n",
                      manifest_file_number_, pending_manifest_file_number_);
       env_->DeleteFile(
-          DescriptorFileName(dbname_, pending_manifest_file_number_));
+              DescriptorFileName(dbname_, pending_manifest_file_number_))
+          .PermitUncheckedError();
```
jay-zhuang, line 4141: should it log a error if a file failed to delete?

table/table_factory.cc
```diff
@@ -37,7 +37,8 @@ Status TableFactory::CreateFromString(const ConfigOptions& config_options_in,
     factory->reset(new CuckooTableFactory());
 #endif  // ROCKSDB_LITE
   } else {
-    return Status::NotSupported("Could not load table factory: ", name);
+    status = Status::NotSupported("Could not load table factory: ", name);
+    return status;
```
jay-zhuang, line 41: Why this change is needed?

jay-zhuang (approved):
LGTM. Just a few minor questions.

db/write_thread.h
```diff
@@ -83,6 +83,8 @@ class WriteThread {
     std::atomic<size_t> running;
     size_t size = 0;
 
+    ~WriteGroup() { status.PermitUncheckedError(); }
```
siying, line 86: Can you add a comment about why this can be permitted?

db/db_impl/db_impl_compaction_flush.cc
zhichao-cao: @anand1976 Mark suggested to have 2 versions of `SetBGError`. One returns the status after we deal with the BG error, and in this case, the caller will make some actions based on the returned error. Another one is a `void` method, we just set the BG error

db/version_set.cc
riversand963: Good catch. Will update.

table/table_factory.cc
riversand963: If we reach here and proceed, the `status` obj created in line 17 will be unchecked.

db/compaction/compaction_job.cc
```diff
@@ -1062,6 +1063,8 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
   }
   if (status.ok()) {
     status = c_iter->status();
+  } else {
+    c_iter->status().PermitUncheckedError();
```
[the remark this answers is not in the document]
riversand963: Will remove the change here.

db/write_thread.h
riversand963: Actually, I think it's better not to call `PermitUncheckedError()` in destructor. Let me update the code.

db/db_impl/db_impl_write.cc
```diff
@@ -999,8 +999,9 @@ WriteBatch* DBImpl::MergeBatch(const WriteThread::WriteGroup& write_group,
     merged_batch = tmp_batch;
     for (auto writer : write_group) {
       if (!writer->CallbackFailed()) {
-        WriteBatchInternal::Append(merged_batch, writer->batch,
-                                   /*WAL_only*/ true);
+        Status s = WriteBatchInternal::Append(merged_batch, writer->batch,
+                                              /*WAL_only*/ true);
+        assert(s.ok());
```
siying, line 1004: Can you add a comment to explain why?

db/db_impl/db_impl_write.cc
```diff
@@ -1925,10 +1926,10 @@ Status DB::DeleteRange(const WriteOptions& opt,
                        const Slice& begin_key, const Slice& end_key) {
   WriteBatch batch;
   Status s = batch.DeleteRange(column_family, begin_key, end_key);
-  if (!s.ok()) {
-    return s;
+  if (s.ok()) {
+    s = Write(opt, &batch);
   }
-  return Write(opt, &batch);
+  return s;
```
siying, line 1932: I don't see a difference between the changes.

siying (commented):
My comments have been addressed. Thanks!

db/db_impl/db_impl_write.cc
riversand963: Done.

db/db_impl/db_impl_write.cc
riversand963: Updated.

PR #7455, opened 2020-09-29
add Status check assertions for repair_test

ajkr:


db/db_test_util.cc
```diff
@@ -1407,29 +1407,33 @@ void DBTestBase::CopyFile(const std::string& source,
...
-      env_->GetFileSize(file_path, &file_size);
-      res[file_path] = file_size;
-      if (total_size) {
-        *total_size += file_size;
+  Status s = env_->GetChildren(dbname_, &files);
+  if (s.ok()) {
+    for (auto& file_name : files) {
+      uint64_t number;
+      FileType type;
+      std::string file_path = dbname_ + "/" + file_name;
```
riversand963, line 1422: We need to construct the `file_path` only if the check on line 1423 is true.

riversand963 (approved):
LGTM.

db/db_test_util.cc
ajkr: Fixed.

ajkr (commented):
Thanks for the review!

db/table_cache.cc
```diff
@@ -171,7 +171,8 @@ Status TableCache::FindTable(const ReadOptions& ro,
 
   if (*handle == nullptr) {
     if (no_io) {  // Don't do IO and return a not-found status
-      return Status::Incomplete("Table not found in table_cache, no_io is set");
+      s = Status::Incomplete("Table not found in table_cache, no_io is set");
```
mrambacher, line 174: Please fix the comment on the earlier line

db/repair.cc
```diff
@@ -163,31 +164,40 @@ class Repairer {
...
   }
 
-  ~Repairer() {
-    if (db_lock_ != nullptr) {
-      env_->UnlockFile(db_lock_);
+  Status Close() {
+    Status s = Status::OK();
+    if (!closed_) {
+      if (db_lock_ != nullptr) {
+        s = env_->UnlockFile(db_lock_);
```
mrambacher, line 171: What happens if the Unlock fails?  The file will be marked as Closed but the db_lock_ is still hanging out?

db/table_cache.cc
ajkr: Deleted the comment since it was just inaccurately repeating the code.

db/repair.cc
ajkr: Fixed to set to nullptr since `UnlockFile()` is not retryable and deletes the pointer passed to it even in the failure case.

ajkr (commented):
Thanks for the review!

db/table_cache.cc
```diff
@@ -170,8 +170,9 @@ Status TableCache::FindTable(const ReadOptions& ro,
                            const_cast<bool*>(&no_io));
 
   if (*handle == nullptr) {
-    if (no_io) {  // Don't do IO and return a not-found status
-      return Status::Incomplete("Table not found in table_cache, no_io is set");
+    if (no_io) {
+      s = Status::Incomplete("Table not found in table_cache, no_io is set");
```
ajkr, line 174: Appveyor says:

```
    35>c:\projects\rocksdb\db\table_cache.cc(173): error C2065: 's': undeclared identifier [C:\projects\rocksdb\build\rocksdb.vcxproj]
    35>c:\projects\rocksdb\db\table_cache.cc(174): error C2065: 's': undeclared identifier [C:\projects\rocksdb\build\rocksdb.vcxproj]
```

I wonder why it thinks `s` is undeclared...

ajkr: It must have automatically rebased without any conflict and lost the declaration. I've never seen a clean rebase cause a problem before.

PR #7457, opened 2020-09-29
Change ParseInternalKey() to return Status instead of bool 

### LINKED ISSUE #7430: investigate making `ParseInternalKey()` return a `Status`
`ParseInternalKey()` is unusual as it returns a `bool` instead of a `Status`. Now we have assertions to ensure clients actually check the `Status` objects they are returned, so `Status` is safer than `bool`. We can try making it a `Status` and see if performance regresses.

ramvadiv:
Fixes #7430 

Change ParseInternalKey() to return Status instead of bool. 

db_bench (seekrandom) based before/after results with value size of 100 bytes and 16 bytes can be found at (tests ran on an udb server):
https://www.dropbox.com/s/47bwamdy5ozngph/PIK_ret_Status_results.xlsx?dl=0

db_bench_results


db/range_del_aggregator.cc
```diff
@@ -33,15 +33,16 @@ TruncatedRangeDelIterator::TruncatedRangeDelIterator(
   if (smallest != nullptr) {
     pinned_bounds_.emplace_back();
     auto& parsed_smallest = pinned_bounds_.back();
-    if (!ParseInternalKey(smallest->Encode(), &parsed_smallest)) {
+    if (ParseInternalKey(smallest->Encode(), &parsed_smallest) !=
+        Status::OK()) {
```
ajkr, line 37: same

db/range_del_aggregator.cc
```diff
@@ -33,15 +33,16 @@ TruncatedRangeDelIterator::TruncatedRangeDelIterator(
...
       assert(false);
     }
     smallest_ = &parsed_smallest;
   }
   if (largest != nullptr) {
     pinned_bounds_.emplace_back();
     auto& parsed_largest = pinned_bounds_.back();
-    if (!ParseInternalKey(largest->Encode(), &parsed_largest)) {
+    if (ParseInternalKey(largest->Encode(), &parsed_largest) != Status::OK()) {
```
ajkr, line 45: same

db/range_del_aggregator.h
```diff
@@ -283,7 +283,7 @@ class RangeDelAggregator {
 
   bool ShouldDelete(const Slice& key, RangeDelPositioningMode mode) {
     ParsedInternalKey parsed;
-    if (!ParseInternalKey(key, &parsed)) {
+    if (ParseInternalKey(key, &parsed) != Status::OK()) {
```
ajkr, line 286: same, I think the `return false` should even be deleted.

ajkr (approved):
LGTM! There could probably be a follow-up task to combine this with #7420 by pushing down the option's value into `ParseInternalKey()` (defaulting to false when unknown) and then including the corrupt key in the `Status::Corruption` there. 

The github convention for linking an issue is to put a line "Fixes #7430." in the description (rather than title), then it'll auto-close the issue when this PR merges.

java/rocksjni/write_batch_test.cc
```diff
@@ -66,7 +66,7 @@ jbyteArray Java_org_rocksdb_WriteBatchTest_getContents(JNIEnv* env,
     ROCKSDB_NAMESPACE::Status pikStatus =
         ROCKSDB_NAMESPACE::ParseInternalKey(iter->key(), &ikey);
     if (!pikStatus.ok()) {
-      assert(0);
+      assert(false);
```
mrambacher, line 69: Why not just assert(pikStatus.ok())

ramvadiv: Thanks Mark. Fixed.

db/range_del_aggregator.h
ramvadiv: Thanks a lot for your review/comments Andrew. I have handled all of your other comments. I am not very clear about this point - Shouldn't we still return false in this case?

ajkr: I don't know. Returning false originally felt safe here because it conservatively preserves data that we cannot conclusively determine whether it's deleted due to a corruption. OTOH, it doesn't seem right to have a code path that makes a guess about whether data is deleted or not. The final solution should involve propagating the `Status`, so I guess it doesn't matter much what we do in the meantime (leave the `return false;` guess here, or let the code continue and probably crash).

PR #7459, opened 2020-09-29
Report error of GetChildren

riversand963:
As title

Test plan:
make check

### DIFF
--- a/db/db_impl/db_impl_files.cc
+++ b/db/db_impl/db_impl_files.cc
@@ -762,10 +762,13 @@ Status DBImpl::FinishBestEffortsRecovery() {
   uint64_t next_file_number = versions_->current_next_file_number();
   uint64_t largest_file_number = next_file_number;
   std::set<std::string> files_to_delete;
+  Status s;
   for (const auto& path : paths) {
     std::vector<std::string> files;
-    // Should we handle it?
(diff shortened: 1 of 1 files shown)


db/db_impl/db_impl_files.cc
```diff
@@ -762,10 +762,13 @@ Status DBImpl::FinishBestEffortsRecovery() {
   uint64_t next_file_number = versions_->current_next_file_number();
   uint64_t largest_file_number = next_file_number;
   std::set<std::string> files_to_delete;
+  Status s;
   for (const auto& path : paths) {
     std::vector<std::string> files;
-    // Should we handle it?
-    env_->GetChildren(path, &files).PermitUncheckedError();
+    s = env_->GetChildren(path, &files);
+    if (!s.ok()) {
```
mrambacher, line 769: Why not just return s here and not break and return down below?

mrambacher (approved):
LGTM.

db/db_impl/db_impl_files.cc
riversand963: Thanks for the review. Sure we can return here.
One potential benefit of breaking out and returning below is that we can avoid duplicating some clean-up code, assuming RAII is not used (which we probably should). This is not necessary in this function though.

PR #7460, opened 2020-09-29
Enable ASSERT_STATUS_CHECKED for db_universal_compaction_test

anand1976:


db/db_universal_compaction_test.cc
```diff
@@ -1239,12 +1239,13 @@ TEST_P(DBTestUniversalCompaction, UniversalCompactionFourPaths) {
   options.num_levels = 1;
 
   std::vector<std::string> filenames;
-  env_->GetChildren(options.db_paths[1].path, &filenames);
+  env_->GetChildren(options.db_paths[1].path, &filenames)
```
mrambacher, line 1242: Should this be EXPECT_OK in the test?  When would this fail in the test scenarios and not be checked?

db/db_universal_compaction_test.cc
```diff
@@ -1239,12 +1239,13 @@ TEST_P(DBTestUniversalCompaction, UniversalCompactionFourPaths) {
...
+      .PermitUncheckedError();
   // Delete archival files.
   for (size_t i = 0; i < filenames.size(); ++i) {
-    env_->DeleteFile(options.db_paths[1].path + "/" + filenames[i]);
+    ASSERT_OK(env_->DeleteFile(options.db_paths[1].path + "/" + filenames[i]));
   }
-  env_->DeleteDir(options.db_paths[1].path);
+  env_->DeleteDir(options.db_paths[1].path).PermitUncheckedError();
```
mrambacher, line 1248: How does this fail in the test?  Why is it not an ASSERT/EXPECT_OK?

db/db_test_util.cc
```diff
@@ -1211,7 +1212,8 @@ std::string DBTestBase::DumpSSTableList() {
 
 void DBTestBase::GetSstFiles(Env* env, std::string path,
                              std::vector<std::string>* files) {
-  env->GetChildren(path, files);
+  // Path may not exist
+  env->GetChildren(path, files).PermitUncheckedError();
```
mrambacher, line 1216: Would the Status ever be non-OK and the files be non-empty?  Why not check if ok() and then do the following code?

db/db_universal_compaction_test.cc
```diff
@@ -1579,12 +1581,13 @@ TEST_P(DBTestUniversalCompaction, UniversalCompactionSecondPathRatio) {
       new SpecialSkipListFactory(KNumKeysByGenerateNewFile - 1));
 
   std::vector<std::string> filenames;
-  env_->GetChildren(options.db_paths[1].path, &filenames);
+  env_->GetChildren(options.db_paths[1].path, &filenames)
+      .PermitUncheckedError();
```
mrambacher, line 1585: Why not an EXPECT/ASSERT here?

db/db_universal_compaction_test.cc
```diff
@@ -1579,12 +1581,13 @@ TEST_P(DBTestUniversalCompaction, UniversalCompactionSecondPathRatio) {
...
   // Delete archival files.
   for (size_t i = 0; i < filenames.size(); ++i) {
-    env_->DeleteFile(options.db_paths[1].path + "/" + filenames[i]);
+    ASSERT_OK(env_->DeleteFile(options.db_paths[1].path + "/" + filenames[i]));
   }
-  env_->DeleteDir(options.db_paths[1].path);
+  env_->DeleteDir(options.db_paths[1].path).PermitUncheckedError();
```
mrambacher, line 1590: Why not EXPECT/ASSERT?

db/db_universal_compaction_test.cc
anand1976: The directory may not exist.

mrambacher: So rather than ignoring the error, can we check for the case were the directory does not exist and do the skip all of hte deletes?

anand1976:
@mrambacher I addressed your comments. PTAL.

Makefile
```diff
@@ -661,6 +661,7 @@ ifdef ASSERT_STATUS_CHECKED
 		column_family_test \
 		file_reader_writer_test \
 		corruption_test \
+		db_universal_compaction_test
```
riversand963, line 664: Nit: a trailing `\` can be added.

riversand963 (approved):
LGTM.

PR #7463, opened 2020-09-30
Return error if Get() fails in Prefetching Filter blocks

akankshamahajan15:
Summary: Right now all I/O failures under
PartitionFilterBlock::CacheDependencies() is swallowed. Return error in
case prefetch fails.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/file/file_prefetch_buffer.cc
+++ b/file/file_prefetch_buffer.cc
@@ -94,6 +94,14 @@ Status FilePrefetchBuffer::Prefetch(const IOOptions& opts,
   if (!s.ok()) {
     return s;
   }
+
+#ifndef NDEBUG
+  if (result.size() < read_len) {
+    // Fake an IO error to force db_stress fault injection to ignore
+    // truncated read errors
+    IGNORE_STATUS_IF_ERROR(Status::IOError());
+  }
+#endif
   buffer_offset_ = rounddown_offset;
   buffer_.Size(static_cast<size_t>(chunk_len) + result.size());
   return s;
--- a/file/file_prefetch_buffer.cc
+++ b/file/file_prefetch_buffer.cc
@@ -91,17 +91,11 @@ Status FilePrefetchBuffer::Prefetch(const IOOptions& opts,
   size_t read_len = static_cast<size_t>(roundup_len - chunk_len);
   s = reader->Read(opts, rounddown_offset + chunk_len, read_len, &result,
                    buffer_.BufferStart() + chunk_len, nullptr, for_compaction);
-#ifndef NDEBUG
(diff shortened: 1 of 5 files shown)


akankshamahajan15:
Failure:  ci/circleci: build-linux-clang10-ubsan  not related to this change.

file/file_prefetch_buffer.cc
```diff
@@ -91,17 +91,11 @@ Status FilePrefetchBuffer::Prefetch(const IOOptions& opts,
...
-  if (!s.ok() || result.size() < read_len) {
-    // Fake an IO error to force db_stress fault injection to ignore
-    // truncated read errors
-    IGNORE_STATUS_IF_ERROR(Status::IOError());
-  }
-#endif
-  if (s.ok()) {
-    buffer_offset_ = rounddown_offset;
-    buffer_.Size(static_cast<size_t>(chunk_len) + result.size());
+  if (!s.ok()) {
```
anand1976, line 94: We still need to call ```IGNORE_STATUS_IF_ERROR()``` if ```result.size() < read_len```. That's not really an error and can happen if the readahead goes past the end of file.

table/block_based/partitioned_filter_block.cc
```diff
@@ -477,17 +475,18 @@ void PartitionedFilterBlockReader::CacheDependencies(const ReadOptions& ro,
         prefetch_buffer.get(), ro, handle, UncompressionDict::GetEmptyDict(),
         &block, BlockType::kFilter, nullptr /* get_context */, &lookup_context,
         nullptr /* contents */);
-
-    assert(s.ok() || block.GetValue() == nullptr);
```
anand1976, line 481: I think this assertion is still valid?

table/block_based/partitioned_filter_block.cc
```diff
@@ -426,12 +426,7 @@ void PartitionedFilterBlockReader::CacheDependencies(const ReadOptions& ro,
   Status s = GetOrReadFilterBlock(false /* no_io */, nullptr /* get_context */,
                                   &lookup_context, &filter_block);
   if (!s.ok()) {
-    ROCKS_LOG_WARN(rep->ioptions.info_log,
```
anand1976, line 429: We should retain this log message

table/block_based/partitioned_filter_block.cc
akankshamahajan15: yes it is. I recently got to know that assertions are excluded in optimized builds.

anand1976 (approved):

PR #7466, opened 2020-09-30
Add a rocksdb lib target with link_whole=True

riversand963:
We would like to build a shared library with all fbcode dependencies statically linked within.
This resulting .so should not drop any symbols definitions in the building process.
To ensure that, we use `link_whole=True` according to
https://buck.build/rule/cxx_library.html#link_whole.
Since `link_whole` is `False` by default, adding a `link_whole=False` to existing libraries won't
change any behavior.

Test plan:
build a .so and test internally.


pdillinger (approved):
If it works for you, it works for me :)

riversand963:

PR #7477, opened 2020-09-30
Fix flaky intra-L0 consistency failure regression tests

ajkr:
Do not assert the number of L0 files after intra-L0 compaction is eligible to run since compaction could complete and reduce the number of L0 files before the assertion executes.


pdillinger (approved):
Seems OK

PR #7478, opened 2020-09-30
Fix TSAN failure for backupable_db_test

jay-zhuang:
It's a transient failure, but can be reproduce with running the test 100
times:
https://app.circleci.com/pipelines/github/facebook/rocksdb/3760/workflows/de909685-f22b-45ba-a8f3-6ebb78a54e96/jobs/37039

test plan: re-run the test 100 times


ajkr (approved):
LGTM.

PR #7479, opened 2020-09-30
Add a host location property to TableProperties

anand1976:
This PR adds support for writing a location identifier of the DB host to SST files as a table property. By default, the hostname is used, but can be overridden by the user. There have been some recent corruptions in files written by ```SstFileWriter``` before checksumming, so this property can be used to trace it back to the writing host and checking the host for hardware isues.

Test plan:
Add new unit tests


include/rocksdb/options.h
```diff
@@ -1157,6 +1157,13 @@ struct DBOptions {
   //
   // Default: 1000000 (microseconds).
   uint64_t bgerror_resume_retry_interval = 1000000;
+
+  // A string identifying the location of the machine hosting the DB. This
+  // will be written as a property in every SST file written by the DB (or
+  // by offline writers such as SstFileWriter and RepairDB)
+  //
+  // Default: hostname
+  std::string db_host_location;
```
mrambacher, line 1166: Can you give a better comment as to what this would be used for?  Why would a user want to set this and why do we want this in the SST files?

mrambacher (commented):
I am a bit worried about how this change will behave for options.  I know we attempt to validate that the options have not changed from what was set in the file but it seems like that could very well happen here (if the option was not set on startup).  It would also make an Options file more tied to a specific host (if we care).

There also needs to be changes to options_settable_test.

anand1976:
@mrambacher 

> I am a bit worried about how this change will behave for options. I know we attempt to validate that the options have not changed from what was set in the file but it seems like that could very well happen here (if the option was not set on startup). It would also make an Options file more tied to a specific host (if we care).

I didn't understand the concern. We do sanitize other options, such as ```wal_dir```, before writing out to the OPTIONS file.  I don't think we read back the options file and try to validate.

HISTORY.md
```diff
@@ -8,6 +8,7 @@
...
 * AdvancedColumnFamilyOptions::force_consistency_checks is now true by default, for more proactive DB corruption detection at virtually no cost (estimated two extra CPU cycles per million on a major production workload). Corruptions reported by these checks now mention "force_consistency_checks" in case a false positive corruption report is suspected and the option needs to be disabled (unlikely). Since existing column families have a saved setting for force_consistency_checks, only new column families will pick up the new default.
+* A new option, db_host_location, has been added to DBOptions. If left empty, this option defaults to the hostname (FQDN). Either the default or the user specified string is written to SST files as a table property
```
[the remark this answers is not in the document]
anand1976: Ok, we can do that. I'm not keen on introducing another option field to disable this though. How about default initialize ```options.db_host_location``` to some well known string, like "default" or "hostname"? We can parse it in ```SanitizeOptions``` and replace it with the actual hostname. To disable the property completely, the user can just set ```options.db_host_location``` to an empty string.

db/db_impl/db_impl_open.cc
```diff
@@ -181,6 +183,18 @@ DBOptions SanitizeOptions(const std::string& dbname, const DBOptions& src) {
                    "file size check will be skipped during open.");
   }
 
+  if (result.db_host_location.empty()) {
+    result.db_host_location.resize(MAX_HOST_NAME_LEN, 0);
```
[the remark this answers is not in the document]
anand1976: ```env->GetHostName()``` requires a pre-allocated ```char``` array.

db/db_impl/db_impl_open.cc
```diff
@@ -181,6 +183,18 @@ DBOptions SanitizeOptions(const std::string& dbname, const DBOptions& src) {
...
+    if (!result.env
+             ->GetHostName(&result.db_host_location.at(0), MAX_HOST_NAME_LEN)
+             .ok()) {
+      ROCKS_LOG_INFO(result.info_log,
+                     "db_host_location property will not be set");
+    } else {
+      result.db_host_location.at(MAX_HOST_NAME_LEN - 1) = 0;
```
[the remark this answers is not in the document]
anand1976: ```env->GetHostName()``` does not guarantee a null terminated string if the hostname exceeds the input array length.

HISTORY.md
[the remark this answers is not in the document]
siying: Can it be something like default value to be "___automatic___"?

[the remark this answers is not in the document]
mrambacher: How about this as a solution:
- We do not sanitize the DBOption but leave it "as is"
- If the "host_location" is empty, nothing is written to the SST file
- If the "host_location" is "localhost" or "default", the host name is written to the SST file (the Env is available in the options)

Also, it is not really a "location" but more of a "name" or "id".  

I still do not understand the use case for this location parameter and why I would want to use it.  What about the other table formats?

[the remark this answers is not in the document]
anand1976: Both sound like good ideas. Personally I like "__hostname__" or something like that as the "__" hints that it will be substituted with something, and "__hostname__" makes it clear what that something is. I think there's little risk of anyone taking it literally.

Regarding the option name, maybe "db_host_id" sounds better.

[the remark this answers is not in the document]
anand1976: > How about this as a solution:
> 
> * We do not sanitize the DBOption but leave it "as is"

That's a good point. That way the OPTIONS file can be read back in and validated without issues (which may have been your original concern).

options/options_helper.cc
```diff
@@ -164,6 +164,7 @@ DBOptions BuildDBOptions(const ImmutableDBOptions& immutable_db_options,
       immutable_db_options.max_bgerror_resume_count;
   options.bgerror_resume_retry_interval =
       immutable_db_options.bgerror_resume_retry_interval;
+  //  options.db_host_location = immutable_db_options.db_host_location;
```
pdillinger, line 167: stale code?

table/format.cc
```diff
@@ -403,4 +404,17 @@ Status UncompressBlockContents(const UncompressionInfo& uncompression_info,
                                                    ioptions, allocator);
 }
 
+Status SanitizeDbHostIdProperty(Env* env, std::string& db_host_id) {
```
pdillinger, line 407: Google style: use pointer for mutable parameters

table/format.cc
```diff
@@ -403,4 +404,17 @@ Status UncompressBlockContents(const UncompressionInfo& uncompression_info,
...
+  if (db_host_id == kDbHostId) {
+    db_host_id.clear();
+    db_host_id.resize(kMaxHostNameLen);
+    Status s = env->GetHostName(&db_host_id.at(0), kMaxHostNameLen);
+    if (!s.ok()) {
+      return s;
```
pdillinger, line 413: Doesn't this leave the db_host_id non-empty?

table/format.cc
```diff
@@ -403,4 +404,17 @@ Status UncompressBlockContents(const UncompressionInfo& uncompression_info,
                                                    ioptions, allocator);
 }
 
+Status SanitizeDbHostIdProperty(Env* env, std::string& db_host_id) {
+  if (db_host_id == kDbHostId) {
+    db_host_id.clear();
+    db_host_id.resize(kMaxHostNameLen);
```
pdillinger, line 410: I also find this extremely weird, to the point that you seem to have bugs yourself: using a fixed-length std::string to contain a \0 terminated C-style string. Performance should not be a concern to copy from a stack allocated fixed-size buffer to a std::string.

table/format.cc
```diff
@@ -403,4 +404,17 @@ Status UncompressBlockContents(const UncompressionInfo& uncompression_info,
...
+    Status s = env->GetHostName(&db_host_id.at(0), kMaxHostNameLen);
+    if (!s.ok()) {
+      return s;
+    }
+    db_host_id.at(kMaxHostNameLen - 1) = 0;
```
pdillinger, line 415: Use '\0' for clarity for null termination character

table/format.cc
```diff
@@ -403,4 +404,17 @@ Status UncompressBlockContents(const UncompressionInfo& uncompression_info,
                                                    ioptions, allocator);
 }
 
+Status SanitizeDbHostIdProperty(Env* env, std::string& db_host_id) {
+  if (db_host_id == kDbHostId) {
+    db_host_id.clear();
+    db_host_id.resize(kMaxHostNameLen);
+    Status s = env->GetHostName(&db_host_id.at(0), kMaxHostNameLen);
```
pdillinger, line 411: If GetHostName is really allowed to not include a terminating nul character, that should be documented in the public API. And best I can tell, the only other place we call this function does not account for that presumed behavior, I guess inherited from posix gethostname. Can't we do something a bit more sane, even without changing the Env function declaration?

include/rocksdb/options.h
```diff
@@ -348,6 +348,8 @@ struct DbPath {
   DbPath(const std::string& p, uint64_t t) : path(p), target_size(t) {}
 };
 
+static const std::string kDbHostId = "__hostname__";
```
pdillinger, line 351: This name doesn't leave any room for alternatives. How about `kHostnameForDbHostId`?

table/format.h
```diff
@@ -331,6 +331,11 @@ extern Status UncompressBlockContentsForCompressionType(
     BlockContents* contents, uint32_t compress_format_version,
     const ImmutableCFOptions& ioptions, MemoryAllocator* allocator = nullptr);
 
+static const size_t kMaxHostNameLen = 256;
+
+// Replace db_host_id contents with the real hostname if necessary
+extern Status SanitizeDbHostIdProperty(Env* env, std::string& db_host_id);
```
pdillinger, line 337: Maybe better to call this `ReifyDbHostIdProperty`

db/db_impl/db_impl_open.cc
```diff
@@ -23,6 +23,8 @@
 #include "test_util/sync_point.h"
 #include "util/rate_limiter.h"
 
+#define MAX_HOST_NAME_LEN 256
```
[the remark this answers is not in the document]
pdillinger: Appears stale / unused

HISTORY.md
```diff
@@ -8,6 +8,7 @@
...
 * AdvancedColumnFamilyOptions::force_consistency_checks is now true by default, for more proactive DB corruption detection at virtually no cost (estimated two extra CPU cycles per million on a major production workload). Corruptions reported by these checks now mention "force_consistency_checks" in case a false positive corruption report is suspected and the option needs to be disabled (unlikely). Since existing column families have a saved setting for force_consistency_checks, only new column families will pick up the new default.
+* A new option, db_host_id, has been added to DBOptions. The option defaults to the hostname (FQDN). if set to something else,  the user specified string is written to SST files as a table property. If its empty, the property is not written.
```
pdillinger, line 11: Myself coming from a case of people not fully digesting release notes and their potential impacts...

I suggest starting with the fact that SST files now include hostnames, but an option is added to override/exclude that.

pdillinger (changes requested):
Some things left to fix

table/format.cc
anand1976: Good point.

table/format.cc
anand1976: Yes, a temporary buffer is a better idea.

table/format.cc
anand1976: The current behavior provides an indication that the input buffer length is insufficient, so the caller can retry if desired. Changing the behavior would eliminate that flexibility. I would just document the API clearly, and maybe provide a wrapper inside RocksDB for simplicity.

table/plain/plain_table_builder.cc
```diff
@@ -100,6 +100,10 @@ PlainTableBuilder::PlainTableBuilder(
   properties_.column_family_name = column_family_name;
   properties_.db_id = db_id;
   properties_.db_session_id = db_session_id;
+  properties_.db_host_id = ioptions.db_host_id;
+  if (!ReifyDbHostIdProperty(ioptions_.env, &properties_.db_host_id).ok()) {
+    ROCKS_LOG_WARN(ioptions_.info_log, "db_host_id property will not be set");
+  }
```
mrambacher, line 106: Warning seems pretty harsh for something the user chose to do...

table/sst_file_writer.cc
```diff
@@ -7,6 +7,7 @@
 
 #include <vector>
 
+#include "db/db_impl/db_impl.h"
```
mrambacher, line 10: Why is this necessary?

include/rocksdb/options.h
```diff
@@ -1180,6 +1182,18 @@ struct DBOptions {
...
+  // will be written as a property in every SST file written by the DB (or
+  // by offline writers such as SstFileWriter and RepairDB). It can be useful
+  // for troubleshooting in memory corruption caused by a failing host when
+  // writing a file, by tracing back to the writing host. These corruptions
+  // may not be caught by the checksum since they happen before checksumming.
+  // If left as default, the table writer will substitute it with the actual
+  // hostname when writing the SST file.
+  //
+  // Default: hostname
+  std::string db_host_id = kHostnameForDbHostId;
```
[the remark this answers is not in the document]
anand1976: It feels more intuitive to just set it to an empty string. Good point about the test. Let me add one.

options/options_settable_test.cc
```diff
@@ -226,6 +226,7 @@ TEST_F(OptionsSettableTest, DBOptionsAllFieldsSettable) {
       {offsetof(struct DBOptions, wal_filter), sizeof(const WalFilter*)},
       {offsetof(struct DBOptions, file_checksum_gen_factory),
        sizeof(std::shared_ptr<FileChecksumGenFactory>)},
+      {offsetof(struct DBOptions, db_host_id), sizeof(std::string)},
```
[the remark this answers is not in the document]
anand1976: It is added to the test string. This line here is to skip filling the option address with garbage, which would cause a segfault.

table/sst_file_writer.cc
anand1976: Not needed anymore, since we don't need to sanitize options here anymore.

include/rocksdb/options.h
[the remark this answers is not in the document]
mrambacher: I am fine with empty meaning "do not write one", but that should be part of the comment here.

table/format.cc
```diff
@@ -403,4 +404,27 @@ Status UncompressBlockContents(const UncompressionInfo& uncompression_info,
                                                    ioptions, allocator);
 }
 
+Status GetHostName(Env* env, std::string* result) {
+  char hostname_buf[kMaxHostNameLen];
```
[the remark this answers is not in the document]
anand1976: How would you access the storage though? You'll still need a temporary buffer. In any case, performance is not a concern.

include/rocksdb/env.h
```diff
@@ -452,9 +452,15 @@ class Env {
...
 
-  // Get the current host name.
+  // Get the current host name as a null terminated string iff the string
+  // length is < len. The hostname should otherwise be truncated to len.
   virtual Status GetHostName(char* name, uint64_t len) = 0;
 
+  // Get the current hostname from the given env as a null terminated
+  // string in result. The result may be truncated if the hostname is too
+  // long
+  static Status GetHostName(Env* env, std::string* result);
```
pdillinger, line 462: It seems a little weird to me making this static. If making it non-static, virtual or not, you could use the name `GetHostNameString` to avoid compiler issues in combining overloading and overriding. Up to you.

table/sst_file_reader_test.cc
```diff
@@ -90,6 +90,7 @@ class SstFileReaderTest : public testing::Test {
     if (check_global_seqno) {
       auto properties = reader.GetTableProperties();
       ASSERT_TRUE(properties);
+      ASSERT_FALSE(properties->db_host_id.empty());
```
pdillinger, line 93: I'm not a fan of "copying the code into the unit test" as an oracle, but it might be worth comparing against GetHostName on the Env here. (I'm not too concerned about the case of someone changing their host name during this test case running.) Up to you

pdillinger (approved):
👍

PR #7480, opened 2020-09-30
Add trace_analyzer_test to ASSERT_STATUS_CHECKED list

zhichao-cao:
summary: Add trace_analyzer_test to ASSERT_STATUS_CHECKED list
test plan: ASSERT_STATUS_CHECKED=1 make -j48 trace_analyzer_test

### DIFF
--- a/tools/trace_analyzer_tool.cc
+++ b/tools/trace_analyzer_tool.cc
@@ -1162,15 +1162,18 @@ Status TraceAnalyzer::ReProcessing() {
 
 // End the processing, print the requested results
 Status TraceAnalyzer::EndProcessing() {
+  Status s;
   if (trace_sequence_f_) {
-    trace_sequence_f_->Close();
+    s = trace_sequence_f_->Close();
   }
   if (FLAGS_no_print) {
-    return Status::OK();
+    return s;
   }
   PrintStatistics();
-  CloseOutputFiles();
-  return Status::OK();
+  if (s.ok()) {
+    s = CloseOutputFiles();
+  }
+  return s;
 }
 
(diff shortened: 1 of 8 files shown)


riversand963 (approved):

tools/trace_analyzer_tool.cc
```diff
@@ -1428,7 +1429,7 @@ Status TraceAnalyzer::OpenStatsOutputFiles(const std::string& type,
                          &new_stats.a_qps_f);
   }
 
-  return Status::OK();
+  return s;
```
mrambacher, line 1432: That entire method is kind of screwy.  It keeps calling Open over and over again and never checking the Status.  Of course, if the status was not OK, the program will exit.  I am not sure what is best to do here.

tools/trace_analyzer_tool.cc
```diff
@@ -1162,15 +1162,16 @@ Status TraceAnalyzer::ReProcessing() {
...
   if (trace_sequence_f_) {
-    trace_sequence_f_->Close();
+    s = trace_sequence_f_->Close();
   }
   if (FLAGS_no_print) {
-    return Status::OK();
+    return s;
   }
   PrintStatistics();
   CloseOutputFiles();
```
mrambacher, line 1173: Note that CloseOutputFiles does a bunch of file Close without checking the status returned.  You are probably just never hitting a test here?

zhichao-cao: Let me update the status check in CloseOutputFiles(). Yes, testing does not hit that case.

PR #7481, opened 2020-09-30
Reduce the number of iterations in DBTest.FileCreationRandomFailure

ltamasi:
Summary:
`DBTest.FileCreationRandomFailure` frequently times out during our
continuous test runs. (It's a case of "stress test posing as unit test.")
The patch reduces the number of iterations to avoid this. Note that
the lower numbers are still sufficient to trigger both flushes and
compactions, so test coverage is still the same.

Test Plan:
`make check`


riversand963 (approved):
LGTM.

PR #7483, opened 2020-09-30
[db_iter.cc] Make sure assert(false) handles failures too

riversand963:
In opt mode, assertions are just no-ops. Therefore, we need to report errors instead of just doing an `assert(false)`.

Test plan:
make check


anand1976 (approved):
LGTM

PR #7484, opened 2020-10-01
Add status check enforcement for column_family_test.cc

akankshamahajan15:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/column_family_test.cc
+++ b/db/column_family_test.cc
@@ -77,11 +77,13 @@ class ColumnFamilyTestBase : public testing::Test {
 
   ~ColumnFamilyTestBase() override {
     std::vector<ColumnFamilyDescriptor> column_families;
+#ifndef ROCKSDB_LITE
     for (auto h : handles_) {
       ColumnFamilyDescriptor cfdescriptor;
-      h->GetDescriptor(&cfdescriptor);
+      EXPECT_OK(h->GetDescriptor(&cfdescriptor));
       column_families.push_back(cfdescriptor);
     }
+#endif  // !ROCKSDB_LITE
     Close();
     ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->DisableProcessing();
     Destroy(column_families);
@@ -168,7 +170,7 @@ class ColumnFamilyTestBase : public testing::Test {
   void Close() {
     for (auto h : handles_) {
       if (h) {
-        db_->DestroyColumnFamilyHandle(h);
+        ASSERT_OK(db_->DestroyColumnFamilyHandle(h));
       }
     }
     handles_.clear();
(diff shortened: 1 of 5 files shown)


db/column_family_test.cc
```diff
@@ -280,7 +282,8 @@ class ColumnFamilyTestBase : public testing::Test {
       ColumnFamilyDescriptor desc;
       ASSERT_OK(handles_[cfi]->GetDescriptor(&desc));
       RocksDBOptionsParser::VerifyCFOptions(ConfigOptions(), desc.options,
-                                            current_cf_opt);
+                                            current_cf_opt)
+          .PermitUncheckedError();
```
mrambacher, line 286: Why is this status being ignored?  After spending a lot of time in the Options code, I am especially curious about this one...

mrambacher (approved):
LGTM.

There are several other PRs that are also changing db_impl_write at the same time.  I have suggested a new signature to the "BGError" so that the status does not need to be ignored everywhere.

db/column_family_test.cc
```diff
@@ -77,11 +77,13 @@ class ColumnFamilyTestBase : public testing::Test {
 
   ~ColumnFamilyTestBase() override {
     std::vector<ColumnFamilyDescriptor> column_families;
+#ifndef ROCKSDB_LITE
```
riversand963, line 80: Maybe always call `GetDescriptor`, and check the return value separately.

db/column_family_test.cc
akankshamahajan15: In the test itself it expects the arguments to be invalid so it returns invalid in few cases and OK in other.

akankshamahajan15:
> LGTM.
> 
> There are several other PRs that are also changing db_impl_write at the same time. I have suggested a new signature to the "BGError" so that the status does not need to be ignored everywhere.

Sounds good. I will add the comment as TODO and when its confirmed to add a new signature to return void, I will change the signature here as well.

db/column_family_test.cc
akankshamahajan15: @mrambacher, Oh my bad. This is different. It gives an error that 
"RocksDBOptionsParser::VerifyCFOptions( ConfigOptions(), desc.options, current_cf_opt)
Invalid argument: [RocksDBOptionsParser]: failed the verification on ColumnFamilyOptions::periodic_compaction_seconds--- The specified one is 0 while the persisted one is 18446744073709551614."
Line: https://github.com/facebook/rocksdb/blob/7d472accdca996d7d83ae8ce78ad17f799696926/options/options_parser.cc#L679

akankshamahajan15: I am not sure if this is expected or not.

mrambacher: I do not think that is expected.  I will take a quick look...

mrambacher: The problem is that when you add a ColumnFamily to the DB, the ColumnFamilyOptions get sanitized and some of the default values might change.  If you change the line to this:
      // Need to sanitize the default column family options before comparing them...
      ASSERT_OK(RocksDBOptionsParser::VerifyCFOptions(ConfigOptions(), desc.options,
                                                      SanitizeOptions(dbfull()->immutable_db_options(), current_cf_opt)));

The tests should pass.

akankshamahajan15: Yes that works. Thanks a lot!

riversand963 (approved):

PR #7485, opened 2020-10-01
Remove explicit qualifier when constructor have no argument or multi argument

darionyaphet:


ajkr (changes requested):
I think it's fine to leave there since it's harmless and the number of arguments for some of these constructors might change to one in the future. Then this PR would introduce an unnecessary risk we forget to add it back.

darionyaphet:
> I think it's fine to leave there since it's harmless and the number of arguments for some of these constructors might change to one in the future. Then this PR would introduce an unnecessary risk we forget to add it back.

thanks and closed ~

PR #7488, opened 2020-10-01
Periodically flush info log out of application buffer

ajkr:
This PR schedules a background thread (shared across all DB instances)
to flush info log every ten seconds. This improves debuggability in case
of RocksDB hanging since it ensures the log messages leading up to the hang
will eventually become visible in the log.

The bulk of this PR is moving monitoring/stats_dump_scheduler* to db/periodic_work_scheduler*
and making the corresponding name changes since now the scheduler handles info
log flushing, not just stats dumping.


riversand963 (approved):
The changes LGTM. The test failure seems legit though.

ajkr:
> The changes LGTM. The test failure seems legit though.

Thanks for the fast review! Sorry I forgot to import so now need the internal diff stamped. The test failure looks likely fixed.

PR #7491, opened 2020-10-01
Ribbon: initial (general) algorithms and basic unit test 

pdillinger:
This is intended as the first commit toward a near-optimal alternative to static Bloom filters for SSTs. Stephan Walzer and I have agreed upon the name "Ribbon" for a PHSF based on his linear system construction in "Efficient Gauss Elimination for Near-Quadratic Matrices with One Short Random Block per Row, with Applications" ("SGauss") and my much faster "on the fly" algorithm for gaussian elimination (or for this linear system, "banding"), which can be faster than peeling while also more compact and flexible. See util/ribbon_alg.h for more detailed introduction and background. RIBBON = Rapid Incremental Boolean Banding ON-the-fly

This commit just adds generic (templatized) core algorithms and a basic unit test showing some features, including the ability to construct structures within 2.5% space overhead vs. information theoretic lower bound. (Compare to cache-local Bloom filter's ~50% space overhead -> ~30% reduction anticipated.) This commit does not include the storage scheme necessary to make queries fast, especially for filter queries, nor fractional "result bits", but there is some description already and those implementations will come soon. Nor does this commit add FilterPolicy support, for use in SST files, but that will also come soon.


pdillinger:
Oops, forgot to make sure it compiles with TEST_UINT128_COMPAT

util/ribbon_alg.h
```diff
@@ -0,0 +1,821 @@
...
+      // Compute next solution bit at row i, column j (see derivation below)
+      CoeffRow tmp = state[j] << 1;
+      int bit = BitParity(tmp & cr) ^ ((rr >> j) & 1);
+      tmp |= static_cast<CoeffRow>(bit);
+
+      // Now tmp is solution at column j from row i for next kCoeffBits
+      // more rows. Thus, for valid solution, the dot product of the
+      // solution column with the coefficient row has to equal the result
+      // at that column,
+      //   BitParity(tmp & cr) == ((rr >> j) & 1)
```
jay-zhuang, line 745: Question: how to make sure `BitParity(tmp & cr) == ((rr >> j) & 1)` as `tmp` could be changed for next row?
For example, to simplify, assume `kResultBits = 1`, `r = 2`. then tmp only has 2 bits, the first bit `tmp[0]` (`tmp[x]` means the `x` bit in `tmp`) it makes sure: `(tmp[0]+cr[0])^(tmp[1]+cr[1]) == rr[0]`. But `tmp[1]` is not assigned yet and will be generated later. How do we make sure we don't need to go back and modify `tmp[0]`?

util/ribbon_impl.h
```diff
@@ -0,0 +1,487 @@
...
+  // hash seed, until either successful or unsuccessful with max_seed
+  // (minimum one seed attempted). Returns true if successful. In that
+  // case, use GetSeed() to get the successful seed.
+  template <typename InputIterator>
+  bool ResetAndFindSeedToSolve(Index num_slots, InputIterator begin,
+                               InputIterator end, Seed max_seed) {
+    StandardHasher<TypesAndSettings>::ResetSeed();
+    do {
+      Reset(num_slots);
+      bool success = AddRange(begin, end);
```
jay-zhuang, line 403: What would be the success rate? Say `r=128`, `1 million keys`, `kFactor=1.025`. `rr = 8 bits`. Is there a formula for that?

util/ribbon_impl.h
```diff
@@ -0,0 +1,487 @@
...
+      if (success) {
+        return true;
+      }
+    } while (StandardHasher<TypesAndSettings>::NextSeed(max_seed));
+    // no seed through max_seed worked
+    return false;
```
jay-zhuang, line 409: Question: what's the expectation from the user if it's failed? Increase the `kNumSlots` and retry? If that's the case, a helper function to do that might be helpful.

util/ribbon_alg.h
```diff
@@ -0,0 +1,821 @@
...
+  for (Index i = num_slots; i > 0;) {
+    --i;
+    CoeffRow cr = *const_cast<BandingStorage &>(ss).CoeffRowPtr(i);
+    ResultRow rr = *const_cast<BandingStorage &>(ss).ResultRowPtr(i);
+    // solution row
+    ResultRow sr = 0;
+    for (Index j = 0; j < kResultBits; ++j) {
+      // Compute next solution bit at row i, column j (see derivation below)
+      CoeffRow tmp = state[j] << 1;
+      int bit = BitParity(tmp & cr) ^ ((rr >> j) & 1);
```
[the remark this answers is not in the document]
pdillinger: I believe some people (not sure about compilers) consider it bad style to cast between integral types and bool. And Google style calls for 'int' for small integer values.

But it might generate the same code, with one less static_cast, to do

    bool bit = (parity) != 0;
    tmp |= bit ? CoeffRow{1} : CoeffRow{0};

util/ribbon_impl.h
```diff
@@ -0,0 +1,487 @@
...
+    assert(num_slots >= kCoeffBits);
+    if (num_slots > num_slots_allocated_) {
+      // Do not need to init the memory
+      solution_rows_.reset(new ResultRow[num_slots]);
+      num_slots_allocated_ = num_slots;
+    }
+    num_starts_ = num_starts;
+  }
+
+  Index GetNumStarts() const { return num_starts_; }
```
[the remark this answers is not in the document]
pdillinger: I should add a comment for that. It's because num_starts is on the extreme critical path for queries: it's value is needed (for FastRange) to know where to start probing in memory. We don't want any extra instructions for getting that value.

util/ribbon_alg.h
pdillinger: Our initial back-substitution state is all zeros, though the initial state is unused because that would mean referencing an "out of bounds" slot, and unoccupied banding rows have cr as all zeros.

Each step through here we set new tmp[x+1] (bit x + 1) to old tmp[x] (bit x). Since we're moving backward through the rows, the second bit of tmp (aka tmp[1] or (tmp >> 1) & 1) will hold the value assigned to the "next" (but most recently assigned) row.

util/ribbon_impl.h
```diff
@@ -0,0 +1,487 @@
...
+// concept RehasherTypesAndSettings: like TypesAndSettings but
+// does not require Key or HashFn.
+template <class RehasherTypesAndSettings>
+class StandardRehasherAdapter : public RehasherTypesAndSettings {
+ public:
+  using Hash = typename RehasherTypesAndSettings::Hash;
+  using Key = Hash;
+  using Seed = typename RehasherTypesAndSettings::Seed;
+
+  static Hash HashFn(const Hash& input, Seed seed) {
```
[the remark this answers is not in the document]
pdillinger: XXH3p_len_4to8_64b (used in XXH3p_64bits_withSeed) has some extra mixing that seems to be unnecessary in this application. (Trying to minimize unnecessary hash computation.)

    xxh_u64 const mix64 = len + ((keyed ^ (keyed >> 51)) * PRIME32_1);
    XXH3p_avalanche((mix64 ^ (mix64 >> 47)) * PRIME64_2)

I should probably say that StandardRehasher is not intended for use (at least untested) with hash-sized values that are not already uniformly distributed.

util/ribbon_impl.h
pdillinger: There's no code here for that yet. Other code will come that choses an appropriate factor for appropriate success rate (generally staying above 50%).

util/ribbon_impl.h
pdillinger: I think that's going to be application specific. Increasing kNumSlots could be an approach, but I expect us to fall back on Bloom filter if we find ourselves with extraordinarily bad luck, or something is broken. We should be able to make that reliably extremely rare without approaches like increasing kNumSlots.

util/ribbon_alg.h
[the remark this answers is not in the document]
pdillinger: That new version is generally about the same, except generating one *less* instruction on gcc 4.9.x :)  https://godbolt.org/z/9MP8zT

util/ribbon_alg.h
jay-zhuang: I see, as the last `r` rows is an upper triangular matrix which will be used to set the initial state. I was confused that it is a band matrix.

util/ribbon_alg.h
```diff
@@ -0,0 +1,821 @@
...
+// A natural approach to scaling Ribbon beyond ~10m keys is splitting
+// (or "sharding") the inputs into multiple linear systems with their
+// own hash seeds. This can also help to control peak memory consumption.
+// TODO: much more to come
+//
+// ######################################################################
+// #################### Ribbon on-the-fly banding #######################
+//
+// "Banding" is what we call the process of reducing the inputs to an
+// upper-triangluar r-band matrix ready for finishing a solution with
```
jay-zhuang, line 245: typo: triangular

util/ribbon_alg.h
```diff
@@ -0,0 +1,821 @@
...
+// this ("SGauss"), the awesome properties of their construction enable
+// an even simpler, faster, and more backtrackable algorithm. In simplest
+// terms, the SGauss algorithm requires sorting the inputs by start
+// columns, but it's possible to make Gaussian elimination resemble hash
+// table insertion!
+//
+// The enhanced algorithm is based on these observations:
+// - When processing a coefficient row with first 1 in column j,
+//   - If it's the first at column j to be processed, it can be part of
+//     the banding at row j. (And that descision never overwritten, with
```
jay-zhuang, line 256: typo: decision

util/ribbon_alg.h
```diff
@@ -0,0 +1,821 @@
...
+//   // column to associate with that hash value. (Starts can be chosen
+//   // uniformly or "smash" extra entries into the beginning and end for
+//   // better utilization at those extremes of the structure. Details in
+//   // ribbon.impl.h)
+//   Index GetStart(Hash, Index num_starts) const;
+//
+//   // Given a hash value, return the r-bit sequence of coefficients to
+//   // associate with it. It's generally OK if
+//   //   sizeof(CoeffRow) > sizeof(Hash)
+//   // as long as the hash itself is not too prone to collsions for the
```
jay-zhuang, line 416: typo: collisions

util/ribbon_alg.h
```diff
@@ -0,0 +1,821 @@
...
+  return expected ==
+         SimpleQueryHelper(hasher.GetStart(hash, sss.GetNumStarts()),
+                           hasher.GetCoeffRow(hash), sss);
+}
+
+// #################### InterleavedSolutionStorage ######################
+
+// InterleavedSolutionStorage is row-major at a high level, for good
+// locality, and column-major at a low level, for CPU efficiency
+// especially in filter querys or relatively small number of result bits
```
jay-zhuang, line 806: queries?

util/ribbon_impl.h
```diff
@@ -0,0 +1,503 @@
...
+};
+
+// StandardRehasher (and StandardRehasherAdapter): A variant of
+// StandardHasher that uses the same type for keys as for hashes.
+// This is primarily intended for building a Ribbon filter/PHSF
+// from existing hashes without going back to original inputs in order
+// to apply a different seed. This hasher seeds a 1-to-1 mixing
+// transformation to apply a seed to an existing hash (or hash-sized key).
+//
+// Testing suggests essentially no degredation of solution success rate
```
jay-zhuang, line 239: degradation?

jay-zhuang (approved):
Nice algorithm 👍👍👍

pdillinger:
Thanks, I'll fix those typos (and more) in next PR

PR #7500, opened 2020-10-02
Add few unit test cases in ASSERT_STATUS_CHECKED

akankshamahajan15:
Summary: Add status enforcement for following tests:
  1. import_column_family_test
  2. memory_test
  3. table_test

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -4710,8 +4710,10 @@ Status DBImpl::CreateColumnFamilyWithImport(
 
   import_job.Cleanup(status);
   if (!status.ok()) {
-    DropColumnFamily(*handle);
-    DestroyColumnFamilyHandle(*handle);
+    // TODO: Should handle this error or log this?
+    DropColumnFamily(*handle).PermitUncheckedError();
+    // Always returns Status::OK()
+    assert(DestroyColumnFamilyHandle(*handle).ok());
     *handle = nullptr;
   }
   return status;
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -4710,8 +4710,12 @@ Status DBImpl::CreateColumnFamilyWithImport(
 
   import_job.Cleanup(status);
   if (!status.ok()) {
-    // TODO: Should handle this error or log this?
-    DropColumnFamily(*handle).PermitUncheckedError();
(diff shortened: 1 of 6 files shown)


db/db_impl/db_impl.cc
```diff
@@ -4701,8 +4701,10 @@ Status DBImpl::CreateColumnFamilyWithImport(
 
   import_job.Cleanup(status);
   if (!status.ok()) {
-    DropColumnFamily(*handle);
-    DestroyColumnFamilyHandle(*handle);
+    // TODO: Should handle this error or log this?
+    DropColumnFamily(*handle).PermitUncheckedError();
```
riversand963, line 4705: I think we should handle the error here, and also check `versions_->io_status` if necessary because this method potentially writes to MANIFEST.

akankshamahajan15: If I add ```status = DropColumnFamily(*handle)```, then it will overwrite the previous errored status and new error will be returned(instead of returning previous error).  Shouldn't we log the error instead as we want to drop and delete the column handle after status is not ok?

riversand963: I am thinking about using another `tmp_status` object, instead of assigning to `status`.

riversand963: Logging the new error SGTM

db/db_impl/db_impl.cc
```diff
@@ -4710,8 +4710,19 @@ Status DBImpl::CreateColumnFamilyWithImport(
...
   if (!status.ok()) {
-    DropColumnFamily(*handle);
-    DestroyColumnFamilyHandle(*handle);
+    Status temp_s = DropColumnFamily(*handle);
+    if (!temp_s.ok()) {
+      ROCKS_LOG_ERROR(immutable_db_options_.info_log,
+                      "DropColumnFamily failed with error %s",
+                      temp_s.ToString().c_str());
+    }
+    if (versions_->io_status().IsIOError()) {
```
riversand963, line 4719: Sorry for the confusion. Checked the code again. If `!io_status.ok()`, then `temp_s` must be non-ok, thus we just need to log once.

zhichao-cao (approved):
LGTM, thanks!

PR #7506, opened 2020-10-05
Range Locking: Implementation of range locking

spetrunia:
Range Locking - an implementation based on the locktree library

- Add a RangeTreeLockManager and RangeTreeLockTracker which implement
  range locking using the locktree library.
- Point locks are handled as locks on single-point ranges.
- Add a unit test: range_locking_test


cheng-chang (commented):
The RangeLockManager and RangeLockTracker implementation are specific to the range tree module, so can we move them to the range_tree directory? Hopefully only the `RangeLockManager` and `LockTracker` interface are exposed to `Transaction` and `TransactionDB`.

utilities/transactions/lock/lock_mgr.h
```diff
@@ -0,0 +1,77 @@
...
+  virtual void Resize(uint32_t) = 0;
+  virtual std::vector<DeadlockPath> GetDeadlockInfoBuffer() = 0;
+
+  // TransactionDB will call this at start
+  virtual void init(TransactionDB*){};
+  virtual ~BaseLockMgr() {}
+
+  // PessimisticTransactionDB will call this to determine if the lock manager
+  // supports range locking (this is a dynamic_cast replacement).
+  virtual RangeLockMgr* getRangeLockMgr() { return nullptr; }
```
cheng-chang, line 70: `RangeLockMgr` depends on toku's range tree, so this makes the `BaseLockMgr` depend on toku's range tree too. Ideally, we would have a standalone `LockManager` interface, and implement `PointLockManager` and `RangeTreeLockManager` separately.

spetrunia: Ok, I'm trying to do that. (part #1 is here: 
https://github.com/facebook/rocksdb/pull/7506/commits/e32cc90100c22e578f44e188300da94a312144c2 )

cheng-chang:
How about opening a separate PR for codes under the `lib` directory? We can merge that PR quickly.
Then after merging https://github.com/facebook/rocksdb/pull/7443, the remaining contents of this PR should be easier to review.

spetrunia:
> How about opening a separate PR for codes under the `lib` directory? We can merge that PR quickly.
> Then after merging #7443, the remaining contents of this PR should be easier to review.

Ok, will do

spetrunia:
@cheng-chang , ok I've created https://github.com/facebook/rocksdb/pull/7753 with just the library code.

cheng-chang:
@spetrunia the PRs for interface and lib are being merged. After they land in master branch, can you create a new PR for the last remaining part? The commit message in this PR is not that clean.

utilities/transactions/transaction_base.h
```diff
@@ -330,8 +330,14 @@ class TransactionBaseImpl : public Transaction {
   // Optimistic Transactions will keep note the requested locks (not actually
   // locked), and do conflict checking until commit time based on the tracked
   // lock requests.
+  //
+  // Declared as public because
+  //  - RangeLocking's lock escalation may replace it
```
cheng-chang, line 335: I didn't see that it's replaced anywhere?

utilities/transactions/transaction_base.h
```diff
@@ -330,8 +330,14 @@ class TransactionBaseImpl : public Transaction {
...
+  //  - RangeTreeLockManager::UnLock needs access to it
```
cheng-chang, line 336: can we have a public getter method instead of exposing this member directly as public?

utilities/transactions/pessimistic_transaction_db.h
```diff
@@ -172,7 +175,7 @@ class PessimisticTransactionDB : public TransactionDB {
   friend class WriteUnpreparedTransactionTest_RecoveryTest_Test;
   friend class WriteUnpreparedTransactionTest_MarkLogWithPrepSection_Test;
 
-  std::unique_ptr<LockManager> lock_manager_;
+  std::shared_ptr<LockManager> lock_manager_;
```
cheng-chang, line 178: who needs to share this lockmanager with?

cheng-chang:
@spetrunia The PRs for interface and lib have been merged to master branch. Can you update this PR?

spetrunia:
@cheng-chang , Ok I've re-based the PR against the current master, but I haven't yet acted on your last input. This remains on my todo.

utilities/transactions/pessimistic_transaction_db.h
spetrunia: It shares it with TransactionDBOptions::lock_mgr_handle.
The workflow is:

* the user passes the manager in TransactionDBOptions::lock_mgr_handle
* TransactionDB gets another pointer to it in NewLockManager, here:

https://github.com/spetrunia/rocksdb/blob/1dfeb2bfa089adc78378459b40650b5b5f3288b7/utilities/transactions/lock/lock_manager.cc#L17

utilities/transactions/transaction_base.h
spetrunia: Ok, the pointed object (RangeTreeLockTracker) is not replaced, but it's contents are modified in `RangeTreeLockManager::on_escalate`  (under a mutex):
https://github.com/spetrunia/rocksdb/blob/1dfeb2bfa089adc78378459b40650b5b5f3288b7/utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc#L375

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  ASSERT_TRUE(s.ok());
+
+  // txn2
+  auto txn2 = NewTxn();
+  txn2->SetSnapshot();
+  std::string val2;
+  // This will see the original value as nothing is committed
+  // This is also Get, so it is doesn't acquire any locks.
+  s = txn2->Get(ReadOptions(), cfh, key_slice, &val2);
+  ASSERT_TRUE(s.ok());
```
cheng-chang, line 172: can we assert that `val2 == initial` here?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+
+  auto txn0 = NewTxn();
+  txn0->Put(key_slice, Slice("initial"));
+  txn0->Commit();
+
+  // txn1
+  auto txn1 = NewTxn();
+  txn1->SetSnapshot();
+  std::string val1;
+  s = txn1->Get(ReadOptions(), cfh, key_slice, &val1);
```
cheng-chang, line 158: can we assert the value of val1 here?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  val2 = val2 + std::string("-txn2");
+  // Now, this call should do Snapshot Validation and fail:
+  s = txn2->Put(cfh, key_slice, Slice(val2));
+  ASSERT_TRUE(s.IsBusy());
+
+  s = txn2->Commit();
+  ASSERT_TRUE(s.ok());
+
+  /*
+    // Not meaningful if s.IsBusy() above is true:
```
cheng-chang, line 188: is this block still useful? I assume it was here for debugging?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
+#ifndef ROCKSDB_LITE
+#ifndef OS_WIN
+
+#ifndef __STDC_FORMAT_MACROS
+#define __STDC_FORMAT_MACROS
```
cheng-chang, line 5: what's this used for?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  // Get a range lock
+  s = txn0->GetRangeLock(cf, Endpoint("z"), Endpoint("z"));
+  ASSERT_EQ(s, Status::OK());
+
+  s = txn1->GetRangeLock(cf, Endpoint("b"), Endpoint("e"));
+  ASSERT_EQ(s, Status::OK());
+
+  auto s2 = range_lock_mgr->GetRangeLockStatusData();
+
+  ASSERT_EQ(s2.size(), 2);
```
cheng-chang, line 218: can we check the contents of the returned status here?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  WriteOptions write_options;
+  TransactionOptions txn_options;
+  auto cf = db->DefaultColumnFamily();
+  Status s;
+
+  Transaction* txn0 = db->BeginTransaction(write_options, txn_options);
+  Transaction* txn1 = db->BeginTransaction(write_options, txn_options);
+
+  // Get a range lock
+  s = txn0->GetRangeLock(cf, Endpoint("z"), Endpoint("z"));
```
cheng-chang, line 210: can we just use `ASSERT_OK(...)` here?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  ASSERT_EQ(s, Status::OK());
+
+  s = txn1->GetRangeLock(cf, Endpoint("b"), Endpoint("e"));
```
cheng-chang, line 213: can we just use `ASSERT_OK(...)` here so that we do not need `s`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+
+TEST_F(RangeLockingTest, MyRocksLikeUpdate) {
+  WriteOptions write_options;
+  TransactionOptions txn_options;
+  Transaction* txn0 = db->BeginTransaction(write_options, txn_options);
+  auto cf = db->DefaultColumnFamily();
+  Status s;
+
+  // Get a range lock for the range we are about to update
+  s = txn0->GetRangeLock(cf, Endpoint("a"), Endpoint("c"));
```
cheng-chang, line 121: `ASSERT_OK` without `s`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  bool try_range_lock_called = false;
+
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->SetCallBack(
+      "RangeTreeLockManager::TryRangeLock:enter",
+      [&](void* /*arg*/) { try_range_lock_called = true; });
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->EnableProcessing();
+
+  // For performance reasons, the following must NOT call lock_mgr->TryLock():
+  // We verify that by checking the value of try_range_lock_called.
+  s = txn0->Put(cf, Slice("b"), Slice("value"), /*assume_tracked=*/true);
```
cheng-chang, line 133: `ASSERT_OK` without `s`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  ASSERT_EQ(s, Status::OK());
+
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->DisableProcessing();
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->ClearAllCallBacks();
+  ASSERT_EQ(try_range_lock_called, false);
```
cheng-chang, line 138: `ASSERT_FALSE`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  std::string value;
+  ReadOptions read_options;
+  auto cf = db->DefaultColumnFamily();
+
+  Transaction* txn0 = db->BeginTransaction(write_options, txn_options);
+  Transaction* txn1 = db->BeginTransaction(write_options, txn_options);
+
+  // Get a range lock
+  {
+    auto s = txn0->GetRangeLock(cf, Endpoint("a"), Endpoint("c"));
```
cheng-chang, line 81: `ASSERT_OK`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+
+  // Check that range Lock inhibits an overlapping point lock
+  {
+    auto s = txn1->GetForUpdate(read_options, cf, Slice("b"), &value);
+    ASSERT_TRUE(s.IsTimedOut());
+  }
+
+  // Get a point lock, check that it inhibits range locks
+  {
+    auto s = txn0->Put(cf, Slice("n"), Slice("value"));
```
cheng-chang, line 99: `ASSERT_OK`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+    options.create_if_missing = true;
+    dbname = test::PerThreadDBPath("range_locking_testdb");
+
+    DestroyDB(dbname, options);
+    Status s;
+
+    range_lock_mgr.reset(NewRangeLockManager(nullptr));
+    txn_db_options.lock_mgr_handle = range_lock_mgr;
+
+    s = TransactionDB::Open(options, txn_db_options, dbname, &db);
```
cheng-chang, line 47: `ASSERT_OK`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  ASSERT_EQ(0, range_lock_mgr->SetMaxLockMemory(2000));
+
+  // Insert until we see lock escalations
+  auto txn0 = NewTxn();
+
+  // Get the locks until we hit an escalation
+  for (int i = 0; i < 2020; i++) {
+    char buf[32];
+    snprintf(buf, sizeof(buf) - 1, "%08d", i);
+    auto s = txn0->GetRangeLock(cf, Endpoint(buf), Endpoint(buf));
```
cheng-chang, line 241: `ASSERT_OK`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  // Initially not using any lock memory
+  ASSERT_EQ(counters.current_lock_memory, 0);
+  ASSERT_EQ(counters.escalation_count, 0);
+
+  ASSERT_EQ(0, range_lock_mgr->SetMaxLockMemory(2000));
+
+  // Insert until we see lock escalations
+  auto txn0 = NewTxn();
+
+  // Get the locks until we hit an escalation
```
cheng-chang, line 237: It's not very clear to me when an escalation happens.
Does it happen when the lock memory threshold is exceeded?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+  auto counters = range_lock_mgr->GetStatus();
+
+  // Initially not using any lock memory
+  ASSERT_EQ(counters.current_lock_memory, 0);
+  ASSERT_EQ(counters.escalation_count, 0);
+
+  ASSERT_EQ(0, range_lock_mgr->SetMaxLockMemory(2000));
+
+  // Insert until we see lock escalations
+  auto txn0 = NewTxn();
```
cheng-chang, line 235: nit: there is no `txn1`, just `txn`?

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
...
+
+  // Get the locks until we hit an escalation
+  for (int i = 0; i < 2020; i++) {
+    char buf[32];
+    snprintf(buf, sizeof(buf) - 1, "%08d", i);
+    auto s = txn0->GetRangeLock(cf, Endpoint(buf), Endpoint(buf));
+    ASSERT_EQ(s, Status::OK());
+  }
+  counters = range_lock_mgr->GetStatus();
+  ASSERT_GE(counters.escalation_count, 0);
```
cheng-chang, line 245: can we assert the count to be 1 here?

utilities/transactions/transaction_base.h
```diff
@@ -250,6 +250,8 @@ class TransactionBaseImpl : public Transaction {
 
   WriteBatch* GetCommitTimeWriteBatch() override;
 
+  LockTracker& GetTrackedLocks() { return *tracked_locks_; }
```
cheng-chang, line 253: can we just expose it as `const`? `tracked_locks_` should only be updated internally by the transaction? Also, can we comment on the thread-safety of the returned `LockTracker`?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h
```diff
@@ -0,0 +1,128 @@
+//
```
cheng-chang, line 1: nit: missing license

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
```diff
@@ -0,0 +1,65 @@
+//
```
cheng-chang, line 1: nit: missing license

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
+#ifndef ROCKSDB_LITE
```
cheng-chang, line 1: nit: missing license

utilities/transactions/lock/range/range_locking_test.cc
```diff
@@ -0,0 +1,301 @@
+#ifndef ROCKSDB_LITE
```
cheng-chang, line 1: nit: missing license

utilities/transactions/lock/range/range_locking_test.cc
spetrunia: This seems to be something no longer needed. I'll remove these.

utilities/transactions/lock/range/range_locking_test.cc
cheng-chang: Is there an API to trigger lock escalation manually?

cheng-chang (commented):
Can we remove the file `range_tree/TODO`?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+#include "util/mutexlock.h"
+#include "utilities/transactions/lock/lock_tracker.h"
+
+// Range Locking:
+#include "lib/locktree/lock_request.h"
+#include "lib/locktree/locktree.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+class RangeLockList;
```
cheng-chang, line 21: `RangeLockList` is declared below, why is this forward declaration needed?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+
+/*
```
cheng-chang, line 23: nit: can we use the `//` comment style consistently?

utilities/transactions/lock/range/range_locking_test.cc
spetrunia: Ok, will do

utilities/transactions/lock/range/range_locking_test.cc
spetrunia: Yes, done.

utilities/transactions/lock/range/range_locking_test.cc
spetrunia: Done

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+
+  Locks are kept in toku::range_buffer because toku::locktree::release_locks()
+  accepts that as an argument.
+
+  Note: the list of locks may differ slighly from the contents of the lock
+  tree, due to concurrency between lock acquisition, lock release, and lock
+  escalation. See MDEV-18227 and RangeTreeLockManager::UnLock for details.
+  This property is currently harmless.
+*/
+class RangeLockList /*: public PessimisticTransaction::Lock--Storage */ {
```
cheng-chang, line 34: remove the comment?

utilities/transactions/lock/range/range_locking_test.cc
spetrunia: Yes. It happens when the total amount of memory used for locks exceeds the limit specified by `SetMaxLockMemory`.
The memory is used for lock endpoints (whose size is known) + sizeof(certain_data_structures) so it is not possible to predict how many locks exactly one needs to take before they hit the limit.

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+      it = buffers_
+               .emplace(cf_id, std::shared_ptr<toku::range_buffer>(
+                                   new toku::range_buffer()))
+               .first;
+      it->second->create();
+    }
+    it->second->append(left_key, right_key);
+  }
+
+  std::unordered_map<uint32_t, std::shared_ptr<toku::range_buffer>> buffers_;
```
cheng-chang, line 64: why are these member variables public?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+  tree, due to concurrency between lock acquisition, lock release, and lock
+  escalation. See MDEV-18227 and RangeTreeLockManager::UnLock for details.
+  This property is currently harmless.
+*/
+class RangeLockList /*: public PessimisticTransaction::Lock--Storage */ {
+ public:
+  virtual ~RangeLockList() { clear(); }
+
+  void clear() {
+    for (auto it : buffers_) {
```
cheng-chang, line 39: should we lock `mutex_` here? What's the thread-safety concern here?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+  Note: the list of locks may differ slighly from the contents of the lock
+  tree, due to concurrency between lock acquisition, lock release, and lock
+  escalation. See MDEV-18227 and RangeTreeLockManager::UnLock for details.
+  This property is currently harmless.
+*/
+class RangeLockList /*: public PessimisticTransaction::Lock--Storage */ {
+ public:
+  virtual ~RangeLockList() { clear(); }
+
+  void clear() {
```
cheng-chang, line 38: nit:`Clear` to keep style consistent?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+    for (auto it : buffers_) {
+      it.second->destroy();
+    }
+    buffers_.clear();
+  }
+
+  RangeLockList() : releasing_locks_(false) {}
+
+  void append(uint32_t cf_id, const DBT* left_key, const DBT* right_key) {
```
cheng-chang, line 47: nit: `Append` to keep style consistent?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  std::unordered_map<uint32_t, std::shared_ptr<toku::range_buffer>> buffers_;
+
+  // Synchronization. See RangeTreeLockManager::UnLock for details
+  port::Mutex mutex_;
+  bool releasing_locks_;
+};
+
+// Tracks range locks
```
cheng-chang, line 71: the comment is not very useful, maybe just remove it?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+      ColumnFamilyId /*column_family_id*/) const override {
+    return nullptr;
+  }
+
+  // Non-override (TODO: make private!)
+  RangeLockList* getList() { return range_list.get(); }
+  RangeLockList* getOrCreateList();
+
+ private:
+  std::shared_ptr<RangeLockList> range_list;
```
cheng-chang, line 132: nit: `range_list_` to keep style consistent?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+  ColumnFamilyIterator* GetColumnFamilyIterator() const override {
+    return nullptr;
+  }
+
+  KeyIterator* GetKeyIterator(
+      ColumnFamilyId /*column_family_id*/) const override {
+    return nullptr;
+  }
+
+  // Non-override (TODO: make private!)
```
cheng-chang, line 127: no need to be public?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
```diff
@@ -0,0 +1,65 @@
...
+// A lock tracker to be used together with RangeTreeLockManager
+//
+#ifndef ROCKSDB_LITE
+#ifndef OS_WIN
+
+#include "range_tree_lock_tracker.h"
+
```
cheng-chang, line 8: unnecessary blank line?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
```diff
@@ -0,0 +1,65 @@
...
+  std::string start_key, end_key;
+
+  serialize_endpoint(lock_req.start_endp, &start_key);
+  serialize_endpoint(lock_req.end_endp, &end_key);
+
+  toku_fill_dbt(&start_dbt, start_key.data(), start_key.size());
+  toku_fill_dbt(&end_dbt, end_key.data(), end_key.size());
+
+  RangeLockList *rl = getOrCreateList();
+  rl->append(lock_req.column_family_id, &start_dbt, &end_dbt);
```
cheng-chang, line 42: `start_dbt` and `end_dbt` are destroyed after `Track`. Here we pass their addresses to `rl`, is this safe since after `Track`, the addresses are no longer meaningful?

utilities/transactions/lock/range/range_locking_test.cc
spetrunia: Wait, we can't use `ASSERT_OK` here, because it's a macros and it tries to return a value while this is a constructor.

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
cheng-chang: also, `range_list` is only owned by `LockTracker`? If so, should we make it `unique_ptr`?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
```diff
@@ -0,0 +1,65 @@
...
+  // "Cheat" and return the status which says the point is not locked.
+  PointLockStatus p;
+  p.locked = false;
+  p.exclusive = true;
+  p.seq = 0;
+  return p;
+}
+
+void RangeTreeLockTracker::Clear() {
+  // This will delete the RangeLockList and cause a proper cleanup
```
cheng-chang, line 59: it's not very clear why a "proper cleanup" will be performed here. Maybe make it `unique_ptr` and `reset` here?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
```diff
@@ -0,0 +1,65 @@
...
+  toku_fill_dbt(&start_dbt, start_key.data(), start_key.size());
+  toku_fill_dbt(&end_dbt, end_key.data(), end_key.size());
+
+  RangeLockList *rl = getOrCreateList();
+  rl->append(lock_req.column_family_id, &start_dbt, &end_dbt);
+}
+
+PointLockStatus RangeTreeLockTracker::GetPointLockStatus(
+    ColumnFamilyId /*cf_id*/, const std::string & /*key*/) const {
+  // TODO: what to do here if we are holding a range lock that is embedding the
```
cheng-chang, line 47: If there is a range lock containing the key, then from point lock's perspective, the point is locked?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+  Storage for locks that are currently held by a transaction.
+
+  Locks are kept in toku::range_buffer because toku::locktree::release_locks()
+  accepts that as an argument.
+
+  Note: the list of locks may differ slighly from the contents of the lock
+  tree, due to concurrency between lock acquisition, lock release, and lock
+  escalation. See MDEV-18227 and RangeTreeLockManager::UnLock for details.
+  This property is currently harmless.
+*/
```
cheng-chang, line 33: add a comment on the thread safety?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
```diff
@@ -0,0 +1,148 @@
...
+  std::unordered_map<uint32_t, std::shared_ptr<toku::range_buffer>> buffers_;
+
+  // Synchronization. See RangeTreeLockManager::UnLock for details
+  port::Mutex mutex_;
+  bool releasing_locks_;
+};
+
+// Tracks range locks
+
+class RangeTreeLockTracker : public LockTracker {
```
cheng-chang, line 73: add a comment on the thread safety?

utilities/transactions/transaction_base.h
spetrunia: Yes. Done.

utilities/transactions/lock/range/range_locking_test.cc
spetrunia: Right. Removed it.

utilities/transactions/lock/range/range_locking_test.cc
spetrunia: Currently, no. It is because of the following property:
- Initially, there is only one transaction, and locktree uses a "Single Transaction Optimization" (STO) which uses a compact data structure (basically, an unsorted array of endpoints) to store the locks.
- When it hits the memory limit for the first time, it switches off the STO mode.
- Now, it is using a much less compact data structure (but one which allows searches and doesn't assume all locks have the same owner).
- Because of that, it quickly hits the limit again. 

So, the count is bigger than 1 here.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h
```diff
@@ -0,0 +1,128 @@
...
+#include "lib/locktree/locktree.h"
+#include "range_tree_lock_tracker.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+using namespace toku;
+
+typedef DeadlockInfoBufferTempl<RangeDeadlockPath> RangeDeadlockInfoBuffer;
+
+/*
```
cheng-chang: nit: use `//` style comment?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h
```diff
@@ -0,0 +1,128 @@
...
+
+  void Resize(uint32_t) override;
+  std::vector<DeadlockPath> GetDeadlockInfoBuffer() override;
+
+  std::vector<RangeDeadlockPath> GetRangeDeadlockInfoBuffer() override;
+  void SetRangeDeadlockInfoBufferSize(uint32_t target_size) override;
+
+  // Get a lock on a range
+  //  @note only exclusive locks are currently supported (requesting a
+  //  non-exclusive lock will get an exclusive one)
```
cheng-chang, line 41: what about returning an error status when requesting a non-exclusive one?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+#include "test_util/sync_point.h"
+#include "util/cast_util.h"
+#include "util/hash.h"
+#include "util/thread_local.h"
+#include "utilities/transactions/pessimistic_transaction_db.h"
+#include "utilities/transactions/transaction_db_mutex_impl.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+/////////////////////////////////////////////////////////////////////////////
```
cheng-chang, line 23: unnecessary comment?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+  serialize_endpoint(end_endp, &end_key);
+
+  toku_fill_dbt(&start_key_dbt, start_key.data(), start_key.size());
+  toku_fill_dbt(&end_key_dbt, end_key.data(), end_key.size());
+
+  // Use the txnid as "extra" in the lock_request. Then, KillLockWait()
+  // will be able to use kill_waiter(txn_id) to kill the wait if needed
+  // TODO: KillLockWait is gone and we are no longer using
+  // locktree::kill_waiter call. Do we need this anymore?
+  TransactionID wait_txn_id = txn->GetID();
```
cheng-chang, line 80: is this still needed?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+
+  auto lt = get_locktree_by_cfid(column_family_id);
+
+  request.set(lt, (TXNID)txn, &start_key_dbt, &end_key_dbt,
+              exclusive ? toku::lock_request::WRITE : toku::lock_request::READ,
```
cheng-chang, line 85: The comment mentioned that non-exclusive lock is not supported. Then what's the semantics of `READ` here?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+              false /* not a big txn */, (void*)wait_txn_id);
+
+  // This is for "periodically wake up and check if the wait is killed" feature
+  // which we are not using.
+  uint64_t killed_time_msec = 0;
+  uint64_t wait_time_msec = txn->GetLockTimeout();
+
+  // convert microseconds to milliseconds
+  if (wait_time_msec != (uint64_t)-1)
+    wait_time_msec = (wait_time_msec + 500) / 1000;
```
cheng-chang, line 95: even with the `if` check, `wait_time_msec + 500` is still possible to overflow?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+
+  std::vector<RangeDeadlockInfo> di_path;
+  request.m_deadlock_cb = [&](TXNID txnid, bool is_exclusive,
+                              const DBT* start_dbt, const DBT* end_dbt) {
+    EndpointWithString start;
+    EndpointWithString end;
+    deserialize_endpoint(start_dbt, &start);
+    deserialize_endpoint(end_dbt, &end);
+
+    di_path.push_back({((PessimisticTransaction*)txnid)->GetID(),
```
cheng-chang, line 105: can we use `emplace_back` and get rid of the `{}` construction inside?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+      dlock_buffer_.AddNewPath(
+          RangeDeadlockPath(di_path, request.get_start_time()));
+      return Status::Busy(Status::SubCode::kDeadlock);
+    }
+    default:
+      assert(0);
+      return Status::Busy(Status::SubCode::kLockLimit);
+  }
+
+  // Don't: save the acquired lock in txn->owned_locks.
```
cheng-chang, line 171: remove the dead code?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+    std::string* key_ptr;
+    autovector<TransactionID> wait_ids;
+    bool done = false;
+
+    static void lock_wait_callback(void* cdata, TXNID /*waiter*/,
+                                   TXNID waitee) {
+      TEST_SYNC_POINT("RangeTreeLockManager::TryRangeLock:WaitingTxn");
+      auto self = (struct st_wait_info*)cdata;
+      // we know that the waiter is self->txn->GetID() (TODO: assert?)
+      if (!self->done) {
```
cheng-chang, line 130: will the callback be called multiple times? Maybe each time on a new waitee?

cheng-chang: What if txn 0 needs to wait for txn 1, then when txn 1 commits, txn 0 needs to wait for txn 2?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+  // Don't: save the acquired lock in txn->owned_locks.
+  // It is now responsibility of RangeTreeLockManager
+  //  RangeLockList* range_list=
+  //  ((RangeTreeLockTracker*)txn->tracked_locks_.get())->getOrCreateList();
+  //  range_list->append(column_family_id, &start_key_dbt, &end_key_dbt);
+
+  return Status::OK();
+}
+
+static void range_lock_mgr_release_lock_int(toku::locktree* lt,
```
cheng-chang, line 180: what's the meaning of `int`? Does it mean intention?

spetrunia: "internal".  I see this function is called only from one place now, so I'll remove it.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+
+  // tracked_locks_->range_list may hold nullptr if the transaction has never
+  // acquired any locks.
+  RangeLockList* range_list = ((RangeTreeLockTracker*)range_tracker)->getList();
+
+  if (range_list) {
+    {
+      MutexLock l(&range_list->mutex_);
+      /*
+        The lt->release_locks() call below will walk range_list->buffer_. We
```
cheng-chang: oh, this is too tricky

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
spetrunia: It is safe, because the code inside RangeLockList::append() will copy the contents of DBT structures. It will not retain pointers to them.
The copying happens here:
```
  #0  toku::range_buffer::append_range (...) at utilities/transactions/lock/range/range_tree/lib/locktree/range_buffer.cc:235
  #1  0x0000555555aeea72 in toku::range_buffer::append (...) at utilities/transactions/lock/range/range_tree/lib/locktree/range_buffer.cc:205
  #2  0x00005555559e614c in rocksdb::RangeLockList::Append (...) at utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc:76
  #3  0x00005555559e5e77 in rocksdb::RangeTreeLockTracker::Track (...) at utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc:41
  #4  0x0000555555af85eb in rocksdb::PessimisticTransaction::GetRangeLock (...) at utilities/transactions/pessimistic_transaction.cc:682
```
https://github.com/facebook/rocksdb/blob/master/utilities/transactions/lock/range/range_tree/lib/locktree/range_buffer.cc#L235

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+  if (ltree_map_.find(column_family_id) == ltree_map_.end()) {
+    DICTIONARY_ID dict_id = {.dictid = column_family_id};
+    toku::comparator cmp;
+    cmp.create(compare_dbt_endpoints, (void*)cfh->GetComparator());
+    toku::locktree* ltree = ltm_.get_lt(dict_id, cmp,
+                                        /* on_create_extra*/ nullptr);
+    // This is ok to because get_lt has copied the comparator:
+    cmp.destroy();
+    ltree_map_.emplace(column_family_id, ltree);
+  } else {
```
cheng-chang, line 364: remove this dead block?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h
spetrunia: I think returning a non-exclusive lock is better.  It still meets the call contract  (and with lock escalation, there is no guarantee that the lock is obtained on the exact range requested. A broader lock can be acquired, instead).

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
spetrunia: We don't support shared range locks, but we do support shared point locks.  

(Rationale for this: initially, locktree didn't support shared locks, and so Range Locking code didn't support it either.  Then, there was a use case for MyRocks which required that shared point locks are supported. So, limited support for shared point locks was added. Sharing point locks are much easier as one doesn't have to deal with locks on partially-overlapping ranges.)

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+
+  return nullptr;
+}
+
+struct LOCK_PRINT_CONTEXT {
+  RangeLockManagerHandle::RangeLockStatus* data;  // Save locks here
+  uint32_t cfh_id;  // Column Family whose tree we are traversing
+};
+
+// Report left endpoints of the acquired locks
```
cheng-chang, line 434: what will be the usage of the left endpoints?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+  it = ltree_map_.find(column_family_id);
+  if (it == ltree_map_.end()) {
+    return nullptr;
+  } else {
+    // Found lock map.  Store in thread-local cache and return.
+    ltree_map_cache->insert({column_family_id, it->second});
+    return it->second;
+  }
+
+  return nullptr;
```
cheng-chang, line 426: unnecessary?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+  ctx->data->insert({ctx->cfh_id, info});
+}
+
+LockManager::RangeLockStatus RangeTreeLockManager::GetRangeLockStatus() {
+  LockManager::RangeLockStatus data;
+  {
+    InstrumentedMutexLock l(&ltree_map_mutex_);
+    for (auto it : ltree_map_) {
+      LOCK_PRINT_CONTEXT ctx = {&data, it.first};
+      it.second->dump_locks((void*)&ctx, push_into_lock_status_data);
```
cheng-chang, line 475: looks like `LOCK_PRINT_CONTEXT` is defined inside the tokudb library? Otherwise, how can the library know how to interpret the `void*`?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+    ltree_map_.erase(lock_maps_iter);
+  }  // lock_map_mutex_
+
+  // TODO: why do we delete first and clear the caches second? Shouldn't this be
+  // done in the reverse order? (if we do it in the reverse order, how will we
+  // prevent somebody from re-populating the cache?)
+
+  // Clear all thread-local caches. We collect a vector of caches but we dont
+  // really need them.
+  autovector<void*> local_caches;
```
cheng-chang, line 395: what if a lock tree is already released by `ltm_.release_lt`, but not cleared from thread local cache yet, and someone accesses it from the local cache?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+    ltree_map_.emplace(column_family_id, ltree);
+  } else {
+    // column_family already exists in lock map
+    // assert(false);
+  }
+}
+
+void RangeTreeLockManager::RemoveColumnFamily(const ColumnFamilyHandle* cfh) {
+  uint32_t column_family_id = cfh->GetID();
+  // Remove lock_map for this column family.  Since the lock map is stored
```
cheng-chang, line 371: it's a map from `uint32_t` (cf id) to `locktree*`, which shared ptr are you referring to?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+/*
+  @brief  Lock Escalation Callback function
+
+  @param txnid   Transaction whose locks got escalated
+  @param lt      Lock Tree where escalation is happening
+  @param buffer  Escalation result: list of locks that this transaction now
+                 owns in this lock tree.
+  @param void*   Callback context
+*/
+
```
cheng-chang: nit: remove empty line

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+    for (auto it2 = it->path.begin(); it2 != it->path.end(); ++it2) {
+      path.push_back(
+          {it2->m_txn_id, it2->m_cf_id, it2->m_exclusive, it2->m_start.slice});
+    }
+    res.push_back(DeadlockPath(path, it->deadlock_time));
+  }
+  return res;
+}
+
+/*
```
cheng-chang: nit: use `//` style for comment consistently?

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,572 @@
...
+        lt->release_locks((TXNID)txn, it.second.get(), all_keys);
+
+        it.second->destroy();
+        it.second->create();
+
+        toku::lock_request::retry_all_lock_requests(lt, nullptr);
+      }
+    }
+    range_list->clear();  // TODO: need this?
+    range_list->releasing_locks_ = false;
```
cheng-chang: I thought modifying releasing_locks_ needs to lock the mutex?

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
spetrunia: Ok, moved other code around and that allowed to make them private.

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
spetrunia: It is locked, in a sense that an attempt to get the lock on a key will fail.
But it is not "tracked", in a sense that we haven't made a note of what was its SequenceNumber when the range lock was acquired.

I don't think this comment is relevant as `class RangeTreeLockTracker` has  `RangeTreeLockTracker::IsPointLockSupported()` which returns FALSE, so there shouldn't be anybody calling this method.

(and providing an implementation that checks if there's a lock on a range that contains this point - this is going to be hard as the RangeLockList that stores the locks we've acquired is implemented as unsorted array, which doesn't allow lookups).

I think I'll just modify the comment to say that this function is not expected to be implemented.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
spetrunia: This doesn't seem to work, as `RangeDeadlockInfo` doesn't have a constructor. 

It's the same with point-based `DeadlockInfo`.  It is also filled with `push_back`, not `emplace_back`:
https://github.com/facebook/rocksdb/blob/master/utilities/transactions/lock/point/point_lock_manager.cc#L430

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
spetrunia: In this operation, no.
The value is obtained like so
```
  uint64_t wait_time_msec = txn->GetLockTimeout();
```
where txn->GetLockTimeout() returns int64_t, a signed value. The only possible negative value (with highest bit set)  is "-1" which means "no timeout". 


But I do see an issue with `wait_time_msec=-1`.  We will pass that value to the `request.wait(wait_time_msec, ...`) call.
`lock_request::wait` has no handling for the case where its `wait_time_ms` parameter is `(unt64_t)-1`: it immediately computes:

```
  uint64_t t_end = t_start + wait_time_ms * 1000;
```
which causes an overflow. This in turn, will cause lock that was acquired immediately to be counted as a "long wait".

I think I can fix that by converting "no timeout" to "wait for one year"

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
spetrunia: No. I'll remove it.

cheng-chang (approved):

cheng-chang:
@spetrunia Thanks for updating the PR, there are still some minor comments remaining to be resolved. Otherwise, LGTM.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
spetrunia: Right.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
spetrunia: Ok, this uncovered a fairly big issue.  locktree library had an API for reporting lock waits in form  "transaction X is now waiting on transaction Y", and RangeTreeLockManager was trying to use it.  But, it didn't provide all the required information, and also RangeTreeLockManager was not using it correctly.  Now all fixed, and test coverage is added.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
spetrunia: Ok, there was a bug there. Now, I've made it work like point-based lock manager does

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.h
spetrunia: No. For this function, thread safety is ensured by the caller. I've now made it private.

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
spetrunia: Done.

utilities/transactions/lock/range/range_tree/lib/locktree/lock_request.cc
```diff
@@ -424,26 +420,20 @@ void lock_request::retry_all_lock_requests_info(
   toku_external_mutex_unlock(&info->mutex);
 }
 
-void lock_request::add_conflicts_to_waits(
-    txnid_set *conflicts, GrowableArray<TXNID> *wait_conflicts) {
+void lock_request::add_conflicts_to_waits(txnid_set *conflicts,
+                                          lock_wait_infos *wait_conflicts) {
+  wait_conflicts->push_back({m_lt, get_txnid(), m_extra});
```
cheng-chang, line 425: missing field 'waitees' initializer

utilities/transactions/lock/range/range_tree/lib/locktree/locktree.cc
```diff
@@ -185,7 +185,10 @@ static bool determine_conflicting_txnids(
     const TXNID other_txnid = lock.txnid;
     if (other_txnid != txnid) {
       if (conflicts) {
-        conflicts->add(other_txnid);
+        if (other_txnid == TXNID_SHARED) {
+          for (TXNID shared_id : *lock.owners) conflicts->add(shared_id);
```
cheng-chang, line 189: missing braces

utilities/transactions/lock/range/range_tree/lib/locktree/locktree.cc
```diff
@@ -185,7 +185,10 @@ static bool determine_conflicting_txnids(
...
+        } else
```
cheng-chang, line 190: missing braces

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h
```diff
@@ -0,0 +1,130 @@
...
+
+// For DeadlockInfoBuffer:
+#include "util/thread_local.h"
+#include "utilities/transactions/lock/point/point_lock_manager.h"
+#include "utilities/transactions/lock/range/range_lock_manager.h"
+
+// Lock Tree library:
+#include "lib/locktree/lock_request.h"
+#include "lib/locktree/locktree.h"
+#include "range_tree_lock_tracker.h"
```
cheng-chang, line 18: Included paths should be relative to the project (e.g. fbcode) root

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h
```diff
@@ -0,0 +1,130 @@
...
+                 bool exclusive) override;
+
+  void UnLock(PessimisticTransaction* txn, const LockTracker& tracker,
+              Env* env) override;
+  void UnLock(PessimisticTransaction* txn, ColumnFamilyId column_family_id,
+              const std::string& key, Env* env) override;
+  void UnLock(PessimisticTransaction*, ColumnFamilyId, const Endpoint&,
+              const Endpoint&, Env*) override{
+      // TODO: range unlock does nothing...
+  };
```
cheng-chang, line 56: extra ';' after member function definition

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h
```diff
@@ -0,0 +1,130 @@
...
+
+  RangeTreeLockManager(
```
cheng-chang, line 58: Single-argument constructor missing explicit specifier.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h
```diff
@@ -0,0 +1,130 @@
...
+      std::shared_ptr<TransactionDBMutexFactory> mutex_factory);
+
+  ~RangeTreeLockManager();
```
cheng-chang, line 61: '~RangeTreeLockManager' overrides a destructor but is not marked 'override'

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+#ifndef OS_WIN
+
+#include "utilities/transactions/lock/range/range_tree/range_tree_lock_manager.h"
+
+#include <algorithm>
+#include <cinttypes>
+#include <mutex>
+
+#include "monitoring/perf_context_imp.h"
+#include "range_tree_lock_tracker.h"
```
cheng-chang, line 16: Included paths should be relative to the project root

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+#include "utilities/transactions/pessimistic_transaction_db.h"
+#include "utilities/transactions/transaction_db_mutex_impl.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+RangeLockManagerHandle* NewRangeLockManager(
+    std::shared_ptr<TransactionDBMutexFactory> mutex_factory) {
+  std::shared_ptr<TransactionDBMutexFactory> use_factory;
+
+  if (mutex_factory)
```
cheng-chang, line 32: if/for/while/do-while blocks should be enclosed in curly braces.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+    use_factory = mutex_factory;
+  else
```
cheng-chang, line 34: if/for/while/do-while blocks should be enclosed in curly braces.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+  request.set(lt.get(), (TXNID)txn, &start_key_dbt, &end_key_dbt,
+              exclusive ? toku::lock_request::WRITE : toku::lock_request::READ,
+              false /* not a big txn */, &wait_key);
+
+  // This is for "periodically wake up and check if the wait is killed" feature
+  // which we are not using.
+  uint64_t killed_time_msec = 0;
+  uint64_t wait_time_msec = txn->GetLockTimeout();
+
+  if (wait_time_msec == (uint64_t)-1) {
```
cheng-chang, line 95: static_cast<uint64_t>(-1)

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+    default:
+      assert(0);
+      return Status::Busy(Status::SubCode::kLockLimit);
+  }
+
+  return Status::OK();
+}
+
+// Wait callback that locktree library will call to inform us about
+// the lock waits taht are in progress.
```
cheng-chang, line 150: taht -> that

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+  size_t min_len = std::min(a_len, b_len);
+
+  // Compare the values. The first byte encodes the endpoint type, its value
+  // is either SUFFIX_INFIMUM or SUFFIX_SUPREMUM.
+  Comparator* cmp = (Comparator*)arg;
+  int res = cmp->Compare(Slice(a + 1, min_len - 1), Slice(b + 1, min_len - 1));
+  if (!res) {
+    if (b_len > min_len) {
+      // a is shorter;
+      if (a[0] == SUFFIX_INFIMUM)
```
cheng-chang, line 219: if/for/while/do-while blocks should be enclosed in curly braces.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+        return -1;  //"a is smaller"
+      else {
+        // a is considered padded with 0xFF:FF:FF:FF...
+        return 1;  // "a" is bigger
+      }
+    } else if (a_len > min_len) {
+      // the opposite of the above: b is shorter.
+      if (b[0] == SUFFIX_INFIMUM)
```
cheng-chang, line 227: if/for/while/do-while blocks should be enclosed in curly braces.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+        return 1;  //"b is smaller"
+      else {
+        // b is considered padded with 0xFF:FF:FF:FF...
+        return -1;  // "b" is bigger
+      }
+    } else {
+      // the lengths are equal (and the key values, too)
+      if (a[0] < b[0])
```
cheng-chang, line 235: if/for/while/do-while blocks should be enclosed in curly braces.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+        return -1;
+      else if (a[0] > b[0])
```
cheng-chang, line 237: if/for/while/do-while blocks should be enclosed in curly braces.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+        return 1;
+      else
```
cheng-chang, line 239: if/for/while/do-while blocks should be enclosed in curly braces.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+        return 0;
+    }
+  } else
```
cheng-chang, line 242: if/for/while/do-while blocks should be enclosed in curly braces.

utilities/transactions/lock/range/range_tree/range_tree_lock_manager.cc
```diff
@@ -0,0 +1,477 @@
...
+  delete lock_tree_map_cache;
+}
+}  // anonymous namespace
+
+RangeTreeLockManager::RangeTreeLockManager(
+    std::shared_ptr<TransactionDBMutexFactory> mutex_factory)
+    : mutex_factory_(mutex_factory),
+      ltree_lookup_cache_(new ThreadLocalPtr(&UnrefLockTreeMapsCache)),
+      dlock_buffer_(10) {
+  ltm_.create(on_create, on_destroy, on_escalate, NULL, mutex_factory_);
```
cheng-chang, line 260: prefer `nullptr` instead of `NULL`

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
```diff
@@ -0,0 +1,159 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#ifndef ROCKSDB_LITE
+#ifndef OS_WIN
+
+#include "range_tree_lock_tracker.h"
```
cheng-chang, line 9: Included paths should be relative to the project root

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
```diff
@@ -0,0 +1,159 @@
...
+
+#include "range_tree_lock_manager.h"
```
cheng-chang, line 11: Included paths should be relative to the project root

utilities/transactions/lock/range/range_tree/range_tree_lock_tracker.cc
```diff
@@ -0,0 +1,159 @@
...
+  MutexLock l(&mutex_);
+  // Only the transaction owner thread calls this function.
+  // The same thread does the lock release, so we can be certain nobody is
+  // releasing the locks concurrently.
+  assert(!releasing_locks_.load());
+  auto it = buffers_.find(cf_id);
+  if (it == buffers_.end()) {
+    // create a new one
+    it = buffers_
+             .emplace(cf_id, std::shared_ptr<toku::range_buffer>(
```
cheng-chang, line 71: prefer `make_shared`

cheng-chang (commented):
Most remaining comments are nits from internal static analysis warnings.

spetrunia:
@cheng-chang , I believe everything needed is done now.

PR #7508, opened 2020-10-05
Fix StallWrite crash with mixed of slowdown/no_slowdown writes

jay-zhuang:
Summary: `BeginWriteStall()` removes no_slowdown write from the write
list and updates `link_newer`, which makes `CreateMissingNewerLinks()`
thought all write list has valid `link_newer` and failed to create link
for all writers.
It caused flaky test and SegFault for release build.

Test Plan: Add unittest to reproduce the issue.


anand1976 (approved):
Nice catch. Thanks for fixing it!

PR #7516, opened 2020-10-07
Do not hold mutex when write keys if not necessary

Little-Wallace:
## Problem Summary
RocksDB will acquire the global mutex of db instance for every time when user calls `Write`.  When RocksDB schedules a lot of compaction jobs,   it will compete the mutex with write thread and it will hurt the write performance. 

## Problem Solution:
I want to use log_write_mutex to replace the global mutex in most case so that we do not acquire it in write-thread unless there is a write-stall event or a write-buffer-full event occur.

## Test plan
1. make check
2. CI
3. COMPILE_WITH_TSAN=1 make db_stress
make crash_test
make crash_test_with_multiops_wp_txn
make crash_test_with_multiops_wc_txn
make crash_test_with_atomic_flush


db/db_impl/db_impl.cc
```diff
@@ -571,6 +571,7 @@ Status DBImpl::CloseHelper() {
     mutex_.Lock();
   }
 
+  log_write_mutex_.Lock();
```
yiwu-arbug, line 574: nit: use scoped `MutexLock`

db/db_impl/db_impl_write.cc
```diff
@@ -893,12 +891,13 @@ void DBImpl::MemTableInsertStatusCheck(const Status& status) {
...
 Status DBImpl::PreprocessWrite(const WriteOptions& write_options,
                                bool* need_log_sync,
+                               log::Writer** back_writer,
                                WriteContext* write_context) {
-  mutex_.AssertHeld();
   assert(write_context != nullptr && need_log_sync != nullptr);
   Status status;
 
   if (error_handler_.IsDBStopped()) {
+    InstrumentedMutexLock l(&mutex_);
```
yiwu-arbug, line 900: All the checks in `PreprocessWrite` (e.g `error_handler_.IsDBStopped()` here) are supposed to guarded by `mutex_` previously. You need to make each of them lock-free.

yiwu-arbug: Sorry, I misread the change. `IsDBStopped` and some of the other checks are changed to be lock-free. I'll double check.

db/db_impl/db_impl.cc
Little-Wallace: Ok. But this lock must not exist longer than `mutex_`.

Little-Wallace:
It seems that CI run timeout because of spending too much time on building. Could we just adjust the run time limit?

Little-Wallace:
@yiwu-arbug  PTAL again

Little-Wallace:
I have fixed failed CI caused by data race and deadlock .

riversand963:
Thanks for the PR. I plan to take a look this week.

Little-Wallace:
@riversand963  thanks for your review. I am trying to solve the failed test `BackupEngineTestWithParam.OfflineIntegrationTest`.

db/db_impl/db_impl.cc
```diff
@@ -660,6 +662,7 @@ Status DBImpl::CloseHelper() {
     }
   }
   logs_.clear();
+  log_write_mutex_.Unlock();
```
ronag, line 665: Better to use a scope or lock guard to ensure the mutex is released?

Little-Wallace:
I can not understand why merge https://github.com/facebook/rocksdb/pull/7516/commits/d26cb103aca4a3475761bcb7db2596df8a61d1a8 to this PR and it will cause `WalSet/SmallerSyncedSize` failed.

Little-Wallace:
I tried to reproduce the failed as the following step:

```
mkdir build
cd build && cmake .. && make -j 8
./backup_engine_test --gtest_filter="BackupEngineTestWithParam/BackupEngineTestWithParam.*"
```

And I will always get the pass result:
```
[ RUN      ] BackupEngineTestWithParam/BackupEngineTestWithParam.OfflineIntegrationTest/0
[       OK ] BackupEngineTestWithParam/BackupEngineTestWithParam.OfflineIntegrationTest/0 (8539 ms)
[ RUN      ] BackupEngineTestWithParam/BackupEngineTestWithParam.OfflineIntegrationTest/1
[       OK ] BackupEngineTestWithParam/BackupEngineTestWithParam.OfflineIntegrationTest/1 (9121 ms)

```

Do you have any suggestion?

riversand963:
> @riversand963 I can not understand why merge [d26cb10](https://github.com/facebook/rocksdb/commit/d26cb103aca4a3475761bcb7db2596df8a61d1a8) to this PR and it will cause `WalSet/SmallerSyncedSize` failed.

If I hadn't, DBWriteTest.ConcurrentlyDisabledWAL would have become flaky due to changes made in this PR. See the commit message of the said commit.

db/wal_edit_test.cc
```diff
@@ -49,19 +49,6 @@ TEST(WalSet, Overwrite) {
   ASSERT_EQ(wals.GetWals().at(kNumber).GetSyncedSizeInBytes(), kBytes);
 }
 
-TEST(WalSet, SmallerSyncedSize) {
```
riversand963, line 52: I think we should keep this test, and assert that the synced WAL size is unchanged.

riversand963:
> I tried to reproduce the failed as the following step:
> 
> ```
> mkdir build
> cd build && cmake .. && make -j 8
> ./backup_engine_test --gtest_filter="BackupEngineTestWithParam/BackupEngineTestWithParam.*"
> ```
> 
> And I will always get the pass result:
> 
> ```
> [ RUN      ] BackupEngineTestWithParam/BackupEngineTestWithParam.OfflineIntegrationTest/0
> [       OK ] BackupEngineTestWithParam/BackupEngineTestWithParam.OfflineIntegrationTest/0 (8539 ms)
> [ RUN      ] BackupEngineTestWithParam/BackupEngineTestWithParam.OfflineIntegrationTest/1
> [       OK ] BackupEngineTestWithParam/BackupEngineTestWithParam.OfflineIntegrationTest/1 (9121 ms)
> ```
> 
> Do you have any suggestion?

Do you have access to the CircleCI instance? I plan to ssh to the instances and take a look.

db/wal_edit_test.cc
Little-Wallace: I keep it after merging main-branch to current PR

Little-Wallace:
@riversand963  I did not login to  the CircleCI instance. Is there some manual to introduce me how to debug on the CircleCI instance.

Little-Wallace:
I guess that I need to run backup_engine_test in a parallel mode to reproduce this bug?

Little-Wallace:
I found that the operation on `disable_delete_obsolete_files_` will cause a bug on both main branch and this PR, although the probability on main branch will be much less.

It is because that when one thread is execute `DBImpl::FindObsoleteFiles`, it will hold `mutex_`. But for main branch, it will release the mutex when meet the first log is pending synced. After this thread release mutex, another thread may acquire the mutex and set `disable_delete_obsolete_files_` to 1. But the previous thread does not check `disable_delete_obsolete_files_` after it hold mutex again.

riversand963:
> I found that the operation on `disable_delete_obsolete_files_` will cause a bug on both main branch and this PR, although the probability on main branch will be much less.
> 
> It is because that when one thread is execute `DBImpl::FindObsoleteFiles`, it will hold `mutex_`. But for main branch, it will release the mutex when meet the first log is pending synced. After this thread release mutex, another thread may acquire the mutex and set `disable_delete_obsolete_files_` to 1. But the previous thread does not check `disable_delete_obsolete_files_` after it hold mutex again.

Makes sense. Early lock release introduced by this PR will increase the likelihood of this, but I think we should fix it on main branch since it's not really related to this PR. Let me know if you want to contribute a fix. Thanks

Little-Wallace:
@riversand963 I'm glad to contribute a fix to main branch.  I will propose another PR in two days.

Little-Wallace:
@riversand963 Is there any other suggestion? Can I merge this PR?

riversand963:
> @riversand963 Is there any other suggestion? Can I merge this PR?

I believe this PR is promising, but I think this PR needs more testing because it's not a trivial change. Need to think more about this.
In the meantime, we can regularly merge facebook:main to keep it up-to-date.

Little-Wallace:
It's difficult to test a mutex......

tabokie:
@riversand963 We'd like to help with testing if it helps progressing this PR. Also @Connor1996 and I will go over the changes again since it has diverged a bit from the local PR (https://github.com/tikv/rocksdb/pull/197) we've reviewed before.

riversand963:
> @riversand963 Is there any other suggestion? Can I merge this PR?

I am assuming that you are still going to contribute a separate PR to fix the bug in `FindObsoleteFiles()`. Is this still the case?

riversand963:
@tabokie thanks and go ahead.
I myself will also go over the changes again. I'd also like to get some performance improvement numbers, but I am not sure if I have bandwidth for that, especially for some production workloads.

Little-Wallace:
I have fixed `FindObsoleteFiles()` in this PR, which just changes the position of operation on `pending_purge_obsolete_files_`.

tabokie (commented):
I see the comments in db_impl.h are not updated to reflect the new locking logic. HISTORY.md should be updated as well.

db/wal_edit.cc
```diff
@@ -112,26 +112,33 @@ Status WalSet::AddWal(const WalAddition& wal) {
 
   auto it = wals_.lower_bound(wal.GetLogNumber());
   bool existing = it != wals_.end() && it->first == wal.GetLogNumber();
-  if (existing && !wal.GetMetadata().HasSyncedSize()) {
-    std::stringstream ss;
-    ss << "WAL " << wal.GetLogNumber() << " is created more than once";
-    return Status::Corruption("WalSet::AddWal", ss.str());
+
+  if (!existing) {
```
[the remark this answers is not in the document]
Little-Wallace: This change is applied by @riversand963 . I do not understand whether we can insert a record which is `!wal.GetMetadata().HasSyncedSize()`.

db/db_impl/db_impl.h
```diff
@@ -1271,7 +1271,10 @@ class DBImpl : public DB {
 
   // only used for dynamically adjusting max_total_wal_size. it is a sum of
   // [write_buffer_size * max_write_buffer_number] over all column families
-  uint64_t max_total_in_memory_state_;
+  std::atomic<uint64_t> max_total_in_memory_state_;
+  // If true, we have only one (default) column family. We use this to optimize
+  // some code-paths
+  std::atomic<bool> single_column_family_mode_;
```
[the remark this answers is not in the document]
Little-Wallace: done

db/db_impl/db_impl_write.cc
```diff
@@ -1627,10 +1635,11 @@ Status DBImpl::HandleWriteBufferManagerFlush(WriteContext* write_context) {
...
 uint64_t DBImpl::GetMaxTotalWalSize() const {
-  mutex_.AssertHeld();
-  return mutable_db_options_.max_total_wal_size == 0
-             ? 4 * max_total_in_memory_state_
-             : mutable_db_options_.max_total_wal_size;
+  auto max_total_wal_size = max_total_wal_size_.load(std::memory_order_acquire);
+  if (max_total_wal_size > 0) {
+    return max_total_wal_size;
+  }
+  return 4 * max_total_in_memory_state_;
```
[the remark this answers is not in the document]
Little-Wallace: done

db/wal_edit.cc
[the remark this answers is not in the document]
riversand963: Yes, with the proposed change of not holding mutex when necessary, it's possible that two threads calling `SyncWAL()` can be re-ordered when they join the queue of MANIFEST writing.

riversand963:
Have had one successful run of the following
```
COMPILE_WITH_TSAN=1 make db_stress
make crash_test
```

riversand963:
I think it's generally a good idea to avoid contention for mutex. Just curious, have you by any chance been able to run `db_bench` in opt mode and observed performance improvement?

Little-Wallace:
> I think it's generally a good idea to avoid contention for mutex. Just curious, have you by any chance been able to run `db_bench` in opt mode and observed performance improvement?

I run it and observe the mutex latency decrease by `PerfContext`, but I can not give any picture. Maybe @tabokie  can give some information from document in PingCAP, which I do not have permission.

Little-Wallace:
We used to find out when the number of files in the DB instance increase, the mutex latency will also increase. When the number reaches to more than 100K, the latency of 5% of the write requests will reach more than 10ms, and the latency of 1% will reach more than 50ms.

tabokie:
I haven't got time to run tests yet. @Little-Wallace ran a sysbench load with TiKV before (it's based on tikv's rocksdb.6.4 fork).

Before (time=22h, end data size=900GB):
qps
perf_context

After (time=16h, end data size=800GB):
qps-2
perf_context-2

riversand963 (approved):
Thanks @Little-Wallace for the great work! Overall LGTM.
I have extracted out the fix in https://github.com/facebook/rocksdb/pull/7516/commits/d9e17239fdc5ce0eac219b1c4b11a3f6b5e71c6d since it's also likely present in main, and I'd like to separate fix for existing bug apart from this main PR.

Little-Wallace:
@riversand963  It seems that ci-traivs is not stable. Would you mind re-run it and merge this PR?

riversand963:
@Little-Wallace Travis-CI job restarted, it was an infra issue.
I have accepted this PR, and it will go through an internal review before merging. Thanks again

riversand963:
The comments about synchronization on `logs_`, `alive_logs_files_`, etc. need to be updated.

Little-Wallace:
> I tried `db_bench` with simple benchmarks, e.g. fillrandom, overwrite, fillseq, and have not been able to see perf improvement in terms of ops/sec (kind of expected). I will try more tweaking later.

based on my experience, you need a large write throughput (with a large writebatch) for a long time and set target_file_size_base to a small value (such as 8MB). And when the number of sst increase to one hundred thousand, you can observe speed reduction.

riversand963:
> > I tried `db_bench` with simple benchmarks, e.g. fillrandom, overwrite, fillseq, and have not been able to see perf improvement in terms of ops/sec (kind of expected). I will try more tweaking later.
> 
> based on my experience, you need a large write throughput (with a large writebatch) for a long time and set target_file_size_base to a small value (such as 8MB). And when the number of sst increase to one hundred thousand, you can observe speed reduction.

Thanks for the suggestion. Will try later.

riversand963:
Good thing is that I haven't been able to observe a perf regression so far.

tabokie:
@riversand963 One suggestion for the benchmark is to use a smaller `target_file_size_base`: in TiKV we use compaction guard which typically cuts SST files into 10MB chunks, that is much smaller than RocksDB's 64MB default. The higher file count is why we hit the mutex hard.

riversand963:
While showing end-to-end performance gain requires more efforts, it's easy to show that time spent holding the db mutex has drastically decreased. One of the updated unit tests in this PR, i.e. `perf_context_test` has an assertion.
Before this PR,
```
ASSERT_GT(total_db_mutex_nanos, 2000U);
```
After this PR,
```
ASSERT_LT(total_db_mutex_nanos, 100U);
```

I did another simple benchmarking on a non-vm host.
```bash
TEST_TMPDIR=/dev/shm/rocksdb ./db_bench -benchmarks=fillseq,overwrite -duration=60 -batch_size=100 -perf_level=5
```
Results show
```
db_mutex_lock_nanos = 564408021 (before)
db_mutex_lock_nanos = 11142 (after)
```

mdcallag:
I see large improvements with fillseq (leveled, universal) and overwrite (universal) in IO-bound workloads -- up to 1.5X more throughput. Thank you for making RocksDB better.

https://twitter.com/MarkCallaghanDB/status/1574425353564475394

PR #7519, opened 2020-10-08
Do not rely on the two-argument std::pair constructor being constexpr

ltamasi:
Summary:
The `std::pair(const T1& x, const T2& y);` constructor is `constexpr`
only starting from C++14; relying on this breaks compilation on certain
compilers/platforms we need to support.

Test Plan:
`make check`


ajkr (approved):
LGTM!

mrambacher (commented):
LGTM. Fixes the compilation issues under MacOS.

PR #7520, opened 2020-10-08
Redesign block cache pinning API

ajkr:
The old flag-based APIs (`BlockBasedTableOptions::pin_l0_filter_and_index_blocks_in_cache` and `BlockBasedTableOptions::pin_top_level_index_and_filter`) were insufficient for our needs. For example, it was impossible to pin only unpartitioned meta-blocks, which could prevent block cache contention when turning on dictionary compression or during a migration to partitioned indexes/filters. It was also impossible to pin all meta-blocks in memory while having predictable memory usage via block cache. If we had continued adding flags to address these scenarios, they would have had significant overlap causing confusion. Instead, this PR deprecates the flags and starts a new API with non-overlapping options.

Test Plan:

- new unit test
- added new options to stress/crash test and ran for a while: `$ python tools/db_crashtest.py blackbox --simple --max_key=1000000 -write_buffer_size=1048576 -target_file_size_base=1048576 -max_bytes_for_level_base=4194304 --interval=10 -value_size_mult=33 -column_families=1 -reopen=0`


include/rocksdb/table.h
```diff
@@ -51,6 +51,54 @@ enum ChecksumType : char {
...
+// `CachePinningOptions` contains members indicating the desired block cache
+// pinning behavior for the different categories of blocks.
+struct CachePinningOptions {
+  // The tier of block-based tables whose top-level index of partitioned
+  // meta-blocks will be pinned.
+  //
+  // Note `cache_index_and_filter_blocks` must be true for this option to have
+  // any effect. Otherwise the top-level index of partitioned meta-blocks would
+  // be held in table reader memory, outside the block cache.
+  PinningTier metablock_top_level_index_pinning = PinningTier::kFallback;
```
ajkr, line 88: Carried over unaddressed comment from https://github.com/ajkr/rocksdb/commit/fe6ed27b4281e7275c31b6ca89f417387612c684:

> Please clarify what you mean by "meta-block" as this is not a standard term in public API. And is "metablock" redundant in these fields? Would we consider pinning something that's not a metadata block?

include/rocksdb/table.h
```diff
@@ -51,6 +51,54 @@ enum ChecksumType : char {
   kxxHash64 = 0x3,
 };
 
+// `PinningTier` is used to specify which tier of block-based tables should
+// be affected by a block cache pinning setting (see `CachePinningOptinos`
```
mrambacher, line 55: "CachePinningOptions"

table/block_based/block_based_table_reader.cc
```diff
@@ -987,20 +987,56 @@ Status BlockBasedTable::PrefetchIndexAndFilterBlocks(
...
+  const bool pin_metablock_partition =
+      is_pinned(table_options.cache_pinning_options.metablock_partition_pinning,
+                table_options.pin_l0_filter_and_index_blocks_in_cache
+                    ? PinningTier::kMaybeFlushed
+                    : PinningTier::kNone);
+  const bool pin_unpartitioned_metablock = is_pinned(
+      table_options.cache_pinning_options.unpartitioned_metablock_pinning,
+      table_options.pin_l0_filter_and_index_blocks_in_cache
+          ? PinningTier::kMaybeFlushed
+          : PinningTier::kNone);
```
mrambacher, line 1028: I think you can do this simpler by using the InitializeOptions method of BlockBasedTableFactory.  You should be able to update the cache_pinning_options in that method based on either of the settings.

mrambacher: Additionally, if you update the pinning options in the InitializeOptions, you can mark the old/deprecated options as "kDontSerialize" and they will no longer be written to the output.  This would help make them more "obsolete".

include/rocksdb/table.h
```diff
@@ -79,19 +127,45 @@ struct BlockBasedTableOptions {
...
+  // option still takes effect on relevant `CachePinningOptions` members that
+  // are set to `PinningTier::kFallback`. In particular, the overrides from
+  // setting this option (and the changes users should make to migrate away from
+  // setting this option) are:
+  //
+  // - `metablock_partition_pinning == PinningTier::kFallback` ->
+  //   `metablock_partition_pinning == PinningTier::kMaybeFlushed`
+  // - `unpartitioned_metablock_pinning == PinningTier::kFallback` ->
+  //   `unpartitioned_metablock_pinning == PinningTier::kMaybeFlushed`
+  //
```
mrambacher, line 140: I am not sure I understand the logic here for how the options are used.  It seems like if the old option was false (the default) these values should be whatever their default is (which I assume is "off"?).  If the old option is true and the new options are their default, then set them to the expected "default" values.  Does that make it cleaner?

include/rocksdb/table.h
```diff
@@ -79,19 +127,45 @@ struct BlockBasedTableOptions {
...
 
+  // DEPRECATED: This option will be removed in a future version. For now, this
+  // option still takes effect on relevant `CachePinningOptions` members that
+  // are set to `PinningTier::kFallback`. In particular, the override from
+  // setting this option (and the change users should make to migrate away from
+  // setting this option) is:
+  //
+  // - `metablock_top_level_index_pinning == PinningTier::kFallback` ->
+  //   `metablock_top_level_index_pinning == PinningTier::kAll`
+  //
```
mrambacher, line 155: The same comment applies as above.  The default for the PinningOption should be whatever the behavior for "true" means.  If this value is "false" and the PinningOption is the default, turn off the PinningOption.

include/rocksdb/table.h
ajkr: I plan to rename the members `partition_pinning`, `top_level_index_pinning`, and `unpartitioned_pinning`? Since metadata is redundant, we can rename the structure `MetadataCachePinningOptions`.

include/rocksdb/table.h
ajkr: Yes, I tried rewriting it.

include/rocksdb/table.h
```diff
@@ -51,6 +51,55 @@ enum ChecksumType : char {
...
+  // The tier of block-based tables whose top-level index into metadata
+  // partitions will be pinned. Currently indexes and filters may be
+  // partitioned.
+  //
+  // Note `cache_index_and_filter_blocks` must be true for this option to have
+  // any effect. Otherwise any top-level index into metadata partitions would be
+  // held in table reader memory, outside the block cache.
+  PinningTier top_level_index_pinning = PinningTier::kFallback;
+
+  // The tier of block-based tables whose metadata partitions will be pinned.
```
pdillinger, line 90: I think a potential fallacy here is "why partition metadata if it's going to be pinned?"  Especially for filters, the CPU hit from key-order partitioning is paid even if the partitions are pinned. I don't think there's much of a case for "what about trivial moves? what about re-opening with different options?" etc. Why not instead skip partitioning in the cases where pinning would be used?

ajkr: Good point. For trivial move I was surprised that it might not even reopen the table reader, so the different settings of the lower level might not even be applied until the DB reopens. So I agree that is not a strong case. The configuration you mentioned could get even weirder by setting `top_level_index_pinning == PinningTier::kNone && partition_pinning >= PinningTier::kMaybeFlushed` (the testing shows such a config does work though). My only guess at a valid use case of `partition_pinning >= PinningTier::kMaybeFlushed` would be for people who really don't like large allocations, maybe due to fragmentation.

Do you think this is needed to proceed with this PR? Pinning partitions was already partially supported by the existing option, `pin_l0_filter_and_index_blocks_in_cache`.

pdillinger: I guess we can punt on that improvement.

include/rocksdb/table.h
```diff
@@ -51,6 +51,55 @@ enum ChecksumType : char {
...
+
+  // This tier contains no block-based tables.
+  kNone,
+
+  // This tier contains block-based tables that may have originated from a
+  // memtable flush. In particular, it includes tables from L0 that are smaller
+  // than 1.5 times the current `write_buffer_size`. Note these criteria imply
+  // it can include intra-L0 compaction outputs and ingested files, as long as
+  // they are not abnormally large for L0.
+  kMaybeFlushed,
```
pdillinger, line 72: I don't like this name because it's not clear which way we are approximating. Would something like `kFlushedAndSimilar` be better?

ajkr: That is a much better name, thanks. Updated the PR.

include/rocksdb/table.h
```diff
@@ -51,6 +51,55 @@ enum ChecksumType : char {
...
+};
+
+// `MetadataCachePinningOptions` contains members indicating the desired block
+// cache pinning behavior for the different categories of metadata blocks.
+struct MetadataCachePinningOptions {
+  // The tier of block-based tables whose top-level index into metadata
+  // partitions will be pinned. Currently indexes and filters may be
+  // partitioned.
+  //
+  // Note `cache_index_and_filter_blocks` must be true for this option to have
```
jay-zhuang, line 85: Just a question, can `cache_index_and_filter_blocks` also be deprecated or included in this option? Then the option would be like `MetadataCacheOptions`, which has options to enable/disable, pin/unpin the cache.

pdillinger: I think that's a great point that could influence how we name these options, even if we don't include it right now.

pdillinger (commented):
Probably best to add to db_stress in this PR

include/rocksdb/table.h
ajkr: That's a good question. Indeed there's a "TODO(ajkr)" in `cache_index_and_filter_blocks`'s API comment saying we should change its name. `cache_index_and_filter_blocks_with_high_priority` has the same naming problem, i.e., it says it applies to index and filter blocks, while actually it applies to all unpartitioned metadata and top-level indexes into metadata partitions. I am not that sure right now whether we'll want to simply add a new flag that does the same thing under a different name, or do something else, or do nothing. But it might not hurt to change the name to `MetadataCacheOptions` so we're not restricted later... I will make that change for now unless anyone has other suggestions.

ajkr: Done.

ajkr (commented):
> Probably best to add to db_stress in this PR

Done. Fixed a bug that showed up: prefetching needs to be enabled for unpartitioned metadata blocks that are pinned while `prefetch_all == false`.

include/rocksdb/table.h
```diff
@@ -51,6 +51,55 @@ enum ChecksumType : char {
...
+  // they are not abnormally large compared to flushed files in L0.
+  kFlushedAndSimilar,
+
+  // This tier contains all block-based tables.
+  kAll,
+};
+
+// `MetadataCacheOptions` contains members indicating the desired caching
+// behavior for the different categories of metadata blocks.
+struct MetadataCacheOptions {
```
riversand963, line 80: Maybe it's just me, but I kind of find the new `MetadataCacheOptions` not as intuitive.
When configuring RocksDB's block cache pinning behaviors, I tend to think in the following pattern:
Given a type or types of meta blocks, what pinning behavior should it have?
The current config options seems to make me think in the other way:
For a specific behavior, e.g. 'pin_top_level_index', 'pin_partition', etc, what kind of SSTs should have this behavior?
It's very likely that I have missed some important piece, and let me know if I am wrong or misunderstood terribly.

ajkr: The steps I'd follow when configuring this option are:

1. Find the variable corresponding to the category of block you care about. For example, if you care about top level indexes into metadata partitions, you'd find the variable `top_level_index_pinning`.
2. Find the value of `PinningTier` according to which SSTs you want that category of blocks to be pinned. For example, if you want it to apply for all SSTs, then you'd pick `PinningTier::kAll`.

To me, that process feels aligned with the first pattern you described, but please let me know if I missed something.

riversand963: Thanks for the clarification. That makes sense.

pdillinger: The old way only makes sense if you want some specific use case that matches a boolean option and have hope that the other boolean options don't interfere with yours in a weird way.

Assuming cache_index_and_filter_blocks, the behavior is boolean: pin or don't pin. To cover the other two dimensions (what kind of metadata blocks? what kind of SST files?) without excessive complexity, we make that boolean implicit.

table/block_based/block_based_table_reader.cc
ajkr: I will try this out but need to defer it until another PR since today's the last day for features in the next release, and there are still bugs to fix.

ajkr (commented):
the crash test shows failures. I wonder if they're new or existing as `pin_l0_filter_and_index_blocks_in_cache` isn't ever tested. Let me check.

db/db_block_cache_test.cc
```diff
@@ -888,6 +888,143 @@ TEST_F(DBBlockCacheTest, CacheCompressionDict) {
...
+  table_options.cache_index_and_filter_blocks = true;
+  table_options.metadata_cache_options.top_level_index_pinning =
+      top_level_index_pinning_;
+  table_options.metadata_cache_options.partition_pinning = partition_pinning_;
+  table_options.metadata_cache_options.unpartitioned_pinning =
+      unpartitioned_pinning_;
+  table_options.filter_policy.reset(
+      NewBloomFilterPolicy(10 /* bits_per_key */));
+  table_options.index_type =
+      BlockBasedTableOptions::IndexType::kTwoLevelIndexSearch;
```
pdillinger, line 940: Would be nice to also test unpartitioned filter / index. Perhaps you can re-open and make one more L0 file with unpartitioned filter / index? (I don't need to see unpartitioned filter / index exercised on all pinning tiers.)

include/rocksdb/table.h
pdillinger: Having thought a bit more, I think cached vs. not should generally be the same for all metadata blocks. Unbounded is unbounded, regardless of which subset of metadata you for some reason keep bounded despite other metadata left unbounded.

I'm OK with the naming updates, though.

db/db_block_cache_test.cc
ajkr: Done, although I did end up doing it on all pinning tiers.

pdillinger (approved):
LGTM

table/two_level_iterator.cc
```diff
@@ -43,6 +43,10 @@ class TwoLevelIndexIterator : public InternalIteratorBase<IndexValue> {
     assert(Valid());
     return second_level_iter_.key();
   }
+  Slice user_key() const override {
```
ajkr, line 46: This fixes a corruption bug. If we inherit the base class's `user_key()` implementation, it will `ExtractUserKey()` on `key()`. `key()` simply returns whatever is the current key in the index block, which may be an internal key or a user key.  In case it's a user key, calling `ExtractUserKey()` on it causes an incorrect truncation.

table/block_based/partitioned_index_reader.cc
```diff
@@ -173,7 +173,7 @@ Status PartitionIndexReader::CacheDependencies(const ReadOptions& ro,
       return s;
     }
     if (block.GetValue() != nullptr) {
-      if (block.IsCached()) {
+      if (block.IsCached() || block.GetOwnValue()) {
```
ajkr, line 176: This fixes another edge case leading to corruption. Due to how `partition_map_` is used, it must contain either all partitions or no partitions. When it contains only some partitions, the partitions not in the map are assumed to be empty (https://github.com/facebook/rocksdb/blob/12b78e40bdba980f3692fd2018cc905eef2127f8/table/block_based/block_based_table_reader.cc#L1977-L1993). Such a wrong assumption causes corrupt reads.

One case in which only some partitions would be added to the map is when `enable_index_compression == 1`. It can be the case that some of the partitions are compressed while others are not (e.g., due to ineffective compression ratio). The compressed partitions would be decompressed, loaded to block cache, and then inserted into `partition_map_`. Meanwhile, the uncompressed partitions could be read directly from the file, e.g., when `mmap_read == true`, and such blocks (until this PR) did not get inserted into `partition_map_`.

PR #7521, opened 2020-10-08
Add tests for paranoid checks with range deletion

mrambacher:
Added unit tests that have paranoid_check = true that perform range deletions.  At the moment, the deleted ranges do not appear to be checked as part of the paranoid checks.


ajkr (changes requested):
minor: it looks like a status check enforcement is failing in `GetSnapshotImpl()`. I'd guess it's just due to this line: https://github.com/facebook/rocksdb/blob/9e03e4dd52ead7875b407d72106c5d23594065b4/db/db_impl/db_impl.cc#L2933

PR #7523, opened 2020-10-09
Handoff checksum Implementation

zhichao-cao:
in PR #7419 , we introduce the new Append and PositionedAppend APIs to WritableFile at File System, which enable RocksDB to pass the data verification information (e.g., checksum of the data) to the lower layer. In this PR, we use the new API in WritableFileWriter, such that the file created via WritableFileWrite can pass the checksum to the storage layer. To control which types file should apply the checksum handoff, we add checksum_handoff_file_types to DBOptions. User can use this option to control which file types (Currently supported file tyes: kLogFile, kTableFile, kDescriptorFile.) should use the new Append and PositionedAppend APIs to handoff the verification information.

Test plan: add new unit test, pass make check/ make asan_check


file/writable_file_writer.cc
```diff
@@ -487,8 +504,17 @@ IOStatus WritableFileWriter::WriteDirect() {
         start_ts = FileOperationInfo::StartNow();
       }
       // direct writes must be positional
-      s = writable_file_->PositionedAppend(Slice(src, size), write_offset,
-                                           IOOptions(), nullptr);
+      if (perform_data_verification_) {
+        std::string checksum;
```
pdillinger, line 508: Building on Siying's previous comments about efficiency: I don't think we should do a heap allocation (std::string append) on every file write with handoff checksum.

include/rocksdb/file_system.h
```diff
@@ -47,6 +47,21 @@ class RateLimiter;
 using AccessPattern = RandomAccessFile::AccessPattern;
 using FileAttributes = Env::FileAttributes;
 
+// File types in current RocksDB
+enum FileType : unsigned char {
```
pdillinger, line 51: Why put this in file_system.h if it's used on options.h?

pdillinger (commented):
We seem to have disparate configurations that need to work together to make this work. If I set checksum_handoff_file_types, that does nothing on its own. If I support the DVI versions of FSWritableFile functions, that does nothing on its own. Why should I have to add something on both ends?

Echoing some comments on the design document, I think the DB and/or column families should only have a notion of which files are considered "critical" and let the storage layer handle that appropriately, when there is a distinction to be made in handling them. It then makes sense for there to be a reasonable set of defaults considered "critical."

(Perhaps revisiting old debates.) If handoff checksums are computed in the storage layer anyway, why can't the Env / FileSystem provide a function to generate the DataVerificationInfo it expects from data. And if we do that, why not let DataVerificationInfo live entirely under Env / FileSystem (in appropriate implementations)?

include/rocksdb/file_system.h
zhichao-cao: Will move it to options.h.

include/rocksdb/file_system.h
```diff
@@ -47,6 +47,21 @@ class RateLimiter;
...
+  kLogFile,
```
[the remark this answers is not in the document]
zhichao-cao: Sure. I agree on it.

zhichao-cao:
Thanks for the review. !
> We seem to have disparate configurations that need to work together to make this work. If I set checksum_handoff_file_types, that does nothing on its own. If I support the DVI versions of FSWritableFile functions, that does nothing on its own. Why should I have to add something on both ends?
> 

My point is, this is a three-ways contracts. 1) user want to do handoff on which type of files,  2) RocksDB provides the path, 3) the customized File System support it. From RocksDB's view, it does not know File System support it or not. From user's view, it will set the option only when it knows the customized File System will do the job.

> Echoing some comments on the design document, I think the DB and/or column families should only have a notion of which files are considered "critical" and let the storage layer handle that appropriately, when there is a distinction to be made in handling them. It then makes sense for there to be a reasonable set of defaults considered "critical."
> 

In our case, we hope to calculate the checksum as early as possible, then the customized File System can use/verfiy/pass it. So, it is better to calculate the checksum in WritableFileWrite with the control option.

> (Perhaps revisiting old debates.) If handoff checksums are computed in the storage layer anyway, why can't the Env / FileSystem provide a function to generate the DataVerificationInfo it expects from data. And if we do that, why not let DataVerificationInfo live entirely under Env / FileSystem (in appropriate implementations)?

The motivation of this PR is to move the checksum calculation from Env/File system up, and do it as early as possible. Even in WritableFileWriter, we do memory copy. There will be a follow up PR to bridge the gap, which covers the checksum verification from source buffer, to the place where the handoff checksum is calculated (due to the buffer write and write rate limiter, we are not able to calculate the checksum for the data being append that early)

file/writable_file_writer.cc
zhichao-cao: I replaced it with char checksum_buf[sizeof(uint32_t)], might be a better way to avoid string append.

pdillinger:
> There will be a follow up PR to bridge the gap, which covers the checksum verification from source buffer, to the place where the handoff checksum is calculated (due to the buffer write and write rate limiter, we are not able to calculate the checksum for the data being append that early)

I'm finding it difficult to approve this complication of the Env interface without actually seeing a data integrity improvement that is not possible with the existing interface. Computing checksum on the buffer passed to Append in caller rather than callee to me is not a meaningful data integrity improvement.

zhichao-cao:
> > There will be a follow up PR to bridge the gap, which covers the checksum verification from source buffer, to the place where the handoff checksum is calculated (due to the buffer write and write rate limiter, we are not able to calculate the checksum for the data being append that early)
> 
> I'm finding it difficult to approve this complication of the Env interface without actually seeing a data integrity improvement that is not possible with the existing interface. Computing checksum on the buffer passed to Append in caller rather than callee to me is not a meaningful data integrity improvement.

You mean there is nearly no different between Append itself do the checksum calculation for the buffer and the caller of Append to do it?

pdillinger:
> You mean there is nearly no different between Append itself do the checksum calculation for the buffer and the caller of Append to do it?

If it's the same buffer, I don't see a difference between computing checksum right before call vs. right at beginning of call.

zhichao-cao:
> > You mean there is nearly no different between Append itself do the checksum calculation for the buffer and the caller of Append to do it?
> 
> If it's the same buffer, I don't see a difference between computing checksum right before call vs. right at beginning of call.

Due to the rate limiter, yes, the buffer/size is the same as what we pass down. But based on the WS and Zippy requirement, if RocksDB can provide the calculation in higher layer instead of FileSystem, they can avoid re-calculate it again. In current implementation+the follow up one, we can achieve from WritableFileWriter::append(data)->buffer for WritableFile::Append->WS client validation. Without this interface, even we verify the buffer in WritableFileWriter, FileSystem still need to calculate again, so that's one more calculation.

ajkr:
> I'm finding it difficult to approve this complication of the Env interface without actually seeing a data integrity improvement that is not possible with the existing interface. Computing checksum on the buffer passed to Append in caller rather than callee to me is not a meaningful data integrity improvement.

Just to clarify, is the `Env` interface change you're referring to the one that landed in #7419?

> If it's the same buffer, I don't see a difference between computing checksum right before call vs. right at beginning of call.

I tend to think it's ok to trust the memory's reliability features for data at rest, so computing in callee would work for me. However, any gap in the software-level protection feels like somewhere users will place the blame in corruption investigations, so I could go either way.

> Due to the rate limiter, yes, the buffer/size is the same as what we pass down. But based on the WS and Zippy requirement, if RocksDB can provide the calculation in higher layer instead of FileSystem, they can avoid re-calculate it again. In current implementation+the follow up one, we can achieve from WritableFileWriter::append(data)->buffer for WritableFile::Append->WS client validation. Without this interface, even we verify the buffer in WritableFileWriter, FileSystem still need to calculate again, so that's one more calculation.

Do you mean in the future there'll be integrity protection computed in `WritableFileWriter::Append()` regardless of this interface, and its checksum will be usable by the underlying `WritableFile::Append()`? I am not quite able to imagine how it'll work. Would we need to compute a rolling checksum over all the chunks passed to `WritableFile::Append()` and verify against the rolling checksum of all data received in `WritableFileWriter::Append()`?

zhichao-cao:
> Do you mean in the future there'll be integrity protection computed in `WritableFileWriter::Append()` regardless of this interface, and its checksum will be usable by the underlying `WritableFile::Append()`? I am not quite able to imagine how it'll work. Would we need to compute a rolling checksum over all the chunks passed to `WritableFile::Append()` and verify against the rolling checksum of all data received in `WritableFileWriter::Append()`?

Yes. The motivation of this project, is to handle the potential corruption of WAL and Manifest, with the discussion with Zippy, they will use the new FileSystem Append with verification to handle the checksum handoff for their log. So they also want the hand off can be used for other files like SST files. The protection will be in this way: by using the rolling checksum of the data in the buffer of `WritableFileWriter::Append()`, we can have a protection until  we calculate the checksum of the piece of data being passed to `WritableFile::Append()`. In WS use case, WS will avoid the current checksum calculation of the data in `WritableFile::Append()`, and directly use the checksum for further verification (e.g., on WS client side or even server side). There are some challenges to close the integrity gap in `WritableFileWriter::Append()` especially caused by buffered write and rate-limiter. Therefore, in this PR, the first step is to move the buffer checksum calculation up from `WritableFile::Append()` customized implementation to the logic inside `WritableFileWriter::Append()`.

ajkr:
> > Do you mean in the future there'll be integrity protection computed in `WritableFileWriter::Append()` regardless of this interface, and its checksum will be usable by the underlying `WritableFile::Append()`? I am not quite able to imagine how it'll work. Would we need to compute a rolling checksum over all the chunks passed to `WritableFile::Append()` and verify against the rolling checksum of all data received in `WritableFileWriter::Append()`?
> 
> Yes. The motivation of this project, is to handle the potential corruption of WAL and Manifest, with the discussion with Zippy, they will use the new FileSystem Append with verification to handle the checksum handoff for their log. So they also want the hand off can be used for other files like SST files. The protection will be in this way: by using the rolling checksum of the data in the buffer of `WritableFileWriter::Append()`, we can have a protection until we calculate the checksum of the piece of data being passed to `WritableFile::Append()`. In WS use case, WS will avoid the current checksum calculation of the data in `WritableFile::Append()`, and directly use the checksum for further verification (e.g., on WS client side or even server side). There are some challenges to close the integrity gap in `WritableFileWriter::Append()` especially caused by buffered write and rate-limiter. Therefore, in this PR, the first step is to move the buffer checksum calculation up from `WritableFile::Append()` customized implementation to the logic inside `WritableFileWriter::Append()`.

Understood, thanks for elaborating. In this PR, it looks like the checksums passed to `WritableFile::Append()` are computed with an empty initial state each time, not rolling checksum. Won't we need the rolling checksum of all the chunks to verify against any checksum at a higher layer than rate limiter? If so, should `WritableFile::Append()`'s API provide rolling checksums? Sorry if I misunderstood something.

zhichao-cao:
> > > Do you mean in the future there'll be integrity protection computed in `WritableFileWriter::Append()` regardless of this interface, and its checksum will be usable by the underlying `WritableFile::Append()`? I am not quite able to imagine how it'll work. Would we need to compute a rolling checksum over all the chunks passed to `WritableFile::Append()` and verify against the rolling checksum of all data received in `WritableFileWriter::Append()`?
> > 
> > 
> > Yes. The motivation of this project, is to handle the potential corruption of WAL and Manifest, with the discussion with Zippy, they will use the new FileSystem Append with verification to handle the checksum handoff for their log. So they also want the hand off can be used for other files like SST files. The protection will be in this way: by using the rolling checksum of the data in the buffer of `WritableFileWriter::Append()`, we can have a protection until we calculate the checksum of the piece of data being passed to `WritableFile::Append()`. In WS use case, WS will avoid the current checksum calculation of the data in `WritableFile::Append()`, and directly use the checksum for further verification (e.g., on WS client side or even server side). There are some challenges to close the integrity gap in `WritableFileWriter::Append()` especially caused by buffered write and rate-limiter. Therefore, in this PR, the first step is to move the buffer checksum calculation up from `WritableFile::Append()` customized implementation to the logic inside `WritableFileWriter::Append()`.
> 
> Understood, thanks for elaborating. In this PR, it looks like the checksums passed to `WritableFile::Append()` are computed with an empty initial state each time, not rolling checksum. Won't we need the rolling checksum of all the chunks to verify against any checksum at a higher layer than rate limiter? If so, should `WritableFile::Append()`'s API provide rolling checksums? Sorry if I misunderstood something.

Based on the current information from WS, they can only use the checksum information aligned with the data being passed in `WritableFile::Append()`. Therefore, it is a one-time checksum being passed. The rolling checksum (also reuse the record checksum in WAL and Manifest if possible) I mentioned before, is to bridge the gap between data being added to the write buffer in `WritableFileWriter::Append()` and the time we calculate the one-time checksum. For example, for the working flow of the small writes. Those data being passed in by `WritableFileWriter::Append(data)` are memcpy to the write buffer. We need to calculate the rolling checksum of the "data". When the buffer is sealed to be write to the lower level via `WritableFile::Append()`, a verification of the write buffer's data is needed. This means, the data in the buffer is correct. That is, the content of the write buffer can be trusted. In the next step, with rate limiter, this buffer might be pass down with multiple `WritableFile::Append()` calls. At that time, we calculate the one-time checksum and pass down with the data pieces. In this PR, it is the one-time checksum which aligns with the data being pass down through `WritableFile::Append()`.

ajkr:
Got it, thanks for the explanation.

> When the buffer is sealed to be write to the lower level via WritableFile::Append(), a verification of the write buffer's data is needed.

My understanding is with a `WritableFile::Append()` API taking rolling checksum for the file would be beneficial in eliminating the checksum calculation for the above verification. But it is not feasible due to WS system limitation. I wonder if it is a fundamental limitation that'll be common to other systems, or just something they chose not to implement in H2 (we'll have a lot of difficulty changing the API once the limitation or WS goes away).

zhichao-cao:
> Got it, thanks for the explanation.
> 
> > When the buffer is sealed to be write to the lower level via WritableFile::Append(), a verification of the write buffer's data is needed.
> 
> My understanding is with a `WritableFile::Append()` API taking cumulative file checksum would be beneficial in eliminating the checksum calculation for the above verification. But it is not feasible due to WS system limitation. I wonder if it is a fundamental limitation that'll be common to other systems, or just something they chose not to implement in H2 (we'll have a lot of difficulty changing the API once the limitation goes away).
> 
> edit: s/rolling checksum/cumulative file checksum/g

My personal consideration is, in the API, we pass a data structure: DataVerificationInfo. It can bring any verification information we want. First, it is natural to bring the checksum align with the data in `WritableFile::Append()`. It is what I implemented and what WS support in the current stage. In the future, as mentioned in the discussion with WS and Zippy, we can bring more information down to the FS such as the offset of the data in the file, rolling checksum of the whole file, or others. Therefore, I think we have the flexibility to extend it. User can choose to use the one-time checksum, or other verification information in the `WritableFile::Append()` based on what is their customized FileSystem implementation and what they support.

pdillinger:
>  User can choose to use the one-time checksum, or other verification information in the WritableFile::Append() based on what is their customized FileSystem implementation and what they support.

Except not really, because right now the generation of DataVerificationInfo is fixed in the implementation. We are directly tied up in any updates or evolution.

Wouldn't it be easier for both sides if we just provide a way for WritableFileWriter to do no buffering (IMHO at the choice of Env/FS), and pass original data directly to FSWritableFile for each Append? Then the FS/Env implementation can worry about its own buffering and coordinating that with hand-off checksums. The fact that we provide buffering above the FS layer seems to me a convenience to FS implementors, and should be optional so that our implementation does not become intertwined will all possible concerns of buffering for all possible FS implementations. Re-implementing a buffering layer into an already complex, niche FS implementation should not be complicated, and not nearly complicated enough to justify intertwining our code and releases with this problem.

zhichao-cao:
> Except not really, because right now the generation of DataVerificationInfo is fixed in the implementation. We are directly tied up in any updates or evolution.
> 
> Wouldn't it be easier for both sides if we just provide a way for WritableFileWriter to do no buffering (IMHO at the choice of Env/FS), and pass original data directly to FSWritableFile for each Append? Then the FS/Env implementation can worry about its own buffering and coordinating that with hand-off checksums. The fact that we provide buffering above the FS layer seems to me a convenience to FS implementors, and should be optional so that our implementation does not become intertwined will all possible concerns of buffering for all possible FS implementations. Re-implementing a buffering layer into an already complex, niche FS implementation should not be complicated, and not nearly complicated enough to justify intertwining our code and releases with this problem.

It is a possible direction, but may not be this PR. If we let WritableFileWriter to do no buffering and directly pass what is appended to FS, we will not be able to control write rate limiter, the FS need to deal with the small writes. This can be a future implementation to give an option to control if the user want to let FS to fully control the write process of Append. But here, since Zippy and WS relies on the current flow in WritableFileWriter, it would be better to follow the current buffering and rate limit control in WritableFileWriter.

pdillinger:
> If we let WritableFileWriter to do no buffering and directly pass what is appended to FS, we will not be able to control write rate limiter, the FS need to deal with the small writes.

This is already handled in `WritableFileWriter::WriteBuffered` (which I think is supposed to mean "rely on layer below for any buffering"), so I propose we just need another hook for calling it from Append.

And I propose that the theoretical problem of data being corrupted while waiting for the rate limiter be left for future work. (AFAICT, my proposal allows an FS implementation to cover more of the pipeline with checksums sooner than this PR.)

zhichao-cao:
> > If we let WritableFileWriter to do no buffering and directly pass what is appended to FS, we will not be able to control write rate limiter, the FS need to deal with the small writes.
> 
> This is already handled in `WritableFileWriter::WriteBuffered` (which I think is supposed to mean "rely on layer below for any buffering"), so I propose we just need another hook for calling it from Append.
> 
> And I propose that the theoretical problem of data being corrupted while waiting for the rate limiter be left for future work. (AFAICT, my proposal allows an FS implementation to cover more of the pipeline with checksums sooner than this PR.)

Can you explain more on your proposal? Or a design doc for it. It is hard for me to pick up the overall information from the comments here.

pdillinger:
Re-sync after more internal design discussion and deliberation: In summary, I have argued this system of checksum handoff is of very limited value *unless* our own block-level checksums can be leveraged in generating hand-off checksums, and I have outlined a rough path to achieving that, involving stitching together CRCs as in folly::crc32_combine. (leveraged == when the storage layer verifies a hand-off checksum, that would imply that the block-level checksum is valid.) There at least seems to be agreement that leveraging our own block-level checksums is our best medium-to-long-term prospect for maximizing data integrity with good efficiency. That is the context for review now.

With that in mind, I think we need to decide right now whether we intend to make our embedded block-level checksums extensible or only allow built-in implementations (and therefore adding crc32 support in addition to existing crc32c support, for compatibility with other FB infrastructure). If we only allow built-in implementations, then this PR should move toward using ChecksumType for hand-off instead of named-string checksum types. If we want to make block checksums extensible, I think we need to plan out some details to avoid fragmented design re: custom checksums. (Which names are recognized where, etc.)

zhichao-cao:
> Re-sync after more internal design discussion and deliberation: In summary, I have argued this system of checksum handoff is of very limited value _unless_ our own block-level checksums can be leveraged in generating hand-off checksums, and I have outlined a rough path to achieving that, involving stitching together CRCs as in folly::crc32_combine. (leveraged == when the storage layer verifies a hand-off checksum, that would imply that the block-level checksum is valid.) There at least seems to be agreement that leveraging our own block-level checksums is our best medium-to-long-term prospect for maximizing data integrity with good efficiency. That is the context for review now.
> 
> With that in mind, I think we need to decide right now whether we intend to make our embedded block-level checksums extensible or only allow built-in implementations (and therefore adding crc32 support in addition to existing crc32c support, for compatibility with other FB infrastructure). If we only allow built-in implementations, then this PR should move toward using ChecksumType for hand-off instead of named-string checksum types. If we want to make block checksums extensible, I think we need to plan out some details to avoid fragmented design re: custom checksums. (Which names are recognized where, etc.)

Thanks for the detailed explanation. I have one question here about the block-level checksums. If we allow the custom checksums extensible, we may need to change the block format (block trailer for checksum is 4 bytes). For example, if user want to use SHA1, SHA256, the checksum size will be different. I'm not sure it is a changes RocksDB actively needed. In most cases, block checksum is used internally for data protection.

I may prefer to handle it in this way. In the short to mid-term, we only allow build-in implementations to support limited checksum types for blocks without changes for block format (say crc32, crc32c, and potentially others). In the long term, it depends on we want to change the block format or not. If not, then, user defined checksum types can be used for checksum handoff, however, user needs to tolerate the overhead-- internal checksum/verification before the handoff and calculate the customized checksum for Data in append sent to the FS layer.

pdillinger:
Yes, I forgot to mention checksum size. I see three options:

1) Built-in block checksums only (we likely add plain crc32)
2) Extensible block checksums, must fit in 4 bytes
3) Extensible block checksums of configurable size

I was intending we choose between option 1 and option 2. To me option 3 is "too big" of a change to SST and WAL schemas with no discernible value. 32 bits should be quite enough for data integrity without a need for data identity. And with such a big schema change, old code on new data would likely report it as unpredictable corruption rather than "not supported".

IMHO, option 2 is not a drastic schema change compared to option 1. Even with option 1, we need to make WAL checksums configurable, which I believe we can do with a new record type.

db/blob/blob_file_builder.cc
```diff
@@ -178,11 +179,13 @@ Status BlobFileBuilder::OpenBlobFileIfNeeded() {
   file->SetWriteLifeTimeHint(write_hint_);
 
   Statistics* const statistics = immutable_cf_options_->statistics;
-
+  bool should_checksum_handoff = ShouldChecksumHandoff(
```
pdillinger, line 182: It seems like many close copies of this code are being added in many places.
* Is there a reason this one does not set handoff_checksum_type in FileOptions?
* Can we share this code, perhaps in something called `NewDBWritableFileWriter`? Is the problem DBOptions vs. ImmutableCFOptions, etc.?

include/rocksdb/options.h
```diff
@@ -1192,6 +1192,16 @@ struct DBOptions {
...
+
+  // Use this if your DB want to enable checksum handoff for specific file
+  // types writes. Make sure that the File_system you use support the
+  // crc32c checksum verification
+  // Currently supported file tyes: kWALFile, kTableFile, kDescriptorFile.
+  // NOTE: currently RocksDB only generates crc32c based checksum for the
+  // handoff. If the storage layer has different checksum support, user
+  // should enble this vector is empty. Otherwise,it may cause unexpected
+  // write failures.
+  std::vector<FileType> checksum_handoff_file_types;
```
pdillinger, line 1204: vector is not a good abstraction for a set from a pretty small domain (11 values currently, so a set fits in uint16_t), and ties us to certain inefficiencies long term because this is public API. Inefficiency is not just membership query, but also copying DBOptions.

I'm envisioning a template<ENUM_TYPE,MAX_VALUE> class SmallEnumSet that is a wrapper around an unsigned scalar type of sufficient size to have one bit for each possible enum value (up to 64 for now, expandable in the future), given the largest enum value as a template parameter and assuming values are from 0 to that largest value. You could then have

    using FileTypeSet = SmallEnumSet<FileType, kBlobFile>;

db/blob/blob_file_builder.cc
zhichao-cao: I'm a little confuse, there is no FileType or handoff_checksum_type information in the FileOptions. So we may not be able to use the information in the FileOptios.

You mean, we pass the FileType as the parameter of WritableFileWriter constructor, and process the logic inside?

include/rocksdb/options.h
zhichao-cao: Yeah, it is more efficient not to use the vector. So is SmallEnumSet already in hour code base?

pdillinger: No, it would need to be implemented, but for now it should only need Add and Contains.

db/blob/blob_file_builder.cc
pdillinger: Example of code in other places that is not present here:

https://github.com/facebook/rocksdb/blob/484e1ffade99fae516057d03a4dfef418a54770a/db/builder.cc#L138-L140

zhichao-cao: > Example of code in other places that is not present here:
> 
> https://github.com/facebook/rocksdb/blob/484e1ffade99fae516057d03a4dfef418a54770a/db/builder.cc#L138-L140

Oh, I think I miss understood your previous comments. Will add the handoff_checksum_type here. Also, let me double check the code. If possible, I will remove "should_checksum_handoff" and replaced with the FileType as the input, such that ShouldChecksumHandoff is part of the construction function of WritableFileWriter.

include/rocksdb/options.h
zhichao-cao: > No, it would need to be implemented, but for now it should only need Add and Contains.

Sure, will add it later to include/rocksdb/utilities

utilities/data_structure_test.cc
```diff
@@ -0,0 +1,33 @@
...
+class SmallEnumSetTest : public testing::Test {
+ public:
+  SmallEnumSetTest() {}
+  ~SmallEnumSetTest() {}
+};
+
+TEST_F(SmallEnumSetTest, SmallSetTest) {
+  FileTypeSet fs;
+  ASSERT_TRUE(fs.Add(FileType::kIdentityFile));
+  ASSERT_TRUE(fs.Add(FileType::kInfoLogFile));
```
pdillinger, line 23: Need to test a case of Add returning false. (Otherwise, fine 👍 )

Recommended: combine with an existing extremely short-running test binary like slice_test or heap_test.

And this seems more like `util/` (useful outside of RocksDB) than `utilities/` (accessories specific to RocksDB)

include/rocksdb/utilities/data_structure.h
```diff
@@ -0,0 +1,45 @@
...
+// pretty small scale of Enum structure. For now, it can support up
+// to 64 element, and it is expandable in the future.
+template <typename ENUM_TYPE, ENUM_TYPE MAX_VALUE>
+class SmallEnumSet {
+ public:
+  SmallEnumSet() : state_(0) {}
+
+  ~SmallEnumSet() {}
+
+  bool Add(const ENUM_TYPE value) {
```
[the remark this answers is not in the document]
zhichao-cao: Sure.

utilities/data_structure_test.cc
zhichao-cao: Sure, will make the change

db/builder.cc
```diff
@@ -136,7 +137,9 @@ Status BuildTable(
       bool use_direct_writes = file_options.use_direct_writes;
       TEST_SYNC_POINT_CALLBACK("BuildTable:create_file", &use_direct_writes);
 #endif  // !NDEBUG
-      IOStatus io_s = NewWritableFile(fs, fname, &file, file_options);
+      FileOptions cur_file_opts = file_options;
+      cur_file_opts.handoff_checksum_type = ChecksumType::kCRC32c;
```
anand1976, line 141: Can this be set in the ```FileOptions``` constructor?

anand1976 (commented):
LGTM

db/builder.cc
zhichao-cao: > Can this be set in the `FileOptions` constructor?

Yes, I will add a new constructor for FileOptions to set the ChekcsumType.

db/builder.cc
```diff
@@ -136,7 +137,8 @@ Status BuildTable(
       bool use_direct_writes = file_options.use_direct_writes;
       TEST_SYNC_POINT_CALLBACK("BuildTable:create_file", &use_direct_writes);
 #endif  // !NDEBUG
-      IOStatus io_s = NewWritableFile(fs, fname, &file, file_options);
+      FileOptions cur_file_opts(file_options, ChecksumType::kCRC32c);
```
anand1976, line 140: Just to clarify my previous comment, is there any reason why ```ChecksumType::kCRC32c``` needs to be explicitly specified? Can it be the default in ```FileOptions```? Alternatively, it could be set in ```file_options_``` and ```file_options_for_compaction_``` in the ```DBImpl``` constructor, rather than everywhere ```NewWritableFile``` is called..

zhichao-cao: It might be better to set ChecksumType::kCRC32c as the default in FileOptions. Most likely, RocksDB internal will only use it for the handoff. If other types are supported in the future, we can set it explicitly. Thanks for the suggestion.

file/writable_file_writer.cc
```diff
@@ -420,8 +423,16 @@ IOStatus WritableFileWriter::WriteBuffered(const char* data, size_t size) {
 #endif
       {
         auto prev_perf_level = GetPerfLevel();
+
         IOSTATS_CPU_TIMER_GUARD(cpu_write_nanos, clock_);
-        s = writable_file_->Append(Slice(src, allowed), IOOptions(), nullptr);
+        if (perform_data_verification_) {
+          Crc32cHandoffChecksumCalculation(src, allowed, checksum_buf);
+          v_info.checksum = Slice(checksum_buf, sizeof(uint32_t));
```
pdillinger, line 430: Nit: consider std::array in place of C-style array, in part to avoid writing sizeof(uint32_t) twice.

file/writable_file_writer.cc
```diff
@@ -451,6 +462,13 @@ void WritableFileWriter::UpdateFileChecksum(const Slice& data) {
   }
 }
 
+void WritableFileWriter::Crc32cHandoffChecksumCalculation(const char* data,
```
pdillinger, line 465: Suggest a comment about how this code is intended to evolve

include/rocksdb/data_structure.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include "rocksdb/rocksdb_namespace.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// This is a data structure specifically designed as a "Set" for a
+// pretty small scale of Enum structure. For now, it can support up
+// to 64 element, and it is expandable in the future.
+template <typename ENUM_TYPE, ENUM_TYPE MAX_VALUE>
+class SmallEnumSet {
```
pdillinger, line 20: 👍

include/rocksdb/file_system.h
```diff
@@ -777,10 +790,14 @@ class FSWritableFile {
...
   }
 
-  // EXPERIMENTAL / CURRENTLY UNUSED
   // PositionedAppend data with verification information.
   // Note that this API change is experimental and it might be changed in
-  // the future. Currently, RocksDB does not use this API.
+  // the future. Currently, RocksDB only generates crc32c based checksum for
+  // the file writes when the option is set.
+  // Expected behavior: if the supported checksum verification type in this
+  // FSWritableFile does not match the handoff_checksum_type in FileOptions,
```
pdillinger, line 798: Do you mean FileOptions passed in the constructor? Why do we pass the checksum type twice in the constructor and only use it if they agree?

pdillinger (commented):
Getting close. One question.

include/rocksdb/file_system.h
pdillinger: Sorry, I mixed up the FileOptions and FSWritableFile constructors.

"the supported checksum verification type in this FSWritableFile" - I think this confused me a bit because it sounds like the FSWritableFile has a field for the checksum type. What I think you mean now is (rephrased) "the checksum verification type supported by this FSWritableFile". Suggested edit:

    // Expected behavior: if the handoff_checksum_type in FileOptions is not supported by
    // this FSWritableFile, the information in DataVerificationInfo should be ignored (i.e. does
    // not perform checksum verification)

I suggest we also consider changing "should" to "can," so that implementations can choose to "fail safe" in the case of a configuration error rather than "quietly skip verification." But it's up to you, and I'm not blocking approval on it.

pdillinger (approved):
Significant step in an interesting and valuable journey 😉

include/rocksdb/file_system.h
zhichao-cao: Thanks for the clarification. I will use the suggested statement in the comments. Also, using "can" makes it more flexible such that the developer of FSWritableFile can decide the behavior, like "fail" the writes if they want.

zhichao-cao:
> Significant step in an interesting and valuable journey 😉

Thanks for all of the suggestions and comments during the review. Stage 2 is on the way to close the checksum protection gap.

PR #7529, opened 2020-10-10
remove useless variable

914022466:
right_left_diff is always equal to bucket_value, so  right_sum and right_left_diff are useless.
bucket_value cannot be 0. Because if bucket_value is 0, then previous loop will reach the condition `cumulative_sum >= threshold`


monitoring/histogram.cc
```diff
@@ -149,12 +149,8 @@ double HistogramStat::Percentile(double p) const {
       uint64_t left_point = (b == 0) ? 0 : bucketMapper.BucketLimit(b-1);
       uint64_t right_point = bucketMapper.BucketLimit(b);
       uint64_t left_sum = cumulative_sum - bucket_value;
-      uint64_t right_sum = cumulative_sum;
       double pos = 0;
```
darionyaphet, line 152: why don't move define and assign into one line?

```
double pos = (threshold - left_sum) / bucket_value;
```

monitoring/histogram.cc
```diff
@@ -149,12 +149,8 @@ double HistogramStat::Percentile(double p) const {
...
-      uint64_t right_left_diff = right_sum - left_sum;
-      if (right_left_diff != 0) {
-       pos = (threshold - left_sum) / right_left_diff;
-      }
+      pos = (threshold - left_sum) / bucket_value;
```
jay-zhuang, line 153: what if `bucket_value == 0`?

PR #7532, opened 2020-10-10
Abstract out LockManager interface

cheng-chang:
In order to be able to introduce more locking protocols, we need to abstract out the locking subsystem in TransactionDB into a set of interfaces.

PR https://github.com/facebook/rocksdb/pull/7013 introduces interface `LockTracker`. This PR is a follow up to take the first step to abstract out a `LockManager` interface.

Further modifications to the interface may be needed when introducing the first implementation of range lock. But the idea here is to put the range lock implementation based on range tree under the `utilities/transactions/lock/range/range_tree`.

Test Plan:
point_lock_manager_test


spetrunia:
A question about this call:

```c++
class LockManager {
  ...
  // Enable locking for the specified column family.
  // Caller should guarantee that this column family is not already enabled.
  virtual void AddColumnFamily(ColumnFamilyId column_family_id) = 0;
```

Here, the LockManager is expected to initialize its data structures based on just ColumnFamilyId (not ColumnFamilyHandle).
Current PointLockManager can do this because it doesn't care about the Comparator used, it assumes the comparator is memcmp()  (one can see this in `class LockMapStripe`: the keys are indexed with `std::unordered_map<std::string, LockInfo> keys`). 

A Range-based Lock manager (or a more advanced Point-based lock manager) needs to "know" what Comparator is used.

I can see two solutions:
Solution1: Pass ColumnFamilyHandle* instead (this is what current Range Locking code is using)

Solution2: Assume that the lock manager can somehow call
```
  db_.versions_.GetColumnFamily(id)->GetComparator();
```
to get the Comparator. I'm not sure if/how to do that safely though (`versions_` is not public, etc).

Should this issue be addressed as part of this patch?

utilities/transactions/pessimistic_transaction_db.cc
```diff
@@ -597,17 +587,16 @@ void PessimisticTransactionDB::GetAllPreparedTransactions(
   }
 }
 
-TransactionLockMgr::LockStatusData
-PessimisticTransactionDB::GetLockStatusData() {
-  return lock_mgr_.GetLockStatusData();
+LockManager::PointLockStatus PessimisticTransactionDB::GetLockStatusData() {
+  return lock_manager_->GetPointLockStatus();
```
spetrunia, line 591: So, now LockManager got two functions, `GetPointLockStatus` and `GetRangeLockStatus`. 

Should this function (`PessimisticTransactionDB::GetLockStatusData`) also be replaced with two functions? (And if yes, should this be done in this PR or outside of it?)

cheng-chang: I'll add these two methods and add a comment in `TransactionDB` to mention that `GetLockStatusData` is deprecated in favor of `GetPointLockStatus`.

cheng-chang:
> A question about this call:
> 
> ```c++
> class LockManager {
>   ...
>   // Enable locking for the specified column family.
>   // Caller should guarantee that this column family is not already enabled.
>   virtual void AddColumnFamily(ColumnFamilyId column_family_id) = 0;
> ```
> 
> Here, the LockManager is expected to initialize its data structures based on just ColumnFamilyId (not ColumnFamilyHandle).
> Current PointLockManager can do this because it doesn't care about the Comparator used, it assumes the comparator is memcmp() (one can see this in `class LockMapStripe`: the keys are indexed with `std::unordered_map<std::string, LockInfo> keys`).
> 
> A Range-based Lock manager (or a more advanced Point-based lock manager) needs to "know" what Comparator is used.
> 
> I can see two solutions:
> Solution1: Pass ColumnFamilyHandle* instead (this is what current Range Locking code is using)
> 
> Solution2: Assume that the lock manager can somehow call
> 
> ```
>   db_.versions_.GetColumnFamily(id)->GetComparator();
> ```
> 
> to get the Comparator. I'm not sure if/how to do that safely though (`versions_` is not public, etc).
> 
> Should this issue be addressed as part of this patch?

Solution1 looks good to me, I'll do it in this PR.

cheng-chang:
@spetrunia PTAL

spetrunia:
@cheng-chang looks good to me now.

spetrunia (approved):

ajkr (approved):
LGTM.

utilities/transactions/lock/lock_manager.cc
```diff
@@ -0,0 +1,23 @@
...
+
+#include "utilities/transactions/lock/point/point_lock_manager.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+LockManager* NewLockManager(PessimisticTransactionDB* db,
+                            const TransactionDBOptions& opt) {
+  assert(db);
+  // TODO: determine the lock manager implementation based on configuration.
+  return new PointLockManager(db, opt);
```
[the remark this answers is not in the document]
spetrunia: The first one.  When `RangeLockManager` is used, acquiring a point lock on $P is the same as acquiring a range lock on `$P <= key <= $P`.

cheng-chang:
@lth I'm going to merge the PR to unblock Sergey, feel free to leave reviews, I'll address them in a follow up PR.

include/rocksdb/utilities/transaction_db.h
```diff
@@ -202,6 +202,13 @@ struct KeyLockInfo {
   bool exclusive;
 };
 
+struct RangeLockInfo {
+  Endpoint start;
+  Endpoint end;
+  std::vector<TransactionID> ids;
+  bool exclusive;
```
lth, line 209: Does range locking support shared locks? Or is this here just for future extensibility?

utilities/transactions/pessimistic_transaction_db.cc
lth: I talked with @cheng-chang  about this, and I think that in the future, we should always get status from the lock manager objects, and the methods on `PessimisticTransactionDB` will go away.

I think the below calls relating to deadlock info would be similar too.

lth (commented):
Looks pretty good to me too.

Regarding needing the handle for comparators for `AddColumnFamily`, write prepared needed to solve a similar problem, so it'd be good if the two solutions converged eventually. This is implemented in `UpdateCFComparatorMap`. Maybe we can add a TODO?

include/rocksdb/utilities/transaction_db.h
cheng-chang: For future extensibility.

Issue #7531, opened 2020-10-10
Always print hex for keys with corrupt results in stress tests


ajkr:
Currently some failures look like this:

```
***Verification failed for column family 0 key 87500643: Value not found: NotFound:***
```

However this is difficult to debug since the translation from integer used to generate the key into the actual key bytes is complex and configuration-dependent. That makes it difficult to do manual investigation like looking for the key in the SST files. It'd be good to accompany the key integer with the actual key bytes (in hex format) in all error messages printed by the stress tests.

PR #7533, opened 2020-10-10
db_stress prints key in Hex

jay-zhuang:
db_stress prints key in both id and hex. #7531

Test Plan: local test


zhichao-cao (approved):
LGTM, thanks!

Issue #7530, opened 2020-10-10
Preserve expected values file for failed crash tests


ajkr:
In crash tests ("tools/db_crashtest.py"), `expected_values_file` is assigned `tempfile.NamedTemporaryFile()`, which is deleted as soon as it's closed according to Python docs. For debugging corruptions, it'd be good to preserve this file when the crash test fails. That way we can rerun the final failed command using the same expected state that the failed command had, and hopefully repro the corruption.

PR #7534, opened 2020-10-11
db_crashtest preserves expected values file for failed crash tests

jay-zhuang:
If crash test fails, don't delete the `expected_values_file` for later
debug. More details: #7530

Test Plan: local host

### DIFF
--- a/tools/db_crashtest.py
+++ b/tools/db_crashtest.py
@@ -25,7 +25,7 @@ import argparse
 #   for txn:
 #       default_params < {blackbox,whitebox}_default_params < txn_params < args
 
-expected_values_file = tempfile.NamedTemporaryFile()
+expected_values_file = tempfile.NamedTemporaryFile(delete=False)
 
 default_params = {
     "acquire_snapshot_one_in": 10000,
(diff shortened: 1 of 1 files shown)


ajkr (approved):
LGTM, thanks for the fast fix!

BTW, there is an upload step for failed runs in CI (see internal wiki `w MySQL/RocksDB/Oncall/Stress_test/`). We might need to somehow add this file to include it in the tar.

jay-zhuang:
> LGTM, thanks for the fast fix!
> 
> BTW, there is an upload step for failed runs in CI (see internal wiki `w MySQL/RocksDB/Oncall/Stress_test/`). We might need to somehow add this file to include it in the tar.

thanks for pointing out. Changed the `expected_values_file` the store place also to the same `tmp_dir` if it's specified. Then it will be upload together with db, also will be clean up together. for example, if the tmp dir is `/tmp/rocksdb`, it will create a db folder and a expected value file:
```
zjay@zjay-mbp /tmp/rocksdb% ll
total 0
-rw-r--r--  1 zjay  wheel     0B Oct 12 10:57 rocksdb_crashtest_expected
drwxr-xr-x  2 zjay  wheel    64B Oct 12 10:57 rocksdb_crashtest_whitebox
```

PR #7536, opened 2020-10-11
Add macos tests to circleci

cheng-chang:
as title

Test Plan:
see the new `build-macos` tests pass in circleci


cheng-chang:
The tests on macos passed.
But the 2xlarge machine type is not supported in macos, the tests took 30min.
While on linux 2xlarge, the same set of tests took 12min.

jay-zhuang (approved):
LGTM.

PR #7539, opened 2020-10-12
Change ErrorHandler methods to return const Status&

mrambacher:
This change eliminates the need for a lot of the PermitUncheckedError calls on return from ErrorHandler methods.  The calls are no longer needed as the status is returned as a reference rather than a copy.  Additionally, this means that the originating status (recovery_error_, bg_error_) is not cleared implicitly as a result of calling one of these methods.

For this class, I do not know if the proper behavior should be to call PermitUncheckedError in the destructor or if the checked state should be cleared when the status is cleared.  I did tests both ways.  Without the code in the destructor, the status will need to be cleared in at least some of the places where it is set to OK.  When running tests, I found no instances where this class was destructed with a non-OK, non-checked Status.


db/error_handler.cc
```diff
@@ -353,8 +353,8 @@ Status ErrorHandler::SetBGError(const IOStatus& bg_io_err,
     if (recovery_in_prog_ && recovery_error_.ok()) {
       recovery_error_ = bg_err;
     }
-    EventHelpers::NotifyOnBackgroundError(db_options_.listeners, reason, &s,
-                                          db_mutex_, &auto_recovery);
+    EventHelpers::NotifyOnBackgroundError(db_options_.listeners, reason,
+                                          &bg_err, db_mutex_, &auto_recovery);
```
anand1976, line 357: Looks like a bug in the original code. Good catch.

db/error_handler.h
```diff
@@ -53,15 +50,16 @@ class ErrorHandler {
...
 
-   Status SetBGError(const Status& bg_err, BackgroundErrorReason reason);
+   const Status& SetBGError(const Status& bg_err, BackgroundErrorReason reason);
 
-   Status SetBGError(const IOStatus& bg_io_err, BackgroundErrorReason reason);
+   const Status& SetBGError(const IOStatus& bg_io_err,
+                            BackgroundErrorReason reason);
 
-   Status GetBGError() { return bg_error_; }
+   const Status& GetBGError() const { return bg_error_; }
```
anand1976, line 58: Is it necessary to return ```const Status&``` from the ```Get*Error``` functions? Since its a getter, the caller would be expected to check the status.

mrambacher: Actually it is important to return a & from these functions for a couple of reasons:
- Without a reference we needlessly make a copy of the member variable;
- If we do not return a reference, then the act of calling this function will cause the *_error_ to be marked as checked. This is because the Status copy constructor will clear the checked flag on the original;
- If we do not return a reference, then the returned copy must be explicitly checked somewhere.  This lead to a whole bunch of calls to permit unchecked errors when we returned a status we never cared about (and was checked implicitly as a member variable already).

anand1976: > * If we do not return a reference, then the returned copy must be explicitly checked somewhere.  This lead to a whole bunch of calls to permit unchecked errors when we returned a status we never cared about (and was checked implicitly as a member variable already).

It seems strange that the caller of ```GetBGError``` will not check the return status. That would be an obvious bug. I think we should let ```GetBGError``` return Status and fix the callers. FWIW, I didn't see any instance of ```PermitUncheckedError``` on the status returned by ```GetBGError```.

mrambacher: Could be.  OTOH, is there a reason to return a copy of a member variable, rather than a const & ?  I thought the latter was always considered good programming practice...

pdillinger: These Set* and Clear* seem quite different from Get* for obligation to check return status. Why do Set* and Clear* return a status anyway?

> I thought the latter was always considered good programming practice...

It mixes in the concerns of object lifetime and thread shared data, which can cause subtle bugs. Use with caution.

mrambacher: I have no idea why Clear* and Set* return Status.  I agree it would be better if they did not, but I do not know feel as if I should tackle that issue.

mrambacher:
Can we decide what to do with this PR?  It has been sitting idle for awhile now...

include/rocksdb/status.h
```diff
@@ -62,6 +62,13 @@ class Status {
   bool operator==(const Status& rhs) const;
   bool operator!=(const Status& rhs) const;
 
+  inline bool IsChecked() const {
```
pdillinger, line 65: Was this used for debugging? It might invite misuse to leave it

db/error_handler.h
```diff
@@ -53,13 +50,14 @@ class ErrorHandler {
                                      Status::Code code,
                                      Status::SubCode subcode);
 
-   Status SetBGError(const Status& bg_err, BackgroundErrorReason reason);
+   const Status& SetBGError(const Status& bg_err, BackgroundErrorReason reason);
```
pdillinger, line 53: I see that because of some complex, possibly erroneous logic in StartRecoverFromRetryableBGIOError we can't equivalently make SetBGError void and have interested callers call GetBGError(). Can think about that later.

db/error_handler.h
pdillinger: > This lead to a whole bunch of calls to permit unchecked errors when we returned a status we never cared about (and was checked implicitly as a member variable already).

I don't think you ever clarified whether this actually applies to GetBGError, because it seems to describe callers to SetBGError. In quick review, I don't see any callers of GetBGError that would have distinct behavior for returning Status& vs. Status. Anand and I prefer this one return Status.

pdillinger (changes requested):
I also see more obsolete calls to PermitUncheckedError after SetBGError than are corrected here. (Perhaps they were added after this PR was prepared.)

db/error_handler.h
mrambacher: I can change the GetBGError back to Status if you really want, but would question why.  When we return a Status (not reference), we end up with the potential for making an extra ephemeral copy.  A majority of the calls to GetBGError are "Status s = GetBGError();", which means we copy the reference on assignment anyway.

I can change it back, but it seems more efficient and proper to use a const reference...

mrambacher:
I fixed all of the ErrorHandler::SetBGError().PermitUncheckedError calls except one.  The remainder causes one of the tests to fail (ColumnFamilyTest::WriteBatchFailure), and I do not have cycles to investigate that test at the moment.

pdillinger (approved):
LGTM now, thanks. Do a fresh merge with master and I'll be able to land it.

anand1976 (approved):
LGTM. Thanks for the PR.

anand1976:
@mrambacher Sorry for the delay in stamping the PR. You'll have to merge with master one more time to resolve the conflict.

PR #7540, opened 2020-10-12
Introduce BlobFileCache and add support for blob files to Get()

ltamasi:
Summary:
The patch adds blob file support to the `Get` API by extending `Version` so that
whenever a blob reference is read from a file, the blob is retrieved from the corresponding
blob file and passed back to the caller. (This is assuming the blob reference is valid
and the blob file is actually part of the given `Version`.) It also introduces a cache
of `BlobFileReader`s called `BlobFileCache` that enables sharing `BlobFileReader`s
between callers. `BlobFileCache` uses the same backing cache as `TableCache`, so
`max_open_files` (if specified) limits the total number of open (table + blob) files.

TODO: proactively open/cache blob files and pin the cache handles of the readers in the
metadata objects similarly to what `VersionBuilder::LoadTableHandlers` does for
table files.

Test Plan:
`make check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -588,6 +588,7 @@ ifdef ASSERT_STATUS_CHECKED
 		crc32c_test \
 		dbformat_test \
 		db_basic_test \
+		db_blob_basic_test \
 		db_flush_test \
 		db_with_timestamp_basic_test \
 		db_with_timestamp_compaction_test \
@@ -687,6 +688,7 @@ endif
 # Not necessarily well thought out or up-to-date, but matches old list
 TESTS_PLATFORM_DEPENDENT := \
 	db_basic_test \
+	db_blob_basic_test \
 	db_with_timestamp_basic_test \
 	db_encryption_test \
 	db_test2 \
@@ -1460,6 +1462,9 @@ slice_transform_test: $(OBJ_DIR)/util/slice_transform_test.o $(TEST_LIBRARY) $(L
 	$(AM_LINK)
 
 db_basic_test: $(OBJ_DIR)/db/db_basic_test.o $(TEST_LIBRARY) $(LIBRARY)
+	$(AM_LINK)
+
+db_blob_basic_test: $(OBJ_DIR)/db/blob/db_blob_basic_test.o $(TEST_LIBRARY) $(LIBRARY)
 	$(AM_LINK)
 
 db_with_timestamp_basic_test: $(OBJ_DIR)/db/db_with_timestamp_basic_test.o $(TEST_LIBRARY) $(LIBRARY)
--- a/Makefile
+++ b/Makefile
@@ -577,6 +577,7 @@ ifdef ASSERT_STATUS_CHECKED
 		lru_cache_test \
 		blob_file_addition_test \
 		blob_file_builder_test \
+		blob_file_cache_test \
 		blob_file_garbage_test \
 		blob_file_reader_test \
 		bloom_test \
@@ -1917,6 +1918,9 @@ blob_file_addition_test: $(OBJ_DIR)/db/blob/blob_file_addition_test.o $(TEST_LIB
 	$(AM_LINK)
 
 blob_file_builder_test: $(OBJ_DIR)/db/blob/blob_file_builder_test.o $(TEST_LIBRARY) $(LIBRARY)
+	$(AM_LINK)
+
+blob_file_cache_test: $(OBJ_DIR)/db/blob/blob_file_cache_test.o $(TEST_LIBRARY) $(LIBRARY)
 	$(AM_LINK)
 
 blob_file_garbage_test: $(OBJ_DIR)/db/blob/blob_file_garbage_test.o $(TEST_LIBRARY) $(LIBRARY)
--- a/db/blob/blob_file_cache.h
+++ b/db/blob/blob_file_cache.h
@@ -1,0 +1,44 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#pragma once
+
+#include <cinttypes>
+
+#include "rocksdb/rocksdb_namespace.h"
+#include "util/mutexlock.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+class Cache;
+struct ImmutableCFOptions;
+struct FileOptions;
+class HistogramImpl;
+class Status;
+class BlobFileReader;
+class Slice;
+
+class BlobFileCache {
+ public:
+  BlobFileCache(Cache* cache, const ImmutableCFOptions* immutable_cf_options,
+                const FileOptions* file_options, uint32_t column_family_id,
+                HistogramImpl* blob_file_read_hist);
+
+  BlobFileCache(const BlobFileCache&) = delete;
+  BlobFileCache& operator=(const BlobFileCache&) = delete;
+
+  Status GetBlobFileReader(uint64_t blob_file_number,
+                           BlobFileReader** blob_file_reader);
+
+ private:
+  Cache* cache_;
+  Striped<port::Mutex, Slice> mutex_;
+  const ImmutableCFOptions* immutable_cf_options_;
+  const FileOptions* file_options_;
+  uint32_t column_family_id_;
+  HistogramImpl* blob_file_read_hist_;
+};
+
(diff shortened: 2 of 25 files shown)


db/blob/db_blob_basic_test.cc
```diff
@@ -0,0 +1,188 @@
...
+ protected:
+  DBBlobBasicTest()
+      : DBTestBase("/db_blob_basic_test", /* env_do_fsync */ false) {}
+};
+
+TEST_F(DBBlobBasicTest, GetBlob) {
+  Options options;
+  options.enable_blob_files = true;
+  options.min_blob_size = 0;
+  options.disable_auto_compactions = true;
```
riversand963, line 24: Maybe add a `TODO` to enable auto compactions?

db/blob/db_blob_index_test.cc
```diff
@@ -171,8 +174,13 @@ TEST_F(DBBlobIndexTest, Get) {
     ASSERT_EQ("value", GetImpl("key", &is_blob_index));
     ASSERT_FALSE(is_blob_index);
     // Verify blob index
-    ASSERT_TRUE(Get("blob_key", &value).IsNotSupported());
-    ASSERT_EQ("NOT_SUPPORTED", GetImpl("blob_key"));
+    if (tier <= kImmutableMemtables) {
```
riversand963, line 177: Should the comment on line 159 be updated?

riversand963 (commented):
Thanks @ltamasi for the PR. Mostly LGTM with a few minor comments and questions.

Makefile
```diff
@@ -682,6 +684,7 @@ endif
 # Not necessarily well thought out or up-to-date, but matches old list
 TESTS_PLATFORM_DEPENDENT := \
 	db_basic_test \
+	db_blob_basic_test \
```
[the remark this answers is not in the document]
ltamasi: I've always thought that these are just the tests we want to run on every supported platform but as it turns out it's just a contiguous subset of the tests (similarly to the groups 1..4 used for load balancing).

db/version_set.h
```diff
@@ -1265,6 +1278,7 @@ class VersionSet {
   WalSet wals_;
 
   std::unique_ptr<ColumnFamilySet> column_family_set_;
+  Cache* table_cache_;
```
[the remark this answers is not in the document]
ltamasi: That's part of a small cleanup: before the patch, `VersionSet` did not have a pointer to the backing cache, so anytime it needed it, it had to retrieve it from `ColumnFamilySet`, resulting in awkward code like this for example:

```
void VersionSet::Reset() {
  if (column_family_set_) {
    Cache* table_cache = column_family_set_->get_table_cache();
    WriteBufferManager* wbm = column_family_set_->write_buffer_manager();
    WriteController* wc = column_family_set_->write_controller();
    column_family_set_.reset(
        new ColumnFamilySet(dbname_, db_options_, file_options_, table_cache,
                            wbm, wc, block_cache_tracer_, io_tracer_));
  }
```

db/blob/blob_file_cache.h
```diff
@@ -0,0 +1,47 @@
...
+
+ private:
+  Cache* cache_;
+  Striped<port::Mutex, Slice> mutex_;
+  const ImmutableCFOptions* immutable_cf_options_;
+  const FileOptions* file_options_;
+  uint32_t column_family_id_;
+  HistogramImpl* blob_file_read_hist_;
+
+  static constexpr size_t kNumberOfMutexStripes = 1 << 7;
```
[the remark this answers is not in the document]
ltamasi: Not really ;) I cribbed the striped mutex idea as well as the value from `TableCache`.

db/blob/db_blob_index_test.cc
ltamasi: Yes, nice catch, thanks!

db/version_set.cc
```diff
@@ -1865,6 +1925,18 @@ void Version::Get(const ReadOptions& read_options, const LookupKey& k,
         // TODO: update per-level perfcontext user_key_return_count for kMerge
         break;
       case GetContext::kFound:
+        if (is_blob_index) {
+          if (do_merge && value) {
```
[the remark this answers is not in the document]
ltamasi: Yes, I think it does: `do_merge` is actually a misnomer since it's supposed to mean "get value" (it's what differentiates the `Get` and `GetMergeOperands` APIs). From `version_set.h`:

```
  // Lookup the value for key or get all merge operands for key.
  // If do_merge = true (default) then lookup value for key.
```

db/blob/db_blob_basic_test.cc
ltamasi: I think the very first test case (which does two `Get`s, the second with `read_tier` set to `kBlockCacheTier`) needs it to be set to ensure the SST and the block are found in the cache. For the rest, I think we could remove it now.

db/version_set.h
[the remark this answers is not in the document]
riversand963: Maybe it's personal preference, and I may have been spoiled by powerful modern hardware. XD
Access to the table cache pointer is not frequent, thus I think it's fine to do the computation `Cache* table_cache = column_family_set_->get_table_cache();` each time. In such cases, I prefer doing computation on-demand (maybe using a function to wrap it) instead of introducing new data member to class.
Anyway, I do not have a strong opinion, and it's up to your choice.

db/blob/db_blob_basic_test.cc
ltamasi: Actually, we can remove it from the first test case too: automatic compactions wouldn't kick in with a single L0 file and nothing else.

db/version_set.h
```diff
@@ -776,6 +777,13 @@ class Version {
     return storage_info_.user_comparator_;
   }
 
+  // Interprets *value as a blob reference, and (assuming the corresponding
+  // blob file is part of this Version) retrieves the blob and saves it in
+  // *value, replacing the blob reference.
+  // REQUIRES: *value stores an encoded blob reference
+  Status GetBlob(const ReadOptions& read_options, const Slice& user_key,
```
riversand963, line 784: Any reason why use `value` as both input and output argument?

db/blob/db_blob_index_test.cc
```diff
@@ -154,7 +157,8 @@ TEST_F(DBBlobIndexTest, Write) {
 }
 
 // Get should be able to return blob index if is_blob_index is provided,
-// otherwise return Status::NotSupported status.
+// otherwise it should return Status::NotSupported (when reading from memtable)
+// or Status::Corruption (when reading from SST).
```
riversand963, line 161: On second thought, can you remind me why `Corruption` is returned if `is_blob_index == nullptr` and read from SST? I am thinking maybe `InvalidArgument` is a better fit?

db/version_set.h
ltamasi: It minimizes data movement. `value` here is the actual `PinnableSlice` passed in by the user. It makes sense for `Version::Get` to pass it (in `GetContext`) to `TableCache::Get` to eliminate additional copies/moves in the common case when the value is not a `BlobIndex`. On the other hand, when it *is* a `BlobIndex`, the value read from the table (which is in `value` at that point) becomes the input for `GetBlob`, and I feel it again makes sense to use it as output as well to eliminate the need for any further copying/moving. (In other words, if we were to use separate input and output arguments for `GetBlob`, we would eventually need to move `output_value` back to `value` at the call site.)

db/blob/db_blob_index_test.cc
ltamasi: I should probably tweak that comment a bit more. :) Here's the deal: that's actually a test case related to the old `StackableDB` based BlobDB. Namely, `GetImplOptions::is_blob_index` is only non-null when the old BlobDB is used. Historically, ending up with a `BlobIndex` when `is_blob_index` is `nullptr` meant that the user had accidentally opened the base DB of a `BlobDB` instance (i.e. used `DB::Open` instead of `BlobDB::Open`), thus we returned `NotSupported` with a message instructing the user to use `BlobDB::Open` instead.

With the new integrated BlobDB in the picture, the above assumption no longer holds, that is, we can't really tell apart a) the above case of the application accidentally opening the base DB of a stacked BlobDB and b) a real corruption when using the new integrated BlobDB (an example would be a `BlobIndex` pointing to a blob file not in the `Version`). Since we have two pick one of the two, I went with the second one since I think the first case is extremely unlikely and the message would be pretty confusing for users.

riversand963 (approved):
LGTM! My questions and concerns have been addressed.

PR #7541, opened 2020-10-12
cmake: Use exported target for bz2

bachp:
Without this change, CMake puts the entire path into the generated file.
This is not portable and makes it fail in environment like Yocto.

Not that this requires a CMake version bump, as `BZip2::BZip2` is only available since CMake 3.7.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@
 # 3. cmake ..
 # 4. make -j
 
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.7.2)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/modules/")
 include(ReadVersion)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -32,7 +32,7 @@
 # 3. cmake ..
 # 4. make -j
 
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.7.2)
```
jay-zhuang, line 35: Do we have to increase minimum cmake version?

bachp: Yes as mentioned in the PR description:
> Note that this requires a CMake version bump, as `BZip2::BZip2` is only available since CMake 3.7.

jay-zhuang: Seems circleci image `ubuntu-1604:201903-01` only has `3.5.1` installed: https://app.circleci.com/pipelines/github/facebook/rocksdb/4088/workflows/1cf5ed7a-f437-482c-9012-1e4aa4904eb5/jobs/44573
Not sure how to upgrade to a newer version.

bachp: That's what I was afraid of. I see two options:
1. Wait merging this release until after ubuntu 16.04 is no longer supported
2. Keep a more up to date copy of `FindBzip2.cmake` in this repository until we can update to a newer version.

jay-zhuang: would be great if we could have option 2 before we could upgrade cmake.

PR #7543, opened 2020-10-13
Return error if Get/Multi() fails in Prefetching Filter blocks

akankshamahajan15:
Summary: Right now all I/O failures under
PartitionFilterBlock::CacheDependencies() is swallowed. Return error in
case prefetch fails.

On returning error in PartitionedFilterBlockReader::CacheDependencies was causing stress test failure because PrefetchBuffer is initialized with enable_ = true, as result when PosixMmapReadableFile::Read is called from Prefetch, scratch is ignored causing buffer to fill with garbage values. Initializing prefetch buffer by CreatePrefetchBuffer that sets enable_ with !ioptions.allow_mmap_reads fixed the problem as it returns without prefetching data if allow_mmap_reads is set.

Test Plan: make check -j64;
python -u tools/db_crashtest.py --simple blackbox

Reviewers:

Subscribers:

Tasks:

Tags:


anand1976 (approved):
LGTM

PR #7549, opened 2020-10-13
Add a missing bug fix in HISTORY.md

jay-zhuang:


ajkr (commented):
Do you plan to add the note on 6.14.fb branch as well? I usually just append it to the "Unreleased" section and say "Since 6.14, \<bug fix note\>". But if you want to try keeping the "6.14" section on 6.14.fb and master branches in sync, we can try that too.

jay-zhuang:
> Do you plan to add the note on 6.14.fb branch as well? I usually just append it to the "Unreleased" section and say "Since 6.14, <bug fix note>". But if you want to try keeping the "6.14" section on 6.14.fb and master branches in sync, we can try that too.

I see. I thought `Since x.xx` means the bug happens since version `x.xx`. Will just use that.

ajkr (approved):
LGTM.

PR #7553, opened 2020-10-14
Fix for clang_analyzer build failure in table_test

akankshamahajan15:
Summary: fix for clang_analyzer build failure in table_test because of
potential memory leak of memtable in case of ASSERT failure.

Test Plan: USE_CLANG=1 make analyze;
           make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:


zhichao-cao (approved):
LGTM, thanks for the fix!

PR #7554, opened 2020-10-14
Test for LoadLatestOptions

mrambacher:
Make LoadLatestOptions return PathNotFound if the options file does not exist.  Added tests for the LoadOptions related methods.


zhichao-cao (commented):
LGTM, thanks for the quick fix!

zhichao-cao (approved):

zhichao-cao:
Thanks for updating the testing case for the bad last option files.

options/options_parser.cc
```diff
@@ -289,13 +289,13 @@ Status RocksDBOptionsParser::Parse(const ConfigOptions& config_options_in,
...
+      // If the option file is newer than the current version
+      // there may be unknown options.
+      if (!config_options.ignore_unknown_options &&
           section == kOptionSectionVersion) {
-        if (db_version[0] < ROCKSDB_MAJOR || (db_version[0] == ROCKSDB_MAJOR &&
-                                              db_version[1] <= ROCKSDB_MINOR)) {
-          config_options.ignore_unknown_options = false;
+        if (db_version[0] > ROCKSDB_MAJOR ||
+            (db_version[0] == ROCKSDB_MAJOR && db_version[1] > ROCKSDB_MINOR)) {
+          config_options.ignore_unknown_options = true;
```
riversand963, line 298: The change here broke one of the internal tests.
The test bumps up the version number in the OPTIONS file and adds an unknown option. The test expects an error to be returned if `ignore_unknown_options` is false.
The old logic and the new logic are slightly different.
The old code makes the check more strict for old versions, while the new logic silently relaxes the check for new versions.

ajkr: API changes are required to be mentioned in HISTORY.md, and should not be included in PRs titled "Test for X".

mrambacher: Personally, I do not think this code should be here at all.  If the caller said to ignore unknown options, we ignore them regardless of the version.  In the old code, we ignored the user's preference (the default is FALSE, so they had to explicitly say to ignore the value but that was disabled).  

In the new code, we ignore options if the options file is newer.   I would argue this is (slightly) better in that we might have overridden the default value and a newer option file would be likely to have values that are unknown.  In this case, I would also argue that the internal test is broken.

If you agree, I will push a change that removes this code entirely and we can see what else breaks....

riversand963: > Personally, I do not think this code should be here at all. If the caller said to ignore unknown options, we ignore them regardless of the version. In the old code, we ignored the user's preference (the default is FALSE, so they had to explicitly say to ignore the value but that was disabled).
> 

Regarding the old code. I wonder what the behavior should be if 1) a new RocksDB release gets rid of an option "foo", while an older release has "foo" in OPTIONS file, and 2) new RocksDB binary tries to load the old OPTIONS file with `ignore_unknown_options=true`. The old code will change `ignore_unknown_options` to false.

> In the new code, we ignore options if the options file is newer. I would argue this is (slightly) better in that we might have overridden the default value and a newer option file would be likely to have values that are unknown. In this case, I would also argue that the internal test is broken.
> 

Would it be possible to let (ignore_unknown_options = false) take highest precedence here? I am thinking of something like
```
if (!ignore_unknown_options) {
  This means user always wants to be notified of any error encountered in loading options.
  always report error;
} else if (options file is older release) {
  User does not seem to care too much, but we want to enforce some extra checks.
  always report error;
} else {
  we may ignore unknown options.
}
```

> If you agree, I will push a change that removes this code entirely and we can see what else breaks....

mrambacher: I would argue that "ignore_unknown_options=false" should not take higher precedence because it is the default.  The user had to do something explicit to set the value to true, so that should take higher precedent.  The old code would ignore the user's wish.  Is there a reason not to delete these lines of code entirely?

riversand963: Removing these lines will still be less strict than what we used to have, and I think the old behavior arguably makes sense. If trying to load an OPTIONS file generated by older RocksDB release, an unknown option name should be reported because this should not happen (I am still not sure about the case in which old option names are removed).

Whether we stick to the old code or remove these lines, we should not silently change `ignore_unknown_options` from false to true.

mrambacher: The old behavior would silently change the flag as well -- just from true to false.  

- In the original code, ignore would change from true to false if the version was OLD
- In the current code, ignore would change from false to true if the version was NEW
- In the proposed code, ignore will NOT change

But I would argue the ORIGINAL code was wrong as the ONLY way that ignore came in as true was if the user explicitly asked for it, which means that we would be silently overriding their requests, which I would argue is wrong.  

As for older options being removed, I do not think that happens -- they are typically marked as deprecated and the code just ignores them.  If they were removed from the system, why should an "old option that no longer exists" be treated any differently than a "new option that does not yet exist"?

riversand963: > The old behavior would silently change the flag as well -- just from true to false.
> 
> * In the original code, ignore would change from true to false if the version was OLD
>
I think this is fine.

> * In the current code, ignore would change from false to true if the version was NEW
>
I think this is problematic. By setting the flag to false, the application indicates that errors should be reported. By silently changing it to true, we break the contract in a more dangerous way because there is a risk of application not being notified of an potential issue.

> * In the proposed code, ignore will NOT change
>
This feels borderline, but it is a change to API behavior which relaxes some checks. If the user sets `ignore_unknown_options` to `true`, and RocksDB detects some issue which should not happen, should RocksDB notify the user even if the user wants to ignore? IMO, it is better to still report the error. If a bad OPTIONS file is loaded, it may lead to unexpected consequences. In this specific case of old OPTIONS file containing unknown option name (to a new release), it is highly likely that the OPTIONS file has been tampered with, and application should not ignore this case.

> As for older options being removed, I do not think that happens -- they are typically marked as deprecated and the code just ignores them. If they were removed from the system, why should an "old option that no longer exists" be treated any differently than a "new option that does not yet exist"?
> Not removing options SGTM. Just curious and I am not looking for a definite solution, should we ever worry about too many options? If the number of deprecated options increase, it may one day become difficult for users and developers to browse them.

mrambacher: > > The old behavior would silently change the flag as well -- just from true to false.
> > 
> > * In the original code, ignore would change from true to false if the version was OLD
> 
> I think this is fine.
> 
> > * In the current code, ignore would change from false to true if the version was NEW
> 
> I think this is problematic. By setting the flag to false, the application indicates that errors should be reported. By silently changing it to true, we break the contract in a more dangerous way because there is a risk of application not being notified of an potential issue.
> 
> > * In the proposed code, ignore will NOT change
> 
> This feels borderline, but it is a change to API behavior which relaxes some checks. If the user sets `ignore_unknown_options` to `true`, and RocksDB detects some issue which should not happen, should RocksDB notify the user even if the user wants to ignore? IMO, it is better to still report the error. If a bad OPTIONS file is loaded, it may lead to unexpected consequences. In this specific case of old OPTIONS file containing unknown option name (to a new release), it is highly likely that the OPTIONS file has been tampered with, and application should not ignore this case.

I do not think the proposed change relaxes checks.  In the OLD code, it relaxed the checks (going from true to false on older versions).  The proposed change keeps the checks as the user requested them. 

> > As for older options being removed, I do not think that happens -- they are typically marked as deprecated and the code just ignores them. If they were removed from the system, why should an "old option that no longer exists" be treated any differently than a "new option that does not yet exist"?
> 
> Not removing options SGTM. Just curious and I am not looking for a definite solution, should we ever worry about too many options? If the number of deprecated options increase, it may one day become difficult for users and developers to browse them.

There are two different points on the deprecated options.  They still live in the code as "deprecated" options that are read and ignored.  So if a deprecated option is in the options file, it is read and treated as valid.  However, when the code writes out an options file, deprecated options are not written back out.

riversand963: iiuc, `ignore_unknown_options` being true is less strict than false, since it being false will report error, as what I read in configurable.cc
```
  if (config_options.ignore_unknown_options) {
    s = Status::OK();
  } else if (s.ok() && unused == nullptr && !remaining.empty()) {
    s = Status::NotFound("Could not find option: ", remaining.begin()->first);
  }
```
 Therefore, the old code is more strict since the flag can be changed from true to false. In contrast, the proposed change will keep the flag as true if the application specifies it, thus it is less strict.

mrambacher: The original code prevents the user from ever making the determination that they wanted to ignore options in an old options file.  It ignores something they would have had to explicitly set and not take the default value.  If you wrote the same test as is failing (adding an unknown option) with an older version instead of a newer one, there is no combination of parameters that would make that test pass.  The proposed change makes it so that we respect the parameters passed in and do not assume "we know better".

riversand963: If options are never removed, then how can an OPTIONS file generated by an old RocksDB release contain option names unknown to a new RocksDB release?

> same test as is failing (adding an unknown option) with an older version
>
You mean a test case in which RocksDB tries to load an OPTIONS file generated by old release but encounters an unknown option (manually added)? According to the comment of the old code, the intention is to fail this anyway, so the test will assert on failure to load it.

mrambacher:
I reverted the code/tests to the old way, though I strongly feel that it is wrong.   It makes the code and meaning of values inconsistent and makes the code difficult to document.  For example, the ConfigOptions has an "ignore_unknown_options" value.  How do you document when it applies?  If you are calling methods like "GetDBOptionsFromMap" it seems like it applies but if you are calling "LoadLatestOptions" we may choose to ignore it.  The GetDBOptionsFromMap takes in an "ignore_unknown_options" parameter.  Does it always apply?  

One of the banes of good software is having exceptions to the rule.  If a value sometimes means X and sometimes means Y, the user and other developers are going to get confused and those exception becomes amplified over time.  It makes documentation hard and confusing and impossible to maintain.  Another bane is assuming that the developer knows more than the user and should override them because we know what they really want.  In this case, we are completely ignoring the explicit instruction of the user for no apparent gain.

ajkr:
Interesting discussion on the existing behavior of `ignore_unknown_options`. Considering how it worked and considering our process for deleting options, it seems we have an undocumented requirement that users do not skip over too many versions when upgrading. For example, imagine we decide today we intend to delete `pin_l0_filter_and_index_blocks_in_cache` in RocksDB 7.0, then:

- In 6.15 we can stop writing `pin_l0_filter_and_index_blocks_in_cache` to OPTIONS files
- In 7.0 we can delete `pin_l0_filter_and_index_blocks_in_cache` and stop allowing it to be read from OPTIONS files

If a user tried to upgrade directly from 6.14 to 7.0, their `LoadLatestOptions()` would fail no matter how they set `ignore_unknown_options` because we never permit unknown options from older versions. However if they did it in multiple steps (6.14 -> something in [6.15, 7.0) -> 7.0), it would succeed since opening the DB on the intermediate version would write an OPTIONS file without `ignore_unknown_options`.

riversand963:
Thanks @mrambacher for the fix.
> I reverted the code/tests to the old way, though I strongly feel that it is wrong.
>
Is there a PR?

riversand963:
The disagreement is mainly about old code vs removing it. We all agree that new behavior is problematic. Can we take a two-step approach? We can revert to old behavior to fix the tests and unblock dependencies, then next step can be discussed.

mrambacher:
PR #7612 

I see no reason to take the two step approach.  We will simply be rehashing the discussion here and I do not see where we make further progress.

mrambacher:
> Interesting discussion on the existing behavior of `ignore_unknown_options`. Considering how it worked and considering our process for deleting options, it seems we have an undocumented requirement that users do not skip over too many versions when upgrading. For example, imagine we decide today we intend to delete `pin_l0_filter_and_index_blocks_in_cache` in RocksDB 7.0, then:
> 
> * In 6.15 we can stop writing `pin_l0_filter_and_index_blocks_in_cache` to OPTIONS files
> * In 7.0 we can delete `pin_l0_filter_and_index_blocks_in_cache` and stop allowing it to be read from OPTIONS files
> 
> If a user tried to upgrade directly from 6.14 to 7.0, their `LoadLatestOptions()` would fail no matter how they set `ignore_unknown_options` because we never permit unknown options from older versions. However if they did it in multiple steps (6.14 -> something in [6.15, 7.0) -> 7.0), it would succeed since opening the DB on the intermediate version would write an OPTIONS file without `pin_l0_filter_and_index_blocks_in_cache`.

We actually have an undocumented requirement that we can never delete a deprecated option from the code.  Once an option is created, it must be preserved in the list of potential options forever.  

If we changed it that a user could ignore unknown options regardless of the version of the Options file (as I proposed but appears to have been rejected), we could hit the problem you mentioned when a version is skipped where options are officially removed.

ajkr:
> We actually have an undocumented requirement that we can never delete a deprecated option from the code. Once an option is created, it must be preserved in the list of potential options forever.
>
> If we changed it that a user could ignore unknown options regardless of the version of the Options file (as I proposed but appears to have been rejected), we could hit the problem you mentioned when a version is skipped where options are officially removed.

I tested it out just now and did not observe the mentioned problem. Upgrading through an intermediate version with `kDontSerialize` on `pin_l0_filter_and_index_blocks_in_cache` allowed the OPTIONS file to be loaded successfully after it was deleted: https://gist.github.com/ajkr/44df1c69e78629e91e7cf652661d8dc0. Meanwhile, upgrading directly to the version with `pin_l0_filter_and_index_blocks_in_cache` deleted caused `LoadLatestOptions()` to fail: https://gist.github.com/ajkr/d2037c644b753f554fdf0feb6428b11f. Also note the latter failure returns `NotFound`, while I thought it's supposed to return `InvalidArgument`.

mrambacher:
> > We actually have an undocumented requirement that we can never delete a deprecated option from the code. Once an option is created, it must be preserved in the list of potential options forever.
> > If we changed it that a user could ignore unknown options regardless of the version of the Options file (as I proposed but appears to have been rejected), we could hit the problem you mentioned when a version is skipped where options are officially removed.
> 
> I tested it out just now and did not observe the mentioned problem. Upgrading through an intermediate version with `kDontSerialize` on `pin_l0_filter_and_index_blocks_in_cache` allowed the OPTIONS file to be loaded successfully after it was deleted: https://gist.github.com/ajkr/44df1c69e78629e91e7cf652661d8dc0. Meanwhile, upgrading directly to the version with `pin_l0_filter_and_index_blocks_in_cache` deleted caused `LoadLatestOptions()` to fail: https://gist.github.com/ajkr/d2037c644b753f554fdf0feb6428b11f. Also note the latter failure returns `NotFound`, while I thought it's supposed to return `InvalidArgument`.

Because when you marked it as "dontSerialize", the option is not written to the file.  So if you look at the options file after the first step, you will not see your option in it any longer.  So in your case, the option is valid in 6.14, not written to 6.15, and not processed at all in 7.0.  If you upgraded directly from a 6.14 to 7.0, you would see this problem.  

Which I think leads us back to you cannot ever upgrade without the intermediate step (6.14 -> 6.15 -> 7.0 is fine, 6.14 -> 7.0 is bad).  And there is no way to "force" the upgrade from 6.14 to 7.0 that I am aware of.

PR #7359, opened 2020-09-08
Fix some errors showing up in Travis builds

pdillinger:
Summary: Also enables a pull request to trigger all the Travis
configurations by writing FULL_CI in the commit message. (See what I did
there?)

First issue

    make: *** No rule to make target 'jl/util/crc32c_ppc_asm.o', needed by 'rocksdbjava'.  Stop.

Second issue

    tools/db_bench_tool.cc:5514:38: error: ‘gen_exp.rocksdb::Benchmark::GenerateTwoTermExpKeys::keyrange_size_’ may be used uninitialized in this function

Test Plan: CI

PR #7558, opened 2020-10-14
Fix RocksJava ppc64le build

adamretter:
Fix a build issue with RocksJava on ppc64le that was introduced in https://github.com/facebook/rocksdb/pull/6660

This was fixed in https://github.com/facebook/rocksdb/pull/7359 and this commit is just a partial back-port of the minimum necessary parts of that PR.


ajkr (approved):
LGTM.

ajkr:
It looks like we neither have tooling nor GitHub UI permission to merge to non-master branch. I manually pushed it here: https://github.com/facebook/rocksdb/commit/bfdb0a7651672427ee62ed9e489a8f7f4235e2fc.

PR #7563, opened 2020-10-16
Revert `Status`es returned from pre-`Configurable` options functions

mrambacher:
Further refinement of the earlier PR.  Now the Status is NotFound with a subcode of PathNotFound. Also the existing functions for options parsing/loading are reverted to return InvalidArgument no matter in which way the user-provided arguments are deemed invalid.


ajkr (approved):
LGTM!

PR #7566, opened 2020-10-17
Fix many tests to run with MEM_ENV and ENCRYPTED_ENV; Introduce a MemoryFileSystem class

mrambacher:
This PR does a few things:

1.  The MockFileSystem class was split out from the MockEnv.  This change would theoretically allow a MockFileSystem to be used by other Environments as well (if we created a means of constructing one).  The MockFileSystem implements a FileSystem in its entirety and does not rely on any Wrapper implementation.

2.  Make the RocksDB test suite work when MOCK_ENV=1 and ENCRYPTED_ENV=1 are set.  To accomplish this, a few things were needed:
- The tests that tried to use the "wrong" environment (Env::Default() instead of env_) were updated
- The MockFileSystem was changed to support the features it was missing or mishandled (such as recursively deleting files in a directory or supporting renaming of a directory).

3.  Updated the test framework to have a ROCKSDB_GTEST_SKIP macro.  This can be used to flag tests that are skipped.  Currently, this defaults to doing nothing (marks the test as SUCCESS) but will mark the tests as SKIPPED when RocksDB is upgraded to a version of gtest that supports this (gtest-1.10).

I have run a full "make check" with MEM_ENV, ENCRYPTED_ENV,  both, and neither under both MacOS and RedHat.  A few tests were disabled/skipped for the MEM/ENCRYPTED cases.  The error_handler_fs_test fails/hangs for MEM_ENV (presumably a timing problem) and I will introduce another PR/issue to track that problem.  (I will also push a change to disable those tests soon).  There is one more test in DBTest2 that also fails which I need to investigate or skip before this PR is merged.

Theoretically, this PR should also allow the test suite to run against an Env loaded from the registry, though I do not have one to try it with currently.  

Finally, once this is accepted, it would be nice if there was a CircleCI job to run these tests on a checkin so this effort does not become stale.  I do not know how to do that, so if someone could write that job, it would be appreciated :)


db/db_impl/db_secondary_test.cc
```diff
@@ -469,6 +469,7 @@ TEST_F(DBSecondaryTest, MissingTableFile) {
         if (s.IsPathNotFound()) {
           ++table_files_not_exist;
         } else if (!s.ok()) {
+          printf("MJR: Status=%s\b", s.ToString().c_str());
```
jay-zhuang, line 472: what does MJR stand for?

test_util/testutil.cc
```diff
@@ -476,46 +476,62 @@ size_t GetLinesCount(const std::string& fname, const std::string& pattern) {
...
+Status CorruptFile(Env* env, const std::string& fname, int offset,
+                   int bytes_to_corrupt, bool verify_checksum /*=true*/) {
+  uint64_t size;
+  Status s = env->GetFileSize(fname, &size);
+  if (!s.ok()) {
+    return s;
+  } else if (offset < 0) {
     // Relative to end of file; make it absolute
-    if (-offset > sbuf.st_size) {
+    if (-offset > (int)size) {
```
jay-zhuang, line 487: nit: using `static_cast<int>(size)`

test_util/testutil.cc
```diff
@@ -476,46 +476,62 @@ size_t GetLinesCount(const std::string& fname, const std::string& pattern) {
...
     }
   }
-  if (offset > sbuf.st_size) {
-    offset = static_cast<int>(sbuf.st_size);
+  if (offset > (int)size) {
+    offset = static_cast<int>(size);
   }
-  if (offset + bytes_to_corrupt > sbuf.st_size) {
-    bytes_to_corrupt = static_cast<int>(sbuf.st_size - offset);
+  if (offset + bytes_to_corrupt > (int)size) {
```
jay-zhuang, line 496: nit: using `static_cast<int>(size)`?

jay-zhuang (approved):
LGTM. Added circleci tests.

db/db_impl/db_secondary_test.cc
mrambacher: Sorry.  my initials.  Leftover debugging.  I will remove it.

jay-zhuang:
Seems mem-env test hang: https://app.circleci.com/pipelines/github/facebook/rocksdb/4268/workflows/8081e2f3-8fd5-4ed0-b5a4-20b436a5d065/jobs/49303

mrambacher:
Pushed a new commit that should address the hangs in error_handler_fs_test

jay-zhuang:
@mrambacher would you please rebase the code? Thanks.

mrambacher:
> @mrambacher would you please rebase the code? Thanks.

Done!

PR #7570, opened 2020-10-19
Write min_log_number_to_keep to MANIFEST during atomic flush under 2 phase commit

cheng-chang:
When 2 phase commit is enabled, if there are prepared data in a WAL, the WAL should be kept, the minimum log number for such a WAL is written to MANIFEST during flush. In atomic flush, such information is not written to MANIFEST.

Test Plan:
Added a new unit test `DBAtomicFlushTest.ManualFlushUnder2PC`, this test fails in atomic flush without this PR, after this PR, it succeeds.


db/memtable_list.cc
```diff
@@ -726,6 +727,17 @@ Status InstallMemtableAtomicFlushResults(
     assert(0 == num_entries);
   }
 
+  if (vset->db_options()->allow_2pc) {
+    uint64_t min_log_number_to_keep = port::kMaxUint64;
```
riversand963, line 731: Nit: `port::kMaxUint64`.

db/memtable_list.cc
```diff
@@ -726,6 +727,17 @@ Status InstallMemtableAtomicFlushResults(
     assert(0 == num_entries);
   }
 
+  if (vset->db_options()->allow_2pc) {
```
riversand963, line 730: Should the non-2pc case be handled too?

riversand963 (commented):
Thanks @cheng-chang for the fix. Left a few comments.

db/memtable_list.cc
cheng-chang: The algorithm for computing the min_log_number has problems, the non-2pc case is fixed in https://github.com/facebook/rocksdb/pull/7660. I'll update this PR after https://github.com/facebook/rocksdb/pull/7660 is merged.

db/memtable_list.cc
cheng-chang: hmm, already used it

cheng-chang:
@riversand963 PTAL, thanks.

HISTORY.md
```diff
@@ -21,6 +21,7 @@
 * Fixed a potential bug caused by evaluating `TableBuilder::NeedCompact()` before `TableBuilder::Finish()` in compaction job. For example, the `NeedCompact()` method of `CompactOnDeletionCollector` returned by built-in `CompactOnDeletionCollectorFactory` requires `BlockBasedTable::Finish()` to return the correct result. The bug can cause a compaction-generated file not to be marked for future compaction based on deletion ratio.
 * Fixed a seek issue with prefix extractor and timestamp.
 * Fixed a bug of encoding and parsing BlockBasedTableOptions::read_amp_bytes_per_bit as a 64-bit integer.
+* Fixed a bug of a recovery corner case, details in PR7621.
```
riversand963, line 24: Is this PR related to this item?

cheng-chang: committed by mistake, removed

db/memtable_list.cc
```diff
@@ -752,14 +746,21 @@ Status InstallMemtableAtomicFlushResults(
...
-        PrecomputeMinLogNumberToKeepNon2PC(vset, cfds, edit_lists);
-    const auto& wals = vset->GetWalSet().GetWals();
-    if (!wals.empty() && min_wal_number_to_keep > wals.begin()->first) {
+  if (vset->db_options()->track_and_verify_wals_in_manifest &&
+      !vset->GetWalSet().GetWals().empty()) {
+    if (!vset->db_options()->allow_2pc) {
+      min_wal_number_to_keep =
+          PrecomputeMinLogNumberToKeepNon2PC(vset, cfds, edit_lists);
+    }
+    if (min_wal_number_to_keep > vset->GetWalSet().GetWals().begin()->first) {
```
riversand963, line 763: Can we do something like:
```
WalNumber min_wal_number_to_keep = allow_2pc ? PrecomputeMinLogNumberToKeep2PC() : PrecomputeMinLogNumberToKeepNon2PC()?
edit.SetMinLogNumberToKeep(min_wal_ number_to_keep);
```
I guess I am not sure I miss anything when I see `min_wal_number_to_keep` is evaluated in allow_2pc mode but only conditionally in non-2pc mode. In either case, it would be OK for the db to track min_wal_number_to_keep, right?

db/memtable_list.cc
```diff
@@ -674,20 +674,13 @@ void MemTableList::RemoveMemTablesOrRestoreFlags(
...
     // if the performance has some problem.
-    bool should_skip = false;
-    for (MemTable* m_to_flush : memtables_to_flush) {
-      if (m == m_to_flush) {
-        should_skip = true;
-        break;
-      }
-    }
-    if (should_skip) {
+    if (memtables_to_flush.count(m)) {
```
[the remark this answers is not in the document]
cheng-chang: removed

db/memtable_list.cc
cheng-chang: I agree that it's a bit confusing. But according to https://github.com/facebook/rocksdb/blob/master/db/version_set.cc#L4254, the min_log_number_to_keep in VersionEdit now should only be set in 2pc mode.

db/column_family.cc
```diff
@@ -708,9 +708,8 @@ uint64_t ColumnFamilyData::OldestLogToKeep() {
   auto current_log = GetLogNumber();
 
   if (allow_2pc_) {
-    autovector<MemTable*> empty_list;
-    auto imm_prep_log =
-        imm()->PrecomputeMinLogContainingPrepSection(empty_list);
+    std::unordered_set<MemTable*> empty_set;
```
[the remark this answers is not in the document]
cheng-chang: We can use std::optional, but that requires c++17. I'll change the parameter to be a pointer, then passing nullptr to indicate that it's optional.

cheng-chang:
@riversand963 thanks for comments, PTAL.

db/column_family.cc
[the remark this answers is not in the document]
riversand963: If you go with pointer type, then maybe add a comment to explain what `nullptr` means here.
I am also OK with keeping it `const std::unordered_set&`.

riversand963 (approved):

db/db_impl/db_impl_files.cc
```diff
@@ -681,6 +683,37 @@ uint64_t FindMinPrepLogReferencedByMemTable(
   return min_log;
 }
 
+uint64_t FindMinPrepLogReferencedByMemTable(
+    VersionSet* vset, const autovector<ColumnFamilyData*>& cfds_to_flush,
+    const autovector<const autovector<MemTable*>*>& memtables_to_flush) {
+  uint64_t min_log = 0;
```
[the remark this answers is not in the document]
cheng-chang: the current logic assumes the return value is 0 if no log contains prep log (since no log will use 0 as log number).

PR #7572, opened 2020-10-19
Fix unchecked statuses for transaction_test

cheng-chang:
When `ASSERT_STATUS_CHECKED` is enabled, `transaction_test` does not pass without this PR.

Test Plan:

`ASSERT_STATUS_CHECKED=1 make   -j32 transaction_test && ./transaction_test`


db/db_iter.h
```diff
@@ -124,6 +124,7 @@ class DBIter final : public Iterator {
   void operator=(const DBIter&) = delete;
 
   ~DBIter() override {
+    status_.PermitUncheckedError();
```
zhichao-cao, line 127: We prefer not to call PermitUncheckedError() in the destructor. Is it possible to force check it when using the DBIter? Or changing the testing code to check it?

cheng-chang:
@zhichao-cao thanks for review, PTAL.

test_util/transaction_test_util.cc
```diff
@@ -350,6 +349,7 @@ Status RandomTransactionInserter::Verify(DB* db, uint16_t num_sets,
             static_cast<int>(key.size()), key.data(), int_value);
         total += int_value;
       }
+      iter->status().PermitUncheckedError();
```
mrambacher, line 352: Though I agree it is not likely the status is not OK here, you should probably extract the status and return it if it is not OK, rather than just ignoring it.

test_util/transaction_test_util.cc
```diff
@@ -39,7 +39,6 @@ RandomTransactionInserter::RandomTransactionInserter(
       cmt_delay_ms_(cmt_delay_ms) {}
 
 RandomTransactionInserter::~RandomTransactionInserter() {
-  last_status_.PermitUncheckedError();
```
mrambacher: Since this is a test utility, is there a reason not to uncheck the status in the destructor rather than in other places in the code?

test_util/transaction_test_util.cc
cheng-chang: the status must be ok, this is just to work around the status check.

test_util/transaction_test_util.cc
cheng-chang: swallowing the error in destructor is generally bad practice I think.

zhichao-cao (approved):
LGTM, thanks for the changes.

PR #7573, opened 2020-10-19
Integrate BlobFileBuilder into the compaction process

ltamasi:
Summary:
Similarly to how #7345
integrated blob file writing into the flush process,
the patch adds support for writing blob files to the compaction logic.
Namely, if `enable_blob_files` is set, large values encountered during
compaction are extracted to blob files and replaced with blob indexes.
The resulting blob files are then logged to the MANIFEST as part of the
compaction job's `VersionEdit` and added to the `Version` alongside any
table files written by the compaction. Any errors during blob file building fail
the compaction job.

There will be a separate follow-up patch to perform blob garbage collection
during compactions.

In addition, the patch continues to chip away at the mess around computing
various compaction related statistics by eliminating some code duplication
and by making the `num_output_files` and `bytes_written` stats more consistent
for flushes, compactions, and recovery.

Test Plan:
`make check`


db/builder.cc
```diff
@@ -330,8 +335,6 @@ Status BuildTable(
         ignored = fs->DeleteFile(blob_file_path, IOOptions(), dbg);
         ignored.PermitUncheckedError();
       }
-
-      blob_file_additions->clear();
```
riversand963, line 334: It's possible that no sst is generated, but blob files are written and recorded in the MANIFEST. How will these blobs be accessed?

riversand963: Or is this just for the accuracy of stats, e.g. blobs written, blob_bytes, etc?

db/db_compaction_test.cc
```diff
@@ -5845,6 +5845,172 @@ TEST_F(DBCompactionTest, ChangeLevelErrorPathTest) {
   ASSERT_EQ("0,5", FilesPerLevel(0));
 }
 
+TEST_F(DBCompactionTest, CompactionWithBlob) {
+  Options options;
```
riversand963, line 5849: Set `options.env` to `env_`.

db/db_compaction_test.cc
```diff
@@ -5845,6 +5845,172 @@ TEST_F(DBCompactionTest, ChangeLevelErrorPathTest) {
...
+  ASSERT_EQ(blob_file->GetTotalBlobCount(), 2);
+
+  const InternalStats* const internal_stats = cfd->internal_stats();
+  assert(internal_stats);
+
+  const uint64_t expected_bytes =
+      table_file->fd.GetFileSize() + blob_file->GetTotalBlobBytes();
+
+  const auto& compaction_stats = internal_stats->TEST_GetCompactionStats();
+  ASSERT_TRUE(compaction_stats.size() >= 2);
```
riversand963, line 5922: Nit: `ASSERT_LE(2, compaction_stats.size())`

db/db_compaction_test.cc
```diff
@@ -5845,6 +5845,172 @@ TEST_F(DBCompactionTest, ChangeLevelErrorPathTest) {
...
+  std::string sync_point_;
+};
+
+INSTANTIATE_TEST_CASE_P(DBCompactionTestBlobError, DBCompactionTestBlobError,
+                        ::testing::ValuesIn(std::vector<std::string>{
+                            "BlobFileBuilder::WriteBlobToFile:AddRecord",
+                            "BlobFileBuilder::WriteBlobToFile:AppendFooter"}));
+
+TEST_P(DBCompactionTestBlobError, CompactionError) {
+  Options options;
```
riversand963, line 5945: Same, `options.env = env_`;

db/db_compaction_test.cc
```diff
@@ -5845,6 +5845,172 @@ TEST_F(DBCompactionTest, ChangeLevelErrorPathTest) {
...
+  ASSERT_TRUE(l1_files.empty());
+
+  const auto& blob_files = storage_info->GetBlobFiles();
+  ASSERT_TRUE(blob_files.empty());
+
+  const InternalStats* const internal_stats = cfd->internal_stats();
+  assert(internal_stats);
+
+  const auto& compaction_stats = internal_stats->TEST_GetCompactionStats();
+  ASSERT_TRUE(compaction_stats.size() >= 2);
```
riversand963, line 6009: `ASSERT_LE` can be used.

riversand963 (commented):
Thanks @ltamasi for the PR.

db/builder.cc
```diff
@@ -212,6 +212,15 @@ Status BuildTable(
     } else if (!c_iter.status().ok()) {
       s = c_iter.status();
     }
+
+    if (blob_file_builder) {
```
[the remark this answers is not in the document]
ltamasi: Sure. I can change that back.

db/db_impl/db_impl_open.cc
```diff
@@ -1368,7 +1368,6 @@ Status DBImpl::WriteLevel0TableForRecovery(int job_id, ColumnFamilyData* cfd,
   // Note that if file_size is zero, the file has been deleted and
   // should not be added to the manifest.
   const bool has_output = meta.fd.GetFileSize() > 0;
-  assert(has_output || blob_file_additions.empty());
```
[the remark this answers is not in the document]
ltamasi: Yes: it is possible for the SST building to fail (and `has_output` to be false) after some blob files have already been written successfully. Note that the `VersionEdit` will not be applied for such failed jobs, so these blob files will not be logged in the MANIFEST. My goal here is simply to be consistent with compactions, which do count successfully written files even if the job eventually fails.

db/builder.cc
ltamasi: > Or is this just for the accuracy of stats, e.g. blobs written, blob_bytes, etc?

Exactly, please see above.

riversand963 (approved):

PR #7574, opened 2020-10-19
Fix for stress test failure

akankshamahajan15:
Summary:  Ignore read error in 'FilePrefetchBuffer::TryReadFromCache' as status is ignored
and  bool value is returned. Return error if prefetch fails in
'PrefetchTail' as we have planned to return Prefetch failures to users.

Test Plan: make check -j64,
python -u tools/db_crashtest.py blackbox

Reviewers:

Subscribers:

Tasks:

Tags:


zhichao-cao (approved):
LGTM, thanks!

PR #7575, opened 2020-10-20
Fix write_batch_test when ASSERT_STATUS_CHECKED=1

cheng-chang:
Without this PR, `ASSERT_STATUS_CHECKED=1 make   -j32 write_batch_test && ./write_batch_test` fails.

Test Plan:
ASSERT_STATUS_CHECKED=1 make   -j32 write_batch_test && ./write_batch_test


mrambacher (approved):
This generally LGTM.  I must admit to not being a fan of the "assert(s.ok())", as it seems like it could come back to bite us, but I cannot see where it is much different than ignoring the errors and do not see much of an alternative.

zhichao-cao (approved):
LGTM!

PR #7577, opened 2020-10-20
Update "ReadOptions.value_size_soft_limit" to also include keys size.

akankshamahajan15:
Summary: Update "ReadOptions.value_size_soft_limit" to include keys size
(keys found) along with size of their values.
Initally this option considered only the size of values.

Test Plan: make check -j64,
           Tests related to value_size_soft_limit

Reviewers:

Subscribers:

Tasks: T77798449

Tags:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -9,6 +9,7 @@
 
 ### Public API Change
 * Deprecate `BlockBasedTableOptions::pin_l0_filter_and_index_blocks_in_cache` and `BlockBasedTableOptions::pin_top_level_index_and_filter`. These options still take effect until users migrate to the replacement APIs in `BlockBasedTableOptions::metadata_cache_options`. Migration guidance can be found in the API comments on the deprecated options.
+* Update "ReadOptions.value_size_soft_limit" to include keys size (keys that are found) along with their value size in the cumulative size (which initially included size of values only). The MultiGet is Aborted once the total cumulative size exceeds "value_size_soft_limit".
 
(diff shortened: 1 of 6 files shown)


pdillinger (commented):
Except we called it "value_size_soft_limit." I don't think there's an easy answer to supporting a limit on key+value size, since we have to either
(a) change behavior and have a misleading name (this PR draft)
(b) change the name & behavior (break recent API addition)
(c) add a new option for combined key+value size limit

If (c) is kinda hard, I probably prefer (b) to (a) since this is a recent API addition.

Some thinking out loud:

I also have concerns about feature creep here. Suppose that (I haven't confirmed any of this, just speculating) service owner X realizes that their service keys are really just a substring of their RocksDB keys, so RocksDB sometimes returns less data than they want because RocksDB keys are longer than their corresponding service keys. Maybe the fact that we're using a soft limit can make up for that, though, as there's a chance you can include all the returned KVs, beyond RocksDB soft limit setting == service hard limit setting, (instead of dropping one KV to satisfy service hard limit) if it happens to be under the service hard limit thanks to service keys being smaller than corresponding RocksDB keys.

So maybe soft limit leaves open enough customizability on the user end to get what they want even in scenarios like that, though we might want to provide choice of value size soft limit or key+value size soft limit.

pdillinger:
I just saw internal discussion where this change request *might* not be needed. Let's see. :)

akankshamahajan15:
No need of these changes. status updated internally.

PR #7578, opened 2020-10-20
Add API to verify whole sst file checksum

riversand963:
Existing API `VerifyChecksum()` allows application to verify sst files' block checksums.
Since whole file, user-specified checksum is tracked in MANIFEST, we can expose a new
API to verify sst files' file checksums.

```
// Compute table file checksums if applicable and compare with MANIFEST.
// Returns OK if no file has mismatching whole-file checksum.
Status DB::VerifyFileChecksums(const ReadOptions& /*read_options*/);
```

Test plan:
make check


db/corruption_test.cc
```diff
@@ -824,6 +831,43 @@ TEST_F(CorruptionTest, DisableKeyOrderCheck) {
   ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->ClearAllCallBacks();
 }
 
+TEST_F(CorruptionTest, VerifyWholeTableChecksum) {
+  delete db_;
+  db_ = nullptr;
+  Options options;
```
mrambacher, line 837: Please either use Options options = GetDefaultOptions() or be sure to set env_.  Otherwise, this test will fail for the non-default Env cases

db/db_impl/db_impl.cc
```diff
@@ -4727,7 +4728,30 @@ Status DBImpl::CreateColumnFamilyWithImport(
...
+    if (!file_checksum_gen_factory) {
+      s = Status::InvalidArgument(
+          "use_file_checksum cannot be true if file checksum is disabled");
+      return s;
+    } else {
+      FileChecksumGenContext file_checksum_context;
+      std::unique_ptr<FileChecksumGenerator> file_checksum_gen =
+          file_checksum_gen_factory->CreateFileChecksumGenerator(
+              file_checksum_context);
+      file_checksum_func_name = file_checksum_gen->Name();
```
mrambacher, line 4751: This seems like an awful expensive operation to get the name of the checksum func.  Is there a cheaper alternative?

include/rocksdb/convenience.h
```diff
@@ -494,6 +494,10 @@ Status VerifySstFileChecksum(const Options& options,
                              const ReadOptions& read_options,
                              const std::string& file_path);
 
+// Verify the checksums of table files in db
```
mrambacher, line 497: This needs more information as to what the arguments mean.

db/corruption_test.cc
riversand963: Good catch!

db/db_impl/db_impl.cc
riversand963: I checked, and didn't find one. From `FileChecksumGenFactory`, the name is different from the name returned from `FileChecksumGenerator`. Let me know if I miss anything @zhichao-cao .

zhichao-cao: Currently, no. User provides the shared_ptr of FileChecksumGenFactory and it brings the factory name. The checksum function name is brought by the generator itself. We can add the get checksum function name to the Factory, but it will break the flexibility. It is possible that there are more than one generator in a factory and only the one match the file_checksum_context will be returned.

riversand963: Is it possible that the application specified a different name for the generator in `FileChecksumContext` even if the same underlying algorithm is used?

zhichao-cao: It is possible. We do not have the role or contract for the naming of the generator. E.g., we can call it "crc32c" or "CRC32C".

riversand963: Thanks @zhichao-cao for the info. Then we need to handle this case, too. If the name can be anything, it is a little hard.

zhichao-cao: Just double check your question: user implement the role to get the FileChecksumGenerator based on the name passed in FileChecksumContext.

riversand963: hmm, I wonder how much effort we should put here to detect the equality of checksum function name. Should we just require that they be the same (case sensitive) so that file checksum verification can be done?

zhichao-cao: Actually, if we consider a production case, application should always use the same FileChecksumGenFactory. So the name should be exactly the same. Only in the case that the DB is opened by a different application, then, there might be the possibility that same algorithm is used with different name.

db/db_impl/db_impl.cc
```diff
@@ -4726,8 +4727,35 @@ Status DBImpl::CreateColumnFamilyWithImport(
...
+                              bool use_file_checksum) {
   Status s;
+
+  std::string file_checksum_func_name(kUnknownFileChecksumFuncName);
+  if (use_file_checksum) {
+    FileChecksumGenFactory* const file_checksum_gen_factory =
+        immutable_db_options_.file_checksum_gen_factory.get();
+    if (!file_checksum_gen_factory) {
+      s = Status::InvalidArgument(
+          "use_file_checksum cannot be true if file checksum is disabled");
```
pdillinger, line 4748: use_file_checksum is not part of the public API, so doesn't make sense in a Status result passed back to the user.

db/db_impl/db_impl.cc
```diff
@@ -4789,6 +4826,44 @@ Status DBImpl::VerifyChecksum(const ReadOptions& read_options) {
...
 
+Status DBImpl::VerifySstFileChecksum(const FileMetaData& fmeta,
+                                     const std::string& fname,
+                                     const std::string& file_checksum_func_name,
+                                     const ReadOptions& read_options) {
+  Status s;
+  if (fmeta.file_checksum == kUnknownFileChecksum) {
+    return s;
+  }
+  if (fmeta.file_checksum_func_name != file_checksum_func_name) {
```
pdillinger, line 4837: This is the wrong approach. A checksum generator factory can support different functions, though it does have a default function. To support migration, it should be OK for the SST files to have a mix of checksum functions, as long as they're all supported by the configured factory. You have to request the function used in the SST file from the generator factory, then verify that the function returned matches what you requested. If that doesn't match, then you have an error.

pdillinger (changes requested):

pdillinger:
Also update PR description

db/db_impl/db_impl.cc
riversand963: Sorry, this comes from a previous version. Will update.

db/db_impl/db_impl.cc
zhichao-cao: Looks like the file_checksum_context does not carry the name, in this case, you will get the default generator. So the file_checksum_func_name is the default generator's name. As mentioned before, in most cases, there is only one generator type in a factory. But it is possible, there are more than one generator in a factory. If not the default one is used in the file, we will have wrong verification results.

db/db_impl/db_impl.cc
```diff
@@ -4789,6 +4826,44 @@ Status DBImpl::VerifyChecksum(const ReadOptions& read_options) {
...
+        << " checksum func name does not "
+           "match ("
+        << fmeta.file_checksum_func_name << "vs. " << file_checksum_func_name
+        << ")";
+    s = Status::InvalidArgument(oss.str());
+    return s;
+  }
+  std::string file_checksum;
+  std::string func_name;
+  s = ROCKSDB_NAMESPACE::GenerateOneFileChecksum(
```
zhichao-cao, line 4849: I will quickly create a PR to fix `GenerateOneFileChecksum`.  Zitan introduced the `requested_checksum_func_name` in `file_checksum_contex`, but this function is not updated accordingly......

riversand963: @zhichao-cao, so after your fix, `GenerateOneFileChecksum()` will take `func_name` as input.
It will be used to set `file_checksum_contex.requested_checksum_func_name` which will be used to request a function from factory. If no function is returned or the returned func has a different name, then there is an error. Is this right?

zhichao-cao: In GenerateOneFileChecksum, we pass in the `requested_checksum_func_name` and it will get the checksum together with the `generator_checksum_func_name`. GenerateOneFileChecksum does not compare and verify if `requested_checksum_func_name == generator_checksum_func_name`. I think caller will do the verification if they want, and decide to use the checksum or not.

db/db_impl/db_impl.cc
riversand963: I think the solution will be to leverage your change in #7586.

db/db_impl/db_impl.cc
riversand963: Updated.

include/rocksdb/db.h
```diff
@@ -1443,6 +1443,11 @@ class DB {
       const ExportImportFilesMetaData& metadata,
       ColumnFamilyHandle** handle) = 0;
 
+  // Verify the checksums of files in db. Currently only tables are checked.
```
pdillinger, line 1446: It's not clear how this is different from VerifyChecksum. Please make that clear.

db/db_impl/db_impl.cc
```diff
@@ -4743,23 +4765,31 @@ Status DBImpl::VerifyChecksum(const ReadOptions& read_options) {
...
+        const FileMetaData* fmeta = fd_with_krange.file_metadata;
+        assert(fmeta);
         std::string fname = TableFileName(cfd->ioptions()->cf_paths,
                                           fd.GetNumber(), fd.GetPathId());
-        s = ROCKSDB_NAMESPACE::VerifySstFileChecksum(opts, file_options_,
-                                                     read_options, fname);
+        if (use_file_checksum) {
+          s = VerifySstFileChecksum(*fmeta, fname, read_options);
+        } else {
+          s = ROCKSDB_NAMESPACE::VerifySstFileChecksum(opts, file_options_,
```
pdillinger, line 4790: We need to rename this function to reduce confusion. Maybe `VerifySstInternalChecksums`?

db/db_impl/db_impl.cc
```diff
@@ -4727,8 +4728,29 @@ Status DBImpl::CreateColumnFamilyWithImport(
...
 
+Status DBImpl::VerifyFileChecksums(const ReadOptions& read_options) {
+  return VerifyChecksum(read_options, /*use_file_checksum=*/true);
+}
+
 Status DBImpl::VerifyChecksum(const ReadOptions& read_options) {
+  return VerifyChecksum(read_options, /*use_file_checksum=*/false);
+}
+
+Status DBImpl::VerifyChecksum(const ReadOptions& read_options,
```
pdillinger, line 4739: Not a fan of overloading in derived class a function that is in base class, but not that overload.

pdillinger (changes requested):
The naming is so confusing. What can we do to make that better rather than worse?

db/db_impl/db_impl.cc
riversand963: How about `VerifySstFileBlockChecksums()`?

riversand963: Actually, `rocksdb::VerifySstFileChecksum()` is a public API. Maybe rename it in a separate PR? I plan to backport this PR to 6.14.

db/db_impl/db_impl.cc
riversand963: Sure, will rename it to `VerifyChecksumInternal()`.

include/rocksdb/db.h
riversand963: Done

db/db_impl/db_impl.cc
pdillinger: Oh, OK. I must have been looking at convenience.cc when I saw db/

pdillinger (approved):
Much better. I don't have a great suggestion for improving the names, and the comments are pretty clear about what to expect.

PR #7580, opened 2020-10-21
Make FileType Public and Replace kLogFile with kWalFile

zhichao-cao:
As suggested by @pdillinger ,The name of kLogFile is misleading, in some tests, kLogFile is defined as info log. Replace it with kWalFile and move it to public, which will be used in #7523 

Test plan: make check

### DIFF
--- a/include/rocksdb/options.h
+++ b/include/rocksdb/options.h
@@ -24,6 +24,7 @@
 #include "rocksdb/file_checksum.h"
 #include "rocksdb/listener.h"
 #include "rocksdb/sst_partitioner.h"
+#include "rocksdb/types.h"
 #include "rocksdb/universal_compaction.h"
 #include "rocksdb/version.h"
 #include "rocksdb/write_buffer_manager.h"
--- a/db/column_family_test.cc
+++ b/db/column_family_test.cc
@@ -3271,7 +3271,7 @@ TEST_P(ColumnFamilyTest, DISABLED_LogTruncationTest) {
     FileType type;
     if (!(ParseFileName(filenames[i], &number, &type))) continue;
 
-    if (type != kLogFile) continue;
+    if (type != kWalFile) continue;
 
     logfs.push_back(filenames[i]);
   }
--- a/db/corruption_test.cc
+++ b/db/corruption_test.cc
@@ -249,8 +249,8 @@ TEST_F(CorruptionTest, Recovery) {
   // is not available for WAL though.
   CloseDb();
 #endif
-  Corrupt(kLogFile, 19, 1);      // WriteBatch tag for first record
-  Corrupt(kLogFile, log::kBlockSize + 1000, 1);  // Somewhere in second block
+  Corrupt(kWalFile, 19, 1);  // WriteBatch tag for first record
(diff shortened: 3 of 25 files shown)


include/rocksdb/options.h
```diff
@@ -57,6 +57,21 @@ class FileSystem;
 struct Options;
 struct DbPath;
 
+// To indicate the file types
+enum FileType {
```
mrambacher, line 61: Since this is not public, can you add a brief comment as to what all of these mean?  

Also, should there be a kUnknownFile for types that we do not know about?

include/rocksdb/options.h
```diff
@@ -57,6 +57,21 @@ class FileSystem;
...
+  kWalFile,
+  kDBLockFile,
+  kTableFile,
+  kDescriptorFile,
+  kCurrentFile,
+  kTempFile,
+  kInfoLogFile,  // Either the current one, or an old one
+  kMetaDatabase,
```
mrambacher, line 69: What is a "kMetaDatabase"?  Is this something that has not been developed or is it dead code?  Not sure if this belongs in the public API...

mrambacher (approved):
LGTM once comments are added.

include/rocksdb/options.h
zhichao-cao: Sure. Totally agree, some of the name does not directly indicate the file type. I will add the comments. Since I move it from file_name.h to options.h, I prefer not to add kUnknownFile at this moment. In the future, if we have the use case, we can add it easily.

include/rocksdb/options.h
```diff
@@ -57,6 +57,21 @@ class FileSystem;
 struct Options;
 struct DbPath;
 
+// To indicate the file types
```
pdillinger, line 60: Perhaps

    // The types of files RocksDB uses in a DB directory. (Available for advanced options.)

pdillinger (commented):
FileType is touching implementation details, so it is definitely of the "advanced options" flavor. Users should rarely need FileType. Even if checksum_handoff_file_types belongs in DBOptions, can we put this in some header included by options.h? Or we could even consider a forward declaration in options.h. If we get the enum definition out of options.h, I don't think it's so critical to get clear comments on all the enumerators. This enum is for advanced usage.

I think we generally prefer 'enum class' for new enums (or in this case, newly-public enums). What do you think?

include/rocksdb/options.h
mrambacher: Perhaps the enum should be defined in "advanced_options.h" instead of options.h?

I am not in favor of classes that are defined in the public API only having forward declarations.

riversand963: If we want to expose file type as public, how about considering the file "include/rocksdb/types.h"? According to the file's comment, it is for all public custom types.

riversand963 (approved):
LGTM with one minor comment.

include/rocksdb/options.h
zhichao-cao: Sure, I will move to "include/rocksdb/types.h", it makes more sense for me.

PR #7581, opened 2020-10-22
Bug fix to remove function calling in assert statement

akankshamahajan15:
Summary: Remove function calling in assert statement as assert is a no
op in opt build and that function might not be called. This causes hang
in closing RocksDB when refit level is set.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:


riversand963 (approved):
Please also update HISTORY.md. Thanks @akankshamahajan15 for the fix.

mrambacher (approved):
LGTM.

PR #7586, opened 2020-10-22
Updated GenerateOneFileChecksum to use requested_checksum_func_name

zhichao-cao:
CreateFileChecksumGenerator may uses requested_checksum_func_name in generator context to decide which generator will be used. GenerateOneFileChecksum has not being updated to use it, which will always get the generator when the name is empty. Fix it.

test plan: make check

### DIFF
--- a/db/external_sst_file_ingestion_job.cc
+++ b/db/external_sst_file_ingestion_job.cc
@@ -192,8 +192,9 @@ Status ExternalSstFileIngestionJob::Prepare(
     // Step 1: generate the checksum for ingested sst file.
     if (need_generate_file_checksum_) {
       for (size_t i = 0; i < files_to_ingest_.size(); i++) {
-        std::string generated_checksum, generated_checksum_func_name,
-            requested_checksum_func_name;
+        std::string generated_checksum;
+        std::string generated_checksum_func_name;
+        std::string requested_checksum_func_name;
         IOStatus io_s = GenerateOneFileChecksum(
             fs_.get(), files_to_ingest_[i].internal_file_path,
             db_options_.file_checksum_gen_factory.get(),
@@ -841,8 +842,9 @@ IOStatus ExternalSstFileIngestionJob::GenerateChecksumForIngestedFile(
     // file checksum generated during Prepare(). This step will be skipped.
     return IOStatus::OK();
   }
-  std::string file_checksum, file_checksum_func_name,
-      requested_checksum_func_name;
+  std::string file_checksum;
+  std::string file_checksum_func_name;
+  std::string requested_checksum_func_name;
   IOStatus io_s = GenerateOneFileChecksum(
       fs_.get(), file_to_ingest->internal_file_path,
       db_options_.file_checksum_gen_factory.get(), requested_checksum_func_name,
--- a/db/external_sst_file_ingestion_job.cc
+++ b/db/external_sst_file_ingestion_job.cc
@@ -192,10 +192,12 @@ Status ExternalSstFileIngestionJob::Prepare(
     // Step 1: generate the checksum for ingested sst file.
     if (need_generate_file_checksum_) {
       for (size_t i = 0; i < files_to_ingest_.size(); i++) {
-        std::string generated_checksum, generated_checksum_func_name;
+        std::string generated_checksum, generated_checksum_func_name,
+            requested_checksum_func_name;
         IOStatus io_s = GenerateOneFileChecksum(
             fs_.get(), files_to_ingest_[i].internal_file_path,
-            db_options_.file_checksum_gen_factory.get(), &generated_checksum,
+            db_options_.file_checksum_gen_factory.get(),
+            requested_checksum_func_name, &generated_checksum,
             &generated_checksum_func_name,
             ingestion_options_.verify_checksums_readahead_size,
             db_options_.allow_mmap_reads, io_tracer_);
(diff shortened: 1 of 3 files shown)


file/file_util.cc
```diff
@@ -123,22 +123,33 @@ bool IsWalDirSameAsDBPath(const ImmutableDBOptions* db_options) {
...
     return IOStatus::InvalidArgument("Checksum factory is invalid");
   }
   assert(file_checksum != nullptr);
   assert(file_checksum_func_name != nullptr);
 
   FileChecksumGenContext gen_context;
+  gen_context.requested_checksum_func_name = requested_checksum_func_name;
   std::unique_ptr<FileChecksumGenerator> checksum_generator =
       checksum_factory->CreateFileChecksumGenerator(gen_context);
+  if (checksum_generator == nullptr) {
```
riversand963, line 146: QQ: what does it mean if `checksum_generator != nullptr && checksum_generator->Name() != requested_checksum_func_name`?

riversand963 (commented):
Thanks @zhichao-cao for the fix.

file/file_util.cc
zhichao-cao: This should not happened, but yes, this behavior undefined. In regular case, the developer of factory should return the checksum_generator->Name() matches with requested_checksum_func_name.

riversand963: Maybe use assertion explicitly disallow this case in debug mode?

zhichao-cao: Sure

zhichao-cao: @riversand963 I think, due to the backward compatibility issue, we cannot do assert here. Before `requested_checksum_func_name` was introduced, we do not give the name to `CreateFileChecksumGenerator`. So checksum_generator->Name() will returns the name and requested_checksum_func_name is empty, it is expected.

db/external_sst_file_ingestion_job.cc
```diff
@@ -192,10 +192,12 @@ Status ExternalSstFileIngestionJob::Prepare(
     // Step 1: generate the checksum for ingested sst file.
     if (need_generate_file_checksum_) {
       for (size_t i = 0; i < files_to_ingest_.size(); i++) {
-        std::string generated_checksum, generated_checksum_func_name;
+        std::string generated_checksum, generated_checksum_func_name,
```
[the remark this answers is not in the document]
zhichao-cao: I think make format force to arrange it.......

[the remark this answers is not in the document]
riversand963: I mean
```
std::string generated_checksum;
std::string generated_checksum_func_name;
std::string requested_checksum_func_name;
```
formatter won't adjust this.

file/file_util.cc
riversand963: How about this?
```
assert(!checksum_generator || requested_checksum_func_name.empty() || requested_checksum_func_name == checksum_generator->Name());
```

file/file_util.cc
```diff
@@ -123,22 +123,33 @@ bool IsWalDirSameAsDBPath(const ImmutableDBOptions* db_options) {
...
+    std::string msg =
+        "Cannot get the file checksum generator based on the requested "
+        "checksum function name: " +
+        requested_checksum_func_name;
+    return IOStatus::InvalidArgument(msg);
```
riversand963, line 151: It will be useful to also include the name of the factory in `msg`.

riversand963 (approved):
LGTM with me. Will defer to @pdillinger in case he has any further concern.

file/file_util.cc
```diff
@@ -123,22 +123,41 @@ bool IsWalDirSameAsDBPath(const ImmutableDBOptions* db_options) {
...
+        requested_checksum_func_name +
+        " from checksum factory: " + checksum_factory->Name();
+    return IOStatus::InvalidArgument(msg);
+  }
+
+  // For backward compatable, requested_checksum_func_name can be empty.
+  // If we give the requested checksum function name, we expect it is the
+  // same name of the checksum generator.
+  assert(!checksum_generator || requested_checksum_func_name.empty() ||
+         requested_checksum_func_name == checksum_generator->Name());
```
pdillinger, line 159: I suppose it's possible that users could have name aliases for the same function, but this assertion seems quite reasonable to me in DEBUG builds for now.

file/file_util.cc
```diff
@@ -123,22 +123,41 @@ bool IsWalDirSameAsDBPath(const ImmutableDBOptions* db_options) {
...
+    std::string* file_checksum_func_name,
+    size_t verify_checksums_readahead_size, bool allow_mmap_reads,
+    std::shared_ptr<IOTracer>& io_tracer) {
   if (checksum_factory == nullptr) {
     return IOStatus::InvalidArgument("Checksum factory is invalid");
   }
   assert(file_checksum != nullptr);
   assert(file_checksum_func_name != nullptr);
 
   FileChecksumGenContext gen_context;
```
pdillinger, line 142: Shouldn't we populate the file_name field also?

pdillinger (approved):
LGTM except recommend populating file_name

file/file_util.cc
zhichao-cao: Yes, only in the case that, two different group of users they use different name.

file/file_util.cc
zhichao-cao: Good catch, will add it.

PR #7587, opened 2020-10-22
Add Encode/DecodeFixedGeneric, coding_lean.h

pdillinger:
Summary: To minimize dependencies for Ribbon filter code in progress,
core part of coding.h for fixed sizes has been moved to coding_lean.h.
Also, generic versions of these functions have been added to math128.h
(since the generic versions are likely only to be used along with
Unsigned128).

Test Plan: Unit tests added for new functions

### DIFF
--- a/util/math128.h
+++ b/util/math128.h
@@ -223,7 +223,7 @@ inline Unsigned128 DecodeFixed128(const char* ptr) {
 // A version of EncodeFixed* for generic algorithms. Likely to be used
 // with Unsigned128, so lives here for now.
 template <typename T>
-inline void EncodeFixedGeneric(char* dst, T value) {
+inline void EncodeFixedGeneric(char* /*dst*/, T /*value*/) {
   // Unfortunately, GCC does not appear to optimize this simple code down
   // to a trivial load on Intel:
   //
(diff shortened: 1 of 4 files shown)


util/math128.h
```diff
@@ -220,4 +220,66 @@ inline Unsigned128 DecodeFixed128(const char* ptr) {
   return (rv << 64) | DecodeFixed64(ptr);
 }
 
+// A version of EncodeFixed* for generic algorithms. Likely to be used
+// with Unsigned128, so lives here for now.
+template <typename T>
+inline void EncodeFixedGeneric(char* /*dst*/, T /*value*/) {
```
jay-zhuang, line 226: Should we just name it `EncodeFixed()`, `DecodeFixed()`?
`FastRangeGeneric()` makes me feel it should not be used directly.

jay-zhuang (approved):
LGTM

util/math128.h
pdillinger: In my opinion it should only be used for generic algorithms. Otherwise, we should be explicit about 32 vs. 64 etc. because we don't want to accidentally change some serialization format based on just changing the type of a variable. (Unless it's obviously going to do so, e.g. Thrift, but that's not normal in RocksDB.)

PR #7589, opened 2020-10-23
Fix MultiGet unable to query timestamp data issue

jay-zhuang:
The filter query key should not contain timestamp. The timestamp is
stripped for Get(), but not MultiGet().

### DIFF
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -492,6 +492,52 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
   Close();
 }
 
+TEST_F(DBBasicTestWithTimestamp, MultiGetWithFastLocalBloom) {
+  Options options = CurrentOptions();
+  options.env = env_;
+  options.create_if_missing = true;
+  BlockBasedTableOptions bbto;
+  bbto.filter_policy.reset(NewBloomFilterPolicy(10, false));
+  bbto.cache_index_and_filter_blocks = true;
+  bbto.whole_key_filtering = true;
+  options.table_factory.reset(NewBlockBasedTableFactory(bbto));
+  const size_t kTimestampSize = Timestamp(0, 0).size();
+  TestComparator test_cmp(kTimestampSize);
+  options.comparator = &test_cmp;
+  DestroyAndReopen(options);
+
+  // Write any value
+  WriteOptions write_opts;
+  std::string ts_str = Timestamp(1, 0);
+  Slice ts = ts_str;
+  write_opts.timestamp = &ts;
+
+  std::string keystr = "foo";
+  std::string vstr = "bar";
+  Slice k = keystr;
+  Slice v = vstr;
+  ASSERT_OK(db_->Put(write_opts, k, v));
+
+  // Flush to sstable
+  FlushOptions fo;
+  fo.wait = true;
+  ASSERT_OK(db_->Flush(fo));
+
+  // Read with MultiGet
+  ReadOptions read_opts;
+  read_opts.timestamp = &ts;
+  size_t batch_size = 1;
+  std::vector<Slice> keys(batch_size);
+  std::vector<PinnableSlice> values(batch_size);
+  std::vector<Status> statuses(batch_size);
+  keys[0] = k;
+  ColumnFamilyHandle* cfh = db_->DefaultColumnFamily();
+  db_->MultiGet(read_opts, cfh, batch_size, keys.data(), values.data(),
+                statuses.data());
+
+  ASSERT_OK(statuses[0]);
+}
+
 TEST_F(DBBasicTestWithTimestamp, MaxKeysSkipped) {
   Options options = CurrentOptions();
   options.env = env_;
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -705,6 +705,80 @@ TEST_F(DBBasicTestWithTimestamp, MultiGetVersion) {
   ASSERT_OK(statuses[0]);
 }
 
+TEST_F(DBBasicTestWithTimestamp, MultiGetCache) {
+  Options options = CurrentOptions();
+  options.env = env_;
+  options.create_if_missing = true;
+
+  options.prefix_extractor.reset(NewCappedPrefixTransform(5));
+  options.row_cache = NewLRUCache(8192);
+
+  BlockBasedTableOptions bbto;
+  bbto.filter_policy.reset(NewBloomFilterPolicy(10, false));
+  bbto.cache_index_and_filter_blocks = true;
+  bbto.whole_key_filtering = false;
+  options.memtable_prefix_bloom_size_ratio = 0.1;
+  options.table_factory.reset(NewBlockBasedTableFactory(bbto));
+  const size_t kTimestampSize = Timestamp(0, 0).size();
+  TestComparator test_cmp(kTimestampSize);
+  options.comparator = &test_cmp;
+  DestroyAndReopen(options);
+
+  // Write any value
+  WriteOptions write_opts;
+  std::string ts_str = Timestamp(1, 0);
+  Slice ts = ts_str;
+  write_opts.timestamp = &ts;
+
+  for (int i = 0; i < 3; i++) {
+    auto key = ToString(i * 10);
+    auto value = ToString(i * 10);
+    Slice key_slice = key;
+    Slice value_slice = value;
+    ASSERT_OK(db_->Put(write_opts, key_slice, value_slice));
+    // Flush to sstable
+    FlushOptions fo;
+    fo.wait = true;
+    ASSERT_OK(db_->Flush(fo));
+  }
+
+  db_->CompactRange(CompactRangeOptions(), nullptr, nullptr);
+
+  std::string keystr = "foo";
+  std::string vstr = "bar";
+  Slice k = keystr;
+  Slice v = vstr;
+  ASSERT_OK(db_->Put(write_opts, k, v));
+
+  // Flush to sstable
+  FlushOptions fo;
+  fo.wait = true;
+  ASSERT_OK(db_->Flush(fo));
+
+  // Read with MultiGet
+  ts_str = Timestamp(2, 0);
+  ts = ts_str;
+  ReadOptions read_opts;
+  read_opts.timestamp = &ts;
+  size_t batch_size = 1;
+  std::vector<Slice> keys(batch_size);
+  std::vector<PinnableSlice> values(batch_size);
+  std::vector<Status> statuses(batch_size);
+  keys[0] = k;
+  ColumnFamilyHandle* cfh = db_->DefaultColumnFamily();
+  db_->MultiGet(read_opts, cfh, batch_size, keys.data(), values.data(),
+                statuses.data());
+
+  ASSERT_OK(statuses[0]);
+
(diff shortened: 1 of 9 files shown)


pdillinger:
This appears to be an issue with all the Bloom filter configurations, nothing specific to kFastLocalBloom, which is not a publicly visible name for anything anyway (but is the implementation used with format_version >= 5).

table/block_based/full_filter_block.cc
```diff
@@ -245,7 +245,7 @@ void FullFilterBlockReader::MayMatch(
   MultiGetRange filter_range(*range, range->begin(), range->end());
   for (auto iter = filter_range.begin(); iter != filter_range.end(); ++iter) {
     if (!prefix_extractor) {
-      keys[num_keys++] = &iter->ukey;
+      keys[num_keys++] = &iter->ukey_without_ts;
     } else if (prefix_extractor->InDomain(iter->ukey)) {
```
pdillinger, line 249: Aren't all of these wrong?

    $ git grep -- '->ukey' | grep filter
    table/block_based/filter_block.h:      const Slice ukey = iter->ukey;
    table/block_based/filter_block.h:      const Slice ukey = iter->ukey;
    table/block_based/full_filter_block.cc:      keys[num_keys++] = &iter->ukey;
    table/block_based/full_filter_block.cc:    } else if (prefix_extractor->InDomain(iter->ukey)) {
    table/block_based/full_filter_block.cc:      prefixes.emplace_back(prefix_extractor->Transform(iter->ukey));
    $

riversand963: hmm, I think prefix is a prefix of the user key, thus should be shorter than or equal to user key. The obtained prefix from a user key should not contain timestamp.

pdillinger: In fact, I'm not really seeing instances of ->ukey that I would suspect *not* to be impacted by user timestamp. Perhaps we should rename `ukey` to `ukey_with_ts`. (I'm not saying you have to fix everything in this PR.)

pdillinger: > hmm, I think prefix is a prefix of the user key, thus should be shorter than or equal to user key. The obtained prefix from a user key should not contain timestamp.

I'm not aware of anything that guarantees that. AFAIK, prefix extractor can be identity function.

table/multiget_context.h
```diff
@@ -22,6 +24,7 @@ struct KeyContext {
   const Slice* key;
   LookupKey* lkey;
   Slice ukey;
+  Slice ukey_without_ts;
```
pdillinger, line 27: Assuming this is kept, it's probably good to update existing calls to StripTimestampFromUserKey in memtable.cc

pdillinger: Sorry, those might not be for multiget, but I'm pretty sure memtable Bloom multiget is also broken (up to you whether to fix in this PR).

table/block_based/full_filter_block.cc
riversand963: If so, we need to update the code to make sure prefix extractor operates only on user key without ts.

table/multiget_context.h
jay-zhuang: thanks @pdillinger, updated the calls to memtable bloomfilter.

table/block_based/full_filter_block.cc
jay-zhuang: Updated, and added test for that.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+  size_t batch_size = 1;
+  std::vector<Slice> keys(batch_size);
+  std::vector<PinnableSlice> values(batch_size);
+  std::vector<Status> statuses(batch_size);
+  keys[0] = k;
+  ColumnFamilyHandle* cfh = db_->DefaultColumnFamily();
+  db_->MultiGet(read_opts, cfh, batch_size, keys.data(), values.data(),
+                statuses.data());
+
+  ASSERT_OK(statuses[0]);
```
riversand963, line 536: Given that `test_cmp` will go out of scope when this test method returns, we need to call `Close()` at the end, otherwise I think ASAN may complain. This applies to all unit tests here.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+  // Write any value
+  WriteOptions write_opts;
+  std::string ts_str = Timestamp(1, 0);
+  Slice ts = ts_str;
+  write_opts.timestamp = &ts;
+
+  std::string keystr = "foo";
+  std::string vstr = "bar";
+  Slice k = keystr;
+  Slice v = vstr;
```
riversand963, line 563: Nit: `k` and `v` are not needed.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+  ASSERT_OK(db_->Put(write_opts, k, v));
+
+  // Flush to sstable
+  FlushOptions fo;
+  fo.wait = true;
+  ASSERT_OK(db_->Flush(fo));
```
riversand963, line 569: Nit: `Flush()` will do.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+  // Write any value
+  WriteOptions write_opts;
+  std::string ts_str = Timestamp(1, 0);
+  Slice ts = ts_str;
+  write_opts.timestamp = &ts;
+
+  std::string keystr = "foo";
+  std::string vstr = "bar";
+  Slice k = keystr;
+  Slice v = vstr;
```
riversand963, line 611: `k` and `v` not needed.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+  std::string ts_str = Timestamp(1, 0);
+  Slice ts = ts_str;
+  write_opts.timestamp = &ts;
+
+  // random data
+  for (int i = 0; i < 3; i++) {
+    auto key = ToString(i * 10);
+    auto value = ToString(i * 10);
+    Slice key_slice = key;
+    Slice value_slice = value;
```
riversand963, line 657: `key_slice` and `value_slice` not needed.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+    ASSERT_OK(db_->Put(write_opts, key_slice, value_slice));
+    // Flush to sstable
+    FlushOptions fo;
+    fo.wait = true;
+    ASSERT_OK(db_->Flush(fo));
```
riversand963, line 662: `Flush()` will do.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+  }
+
+  // Make num_levels to 2 to do key range filtering of sst files
+  db_->CompactRange(CompactRangeOptions(), nullptr, nullptr);
+
+  std::string keystr = "foo";
+  std::string vstr = "bar";
+  Slice k = keystr;
+  Slice v = vstr;
```
riversand963, line 671: Ditto.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+    Slice value_slice = value;
+    ASSERT_OK(db_->Put(write_opts, key_slice, value_slice));
+    // Flush to sstable
+    FlushOptions fo;
+    fo.wait = true;
+    ASSERT_OK(db_->Flush(fo));
+  }
+
+  // Make num_levels to 2 to do key range filtering of sst files
+  db_->CompactRange(CompactRangeOptions(), nullptr, nullptr);
```
riversand963, line 666: `ASSER_OK()` for the result of `CompactRange()`?

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+  std::string keystr = "foo";
+  std::string vstr = "bar";
+  Slice k = keystr;
+  Slice v = vstr;
+  ASSERT_OK(db_->Put(write_opts, k, v));
+
+  // Flush to sstable
+  FlushOptions fo;
+  fo.wait = true;
+  ASSERT_OK(db_->Flush(fo));
```
riversand963, line 677: `Flush()`

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
```
riversand963, line 522: `Flush()` will do.

db/db_with_timestamp_basic_test.cc
```diff
@@ -490,6 +490,209 @@ TEST_F(DBBasicTestWithTimestamp, ReseekToNextUserKey) {
...
+  // Write any value
+  WriteOptions write_opts;
+  std::string ts_str = Timestamp(1, 0);
+  Slice ts = ts_str;
+  write_opts.timestamp = &ts;
+
+  std::string keystr = "foo";
+  std::string vstr = "bar";
+  Slice k = keystr;
+  Slice v = vstr;
```
riversand963, line 516: `k` and `v` not needed.

db/version_set.cc
```diff
@@ -456,14 +456,15 @@ class FilePickerMultiGet {
...
                    fp_ctx.start_index_in_curr_level ||
-               user_comparator_->Compare(user_key,
-                                         ExtractUserKey(f->smallest_key)) <= 0);
+               user_comparator_->CompareWithoutTimestamp(
+                   user_key, false, ExtractUserKey(f->smallest_key), true) <=
+                   0);
 
-        int cmp_smallest = user_comparator_->Compare(
-            user_key, ExtractUserKey(f->smallest_key));
+        int cmp_smallest = user_comparator_->CompareWithoutTimestamp(
```
riversand963, line 463: Can we save one comparison even for non-opt build?

riversand963 (approved):
Thanks @jay-zhuang for the fix! Left some minor comments.

riversand963:
Another question: maybe this should also be fixed in another PR: https://github.com/facebook/rocksdb/blob/91f3b72ebc96e8083f23aca0aa516406c3e0df43/table/block_based/block_based_table_reader.cc#L2184

jay-zhuang:
> Another question: maybe this should also be fixed in another PR:
> 
> https://github.com/facebook/rocksdb/blob/91f3b72ebc96e8083f23aca0aa516406c3e0df43/table/block_based/block_based_table_reader.cc#L2184

Good point, as it's for `Get()`, will check that in a separate diff.

PR #7595, opened 2020-10-25
Ignore cmake-build-debug

darionyaphet:


jay-zhuang:
I guess you're using CLion, I would suggest to set a global ignore file to exclude `cmake-build-*`: https://stackoverflow.com/questions/7335420/global-git-ignore

ajkr (changes requested):
Classifying according to latest comments.

PR #7596, opened 2020-10-25
Remove useless assert

darionyaphet:


riversand963 (changes requested):
Thanks @darionyaphet .
In debug mode, we perform the assertion to terminate the process early, which I believe have a value for trouble-shooting purpose.
In opt mode, we do not rely on these assertions and they are no-ops.

darionyaphet:
> Thanks @darionyaphet .
> In debug mode, we perform the assertion to terminate the process early, which I believe have a value for trouble-shooting purpose.
> In opt mode, we do not rely on these assertions and they are no-ops.

I see thanks & closed

PR #7597, opened 2020-10-25
RocksJava: Improve 'get' performance

alucarded:
Summary:

1. Parametrize GetBenchmarks with key and value sizes
2. Add benchmarks for existing 'get' methods:
-  with preallocated value byte array: RocksDB#get(ColumnFamilyHandle, byte[], byte[])
- using ByteBuffer: RocksDB#get(ColumnFamilyHandle, ReadOptions, ByteBuffer, ByteBuffer)
3. Add experimental RocksDB#getCritical(ColumnFamilyHandle, byte[], byte[]), which uses [GetPrimitiveArrayCritical](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetPrimitiveArrayCritical_ReleasePrimitiveArrayCritical) instead of [GetByteArrayRegion](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#Get_PrimitiveType_ArrayRegion_routines). Add benchmark for the new method.
4. Add experimental RocksDB#getUnsafe(ColumnFamilyHandle, ReadOptions, byte[]), which uses sun.misc.Unsafe to access value data returned with native PinnableSlice object. Java PinnableSlice object ownes the native counterpart. Add benchmark for the new method.
5. [Benchmark results ("Score" column contains throughput in operations per second)](https://docs.google.com/spreadsheets/d/1n2sUdRi-WycjgemVvHnYnzLomEthyWwbmI2cCidfqR4/edit?usp=sharing)


adamretter:
I will shortly do some final benchmarking, where I just check a couple of this before we merge this.

java/jmh/src/main/java/org/rocksdb/jmh/GetBenchmarks.java
```diff
@@ -131,9 +159,70 @@ private int next() {
...
+
+  @Benchmark
+  public void preallocatedByteBufferGet() throws RocksDBException {
+    int res = db.get(getColumnFamily(), readOptions, getKeyBuf(), getValueBuf());
+    // For testing correctness:
+    //    assert res > 0;
+    //    final byte[] ret = new byte[valueSize];
+    //    valueBuf.get(ret);
+    //    System.out.println(str(ret));
+    //    valueBuf.flip();
```
javeme, line 219: remove the unused code?

java/rocksjni/pinnable_slice.cc
```diff
@@ -0,0 +1,16 @@
...
+ * Signature: (J)V
+ */
+void Java_org_rocksdb_PinnableSlice_deletePinnableSlice(JNIEnv*, jclass,
+                                                        jlong handle) {
+  auto* pinnable_slice =
+      reinterpret_cast<ROCKSDB_NAMESPACE::PinnableSlice*>(handle);
+  assert(pinnable_slice != nullptr);
+  pinnable_slice->Reset();
+  delete pinnable_slice;
+}
```
javeme, line 16: expect blank line here

java/rocksjni/rocksjni.cc
```diff
@@ -1181,6 +1181,86 @@ jint Java_org_rocksdb_RocksDB_getDirect(JNIEnv* env, jobject /*jdb*/,
...
+    return kStatusError;
+  }
+  ret_arr[0] = reinterpret_cast<jlong>(value_slice);
+  ret_arr[1] = reinterpret_cast<jlong>(value_slice->data());
+
+  env->ReleasePrimitiveArrayCritical(jret_data, ret_arr, 0);
+
+  *has_exception = false;
+
+  return static_cast<jint>(value_slice->size_);
```
javeme, line 1233: can we construct a PinnableSlice instance and just return it here?

java/jmh/src/main/java/org/rocksdb/jmh/GetBenchmarks.java
```diff
@@ -131,9 +159,70 @@ private int next() {
...
+    db.get(getColumnFamily(), getKeyArr(), getValueArr());
+  }
+
+  @Benchmark
+  public void preallocatedCriticalGet() throws RocksDBException {
+    db.getCritical(getColumnFamily(), getKeyArr(), getValueArr());
+  }
+
+  @Benchmark
+  public void preallocatedByteBufferGet() throws RocksDBException {
```
javeme, line 212: seems this test has the best throughput

java/rocksjni/rocksjni.cc
```diff
@@ -1181,6 +1181,86 @@ jint Java_org_rocksdb_RocksDB_getDirect(JNIEnv* env, jobject /*jdb*/,
...
+    JNIEnv* env, ROCKSDB_NAMESPACE::DB* db,
+    const ROCKSDB_NAMESPACE::ReadOptions& read_options,
+    ROCKSDB_NAMESPACE::ColumnFamilyHandle* column_family_handle,
+    jbyteArray jkey, jint jkey_off, jint jkey_len, jlongArray jret_data,
+    bool* has_exception) {
+  static const int kNotFound = -1;
+  static const int kStatusError = -2;
+
+  char* key =
+      reinterpret_cast<char*>(env->GetPrimitiveArrayCritical(jkey, nullptr));
```
javeme, line 1194: use direct buffer to pass key that to prevent GetPrimitiveArrayCritical+ReleasePrimitiveArrayCritical?

java/rocksjni/rocksjni.cc
```diff
@@ -1587,6 +1668,86 @@ jint rocksdb_get_helper(
...
+    // Exception occurred
+    *has_exception = true;
+    return kStatusError;
+  }
+
+  char* key_start_ptr = key + jkey_off;
+  ROCKSDB_NAMESPACE::Slice key_slice(key_start_ptr, jkey_len);
+
+  // TODO(yhchiang): we might save one memory allocation here by adding
+  // a DB::Get() function which takes preallocated jbyte* as input.
```
javeme, line 1693: looking forward to this feature

javeme (commented):
Looking forward to continuous work~

adamretter:
Based on the Benchmark results provided by @alucarded ([Benchmark results ("Score" column contains throughput in operations per second)](https://docs.google.com/spreadsheets/d/1n2sUdRi-WycjgemVvHnYnzLomEthyWwbmI2cCidfqR4/edit?usp=sharing)). I have plotted a very simple rplot to show the results more quickly
Rplot

----
In summary, preallocated methods offer better performance as you can potentially allocate once and amortize the cost over subsequent calls.

1. Pursuing API interfaces for `preallocatedByteByfferGet` and `preallocatedGet` seems obvious as these are the two APIs that end-users are comfortable working with.

2. The thing that detracts from implementing an `unsafe` API, is that at some point the user has to perform additional the work to get the data into/out of that structure, and so the performance increase is then negated.

3. The `preallocatedCriticalGet` could be interesting, but whether to use the JNI Critical API methods or not is very much a choice for the user that is embedding RocksJava. The Critical methods can have an impact on the ability of the JVM to perform GC. So it will depend on the users application and memory use profile as to whether Critical is a good fit for them. We could add this as a potential option.

@adamretter and @alanpaxton will pick up this work that our colleague @alucarded started, and continue with it. Likely producing one or two PR's that supersede and incorporate the moist important aspects of this PR.

PR #7601, opened 2020-10-26
Track WAL in MANIFEST: LogAndApply WAL events to MANIFEST

cheng-chang:
When a WAL is synced, an edit is written to MANIFEST.
After flushing memtables, the obsoleted WALs are piggybacked to MANIFEST while writing the new L0 files to MANIFEST.

Test Plan:
`track_and_verify_wals_in_manifest` is enabled by default for all tests extending `DBBasicTest`, and in db_stress_test.
Unit test `wal_edit_test`, `version_edit_test`, and `version_set_test` are also updated.
Watch all tests to pass.


db/db_impl/db_impl.h
```diff
@@ -2184,12 +2186,19 @@ extern CompressionType GetCompressionFlush(
...
+extern uint64_t PrecomputeMinLogNumberToKeep2PC(
     VersionSet* vset, const ColumnFamilyData& cfd_to_flush,
-    autovector<VersionEdit*> edit_list,
+    const autovector<VersionEdit*>& edit_list,
     const autovector<MemTable*>& memtables_to_flush,
     LogsWithPrepTracker* prep_tracker);
 
+// In non-2PC mode,
+// WALs with log number < the returned number can be deleted after the
+// cfd_to_flush column family is flushed successfully.
```
ltamasi, line 2197: Nit: please fix the formatting.

db/db_impl/db_impl_open.cc
```diff
@@ -589,18 +589,20 @@ Status DBImpl::Recover(
...
       // otherwise, in the future, if WAL tracking is enabled again,
       // since the WALs deleted when WAL tracking is disabled are not persisted
       // into MANIFEST, WAL check may fail.
       VersionEdit edit;
-      for (const auto& wal : versions_->GetWalSet().GetWals()) {
-        WalNumber number = wal.first;
-        edit.DeleteWal(number);
-      }
+      WalNumber max_wal_number =
+          std::prev(versions_->GetWalSet().GetWals().end())->first;
```
ltamasi, line 604: No need for `std::prev`: how about `versions_->GetWalSet().GetWals().rbegin()->first` ?

db/db_range_del_test.cc
```diff
@@ -159,7 +159,7 @@ TEST_F(DBRangeDelTest, MaxCompactionBytesCutsOutputFiles) {
   // Want max_compaction_bytes to trigger the end of compaction output file, not
   // target_file_size_base, so make the latter much bigger
   opts.target_file_size_base = 100 * opts.max_compaction_bytes;
-  Reopen(opts);
+  DestroyAndReopen(opts);
```
ltamasi, line 162: Just to double-check: is this related to this project (and if yes, how)?

db/memtable_list.cc
```diff
@@ -473,12 +473,27 @@ Status MemTableList::TryInstallMemtableFlushResults(
...
+        edit_list.back()->SetMinLogNumberToKeep(min_wal_number_to_keep);
+      } else {
+        min_wal_number_to_keep =
+            PrecomputeMinLogNumberToKeepNon2PC(vset, *cfd, edit_list);
+      }
+
+      std::unique_ptr<VersionEdit> wal_deletion;
+      if (vset->db_options()->track_and_verify_wals_in_manifest) {
+        const auto& wals = vset->GetWalSet().GetWals();
+        if (wals.lower_bound(min_wal_number_to_keep) != wals.begin()) {
```
ltamasi, line 492: I don't think you need `lower_bound`: you could simply compare `min_wal_number_to_keep` with `wals.begin()->first`, right?

db/memtable_list.cc
```diff
@@ -716,12 +735,36 @@ Status InstallMemtableAtomicFlushResults(
...
+  if (vset->db_options()->track_and_verify_wals_in_manifest) {
+    uint64_t min_wal_number_to_keep =
+        PrecomputeMinLogNumberToKeepNon2PC(vset, *cfds[0], edit_lists[0]);
+    for (size_t i = 1; i < cfds.size(); i++) {
+      min_wal_number_to_keep = std::min(
+          min_wal_number_to_keep,
+          PrecomputeMinLogNumberToKeepNon2PC(vset, *cfds[i], edit_lists[i]));
+    }
+    const auto& wals = vset->GetWalSet().GetWals();
+    if (wals.lower_bound(min_wal_number_to_keep) != wals.begin()) {
```
ltamasi, line 751: Same comment about `lower_bound` as above applies here as well.

db/version_edit.cc
```diff
@@ -576,7 +576,7 @@ Status VersionEdit::DecodeFrom(const Slice& src) {
           return s;
         }
 
-        wal_deletions_.emplace_back(std::move(wal_deletion));
+        wal_deletion_ = wal_deletion;
```
ltamasi, line 579: Any reason we can no longer move?

db/version_edit.h
```diff
@@ -463,22 +464,27 @@ class VersionEdit {
   bool IsWalAddition() const { return !wal_additions_.empty(); }
 
   // Delete a WAL (either directly deleted or archived).
-  void DeleteWal(WalNumber number) {
-    assert(NumEntries() == wal_deletions_.size());
-    wal_deletions_.emplace_back(number);
+  // AddWal and DeleteWalsBefore cannot be called on the same VersionEdit.
+  void DeleteWalsBefore(WalNumber number) {
+    assert(NumEntries() == !wal_deletion_.IsEmpty());
```
ltamasi, line 469: How about `assert((NumEntries() > 0) == !wal_deletion_.IsEmpty());` to make the intention clearer?

db/version_set_test.cc
```diff
@@ -1196,10 +1196,8 @@ TEST_F(VersionSetTest, WalEditsNotAppliedToVersion) {
     edits.back()->AddWal(i, WalMetadata(i));
   }
   // Delete the first half of the WALs.
-  for (uint64_t i = 1; i <= kNumWals; i++) {
-    edits.emplace_back(new VersionEdit);
-    edits.back()->DeleteWal(i);
-  }
+  edits.emplace_back(new VersionEdit);
```
ltamasi, line 1199: Seems like we're leaking this edit.

db/version_set_test.cc
```diff
@@ -1232,10 +1230,8 @@ TEST_F(VersionSetTest, NonWalEditsAppliedToVersion) {
...
```
ltamasi, line 1233: And this one.

db/version_set_test.cc
```diff
@@ -1603,11 +1643,9 @@ TEST_F(VersionSetTest, AtomicGroupWithWalEdits) {
   edits.back()->SetDBId(kDBId);
   edits.back()->MarkAtomicGroup(--remaining);
   // Delete the first added 4 WALs.
-  for (uint64_t i = 1; i < kNumWals; i++) {
-    edits.emplace_back(new VersionEdit);
-    edits.back()->DeleteWal(i);
-    edits.back()->MarkAtomicGroup(--remaining);
-  }
+  edits.emplace_back(new VersionEdit);
```
ltamasi, line 1646: Same here.

ltamasi (commented):
Thanks for the PR @cheng-chang ! Please see some questions/comments below.

db/db_impl/db_impl.cc
```diff
@@ -1308,27 +1312,52 @@ Status DBImpl::UnlockWAL() {
...
-    auto& log = *it;
-    assert(log.getting_synced);
-    if (status.ok() && logs_.size() > 1) {
-      logs_to_free_.push_back(log.ReleaseWriter());
+    auto& wal = *it;
+    assert(wal.getting_synced);
+    if (logs_.size() > 1) {
+      if (immutable_db_options_.track_and_verify_wals_in_manifest) {
+        synced_wals.AddWal(wal.number,
+                           WalMetadata(wal.writer->file()->GetFileSize()));
```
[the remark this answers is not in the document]
cheng-chang: If we do it outside of the `if` block, then assume that there is just one log now, then everytime `SyncWAL` syncs this log, a VersionEdit will be written to MANIFEST. We only want to write one VersionEdit for syncing each WAL.

db/db_range_del_test.cc
cheng-chang: Reopening DB with a different comparator will cause writing to MANIFEST to fail: https://github.com/facebook/rocksdb/blob/master/db/version_set.cc#L4579. 

Without this PR, this `Reopen` succeeds, because the test never triggers a MANIFEST write.

db/memtable_list.cc
cheng-chang: if we also check whether wals is empty, yes.

db/version_edit.cc
cheng-chang: we can still move, but it's the same as copy for now since it only holds a uint64_t.

db/version_set_test.cc
cheng-chang: `autovector<std::unique_ptr<VersionEdit>> edits;`, unique_ptr should `delete` them on destruction.

db/version_set_test.cc
cheng-chang: also `autovector<std::unique_ptr<VersionEdit>> edits;`

db/version_set_test.cc
cheng-chang: also `autovector<std::unique_ptr<VersionEdit>> edits;`

cheng-chang:
@ltamasi Thanks for the review! PTAL.

db/db_impl/db_impl.cc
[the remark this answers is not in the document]
ltamasi: > If we do it outside of the `if` block, then assume that there is just one log now, then everytime `SyncWAL` syncs this log, a VersionEdit will be written to MANIFEST.

I thought that's what we wanted (otherwise, you will only track closed WALs)?

Could you clarify under which conditions we want to log this in the MANIFEST?

db/version_set_test.cc
ltamasi: Thanks, I missed that.

db/memtable_list.cc
```diff
@@ -473,12 +473,27 @@ Status MemTableList::TryInstallMemtableFlushResults(
...
+      uint64_t min_wal_number_to_keep = 0;
       if (vset->db_options()->allow_2pc) {
         assert(edit_list.size() > 0);
+        min_wal_number_to_keep = PrecomputeMinLogNumberToKeep2PC(
+            vset, *cfd, edit_list, memtables_to_flush, prep_tracker);
         // We piggyback the information of  earliest log file to keep in the
         // manifest entry for the last file flushed.
-        edit_list.back()->SetMinLogNumberToKeep(PrecomputeMinLogNumberToKeep(
-            vset, *cfd, edit_list, memtables_to_flush, prep_tracker));
+        edit_list.back()->SetMinLogNumberToKeep(min_wal_number_to_keep);
```
ltamasi, line 483: Would it make sense to do this unconditionally (i.e. in non-2PC mode as well)?

cheng-chang: the min_log_number is only used for 2pc: https://github.com/facebook/rocksdb/blob/master/db/version_set.cc#L4244

db/db_impl/db_impl.cc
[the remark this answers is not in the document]
cheng-chang: We only track closed WALs. The reason behind this is, if we log the syncing event to MANIFEST for every syncing, then there might be too many edits, and since logging to MANIFEST needs another sync, it will negatively impact write throughput if synced writes are frequent.

db/version_edit.h
```diff
@@ -463,22 +464,27 @@ class VersionEdit {
   bool IsWalAddition() const { return !wal_additions_.empty(); }
 
   // Delete a WAL (either directly deleted or archived).
-  void DeleteWal(WalNumber number) {
-    assert(NumEntries() == wal_deletions_.size());
-    wal_deletions_.emplace_back(number);
+  // AddWal and DeleteWalsBefore cannot be called on the same VersionEdit.
+  void DeleteWalsBefore(WalNumber number) {
+    assert((NumEntries() > 0) == !wal_deletion_.IsEmpty());
```
ltamasi, line 469: On second thought, this should probably be `NumEntries() == 1` ? Otherwise, LGTM, thanks!

ltamasi (approved):

PR #7604, opened 2020-10-27
[CodeMod] Remove unused includes

riversand963:
This is a PR generated **semi-automatically** by an internal tool to remove unused includes and `using` statements.

Test plan:
make check

### DIFF
--- a/cache/lru_cache.cc
+++ b/cache/lru_cache.cc
@@ -11,7 +11,6 @@
 
 #include <assert.h>
 #include <stdio.h>
-#include <stdlib.h>
 #include <string>
 
 #include "util/mutexlock.h"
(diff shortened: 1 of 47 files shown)


riversand963:
> Can I ask how this tool decides what can be removed?
> 
> Personally, the problem I see is that a lot of header files include things that they should not. Some are unnecessary includes and some should be forward declares. I view cleaning those header files up as much more bang for the buck than cleaning up C++ files.

It was generated by an internal src code analysis service, and I am not sure about the implementation. An internal diff was generated, then I applied the diff to Github repo as a test.

The way header files are organized can be improved. In addition to that, I think part of this PR does the job by removing unnecessary inclusion of system headers, unnecessary `using` statements, etc.

ajkr (approved):
LGTM

zhichao-cao:
Is there anything that influence the windows build and testing?

PR #7605, opened 2020-10-28
Require only one `Logger::Logv()` implementation

ajkr:
A user who extended `Logger` recently pointed out it is unusual to
require they implement the two-argument `Logv()` overload when they've
already implemented the three-argument `Logv()` overload. I agree with
that and think we can fix it by only calling the two-argument overload
from the default implementation of the three-argument overload. Then
when the three-argument overload is overridden, RocksDB would not
rely on the two-argument overload. Only `Logger::LogHeader()` needed
adjustment to achieve this.

### DIFF
--- a/include/rocksdb/env.h
+++ b/include/rocksdb/env.h
@@ -1057,7 +1057,8 @@ class Logger {
   // Write an entry to the log file with the specified format.
   //
   // Users who override the `Logv()` overload taking `InfoLogLevel` do not need
-  // to implement this, unless they explicitly invoke it.
+  // to implement this, unless they explicitly invoke it in
+  // `Logv(InfoLogLevel, ...)`.
   virtual void Logv(const char* /* format */, va_list /* ap */) {
     assert(false);
   }
--- a/include/rocksdb/env.h
+++ b/include/rocksdb/env.h
@@ -1051,11 +1051,16 @@ class Logger {
   virtual void LogHeader(const char* format, va_list ap) {
     // Default implementation does a simple INFO level log write.
     // Please override as per the logger class requirement.
-    Logv(format, ap);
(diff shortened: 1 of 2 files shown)


include/rocksdb/env.h
```diff
@@ -1051,11 +1051,16 @@ class Logger {
...
     // Please override as per the logger class requirement.
-    Logv(format, ap);
+    Logv(InfoLogLevel::INFO_LEVEL, format, ap);
   }
 
   // Write an entry to the log file with the specified format.
-  virtual void Logv(const char* format, va_list ap) = 0;
+  //
+  // Users who override the `Logv()` overload taking `InfoLogLevel` do not need
+  // to implement this, unless they explicitly invoke it.
```
riversand963, line 1060: Nit: maybe explicitly state that "invoke it in `Logv(InfoLogLevel, ...)`.

riversand963 (approved):
LGTM with one minor comment.

include/rocksdb/env.h
ajkr: Done.

ajkr (commented):
Thanks for the review!

ajkr:
I'll land this, but this API still has some odd parts that could use cleanup:

1. `Logv(InfoLogLevel::HEADER_LEVEL, ...)` calls `LogHeader()` whose default implementation calls `Logv(InfoLogLevel::INFO_LEVEL, ...)`. Why do we need both `LogHeader()` and `Logv(InfoLogLevel::HEADER_LEVEL, ...)`?
2. `LogWrapper` overrides the two-argument `Logv()` to delegate to the wrapped `Logger`'s two-argument `Logv()`. But then the wrapped `Logger` has no way to find out the `InfoLogLevel` value.

PR #7607, opened 2020-10-28
Disable file deletion during recovery until switching to new MANIFEST

cheng-chang:
Consider the following sequence of events:
1. Db flushed an SST with file number N, appended to MANIFEST, and tried to sync the MANIFEST.
2. Syncing MANIFEST failed and db crashed.
3. Db tried to recover with this MANIFEST. In the meantime, no entry about the newly-flushed SST was found in the MANIFEST. Therefore, RocksDB replayed WAL and tried to flush to an SST file reusing the same file number N. This failed because file system does not support overwrite. Then Db deleted this file.
4. Db crashed again.
5. Db tried to recover. When db read the MANIFEST, there was an entry referencing N.sst. This could happen probably because the append in step 1 finally reached the MANIFEST and became visible. Since N.sst had been deleted in step 3, recovery failed.

It is possible that N.sst created in step 1 is valid. Although step 3 would still fail since the MANIFEST was not synced properly in step 1 and 2, deleting N.sst would make it impossible for the db to recover even if the remaining part of MANIFEST was appended and visible after step 5.

With this PR, there are two cases:

1. in step 3, if N.sst is not recovered from the MANIFEST, then it won't show up in the newly created MANIFEST, so in step 5, DB won't try to recover N.sst. 
2. in step 3, DB crashes while recovering from MANIFEST, then since file deletion is disabled during recovering from MANIFEST, N.sst won't be deleted, then in step 5, if N.sst shows up in MANIFEST, since N.sst was not deleted, the recovery can succeed.

Test Plan:
make check


cheng-chang:
@mrambacher I updated the java test for read only open to make it less nested. In the original test, we try to reopen as read only while the original db is not closed. Is this intentional?

riversand963 (changes requested):
Note that file deletion in `BuildTable()` called by `WriteLevel0TableForRecovery()` is not actually disabled by this PR. `DisableFileDeletions()` disables file deletion that go through `FindObsoleteFiles()`/`PurgeObsoleteFiles()`, but does nothing to file deletion operation that are directly called on `fs` or `env`. We need to disable this file deletion during recovery.
Unit test for the case described in the PR description is preferred.

riversand963:
Actually,How about we re-purpose and rename `FinishBestEffortsRecovery()` so that it is called whether `best_efforts_recovery` is true or not. In this method, we will switch to a new MANIFEST.
If we do this, then later `WriteLevel0TableForRecovery` will not fail because we have already removed the sst files whose numbers may conflict with this newly-started db instance.

cheng-chang:
> Actually,How about we re-purpose and rename `FinishBestEffortsRecovery()` so that it is called whether `best_efforts_recovery` is true or not. In this method, we will switch to a new MANIFEST.
> If we do this, then later `WriteLevel0TableForRecovery` will not fail because we have already removed the sst files whose numbers may conflict with this newly-started db instance.

Sounds good, in this way, the sst file number is also not reused. I'll open another PR for this.

PR #7608, opened 2020-10-29
Simplify a test case in Java ReadOnlyTest 

cheng-chang:
The original test nests a lot of `try` blocks. This PR flattens these blocks into independent blocks, so that each `try` block closes the DB before opening the next DB instance.

Test Plan:
watch the existing java tests to pass


zhichao-cao (approved):
The refactor LGTM, thanks!

adamretter:
@cheng-chang @zhichao-cao I am afraid that there are problems with this PR. Several RocksObjects are now never free'd.

adamretter (changes requested):
This PR has problems. Objects that were previously free'd no longer are. Please see the documentation about this here - https://github.com/facebook/rocksdb/wiki/RocksJava-Basics#memory-management

zhichao-cao:
> This PR has problems. Objects that were previously free'd no longer are. Please see the documentation about this here - https://github.com/facebook/rocksdb/wiki/RocksJava-Basics#memory-management

Thanks for pointing out. I did not realize the issue.

cheng-chang:
> This PR has problems. Objects that were previously free'd no longer are. Please see the documentation about this here - https://github.com/facebook/rocksdb/wiki/RocksJava-Basics#memory-management

Thanks for pointing out the link, I didn't realize every object needs to be closed manually. The db instances and column family handles are closed. I'm wondering is it necessary to close the options and descriptors objects manually?

java/src/test/java/org/rocksdb/ReadOnlyTest.java
```diff
@@ -31,81 +31,51 @@ public void readOnlyOpen() throws RocksDBException {
...
-      try (final RocksDB db2 = RocksDB.openReadOnly(
-          dbFolder.getRoot().getAbsolutePath())) {
-        assertThat("value").
-            isEqualTo(new String(db2.get("key".getBytes())));
+    }
+    try (final RocksDB db = RocksDB.openReadOnly(dbFolder.getRoot().getAbsolutePath())) {
+      assertThat("value").isEqualTo(new String(db.get("key".getBytes())));
+    }
+
+    final ColumnFamilyOptions cfOpts = new ColumnFamilyOptions();
```
adamretter, line 39: This is never closed.

java/src/test/java/org/rocksdb/ReadOnlyTest.java
```diff
@@ -31,81 +31,51 @@ public void readOnlyOpen() throws RocksDBException {
...
+    cfDescriptors.add(
+        new ColumnFamilyDescriptor(RocksDB.DEFAULT_COLUMN_FAMILY, new ColumnFamilyOptions()));
+    cfDescriptors.add(new ColumnFamilyDescriptor("new_cf2".getBytes(), new ColumnFamilyOptions()));
+    final List<ColumnFamilyHandle> readOnlyColumnFamilyHandleList2 = new ArrayList<>();
+    try (final RocksDB db = RocksDB.openReadOnly(dbFolder.getRoot().getAbsolutePath(),
+             cfDescriptors, readOnlyColumnFamilyHandleList2)) {
+      assertThat(new String(db.get(readOnlyColumnFamilyHandleList2.get(1), "key2".getBytes())))
+          .isEqualTo("value2");
+    } finally {
+      for (ColumnFamilyHandle handle : readOnlyColumnFamilyHandleList2) {
```
adamretter, line 77: In the latest codebase. Any column family handles created by a `RocksDB#open` call, will be auto-magically cleaned up in the corresponding `RocksDB#close` call.

So you may be able to benefit from that, and tidy this up further.

adamretter (changes requested):

java/src/test/java/org/rocksdb/ReadOnlyTest.java
cheng-chang: It's weird to close an option object. What kind of resource can't the `finalize` close?

cheng-chang: Why isn't `ColumnFamilyDescriptor` closed explicitly?

cheng-chang: It's weird to close an option object, but I understand the motivation behind this to free the memory asap. https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/ColumnFamilyDescriptor.java#L24, it's not closed explicitly here.

PR #7617, opened 2020-10-30
Add recent fixes to HISTORY.md

ajkr:
The recently reverted behavior changes were released to at least one
place internally, so we should mention the reverts in release notes.


riversand963 (approved):
Thanks @ajkr for backporting the fixes!

PR #7618, opened 2020-10-30
Report if unpinnable value encountered during backward iteration

riversand963:
There is an undocumented behavior about a certain combination of options and operations.
- inplace_update_support = true, and
- call `SeekForPrev()`, `SeekToLast()`, and/or `Prev()` on unflushed data.

We should stop the backward iteration and report an error of `Status::NotSupported`.

Test plan:
make check

### DIFF
--- a/db/db_basic_test.cc
+++ b/db/db_basic_test.cc
@@ -3363,43 +3363,6 @@ TEST_F(DBBasicTest, VerifyFileChecksums) {
 }
 #endif  // !ROCKSDB_LITE
 
-TEST_F(DBBasicTest, BackwardIteration) {
-  Options options = CurrentOptions();
-  options.create_if_missing = true;
-  options.inplace_update_support = false;
-  options.env = env_;
-  DestroyAndReopen(options);
-  constexpr int kNumKeys = 10;
-
-  // Write kNumKeys to WAL.
-  for (int i = 0; i < kNumKeys; ++i) {
-    ASSERT_OK(Put(Key(i), "val"));
-  }
-  ReadOptions read_opts;
-  read_opts.total_order_seek = true;
-  {
-    std::unique_ptr<Iterator> iter(db_->NewIterator(read_opts));
-    int count = 0;
-    for (iter->SeekToLast(); iter->Valid(); iter->Prev()) {
-      ++count;
-    }
-    ASSERT_EQ(kNumKeys, count);
-  }
-
-  // Reopen and rebuild the memtable from WAL.
-  options.create_if_missing = false;
-  options.avoid_flush_during_recovery = true;
-  options.inplace_update_support = true;
-  options.allow_concurrent_memtable_write = false;
-  Reopen(options);
-  {
-    std::unique_ptr<Iterator> iter(db_->NewIterator(read_opts));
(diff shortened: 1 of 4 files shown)


pdillinger (commented):
I think there should be *at least* one of
* a unit test for the new behavior
* stress testing of in_place_update (does not appear enabled)

riversand963:

> I think there should be _at least_ one of
> 
> * a unit test for the new behavior
>
Unit test requires debug mode, thus the `assert(false)` will always terminate the process. The test coverage is the same as before. Maybe I can try to add a unit test and expect the test to die using gtest `EXPECT_DEATH` or similar, but I seem to get warning as
```
Death tests use fork(), which is unsafe particularly in a threaded context. For this test, Google Test couldn't detect the number of threads.
```

> * stress testing of in_place_update (does not appear enabled)
>
Similarly, if db_stress is built with debug mode, the process will terminate once hits this.
If db_stress is built with opt mode, afaik, we only do that for blackbox test, but it seems broken now, according to @ajkr in another chat thread. I think a separate PR will be suitable for 1) fixing db_stress in opt mode, and 2) adding test coverage for `inplace_update=true`.

pdillinger:
Why assert(false)? That would mean anyone reading options from a config file and doing backward iteration can reach an assertion violation state. Assertions are not supposed to depend on runtime inputs, with the possible exception of checksum-protected data that we generate. I consider it generally dubious to use assertions to check public API preconditions more complex than 'x is not nullptr'.

I consider #6788 an exemplar of using Status to report precondition violations in public API calls.

db/db_basic_test.cc
```diff
@@ -3326,6 +3326,43 @@ TEST_F(DBBasicTest, VerifyFileChecksums) {
 }
 #endif  // !ROCKSDB_LITE
 
+TEST_F(DBBasicTest, BackwardIteration) {
```
pdillinger, line 3329: Maybe db_iterator_test is a better place, or maybe you consider this a test about inplace_update_support. 

¯\\_(ツ)_/¯

db/db_basic_test.cc
```diff
@@ -3326,6 +3326,43 @@ TEST_F(DBBasicTest, VerifyFileChecksums) {
...
+  // Reopen and rebuild the memtable from WAL.
+  options.create_if_missing = false;
+  options.avoid_flush_during_recovery = true;
+  options.inplace_update_support = true;
+  options.allow_concurrent_memtable_write = false;
+  Reopen(options);
+  {
+    std::unique_ptr<Iterator> iter(db_->NewIterator(read_opts));
+    iter->SeekToLast();
+    ASSERT_TRUE(iter->status().IsNotSupported());
```
pdillinger, line 3361: Suggest

    // Backward iteration not supported due to inplace_update_support = true

pdillinger (approved):
👍 👍

db/db_basic_test.cc
riversand963: Done.

db/db_basic_test.cc
riversand963: Added.

PR #7621, opened 2020-10-31
Fix a recovery corner case

cheng-chang:
Consider the following sequence of events:

1. Db flushed an SST with file number N, appended to MANIFEST, and tried to sync the MANIFEST.
2. Syncing MANIFEST failed and db crashed.
3. Db tried to recover with this MANIFEST. In the meantime, no entry about the newly-flushed SST was found in the MANIFEST. Therefore, RocksDB replayed WAL and tried to flush to an SST file reusing the same file number N. This failed because file system does not support overwrite. Then Db deleted this file.
4. Db crashed again.
5. Db tried to recover. When db read the MANIFEST, there was an entry referencing N.sst. This could happen probably because the append in step 1 finally reached the MANIFEST and became visible. Since N.sst had been deleted in step 3, recovery failed.

It is possible that N.sst created in step 1 is valid. Although step 3 would still fail since the MANIFEST was not synced properly in step 1 and 2, deleting N.sst would make it impossible for the db to recover even if the remaining part of MANIFEST was appended and visible after step 5.

After this PR, in step 3, immediately after recovering from MANIFEST, a new MANIFEST is created, then we find that N.sst is not referenced in the MANIFEST, so we delete it, and we'll not reuse N as file number. Then in step 5, since the new MANIFEST does not contain N.sst, the recovery failure situation in step 5 won't happen.

Test Plan:
1. some tests are updated, because these tests assume that new MANIFEST is created after WAL recovery.
2. a new unit test is added in db_basic_test to simulate step 3.


jbowens:
@cheng-chang What if the original MANIFEST write was writing a compaction output sstable, not a flushed sstable. Doesn't the same problem exist? In step 3, the flush to `N.sst` could successfully overwrite the compaction output sstable, but in step 5 it would be read from the MANIFEST with the original compaction output's metadata in a non-zero level.

db/db_impl/db_impl.h
```diff
@@ -1198,11 +1198,15 @@ class DBImpl : public DB {
   // REQUIRES: db mutex held when calling this function, but the db mutex can
   // be released and re-acquired. Db mutex will be held when the function
   // returns.
-  // After best-efforts recovery, there may be SST files in db/cf paths that are
-  // not referenced in the MANIFEST. We delete these SST files. In the
+  // After recovery, there may be SST files in db/cf paths that are
+  // not referenced in the MANIFEST (e.g. it's best effort recovery,
+  // or the SST files are referenced by unsynced VersionEdits in MANIFEST).
```
riversand963, line 1203: I feel it's a little confusing to say "unsynced VersionEdits in MANIFEST" in the context of recovery. If the program reaches this point, the prior crash or shutdown has already happened. If a version edit was unsynced before crash, it's very likely that the entry does not exist in the MANIFEST at all.

db/db_impl/db_impl.h
```diff
@@ -1198,11 +1198,15 @@ class DBImpl : public DB {
...
+  // We delete these SST files. In the
   // meantime, we find out the largest file number present in the paths, and
   // bump up the version set's next_file_number_ to be 1 + largest_file_number.
-  Status FinishBestEffortsRecovery();
+  // We also switch to a new MANIFEST, so if there are unsynced edits in the
```
riversand963, line 1207: Ditto.

db/version_set.cc
```diff
@@ -4118,7 +4118,7 @@ Status VersionSet::ProcessManifestWrites(
         TEST_KILL_RANDOM("VersionSet::LogAndApply:BeforeAddRecord",
                          rocksdb_kill_odds * REDUCE_ODDS2);
 #ifndef NDEBUG
-        if (batch_edits.size() > 1 && batch_edits.size() - 1 == idx) {
+        if (batch_edits.size() > 0 && batch_edits.size() - 1 == idx) {
```
riversand963, line 4121: Any reason for this change?

db/db_impl/db_impl_open.cc
```diff
@@ -516,6 +515,11 @@ Status DBImpl::Recover(
   } else {
     s = SetIdentityFile(env_, dbname_, db_id_);
   }
+  if (s.ok() && !read_only) {
+    // FinishRecovery may switch to new MANIFEST, which requires db_id to be
+    // set.
+    s = FinishRecovery();
```
riversand963, line 521: I wonder whether the writing of db_id can be included in `FinishRecovery()`.

db/db_impl/db_impl_files.cc
```diff
@@ -741,7 +741,7 @@ uint64_t PrecomputeMinLogNumberToKeep(
   return min_log_number_to_keep;
 }
 
-Status DBImpl::FinishBestEffortsRecovery() {
+Status DBImpl::FinishRecovery() {
```
riversand963, line 744: Not necessarily in this PR, but can we use `SstFileManager` to perform file deletions?

riversand963 (commented):
Thanks @cheng-chang for the fix. Left a few comments.

ajkr:
Should the underlying storage system guarantee that an unsynced write that is not visible to one reader does not become visible to a later/different reader? We can probably work with it for MANIFEST since a new file is created whenever a DB instance opens, though I don't see how it'd work for WAL with `avoid_flush_during_recovery` enabled, for example.

cheng-chang:
> @cheng-chang What if the original MANIFEST write was writing a compaction output sstable, not a flushed sstable. Doesn't the same problem exist? In step 3, the flush to `N.sst` could successfully overwrite the compaction output sstable, but in step 5 it would be read from the MANIFEST with the original compaction output's metadata in a non-zero level.

Yes, the problem also exists in your case. The current solution also works for this case.

cheng-chang:
> Should the underlying storage system guarantee that an unsynced write that is not visible to one reader does not become visible to a later/different reader? We can probably work with it for MANIFEST since a new file is created whenever a DB instance opens, though I don't see how it'd work for WAL with `avoid_flush_during_recovery` enabled, for example. In that case, the recovered DB instance would be writing to what it thinks is the tail, while an unsynced write from the previously crashed instance might just show up later?

The sync operation might just be triggered when the DB crashed, so the syncing might be in progress and succeed later.

db/db_impl/db_impl.h
cheng-chang: reworded

db/version_set.cc
cheng-chang: reworded

db/db_impl/db_impl.h
cheng-chang: reworded

db/db_impl/db_impl_files.cc
cheng-chang: can try in another PR, the original code for deleting files is reused.

db/version_set.cc
cheng-chang: a test relies on the fact that, if the edit is the last in the batch, the next sync point will be executed.

db/db_impl/db_impl_open.cc
cheng-chang: `FinishRecovery` feels too general, I renamed it to `DeleteUnreferencedSstFilesAndSwitchToNewManifest`, although a long name, but the intention is more clear. 

Also moved the db id related logic to a new method `SetDBId`.

cheng-chang:
@riversand963 
Thanks for comments, PTAL

jbowens:
> The sync operation might just be triggered when the DB crashed, so the syncing might be in progress and succeed later. @riversand963 knows more about this problem happened in production.

Wouldn't the read during recovery read the unsynced data from the dirty OS buffer cache? Or is this in the context of direct IO?

cheng-chang:
> > The sync operation might just be triggered when the DB crashed, so the syncing might be in progress and succeed later. @riversand963 knows more about this problem happened in production.
> 
> Wouldn't the read during recovery read the unsynced data from the dirty OS buffer cache? Or is this in the context of direct IO?

If the data is still in buffer cache, then it's good. But considering either direct I/O or the filesystem is a remote distributed filesystem. We observed this behavior for a rocksdb instance running on top of a remote distributed fs.

db/cuckoo_table_db_test.cc
```diff
@@ -63,6 +63,14 @@ class CuckooTableDBTest : public testing::Test {
     ASSERT_OK(DB::Open(opts, dbname_, &db_));
   }
 
+  void DestroyAndReopen(Options* options = nullptr) {
+    ASSERT_OK(db_->Close());
+    delete db_;
+    db_ = nullptr;
+    ASSERT_OK(DestroyDB(dbname_, Options()));
```
riversand963, line 70: Can `Options()` be `*options` if `options` is not null?

riversand963 (commented):
I think it will be good if we can avoid incurring one extra switching and syncing of the MANIFEST during recovery, especially for a distributed file system.

db/db_impl/db_impl_files.cc
```diff
@@ -797,11 +833,10 @@ Status DBImpl::FinishBestEffortsRecovery() {
   assert(versions_->GetColumnFamilySet());
   ColumnFamilyData* default_cfd = versions_->GetColumnFamilySet()->GetDefault();
   assert(default_cfd);
-  // Even if new_descriptor_log is false, we will still switch to a new
-  // MANIFEST and update CURRENT file, since this is in recovery.
+  // Switch to a new MANIFEST.
   s = versions_->LogAndApply(
       default_cfd, *default_cfd->GetLatestMutableCFOptions(), &edit, &mutex_,
-      directories_.GetDbDir(), /*new_descriptor_log*/ false);
+      directories_.GetDbDir(), /*new_descriptor_log*/ true);
```
riversand963, line 839: By setting `true` to `false` here (keep original code unchanged), we can save one extra MANIFEST/CURRENT switch.

riversand963: In `write_dbid_to_manifest = true` case.

cheng-chang: I didn't notice that `VersionSet::descriptor_log_` is opened lazily. So when writing db id, even if `new_descriptor_log` is `false`, it will switch to a new manifest. If not writing db id, the same logic applies here, even if it's false, a new manifest will be created.
Thanks for catching this.

cheng-chang:
@riversand963 
Thanks for comments, PTAL

HISTORY.md
```diff
@@ -20,6 +20,9 @@
...
 
+### Behavior Changes
+* The dictionary compression settings specified in `ColumnFamilyOptions::compression_opts` now additionally affect files generated by flush and compaction to non-bottommost level. Previously those settings at most affected files generated by compaction to bottommost level, depending on whether `ColumnFamilyOptions::bottommost_compression_opts` overrode them. Users who relied on dictionary compression settings in `ColumnFamilyOptions::compression_opts` affecting only the bottommost level can keep the behavior by moving their dictionary settings to `ColumnFamilyOptions::bottommost_compression_opts` and setting its `enabled` flag.
```
riversand963, line 24: This line should not be in this PR because it's irrelevant.

cheng-chang: removed

db/db_impl/db_impl_open.cc
```diff
@@ -1239,7 +1211,7 @@ Status DBImpl::RecoverLogFiles(const std::vector<uint64_t>& wal_numbers,
       // write MANIFEST with update
       status = versions_->LogAndApply(cfds, cf_opts, edit_lists, &mutex_,
                                       directories_.GetDbDir(),
-                                      /*new_descriptor_log=*/true);
+                                      /*new_descriptor_log=*/false);
```
riversand963, line 1214: Actually, https://github.com/facebook/rocksdb/pull/5856 requires that these version edits be written to a new MANIFEST. Thanks to the atomicity of `rename`, the CURRENT file either points to a new MANIFEST with all these version edits that bump the `log_number` for each column family, or  points to the old MANIFEST with non-of these version edits. We thought about using atomic group, but for compatibility, we chose not to when we landed https://github.com/facebook/rocksdb/pull/5856. For the purpose of this PR, I think it's better to keep the old behavior for correctness.

db/db_impl/db_impl_files.cc
```diff
@@ -797,8 +833,10 @@ Status DBImpl::FinishBestEffortsRecovery() {
   assert(versions_->GetColumnFamilySet());
   ColumnFamilyData* default_cfd = versions_->GetColumnFamilySet()->GetDefault();
   assert(default_cfd);
-  // Even if new_descriptor_log is false, we will still switch to a new
-  // MANIFEST and update CURRENT file, since this is in recovery.
+  // If SetDBId writes db_id to MANIFEST, then a new MANIFEST is created in
```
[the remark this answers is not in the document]
cheng-chang: removed the comments completely

db/db_impl/db_impl_open.cc
cheng-chang: I was not aware of this implication. Changed back to `true`.

db/version_set.cc
riversand963: Which test relies on triggering the following test sync point when `batch_edits.size() == 1`?

cheng-chang: I tried to rollback this change, and the tests pass. The only behavior change of the recent commits is that when writing to MANIFEST in `DeleteUnreferencedSstFiles`, if might not switch to a new MANIFEST. The test might fail when writing to a new MANIFEST with a single edit. Anyway, I rollbacked this change.

riversand963 (approved):
Thanks @cheng-chang for the PR.
The fact that all Win builds fail makes me a little worried. Maybe re-run them and make sure nothing is broken before landing.

PR #7622, opened 2020-10-31
Fix util/bloom_test.cc (close #7019)

### LINKED ISSUE #7019: bloom_test fails with non-default bits_per_key
The bloom_test fails if the --bits_per_key is specified.

[  FAILED  ] BlockBasedBloomTest.VaryingLengths
[  FAILED  ] Full/FullBloomTest.FullVaryingLengths/0, where GetParam() = 0
[  FAILED  ] Full/FullBloomTest.FullVaryingLengths/1, where GetParam() = 2

The tests fail both if bits_per_key is greater than or less than the default (which is 10).

vdimir:
This PR adjusts bloom filter false positive estimation for `3 <= bits_per_key <= 30`. I try to keep numbers for bits_per_key 10 as it was hardcoded.
Actually, it still contains some magic constants as padding for theoretical estimation. I'm not sure what is the right way. If it's not suitable, I hope someone help to fix it or gives an advice.

### DIFF
--- a/util/bloom_test.cc
+++ b/util/bloom_test.cc
@@ -30,7 +30,7 @@ int main() {
 
 using GFLAGS_NAMESPACE::ParseCommandLineFlags;
 
-DEFINE_int32(bits_per_key, 10, "");
+DEFINE_int32(bits_per_key, 10, "number of bits in filter per one key to define filter size");
 
 namespace ROCKSDB_NAMESPACE {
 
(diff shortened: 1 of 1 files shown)


util/bloom_test.cc
```diff
@@ -56,6 +56,23 @@ static int NextLength(int length) {
   return length;
 }
 
+static double EstimateRate(double k, double m, double n) {
```
pdillinger, line 59: Already available from the builder. See for example https://github.com/facebook/rocksdb/blob/b6d8e3674119c2d15f0d1143720871603683c502/util/filter_bench.cc#L399

PR #7627, opened 2020-11-02
Compute NeedCompact() after table builder Finish()

riversand963:
In `BuildTable()`, we call `builder->Finish()` before evaluating `builder->NeedCompact()`.
However, we call `builder->NeedCompact()` before `builder->Finish()` in compaction job. This can be wrong because the table properties collectors may rely on the success of `Finish()` to provide correct result for `NeedCompact()`.

Test plan (on devserver):
make check


ajkr (approved):
LGTM, thanks for the fix!

It might be beneficial to have an end-to-end test for ratio-based DTC. I believe that would also repro the bug. There is an end-to-end test for sliding window DTC that may be a useful starting point: `DBTablePropertiesTest.DeletionTriggeredCompactionMarking`.

Also if we agree it fixes a bug in the `CompactOnDeletionCollectorFactory` that we supply, I guess it should have a release note.

PR #7628, opened 2020-11-02
Avoid skipping a test in db_wal_test

riversand963:
Recent test report shows that some tests have been skipped.

For DBWALTest that inherits from DBTestBase, the following will always be
true, since `env_` is an instance of `SpecialEnv`, not `Env::Default()`. Thus the test
will always be skipped.

```
if (options.env != Env::Default()) {
  ROCKSDB_GTEST_SKIP("Test requires default environment");
  return;
}
```

Test plan:
./db_wal_test --gtest_filter=DBWALTest.TruncateLastLogAfterRecoverWithoutFlush
MEM_ENV=1 ./db_wal_test --gtest_filter=DBWALTest.TruncateLastLogAfterRecoverWithoutFlush
make check

### DIFF
--- a/db/db_wal_test.cc
+++ b/db/db_wal_test.cc
@@ -1699,7 +1699,7 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
   options.env = env_;
   options.avoid_flush_during_recovery = true;
   if (mem_env_) {
-    ROCKSDB_GTEST_SKIP("Test requires default environment");
+    ROCKSDB_GTEST_SKIP("Test requires non-mem environment");
     return;
   }
   // Test fallocate support of running file system.
--- a/db/db_wal_test.cc
+++ b/db/db_wal_test.cc
@@ -1697,10 +1697,7 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
   constexpr size_t kKB = 1024;
   Options options = CurrentOptions();
   options.avoid_flush_during_recovery = true;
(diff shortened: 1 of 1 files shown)


mrambacher (commented):
I do not think this change will work when using MEM_ENV and possibly not ENCRYPTED_ENV and possibly not any other Env loaded that way.  

Does the test pass with these values set?

riversand963:
> I do not think this change will work when using MEM_ENV and possibly not ENCRYPTED_ENV and possibly not any other Env loaded that way.
> 
> Does the test pass with these values set?

If `MEM_ENV` is set, the `open()` call will fail because this POSIX open() call tries to create a file on a non-existing path returned by `mem_env_`. We can skip the test if `MEM_ENV` is set.

db/db_wal_test.cc
```diff
@@ -1696,8 +1696,9 @@ TEST_F(DBWALTest, RestoreTotalLogSizeAfterRecoverWithoutFlush) {
 TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
   constexpr size_t kKB = 1024;
   Options options = CurrentOptions();
+  options.env = env_;
   options.avoid_flush_during_recovery = true;
-  if (options.env != Env::Default()) {
+  if (mem_env_) {
     ROCKSDB_GTEST_SKIP("Test requires default environment");
```
mrambacher, line 1702: Nit: The skip message should now be about MEM_ENV, not default

mrambacher (approved):
LGTM

PR #7629, opened 2020-11-02
Avoid skipping a test in db_test2

riversand963:
Summary:
Test report shows that this test has been skipped recently due to
a condition that will never meet. `env_` is not equal to
`Env::Default()` for DBTest2 that inherits from DBTestBase.

Test Plan:
make check
./db_test2 --gtest_filter=DBTest2.PinnableSliceAndMmapReads


mrambacher (approved):
If this works for the MEM_ENV case, I am fine with it.  At one point, the MEM_ENV did not do well with Memory mapped files, but I think I fixed that already...

jay-zhuang (approved):

PR #7632, opened 2020-11-03
Use NPHash64 in more places

pdillinger:
Summary: Since the hashes should not be persisted in output_validator
nor mock_env.

Also updated NPHash64 to use 64-bit seed, and comments.

Test Plan: make check, and new build setting that enables modification
to NPHash64, to check for behavior depending on specific values. Added
that setting to one of the CircleCI configurations.


ajkr (approved):
LGTM.

PR #7633, opened 2020-11-03
Always apply bottommost_compression_opts when enabled

### LINKED ISSUE #7631: `ColumnFamilyOptions::bottommost_compression_opts` sometimes ignored

### Expected behavior

Enabling `CompressionOptions::enabled` in `ColumnFamilyOptions::bottommost_compression_opts` should be sufficient to make those options take effect.

### Actual behavior

Those options take effect conditionally on `ColumnFamilyOptions::bottommost_compression`. If `ColumnFamilyOptions::bottommost_compression` uses the default value to fallback to `ColumnFamilyOptions::compression`, those options are ignored.

Code: https://github.com/facebook/rocksdb/blob/a388c8cc6b5c5e5edefdc58b856cce396dfad7f7/db/compaction/compaction_picker.cc#L145

Unclear if this is a bug or a behavior change. Will probably call it a behavior change for `ColumnFamilyOptions::bottommost_compression_opts` since it's not particularly obvious how those options should be interpreted when there's no setting for `ColumnFamilyOptions::bottommost_compression`

ajkr:
Previously, even when `bottommost_compression_opts`'s `enabled` flag was set, it only took effect when
`bottommost_compression` was also set to something other than `kDisableCompressionOption`.
This wasn't documented and, if we kept the old behavior, it'd make
things complicated like the migration instructions in #7619. We can
simplify the API by making `bottommost_compression_opts` always take
effect when its `enabled` flag is set.

Fixes #7631.


db/db_options_test.cc
```diff
@@ -952,6 +952,63 @@ TEST_F(DBOptionsTest, ChangeCompression) {
...
+  // Verify the bottommost compression options still take effect even when the
+  // bottommost compression type is left at its default value. Verify for both
+  // automatic and manual compaction.
+  if (!Snappy_Supported() || !LZ4_Supported()) {
+    return;
+  }
+
+  const int kUpperCompressionLevel = 1;
+  const int kBottommostCompressionLevel = 2;
+  const int kNumL0Files = 2;
```
ltamasi, line 965: Minor: these could be marked `constexpr`.

db/db_options_test.cc
```diff
@@ -952,6 +952,63 @@ TEST_F(DBOptionsTest, ChangeCompression) {
...
+  options.bottommost_compression_opts.level = kBottommostCompressionLevel;
+  options.bottommost_compression_opts.enabled = true;
+  Reopen(options);
+
+  CompressionType compression_used = CompressionType::kDisableCompressionOption;
+  CompressionOptions compression_opt_used;
+  bool compacted = false;
+  SyncPoint::GetInstance()->SetCallBack(
+      "CompactionPicker::RegisterCompaction:Registered", [&](void* arg) {
+        Compaction* c = reinterpret_cast<Compaction*>(arg);
```
ltamasi, line 980: Minor: `static_cast` should be sufficient here.

db/db_options_test.cc
```diff
@@ -952,6 +952,63 @@ TEST_F(DBOptionsTest, ChangeCompression) {
...
+  options.compression = CompressionType::kLZ4Compression;
+  options.compression_opts.level = kUpperCompressionLevel;
+  options.bottommost_compression_opts.level = kBottommostCompressionLevel;
+  options.bottommost_compression_opts.enabled = true;
+  Reopen(options);
+
+  CompressionType compression_used = CompressionType::kDisableCompressionOption;
+  CompressionOptions compression_opt_used;
+  bool compacted = false;
+  SyncPoint::GetInstance()->SetCallBack(
```
ltamasi, line 978: We should probably call `DisableProcessing` and `ClearAllCallBacks` after the compactions to be safe.

HISTORY.md
```diff
@@ -17,6 +17,7 @@
 
 ### Behavior Changes
 * The dictionary compression settings specified in `ColumnFamilyOptions::compression_opts` now additionally affect files generated by flush and compaction to non-bottommost level. Previously those settings at most affected files generated by compaction to bottommost level, depending on whether `ColumnFamilyOptions::bottommost_compression_opts` overrode them. Users who relied on dictionary compression settings in `ColumnFamilyOptions::compression_opts` affecting only the bottommost level can keep the behavior by moving their dictionary settings to `ColumnFamilyOptions::bottommost_compression_opts` and setting its `enabled` flag.
+* When the `enabled` flag is set in `ColumnFamilyOptions::bottommost_compression_opts`, those compression options now take effect regardless of the value in `ColumnFamilyOptions::bottommost_compression`. Previously, those compression options only took effect when `ColumnFamilyOptions::bottommost_compression` was set to something other than `kDisableCompressionOption`.
```
ltamasi, line 20: We could consider calling out what compression type will be used in the `bottommost_compression == kDisableCompressionOption && bottommost_compression_opts.enabled` case (level-specific one if configured, `compression` otherwise).

ltamasi (approved):

ajkr (commented):
Thanks for the review!

PR #7635, opened 2020-11-03
Ribbon: major re-work of hashing, seeds, and more

pdillinger:
Summary:
* Fully optimized StandardHasher, in terms of efficiently generating Start, CoeffRow, and ResultRow from a stock hash value, with sufficient independence between them to have no measurably degraded behavior. (Degraded behavior would be an FP rate higher than explainable by 2^-b and, if using a 32-bit stock hash function, expected stock hash collisions.) Details in code comments.
* Our standard 64-bit and 32-bit hash functions do not exhibit sufficient independence on sequential seeds (for one Ribbon construction attempt to have independent probability from the next). I have worked around this in the Ribbon code by "pre-mixing" "ordinal seeds," sequentially tried and appropriate for storage in persisted metadata, into "raw seeds," ready for application and appropriate for in-memory storage. This way the pre-mixing step (though fast) is only applied on loading or configuring the structure, not on each query or banding add.
* Fix a subtle flaw in which backtracking not clearing ResultRow data could lead to elevated FP rate on keys that were backtracked on and should (for generality) exhibit the same FP rate as novel keys.
* Added a basic test for PhsfQuery and construction algorithms (map or "retrieval structure" rather than set or filter), and made a few trivial related fixes.
* Better random configuration generation in unit tests
* Some other minor cleanup / clarification / etc.

Test Plan: unit tests included


util/ribbon_impl.h
```diff
@@ -126,15 +122,31 @@ struct AddInputSelector<Key, ResultRow, true /*IsFilter*/> {
 // can do" with available hash information in terms of FP rate and
 // compactness. (64 bits recommended and sufficient for PHSF practical
 // purposes.)
+//
+// Another feature of this hasher is a minimal "premixing" of seeds before
+// they are provided to TypesAndSettings::HashFn in case that function does
+// not provide sufficiently independent hashes when iterating merely
+// sequentially on seeds. (This for example works around a problem with the
+// preview version 0.7.2 of XXH3 used in RocksDB, a.k.a. XXH3p or Hash64, and
```
jay-zhuang, line 130: Why not just use the new version of XXH3 for ribbon? to avoid this next seed calculation.

jay-zhuang (approved):
LGTM. Just a few questions.

util/hash.h
```diff
@@ -29,6 +29,8 @@ namespace ROCKSDB_NAMESPACE {
 
 // Stable/persistent 64-bit hash. Higher quality and generally faster than
 // Hash(), especially for inputs > 24 bytes.
+// KNOWN FLAW: incrementing seed by 1 might not give sufficiently independent
+// results from previous seed. Recommend incrementing by a large odd number.
```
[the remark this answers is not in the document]
pdillinger: https://github.com/facebook/rocksdb/blob/4c2aef04bdd8538a9c2d6199f71d50fa9ab83a7e/util/ribbon_test.cc#L449-L452

This number will be elevated if the independence is insufficient, which will fail the test, at least with high --thoroughness setting. Or the max seeds for ResetAndFindSeedToSolve will be exceeded and fail.

util/ribbon_impl.h
pdillinger: We would then need to maintain two versions of XXH code in RocksDB, which not only increases maintenance complexity, but also working code size when using both versions.

But even that concern is based on an assumption I expect to be false: I expect only to use StandardRehasher in RocksDB, and only use one seed (default, like Bloom filter) for hashing to initial 64 bits.

So this fix for handling seeds was really about validating and testing the implementation's generality than it was about what we expect to use in RocksDB. (I might have even deceived myself for a while on that point.)

util/ribbon_impl.h
```diff
@@ -226,33 +285,71 @@ class StandardHasher {
...
-  void ResetSeed(Seed seed = 0) { seed_ = seed; }
 
  protected:
-  Seed seed_;
+  static constexpr Hash kMixMask = static_cast<Hash>(0xf0f0f0f0f0f0f0f0U);
+  static constexpr unsigned kMixShift = 4U;
+  static constexpr Hash kToRawSeedFactor =
+      static_cast<Hash>(0xc78219a23eeadd03U);
+  static constexpr Hash kFromRawSeedFactor =
+      static_cast<Hash>(0xfe1a137d14b475abU);
```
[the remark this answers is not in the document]
pdillinger: Yes. If some number is congruent to 1 in mod 2^64, it's also congruent to 1 in mod 2^x for any whole x < 64 (since 2^x divides 2^64).

    >>> (0xadd03 * 0x475ab) & 0xfffff
    1

util/ribbon_impl.h
pdillinger: > So this fix for handling seeds was really about validating and testing the implementation's generality than it was about what we expect to use in RocksDB. (I might have even deceived myself for a while on that point.)

Correction: StandardRehasher needed to do some kind of mixing on "ordinal seeds" anyway for it to work properly, and doing that ahead of time saves maybe a few per-query cycles. The arguably flawed handling of seeds by our built-in hash functions just made it an obvious choice to share the seed pre-mixing infrastructure for both use-cases (sufficiently rehashing/re-seeding a precomputed hash, or independently seeding a stock hash function).

PR #7636, opened 2020-11-04
[NO LAND] Remove DB::VerifyFileChecksums

riversand963:
Remove the API `DB::VerifyFileChecksums()` added in #7578 

This PR will not go to master.
I plan to squash all commits of this PR and cherry-pick to 6.14.fb branch if all tests pass and code change gets approved.

Test plan:
make check


riversand963:
After this lands, I am going to cherry-pick this to 6.14.fb and then revert this in master because I think this is not the best solution in the mid-to-long term.

db/db_impl/db_impl_open.cc
```diff
@@ -475,6 +475,12 @@ Status DBImpl::Recover(
     assert(!files_in_dbname.empty());
     s = versions_->TryRecover(column_families, read_only, files_in_dbname,
                               &db_id_, &missing_table_file);
+#ifndef ROCKSDB_LITE
+    // TODO: remove the VerifyFileChecksums() call because it's very expensive.
+    if (s.ok() && immutable_db_options_.file_checksum_gen_factory) {
```
pdillinger, line 480: This seems quite general. I would expect some external users to use file_checksum_gen_factory. What if we used a weak symbol (see e.g. jemalloc_helper.h) as an unofficial temporary hook (callback) here?

pdillinger (changes requested):
To satisfy me that you haven't modified the behavior of VerifyChecksum in #7578, please move this code from db_impl.cc under the `if (use_file_checksum)`

    const FileMetaData* fmeta = fd_with_krange.file_metadata;
    assert(fmeta);

Otherwise, it's a simple reduction from the new to the old.

> After this lands, I am going to cherry-pick this to 6.14.fb and then revert this in master

I don't think this ever has to go in master. If CI results look good, just pick it without landing to master (IMHO).

db/db_impl/db_impl_open.cc
riversand963: IMO, this line should not be here if I were not trying to create a patch for 6.14.fb. This operation is very expensive and will be removed from here asap after we cherrypick to 6.14

riversand963:
> To satisfy me that you haven't modified the behavior of VerifyChecksum in #7578, please move this code from db_impl.cc under the `if (use_file_checksum)`
> 
> ```
> const FileMetaData* fmeta = fd_with_krange.file_metadata;
> assert(fmeta);
> ```
> 
> Otherwise, it's a simple reduction from the new to the old.
>
Could you elaborate more on this? I thought as long as we do not introduce API change, we can do a patch release.
> 
> > After this lands, I am going to cherry-pick this to 6.14.fb and then revert this in master
> 
> I don't think this ever has to go in master. If CI results look good, just pick it without landing to master (IMHO).

SG.

db/db_impl/db_impl_open.cc
riversand963: This will be executed only if people set `options.best_efforts_recovery` to true.
Not sure if I follow the callback thing. What should we use a callback for?

pdillinger:
> > To satisfy me that you haven't modified the behavior of VerifyChecksum in #7578, please move this code from db_impl.cc under the `if (use_file_checksum)`
> > ```
> > const FileMetaData* fmeta = fd_with_krange.file_metadata;
> > assert(fmeta);
> > ```
> > 
> > 
> > Otherwise, it's a simple reduction from the new to the old.
> 
> Could you elaborate more on this? I thought as long as we do not introduce API change, we can do a patch release.

In an internal discussion @ajkr mentioned that the patch could "expose existing users of VerifyChecksum() to a risk that they wouldn't expect from a patch release." I didn't fully consider that in my review of #7578, so I'm proposing a change to minimize that risk, by making it easy to reason no change in behavior for VerifyChecksum in the patch. Maybe `assert(fmeta)` is extremely safe, but I can't verify that without some digging. That's why I propose we only do it for VerifyFileChecksums (use_file_checksum).

db/db_impl/db_impl_open.cc
pdillinger: OK, I missed that bit of context, but still, from just our internal customer perspective, it seems very dangerous not to have any way to disable scanning the whole DB on startup other than through disabling best_efforts_recovery.

What I was envisioning was

    extern bool EnableVerifyFileChecksumsOnRecovery() __attribute__((__weak__));
    ...
    if (s.ok() && EnableVerifyFileChecksumsOnRecovery && EnableVerifyFileChecksumsOnRecovery()) {
      ...

riversand963:
Sure, I moved
```
assert(fmeta);
```
to the `if (use_file_checksum)` block.

db/db_impl/db_impl_open.cc
riversand963: Valid concern. I am actually more concerned with external users who have to stick with 6.14 but want to set `options.best_efforts_recovery` to `true`. For internal customer, I think they are ok with performing this check by default.

riversand963: Done with a weak symbol `RocksDbFileChecksumsVerificationEnabledOnRecovery`.

riversand963:
I plan to squash all commits of this PR and cherry-pick to 6.14.fb branch if all tests pass and code change gets approved.

pdillinger (approved):
You can blame me if anyone complains about the ugliness of the solution, but I think it's the most reasonable available to keep the customer happy. Thanks!

db/db_impl/db_impl_open.cc
```diff
@@ -1429,6 +1436,22 @@ Status DBImpl::WriteLevel0TableForRecovery(int job_id, ColumnFamilyData* cfd,
   return s;
 }
 
+Status DBImpl::MaybeVerifyFileChecksums() {
+  Status s;
+#if !defined(ROCKSDB_LITE) && defined(OS_LINUX)
```
pdillinger, line 1441: Actually CorruptionTest is failing on macos and probably others. Update the relevant test check to be Linux only (for this patch)?

riversand963: Sure, will enable the failing check only in Linux.

riversand963: MacOS test failure is fixed.

darionyaphet:
Hi @riversand963 why we don't need `VerifyFileChecksums ` any more ?

riversand963:
> Hi @riversand963 Currently, it's not supported. why we don't need `VerifyFileChecksums ` any more ?

`VerifyFileChecksums()` is still in master and will be available in 6.15. The purpose of this PR is to remove the new API from 6.14.fb branch once #7578 is cherry-picked to 6.14.fb branch because we do **not** include API change in patch releases.

`DB::VerifyFileChecksums()` is a new API that does not before. It is different from the existing `DB::VerifyChecksum()`.

PR #7637, opened 2020-11-05
Disable fsync in db_range_del_test

cheng-chang:
This test often times out in internal test infra.

Test Plan:
watch test to pass


ajkr (approved):
LGTM!

PR #7638, opened 2020-11-05
ColumnFamilyTest often times out in internal test infra

cheng-chang:
Tries to fix by skipping fsync.

Test Plan:
watch the tests to pass


riversand963 (approved):
Mostly LGTM, and make sure the change does not break tests.

PR #7644, opened 2020-11-05
Fix a seek issue with prefix extractor and timestamp

jay-zhuang:
During seek, prefix compare should not include timestamp.

Test Plan: added unittest

### DIFF
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -284,8 +284,10 @@ TEST_F(DBBasicTestWithTimestamp, SeekWithPrefix) {
     std::unique_ptr<Iterator> iter(db_->NewIterator(read_opts));
     iter->Seek("foo");
     ASSERT_TRUE(iter->Valid());
+    ASSERT_OK(iter->status());
     iter->Seek("bbb");
     ASSERT_FALSE(iter->Valid());
+    ASSERT_OK(iter->status());
   }
 
   Close();
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -244,7 +244,7 @@ TEST_F(DBBasicTestWithTimestamp, SimpleForwardIterate) {
   Close();
 }
 
-TEST_F(DBBasicTestWithTimestamp, Seek) {
+TEST_F(DBBasicTestWithTimestamp, SeekWithPrefix) {
   Options options = CurrentOptions();
   options.env = env_;
   options.create_if_missing = true;
@@ -265,26 +265,25 @@ TEST_F(DBBasicTestWithTimestamp, Seek) {
   write_opts.timestamp = &ts;
 
   ASSERT_OK(db_->Put(write_opts, "foo1", "bar"));
-
   Flush();
 
   ASSERT_OK(db_->Put(write_opts, "foo2", "bar"));
-
   Flush();
 
+  // Move sst file to next level
   ASSERT_OK(db_->CompactRange(CompactRangeOptions(), nullptr, nullptr));
 
   ASSERT_OK(db_->Put(write_opts, "foo3", "bar"));
-
   Flush();
 
+  ReadOptions read_opts;
+  std::string read_ts = Timestamp(1, 0);
+  ts = read_ts;
+  read_opts.timestamp = &ts;
   {
-    ReadOptions read_opts;
-    std::string read_ts = Timestamp(1, 0);
-    ts = read_ts;
-    read_opts.timestamp = &ts;
     std::unique_ptr<Iterator> iter(db_->NewIterator(read_opts));
     iter->Seek("foo");
+    ASSERT_TRUE(iter->Valid());
   }
 
   Close();
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -341,7 +341,6 @@ TEST_F(DBBasicTestWithTimestamp, SeekWithPrefixLargerThanKey) {
   Close();
 }
 
-
 TEST_F(DBBasicTestWithTimestamp, SimpleForwardIterateLowerTsBound) {
   constexpr int kNumKeysPerFile = 128;
   constexpr uint64_t kMaxKey = 1024;
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -244,6 +244,52 @@ TEST_F(DBBasicTestWithTimestamp, SimpleForwardIterate) {
   Close();
 }
 
+TEST_F(DBBasicTestWithTimestamp, Seek) {
+  Options options = CurrentOptions();
+  options.env = env_;
+  options.create_if_missing = true;
+  options.prefix_extractor.reset(NewFixedPrefixTransform(5));
+  BlockBasedTableOptions bbto;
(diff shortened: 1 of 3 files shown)


db/db_with_timestamp_basic_test.cc
```diff
@@ -244,6 +244,53 @@ TEST_F(DBBasicTestWithTimestamp, SimpleForwardIterate) {
   Close();
 }
 
+TEST_F(DBBasicTestWithTimestamp, SeekWithPrefix) {
+  Options options = CurrentOptions();
+  options.env = env_;
+  options.create_if_missing = true;
+  options.prefix_extractor.reset(NewFixedPrefixTransform(3));
```
riversand963, line 251: Can we use a prefix extractor that does not simply extract a prefix from user key?
If the prefix extractor takes <user_key, ts> and just returns a substring of user_key, then the fix introduced in this PR won't actually make a difference.
Is there a prefix extractor that takes <user_key, ts> and returns a result based on **both** user_key and ts?

riversand963 (approved):
Thanks @jay-zhuang for the fix. Mostly LGTM except a comment on test.

db/db_with_timestamp_basic_test.cc
jay-zhuang: I never thought about such usecase. Are you saying there could be usecase like: get prefix with the whole key + part of timestamp?
For example, with the following data:
```
foo1, ts1 -> value1
foo1, ts2 -> value2
foo1, ts3 -> value3
```
And prefix extractor is: `key + first_part_of(ts)`. Is that correct? Or a valid case? And I assume the prefix extractor can only return a prefix of the user key, so to include timestamp, it has to include the whole key. Is that right?

This unittest just to reproduce the issue that the prefix key is compared with `CompareWithoutTimestamp()` instead of `Compare()`.

riversand963: I believe so, according to a discussion with @pdillinger. https://github.com/facebook/rocksdb/pull/7589#discussion_r511062152

Some of our existing code feeds <user_key, ts> as input to prefix extractor, and we cannot assume that the `ts` part does not affect output of prefix extraction.

jay-zhuang: I don't think we should include timestamp for prefix extractor:
1. it's an internal implementation detail that we should not expose to the user. That we have timestamp right after user key and how the timestamp is encoded (I think you mentioned that we might remove timestamp if the data is on the bottom level like seq_num and maybe delta encoding?);
2. All user's existing prefix extractors need to update to handle the timestamp, as the prefix may or may not include the timestamp: https://github.com/facebook/rocksdb/blob/fdf882ded218344c136c97daf76dfb59e4bc155f/util/slice.cc#L26
3. I don't feel it's a useful usecase (if it's still valid) to include timestamp in prefix extractor. If we do need it, we should have a new prefix extractor interface to explicitly having the timestamp.

What you think?

riversand963: Sorry for the confusion, but I don't think the unit test is actually testing what we meant to fix. In fact, if just change version_set.cc as follows, the unit test still can pass.
```
diff --git a/db/version_set.cc b/db/version_set.cc
index 0fe09ef34..1d4ef1853 100644
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -1100,9 +1100,9 @@ void LevelIterator::Seek(const Slice& target) {
     Slice file_user_key = ExtractUserKey(file_iter_.key());
     if (prefix_extractor_->InDomain(target_user_key) &&
         (!prefix_extractor_->InDomain(file_user_key) ||
-         user_comparator_.Compare(
-             prefix_extractor_->Transform(target_user_key),
-             prefix_extractor_->Transform(file_user_key)) != 0)) {
+         user_comparator_.CompareWithoutTimestamp(
+             prefix_extractor_->Transform(target_user_key), false,
+             prefix_extractor_->Transform(file_user_key), false) != 0)) {
       SetFileIterator(nullptr);
     }
   }
```
Namely, without stripping the timestamp from user key, the unit test can still pass.
I was saying we need a test case in which your change will make the unit pass, but the above will not.

riversand963: My original comment was proposing a test case which, without your change, will mistakenly include timestamp for prefix extraction and yield a different result.

riversand963:
Or, we can keep unit test, but change only one line in version_set.cc as follows
```
diff --git a/db/version_set.cc b/db/version_set.cc
index 0fe09ef34..1d4ef1853 100644
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -1100,9 +1100,9 @@ void LevelIterator::Seek(const Slice& target) {
     Slice file_user_key = ExtractUserKey(file_iter_.key());
     if (prefix_extractor_->InDomain(target_user_key) &&
         (!prefix_extractor_->InDomain(file_user_key) ||
-         user_comparator_.Compare(
-             prefix_extractor_->Transform(target_user_key),
-             prefix_extractor_->Transform(file_user_key)) != 0)) {
+         user_comparator_.CompareWithoutTimestamp(
+             prefix_extractor_->Transform(target_user_key), false,
+             prefix_extractor_->Transform(file_user_key), false) != 0)) {
       SetFileIterator(nullptr);
     }
   }
```
Then, we update the PR description so that it describes the real fix here: change `Compre` to `CompareWithoutTimestamp`.
Then, I think this PR is ready to go.

riversand963 (changes requested):
Thanks @jay-zhuang for the fix!

jay-zhuang:
> Or, we can keep unit test, but change only one line in version_set.cc as follows
> 
> ```
> diff --git a/db/version_set.cc b/db/version_set.cc
> index 0fe09ef34..1d4ef1853 100644
> --- a/db/version_set.cc
> +++ b/db/version_set.cc
> @@ -1100,9 +1100,9 @@ void LevelIterator::Seek(const Slice& target) {
>      Slice file_user_key = ExtractUserKey(file_iter_.key());
>      if (prefix_extractor_->InDomain(target_user_key) &&
>          (!prefix_extractor_->InDomain(file_user_key) ||
> -         user_comparator_.Compare(
> -             prefix_extractor_->Transform(target_user_key),
> -             prefix_extractor_->Transform(file_user_key)) != 0)) {
> +         user_comparator_.CompareWithoutTimestamp(
> +             prefix_extractor_->Transform(target_user_key), false,
> +             prefix_extractor_->Transform(file_user_key), false) != 0)) {
>        SetFileIterator(nullptr);
>      }
>    }
> ```
> 
> Then, we update the PR description so that it describes the real fix here: change `Compre` to `CompareWithoutTimestamp`.
> Then, I think this PR is ready to go.

I added an unit test to cover the case that prefix extractor includes timestamp, which will return invalid result if timestamp is included. As prefix bloom filter uses prefix without timestamp, so as here.

db/db_with_timestamp_basic_test.cc
jay-zhuang: That makes sense. Added one.

pdillinger: https://github.com/facebook/rocksdb/blob/c3911f1a72add1ec7d86ad337c9f39182ccdb16a/include/rocksdb/options.h#L225-L232

I think we have to assume things might break if "prefix" is not actually a prefix of user key (without timestamp), but I haven't thought through what might break.

riversand963 (approved):

PR #7649, opened 2020-11-09
Track WAL in MANIFEST: Track deleted WALs in MANIFEST after recovering from the WALs

cheng-chang:
After replaying the WALs, the memtables are flushed synchronously to L0 instead of being flushed in background. Currently, we only track WAL obsoletion events in the code path of background flush jobs. This PR tracks these events in RecoverLogFiles.

After this change, we can enable `track_and_verify_wal_in_manifest` in `db_stress`.

Test Plan:
`python tools/db_crashtest.py whitebox`


riversand963 (approved):
Thanks @cheng-chang for the fix

PR #7653, opened 2020-11-10
Inject the random write error to stress test

zhichao-cao:
Inject the random write error to stress test, it requires set reopen=0 and disable_wal=true. 

test plan: pass db_stress and python3 db_crashtest.py blackbox

### DIFF
--- a/db_stress_tool/db_stress_tool.cc
+++ b/db_stress_tool/db_stress_tool.cc
@@ -97,14 +97,25 @@ int db_stress_tool(int argc, char** argv) {
   }
 
 #ifndef NDEBUG
-  if (FLAGS_read_fault_one_in || FLAGS_sync_fault_injection) {
+  if (FLAGS_read_fault_one_in || FLAGS_sync_fault_injection ||
+      FLAGS_write_fault_one_in) {
     FaultInjectionTestFS* fs =
         new FaultInjectionTestFS(raw_env->GetFileSystem());
     fault_fs_guard.reset(fs);
-    fault_fs_guard->SetFilesystemDirectWritable(true);
+    if (FLAGS_write_fault_one_in) {
+      fault_fs_guard->SetFilesystemDirectWritable(false);
+    } else {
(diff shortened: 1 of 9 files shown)


utilities/fault_injection_fs.cc
```diff
@@ -99,7 +99,8 @@ IOStatus TestFSWritableFile::Append(const Slice& data, const IOOptions&,
   state_.buffer_.append(data.data(), data.size());
   state_.pos_ += data.size();
   fs_->WritableFileAppended(state_);
-  return IOStatus::OK();
+  IOStatus io_s = fs_->InjectWriteError(state_.filename_);
```
ajkr, line 102: is there an assumption that our retry logic must always throw away the file? (If it doesn't, I believe retrying on a same file would result in the data being double appended.)

ajkr (approved):
LGTM!

utilities/fault_injection_fs.cc
zhichao-cao: Yes, the flush and compaction will remove the obsolete files.

anand1976 (approved):

db_stress_tool/db_stress_tool.cc
```diff
@@ -97,20 +97,42 @@ int db_stress_tool(int argc, char** argv) {
...
+      fault_fs_guard->SetFilesystemDirectWritable(true);
+    }
     fault_env_guard =
         std::make_shared<CompositeEnvWrapper>(raw_env, fault_fs_guard);
     raw_env = fault_env_guard.get();
   }
+  if (FLAGS_write_fault_one_in) {
+    SyncPoint::GetInstance()->SetCallBack(
+        "BuildTable:BeforeFinishBuildTable",
+        [&](void*) { fault_fs_guard->EnableWriteErrorInjection(); });
```
ajkr, line 117: @zhichao-cao Any chance you remember why `EnableWriteErrorInjection()` happens at the time of "BuildTable:BeforeFinishBuildTable", as opposed to simply after `Open()`?

PR #7654, opened 2020-11-10
Add arena_block_size flag to db_bench

mulugetam:
db_bench currently does not allow overriding the default `arena_block_size `calculation ([memtable size/8](https://github.com/facebook/rocksdb/blob/master/db/column_family.cc#L216)). For memtables whose size is in gigabytes, the `arena_block_size` defaults to hundreds of megabytes (affecting performance). 

Exposing this option in db_bench would allow us to test the workloads with various `arena_block_size` values.


ajkr (approved):
LGTM, thanks.

PR #7655, opened 2020-11-11
Add full_history_ts_low_ to FlushJob

riversand963:
#7556 enables `CompactionIterator` to perform garbage collection during compaction according
to a lower bound (user-defined) timestamp `full_history_ts_low_`.
This PR adds a data member `full_history_ts_low_` of type `std::string` to `FlushJob`, and
`full_history_ts_low_` does not change during flush. `FlushJob` will pass a pointer to this data member
to the `CompactionIterator` used during flush.

Also refactored flush_job_test.cc to re-use some existing code, which is actually the majority of this PR.

Test plan
make check

### DIFF
--- a/db/builder.cc
+++ b/db/builder.cc
@@ -69,8 +69,8 @@ TableBuilder* NewTableBuilder(
 }
 
 Status BuildTable(
-    const std::string& dbname, VersionSet* versions, Env* env, FileSystem* fs,
-    const ImmutableCFOptions& ioptions,
+    const std::string& dbname, VersionSet* versions,
+    const ImmutableDBOptions& db_options, const ImmutableCFOptions& ioptions,
     const MutableCFOptions& mutable_cf_options, const FileOptions& file_options,
     TableCache* table_cache, InternalIterator* iter,
     std::vector<std::unique_ptr<FragmentedRangeTombstoneIterator>>
@@ -120,6 +120,10 @@ Status BuildTable(
   EventHelpers::NotifyTableFileCreationStarted(
       ioptions.listeners, dbname, column_family_name, fname, job_id, reason);
 #endif  // !ROCKSDB_LITE
+  Env* env = db_options.env;
+  assert(env);
+  FileSystem* fs = db_options.fs.get();
+  assert(fs);
   TableProperties tp;
   if (iter->Valid() || !range_del_agg->IsEmpty()) {
     TableBuilder* builder;
@@ -175,7 +179,6 @@ Status BuildTable(
                                   blob_file_additions)
             : nullptr);
 
-    std::shared_ptr<Logger> info_log;
     CompactionIterator c_iter(
         iter, internal_comparator.user_comparator(), &merge, kMaxSequenceNumber,
         &snapshots, earliest_write_conflict_snapshot, snapshot_checker, env,
@@ -185,7 +188,7 @@ Status BuildTable(
         /*compaction=*/nullptr,
         /*compaction_filter=*/nullptr, /*shutting_down=*/nullptr,
         /*preserve_deletes_seqnum=*/0, /*manual_compaction_paused=*/nullptr,
-        /*info_log=*/info_log, full_history_ts_low);
+        db_options.info_log, full_history_ts_low);
 
     c_iter.SeekToFirst();
     for (; c_iter.Valid(); c_iter.Next()) {
--- a/db/builder.cc
+++ b/db/builder.cc
@@ -91,7 +91,7 @@ Status BuildTable(
     TableProperties* table_properties, int level, const uint64_t creation_time,
     const uint64_t oldest_key_time, Env::WriteLifeTimeHint write_hint,
     const uint64_t file_creation_time, const std::string& db_id,
-    const std::string& db_session_id) {
+    const std::string& db_session_id, const std::string* full_history_ts_low) {
   assert((column_family_id ==
           TablePropertiesCollectorFactory::Context::kUnknownColumnFamily) ==
          column_family_name.empty());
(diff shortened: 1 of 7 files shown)


db/flush_job.h
```diff
@@ -164,6 +164,8 @@ class FlushJob {
   IOStatus io_status_;
 
   const std::shared_ptr<IOTracer> io_tracer_;
+
+  const std::string full_history_ts_low_;
```
jay-zhuang, line 168: Just a question: why it's a `string` instead of `Slice`?

riversand963: A slice just points to some data that may or may not be owned. For flush job, it is better for the same FlushJob object to keep a copy of `full_history_ts_low_` so that it does not change throughout a flush. The cost of copy a small string is negligible compared to the cost of flush itself.

jay-zhuang (approved):
LGTM

db/flush_job_test.cc
```diff
@@ -114,9 +99,37 @@ class FlushJobTest : public testing::Test {
...
+    EXPECT_OK(env_->CreateDirIfMissing(dbname_));
+
+    // TODO(icanadi) Remove this once we mock out VersionSet
+    NewDB();
+
+    db_options_.env = env_;
+    db_options_.fs = fs_;
+    db_options_.db_paths.emplace_back(dbname_,
+                                      std::numeric_limits<uint64_t>::max());
+    db_options_.statistics = ROCKSDB_NAMESPACE::CreateDBStatistics();
```
ltamasi, line 112: Minor: `ROCKSDB_NAMESPACE` seems redundant here.

db/flush_job_test.cc
```diff
@@ -487,6 +507,137 @@ TEST_F(FlushJobTest, Snapshots) {
...
+      : FlushJobTestBase(test::PerThreadDBPath("flush_job_ts_gc_test"),
+                         test::ComparatorWithU64Ts()) {}
+
+  void AddKeyValueToMemtable(MemTable* memtable, Slice key, uint64_t ts,
+                             SequenceNumber seq, ValueType value_type,
+                             Slice value) {
+    std::string key_str(key.data(), key.size());
+    std::string ts_str;
+    PutFixed64(&ts_str, ts);
+    key_str.append(ts_str);
```
ltamasi, line 522: I think you could encode directly into `key_str` here (no need for a separate `ts_str`).

db/flush_job_test.cc
```diff
@@ -487,6 +507,137 @@ TEST_F(FlushJobTest, Snapshots) {
...
+    memtable->Add(seq, value_type, key_str, value);
+  }
+
+ protected:
+  static constexpr uint64_t kStartTs = 10;
+  static constexpr SequenceNumber kStartSeq = 0;
+  SequenceNumber curr_seq_{kStartSeq};
+  std::atomic<uint64_t> curr_ts_{kStartTs};
```
ltamasi, line 530: I'm wondering if this has to be atomic?

db/flush_job_test.cc
```diff
@@ -487,6 +507,137 @@ TEST_F(FlushJobTest, Snapshots) {
   job_context.Clean();
 }
 
+class FlushJobTimestampTest : public FlushJobTestBase {
+ public:
+  FlushJobTimestampTest()
+      : FlushJobTestBase(test::PerThreadDBPath("flush_job_ts_gc_test"),
+                         test::ComparatorWithU64Ts()) {}
+
+  void AddKeyValueToMemtable(MemTable* memtable, Slice key, uint64_t ts,
```
ltamasi, line 516: `test::EncodeInt` (which is used to produce `key` when calling this method) returns an `std::string`; switching from `Slice` to `string` for `key` would eliminate some back-and-forth conversions between the two.

db/flush_job_test.cc
```diff
@@ -487,6 +507,137 @@ TEST_F(FlushJobTest, Snapshots) {
...
+};
+
+TEST_F(FlushJobTimestampTest, AllKeysExpired) {
+  ColumnFamilyData* cfd = versions_->GetColumnFamilySet()->GetDefault();
+  autovector<MemTable*> to_delete;
+
+  {
+    MemTable* new_mem = cfd->ConstructNewMemtable(
+        *cfd->GetLatestMutableCFOptions(), kMaxSequenceNumber);
+    new_mem->Ref();
```
ltamasi, line 540: Do we need to `Unref` / `delete` this?

db/flush_job_test.cc
```diff
@@ -487,6 +507,137 @@ TEST_F(FlushJobTest, Snapshots) {
...
+    }
+    uint64_t ts = curr_ts_.fetch_add(1);
+    SequenceNumber seq = (curr_seq_++);
+    AddKeyValueToMemtable(new_mem, test::EncodeInt(0), ts, seq,
+                          ValueType::kTypeDeletionWithTimestamp, "");
+    cfd->imm()->Add(new_mem, &to_delete);
+  }
+
+  std::vector<SequenceNumber> snapshots;
+  SnapshotChecker* const snapshot_checker = nullptr;
```
ltamasi, line 555: This could be `constexpr SnapshotChecker* snapshot_checker = nullptr;` I think.

db/builder.cc
```diff
@@ -175,12 +175,17 @@ Status BuildTable(
                                   blob_file_additions)
             : nullptr);
 
+    std::shared_ptr<Logger> info_log;
```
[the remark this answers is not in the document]
riversand963: We did log using `db_options_.info_log`. Maybe should pass that to `BuildTable`?

db/flush_job_test.cc
riversand963: Yes it does. I can remove it.

db/flush_job_test.cc
riversand963: It does not. Making it atomic is to simulate the fact that multiple threads can get time without acquiring mutex, which is unlike RocksDB's internal sequence number.

db/flush_job_test.cc
riversand963: We do not have to do it here. `MemTableListVersion::Remove()` unrefs the memtable, and `JocContext::Clean()` deletes it.

db/builder.cc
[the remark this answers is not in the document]
ltamasi: What I mean is that weirdly enough, it seems we haven't passed the logger to `CompactionIterator` historically (i.e. your code, which passes an empty `shared_ptr`, is functionally equivalent to the checked-in code). But yeah, it might be a good idea to improve this and start passing `db_options_.info_log` like you suggested.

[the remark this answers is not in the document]
riversand963: I find it weird too. Let me fix it.

ltamasi (approved):

PR #7658, opened 2020-11-11
Experimental (production candidate) SST schema for Ribbon filter

pdillinger:
Summary: Added experimental public API for Ribbon filter:
NewExperimentalRibbonFilterPolicy(). This experimental API will
take a "Bloom equivalent" bits per key, and configure the Ribbon
filter for the same FP rate as Bloom would have but ~30% space
savings. (Note: optimize_filters_for_memory is not yet implemented
for Ribbon filter. That can be added with no effect on schema.)

Internally, the Ribbon filter is configured using a "one_in_fp_rate"
value, which is 1 over desired FP rate. For example, use 100 for 1%
FP rate. I'm expecting this will be used in the future for configuring
Bloom-like filters, as I expect people to more commonly hold constant
the filter accuracy and change the space vs. time trade-off, rather than
hold constant the space (per key) and change the accuracy vs. time
trade-off, though we might make that available.

### Benchmarking

```
$ ./filter_bench -impl=2 -quick -m_keys_total_max=200 -average_keys_per_filter=100000 -net_includes_hashing
Building...
Build avg ns/key: 34.1341
Number of filters: 1993
Total size (MB): 238.488
Reported total allocated memory (MB): 262.875
Reported internal fragmentation: 10.2255%
Bits/key stored: 10.0029
----------------------------
Mixed inside/outside queries...
  Single filter net ns/op: 18.7508
  Random filter net ns/op: 258.246
    Average FP rate %: 0.968672
----------------------------
Done. (For more info, run with -legend or -help.)
$ ./filter_bench -impl=3 -quick -m_keys_total_max=200 -average_keys_per_filter=100000 -net_includes_hashing
Building...
Build avg ns/key: 130.851
Number of filters: 1993
Total size (MB): 168.166
Reported total allocated memory (MB): 183.211
Reported internal fragmentation: 8.94626%
Bits/key stored: 7.05341
----------------------------
Mixed inside/outside queries...
  Single filter net ns/op: 58.4523
  Random filter net ns/op: 363.717
    Average FP rate %: 0.952978
----------------------------
Done. (For more info, run with -legend or -help.)
```

168.166 / 238.488 = 0.705  -> 29.5% space reduction

130.851 / 34.1341 = 3.83x construction time for this Ribbon filter vs. lastest Bloom filter (could make that as little as about 2.5x for less space reduction)

### Working around a hashing "flaw"

bloom_test discovered a flaw in the simple hashing applied in
StandardHasher when num_starts == 1 (num_slots == 128), showing an
excessively high FP rate.  The problem is that when many entries, on the
order of number of hash bits or kCoeffBits, are associated with the same
start location, the correlation between the CoeffRow and ResultRow (for
efficiency) can lead to a solution that is "universal," or nearly so, for
entries mapping to that start location. (Normally, variance in start
location breaks the effective association between CoeffRow and
ResultRow; the same value for CoeffRow is effectively different if start
locations are different.) Without kUseSmash and with num_starts > 1 (thus
num_starts ~= num_slots), this flaw should be completely irrelevant.  Even
with 10M slots, the chances of a single slot having just 16 (or more)
entries map to it--not enough to cause an FP problem, which would be local
to that slot if it happened--is 1 in millions. This spreadsheet formula
shows that: =1/(10000000*(1 - POISSON(15, 1, TRUE)))

As kUseSmash==false (the setting for Standard128RibbonBitsBuilder) is
intended for CPU efficiency of filters with many more entries/slots than
kCoeffBits, a very reasonable work-around is to disallow num_starts==1
when !kUseSmash, by making the minimum non-zero number of slots
2*kCoeffBits. This is the work-around I've applied. This also means that
the new Ribbon filter schema (Standard128RibbonBitsBuilder) is not
space-efficient for less than a few hundred entries. Because of this, I
have made it fall back on constructing a Bloom filter, under existing
schema, when that is more space efficient for small filters. (We can
change this in the future if we want.)

TODO: better unit tests for this case in ribbon_test, and probably
update StandardHasher for kUseSmash case so that it can scale nicely to
small filters.

### Other related changes

* Add Ribbon filter to stress/crash test
* Add Ribbon filter to filter_bench as -impl=3
* Add option string support, as in "filter_policy=experimental_ribbon:5.678;"
where 5.678 is the Bloom equivalent bits per key.
* Rename internal mode BloomFilterPolicy::kAuto to kAutoBloom
* Add a general BuiltinFilterBitsBuilder::CalculateNumEntry based on
binary searching CalculateSpace (inefficient), so that subclasses
(especially experimental ones) don't have to provide an efficient
implementation inverting CalculateSpace.
* Minor refactor FastLocalBloomBitsBuilder for new base class
XXH3pFilterBitsBuilder shared with new Standard128RibbonBitsBuilder,
which allows the latter to fall back on Bloom construction in some
extreme cases.
* Mostly updated bloom_test for Ribbon filter, though a test like
FullBloomTest::Schema is a next TODO to ensure schema stability
(in case this becomes production-ready schema as it is).
* Add some APIs to ribbon_impl.h for configuring Ribbon filters.
Although these are reasonably covered by bloom_test, TODO more unit
tests in ribbon_test
* Added a "tool" FindOccupancyForSuccessRate to ribbon_test to get data
for constructing the linear approximations in GetNumSlotsFor95PctSuccess.

Test Plan: Some unit tests updated but other testing is left TODO. This
is considered experimental but laying down schema compatibility as early
as possible in case it proves production-quality. Also tested in
stress/crash test.


include/rocksdb/filter_policy.h
```diff
@@ -212,4 +212,15 @@ class FilterPolicy {
...
+
+// An EXPERIMENTAL new Bloom alternative that saves about 30% space
+// compared to Bloom filters, with about 3-4x construction time and
+// similar query times. For example, if you pass in 10 for
+// bloom_equivalent_bits_per_key, you'll get the same 0.95% FP rate
+// as Bloom filter but only using about 7 bits per key. (This
+// way of configuring the new filter is considered experimental
+// and/or transitional, so is expected to go away.)
+extern const FilterPolicy* NewExperimentalRibbonFilterPolicy(
+    double bloom_equivalent_bits_per_key);
```
jay-zhuang, line 224: The interface exposes the filter size (bits_per_key), but also uses `equivalent` of fp to bloom, which I feel might confuse the user.
Maybe we should switch to using fp as interface for long-term, but stick with filter size (bits_per_key) for now, and have a helper function/interface to bridge them?

table/block_based/filter_policy.cc
```diff
@@ -341,6 +387,213 @@ class FastLocalBloomBitsReader : public FilterBitsReader {
...
+  // Save a conditional in Ribbon queries
+  static constexpr bool kAllowZeroStarts = false;
+};
+
+using Standard128RibbonTypesAndSettings =
+    ribbon::StandardRehasherAdapter<Standard128RibbonRehasherTypesAndSettings>;
+
+class Standard128RibbonBitsBuilder : public XXH3pFilterBitsBuilder {
+ public:
+  explicit Standard128RibbonBitsBuilder(double desired_one_in_fp_rate,
```
jay-zhuang, line 415: just question about code style, do we need to add `explicit` even the constructor takes more than 1 parameter?

table/block_based/filter_policy.cc
```diff
@@ -341,6 +387,213 @@ class FastLocalBloomBitsReader : public FilterBitsReader {
...
+      entropy = Lower32of64(hash_entries_.front());
+    }
+    size_t len = SolnType::GetBytesForOneInFpRate(
+        num_slots, desired_one_in_fp_rate_, /*rounding*/ entropy);
+    size_t len_with_metadata = len + 5;
+
+    // Use Bloom filter when it's better for small filters
+    if (num_slots < 1024 && bloom_fallback_.CalculateSpace(static_cast<int>(
+                                num_entries)) < len_with_metadata) {
+      SwapEntriesWith(&bloom_fallback_);
```
jay-zhuang, line 459: Is it worth adding a statistic ticker for the fall back?

jay-zhuang (approved):

table/block_based/filter_policy.cc
pdillinger: I don't think so. For small filters, it's perfectly fine to use Bloom. These aren't the cases we're really worried about optimizing.

For the other two fallback cases, I consider it "bad" or "rare" enough to log a warning for, each time it might happen. We have an internal tool for tracking these. In theory we should only run out of hash seeds with probability 0.05^256, less than one chance in a googol googol googol. In theory.

table/block_based/filter_policy_internal.h
```diff
@@ -127,6 +137,10 @@ class BloomFilterPolicy : public FilterPolicy {
   // behavior with format_version < 5 just in case.)
   int whole_bits_per_key_;
 
+  // For configuring Ribbon filter: a desired value for 1/fp_rate. For
+  // example, 100 -> 1% fp rate.
+  double desired_one_in_fp_rate_;
```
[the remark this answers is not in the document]
pdillinger: My hypothesis is that it's easier to configure things when you can use whole numbers for most common choices, with pretty good granularity. Think powers of 2 or powers of 10 up to about 1000. Beyond roughly 1000, numbers start to get uncomfortable to read and write, and those settings should not be needed for most applications anyway.

And I don't like specifying things as percentages. When someone writes 0.01, did they mean 1% or 1 in 10,000?

I'm reluctant to emphasize bits per key because that can have a complicated relationship with FP rate now that we are moving beyond Bloom and Ribbon128 does not scale infinitely in that way as Bloom does. But maybe people do care about that, in terms of how many bits of memory do I need to dedicate for (more or less) each key in my database.

In any case, this part is relatively hidden and can be changed / updated.

include/rocksdb/filter_policy.h
pdillinger: The complication with Ribbon is that if you pin the bits per key, your FP rate varies a little bit based on number of keys in the filter. If you pin the FP rate (what I've done here with Bloom equiv), your Ribbon bits per key varies a little bit based on number of keys in the filter. And consequently, there's no easy translation between the two.

Though that's assuming a relatively fixed construction success probability and (thus) construction CPU overhead. If you let that be more flexible, you *could* have a fixed relationship between FP rate and Ribbon bits per key, but then you'd "pay extra" for scaling to large number of keys per filter. You already have to pay for larger working size (just like building a hash table); you don't want to compound that with lowered construction success probability. (In fact, in this implementation I slightly favor higher-than-advertised construction success probability for large number of keys per filter.)

table/block_based/filter_policy.cc
pdillinger: Probably left over from when a constructor had just one parameter, was updated (by me) without removing 'explicit' and then I copy-pasted . I'll fix it. :)

pdillinger: Actually, it looks like it could theoretically matter for more than 1 param. I'll leave it. :)

https://en.cppreference.com/w/cpp/language/explicit

table/block_based/filter_policy_internal.h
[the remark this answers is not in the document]
jay-zhuang: at least for public interface, we don't have `_one_in_`, but already has some `_ratio` like: https://github.com/facebook/rocksdb/blob/bcba372352644c76267f4da8c0eb373682a17472/include/rocksdb/advanced_options.h#L298-L306

PR #7659, opened 2020-11-11
Hack to load OPTIONS file for read_amp_bytes_per_bit

riversand963:
A temporary hack to work around a bug in 6.10, 6.11, 6.12, 6.13 and
6.14. The bug will write out 8 bytes to OPTIONS file from the starting
address of BlockBasedTableOptions.read_amp_bytes_per_bit which is
actually a uint32. Consequently, the value of read_amp_bytes_per_bit
written in the OPTIONS file is wrong. From 6.15, RocksDB will
try to parse the read_amp_bytes_per_bit from OPTIONS file as a uint32.
To be able to load OPTIONS file generated by affected releases before
the fix, we need to manually parse read_amp_bytes_per_bit with this hack.

Test plan:
Generate a db with current 6.14.fb (head at https://github.com/facebook/rocksdb/commit/b6db05dbb5364c658c5401a8078d73697bb5f31d). Maybe use db_stress.

Checkout this PR, run
```
 ~/rocksdb/ldb --db=. --try_load_options --ignore_unknown_options idump --count_only
```
Expect success, and should not see
```
Failed: Invalid argument: Error parsing read_amp_bytes_per_bit:17179869184
```

Also
make check


pdillinger (changes requested):
I was imagining this instead, before call to ParseOptionHelper:

    } else if (opt_name == "read_amp_bytes_per_bit") {
      // HACK working around a bug (insert more details)
      assert(type_ == OptionType::kUInt32T);
      *reinterpret_cast<unsigned int*>(opt_addr) =
          static_cast<uint32_t>(ParseUint64(opt_value));
      return Status::OK();

So as not to involve exceptions, and allow truly bad input to still fail as before.

And shouldn't it be pretty easy to add a unit test for this? (Set the option to big number that is zero in lower 32 bits, check that parses as zero.)

riversand963:

options/options_helper.cc
```diff
@@ -969,6 +969,20 @@ Status OptionTypeInfo::Parse(const ConfigOptions& config_options,
...
+      // address of BlockBasedTableOptions.read_amp_bytes_per_bit which is
+      // actually a uint32. Consequently, the value of read_amp_bytes_per_bit
+      // written in the OPTIONS file is wrong. From 6.15, RocksDB will
+      // try to parse the read_amp_bytes_per_bit from OPTIONS file as a uint32.
+      // To be able to load OPTIONS file generated by affected releases before
+      // the fix, we need to manually parse read_amp_bytes_per_bit with this
+      // special hack.
+      assert(type_ == OptionType::kUInt32T);
+      uint64_t read_amp_bytes_per_bit = ParseUint64(opt_value);
+      EncodeFixed32(opt_addr, static_cast<uint32_t>(read_amp_bytes_per_bit));
```
pdillinger, line 984: EncodeFixed32 always uses little endian. We need to respect machine order in loading the value for use (in case someone in the world uses RocksDB with big endian).

pdillinger (changes requested):
Thanks for the unit test!

options/options_helper.cc
riversand963: I think it handles both little and big endian: https://github.com/facebook/rocksdb/blob/master/util/coding_lean.h#L34

mrambacher, line 984: I too am confused as to the need for the EncodeFixed32.  In my experiment here, the following worked as well:

 *addr = static_cast<uint32_t>(ParseUint64(value));

options/options_helper.cc
```diff
@@ -969,6 +969,20 @@ Status OptionTypeInfo::Parse(const ConfigOptions& config_options,
       ConfigOptions copy = config_options;
       copy.invoke_prepare_options = false;
       return parse_func_(copy, opt_name, opt_value, opt_addr);
+    } else if (opt_name == "read_amp_bytes_per_bit") {
+      // A temporary hack to work around a bug in 6.10, 6.11, 6.12, 6.13 and
+      // 6.14. The bug will write out 8 bytes to OPTIONS file from the starting
```
mrambacher, line 974: Can we tone down the comment to just say "A workaround to fix a bug in earlier releases where 8 bytes...

options/options_test.cc
```diff
@@ -842,7 +842,12 @@ TEST_F(OptionsTest, GetBlockBasedTableOptionsFromString) {
       "block_cache=1M;block_cache_compressed=1k;block_size=1024;"
       "block_size_deviation=8;block_restart_interval=4;"
       "format_version=5;whole_key_filtering=1;"
-      "filter_policy=bloomfilter:4.567:false;",
+      "filter_policy=bloomfilter:4.567:false;"
+      // A test for a hack. A bug caused read_amp_bytes_per_bit to be a large
+      // integer in OPTIONS file generated by 6.10 to 6.14. Though bug is
```
mrambacher, line 847: Get rid of "A test for a hack" to make the comment less incinderary.

mrambacher (approved):
If the code is moved out of options_helper, it LGTM.

I wonder if there is some more generic test we should be writing somehow to test that the type of the value (uint32) matches the type of the option (was size_t before the recent fix).  If you have ideas on how to do this validation, LMK and we can work on it in a separate PR.

options/options_helper.cc
riversand963: The `addr` or `opt_addr` is of type `char*`, to correctly store an integer at this address, we need to consider endianness.

riversand963:
Thanks @mrambacher for the review.

> 
> I wonder if there is some more generic test we should be writing somehow to test that the type of the value (uint32) matches the type of the option (was size_t before the recent fix). If you have ideas on how to do this validation, LMK and we can work on it in a separate PR.
>
I do not have a concrete plan yet. Since we can optionally build RocksDB with RTTI, I wonder whether there is anything we can do with this useful runtime type information.

options/options_helper.cc
pdillinger: No, this code is wrong. If you were correct, the entirety of ParseOptionHelper would be wrong. A function like ParseUInt64 returns a value in native endianness. It needs to be stored in the struct in native endianness. EncodeFixed32 stores it in the struct in little endian, including on big endian machines.

PR #7660, opened 2020-11-12
Track WAL in MANIFEST: Update logic for computing min_log_number_to_keep in atomic flush

cheng-chang:
The logic for computing min_log_number_to_keep in atomic flush was incorrect.

For example, when all column families are flushed, the min_log_number_to_keep should be the latest new log. But the incorrect logic calls `PrecomputeMinLogNumberToKeepNon2PC` for each column family, and returns the minimum of them. However, `PrecomputeMinLogNumberToKeepNon2PC(cf)` assumes column families other than `cf` are unflushed, but in case all column families are flushed, this assumption is incorrect.

Without this fix, the WAL referenced by the computed min_log_number_to_keep may actually contain no unflushed data, so the WAL might have actually been deleted from disk on recovery, then an incorrect error `Corruption: missing WAL` will be reported during the recovery.

Test Plan:
run `make crash_test_with_atomic_flush`  on devserver
added a unit test in `db_flush_test`


riversand963:
Thanks @cheng-chang for the fix.
Can we stress test this PR on a private branch for 3 continuous successful runs of `make crash_test`?
In the meantime, can we disable wal tracking in db_stress to collect signals for the rest of the code?

riversand963:
Can the PR description be more specific so that it is clearer to tell which part is incorrect? Otherwise people will have to look at the diff view which lacks some context.

db/db_impl/db_impl_files.cc
```diff
@@ -710,6 +710,42 @@ uint64_t PrecomputeMinLogNumberToKeepNon2PC(
...
 
+uint64_t PrecomputeMinLogNumberToKeepNon2PC(
+    VersionSet* vset, const autovector<ColumnFamilyData*>& cfds_to_flush,
+    const autovector<autovector<VersionEdit*>>& edit_lists) {
+  assert(vset != nullptr);
+  assert(!cfds_to_flush.empty());
+  assert(cfds_to_flush.size() == edit_lists.size());
+
+  const uint64_t kUint64Max = std::numeric_limits<uint64_t>::max();
+  uint64_t min_log_flushed = kUint64Max;
```
riversand963, line 721: How about `min_log_number_to_keep`?

db/version_set.h
```diff
@@ -1136,6 +1137,21 @@ class VersionSet {
     }
     return min_log_num;
   }
+  uint64_t PreComputeMinLogNumberWithUnflushedData(
+      const std::unordered_set<const ColumnFamilyData*>& cfds_to_skip) const {
+    uint64_t min_log_num = std::numeric_limits<uint64_t>::max();
```
riversand963, line 1142: Nit: `port::kMaxUint64`

riversand963 (commented):
Can you add a unit test, too?

cheng-chang:
@riversand963 crash_test_with_atomic_flush passed on devserver.

cheng-chang:
> Can the PR description be more specific so that it is clearer to tell which part is incorrect? Otherwise people will have to look at the diff view which lacks some context.

hard to explain in plain text, it's easier to understand the code.

cheng-chang:
@riversand963 added a unit test in db_flush_test

riversand963:
> > Can the PR description be more specific so that it is clearer to tell which part is incorrect? Otherwise people will have to look at the diff view which lacks some context.
> 
> hard to explain in plain text, it's easier to understand the code.

The purpose of this PR is to fix a bug in WAL tracking, right? We can at least describe what the bug is and what can trigger the bug. It does not have to include all the details, but just "incorrect" sounds too sketchy and ambiguous.

cheng-chang:
> > > Can the PR description be more specific so that it is clearer to tell which part is incorrect? Otherwise people will have to look at the diff view which lacks some context.
> > 
> > 
> > hard to explain in plain text, it's easier to understand the code.
> 
> The purpose of this PR is to fix a bug in WAL tracking, right? We can at least describe what the bug is and what can trigger the bug. It does not have to include all the details, but just "incorrect" sounds too sketchy and ambiguous.

I updated the description, hope it's clearer.

db/db_impl/db_impl.h
```diff
@@ -1017,6 +1017,15 @@ class DBImpl : public DB {
 
   VersionSet* TEST_GetVersionSet() const { return versions_.get(); }
 
+  uint64_t TEST_PrecomputeMinLogNumberToKeepNon2PC(
+      const autovector<ColumnFamilyData*>& cfds_to_flush,
+      const autovector<autovector<VersionEdit*>>& edit_lists);
+
+  uint64_t TEST_GetCurrentLogNumber() const {
+    InstrumentedMutexLock l(mutex());
+    return logs_.back().number;
```
riversand963, line 1026: Do we need to assert `!logs_.empty()`?

db/version_set.h
```diff
@@ -1136,6 +1137,21 @@ class VersionSet {
     }
     return min_log_num;
   }
+  uint64_t PreComputeMinLogNumberWithUnflushedData(
```
riversand963, line 1140: Similar to the comments in line 1123 and 1124, a short explanation of this method can be helpful.

db/db_flush_test.cc
```diff
@@ -646,6 +646,62 @@ TEST_P(DBAtomicFlushTest, ManualAtomicFlush) {
...
 }
 
+TEST_P(DBAtomicFlushTest, PrecomputeMinLogNumberToKeepNon2PC) {
+  Options options = CurrentOptions();
+  options.create_if_missing = true;
+  options.atomic_flush = GetParam();
+  options.write_buffer_size = (static_cast<size_t>(64) << 20);
+  CreateAndReopenWithCF({"pikachu"}, options);
+
+  const size_t num_cfs = handles_.size();
```
riversand963, line 656: Nit: maybe assert on the number of `handles_`.

db/db_impl/db_impl.h
```diff
@@ -1017,6 +1017,15 @@ class DBImpl : public DB {
 
   VersionSet* TEST_GetVersionSet() const { return versions_.get(); }
 
+  uint64_t TEST_PrecomputeMinLogNumberToKeepNon2PC(
```
riversand963, line 1020: Do we need this `TEST_`?

riversand963:
> I updated the description, hope it's clearer.
>
Better. It will still be good to describe what the application will see if the bug occurs. How it affects WAL tracking, etc.

db/db_impl/db_impl.h
cheng-chang: removed

cheng-chang:
@riversand963 PTAL, thanks

riversand963:
> However, `PrecomputeMinLogNumberToKeepNon2PC(cf)` assumes column families other than `cf` are flushed, but in case all column families are flushed, this assumption is incorrect.
>
Assumes column families other than `cf` are unflushed?

> Without this fix, the WAL referenced by the computed min_log_number_to_keep may actually contain no unflushed data, so the WAL might have actually been deleted from disk on recovery, then an incorrect error `Corruption: missing WAL` will be reported.
> 
Maybe it's clearer to say "will be reported during a subsequent recovery".

cheng-chang:
> > However, `PrecomputeMinLogNumberToKeepNon2PC(cf)` assumes column families other than `cf` are flushed, but in case all column families are flushed, this assumption is incorrect.
> 
> Assumes column families other than `cf` are unflushed?
> 
> > Without this fix, the WAL referenced by the computed min_log_number_to_keep may actually contain no unflushed data, so the WAL might have actually been deleted from disk on recovery, then an incorrect error `Corruption: missing WAL` will be reported.
> 
> Maybe it's clearer to say "will be reported during a subsequent recovery".

updated

riversand963:
Thanks @cheng-chang for the update.
Just to confirm, `min(PrecomputeMinLogNumberToKeepNon2PC(cfd) for cfd in cfds)` will be equal to `PrecomputeMinLogNumberToKeepNon2PC(cfds)` if `cfds` does **not** include **all** column families? Because if it does include all column families (flushed and unflushed), then the result will be bound by the minimum of `cfd->log_number` for all unflushed column families.

cheng-chang:
> Thanks @cheng-chang for the update.
> Just to confirm, `min(PrecomputeMinLogNumberToKeepNon2PC(cfd) for cfd in cfds)` will be equal to `PrecomputeMinLogNumberToKeepNon2PC(cfds)` if `cfds` does **not** include **all** column families? Because if it does include all column families (flushed and unflushed), then the result will be bound by the minimum of `cfd->log_number` for all unflushed column families.

Yes.

riversand963 (approved):

db/db_flush_test.cc
```diff
@@ -646,6 +646,63 @@ TEST_P(DBAtomicFlushTest, ManualAtomicFlush) {
...
+  ASSERT_EQ(num_cfs, 2);
+  WriteOptions wopts;
```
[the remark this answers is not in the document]
cheng-chang: can save unnecessary creation of WriteOptions, but compiler may optimize it even if calling WriteOptions() directly in `Put`.

PR #7661, opened 2020-11-12
Add options for integrated blob GC

ltamasi:
Summary:
This patch simply adds a couple of options that will enable users to
configure garbage collection when using the integrated BlobDB
implementation. The actual GC logic will be added in a separate step.

Test Plan:
`make check`

### DIFF
--- a/include/rocksdb/advanced_options.h
+++ b/include/rocksdb/advanced_options.h
@@ -734,7 +734,7 @@ struct AdvancedColumnFamilyOptions {
   // amplification for large-value use cases at the cost of introducing a level
   // of indirection for reads. See also the options min_blob_size,
   // blob_file_size, blob_compression_type, enable_blob_garbage_collection,
-  // and blob_garbage_collection_cutoff below.
+  // and blob_garbage_collection_age_cutoff below.
   //
   // Default: false
   //
@@ -779,7 +779,7 @@ struct AdvancedColumnFamilyOptions {
   // compaction. Valid blobs residing in blob files older than a cutoff get
   // relocated to new files as they are encountered during compaction, which
   // makes it possible to clean up blob files once they contain nothing but
-  // obsolete/garbage blobs. See also blob_garbage_collection_cutoff below.
+  // obsolete/garbage blobs. See also blob_garbage_collection_age_cutoff below.
   //
   // Default: false
   //
@@ -796,7 +796,7 @@ struct AdvancedColumnFamilyOptions {
   // Default: 0.25
   //
   // Dynamically changeable through the SetOptions() API
-  double blob_garbage_collection_cutoff = 0.25;
+  double blob_garbage_collection_age_cutoff = 0.25;
 
   // Create ColumnFamilyOptions with default values for all fields
   AdvancedColumnFamilyOptions();
--- a/include/rocksdb/advanced_options.h
+++ b/include/rocksdb/advanced_options.h
@@ -733,7 +733,8 @@ struct AdvancedColumnFamilyOptions {
   // only pointers to them are stored in SST files. This can reduce write
   // amplification for large-value use cases at the cost of introducing a level
   // of indirection for reads. See also the options min_blob_size,
-  // blob_file_size, and blob_compression_type below.
+  // blob_file_size, blob_compression_type, enable_blob_garbage_collection,
+  // and blob_garbage_collection_cutoff below.
   //
   // Default: false
   //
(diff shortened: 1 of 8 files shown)


include/rocksdb/advanced_options.h
```diff
@@ -773,6 +774,30 @@ struct AdvancedColumnFamilyOptions {
...
+  // Enables garbage collection of blobs. Blob GC is performed as part of
+  // compaction. Valid blobs residing in blob files older than a cutoff get
+  // relocated to new files as they are encountered during compaction, which
+  // makes it possible to clean up blob files once they contain nothing but
+  // obsolete/garbage blobs. See also blob_garbage_collection_cutoff below.
+  //
+  // Default: false
+  //
+  // Dynamically changeable through the SetOptions() API
+  bool enable_blob_garbage_collection = false;
```
riversand963, line 787: I remember there was a discussion about naming such options as adjectives, e.g. `blob_garbage_collection_enabled`?

include/rocksdb/advanced_options.h
```diff
@@ -773,6 +774,30 @@ struct AdvancedColumnFamilyOptions {
...
+  // The cutoff in terms of blob file age for garbage collection. Blobs in
+  // the oldest N blob files will be relocated when encountered during
+  // compaction, where N = garbage_collection_cutoff * number_of_blob_files.
+  // Note that enable_blob_garbage_collection has to be set in order for this
+  // option to have any effect.
+  //
+  // Default: 0.25
+  //
+  // Dynamically changeable through the SetOptions() API
+  double blob_garbage_collection_cutoff = 0.25;
```
riversand963, line 799: From the name, what `cutoff` actually refers to is kind-of unclear. Is it possible to give it a more explicit name?

include/rocksdb/advanced_options.h
riversand963: What is the expected behavior if this options is changed from true to false during a compaction that has already started?

include/rocksdb/advanced_options.h
riversand963: Same question, what is the expected behavior is application adjusts this cutoff in the middle of a compaction?

riversand963 (approved):
Mostly LGTM. Maybe also update `SanitizeOptions(const ImmutableDBOptions& db_options, const ColumnFamilyOptions& src)` when these options are actually ready for use?

options/cf_options.cc
```diff
@@ -408,6 +408,14 @@ static std::unordered_map<std::string, OptionTypeInfo>
          {offsetof(struct MutableCFOptions, blob_compression_type),
           OptionType::kCompressionType, OptionVerificationType::kNormal,
           OptionTypeFlags::kMutable}},
+        {"enable_blob_garbage_collection",
+         {offsetof(struct MutableCFOptions, enable_blob_garbage_collection),
```
mrambacher, line 412: First, would it be better to create a struct that held all of the Blob options in it, rather than adding them one at a time to the various options (ColumnFamilyOptions, MutableCFOptions).  

Second, "blob_garbage_collection_enabled" would be a better name.  That way all of the "blob" options start with "blob" and will be grouped together alphabetically.

mrambacher (commented):
I believe it would be beneficial to create a "BlobColumnFamilyOptions" struct to put all of these options in and manage them as a group.  It keeps them all together and may allow better "pluggability" and such in the future.  It might make it easier to turn on/off blob features on a column family (for example) or share blob settings between column families.

include/rocksdb/advanced_options.h
ltamasi: > I remember there was a discussion about naming such options as adjectives, e.g. `blob_garbage_collection_enabled`?

Hm, I'm not aware of that discussion but the prevailing convention is definitely the imperative style, e.g.:

```
options.h:  bool enable_thread_tracking = false;
options.h:  bool enable_pipelined_write = false;
options.h:  bool enable_write_thread_adaptive_yield = true;
table.h:  bool enable_index_compression = true;
options.h:  bool disable_auto_compactions = false;
options.h:  bool disableWAL;
```

I'd also rather keep it in sync with `enable_blob_files` (which would be a bit of a pain to change at this point since it's been part of the public API for some time).

ltamasi: > What is the expected behavior if this options is changed from true to false during a compaction that has already started?

Since `MutableCFOptions` is copied to a member in `Compaction`, in-progress compactions are never affected by `SetOptions`; they always use the same saved values throughout the compaction. This is true in general; I can call it out in a note for these options though.

include/rocksdb/advanced_options.h
ltamasi: > From the name, what `cutoff` actually refers to is kind-of unclear. Is it possible to give it a more explicit name?

How about putting "age" in there, i.e. `blob_garbage_collection_age_cutoff` ?

include/rocksdb/advanced_options.h
riversand963: > Since `MutableCFOptions` is copied to a member in `Compaction`
>
Thanks for clarification.

include/rocksdb/advanced_options.h
riversand963: I can't think of another, and `age` sounds good to me.

PR #7665, opened 2020-11-12
Update  option "allow_data_in_errors" in BuildOptions

akankshamahajan15:
Summary: "allow_data_in_errors" is not updated in BuildOptions. So it
would assume default value when BuildOptions is called.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks:

Tags:


zhichao-cao (approved):
LGTM, thanks!

PR #7667, opened 2020-11-12
Fail early when `merge_operator` not configured

ajkr:
An application may accidentally write merge operands without properly configuring `merge_operator`. We should alert them as early as possible that there's an API misuse. Previously RocksDB only notified them when a query or background operation needed to merge but couldn't. With this PR, RocksDB notifies them of the problem before applying the merge operand to the memtable (although it may already be in WAL, which seems it'd cause a crash loop until they enable `merge_operator`).


riversand963 (approved):

db/write_batch.cc
```diff
@@ -1715,6 +1715,10 @@ class MemTableInserter : public WriteBatch::Handler {
 
     MemTable* mem = cf_mems_->GetMemTable();
     auto* moptions = mem->GetImmutableMemTableOptions();
+    if (moptions->merge_operator == nullptr) {
```
mrambacher, line 1718: I think this is going to have problems with PermitUncheckedError

mrambacher (commented):
There seem to be other places in WriteBatch::Merge that could be failed if there is no MergeOperator.  Why only the one?

db/write_batch.cc
ajkr: Yes, the proposed refactoring in the other PR to `SeekToColumnFamily()` could help here to make it clearer whether `ret_status` was checked.

ajkr (commented):
> There seem to be other places in WriteBatch::Merge that could be failed if there is no MergeOperator. Why only the one?

I wish we could do the checks in `WriteBatch::Merge()`, but it doesn't always have access to `ColumnFamilyOptions` (the `ColumnFamilyHandle*` is allowed to be nullptr) and isn't called during recovery. Memtable insertion is a better place in terms of those considerations, although its downside is we have to enter read-only mode and we can't prevent the bad merge operand from being written to WAL.

anand1976 (approved):
LGTM

PR #7671, opened 2020-11-13
Add GetStringFromBlockBasedTableOptions APIs in convenience.h

akankshamahajan15:
Summary: Add GetStringFromBlockBasedTableOptions APIs to get string of
BlockBasedTableOptions. The behavior of these APIs is similar to
GetStringFromDBOptions and GetStringFromColumnFamilyOptions.

Test Plan: Add new unit test cases.

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/options/options_test.cc
+++ b/options/options_test.cc
@@ -876,6 +876,36 @@ TEST_F(OptionsTest, GetBlockBasedTableOptionsFromString) {
   // new_opt.read_amp_bytes_per_bit.
   EXPECT_EQ(1U, new_opt.read_amp_bytes_per_bit);
 
+  {
+    // Get string from block based table options.
+    std::string opt_str;
+    BlockBasedTableOptions opt_from_str;
+    ASSERT_OK(
+        GetStringFromBlockBasedTableOptions(config_options, new_opt, &opt_str));
+
+    // Check options from string opt_str is same as previous ones.
+    ASSERT_OK(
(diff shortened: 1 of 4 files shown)


options/options_test.cc
```diff
@@ -1320,6 +1350,32 @@ TEST_F(OptionsTest, ColumnFamilyOptionsSerialization) {
...
+
+  // Phase 3: Set new_options from the derived string
+  ASSERT_OK(GetBlockBasedTableOptionsFromString(
+      config_options, BlockBasedTableOptions(), base_options_file_content,
+      &new_options));
+
+  // Phase 4: new_options == base_options
+  ASSERT_OK(RocksDBOptionsParser::VerifyTableFactory(
+      config_options, NewBlockBasedTableFactory(base_options),
+      NewBlockBasedTableFactory(new_options)));
```
mrambacher, line 1376: I know it is a test, but this leaks memory...

mrambacher (commented):
Can you tell me the use case of these functions?   Given a TableFactory, you can already serialize it and parse its options.  These are really helper/convenience methods around that functionality, but seem only work for a very specific table.  I wonder if we go down this path we end up needing a lot more methods (GetStringFromPlainTableOptions et al), which is one reason I wrote the Configurable class...

akankshamahajan15:
Not needed anymore. Discussed with Mark, TableFactory::GetOptionString can be used to get the string of options.

PR #7672, opened 2020-11-13
Use default FileSystem in GenerateUniqueId

anand1976:
Use ```FileSystem::Default``` to read ```/proc/sys/kernel/uuid```, so it works for ```Envs``` with remote ```FileSystem``` as well.

### DIFF
--- a/env/env_posix.cc
+++ b/env/env_posix.cc
@@ -463,11 +463,12 @@ void PosixEnv::WaitForJoin() {
 
 std::string Env::GenerateUniqueId() {
   std::string uuid_file = "/proc/sys/kernel/random/uuid";
+  std::shared_ptr<FileSystem> fs = FileSystem::Default();
 
-  Status s = FileExists(uuid_file);
+  Status s = fs->FileExists(uuid_file, IOOptions(), nullptr);
   if (s.ok()) {
     std::string uuid;
-    s = ReadFileToString(this, uuid_file, &uuid);
+    s = ReadFileToString(fs.get(), uuid_file, &uuid);
     if (s.ok()) {
       return uuid;
     }


env/file_system.cc
```diff
@@ -83,6 +84,31 @@ FileOptions FileSystem::OptimizeForCompactionTableRead(
   return optimized_file_options;
 }
 
+std::string FileSystem::GenerateUniqueId(const IOOptions& opts,
+                                         IODebugContext* dbg) {
+  std::string uuid_file = "/proc/sys/kernel/random/uuid";
```
riversand963, line 89: How about Windows and MacOS?

riversand963 (commented):
Thanks @anand1976 for the PR.
I feel that generating unique id is a OS functionality which is not necessarily provided by file system. I am under the impression that the separation of Env and FileSystem aims to move file/storage related functionality to `FileSystem` while keeping the rest in `Env`.

env/file_system.cc
anand1976: Windows has its own implementation, and the implementation in this PR should work for MacOS I think? Its taken from ```Env::GenerateUniqueId```.

Regarding your comment about this being OS functionality, its a fair point. I noticed that ```env_librados.cc``` overrides it with its own implementation, but that also relies on system clock and there's nothing file system specific. So this could be considered as part of Env functionality.

riversand963: > Windows has its own implementation, and the implementation in this PR should work for MacOS I think? Its taken from `Env::GenerateUniqueId`.
> 
Verified on MacOS, and it does not work since `/proc/sys/kernel/random/uuid` does not exist (MacOS does not have /proc). This issue can be addressed in a separate PR. This PR itself LGTM.

riversand963 (approved):

env/file_system.cc
riversand963: Never mind, the function handles the case when the /proc is missing, thus should be fine.

PR #7675, opened 2020-11-14
add ArangoDB to USERS.md, and fix typos in that file

jsteemann:
Add ArangoDB to USERS.md.
We are using RocksDB since 2016.


ajkr (approved):
LGTM.

PR #7676, opened 2020-11-15
folly: Use SYS_futex for syscall

kraj:
glibc defines SYS_futex and on newer 32bit CPUs like RISCV-32, arc there
is no 32bit time_t therefore define SYS_futex in terms of SYS_futex_time64

Signed-off-by: Khem Raj <raj.khem@gmail.com>


ajkr (approved):
LGTM. Will a corresponding fix also be sent to upstream folly?

kraj:
> LGTM. Will a corresponding fix also be sent to upstream folly?

hmmm did not plan for that, but I think it makes sense, will follow up surely.

PR #7678, opened 2020-11-16
add find_prev_user_entry_time to PerfContext

liam-thunder:
add find_prev_user_entry_time to perf context, to track time cost in PrevInternal


chenyou-fdu:
not sure why circleci fails / timeout

PR #7679, opened 2020-11-16
Add further tests to ASSERT_STATUS_CHECKED (1)

adamretter:
First batch of adding more tests to ASSERT_STATUS_CHECKED.

* db_iterator_test
* db_memtable_test
* db_merge_operator_test
* db_merge_operand_test
* write_batch_test
* write_batch_with_index_test

### DIFF
--- a/db/db_iterator_test.cc
+++ b/db/db_iterator_test.cc
@@ -2274,6 +2274,7 @@ TEST_P(DBIteratorTest, Refresh) {
   ASSERT_OK(Put("x", "y"));
 
   std::unique_ptr<Iterator> iter(NewIterator(ReadOptions()));
+  ASSERT_OK(iter->status());
   iter->Seek(Slice("a"));
   ASSERT_TRUE(iter->Valid());
   ASSERT_EQ(iter->key().compare(Slice("x")), 0);
@@ -2288,6 +2289,7 @@ TEST_P(DBIteratorTest, Refresh) {
   iter->Next();
   ASSERT_FALSE(iter->Valid());
 
+  ASSERT_OK(iter->status());
   ASSERT_OK(iter->Refresh());
 
   iter->Seek(Slice("a"));
@@ -2312,6 +2314,7 @@ TEST_P(DBIteratorTest, Refresh) {
   iter->Next();
   ASSERT_FALSE(iter->Valid());
 
+  ASSERT_OK(iter->status());
   ASSERT_OK(iter->Refresh());
 
   iter->Seek(Slice("a"));
@@ -2335,6 +2338,7 @@ TEST_P(DBIteratorTest, RefreshWithSnapshot) {
   ReadOptions options;
   options.snapshot = snapshot;
   Iterator* iter = NewIterator(options);
+  ASSERT_OK(iter->status());
 
   iter->Seek(Slice("a"));
   ASSERT_TRUE(iter->Valid());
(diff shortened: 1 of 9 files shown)


mrambacher (approved):
LGTM!

pdillinger (changes requested):
ASC check failed in CI (both Travis & CircleCI - I didn't dig into details)

adamretter:
@mrambacher @pdillinger There is one failing test now in `db_merge_operator_test.cc` which is: ` MergeOperatorPinningTest/MergeOperatorPinningTest.TailingIterator/0`.

The problem is that this code:
```C++
    ReadOptions ro;
    ro.tailing = true;
    Iterator* iter = db_->NewIterator(ro);
```

Creates an `DBIter` which wraps a `ForwardIterator`, the implementation of `DBIter::status()` is then:

```
Status status() const override {
    if (status_.ok()) {
      return iter_.status();
    } else {
      assert(!valid_);
      return status_;
    }
  }
```

The problem in the test is that it is impossible to call `ASSERT_OK` on the `status_` field of `DBIter`.
I could do with some advice on how best to access the internal private state of `DBIter` from that test please...

mrambacher:
When the test fails, is iter_status() ok?  I think the preferred way of doing this would be in the Iterator constructor to call PermitUncheckedError on the status value.

adamretter:
@mrambacher yes, the status is OK.

So... I think U would need to do that for the status of both the wrapped `ForwardIterator` and the outer `DBIter`, as I can't rely on the status of the DBIter being OK. Does that sound reasonable still? Or is it better to expose some internal state to just the test... through an internal API only method, or some sort of friend class?

db/db_iterator_test.cc
```diff
@@ -2308,7 +2312,7 @@ TEST_P(DBIteratorTest, Refresh) {
   iter->Next();
   ASSERT_FALSE(iter->Valid());
 
-  iter->Refresh();
+  ASSERT_OK(iter->Refresh());
 
   iter->Seek(Slice("a"));
   ASSERT_TRUE(iter->Valid());
```
mrambacher, line 2318: Should there also be a check for iter->status being OK prior to the reset?

db/forward_iterator.cc
```diff
@@ -53,6 +53,12 @@ class ForwardLevelIterator : public InternalIterator {
     if (pinned_iters_mgr_ && pinned_iters_mgr_->PinningEnabled()) {
       pinned_iters_mgr_->PinIterator(file_iter_);
     } else {
+      /*
+      TODO(AR) needed for ASSERT_STATUS_CHECKED in
+      MergeOperatorPinningTest/MergeOperatorPinningTest.TailingIterator
+      */
+      status_.PermitUncheckedError();
+
```
mrambacher, line 61: The problem is that status_ is never set (or checked) by this test.  A better solution is to put the PermitUncheckedError into the constructor.  That way, if the status is ever set and not checked by another test, they would see the assert.

I believe that is safe to do (and probably the right thing) without the associated comment.

mrambacher (approved):
LGTM, with the moving of the PermitUncheckedError into the constructor...

pdillinger:
I did a merge with master and got many ASC failures. I should be available to try to land this quickly once it is back to working again.

pdillinger (changes requested):
I did a merge with master and got many ASC failures.

adamretter:
@pdillinger So was this introduced by merging onto master? What should I do next? Should I rebase this onto master and fix any resulting ASC fails?

pdillinger:
> @pdillinger So was this introduced by merging onto master? What should I do next? Should I rebase this onto master and fix any resulting ASC fails?

I don't know, but I don't attempt to land anything touching this many files with a merge base more than a few days old. (This one is two weeks old.) And it seems I can't push to this branch to run CI on the merge, so I'm trusting my local runs (both clang and gcc) until I see otherwise.

adamretter:
@pdillinger I rebased this onto master. I only saw one new failing ASC, so I fixed that. Is the PR good to go now?

pdillinger:
Thanks. I was running 'make check' which runs tests presumed to be broken, but I stand by my heuristic merge age limits. ;)

PR #7680, opened 2020-11-16
Fix the logic of setting read_amp_bytes_per_bit from OPTIONS file

riversand963:
Instead of using `EncodeFixed32` which always serialize a integer to
little endian, we should use the local machine's endianness when
populating a native data structure during options parsing.
Without this fix, `read_amp_bytes_per_bit` may be populated incorrectly
on big-endian machines.

Test Plan:
make check


HISTORY.md
```diff
@@ -1,4 +1,8 @@
 # Rocksdb Change Log
+## Unreleased
+### Bug Fixes
+* Fixed the logic of populating native data structure for `read_amp_bytes_per_bit` during OPTIONS file parsing.
```
pdillinger, line 4: Is this line referring to just this fix? If so, I think we can say it was a fix for big endian only (little endian not broken).

If it's referring to #7659 also, I think there should be more detail about what could go wrong.

riversand963: Updated.

pdillinger (approved):
Thanks!

#7659 seems like a significant bug deserving of description in HISTORY.md, which I don't see yet. If this is going to be cherry-picked for 6.15.0, can't we just put the HISTORY update under there?

riversand963:
Thanks @pdillinger for the review.
>a significant bug deserving of description in HISTORY.md, which I don't see yet.
>
Which bug are you referring to? Parsing and encoding `read_amp_bytes_per_bit` as uint64 or using `EncodeFixed32()` to set `BlockBasedTableOptions::read_amp_bytes_per_bit`? The former is in documented in the HISTORY.md of 6.10, ..., 6.14. 
I am thinking of cherry-picking this PR to 6.10, 6.11, ..., 6.15, and then updating the HISTORY.md of 6.15 to describe the fix combining #7659 and this PR.

PR #7683, opened 2020-11-16
RocksJava static lib dependencies should support MacOS 10.12+

adamretter:
Expands on https://github.com/facebook/rocksdb/pull/7016 so that when `PORTABLE=1` is set the dependencies for RocksJava static target will also be built with backwards compatibility for MacOS as far back as 10.12 (i.e. 2016).


pdillinger (approved):
LGTM

PR #7684, opened 2020-11-17
Make CompactRange and GetApproximateSizes work with timestamp

jay-zhuang:
Add timestamp to the `CompactRange()` and `GetApproximateSizes` range keys if needed.

Test plan: make check

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -3418,10 +3418,11 @@ Status DBImpl::GetApproximateSizes(const SizeApproximationOptions& options,
     // Add timestamp if needed
     std::string start_with_ts, limit_with_ts;
     if (ts_sz > 0) {
-      // Append a minimal timestamp as the range limit is exclusive:
+      // Maximum timestamp means including all key with any timestamp
+      AppendKeyWithMaxTimestamp(&start_with_ts, start, ts_sz);
+      // Append a maximum timestamp as the range limit is exclusive:
       // [start, limit)
-      AppendKeyWithMinTimestamp(&start_with_ts, start, ts_sz);
-      AppendKeyWithMinTimestamp(&limit_with_ts, limit, ts_sz);
+      AppendKeyWithMaxTimestamp(&limit_with_ts, limit, ts_sz);
       start = start_with_ts;
       limit = limit_with_ts;
     }
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -3397,20 +3397,26 @@ void DBImpl::GetApproximateMemTableStats(ColumnFamilyHandle* column_family,
 Status DBImpl::GetApproximateSizes(const SizeApproximationOptions& options,
                                    ColumnFamilyHandle* column_family,
                                    const Range* range, int n, uint64_t* sizes) {
-  auto ts_sz = column_family->GetComparator()->timestamp_size();
+  const Comparator* const ucmp = column_family->GetComparator();
+  assert(ucmp);
+  size_t ts_sz = ucmp->timestamp_size();
   if (ts_sz == 0) {
     return GetApproximateSizesInternal(options, column_family, range, n, sizes);
-  } else {
-    std::string start_str;
-    std::string limit_str;
-    AppendKeyWithMinTimestamp(&start_str, range->start, ts_sz);
+  }
+
+  std::string start_str[n];
+  std::string limit_str[n];
+  Range range_with_ts[n];
+  for (int i = 0; i < n; i++) {
+    AppendKeyWithMinTimestamp(&start_str[i], range[i].start, ts_sz);
     // Append a minimal timestamp as the range limit is exclusive:
     // [start, limit)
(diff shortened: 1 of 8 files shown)


riversand963:
Thanks @jay-zhuang for the PR.
I believe this is the first time we add support for timestamp to `CompactRange` and `GetApproxiamteSize`. Therefore, I suggest that we rename the PR to avoid confusion when we browse the change history in the future.

db/db_impl/db_impl.cc
```diff
@@ -3396,6 +3396,25 @@ void DBImpl::GetApproximateMemTableStats(ColumnFamilyHandle* column_family,
 Status DBImpl::GetApproximateSizes(const SizeApproximationOptions& options,
                                    ColumnFamilyHandle* column_family,
                                    const Range* range, int n, uint64_t* sizes) {
+  auto ts_sz = column_family->GetComparator()->timestamp_size();
+  if (ts_sz == 0) {
+    return GetApproximateSizesInternal(options, column_family, range, n, sizes);
+  } else {
+    std::string start_str;
+    std::string limit_str;
+    AppendKeyWithMinTimestamp(&start_str, range->start, ts_sz);
```
riversand963, line 3405: Argument `range` can actually be an array.

riversand963 (commented):
Thanks @jay-zhuang for the PR. Left some comments.

db/db_impl/db_impl.cc
```diff
@@ -3397,6 +3397,31 @@ void DBImpl::GetApproximateMemTableStats(ColumnFamilyHandle* column_family,
...
                                    ColumnFamilyHandle* column_family,
                                    const Range* range, int n, uint64_t* sizes) {
+  const Comparator* const ucmp = column_family->GetComparator();
+  assert(ucmp);
+  size_t ts_sz = ucmp->timestamp_size();
+  if (ts_sz == 0) {
+    return GetApproximateSizesInternal(options, column_family, range, n, sizes);
+  }
+
+  std::string start_str[n];
```
riversand963, line 3407: I am surprised that this line can compile now since `n` is not const...

riversand963: Though the strings themselves (may) reside on the heap, the `string` object containing pointers to the string are allocated on the stack. The size of available space on the stack for the thread is limited and can be small compared to the heap size. Therefore, if the user input `n` to this API is very large, there is a chance that `start_str`, `limit_str` and `range_with_ts` will overflow. How about just allocating from heap?

db/db_with_timestamp_basic_test.cc
```diff
@@ -179,6 +179,93 @@ class DBBasicTestWithTimestamp : public DBBasicTestWithTimestampBase {
...
+  options.comparator = &test_cmp;
+  DestroyAndReopen(options);
+
+  WriteOptions write_opts;
+  std::string ts_str = Timestamp(1, 0);
+  Slice ts = ts_str;
+  write_opts.timestamp = &ts;
+
+  ASSERT_OK(db_->Put(write_opts, "foo1", "bar"));
+  Flush();
```
riversand963, line 197: Not sure if you have updated this, but I think `ASSERT_OK(Flush())` is better.

riversand963 (commented):
Thanks @jay-zhuang for the PR. Left a few more comments, and the test failure seems worth investigation.

db/db_impl/db_impl.cc
```diff
@@ -3401,16 +3401,33 @@ Status DBImpl::GetApproximateSizes(const SizeApproximationOptions& options,
...
   for (int i = 0; i < n; i++) {
+    Slice start = range[i].start;
+    Slice limit = range[i].limit;
+
+    // Add timestamp if needed
+    std::string start_with_ts, limit_with_ts;
+    if (ts_sz > 0) {
+      // Append a minimal timestamp as the range limit is exclusive:
+      // [start, limit)
+      AppendKeyWithMinTimestamp(&start_with_ts, start, ts_sz);
```
riversand963, line 3423: Should this be `AppendKeyWithMaxTimestamp(&start_with_ts, start, ts_sz);` because we want to create a smallest internal key.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -781,7 +781,38 @@ void DBImpl::NotifyOnFlushCompleted(
...
+  if (ts_sz == 0) {
+    return CompactRangeInternal(options, column_family, begin_without_ts,
+                                end_without_ts);
+  }
+
+  std::string begin_str;
+  std::string end_str;
+
+  if (begin_without_ts != nullptr) {
+    AppendKeyWithMinTimestamp(&begin_str, *begin_without_ts, ts_sz);
```
riversand963, line 798: Ditto.

db/db_impl/db_impl.cc
jay-zhuang: Yeah, it's a little bit counterintuitive, I updated and added test for that.

riversand963 (approved):

PR #7685, opened 2020-11-17
Fuzzing RocksDB

cheng-chang:
This is the initial PR to support adding fuzz tests to RocksDB.
It includes the necessary build infrastructure, and includes an example fuzzer.
There is also a README serving as the tutorial for how to add more tests.

Test Plan:
Manually build and run the fuzz test according to README.


fuzz/sst_file_writer_fuzzer.cc
```diff
@@ -0,0 +1,101 @@
...
+#include "rocksdb/file_system.h"
+#include "rocksdb/sst_file_reader.h"
+#include "rocksdb/sst_file_writer.h"
+#include "src/libfuzzer/libfuzzer_macro.h"
+
+// Keys in SST file writer operations must be unique and in ascending order.
+// For each DBOperation generated by the fuzzer, this function is called on
+// it to deduplicate and sort the keys in the DBOperations.
+protobuf_mutator::libfuzzer::PostProcessorRegistration<DBOperations> reg = {
+    [](DBOperations* input, unsigned int seed) {
```
pdillinger, line 19: Nit: comment out name of unused parameter. (I'm surprised it's not reported by -Wall)

pdillinger (approved):
Awesome work! I agree that we probably don't need to compile the fuzzers with every top-level 'make check' but we should get them compiling in CircleCI soon (this PR or a later one) so that they aren't accidentally broken.

cheng-chang:
> Awesome work! I agree that we probably don't need to compile the fuzzers with every top-level 'make check' but we should get them compiling in CircleCI soon (this PR or a later one) so that they aren't accidentally broken.

I'll add a CircleCI test dedicated to compiling the fuzzers in a follow-up PR.

PR #7686, opened 2020-11-17
Handling misuse of snprintf return value

akankshamahajan15:
Summary: Handle misuse of snprintf return value to avoid Out of bound
read/write.

Test Plan: make check -j64

Reviewers:

Subscribers:

Tasks: T78794449

Tags:

### DIFF
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -2245,32 +2245,25 @@ Status BackupEngineImpl::BackupMeta::StoreToFile(bool sync) {
     return s;
   }
 
-  std::string buf;
-  buf += std::to_string(timestamp_);
-  buf += "\n";
-  buf += std::to_string(sequence_number_);
-  buf += "\n";
+  std::ostringstream buf;
+  buf << timestamp_ << "\n";
+  buf << sequence_number_ << "\n";
 
   if (!app_metadata_.empty()) {
     std::string hex_encoded_metadata =
         Slice(app_metadata_).ToString(/* hex */ true);
-    buf += kMetaDataPrefix.ToString();
-    buf += hex_encoded_metadata;
-    buf += "\n";
+    buf << kMetaDataPrefix.ToString() << hex_encoded_metadata << "\n";
   }
-  buf += std::to_string(files_.size());
-  buf += "\n";
+  buf << files_.size() << "\n";
 
   for (const auto& file : files_) {
     // use crc32c for now, switch to something else if needed
     // WART: The checksums are crc32c, not original crc32
-    buf += file->filename;
-    buf += " crc32 ";
-    buf += std::to_string(ChecksumHexToInt32(file->checksum_hex));
-    buf += "\n";
+    buf << file->filename << " crc32 " << ChecksumHexToInt32(file->checksum_hex)
+        << "\n";
   }
 
-  s = backup_meta_file->Append(Slice(buf));
+  s = backup_meta_file->Append(Slice(buf.str()));
   if (s.ok() && sync) {
     s = backup_meta_file->Sync();
   }
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -2248,8 +2248,14 @@ Status BackupEngineImpl::BackupMeta::StoreToFile(bool sync) {
   std::unique_ptr<char[]> buf(new char[max_backup_meta_file_size_]);
   size_t len = 0, buf_size = max_backup_meta_file_size_;
   len += snprintf(buf.get(), buf_size, "%" PRId64 "\n", timestamp_);
+  if (len >= buf_size) {
+    return Status::Corruption("Buffer too small to fit backup metadata");
+  }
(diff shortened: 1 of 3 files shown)


utilities/backupable/backupable_db.cc
```diff
@@ -2248,8 +2248,14 @@ Status BackupEngineImpl::BackupMeta::StoreToFile(bool sync) {
   std::unique_ptr<char[]> buf(new char[max_backup_meta_file_size_]);
   size_t len = 0, buf_size = max_backup_meta_file_size_;
   len += snprintf(buf.get(), buf_size, "%" PRId64 "\n", timestamp_);
+  if (len >= buf_size) {
+    return Status::Corruption("Buffer too small to fit backup metadata");
```
riversand963, line 2252: iiuc, we use `Corruption` to denote data corruption on db. This case does not seem to fall into that category.

akankshamahajan15: Line 2267 had ```return Status::Corruption("Buffer too small to fit backup metadata");``` .  So I added same message here as well as other places. Should I change the error Status to something else?

riversand963: I see. In this case, I think we can keep it for now until we revisit this later.

utilities/blob_db/blob_db_impl.cc
```diff
@@ -1703,15 +1708,31 @@ std::pair<bool, int64_t> BlobDBImpl::SanityCheck(bool aborted) {
       pos += snprintf(buf + pos, sizeof(buf) - pos,
                       ", expiration range (%" PRIu64 ", %" PRIu64 ")",
                       expiration_range.first, expiration_range.second);
+      if (pos >= static_cast<int>(sizeof(buf))) {
```
riversand963, line 1711: If the check on line 1695 is true, so must the check on line 1708.  Can we just log the same error message once?

utilities/blob_db/blob_db_impl.cc
```diff
@@ -1703,15 +1708,31 @@ std::pair<bool, int64_t> BlobDBImpl::SanityCheck(bool aborted) {
...
+                       "Buffer too small for sanity check.");
+      }
+
       if (!blob_file->Obsolete()) {
         pos += snprintf(buf + pos, sizeof(buf) - pos,
                         ", expire in %" PRIu64 " seconds",
                         expiration_range.second - now);
+        if (pos >= static_cast<int>(sizeof(buf))) {
+          pos = sizeof(buf);
+          ROCKS_LOG_WARN(db_options_.info_log,
```
riversand963, line 1723: Ditto.

utilities/blob_db/blob_db_impl.cc
```diff
@@ -1703,15 +1708,31 @@ std::pair<bool, int64_t> BlobDBImpl::SanityCheck(bool aborted) {
...
+                         "Buffer too small for sanity check.");
+        }
       }
     }
     if (blob_file->Obsolete()) {
       pos += snprintf(buf + pos, sizeof(buf) - pos, ", obsolete at %" PRIu64,
                       blob_file->GetObsoleteSequence());
+      if (pos >= static_cast<int>(sizeof(buf))) {
+        pos = sizeof(buf);
+        ROCKS_LOG_WARN(db_options_.info_log,
```
riversand963, line 1733: Same.

riversand963 (approved):

pdillinger:
Thanks for the fixes. Sorry I missed this PR being published. We should generally move away from static-sized buffers and snprintf for variable-size data. None of these places appear to be so performance critical that we can't use std::string.

akankshamahajan15:
> Thanks for the fixes. Sorry I missed this PR being published. We should generally move away from static-sized buffers and snprintf for variable-size data. None of these places appear to be so performance critical that we can't use std::string.

Make sense. I will make the changes to replace snprintf with std::string.

utilities/blob_db/blob_db_impl.cc
```diff
@@ -1686,35 +1686,34 @@ std::pair<bool, int64_t> BlobDBImpl::SanityCheck(bool aborted) {
...
     auto blob_file = blob_file_pair.second;
-    char buf[1000];
-    int pos = snprintf(buf, sizeof(buf),
-                       "Blob file %" PRIu64 ", size %" PRIu64
-                       ", blob count %" PRIu64 ", immutable %d",
-                       blob_file->BlobFileNumber(), blob_file->GetFileSize(),
-                       blob_file->BlobCount(), blob_file->Immutable());
+    std::string buf;
+
+    buf = buf + "Blob file " + std::to_string(blob_file->BlobFileNumber());
```
pdillinger, line 1691: Any reason not to use `+=`?

pdillinger (approved):
I find removing unnecessary complexity very satisfying. (I don't know about you.) :)

When I said `std::string` I probably should have mentioned `std::ostringstream` also. It should not practically matter in these cases, as we are not appending in a loop, but it might also simplify *slightly* by removing calls to to_string(). (Change to ostringstream at your discretion.)

PR #7690, opened 2020-11-18
Call out a bug in HISTORY

cheng-chang:
It's worth mentioning the corner case bug fixed in PR 7621.

Test Plan:
N/A


zhichao-cao (approved):
LGTM

PR #7693, opened 2020-11-18
Add kManifestWriteNoWAL to BackgroundErrorReason to handle Flush IO Error when WAL is disabled

zhichao-cao:
In the current code base, all the manifest writes with IO error will be set with reason: BackgroundErrorReason::kManifestWrite, which will be mapped to the kHardError if the IO Error is retryable. However, if the system does not use the WAL, all the retryable IO error should be mapped to kSoftError. Create this PR to handle is special case by adding kManifestWriteNoWAL to BackgroundErrorReason.

Test plan: make check, add new testing cases to error_handler_fs_test

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -254,8 +254,15 @@ Status DBImpl::FlushMemTableToOutputFile(
       // TODO: distinguish between MANIFEST write and CURRENT renaming
       if (!versions_->io_status().ok()) {
         // Should handle return error?
-        error_handler_.SetBGError(io_s, BackgroundErrorReason::kManifestWrite)
-            .PermitUncheckedError();
+        if (total_log_size_ > 0) {
+          // If the WAL is empty, we use different error reason
+          error_handler_.SetBGError(io_s, BackgroundErrorReason::kManifestWrite)
+              .PermitUncheckedError();
+        } else {
+          error_handler_
+              .SetBGError(io_s, BackgroundErrorReason::kManifestWriteNoWAL)
+              .PermitUncheckedError();
+        }
       } else if (total_log_size_ > 0) {
         // Should handle return error?
         error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush)
@@ -656,9 +663,16 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
       // be pessimistic and try write to a new MANIFEST.
       // TODO: distinguish between MANIFEST write and CURRENT renaming
       if (!versions_->io_status().ok()) {
-        // Should Handle this error?
-        error_handler_.SetBGError(io_s, BackgroundErrorReason::kManifestWrite)
-            .PermitUncheckedError();
+        // Should handle return error?
+        if (total_log_size_ > 0) {
+          // If the WAL is empty, we use different error reason
+          error_handler_.SetBGError(io_s, BackgroundErrorReason::kManifestWrite)
+              .PermitUncheckedError();
+        } else {
+          error_handler_
+              .SetBGError(io_s, BackgroundErrorReason::kManifestWriteNoWAL)
+              .PermitUncheckedError();
+        }
       } else if (total_log_size_ > 0) {
         // Should Handle this error?
         error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush)
--- a/db/error_handler.cc
+++ b/db/error_handler.cc
@@ -347,6 +347,22 @@ Status ErrorHandler::SetBGError(const Status& bg_err, BackgroundErrorReason reas
   return bg_error_;
 }
 
+// This is the main function for looking at IO related error during the
+// background operations. The main logic is:
+// 1) if the error is caused by data loss, the error is mapped to
+//    unrecoverable error. Application/user must take action to handle
+//    this situation.
+// 2) if the error is a Retryable IO error, auto resume will be called and the
+//    auto resume can be controlled by resume count and resume interval
+//    options. There are three sub-cases:
+//    a) if the error happens during compaction, it is mapped to a soft error.
+//       the compaction thread will reschedule a new compaction.
+//    b) if the error happens during flush and also WAL is empty, it is mapped
+//       to a soft error. Note that, it includes the case that IO error happens
+//       in SST or manifest write during flush.
+//    c) all other errors are mapped to hard error.
+// 3) for other cases, SetBGError(const Status& bg_err, BackgroundErrorReason
+//    reason) will be called to handle other error cases.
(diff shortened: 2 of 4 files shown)


anand1976 (approved):
Maybe add a comment "TODO" somewhere to simplify the ```BackgroundErrorReason``` by separating out the file type out of it.

zhichao-cao:
> Maybe add a comment "TODO" somewhere to simplify the `BackgroundErrorReason` by separating out the file type out of it.

Sure!

riversand963:
>However, if the system does not use the WAL, all the retryable IO error should be mapped to kSoftError. 
> 
Could you elaborate on this choice a little bit more?

zhichao-cao:
> > However, if the system does not use the WAL, all the retryable IO error should be mapped to kSoftError.
> 
> Could you elaborate on this choice a little bit more?

So, for the no WAL case, we do not need to worry about the consistency and correctness of the WAL. In flush, if it fails (SST file writes or Manifest writes) due to a retryable IO error, the expected solution is to redo the flush. Since, we do not have WAL, the KV-pairs can continue to be inserted to the memtable until the memory is used up. The difference is, if WAL is used and the retryable IO error happens, we must stop the insertion. Otherwise, if we cannot switch to a new WAL, there will be inconsistency between memtable and WAL. So the failure of switch to a new WAL can be caused by a bad manifest writes or WAL. So we map all the errors as hard error to make the DB into read mode.

riversand963:
>The difference is, if WAL is used and the retryable IO error happens, we must stop the insertion. Otherwise, if we cannot switch to a new WAL, there will be inconsistency between memtable and WAL. So the failure of switch to a new WAL can be > caused by a bad manifest writes or WAL. So we map all the errors as hard error to make the DB into read mode.
>
What is the scope of IO error in the new Env? If IO error occurs during writing/syncing MANIFEST file, does it mean an IO error will occur for the WAL?

zhichao-cao:
> What is the scope of IO error in the new Env? If IO error occurs during writing/syncing MANIFEST file, does it mean an IO error will occur for the WAL?

We cannot have the assumption that if Manifest writes has IO error, IO error will occur for WAL. The high level take is: if WAL is disabled, we assume that the data persistency is taken care by other mechanism or the application can tolerate the data loss under the retryable IO error during Flush and Compaction. So, we will NOT put the DB in to read only mode. Any write failure during Flush and Compaction is able to roll back to the state before flush or compaction. But for the Manifest write IO error happen in other conditions, we cannot ensure the correctness of the DB, that is why we still put in hard error.

riversand963:
> So the failure of switch to a new WAL can be caused by a bad manifest writes or WAL
>
I guess I am a little confused by the WAL switching part and why a MANIFEST write IO error can cause WAL switching error. The WAL switching caused by this flush should have already succeeded before putting the cfd onto the flush queue. Or are you referring to a future WAL switching?

Consider flush. If an IO error occurs in writing to MANIFEST file, the memtables will be kept. The WAL files containing the data of these memtables have also been closed in `SwitchMemtable()`, which has also succeeded. If we can tell that the WAL files are **correct** and accessible, can we just switch to a new MANIFEST and retry?

zhichao-cao:
> > So the failure of switch to a new WAL can be caused by a bad manifest writes or WAL
> 
> I guess I am a little confused by the WAL switching part and why a MANIFEST write IO error can cause WAL switching error. The WAL switching caused by this flush should have already succeeded before putting the cfd onto the flush queue. Or are you referring to a future WAL switching?
> 
> Consider flush. If an IO error occurs in writing to MANIFEST file, the memtables will be kept. The WAL files containing the data of these memtables have also been closed in `SwitchMemtable()`, which has also succeeded. If we can tell that the WAL files are **correct** and accessible, can we just switch to a new MANIFEST and retry?

Oh, sorry about the confusion part. Let me re-explain: 1) when WAL is enabled. a) if WAL write has a retryable IO Error, which means, the WAL file cannot be trusted, so we must switch to a new WAL file. To achieve this action, we need to stop the writes, seal the memtable, switch to a new WAL, and do flush. So WAL write IO Error will be mapped to hard error. b) if Manifest write error happens, yes, as you mentioned, if no other write error happens, we can just switch to a new Manifest and retry the manifest writes (this logic has not been implemented yet). c) any write error happens during flush will indicate it is SST write or Manifest write errors. To protect the data correctness, we need to make the DB into read only mode. d) since compaction does not related to the DB state, its write failure will let the compaction be reschedule by itself. So we all map it to soft error and does not stop the writes

2) when WAL is disabled. The only difference is, the DB writes can continue (does not have persistency requirements).

riversand963:
Thanks @zhichao-cao for the clarification.
It seems the cases you are trying to distinguish are
1) retryable WAL IO error followed by a MANIFEST write IO error. In this case, DB remains read-only.
2) no WAL IO error (or no WAL at all), followed by MANIFEST write IO error. In this case, switch to a new MANIFEST and retry

If so, I kind of feel the name `kManifestWriteNoWAL` vs. `kManifestWrite` a little confusing. I think using some kind of `ErrorContext` with related data members such as `has_wal_error` seems the direction to pursue before there are too many error reasons in `BackgroundErrorReason`.

A somewhat related question/issue.
In single-column family mode in which we do not sync WAL files during background flush, there are some writes with WAL enabled, but the most recent a few writes are without WAL. Then flush happens for the first time but a MANIFEST write occurs while trying to commit the flush result to the MANIFEST. In this case, there are WALs, and the change in this PR will still classify the error as hard error and put the db to read-only mode.

zhichao-cao:
> If so, I kind of feel the name `kManifestWriteNoWAL` vs. `kManifestWrite` a little confusing. I think using some kind of `ErrorContext` with related data members such as `has_wal_error` seems the direction to pursue before there are too many error reasons in `BackgroundErrorReason`.
> 

Yes, current is a temporal fix, in the next step, I need to refactor this portion. As indicated in the TODO ahead of the BackgroundErrorReason, I will add a context to indicate the other essential information such as the file type (e.g., Manifest, SST) and special context. `BackgroundErrorReason` is only used to indicate the senario that causes the error.

> A somewhat related question/issue.
> In single-column family mode in which we do not sync WAL files during background flush, there are some writes with WAL enabled, but the most recent a few writes are without WAL. Then flush happens for the first time but a MANIFEST write occurs while trying to commit the flush result to the MANIFEST. In this case, there are WALs, and the change in this PR will still classify the error as hard error and put the db to read-only mode.

I use `total_log_size_ `  to decide the WAL has been enabled or not. So you mean, if WAL is enabled, for the first time flush, WAL is empty and Manifest write IO error happens, (`total_log_size_ += log_entry.size();`) `total_log_size_ ` will not be 0 and yes, it will be map to hard error, which is expected.

zhichao-cao:
@riversand963 Thanks for the comments and review! The PR is updated based on the discussions.

PR #7696, opened 2020-11-18
Add Ribbon schema test to bloom_test

pdillinger:
Summary: These new unit tests should ensure that we don't accidentally
change the interpretation of bits for what I call Standard128Ribbon
filter internally, available publicly as NewExperimentalRibbonFilterPolicy.
There is very little intuitive reason for the values we check against in
these tests; I just plug in the right expected values upon watching the
test fail initially.

Most (but not all) of the tests are essentially "whitebox" "round-trip." We
create a filter from fixed keys, and first compare the checksum of those
filter bytes against a saved value. We also run queries against other fixed
keys, comparing which return false positives against a saved set.

Test Plan: test addition and refactoring only

### DIFF
--- a/util/bloom_test.cc
+++ b/util/bloom_test.cc
@@ -322,6 +322,13 @@ class FullBloomTest : public testing::TestWithParam<BloomFilterPolicy::Mode> {
     }
   }
 
+  int GetRibbonSeedFromFilterData() {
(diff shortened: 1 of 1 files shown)


jay-zhuang (approved):
LGTM.

PR #7697, opened 2020-11-19
Do not track empty WALs

cheng-chang:
An empty WAL won't be backed up by the BackupEngine. So if we track the empty WALs in MANIFEST, then when restoring from a backup, it may report corruption that the empty WAL is missing, which is correct because the WAL is actually in the main DB but not in the backup DB, but missing an empty WAL does not logically break DB consistency.

Test Plan:
watch existing tests to pass

### DIFF
--- a/include/rocksdb/utilities/backupable_db.h
+++ b/include/rocksdb/utilities/backupable_db.h
@@ -69,6 +69,7 @@ struct BackupableDBOptions {
   // If false, we won't backup log files. This option can be useful for backing
   // up in-memory databases where log file are persisted, but table files are in
   // memory.
+  // If track_and_verify_wals_in_manifest is true, this must be true.
   // Default: true
   bool backup_log_files;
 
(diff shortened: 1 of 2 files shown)


pdillinger (approved):
I don't know this area of code, but our crash test indicates empty WALs were not returned by GetSortedWalFiles, so were not backed up by BackupEngine, and those empty WALs were expected by the manifest check on restore. So I trust this will keep the empty WALs from being returned by GetSortedWalFiles

pdillinger:
And of course remember to re-import 😁

include/rocksdb/utilities/backupable_db.h
```diff
@@ -69,6 +69,7 @@ struct BackupableDBOptions {
   // If false, we won't backup log files. This option can be useful for backing
   // up in-memory databases where log file are persisted, but table files are in
   // memory.
+  // If track_and_verify_wals_in_manifest is true, this must be true.
```
riversand963, line 72: It would be nice to enforce this rule in addition to this comment.

riversand963 (approved):
Thanks @cheng-chang for the fix

PR #7699, opened 2020-11-19
Port corruption test to use custom env

riversand963:
Allow corruption_test to run on custom env loaded via
`Env::LoadEnv()`.

Test plan:
```
make corruption_test
./corruption_test
```

Also run on in-house custom env.

### DIFF
--- a/db/corruption_test.cc
+++ b/db/corruption_test.cc
@@ -21,6 +21,7 @@
 #include "db/version_set.h"
 #include "env/composite_env_wrapper.h"
 #include "file/filename.h"
+#include "port/stack_trace.h"
 #include "rocksdb/cache.h"
 #include "rocksdb/convenience.h"
 #include "rocksdb/db.h"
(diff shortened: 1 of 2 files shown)


db/corruption_test.cc
```diff
@@ -173,10 +188,48 @@ class CorruptionTest : public testing::Test {
     ASSERT_GE(max_expected, correct);
   }
 
+  void CorruptFile(const std::string& fname, int offset, int bytes_to_corrupt) {
```
zhichao-cao, line 191: This is copied from testutil.cc?

zhichao-cao: Where does this used?

riversand963: I guess this function was not there when I created this PR on my local repo last year...Let me remove it if possible.

riversand963: Weird, it is not used. Will remove it.

zhichao-cao (approved):
LGTM, thanks!

mrambacher (commented):
I would strongly recommend that you also add support for MEM_ENV:
if (test_env_uri) {
 ... 
} else if (getenv("MEM_ENV") != nullptr) {
  env_guard_.reset(new MockEnv(base_env));
  base_env = env_guard.get();
}

With support for MEM_ENV, you will likely find that there are places that are attempting an action against the database but using the wrong env.

One of these days I am going to write a helper to do all of those settings in one swoop...

riversand963:
> I would strongly recommend that you also add support for MEM_ENV:
>
SG. The purpose of this PR is to explain how to use `Env::LoadEnv`, thus it may be better just to keep it simple. We can add `MEM_ENV` in another PR.

PR #7700, opened 2020-11-19
Fix the thread wait case in error_handler

zhichao-cao:
In error_handler auto recovery case, if recovery_in_prog_ is false, the recover is finished or failed. In this case, the auto recovery thread should finish its execution so recovery_thread_ should be null. However, in some cases, it is not null, the caller should not directly returned. Instead, it should wait for a while and create a new thread to execute the new recovery.

test plan: make check, error_handler_fs_test

### DIFF
--- a/db/error_handler.cc
+++ b/db/error_handler.cc
@@ -584,10 +584,18 @@ Status ErrorHandler::StartRecoverFromRetryableBGIOError(IOStatus io_error) {
   if (bg_error_.ok() || io_error.ok()) {
     return Status::OK();
   }
-  if (db_options_.max_bgerror_resume_count <= 0 || recovery_in_prog_ ||
-      recovery_thread_) {
+  if (db_options_.max_bgerror_resume_count <= 0 || recovery_in_prog_) {
     // Auto resume BG error is not enabled, directly return bg_error_.
     return bg_error_;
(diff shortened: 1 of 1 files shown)


db/error_handler.cc
```diff
@@ -543,12 +543,20 @@ Status ErrorHandler::StartRecoverFromRetryableBGIOError(IOStatus io_error) {
...
     return Status::OK();
   }
-  if (db_options_.max_bgerror_resume_count <= 0 || recovery_in_prog_ ||
-      recovery_thread_) {
+  if (db_options_.max_bgerror_resume_count <= 0 || recovery_in_prog_) {
     // Auto resume BG error is not enabled, directly return bg_error_.
     return bg_error_;
   }
 
+  if (recovery_thread_) {
```
anand1976, line 551: Would it make sense to reset ```recovery_thread_``` around line 470, so that way its guaranteed to be null when this is called for a new error?

zhichao-cao: ClearBGError() is not used in the auto recovery thread and it is only used by the manual resume. If we resert recovery_thread_ in it, actually when next auto resume comes, the issue is still there.

anand1976 (approved):

PR #7701, opened 2020-11-19
Fix kPointInTimeRecovery handling of truncated WAL

ajkr:
WAL may be truncated to an incomplete record due to crash while writing
the last record or corruption. In the former case, no hole will be
produced since no ACK'd data was lost. In the latter case, a hole could
be produced without this PR since we proceeded to recover the next WAL
as if nothing happened. This PR changes the record reading code to
always report a corruption for incomplete records in
`kPointInTimeRecovery` mode, and the upper layer will only ignore them
if the next WAL has consecutive seqnum (i.e., we are guaranteed no
hole).

While this solves the hole problem for the case of incomplete
records, the possibility is still there if the WAL is corrupted by
truncation to an exact record boundary.

Test Plan: Interestingly there already was a test for this case
(`DBWALTestWithParams.kPointInTimeRecovery`); it just had a typo bug in
the verification that prevented it from noticing holes in recovery.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,10 @@
 ## Unreleased
 ### Behavior Changes
 * Attempting to write a merge operand without explicitly configuring `merge_operator` now fails immediately, causing the DB to enter read-only mode. Previously, failure was deferred until the `merge_operator` was needed by a user read or a background operation.
+* Users of `WALRecoveryMode::kPointInTimeRecovery` whose DB contains a mix of writes with/without `WriteOptions::disableWAL` may now see recovery terminate at an earlier point in time after a crash.
+
+### Bug Fixes
+* Truncated WALs ending in incomplete records can no longer produce gaps in the recovered data when `WALRecoveryMode::kPointInTimeRecovery` is used. Gaps are still possible when WALs are truncated exactly on record boundaries; for complete protection, users should enable `track_and_verify_wals_in_manifest`.
 
 ### Bug Fixes
 * Fixed the logic of populating native data structure for `read_amp_bytes_per_bit` during OPTIONS file parsing on big-endian architecture. Without this fix, original code introduced in PR7659, when running on big-endian machine, can mistakenly store read_amp_bytes_per_bit (an uint32) in little endian format. Future access to `read_amp_bytes_per_bit` will give wrong values. Little endian architecture is not affected.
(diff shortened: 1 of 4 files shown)


HISTORY.md
```diff
@@ -2,6 +2,10 @@
 ## Unreleased
 ### Behavior Changes
 * Attempting to write a merge operand without explicitly configuring `merge_operator` now fails immediately, causing the DB to enter read-only mode. Previously, failure was deferred until the `merge_operator` was needed by a user read or a background operation.
+* Users of `WALRecoveryMode::kPointInTimeRecovery` whose DB contains a mix of writes with/without `WriteOptions::disableWAL` may now see recovery terminate at an earlier point in time after a crash.
```
ajkr, line 5: I'm sort of torn whether to declare this a behavior change. While it's true this use case may see less data recovered in a case where no corruption happened, that doesn't seem to change the guarantee that `WALRecoveryMode::kPointInTimeRecovery` provides.

anand1976: I'm not sure I understand this. My understanding is there is no change in crash recovery behavior - the tail of the WAL is corrupted and is not recovered. This PR only affects the behavior of corruptions that result in a hole in the WAL, right?

ajkr: Hm, now that I write it out, it seems like a strange case:

- Write to WAL 1 (seqnum 1)
- Crash in the middle of writing seqnum 2 to WAL 1
- Reopen with `avoid_flush_during_recovery` recovers to seqnum 1
- Write with WAL disabled (seqnum 2)
- Write to WAL 2 (seqnum 3)
- Crash
- Reopen and recover to seqnum 1. We couldn't proceed to WAL 2 since it starts with seqnum 3 which means there could be a hole.

Without this PR, we would have recovered to seqnum 3. Technically this didn't cause a corruption given the above scenario wrote seqnum 2 with WAL disabled, but RocksDB cannot know for sure (for all it knows, seqnum 2 could have been a truncated ACK'd record at the end of WAL 1).

anand1976: During DB open, a dummy record is written to the new WAL with the expected sequence number - https://github.com/facebook/rocksdb/blob/master/db/db_impl/db_impl_open.cc#L1616. So after the first crash, WAL 2 should have a dummy record with seqnum 2.

ajkr: Oh that's great. Thanks, I never noticed that trick. Will remove this note, which is nice as then the change is easily backport-able.

anand1976 (approved):
LGTM

PR #7702, opened 2020-11-20
Fix initialization order of DBOptions and kHostnameForDbHostId

anand1976:
Fix initialization order of DBOptions and kHostnameForDbHostId by making the initialization of the latter static rather than dynamic.


ajkr (approved):
LGTM, thanks.

Issue #3147, opened 2017-11-08
bottommost_compression doesn't reject unsupported compression types


mdcallag:
I think this explains a user problem. They set bottommost_compression to kZSTD, but that is not supported, so they don't get compression and use 10X more space than expected.
https://groups.google.com/forum/#!topic/myrocks-dev/N95FCR2UGoA

My server supports kZSTDNotFinalCompression but not kZSTD.

When I try to set compression for a level to kZSTD the server doesn't start as expected:
```
[ERROR] RocksDB: Error opening instance, Status Code: 4, Status: Invalid argument: Compression type ZSTD is not linked wit
h the binary.
```
When I set bottommost_compression to kZSTD the server starts.

This is rejected, as expected:
```
rocksdb_default_cf_options=compression_per_level=kLZ4Compression:kZSTD;bottommost_compression=kLZ4Compression
```

This not rejected, as expected, because the compression types are supported:
```
rocksdb_default_cf_options=compression_per_level=kLZ4Compression:kLZ4Compression;bottommost_compression=kZSTDNotFinalCompression
```

This is accepted but should be rejected because kZSTD is not supported:
```
rocksdb_default_cf_options=compression_per_level=kLZ4Compression:kLZ4Compression;bottommost_compression=kZSTD
```

From RocksDB LOG
```
2017/11/08-10:42:07.701440 7f7a7fc8da40 Compression algorithms supported:
2017/11/08-10:42:07.701441 7f7a7fc8da40         Snappy supported: 1
2017/11/08-10:42:07.701442 7f7a7fc8da40         Zlib supported: 1
2017/11/08-10:42:07.701443 7f7a7fc8da40         Bzip supported: 0
2017/11/08-10:42:07.701444 7f7a7fc8da40         LZ4 supported: 1
2017/11/08-10:42:07.701445 7f7a7fc8da40         ZSTDNotFinal supported: 1
2017/11/08-10:42:07.701451 7f7a7fc8da40         ZSTD supported: 0
...
2017/11/08-10:42:07.701662 7f7a7fc8da40        Options.compression[0]: LZ4
2017/11/08-10:42:07.701664 7f7a7fc8da40        Options.compression[1]: LZ4
2017/11/08-10:42:07.701665 7f7a7fc8da40                  Options.bottommost_compression: ZSTD
```

Leviathan1995:
@yiwu-arbug Hi, is this issue solved? If not, I'd like to try it.

PR #7709, opened 2020-11-21
Now checking bottommost_compression on cf_option

jbosh:
This is in reference to https://github.com/facebook/rocksdb/issues/3147.

### DIFF
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -137,6 +137,12 @@ Status CheckCompressionSupported(const ColumnFamilyOptions& cf_options) {
           CompressionTypeToString(cf_options.compression) +
           " is not linked with the binary.");
     }
+    if (cf_options.bottommost_compression != kDisableCompressionOption && !CompressionTypeSupported(cf_options.bottommost_compression)) {
+      return Status::InvalidArgument(
+          "Compression type " +
+          CompressionTypeToString(cf_options.bottommost_compression) +
+          " is not linked with the binary.");
+    }
   }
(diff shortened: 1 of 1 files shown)


jbosh:
What's the expected fix for this one? Line 65 of OptionsUtilTest.java sets the test to use ZSTD_COMPRESSION but the build machine doesn't appear to have linked ZSTD_COMPRESSION. It seems like I have a few options:

1.  Remove the line and live with no compression for bottommost in the test.
2.  Add a test for "is compression supported" and not set bottommost compression to an unsupported type.
3.  Add zstd to the java tests.

Number one is the easiest. I don't have a Java version of the project running and would prefer not to to solve with option 2 or option 3.

ajkr (changes requested):
> What's the expected fix for this one? Line 65 of OptionsUtilTest.java sets the test to use ZSTD_COMPRESSION but the build machine doesn't appear to have linked ZSTD_COMPRESSION. It seems like I have a few options:
> 
> 1. Remove the line and live with no compression for bottommost in the test.
> 2. Add a test for "is compression supported" and not set bottommost compression to an unsupported type.
> 3. Add zstd to the java tests.
> 
> Number one is the easiest. I don't have a Java version of the project running and would prefer not to to solve with option 2 or option 3.

I think option one sounds good. That should still achieve different compression types in bottommost level vs. upper levels since upper levels will use snappy.

Also, could you please mention the user-visible effect of this PR under "Behavior Changes" section in HISTORY.md?

Issue #7691, opened 2020-11-18
Build error on powerpc64 with Clang 11.0.0


pkubaj:
### Expected behavior
rocksdb doesn't build.

### Actual behavior
It should build.

### Steps to reproduce the behavior
Build rocksdb using Clang 11.0.0 on powerpc64 - I'm using FreeBSD head, but building on Linux should also yield the same results.

Building fails with:
```
util/crc32c.cc:449:54: error: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'unsigned int' [-Werror,-Wshorten-64-to-32]
  return crc32c_ppc(crc, (const unsigned char *)buf, size);
         ~~~~~~~~~~                                  ^~~~
```

jay-zhuang:
I guess we should change `unsigned len` to `unsigned long len` here: https://github.com/facebook/rocksdb/blob/e731f4402258554812c46334dc0d9483e6cc769b/util/crc32c_ppc.c#L70
And
https://github.com/facebook/rocksdb/blob/e731f4402258554812c46334dc0d9483e6cc769b/util/crc32c_ppc.c#L29-L30
@pkubaj we don't have ppc environment with clang11 for test, would you please try that out?

adamretter:
At the moment our ppc64le Jenkins CI and Travis CI systems both use gcc 5.4.0-6ubuntu1~16.04.12 5.4.0 20160609.

I could attempt to setup a ppc64le CLANG 11 environment on the Jenkins CI, assuming that I can get and install a Clang 11 for Ubuntu 16 on ppc64le.

adamretter:
It looks like Clang 11 only became available in the Ubuntu package repo in Ubuntu 20.10 (groovy).
However, it looks like the LLVM project provide pre-built binaries. I will see what I can do...

adamretter:
The ppc64le Clang 11 binaries are available from here: https://releases.llvm.org/download.html#11.0.0

The exact download is: https://github.com/llvm/llvm-project/releases/download/llvmorg-11.0.0/clang+llvm-11.0.0-powerpc64le-linux-ubuntu-16.04.tar.xz

Unfortunately whilst it claims to be usable for Ubuntu 16.04, it appears the binaries have been compiled against GLIBC 2.27, yet Ubuntu 16.04 has GLIBC 2.23. It is Ubuntu 18.04 that ships with GLIBC 2.27. I will report the issue to the LLVM people...

adamretter:
[...]

PR #7713, opened 2020-11-25
Fix Compilation on ppc64le using Clang 11

adamretter:
Closes https://github.com/facebook/rocksdb/issues/7691

The optimised CRC code for PPC64le which was originally imported in https://github.com/facebook/rocksdb/pull/2353 is not compatible with Clang 11. It looks like the code most likely originated from https://github.com/antonblanchard/crc32-vpmsum.

The code relied on a GCC header file `ppc-asm.h` which is not available in Clang.

To solve this, I have taken the same approach as the the upstream project from which the CRC code came https://github.com/antonblanchard/crc32-vpmsum/commit/ffc8018efc1e4f05d22a9fc8dde57109dd09368b#diff-ec3e62c56fbcddeb07230f2a4673c1abd7f0f1cc8e48a2aa560056cfc1b25d60 and simply imported a copy of the GCC header file into our code-base which will be used when Clang is the compiler on pcc64le.

**NOTE**: The new file `util/ppc-asm.h` may have licensing implications which I guess need to be approved by RocksDB/Facebook before this is merged


adamretter:
I cannot fine how to exclude the file `util/ppc-asm.h` from the clang format check... does anyone know how?

util/crc32c.cc
```diff
@@ -346,6 +346,7 @@ static inline void Slow_CRC32(uint64_t* l, uint8_t const **p) {
   table0_[c >> 24];
 }
 
+#if defined HAVE_SSE42 && defined NO_THREEWAY_CRC32C
```
pdillinger, line 349: Lots of builds seem broken, probably from this?

util/ppc-asm.h
```diff
@@ -0,0 +1,390 @@
+/* PowerPC asm definitions for GNU C.
+
+Copyright (C) 2002-2020 Free Software Foundation, Inc.
+
+This file is part of GCC.
+
+GCC is free software; you can redistribute it and/or modify it under
+the terms of the GNU General Public License as published by the Free
+Software Foundation; either version 3, or (at your option) any later
+version.
```
pdillinger, line 10: This appears to be permissible copying. (Internal ref: https://fburl.com/wiki/s1ewaih5)

However, I think it belongs under third-part/gcc/ or even port/ instead of util/.

pdillinger (changes requested):

pdillinger:
> I cannot fine how to exclude the file `util/ppc-asm.h` from the clang format check... does anyone know how?

There's no standard way. We can just ignore the error in rare cases like this. It will not affect future PRs (unless they edit the affected file)

adamretter:
@pdillinger Ah right. Sorry I forgot to backport a similar fix that I made in the Apple Silicon PR to this branch.

pdillinger (changes requested):
Please put ppc-asm.h under third-party/gcc/ or at least port/

adamretter:
@pdillinger Done.

pdillinger (approved):
👍 👍

PR #7721, opened 2020-11-29
Eliminate possible race between LockFile() vs UnlockFile()

azat:
LockFile() accessing LockHoldingInfo (element of locked_files) by
reference after mutex_locked_files had been released.


jay-zhuang (approved):
Thanks @azat for the patch. LGTM.

Otherwise, `it_success` might be invalid after:
https://github.com/facebook/rocksdb/blob/7b2216c90648a9d2ea116c2a34863e6ed9e8c9a0/env/fs_posix.cc#L813-L816

PR #7722, opened 2020-11-30
Fix a race condition in PeriodicWorkScheduler and timer

### LINKED ISSUE #7711: Is there a race in code?
Could you please explain this piece of code? https://github.com/facebook/rocksdb/blob/0355d14dd9c9e41d37bb3a462c5a1ddbc8962392/util/timer.h#L78-L105 

I think there is a race on `thread_` variable, because we can call Start(), then call Start() and Shutdown() in parallel. So, there will be a situation when `running_=false`, but `thread_` is not nullptr, so if `reset` will be called a bit earlier than `join` it will lead to `std::terminate` according to C++ standart. 

It reproduces in stress tests in ClickHouse with this trace https://gist.github.com/alesapin/3749e00d962a0843d61866e4565c5aee

nikitamikhaylov:
There's a race condition in timer `start` and `shutdown`, issue: #7711
And also race condition in `PeriodicWorkScheduler` check pending task and shutdown.


nikitamikhaylov:
Just experiments.

jay-zhuang:
Thanks @nikitamikhaylov for the contribution, let me know if you want to take over the task. For the next step, I think we need to add more tests for concurrent start/shutdown.

env/fs_posix.cc
```diff
@@ -766,9 +766,9 @@ class PosixFileSystem : public FileSystem {
     // closed, all locks the process holds for that *file* are released
     const auto it_success = locked_files.insert({fname, lhi});
     if (it_success.second == false) {
+      LockHoldingInfo& prev_info = it_success.first->second;
```
jay-zhuang, line 769: is this needed?

nikitamikhaylov: This is from #7721 
We also faced with this problem. I can delete this from here, but could you please review that PR?

nikitamikhaylov:
> Thanks @nikitamikhaylov for the contribution, let me know if you want to take over the task. For the next step, I think we need to add more tests for concurrent start/shutdown.

I want, but could you show me a place (a file), where you want to see these tests? I noticed, that test in timer.h aimed to internal data structures. I think it is not Ok to add a test which will create threads from inside the class.

util/timer.h
```diff
@@ -124,6 +125,7 @@ class Timer {
       if (!running_) {
         return false;
       }
+      shutting_down_ = true;
       running_ = false;
```
pdillinger, line 129: If we have two concurrent calls to Shutdown, one will return before the timer is fully shutdown, with tasks potentially happening. User might reasonably expect timer to be inactive when Shutdown returns (regardless of return value, assuming no additional Start).

util/timer.h
```diff
@@ -111,7 +112,7 @@ class Timer {
     if (running_) {
       return false;
     }
-
+    WaitForShuttingDownIfNecessary();
```
pdillinger, line 115: I don't foresee a practical need to have Start block on a pending shutdown. It could just return false instead. If the user wants Start to be guaranteed, Shutdown should have returned without an intervening Start.

pdillinger (commented):
From my observations, a "shutdown requested" field while keeping running_==true might be a better model.

util/timer.h
nikitamikhaylov: But almost all code that uses Timer doesn't check the return value.

util/timer.h
nikitamikhaylov: I agree.

nikitamikhaylov:
For now the semantic of Shutdown is following: if there are several concurrent call, each of them will be blocked until thread is been joined and internal state will be cleared.

nikitamikhaylov:
How to understand what went wrong in `build-linux` check?

util/timer_test.cc
```diff
@@ -390,6 +390,45 @@ TEST_F(TimerTest, DestroyTimerWithRunningFunc) {
   control_thr.join();
 }
 
+TEST_F(TimerTest, ConcurrentStartShutdown) {
```
jay-zhuang, line 393: Would be great to add a deterministic test by using `mock_env_.MockSleepForMicroseconds()` + `TEST_SYNC_POINT()` in timer.h, for example:
https://github.com/facebook/rocksdb/blob/9be75629b4cc5fa0786440f7c92e4f67ca3fdebe/util/timer_test.cc#L178-L183

jay-zhuang (changes requested):

util/timer.h
```diff
@@ -108,7 +111,7 @@ class Timer {
   // Start the Timer
   bool Start() {
     InstrumentedMutexLock l(&mutex_);
-    if (running_) {
+    if (running_ || shutdown_requested_) {
```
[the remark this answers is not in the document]
nikitamikhaylov: Why don't you want Start() to be blocked until previous Shutdown() ended?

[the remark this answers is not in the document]
jay-zhuang: Yes, that's an option to fix #7711 
There's another race condition in PeriodicWorkScheduler, which add task and call start(), another thread could call cancel, then shutdown. We have to change timer.h to fix that, would be better to fix them together.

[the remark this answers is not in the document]
nikitamikhaylov: Ok, I didn't read the code of PeriodicWorkScheduler. Will try to fix them both.

jay-zhuang:
Hi @nikitamikhaylov just checking if you're still working on it. Thanks.

nikitamikhaylov:
> Hi @nikitamikhaylov just checking if you're still working on it. Thanks.

I'm here.

util/timer.h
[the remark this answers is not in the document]
ajkr: > We have to change timer.h to fix that, would be better to fix them together.

What about hold a lock during `PeriodicWorkScheduler::{Register,Unregister}()`? We could also take the timer.h fix from https://github.com/ClickHouse-Extras/rocksdb/pull/1 to make `Timer` less error prone.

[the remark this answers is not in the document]
nikitamikhaylov: @ajkr `We could also take the timer.h fix from ClickHouse-Extras#1 to make Timer's functions thread-safe.`

This is not a final fix, there were some updates... So, for now ClickHouse-Extras has the same code as in this PR and we detected no problems detected by our CI.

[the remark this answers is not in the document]
ajkr: But with the current version, as Jay pointed out, calling `Start()` while `shutdown_requested_ == true` will return `false` immediately, while `PeriodicWorkScheduler` does not check the return value (as you pointed out). I think in that case the periodic work will not be scheduled as expected.

PR #7723, opened 2020-11-30
Warn about practically unfixable TSAN warnings in stack trace

pdillinger:
Summary: TSAN reports that our stack trace handler makes unsafe calls
during a signal handler. I just tried fixing some of them and I don't
think it's fixable unless we can get away from using FILE stdio. Even if
we can use lower level functions only, I'm not sure it's fixed.

I also tried suppressing the reports with function and file level TSAN
suppression, but that doesn't seem to work, perhaps because the
violation is reported on the callee, not the caller.

So I added a warning to be printed whenever these violations would be
reported that they are practically ignorable.

Internal ref: T77844138

Test Plan: run external_sst_file_test with seeded abort(), with TSAN
(TSAN warnings + new warning) and without TSAN (no warning, just stack
trace).

### DIFF
--- a/port/stack_trace.cc
+++ b/port/stack_trace.cc
@@ -144,6 +144,22 @@ static void StackTraceHandler(int sig) {
   fprintf(stderr, "Received signal %d (%s)\n", sig, strsignal(sig));
   // skip the top three signal handler related frames
   PrintStack(3);
+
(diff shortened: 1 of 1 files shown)


akankshamahajan15 (approved):
LGTM!

PR #7724, opened 2020-12-01
[WIP] Fix timer race condition

jay-zhuang:


util/timer.h
```diff
@@ -121,17 +123,20 @@ class Timer {
   bool Shutdown() {
     {
       InstrumentedMutexLock l(&mutex_);
+      shutting_down_ = true;
```
mrambacher, line 126: Perhaps I do not understand some of the cases enough but won't this cause the Start task to wait forever if running_ = false?  Right below here it returns false "if (!running_)", without resetting shutdown or signaling any conditions.

mrambacher (commented):
I think #7722 was also proposed as a fix to the same problem.  I do not know which one is correct or better but wanted to bring it to attention.

nikitamikhaylov:
> I think #7722 was also proposed as a fix to the same problem. I do not know which one is correct or better but wanted to bring it to attention.

I thought that I was easy to fix, but no :)

util/timer.h
jay-zhuang: Nice catch. Thanks @mrambacher

PR #7726, opened 2020-12-01
Use size_t for filter APIs, protect against overflow

pdillinger:
Summary: Deprecate CalculateNumEntry and replace with
ApproximateNumEntries (better name) using size_t instead of int and
uint32_t, to minimize confusing casts and bad overflow behavior
(possible though probably not realistic). Bloom sizes are now explicitly
capped at max size supported by implementations: just under 4GiB for
fv=5 Bloom, and just under 512MiB for fv<5 Legacy Bloom. This
hardening could help to set up for fuzzing.

Also, since RocksDB only uses this information as an approximation
for trying to hit certain sizes for partitioned filters, it's more important
that the function be reasonably fast than for it to be completely
accurate. It's hard enough to be 100% accurate for Ribbon (currently
reversing CalculateSpace) that adding optimize_filters_for_memory
into the mix is just not worth trying to be 100% accurate for num
entries for bytes.

Also:
- Cleaned up filter_policy.h to remove MSVC warning handling and
potentially unsafe use of exception for "not implemented"
- Correct the number of entries limit beyond which current Ribbon
implementation falls back on Bloom instead.
- Consistently use "num_entries" rather than "num_entry"
- Remove LegacyBloomBitsBuilder::CalculateNumEntry as it's essentially
obsolete from general implementation
BuiltinFilterBitsBuilder::CalculateNumEntries.
- Fix filter_bench to skip some tests that don't make sense when only
one or a small number of filters has been generated.

Test Plan: expanded existing unit tests for CalculateSpace /
ApproximateNumEntries. Also manually used filter_bench to verify Legacy and
fv=5 Bloom size caps work (much too expensive for unit test). Note that
the actual bits per key is below requested due to space cap.

    $ ./filter_bench -impl=0 -bits_per_key=20 -average_keys_per_filter=256000000 -vary_key_count_ratio=0 -m_keys_total_max=256 -allow_bad_fp_rate
    ...
    Total size (MB): 511.992
    Bits/key stored: 16.777
    ...
    $ ./filter_bench -impl=2 -bits_per_key=20 -average_keys_per_filter=2000000000 -vary_key_count_ratio=0 -m_keys_total_max=2000
    ...
    Total size (MB): 4096
    Bits/key stored: 17.1799
    ...
    $


table/block_based/filter_policy.cc
```diff
@@ -100,18 +100,21 @@ class FastLocalBloomBitsBuilder : public XXH3pFilterBitsBuilder {
...
 
+    // Max size supported by implementation
+    assert(len_with_metadata <= 0xffffffffU);
+
     // Compute num_probes after any rounding / adjustments
-    int num_probes = GetNumProbes(num_entry, len_with_metadata);
+    int num_probes = GetNumProbes(num_entries, len_with_metadata);
 
-    uint32_t len = len_with_metadata - 5;
+    uint32_t len = static_cast<uint32_t>(len_with_metadata - 5);
```
jay-zhuang, line 117: could `len` also be `size_t`?

include/rocksdb/filter_policy.h
```diff
@@ -51,22 +51,23 @@ class FilterBitsBuilder {
...
-#else
-    abort();
-#endif
-    return 0;
+  // <= the specified number of bytes, though callers (including RocksDB)
+  // should only rely on that for performance, not correctness.
+  // This default implementation is for old API compatibility.
+  virtual size_t CalculateNumEntries(size_t bytes) {
+    bytes = std::min(bytes, size_t{0xffffffff});
+    return static_cast<size_t>(CalculateNumEntry(static_cast<uint32_t>(bytes)));
```
jay-zhuang, line 59: As `CalculateNumEntry` is going to be deprecated, should we not even use it in the default implementation here and just leave it pure virtual?

include/rocksdb/filter_policy.h
```diff
@@ -51,22 +51,23 @@ class FilterBitsBuilder {
...
+  }
+
+  // Old, DEPRECATED version of CalculateNumEntries. This is not
+  // called by RocksDB except as the default implementation of
+  // CalculateNumEntries for API compatibility.
+  virtual int CalculateNumEntry(const uint32_t bytes) {
+    // DEBUG: ideally should not rely on this implementation
+    assert(false);
+    // RELEASE: something reasonably "safe": 2 bytes per entry
+    return static_cast<int>(bytes / 2);
```
jay-zhuang, line 69: sorry, why 2 bytes per entry is safe? Isn't it change the default behavior of the API?

table/block_based/filter_policy.cc
```diff
@@ -610,12 +625,10 @@ class LegacyBloomBitsBuilder : public BuiltinFilterBitsBuilder {
 
   Slice Finish(std::unique_ptr<const char[]>* buf) override;
 
-  int CalculateNumEntry(const uint32_t bytes) override;
-
-  uint32_t CalculateSpace(const int num_entry) override {
+  size_t CalculateSpace(size_t num_entries) override {
```
jay-zhuang, line 628: Not sure if `const size_t` here would be better? What do you think?

table/block_based/filter_policy.cc
pdillinger: The filter implementation in bloom_impl.h uses FastRange32 to pick a 64-byte bucket. Although this could theoretically scale up to 64 << 32 bytes, the closer you get to that, the more skewed your distribution of elements into buckets can get. With a limit around 1 << 32, the skew is no more than ~2% (64 / 63 - 1).

table/block_based/filter_policy.cc
pdillinger: Google style prefers no const for the declaration:
> For a function parameter passed by value, const has no effect on the caller, thus is not recommended in function declarations.

For function definition / override, this is probably most applicable, since parameter definitions are local variables:
> Using const on local variables is neither encouraged nor discouraged.

My personal preference is to have the definition or implementations match the declaration. When I see 'const' in a function signature, I'm expecting a constraint on the caller, which doesn't apply here.

include/rocksdb/filter_policy.h
pdillinger: RocksDB is calling the new implementation. For other old implementations of the API to keep working, we need this default implementation.

include/rocksdb/filter_policy.h
pdillinger: The old behavior was to throw an exception, which in RocksDB can lead to UB including DB corruption. abort() would be better, but since we don't really expect outside callers of this API, just implementors, I've added this:

    // callers (including RocksDB)
    // should only rely on that for performance, not correctness.

Thus, it's ok for correctness to return something wrong. 16 bits per key is probably not too far from reality. Perhaps I should say "conservative" instead of "`"safe"`"

include/rocksdb/filter_policy.h
jay-zhuang: okay, I see, if the user has their own filter with `CalculateNumEntry()` implemented, it would still work.

jay-zhuang (approved):
👍

PR #7728, opened 2020-12-01
Skip minimum rate check in Sandcastle

pdillinger:
Summary: The minimum rate check in RateLimiterTest.Rate can fail in
Facebook's CI system Sandcastle, presumably due to heavily loaded
machines. This change disables the minimum rate check for Sandcastle
runs, and cleans up the code disabling it on other CI environments. (The
amount of conditionally compiled code shall be minimized.)

Test Plan: try new test with and without setting envvar SANDCASTLE=1

### DIFF
--- a/util/rate_limiter_test.cc
+++ b/util/rate_limiter_test.cc
@@ -128,11 +128,13 @@ TEST_F(RateLimiterTest, Rate) {
   }
 
   // This can fail in heavily loaded CI environments
-  if (getenv("SANDCASTLE")
+  bool skip_minimum_rate_check =
 #if (defined(TRAVIS) || defined(CIRCLECI)) && defined(OS_MACOSX)
-      || true
+      true;
+#else
+      getenv("SANDCASTLE");
 #endif
-  ) {
+  if (skip_minimum_rate_check) {
     fprintf(stderr, "Skipped minimum rate check (%d / %d passed)\n",
             samples_at_minimum, samples);
   } else {
--- a/util/rate_limiter_test.cc
+++ b/util/rate_limiter_test.cc
@@ -59,7 +59,6 @@ TEST_F(RateLimiterTest, Modes) {
   }
 }
 
-#if !((defined(TRAVIS) || defined(CIRCLECI)) && defined(OS_MACOSX))
 TEST_F(RateLimiterTest, Rate) {
   auto* env = Env::Default();
   struct Arg {
@@ -90,6 +89,9 @@ TEST_F(RateLimiterTest, Rate) {
     }
   };
 
+  int samples = 0;
+  int samples_at_minimum = 0;
+
   for (int i = 1; i <= 16; i *= 2) {
     int32_t target = i * 1024 * 10;
     Arg arg(target, i / 4 + 1);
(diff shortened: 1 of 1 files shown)


util/rate_limiter_test.cc
```diff
@@ -117,12 +119,26 @@ TEST_F(RateLimiterTest, Rate) {
...
       ASSERT_LE(rate / target, 1.25);
     }
   }
-}
+
+  // This can fail in heavily loaded CI environments
+  if (getenv("SANDCASTLE")
+#if (defined(TRAVIS) || defined(CIRCLECI)) && defined(OS_MACOSX)
+      || true
 #endif
```
mrambacher, line 134: This might be clearer as two separate conditions rather than the || true

util/rate_limiter_test.cc
```diff
@@ -117,12 +119,26 @@ TEST_F(RateLimiterTest, Rate) {
...
+  ) {
+    fprintf(stderr, "Skipped minimum rate check (%d / %d passed)\n",
+            samples_at_minimum, samples);
```
mrambacher, line 137: Should there also be a ROCKSDB_GTEST_SKIP here, so we can somehow potentially in the future know that the test was skipped?

util/rate_limiter_test.cc
```diff
@@ -117,12 +119,26 @@ TEST_F(RateLimiterTest, Rate) {
...
+        ++samples_at_minimum;
+      }
       ASSERT_LE(rate / target, 1.25);
     }
   }
-}
+
+  // This can fail in heavily loaded CI environments
+  if (getenv("SANDCASTLE")
+#if (defined(TRAVIS) || defined(CIRCLECI)) && defined(OS_MACOSX)
```
mrambacher, line 132: Out of curiosity, why is this skipped on MacOS?  One of these is not like the other...

util/rate_limiter_test.cc
pdillinger: IMHO `#else` adds extra complication.

util/rate_limiter_test.cc
pdillinger: I'm not taking ownership of that. Just reducing the grossness of an existing hack job.

util/rate_limiter_test.cc
pdillinger: We're only skipping one aspect of the test, so I don't think ROCKSDB_GTEST_SKIP is appropriate here. I printed out the line above for more visibility into something being skipped.

Let me know if I have done anything to suggest that this is a good test, or to make it more difficult for the next person to work on RateLimiter tests to improve things further.

util/rate_limiter_test.cc
ltamasi: How about something like

```
#if (defined(TRAVIS) || defined(CIRCLECI)) && defined(OS_MACOSX)
#define DISABLE_MINIMUM_RATE_CHECK true
#else
#define DISABLE_MINIMUM_RATE_CHECK false
#endif
```

and then

```
if (DISABLE_MINIMUM_RATE_CHECK || getenv("SANDCASTLE")) { ... }
```

?

ltamasi (approved):
LGTM with one suggestion, thanks!

util/rate_limiter_test.cc
pdillinger: Apparently does still fail in MacOS on CircleCI: Skipped minimum rate check (9 / 10 passed)

PR #7732, opened 2020-12-02
Fix build on FreeBSD/powerpc64(le)

pkubaj:
To build on FreeBSD, arch_ppc_probe needs to be adapted to FreeBSD.

Since FreeBSD uses elf_aux_info as an getauxval equivalent, use it and include necessary headers:
- machine/cpu.h for PPC_FEATURE2_HAS_VEC_CRYPTO,
- sys/auxv.h for elf_aux_info,
- sys/elf_common.h for AT_HWCAP2.

elf_aux_info isn't checked for being available, because it's available since FreeBSD 12.0. rocksdb assumes using Clang on FreeBSD, but powerpc* platforms switch to Clang only since 13.0.

This patch makes rocksdb build on FreeBSD on powerpc64 and powerpc64le platforms.

### DIFF
--- a/util/crc32c.cc
+++ b/util/crc32c.cc
@@ -478,7 +478,7 @@ static int arch_ppc_probe(void) {
 #if defined(__powerpc64__)
   elf_aux_info(AT_HWCAP2, &cpufeatures, sizeof(cpufeatures));
   if (cpufeatures & PPC_FEATURE2_HAS_VEC_CRYPTO) arch_ppc_crc32 = 1;
-#endif /* __powerpc64__ */
+#endif  /* __powerpc64__ */
 
   return arch_ppc_crc32;
 }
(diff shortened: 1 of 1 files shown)


adamretter:
@pkubaj I just ran this against Linux of ppc64le to make sure that was still fine too. It looks good.

Thanks for your contribution.

@pdillinger I think this is ready to merge.

util/crc32c.cc
```diff
@@ -467,6 +467,21 @@ static int arch_ppc_probe(void) {
 
   return arch_ppc_crc32;
 }
+#elif __FreeBSD__
+#include <machine/cpu.h>
+#include <sys/auxv.h>
+#include <sys/elf_common.h>
```
pdillinger, line 473: Please move #include outside of namespace

pdillinger (changes requested):

pdillinger (approved):
Thanks for the contribution!

PR #7733, opened 2020-12-02
Invalidate iterator on transaction clear

lth:
Some clients do not close their iterators until after the transaction finishes. To handle this case, we will invalidate any iterators on transaction clear.


cheng-chang (approved):

utilities/transactions/write_unprepared_transaction_test.cc
```diff
@@ -636,6 +636,52 @@ TEST_P(WriteUnpreparedTransactionTest, IterateAndWrite) {
   }
 }
 
+// Test that using an iterator after transaction clear is not supported
+TEST_P(WriteUnpreparedTransactionTest, IterateAfterClear) {
+  WriteOptions woptions;
+  TransactionOptions txn_options;
+  txn_options.write_batch_flush_threshold = 1;
+
+  enum Action { DO_COMMIT, DO_ROLLBACK };
```
cheng-chang, line 645: Nit: the style prefers naming like `kDoCommit`, `kDoRollback`.

PR #7739, opened 2020-12-03
Do not full scan obsolete files on compaction busy

burtonli:
When ConcurrentTaskLimiter is enabled and there are too many outstanding compactions, BackgroundCompaction returns Status::Busy(), which shouldn't be treat as compaction failure.
This caused performance issue when outstanding compactions reached the limit.


burtonli:
+ @ltamasi

ltamasi (approved):
LGTM. Thanks for the patch @burtonli !

ltamasi:
@burtonli Could you please call out the fix in HISTORY.md?

PR #7743, opened 2020-12-04
Make the TARGETS file Starlark compliant

ndmitchell:
Buck TARGETS files are sometimes parsed with Python, and sometimes with Starlark - this TARGETS file was not Starlark compliant. In Starlark you can't have a top-level if in a TARGETS file, but you can have a ternary `a if b else c`. Therefore I converted TARGETS, and updated the generator for it.

### DIFF
--- a/TARGETS
+++ b/TARGETS
@@ -775,26 +775,24 @@ cpp_library(
     external_deps = ROCKSDB_EXTERNAL_DEPS,
 )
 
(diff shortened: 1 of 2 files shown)


pdillinger (approved):
LGTM. The Travis failures are unrelated to your PR. "Import to Phabrictor" and I'll approve it there also.

PR #7745, opened 2020-12-04
Migrate away from broken macos on Travis

pdillinger:
Summary: Add macos+cmake build on CircleCI instead.

Test Plan: CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ commands:
           name: Install gflags
           command: |
             sudo apt-get update -y && sudo apt-get install -y libgflags-dev
-  
+
   install-gflags-on-macos:
     steps:
       - run:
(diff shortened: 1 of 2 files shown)


riversand963 (approved):

adamretter:
@pdillinger I am not certain, but if I understand the output on Circle CI correctly, then I think we seem to have lost a few CI jobs that previously ran on macOS.

On Circle CI I see a single job: `build-macos` which runs `make check`.

On Travis there used to be 10 separate jobs on macOS
Screenshot 2020-12-07 at 12 07 43

I think at least 5 of the previous Travis macOS jobs `TEST_GROUP` 1 though 4 and `platform_dependent` are covered by `make check` on CircleCI. However, I think that the following are missing (i.e. have no equivalents in CircleCI):
1. `cmake`
2. `status_checked`
3. `examples`
4. `java_test`

Should I send a PR to reinstate these, or is this something that you are already working on?

pdillinger:
> I think at least 5 of the previous Travis macOS jobs `TEST_GROUP` 1 though 4 and `platform_dependent` are covered by `make check` on CircleCI. However, I think that the following are missing (i.e. have no equivalents in CircleCI):
> 
> 1. `cmake`

Added in this PR.

> 2. `status_checked`

Should not be needed on multiple platforms, except that we also use this build to test LIB_MODE=shared, which should be done on multiple platforms.

> 3. `examples`

For Travis, we used to have small executors and had to split out jobs up for acceptable CI latency, accepting extra redundant work. With CircleCI we have large executors available, so can do more in each test to minimize redundant compute.

The 'make static_lib' part of our CircleCI build-examples is essentially redundant with build-release. Parallel make on examples directory for me takes 10s. There should be no reason to run another >1m release build just for compiling examples, but we should compile all code (including examples) on all platforms where it is supported.

Perhaps it makes sense to combine build-release and build-examples and run it on macos also. And since build-release is already pretty fast (~4m), we could consider tacking on a clean LIB_MODE=shared release build.

> 4. `java_test`

Sure, though if we start to use too much CircleCI compute, this is the kind of thing that is low-value on most PRs.

> Should I send a PR to reinstate these, or is this something that you are already working on?

Sure, and I'm not hacking on it currently.

PR #7747, opened 2020-12-05
Carry over min_log_number_to_keep_2pc in new MANIFEST

cheng-chang:
When two phase commit is enabled, `VersionSet::min_log_number_to_keep_2pc` is set during flush.
But when a new MANIFEST is created, the `min_log_number_to_keep_2pc` is not carried over to the new MANIFEST. So if a new MANIFEST is created and then DB is reopened, the `min_log_number_to_keep_2pc` will be lost.  This may cause DB recovery errors.
The bug is reproduced in a new unit test in `version_set_test.cc`.

Test Plan:
The new unit test in `version_set_test.cc` should pass.

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -5278,6 +5278,14 @@ Status VersionSet::WriteCurrentStateToManifest(
       assert(iter != curr_state.end());
       uint64_t log_number = iter->second.log_number;
       edit.SetLogNumber(log_number);
+
+      if (cfd->GetID() == 0) {
+        uint64_t min_log = min_log_number_to_keep_2pc();
+        if (min_log != 0) {
+          edit.SetMinLogNumberToKeep(min_log);
+        }
+      }
+
       std::string record;
       if (!edit.EncodeTo(&record)) {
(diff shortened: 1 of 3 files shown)


db/version_set.cc
```diff
@@ -5278,6 +5278,14 @@ Status VersionSet::WriteCurrentStateToManifest(
       assert(iter != curr_state.end());
       uint64_t log_number = iter->second.log_number;
       edit.SetLogNumber(log_number);
+
+      if (cfd->GetID() == 0) {
```
jay-zhuang, line 5282: Why does it only apply to default CF?

cheng-chang: min_log_number_to_keep is for the whole db, not for specific column family.
So it does not need to be set for every column family, just need to be set once.
Since default CF can never be dropped, we set the min_log to the default CF here.

jay-zhuang: maybe add a comment for that?

jay-zhuang (approved):
LGTM. Thanks @cheng-chang for the quick fix.

PR #7749, opened 2020-12-05
Fix memory leak for ColumnFamily drop with live iterator

pdillinger:
Summary: Uncommon bug seen by ASAN with
ColumnFamilyTest.LiveIteratorWithDroppedColumnFamily, if the last two
references to a ColumnFamilyData are both SuperVersions (during
InstallSuperVersion). The fix is to use UnrefAndTryDelete even in
SuperVersion::Cleanup but with a parameter to avoid re-entering Cleanup
on the same SuperVersion being cleaned up.

ColumnFamilyData::Unref is considered unsafe so removed.

Test Plan: ./column_family_test --gtest_filter=*LiveIter* --gtest_repeat=100

### DIFF
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -449,9 +449,7 @@ void SuperVersion::Cleanup() {
     to_delete.push_back(m);
   }
   current->Unref();
-  if (cfd->Unref()) {
-    delete cfd;
-  }
+  cfd->UnrefAndTryDelete(this);
 }
 
 void SuperVersion::Init(ColumnFamilyData* new_cfd, MemTable* new_mem,
@@ -660,7 +658,7 @@ ColumnFamilyData::~ColumnFamilyData() {
   }
 }
 
-bool ColumnFamilyData::UnrefAndTryDelete() {
+bool ColumnFamilyData::UnrefAndTryDelete(SuperVersion* sv_under_cleanup) {
   int old_refs = refs_.fetch_sub(1);
   assert(old_refs > 0);
 
@@ -670,7 +668,11 @@ bool ColumnFamilyData::UnrefAndTryDelete() {
     return true;
   }
 
-  if (old_refs == 2 && super_version_ != nullptr) {
+  // If called under SuperVersion::Cleanup, we should not re-enter Cleanup on
+  // the same SuperVersion. (But while installing a new SuperVersion, this
+  // cfd could be referenced only by two SuperVersions.)
+  if (old_refs == 2 && super_version_ != nullptr &&
+      super_version_ != sv_under_cleanup) {
     // Only the super_version_ holds me
     SuperVersion* sv = super_version_;
     super_version_ = nullptr;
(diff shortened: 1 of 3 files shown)


pdillinger:
> Hi @pdillinger 👋
> Getting in touch on behalf of Snyk. Could this memory leak represent a security risk? I.e. could a malicious user practically exploit it and trigger a DoS on the victim server?

Dropping a column family is not normally done in response to service user data, but for maintenance or internal data migrations.

db/column_family.cc
```diff
@@ -449,9 +449,7 @@ void SuperVersion::Cleanup() {
     to_delete.push_back(m);
   }
   current->Unref();
-  if (cfd->Unref()) {
-    delete cfd;
-  }
+  cfd->UnrefAndTryDelete(this);
```
jay-zhuang, line 452: Just a question: when does `ColumnFamilyData` have different `SuperVersion`?
I thought they're always point to each other for the same CF.

db/column_family.cc
```diff
@@ -670,7 +668,11 @@ bool ColumnFamilyData::UnrefAndTryDelete() {
     return true;
   }
 
-  if (old_refs == 2 && super_version_ != nullptr) {
+  // If called under SuperVersion::Cleanup, we should not re-enter Cleanup on
+  // the same SuperVersion. (But while installing a new SuperVersion, this
+  // cfd could be referenced only by two SuperVersions.)
+  if (old_refs == 2 && super_version_ != nullptr &&
+      super_version_ != sv_under_cleanup) {
```
jay-zhuang, line 675: If `sv_under_cleanup` is only used by `SuperVersion::Cleanup()`, should this logic also move to there? I feel having a SuperVersion a parameter for `UnrefAndTryDelete` is a little confusing.

db/column_family.cc
pdillinger: From new comment

    // (But while installing a new SuperVersion, this
    // cfd could be referenced only by two SuperVersions.)

db/column_family.cc
pdillinger: When sv_under_cleanup is non-null, we might exclude this code. We might need the code below from any caller. If you have a specific proposal to simplify, let me know. It is not simple as it is here.

jay-zhuang: I think my suggestion is to change `SuperVersion::Cleanup()`, like:
```
  if (cfd->GetSuperVersion() == this) {
    if (cfd->Unref()) {
      delete cfd;
    }
  } else {
    cfd->UnrefAndTryDelete();
  }
```

jay-zhuang (approved):
LGTM.

db/column_family.cc
jay-zhuang: oh, I see, one is the new SV, another one is the old SV.

db/column_family.cc
pdillinger: We would either need to keep the Unref() function around, which I fear invites further bugs, or make SuperVersion a friend of ColumnFamilyData. Making a `friend` suggests we would be violating the natural encapsulation for ColumnFamilyData.

PR #7751, opened 2020-12-07
Refactor ProcessManifestWrites a little bit

riversand963:
Summary:
This PR removes a nested loop inside ProcessManifestWrites. The new
implementation has the same behavior as the old code with simpler logic
and lower complexity.

Test Plan:
make check
Run make crash_test on devserver and succeeds 3 times.


riversand963:
This PR is a follow-up of https://github.com/facebook/rocksdb/pull/7740#discussion_r536429580.

riversand963:
cc @cheng-chang since the PR is about WAL info computation.

ltamasi (approved):
LGTM, thanks for the patch @riversand963 !

PR #7754, opened 2020-12-07
Do not use ASSERT_OK in child threads in ExternalSstFileTest.PickedLevelBug

ltamasi:
Summary:
`googletest` uses exceptions to communicate assertion failures when
`GTEST_THROW_ON_FAILURE` is set, which does not go well with
`std::thread`s, since an exception escaping the top-level function of an
`std::thread` object or an `std::thread` getting destroyed without
having been `join`ed or `detach`ed first results in a call to
`std::terminate`. The patch fixes this by moving the `Status` assertions
of background operations in `ExternalSstFileTest.PickedLevelBug` to the
main thread.

Test Plan:
`make check`

### DIFF
--- a/db/external_sst_file_test.cc
+++ b/db/external_sst_file_test.cc
@@ -1345,15 +1345,18 @@ TEST_F(ExternalSSTFileTest, PickedLevelBug) {
   ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->EnableProcessing();
 
   // While writing the MANIFEST start a thread that will ask for compaction
+  Status bg_compact_status;
   ROCKSDB_NAMESPACE::port::Thread bg_compact([&]() {
-    ASSERT_OK(db_->CompactRange(CompactRangeOptions(), nullptr, nullptr));
+    bg_compact_status =
+        db_->CompactRange(CompactRangeOptions(), nullptr, nullptr);
   });
   TEST_SYNC_POINT("ExternalSSTFileTest::PickedLevelBug:2");
 
   // Start a thread that will ingest a new file
+  Status bg_addfile_status;
   ROCKSDB_NAMESPACE::port::Thread bg_addfile([&]() {
     file_keys = {1, 2, 3};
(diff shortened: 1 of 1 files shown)


riversand963 (approved):

adamretter:
@ltamasi I have seen this in other places whilst looking into the ASSERT_STATUS_CHECKED stuff. Is this a general rule that should be applied throughout the test suite?

db/external_sst_file_test.cc
```diff
@@ -1345,15 +1345,18 @@ TEST_F(ExternalSSTFileTest, PickedLevelBug) {
   ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->EnableProcessing();
 
   // While writing the MANIFEST start a thread that will ask for compaction
+  Status bg_compact_status;
```
adamretter, line 1348: I am not certain, but... to avoid stale reads due to CPU caching etc, doesn't `bg_compact_status` need synchronisation as it is written and read from different threads? I would have thought that `std::atomic` is needed here.

Same goes for `bg_addfile_status` below...

pdillinger: One thread per field -> no write/write conflict
Thread join before read -> no read/write conflict

ltamasi:
> @ltamasi I have seen this in other places whilst looking into the ASSERT_STATUS_CHECKED stuff. Is this a general rule that should be applied throughout the test suite?

The rules around `std::thread`s and exceptions/destruction are defined by the C++ standard, so yes, I think it would be nice to do this across the board. Note that the worst thing that might happen is that the test binary crashes instead of properly reporting a test failure, so fixing this is probably not a super high priority.

db/external_sst_file_test.cc
ltamasi: > Thread join before read -> no read/write conflict

To use the term from the standard, the thread join creates a *synchronizes-with* relationship (see https://en.cppreference.com/w/cpp/thread/thread/join).

ltamasi:
Closing since this is already merged.

PR #7755, opened 2020-12-08
Add artifact suffix to shared library

SeekingMeaning:
On Unix systems, `ARTIFACT_SUFFIX` was added to the static library `librocksdb.a` but not the shared library `librocksdb.so`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -929,7 +929,7 @@ if(ROCKSDB_BUILD_SHARED)
                           LINKER_LANGUAGE CXX
                           VERSION ${rocksdb_VERSION}
                           SOVERSION ${rocksdb_VERSION_MAJOR}
-                          OUTPUT_NAME "rocksdb")
+                          OUTPUT_NAME "rocksdb${ARTIFACT_SUFFIX}")
   endif()
 endif()
 


SeekingMeaning:
Any updates on this? (This is used in Homebrew; see https://github.com/Homebrew/homebrew-core/pull/66445)

adamretter:
@SeekingMeaning Are you saying that the resultant shared library file is currently just `librocksdb`, but that it should be `librocksdb.so`?

SeekingMeaning:
Suffixes like `_je` at https://github.com/facebook/rocksdb/blob/master/thirdparty.inc#L263 aren't added to `librocksdb.so` (when it should be `librocksdb_je.so`)

adamretter:
@SeekingMeaning I wonder, would it be better to use:

```
OUTPUT_NAME $ROCKSDB_SHARED_LIB)
```

It looks like `ROCKSDB_SHARED_LIB` is set at line 908 and includes the suffix that you are looking for.

SeekingMeaning:
Perhaps, although on Unix libraries tend to have the same name in both static and shared form

adamretter:
@jay-zhuang What do you think?

jay-zhuang:
https://github.com/facebook/rocksdb/blob/master/thirdparty.inc is only used for Windows build and Windows build doesn't use the suffix flag. So currently `ARTIFACT_SUFFIX` is always empty. Is there other situation that `ARTIFACT_SUFFIX` would be set?
Another question, we have https://github.com/facebook/rocksdb/blob/master/Makefile, should it follow the same rule to generate the same binary name?

For us, seems the change is harmless, as long as rocksdb dependants are fine. cc. @ajkr

SeekingMeaning:
> Is there other situation that `ARTIFACT_SUFFIX` would be set?

In our case in Homebrew, we use `ARTIFACT_SUFFIX` to add the suffix `_lite` when building RocksDB with `ROCKSDB_LITE` enabled

jay-zhuang (approved):
LGTM

PR #7756, opened 2020-12-08
Ensure that MultiGet works properly with compressed cache

anand1976:
Ensure that when direct IO is enabled and a compressed block cache is
configured, MultiGet inserts compressed data blocks into the compressed
block cache.

Test plan:
Add unit test to db_basic_test


cheng-chang (approved):

cheng-chang:
@anand1976 can you see the reason behind the test failures?

PR #7757, opened 2020-12-08
Add a test for disabling tracking WAL

cheng-chang:
If WAL tracking was enabled, then disabled during reopen, the previously tracked WALs should be removed from MANIFEST.

Test Plan:
a new unit test `DBBasicTest.DisableTrackWal` is added.


jay-zhuang (approved):
👍

PR #7762, opened 2020-12-09
Add a new db_map_fuzzer

cheng-chang:
Execute randomly generated operations on both a DB and a std::map,
then reopen the DB and make sure that iterating the DB produces the
same key-value pairs as iterating through the std::map.

Test Plan:
cd fuzz && make db_map_fuzzer && ./db_map_fuzzer

### DIFF
--- a/fuzz/db_map_fuzzer.cc
+++ b/fuzz/db_map_fuzzer.cc
@@ -87,6 +87,7 @@ DEFINE_PROTO_FUZZER(DBOperations& input) {
         break;
       }
       case OpType::DELETE_RANGE: {
+        // [op.key(), op.value()) corresponds to [begin, end).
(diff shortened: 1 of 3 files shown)


fuzz/db_map_fuzzer.cc
```diff
@@ -0,0 +1,120 @@
...
+        kv[op.key()] = op.value();
+        break;
+      }
+      case OpType::DELETE: {
+        CHECK_OK(db->Delete(rocksdb::WriteOptions(), op.key()));
+        kv.erase(op.key());
+        break;
+      }
+      case OpType::DELETE_RANGE: {
+        CHECK_OK(db->DeleteRange(rocksdb::WriteOptions(),
```
pdillinger, line 90: Might be worth a comment or something that here `op.value()` is a second key for the range.

pdillinger (approved):
👍

PR #7763, opened 2020-12-09
Test was previously Linux only, however it can also be enabled on macOS

adamretter:
It seems `db_logical_block_size_cache_test` can be enabled for macOS as well.

### DIFF
--- a/env/io_posix.cc
+++ b/env/io_posix.cc
@@ -371,7 +371,7 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
 }
 #endif
 
-#ifdef OS_LINUX
+#if defined(OS_LINUX) || defined(OS_MACOSX)
 std::string RemoveTrailingSlash(const std::string& path) {
   std::string p = path;
   if (p.size() > 1 && p.back() == '/') {
--- a/db/db_logical_block_size_cache_test.cc
+++ b/db/db_logical_block_size_cache_test.cc
@@ -5,7 +5,7 @@
 
 #include "test_util/testharness.h"
 
-#ifdef OS_LINUX
+#if defined(OS_LINUX) || defined(OS_MACOSX)
 #include "env/io_posix.h"
 #include "rocksdb/db.h"
 #include "rocksdb/env.h"
--- a/env/io_posix.h
+++ b/env/io_posix.h
@@ -49,7 +49,7 @@ class PosixHelper {
                                                size_t* size);
 };
 
-#ifdef OS_LINUX
+#if defined(OS_LINUX) || defined(OS_MACOSX)
 // Files under a specific directory have the same logical block size.
 // This class caches the logical block size for the specified directories to
 // save the CPU cost of computing the size.


env/io_posix.cc
```diff
@@ -371,7 +371,7 @@ size_t PosixHelper::GetUniqueIdFromFile(int fd, char* id, size_t max_size) {
 }
 #endif
 
-#ifdef OS_LINUX
+#if defined(OS_LINUX) || defined(OS_MACOSX)
```
mrambacher, line 374: What about the #ifdef OS_LINUX around line 468?  Doesn't this one need changed as well?

mrambacher (commented):
Also, shouldn't the code in `fs_posix.cc` that creates and uses a cache be changed to use macOS as well?

env/io_posix.cc
adamretter: > Also, shouldn't the code in fs_posix.cc that creates and uses a cache be changed to use macOS as well?

@mrambacher Can you be a bit more specific... do you mean `logical_block_size_cache_`?

mrambacher: Yes, I meant the logical_block_size_cache_.

If it turns out a bunch of places are using this same defined() rule, it might be worth introducing a define ROCKSDB_SUPPORTS_BLOCK_SIZE (or some other descriptive name) define.  Define it in io_posix.h and use it whenever the check is needed.

pdillinger:
And does this do anything useful if `GetLogicalBlockSizeOfFd` is not meaningfully implemented on MacOS?

adamretter:
@pdillinger @mrambacher So I did quite some digging into this. The current implementation of `PosixHelper::GetLogicalBlockSizeOfFd` actually isn't really a "POSIX" implementation, instead it is strictly Linux only and relies on a modern Linux kernel with sysfs implementation.

Whilst thee is a POSIX compliant API function `ioctl` which can be called on Linux, FreeBSD, or macOS (albeit with slightly different arguments on each platform), it unfortunately would need to open a file-descriptor inside `/dev` which would require either root-access or per-system specific permission changes. So I think the `ioctl` approach is not suitable here either.

| ioctl Property | Sensible Default | Linux | Mac OS | FreeBSD |
|----------|-----------------|-------|---------|----------|
| Logical Block Size   | 512    | BLKSSZGET   | DKIOCGETBLOCKSIZE                   | DIOCGSECTORSIZE |
| Physical Block Size | 4096 | BLKPBSZGET | DKIOCGETPHYSICALBLOCKSIZE | DIOCGSTRIPESIZE |


For macOS, this could be reimplemented using IOKit. I have done some experimenting with this and could likely do it, but it will take some time, as the documentation/examples for macOS IOKit appears to be pretty poor. I wonder if you think it is worth it or not?

For other systems such as Free/Open/Net BSD it could likely be re-implemented using procfs. For Windows, I suspect there are various Win32 API that could be called.

We can either close this, or do platform specific work as described above. Let me know which way you would like me to go...

mrambacher:
Does the Posix statvfs function return the same data: https://stackoverflow.com/questions/54823541/what-do-f-bsize-and-f-frsize-in-struct-statvfs-stand-for

We already use statvfs to determine free space of a file system for Posix.  Not sure why we would/cloud not use it for the block size, though I have not experimented to see if the same values are returned.

adamretter:
> Does the Posix statvfs function return the same data: https://stackoverflow.com/questions/54823541/what-do-f-bsize-and-f-frsize-in-struct-statvfs-stand-for

Sadly not (at least on Linux)... Luckily, I had written some test code for that too; here are the results:

```
statvfs.f_bsize=4096
statvfs.f_frsize=4096
statfs.f_bsize=4096
statfs.f_frsize=4096
ioctl(BLKGETSIZE64)=64308100608
ioctl(BLKSSZGET)=512
ioctl(BLKBSZGET)=4096
ioctl(BLKPBSZGET)=512
```

The current sysfs implementation in RocksDB performs an equivalent of:
```
$ cat /sys/devices/pci0000:00/0000:00:04.0/virtio1/block/vda/queue/logical_block_size

512
```

pdillinger:
I think this is low-priority unless someone really wants it. Filed #7830

PR #7765, opened 2020-12-10
Do not log unnecessary WAL obsoletion events

cheng-chang:
min_wal_number_to_keep should not be decreasing, if it does not increase, then there is no need to log the WAL obsoletions in MANIFEST since a previous one has been logged.

Test Plan:
watch existing tests and stress tests to pass


pdillinger (approved):
👍

PR #7767, opened 2020-12-11
Log sst number in Corruption status

jay-zhuang:
sst file number in corruption error would be very useful for debugging


zhichao-cao (approved):
LGTM, thanks!

PR #7769, opened 2020-12-13
Fix failing RocksJava test compilation and add CI

adamretter:
* Fixes a Java test compilation issue on macOS
* Cleans up CircleCI RocksDBJava build config
* Adds CircleCI for RocksDBJava on MacOS
* Ensures backwards compatibility with older macOS via CircleCI
* Fixes RocksJava static builds ordering
* Adds missing RocksJava static builds to CircleCI for Mac and Linux
* Improves parallelism in RocksJava builds
* Reduces the size of the machines used for RocksJava CircleCI as they don't need to be so large (Saves credits)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -370,6 +370,26 @@ jobs:
             make V=1 J=32 -j32 rocksdbjava jtest | .circleci/cat_ignore_eagain
       - post-steps
 
+  build-macos-java:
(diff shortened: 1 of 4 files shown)


pdillinger (approved):
LGTM (at least until we get pushback on number of CircleCI builds triggered by each PR)

PR #7774, opened 2020-12-16
Support optimize_filters_for_memory for Ribbon filter

pdillinger:
Summary: Primarily this change refactors the optimize_filters_for_memory
code for Bloom filters, based on malloc_usable_size, to also work for
Ribbon filters.

This change also replaces the somewhat slow but general
BuiltinFilterBitsBuilder::ApproximateNumEntries with
implementation-specific versions for Ribbon (new) and Legacy Bloom
(based on a recently deleted version). The reason is to emphasize
speed in ApproximateNumEntries rather than 100% accuracy.

Justification: ApproximateNumEntries (formerly CalculateNumEntry) is
only used by RocksDB for range-partitioned filters, called each time we
start to construct one. (In theory, it should be possible to reuse the
estimate, but the abstractions provided by FilterPolicy don't really
make that workable.) But this is only used as a heuristic estimate for
hitting a desired partitioned filter size because of alignment to data
blocks, which have various numbers of unique keys or prefixes. The two
factors lead us to prioritize reasonable speed over 100% accuracy.

optimize_filters_for_memory adds extra complication, because precisely
calculating num_entries for some allowed number of bytes depends on state
with optimize_filters_for_memory enabled. And the allocator-agnostic
implementation of optimize_filters_for_memory, using malloc_usable_size,
means we would have to actually allocate memory, many times, just to
precisely determine how many entries (keys) could be added and stay below
some size budget, for the current state. (In a draft, I got this
working, and then realized the balance of speed vs. accuracy was all
wrong.)

So related to that, I have made CalculateSpace, an internal-only API
only used for testing, non-authoritative also if
optimize_filters_for_memory is enabled. This simplifies some code. 

Test Plan: unit test updated, and for FilterSize test, range of tested
values is greatly expanded (still super fast)

Also tested `db_bench -benchmarks=fillrandom,stats -bloom_bits=10 -num=1000000 -partition_index_and_filters -format_version=5 [-optimize_filters_for_memory] [-use_ribbon_filter]` with temporary debug output of generated filter sizes.

Bloom+optimize_filters_for_memory:

      1 Filter size: 197 (224 in memory)
    134 Filter size: 3525 (3584 in memory)
    107 Filter size: 4037 (4096 in memory)
    Total on disk: 904,506
    Total in memory: 918,752

Ribbon+optimize_filters_for_memory:

      1 Filter size: 3061 (3072 in memory)
    110 Filter size: 3573 (3584 in memory)
     58 Filter size: 4085 (4096 in memory)
    Total on disk: 633,021 (-30.0%)
    Total in memory: 634,880 (-30.9%)

Bloom (no offm):

      1 Filter size: 261 (320 in memory)
      1 Filter size: 3333 (3584 in memory)
    240 Filter size: 3717 (4096 in memory)
    Total on disk: 895,674 (-1% on disk vs. +offm; known tolerable overhead of offm)
    Total in memory: 986,944 (+7.4% vs. +offm)

Ribbon (no offm):

      1 Filter size: 2949 (3072 in memory)
      1 Filter size: 3381 (3584 in memory)
    167 Filter size: 3701 (4096 in memory)
    Total on disk: 624,397 (-30.3% vs. Bloom)
    Total in memory: 690,688 (-30.0% vs. Bloom)

Note that optimize_filters_for_memory is even more effective for Ribbon filter than for cache-local Bloom, because it can close the unused memory gap even tighter than Bloom filter, because of 16 byte increments for Ribbon vs. 64 byte increments for Bloom.

### DIFF
--- a/table/block_based/filter_policy.cc
+++ b/table/block_based/filter_policy.cc
@@ -36,7 +36,8 @@ Slice FinishAlwaysFalse(std::unique_ptr<const char[]>* /*buf*/) {
 // also known as Hash64 or GetSliceHash64.
 class XXH3pFilterBitsBuilder : public BuiltinFilterBitsBuilder {
  public:
-  XXH3pFilterBitsBuilder(std::atomic<int64_t>* aggregate_rounding_balance)
+  explicit XXH3pFilterBitsBuilder(
+      std::atomic<int64_t>* aggregate_rounding_balance)
       : aggregate_rounding_balance_(aggregate_rounding_balance) {}
 
   ~XXH3pFilterBitsBuilder() override {}
--- a/table/block_based/filter_policy.cc
+++ b/table/block_based/filter_policy.cc
@@ -561,8 +561,7 @@ class Standard128RibbonBitsBuilder : public XXH3pFilterBitsBuilder {
 
     // Find a slight under-estimate for actual average bits per slot
     double min_real_bits_per_slot;
-    if (desired_one_in_fp_rate_ >=
-        double{std::numeric_limits<uint32_t>::max()}) {
+    if (desired_one_in_fp_rate_ >= 1.0 + std::numeric_limits<uint32_t>::max()) {
       // Max of 32 solution columns (result bits)
       min_real_bits_per_slot = 32.0;
     } else {
--- a/table/block_based/filter_policy.cc
+++ b/table/block_based/filter_policy.cc
@@ -27,6 +27,15 @@ namespace ROCKSDB_NAMESPACE {
 
 namespace {
 
+// Metadata trailer size for built-in filters. (This is separate from
+// block-based table block trailer.)
+//
+// Originally this was 1 byte for num_probes and 4 bytes for number of
+// cache lines in the Bloom filter, but now the first trailer byte is
+// usually an implementation marker and remaining 4 bytes have various
+// meanings.
+static constexpr uint32_t kMetadataLen = 5;
+
 Slice FinishAlwaysFalse(std::unique_ptr<const char[]>* /*buf*/) {
   // Missing metadata, treated as zero entries
   return Slice(nullptr, 0);
@@ -65,7 +74,7 @@ class XXH3pFilterBitsBuilder : public BuiltinFilterBitsBuilder {
   size_t AllocateMaybeRounding(size_t target_len_with_metadata,
                                size_t num_entries,
                                std::unique_ptr<char[]>* buf) {
-    size_t target_len = target_len_with_metadata - 5;
+    size_t target_len = target_len_with_metadata - kMetadataLen;
     assert(target_len < target_len_with_metadata);
 
     // Return value set to a default; overwritten in some cases
@@ -105,10 +114,12 @@ class XXH3pFilterBitsBuilder : public BuiltinFilterBitsBuilder {
         for (uint64_t maybe_len_rough :
              {uint64_t{3} * target_len / 4, uint64_t{13} * target_len / 16,
               uint64_t{7} * target_len / 8, uint64_t{15} * target_len / 16}) {
-          size_t maybe_len = RoundDownUsableSpace(maybe_len_rough);
-          double maybe_fp_rate = EstimatedFpRate(num_entries, maybe_len + 5);
+          size_t maybe_len_with_metadata =
+              RoundDownUsableSpace(maybe_len_rough + kMetadataLen);
+          double maybe_fp_rate =
+              EstimatedFpRate(num_entries, maybe_len_with_metadata);
           if (maybe_fp_rate <= for_balance_fp_rate) {
-            rv = maybe_len + /* metadata */ 5;
+            rv = maybe_len_with_metadata;
             rv_fp_rate = maybe_fp_rate;
             break;
           }
@@ -197,7 +208,7 @@ class FastLocalBloomBitsBuilder : public XXH3pFilterBitsBuilder {
         AllocateMaybeRounding(len_with_metadata, num_entries, &mutable_buf);
 
     assert(mutable_buf);
-    assert(len_with_metadata >= 5);
+    assert(len_with_metadata >= kMetadataLen);
 
     // Max size supported by implementation
     assert(len_with_metadata <= 0xffffffffU);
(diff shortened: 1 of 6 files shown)


jay-zhuang:
is ubsan failure related to this change? https://app.circleci.com/pipelines/github/facebook/rocksdb/5157/workflows/c7035437-ff38-4498-90e7-261577ecc84c/jobs/72893

table/block_based/filter_policy.cc
```diff
@@ -492,37 +486,145 @@ class Standard128RibbonBitsBuilder : public XXH3pFilterBitsBuilder {
...
+      min_real_bits_per_slot = upper_bits_per_key - portion_lower;
+      assert(min_real_bits_per_slot >= 0.0);
+      assert(min_real_bits_per_slot <= 32.0);
+    }
+
+    // An overestimate, but this should only be O(1) slots away from truth.
+    double max_slots = len_no_metadata * 8.0 / min_real_bits_per_slot;
+
+    // Let's not bother accounting for overflow to Bloom filter
+    if (max_slots > 1.44 * kMaxRibbonEntries) {
```
jay-zhuang, line 586: Why `1.44` here? Is it because the best Ribbon could do is `1.44x` of bloom? Which is `1/ln(2)` factor in bloom?

table/block_based/filter_policy.cc
```diff
@@ -492,37 +486,145 @@ class Standard128RibbonBitsBuilder : public XXH3pFilterBitsBuilder {
...
+    uint32_t slots = static_cast<uint32_t>(max_slots);
+    slots = SolnType::RoundUpNumSlots(slots);
+
+    // Assert that we have a valid upper bound on slots
+    assert(SolnType::GetBytesForOneInFpRate(
+               SolnType::RoundUpNumSlots(slots + 1), desired_one_in_fp_rate_,
+               /*rounding*/ 0) > len_no_metadata);
+
+    // Iterate up to a few times to rather precisely account for small effects
+    for (int i = 0; slots > 0 && i < 10; ++i) {
```
jay-zhuang, line 600: why `10`? How precise could it get?

table/block_based/filter_policy.cc
```diff
@@ -73,9 +59,112 @@ class XXH3pFilterBitsBuilder : public BuiltinFilterBitsBuilder {
...
+        double for_balance_fp_rate =
+            -balance / double{0x100000000} + target_fp_rate;
+
+        // To simplify, we just try a few modified smaller sizes. This also
+        // caps how much we vary filter size vs. target, to avoid outlier
+        // behavior from excessive variance.
+        for (uint64_t maybe_len_rough :
+             {uint64_t{3} * target_len / 4, uint64_t{13} * target_len / 16,
+              uint64_t{7} * target_len / 8, uint64_t{15} * target_len / 16}) {
+          size_t maybe_len = RoundDownUsableSpace(maybe_len_rough);
```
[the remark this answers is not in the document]
pdillinger: Nice catch :)

table/block_based/filter_policy.cc
```diff
@@ -136,22 +227,12 @@ class FastLocalBloomBitsBuilder : public XXH3pFilterBitsBuilder {
   }
 
   size_t ApproximateNumEntries(size_t bytes) override {
-    size_t bytes_no_meta = bytes >= 5u ? bytes - 5u : 0;
+    size_t bytes_no_meta = bytes >= 5u ? RoundDownUsableSpace(bytes) - 5u : 0;
```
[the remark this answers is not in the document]
pdillinger: `bytes` is probably not aligned. https://github.com/facebook/rocksdb/blob/04b3524ad00cb162020f5e56ae04d77ed1812bc9/table/block_based/block_based_table_builder.cc#L84

table/block_based/filter_policy.cc
pdillinger: That's what I was thinking, but I'll do better with GetNumSlotsFor95PctSuccess

table/block_based/filter_policy.cc
```diff
@@ -492,37 +486,145 @@ class Standard128RibbonBitsBuilder : public XXH3pFilterBitsBuilder {
...
+      // Effectively asking for 100% FP rate, or NaN etc.
+      // Note that NaN is neither < 1.0 nor > 1.0
+      return kMaxRibbonEntries;
+    }
+
+    // Find a slight under-estimate for actual average bits per slot
+    double min_real_bits_per_slot;
+    if (desired_one_in_fp_rate_ >= 1.0 + std::numeric_limits<uint32_t>::max()) {
+      // Max of 32 solution columns (result bits)
+      min_real_bits_per_slot = 32.0;
```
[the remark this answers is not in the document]
pdillinger: I can't really conceive of an application in RocksDB where it makes a difference. When I handle the production version of configuring Ribbon filters, I will consider what to do for "more accuracy than it can handle" case. I think our general approach is to sanitize things to reasonable bounds, though there's certainly a case for informing the user, such as with a logged warning.

table/block_based/filter_policy.cc
```diff
@@ -492,37 +498,152 @@ class Standard128RibbonBitsBuilder : public XXH3pFilterBitsBuilder {
...
+          SolnType::GetBytesForOneInFpRate(slots, desired_one_in_fp_rate_,
+                                           /*rounding*/ 0);
+      if (reqd_bytes <= len_no_metadata) {
+        break;  // done
+      }
+      if (i >= 2) {
+        // should have been enough iterations
+        assert(false);
+        break;
+      }
```
jay-zhuang, line 626: Just a question: why `2` rounds of iterations is enough here?

jay-zhuang (approved):
👍

table/block_based/filter_policy.cc
pdillinger: I think intuitively, CalculateSpace and its reverse engineered version can disagree on rounding in a few places, which means our initial estimate can be off by up to about 2 chunks of 128 slots. I made the bound as tight as the (expanded) unit test would allow, and observed that being off by 2 was much more rare than by 0 or 1. If the stress test finds a violation, that's good evidence I should expand the unit test even more. :)

PR #7776, opened 2020-12-16
Fix various small build issues

adamretter:
* Compatibility with older GCC.
* Compatibility with older jemalloc libraries.
* Remove Docker warning when building i686 binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2253,7 +2253,7 @@ rocksdbjavastaticreleasedocker: rocksdbjavastatic rocksdbjavastaticdockerx86 roc
 
 rocksdbjavastaticdockerx86:
 	mkdir -p java/target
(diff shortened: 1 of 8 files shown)


pdillinger:
Failure in testAllCallbacksInvocation(org.rocksdb.EventListenerTest) that is not present in master and repeated on re-run

pdillinger:
Now there's a breaking Java API change. That should be mentioned in PR description (to be commit message) and HISTORY.md.

adamretter:
@pdillinger Sorry which API change do you refer to and in which release was it established?

pdillinger:
> @pdillinger Sorry which API change do you refer to and in which release was it established?

https://github.com/facebook/rocksdb/pull/7776/commits/71ea824338959f04e05ed72b4edc88a254c35c43#diff-3e3d56fb4fc4f8b7bcad6defca457c3694f0e2231c0d8d627b2a06f0ab3fd6ec

adamretter:
@pdillinger As far as I can see the last release was 6.14.6 and that wasn't in there - https://github.com/facebook/rocksdb/tree/v6.14.6/java/src/main/java/org/rocksdb/

So how can it be an API change if it was never released before?

pdillinger:
This PR includes an API change by changing case of Java method names in *EventListener.java

adamretter:
> This PR includes an API change by changing case of Java method names in *EventListener.java

I understand that. However, my point was that that API and code had never been released before. So is it really an API change that needs documenting with respect to users of RocksDB. As the users cannot have used it before, IMHO it would rather seem like a new API that is about to be released for the first time.

pdillinger:
https://github.com/facebook/rocksdb/blob/6.15.fb/java/src/main/java/org/rocksdb/AbstractEventListener.java#L268

6.16 is current development (for the next ~day)

PR #7780, opened 2020-12-18
Delete WALs only after flush results are written to MANIFEST

cheng-chang:
In the write path, there is an optimization: when a new WAL is created during SwitchMemtable, we update the internal log number of the empty column families to the new WAL. `FindObsoleteFiles` marks a WAL as obsolete if the WAL's log number is less than `VersionSet::MinLogNumberWithUnflushedData`. After updating the empty column families' internal log number, `VersionSet::MinLogNumberWithUnflushedData` might change, so some WALs might become obsolete to be purged from disk.

For example, consider there are 3 column families: 0, 1, 2:
1. initially, all the column families' log number is 1;
2. write some data to cf0, and flush cf0;
3. now a new WAL 2 is created;
4. write data to cf1 and WAL 2;
5. now cf0's flush hasn't finished, flush cf1, and cf1's flush finishes, but since WAL 1 still contains data for the unflushed cf0, no WAL can be deleted from disk;
6. now cf0's flush finishes, so WAL 1 becomes obsolete, but WAL 2 is still active;
7. a new WAL 3 is created, and write data to cf2 and WAL 3, due to the optimization mentioned in the first paragraph, cf0 and cf1's log number is updated to 3 because they are empty;
8. now if the background threads want to purge obsolete files from disk, WAL 2 can be purged because no column family needs it. (Note that cf2 is not flushed yet.) 

When WAL tracking is enabled, we assume WALs will only become obsolete after a flush result is written to MANIFEST in `MemtableList::TryInstallMemtableFlushResults` (or its atomic flush counterpart). The above situation breaks this assumption.

This PR removes the above optimization, and ignores empty column families when computing the minimum log number to keep. So that WALs will only become obsolete after the flush results are written to MANIFEST.

Test Plan:
watch existing tests and stress tests to pass.

### DIFF
--- a/db/column_family.h
+++ b/db/column_family.h
@@ -350,6 +350,11 @@ class ColumnFamilyData {
 
   MemTableList* imm() { return &imm_; }
   MemTable* mem() { return mem_; }
+
+  bool IsEmpty() {
+    return mem()->GetFirstSequenceNumber() == 0 && imm()->NumNotFlushed() == 0;
+  }
+
   Version* current() { return current_; }
   Version* dummy_versions() { return dummy_versions_; }
   void SetCurrent(Version* _current);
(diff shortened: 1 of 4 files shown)


cheng-chang:
A lot of tests rely on the assumption of the removed optimization.

PR #7781, opened 2020-12-18
Track WAL obsoletion when updating empty CF's log number

cheng-chang:
In the write path, there is an optimization: when a new WAL is created during SwitchMemtable, we update the internal log number of the empty column families to the new WAL. `FindObsoleteFiles` marks a WAL as obsolete if the WAL's log number is less than `VersionSet::MinLogNumberWithUnflushedData`. After updating the empty column families' internal log number, `VersionSet::MinLogNumberWithUnflushedData` might change, so some WALs might become obsolete to be purged from disk.

For example, consider there are 3 column families: 0, 1, 2:
1. initially, all the column families' log number is 1;
2. write some data to cf0, and flush cf0, but the flush is pending;
3. now a new WAL 2 is created;
4. write data to cf1 and WAL 2, now cf0's log number is 1, cf1's log number is 2, cf2's log number is 2 (because cf1 and cf2 are empty, so their log numbers will be set to the highest log number);
5. now cf0's flush hasn't finished, flush cf1, a new WAL 3 is created, and cf1's flush finishes, now cf0's log number is 1, cf1's log number is 3, cf2's log number is 3, since WAL 1 still contains data for the unflushed cf0, no WAL can be deleted from disk;
6. now cf0's flush finishes, cf0's log number is 2 (because when cf0 was switching memtable, WAL 3 does not exist yet), cf1's log number is 3, cf2's log number is 3, so WAL 1 can be purged from disk now, but WAL 2 still cannot because `MinLogNumberToKeep()` is 2;
7. write data to cf2 and WAL 3, because cf0 is empty, its log number is updated to 3, so now cf0's log number is 3, cf1's log number is 3, cf2's log number is 3;
8. now if the background threads want to purge obsolete files from disk, WAL 2 can be purged because `MinLogNumberToKeep()` is 3. But there are only two flush results written to MANIFEST: the first is for flushing cf1, and the `MinLogNumberToKeep` is 1, the second is for flushing cf0, and the `MinLogNumberToKeep` is 2. So without this PR, if the DB crashes at this point and try to recover, `WalSet` will still expect WAL 2 to exist.

When WAL tracking is enabled, we assume WALs will only become obsolete after a flush result is written to MANIFEST in `MemtableList::TryInstallMemtableFlushResults` (or its atomic flush counterpart). The above situation breaks this assumption.

This PR tracks WAL obsoletion if necessary before updating the empty column families' log numbers.

Test Plan:
watch existing tests and stress tests to pass.
`make -j48 blackbox_crash_test` on devserver

### DIFF
--- a/db/db_impl/db_impl_write.cc
+++ b/db/db_impl/db_impl_write.cc
@@ -1819,17 +1819,64 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
     return s;
   }
 
-  for (auto loop_cfd : *versions_->GetColumnFamilySet()) {
-    // all this is just optimization to delete logs that
-    // are no longer needed -- if CF is empty, that means it
-    // doesn't need that particular log to stay alive, so we just
-    // advance the log number. no need to persist this in the manifest
-    if (loop_cfd->mem()->GetFirstSequenceNumber() == 0 &&
-        loop_cfd->imm()->NumNotFlushed() == 0) {
-      if (creating_new_log) {
-        loop_cfd->SetLogNumber(logfile_number_);
+  bool empty_cf_updated = false;
+  if (immutable_db_options_.track_and_verify_wals_in_manifest &&
+      !immutable_db_options_.allow_2pc && creating_new_log) {
+    // In non-2pc mode, WALs become obsolete if they do not contain unflushed
+    // data. Updating the empty CF's log number might cause some WALs to become
+    // obsolete. So we should track the WAL obsoletion event before actually
+    // updating the empty CF's log number.
+    uint64_t min_wal_number_to_keep =
+        versions_->PreComputeMinLogNumberWithUnflushedData(logfile_number_);
+    if (min_wal_number_to_keep >
+        versions_->GetWalSet().GetMinWalNumberToKeep()) {
+      // Get a snapshot of the empty column families.
+      // LogAndApply may release and reacquire db
+      // mutex, during that period, column family may become empty (e.g. its
+      // flush succeeds), then it affects the computed min_log_number_to_keep,
+      // so we take a snapshot for consistency of column family data
+      // status. If a column family becomes non-empty afterwards, its active log
+      // should still be the created new log, so the min_log_number_to_keep is
+      // not affected.
+      autovector<ColumnFamilyData*> empty_cfs;
+      for (auto cf : *versions_->GetColumnFamilySet()) {
+        if (cf->IsEmpty()) {
+          empty_cfs.push_back(cf);
+        }
       }
-      loop_cfd->mem()->SetCreationSeq(versions_->LastSequence());
+
+      VersionEdit wal_deletion;
+      wal_deletion.DeleteWalsBefore(min_wal_number_to_keep);
+      s = versions_->LogAndApplyToDefaultColumnFamily(&wal_deletion, &mutex_);
+      if (!s.ok() && versions_->io_status().IsIOError()) {
+        s = error_handler_.SetBGError(versions_->io_status(),
+                                      BackgroundErrorReason::kManifestWrite);
+      }
+      if (!s.ok()) {
+        return s;
+      }
+
+      for (auto cf : empty_cfs) {
+        if (cf->IsEmpty()) {
+          cf->SetLogNumber(logfile_number_);
+          cf->mem()->SetCreationSeq(versions_->LastSequence());
+        }  // cf may become non-empty.
+      }
+      empty_cf_updated = true;
+    }
+  }
+  if (!empty_cf_updated) {
+    for (auto cf : *versions_->GetColumnFamilySet()) {
+      // all this is just optimization to delete logs that
+      // are no longer needed -- if CF is empty, that means it
+      // doesn't need that particular log to stay alive, so we just
+      // advance the log number. no need to persist this in the manifest
+      if (cf->IsEmpty()) {
+        if (creating_new_log) {
+          cf->SetLogNumber(logfile_number_);
+        }
+        cf->mem()->SetCreationSeq(versions_->LastSequence());
+      }
     }
   }
 
--- a/db/column_family.h
+++ b/db/column_family.h
@@ -350,6 +350,11 @@ class ColumnFamilyData {
 
   MemTableList* imm() { return &imm_; }
   MemTable* mem() { return mem_; }
+
+  bool IsEmpty() {
+    return mem()->GetFirstSequenceNumber() == 0 && imm()->NumNotFlushed() == 0;
+  }
+
   Version* current() { return current_; }
   Version* dummy_versions() { return dummy_versions_; }
   void SetCurrent(Version* _current);
--- a/db/version_set.h
+++ b/db/version_set.h
@@ -1127,10 +1127,28 @@ class VersionSet {
     return PreComputeMinLogNumberWithUnflushedData(nullptr);
   }
   // Returns the minimum log number which still has data not flushed to any SST
+  // file.
+  // Empty column families' log number is considered to be
+  // new_log_number_for_empty_cf.
+  uint64_t PreComputeMinLogNumberWithUnflushedData(
+      uint64_t new_log_number_for_empty_cf) const {
+    uint64_t min_log_num = port::kMaxUint64;
+    for (auto cfd : *column_family_set_) {
+      // It's safe to ignore dropped column families here:
+      // cfd->IsDropped() becomes true after the drop is persisted in MANIFEST.
+      uint64_t num =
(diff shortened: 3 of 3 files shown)


cheng-chang:
In https://github.com/facebook/rocksdb/pull/7780, I tried to get rid of the optimization to update empty cf's log number during switch memtable so that WAL can only become obsolete after successful flush, but a lot of tests fail, so I assume there are code paths that assume the existence of this optimization.

ltamasi:
> For example, consider there are 3 column families: 0, 1, 2:
> 
> 1. initially, all the column families' log number is 1;
> 2. write some data to cf0, and flush cf0;
> 3. now a new WAL 2 is created;
> 4. write data to cf1 and WAL 2;
> 5. now cf0's flush hasn't finished, flush cf1, and cf1's flush finishes, but since WAL 1 still contains data for the unflushed cf0, no WAL can be deleted from disk;
> 6. now cf0's flush finishes, so WAL 1 becomes obsolete, but WAL 2 is still active;

I get what's going on up till step 5 but you lost me here. Flushing cf1 in step 5 would also involve a WAL switch, right? So WAL 2 only contains data from cf1, and since cf1's flush has finished, WAL 2 should also be obsolete, no?

> 7. a new WAL 3 is created, and write data to cf2 and WAL 3, due to the optimization mentioned in the first paragraph, cf0 and cf1's log number is updated to 3 because they are empty;
> 8. now if the background threads want to purge obsolete files from disk, WAL 2 can be purged because no column family needs it. (Note that cf2 is not flushed yet.)

cheng-chang:
> > For example, consider there are 3 column families: 0, 1, 2:
> > 
> > 1. initially, all the column families' log number is 1;
> > 2. write some data to cf0, and flush cf0;
> > 3. now a new WAL 2 is created;
> > 4. write data to cf1 and WAL 2;
> > 5. now cf0's flush hasn't finished, flush cf1, and cf1's flush finishes, but since WAL 1 still contains data for the unflushed cf0, no WAL can be deleted from disk;
> > 6. now cf0's flush finishes, so WAL 1 becomes obsolete, but WAL 2 is still active;
> 
> I get what's going on up till step 5 but you lost me here. Flushing cf1 in step 5 would also involve a WAL switch, right? So WAL 2 only contains data from cf1, and since cf1's flush has finished, WAL 2 should also be obsolete, no?
> 
> > 1. a new WAL 3 is created, and write data to cf2 and WAL 3, due to the optimization mentioned in the first paragraph, cf0 and cf1's log number is updated to 3 because they are empty;
> > 2. now if the background threads want to purge obsolete files from disk, WAL 2 can be purged because no column family needs it. (Note that cf2 is not flushed yet.)

Yes, but since cf0's flush hasn't finished, we still need to keep WAL 1, so WAL 2 cannot be purged at step 5.

ltamasi:
> Yes, but since cf0's flush hasn't finished, we still need to keep WAL 1, so WAL 2 cannot be purged at step 5.

That part is clear. What I don't get is that WAL 2 is still considered "active" in step 6.

db/db_impl/db_impl_open.cc
```diff
@@ -622,6 +622,15 @@ Status DBImpl::Recover(
           cfd->CreateNewMemtable(*cfd->GetLatestMutableCFOptions(),
                                  kMaxSequenceNumber);
         }
+      } else if (immutable_db_options_.track_and_verify_wals_in_manifest) {
```
cheng-chang, line 625: If the recovered memtables are flushed, then the WAL obsoletion event will be logged in MANIFEST when logging the flush.

But in https://github.com/facebook/rocksdb/blob/master/db/db_impl/db_impl_open.cc#L1192, if the recovered memtable is empty, the CF's log number will also be updated, which may cause some WALs to become obsolete due to the same reason as updating log number of empty WAL in db_impl_write.cc.

cheng-chang:
> > Yes, but since cf0's flush hasn't finished, we still need to keep WAL 1, so WAL 2 cannot be purged at step 5.
> 
> That part is clear. What I don't get is that WAL 2 is still considered "active" in step 6. I would expect both WAL 1 and WAL 2 to become obsolete once cf0's flush finishes.

I updated the description, hope it's clearer.

ltamasi:
Thanks for the clarification!

> When WAL tracking is enabled, we assume WALs will only become obsolete after a flush result is written to MANIFEST in `MemtableList::TryInstallMemtableFlushResults` (or its atomic flush counterpart). The above situation breaks this assumption.

An alternative approach would be making sure the assumption holds even in the situation described above. I'm wondering if we could do that by bumping the flushed CF's log number when we log the `VersionEdit` for the flush? I.e. in step 6 above, we could update the log number of cf0 to 3 before we log the flush, right?

cheng-chang:
> Thanks for the clarification!
> 
> > When WAL tracking is enabled, we assume WALs will only become obsolete after a flush result is written to MANIFEST in `MemtableList::TryInstallMemtableFlushResults` (or its atomic flush counterpart). The above situation breaks this assumption.
> 
> An alternative approach would be making sure the assumption holds even in the situation described above. I'm wondering if we could do that by bumping the flushed CF's log number when we log the `VersionEdit` for the flush? I.e. in step 6 above, we could update the log number of cf0 to 3 before we log the flush, right?

For this specific case, the alternative approach may work. But in general, there might be other cases I didn't think of when updating the empty column family's log number, which may cause WAL deletion.

db/db_impl/db_impl_write.cc
```diff
@@ -1819,17 +1819,64 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
...
+    for (auto cf : *versions_->GetColumnFamilySet()) {
+      // all this is just optimization to delete logs that
+      // are no longer needed -- if CF is empty, that means it
+      // doesn't need that particular log to stay alive, so we just
+      // advance the log number. no need to persist this in the manifest
+      if (cf->IsEmpty()) {
+        if (creating_new_log) {
+          cf->SetLogNumber(logfile_number_);
+        }
+        cf->mem()->SetCreationSeq(versions_->LastSequence());
```
ltamasi, line 1878: I'm probably missing something but instead of all this can't we simply do

```
    for (auto cf : *versions_->GetColumnFamilySet()) {
      if (cf->IsEmpty()) {
        if (creating_new_log) {
          cf->SetLogNumber(logfile_number_);
        }
        cf->mem()->SetCreationSeq(versions_->LastSequence());
    }
    if (immutable_db_options_.track_and_verify_wals_in_manifest &&
      !immutable_db_options_.allow_2pc && creating_new_log) {
    uint64_t min_wal_number_to_keep =
        versions_->PreComputeMinLogNumberWithUnflushedData(logfile_number_);
    if (min_wal_number_to_keep >
        versions_->GetWalSet().GetMinWalNumberToKeep()) {
      VersionEdit wal_deletion;
      wal_deletion.DeleteWalsBefore(min_wal_number_to_keep);
      s = versions_->LogAndApplyToDefaultColumnFamily(&wal_deletion, &mutex_);
      // ...
    }
```

?

ltamasi: In other words, I'm not sure why this is a problem (from the code comments):

```
// Updating the empty CF's log number might cause some WALs to become
// obsolete.
```

ltamasi: ... and why we need `empty_cfs`.

db/db_impl/db_impl_open.cc
ltamasi: We could consider moving this chunk to `RecoverLogFiles` (which already logs `VersionEdit`s).

db/db_impl/db_impl_write.cc
cheng-chang: `LogAndApply` will release the db mutex while writing to MANIFEST, and afterwards reacquire the db mutex.
Suppose during that interval,  CF0's flush succeeds, then CF0 becomes empty.
So when we compute the min_log_number_to_keep, CF0 is not empty yet. But after we LogAndApply, CF0 becomes empty, which might make the previously computed min_log_number_to_keep incorrect.

cheng-chang: `empty_cfs` is to take a snapshot of the empty CFs before `LogAndApply`, so that after `LogAndApply`, we only update log numbers for the empty CFs in the snapshot, to ensure that the min_log_number_to_keep computed in `LogAndApply` is always correct even if some CFs become empty after `LogAndApply`.

ltamasi: Yeah I think I got it. You want to update the log number in the same set of column families that were empty when you first computed the log number to keep. Maybe that's a simpler way of phrasing it... :)

ltamasi (approved):

PR #7783, opened 2020-12-19
Fix block_cache_test failure

jay-zhuang:
`block_cache_tracer_test` and `block_cache_trace_analyzer_test` are using the same test directory. Which causes build failure if these 2 tests are running in parallel, for example: https://app.circleci.com/pipelines/github/facebook/rocksdb/5211/workflows/8639afbe-9fec-43e2-a6a4-6d47ea9cbcbe/jobs/74598/tests


akankshamahajan15 (approved):
LGTM! Thanks for the fix

PR #7786, opened 2020-12-20
Make StringEnv, StringSink, StringSource use FS classes

mrambacher:
Change the StringEnv and related classes to be based on FileSystem APIs rather than the corresponding Env ones.  The StringSink and StringSource classes were changed to be based on the corresponding FS file classes.

Part of a cleanup to use the newer interfaces.  This change also eliminates some of the casts/wrappers to LegacyFile classes.


anand1976 (approved):
LGTM. You'll need to rebase due to conflicts.

PR #7787, opened 2020-12-20
Flush and sync WAL before checkpointing

cheng-chang:
In a stress test failure, we observe that a WAL exists on disk, but when creating a checkpoint, the WAL is not linked to the checkpoint directory, even though the WAL is still active.

This is possible because if the WAL or WAL directory is not fsynced, there is no guarantee that ListDir will return all the on-disk WALs.

This PR fixes the problem by always `FlushWAL` and `SyncWAL` before checkpointing.

Test Plan:
watch existing tests to pass.

### DIFF
--- a/utilities/checkpoint/checkpoint_impl.cc
+++ b/utilities/checkpoint/checkpoint_impl.cc
@@ -115,6 +115,10 @@ Status CheckpointImpl::CreateCheckpoint(const std::string& checkpoint_dir,
     s = db_->DisableFileDeletions();
     const bool disabled_file_deletions = s.ok();
 
+    if (s.ok()) {
+      s = db_->FlushWAL(/*sync=*/true);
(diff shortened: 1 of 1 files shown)


ajkr:
> This is possible because if the WAL or WAL directory is not fsynced, there is no guarantee that ListDir will return all the on-disk WALs.

I thought it is guaranteed in the stress/crash test scenario since we don't do any power-loss or power-loss simulation, so I don't think unsynced data can be dropped.

cheng-chang:
But if the WAL is already tracked in the MANIFEST when creating the checkpoint, the WAL should have already been synced.

cheng-chang:
Re-analyzed the problem and found another cause, see https://github.com/facebook/rocksdb/pull/7789 for details.

PR #7788, opened 2020-12-20
SyncWAL shouldn't be supported in compacted db

cheng-chang:
`CompactedDB` is a kind of read-only DB, so it shouldn't support `SyncWAL`.

Test Plan:
watch existing tests to pass.


akankshamahajan15 (approved):
LGTM!

PR #7795, opened 2020-12-21
Fix Windows build in block_cache_tracer_test

jay-zhuang:
The test was added to cmake in #7783


akankshamahajan15 (approved):
Thanks for the fix!

PR #7800, opened 2020-12-22
Remove flaky, redundant, and dubious DBTest.SparseMerge

pdillinger:
Summary: This test would occasionally fail like this:

    WARNING: c:\users\circleci\project\db\db_test.cc(1343): error: Expected:
    (dbfull()->TEST_MaxNextLevelOverlappingBytes(handles_[1])) <= (20 * 1048576), actual: 33501540 vs 20971520

And being a super old test, it's not structured in a sound way. And it appears that DBTest2.MaxCompactionBytesTest is a better test of what SparseMerge was intended to test. In fact, SparseMerge fails if I set

    options.max_compaction_bytes = options.target_file_size_base * 1000;

Thus, we are removing this negative-value test.

Test Plan: Q.E.D.


ajkr (commented):
> I also fixed apparently missing column family in CompactRange calls.

This makes me worried it'll fail more because now it's actually testing something.

More seriously, though, I can't see what this is testing. My unverified understanding of what this test does is

- Creates ~50 x 2MB (the default `target_file_size_base`) files in some lower level
- Flushes a small file to L0. It overlaps everything but `TEST_MaxNextLevelOverlappingBytes()` only measures starting at L1+
- Issues L0->L1 compaction. It relies on default `max_compaction_bytes` (50MB) to "cut" output files to prevent future compactions (L1->L2) from being too large. So I'd expect it to produce ~2 L1 files, each overlapping ~50MB in L2.
- But then it goes on to assert the maximum overlap is just 20MB, which is not consistent with the above

ajkr:
> My unverified understanding of what this test does is

OK I verified and it's slightly different. The reason the test works right now is it only compacts the ~100MB data L0->L1, and the `TEST_MaxNextLevelOverlappingBytes()` calls never consider overlap of L0 file with L1.

If we instead call `MoveFilesToLevel(2 /* level */, 1 /* cf */);` after generating the ~100MB data, then it lands in L2. Then it is more aligned with what I described above. Except, the test actually still passes. This turned out to be because the output files are cut in a smarter way than expected. The first file contains key "A" only. The second file contains keys "B100" and "C". So the ~100MB of "B%010d" keys actually fall in the gap between files and thus overlap neither.

pdillinger:
@ajkr Do you agree this test is not appropriate for Universal compaction?

> This turned out to be because the output files are cut in a smarter way than expected. The first file contains key "A" only. The second file contains keys "B100" and "C".

If there's another test for that, would could always remove this one. Do you happen to know?

ajkr:
> @ajkr Do you agree this test is not appropriate for Universal compaction?

I'd actually still expect `max_compaction_bytes` to take effect and split files based on overlap. It generally wouldn't have much use because the unit of compaction picking is full sorted runs in automatic compactions. But now we are picking delete-triggered compaction (and maybe soon `CompactRange()`) at file granularity in universal compaction, which can benefit from proper file cutting.

> If there's another test for that, would could always remove this one. Do you happen to know?

Yeah, I'd also be fine with seeing this one removed considering its current lack of meaningful coverage. There's `DBTest2.MaxCompactionBytesTest`, not sure if it works but it looks like it at least attempts to test the logic for file cutting to prevent future huge compactions.

ajkr (approved):
LGTM!

PR #7812, opened 2020-12-24
Store test logs as artifacts if the build fails in CircleCI

adamretter:
If a workflow fails in CircleCI this will ensure that the `t/` directory is tar'd up and added to the workflow as an artifact. This allows us to download the detailed logs and see what went wrong.


jay-zhuang (approved):
👍

PR #7813, opened 2020-12-24
fix memory leak in db_stress checkpoint test

zhichao-cao:
fix memory leak in db_stress checkpoint test. If s is not ok, checkpoint is not deleted, may cause memory leak.

test plan: make asan_check

### DIFF
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -1505,11 +1505,11 @@ Status StressTest::TestCheckpoint(ThreadState* thread,
       }
     }
   }
+  delete checkpoint;
+  checkpoint = nullptr;
   std::vector<ColumnFamilyHandle*> cf_handles;
   DB* checkpoint_db = nullptr;
   if (s.ok()) {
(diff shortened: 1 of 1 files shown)


db_stress_tool/db_stress_test_base.cc
```diff
@@ -1563,6 +1563,8 @@ Status StressTest::TestCheckpoint(ThreadState* thread,
   if (!s.ok()) {
     fprintf(stderr, "A checkpoint operation failed with: %s\n",
             s.ToString().c_str());
+    delete checkpoint;
+    checkpoint = nullptr;
```
mrambacher, line 1567: Why not just move the delete checkpoint lines @1511 outside of the s.ok() block, either at the "top-level" or inside the previous s.ok() block (in which the checkpoint was successfully created)?

cheng-chang (approved):

db_stress_tool/db_stress_test_base.cc
zhichao-cao: Will do, thanks!

PR #7814, opened 2020-12-27
Disable BasicLockEscalation if cannot determine whether TSAN is enabled

cheng-chang:
BasicLockEscalation will cause false-positive warnings under TSAN (this is a known issue in TSAN, see details in https://gist.github.com/spetrunia/77274cf2d5848e0a7e090d622695ed4e), skip this test if TSAN is enabled, or if we are not sure whether TSAN is enabled.

Test Plan:
watch the tsan contrun test to pass.


zhichao-cao (approved):
LGTM, thanks!

PR #7816, opened 2020-12-28
Return Status from FilePrefetchBuffer::TryReadFromCache()

anand1976:
Return the Status from TryReadFromCache() in an argument to make it easier to report prefetch errors to the user.

Tests:
make crash_test
make check


akankshamahajan15 (approved):
Thanks for the fix. LGTM!

PR #7817, opened 2020-12-28
Fix db_bench duration for multireadrandom benchmark

anand1976:
The multireadrandom benchmark, when run for a specific number of reads (--reads argument), should base the duration on the actual number of keys read rather than number of batches.

Tests:
Run db_bench multireadrandom benchmark

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -5220,7 +5220,7 @@ class Benchmark {
     }
 
     Duration duration(FLAGS_duration, reads_);
-    while (!duration.Done(1)) {
+    while (!duration.Done(entries_per_batch_)) {
       DB* db = SelectDB(thread);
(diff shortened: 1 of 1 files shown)


tools/db_bench_tool.cc
```diff
@@ -5220,7 +5220,7 @@ class Benchmark {
     }
 
     Duration duration(FLAGS_duration, reads_);
-    while (!duration.Done(1)) {
```
zhichao-cao, line 5223: So the previous duration is wrong if one batch has more than 1 entries?

anand1976: Yes, previously the benchmark would run longer than the user intended.

zhichao-cao (approved):
LGTM, thanks for the fix!

PR #7824, opened 2020-12-31
Verify file checksum generator name

ajkr:
Previously we only had a debug assertion to check the right generator was being used for verification. However a user hit a problem in production where their factory was creating the wrong generator for some files, leading to checksum mismatches. It would have been easier to debug if we verified in optimized builds that the generator with the proper name is used. This PR adds such verification.

### DIFF
--- a/file/file_util.cc
+++ b/file/file_util.cc
@@ -124,8 +124,10 @@ bool IsWalDirSameAsDBPath(const ImmutableDBOptions* db_options) {
 }
 
 // requested_checksum_func_name brings the function name of the checksum
-// generator in checksum_factory. Checksum factories may use or ignore
-// requested_checksum_func_name.
+// generator in checksum_factory. Empty string is permitted, in which case the
+// name of the generator created by the factory is unchecked. When
+// `requested_checksum_func_name` is non-empty, however, the created generator's
+// name must match it, otherwise an `InvalidArgument` error is returned.
 IOStatus GenerateOneFileChecksum(
     FileSystem* fs, const std::string& file_path,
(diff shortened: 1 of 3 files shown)


file/file_util.cc
```diff
@@ -151,14 +151,21 @@ IOStatus GenerateOneFileChecksum(
         requested_checksum_func_name +
         " from checksum factory: " + checksum_factory->Name();
     return IOStatus::InvalidArgument(msg);
+  } else {
+    // For backward compatibility, requested_checksum_func_name can be empty.
```
zhichao-cao, line 155: Maybe add a sentence of comments on line 128 to indicate if requested function name does not match the one from factory, IOStatus::InvalidArgument is returned

zhichao-cao (approved):
LGTM, thanks for the fix!

file/file_util.cc
ajkr: Good idea, thanks. Done.

file/file_util.cc
```diff
@@ -151,14 +153,22 @@ IOStatus GenerateOneFileChecksum(
         requested_checksum_func_name +
         " from checksum factory: " + checksum_factory->Name();
     return IOStatus::InvalidArgument(msg);
+  } else {
+    // For backward compatibility and use in file ingestion clients where there
```
ajkr, line 157: I also updated this comment a bit since it looks like empty string is passed during file ingestion , which appears unrelated to compatibility.

ajkr (commented):
Thanks for the review!

PR #7825, opened 2020-12-31
fix thread status synchronization in thread_list_test

ajkr:
The test was flaky because the BG threads could increase
`running_count_` up to `job_count_` before applying their thread status
updates. Then the test thread would see non-deterministic results when
counting threads with each status. The fix is to acquire mutex in test
thread so it sees `running_count_` and thread status updated atomically.
I think simply reordering the two updates would have been insufficient
since the thread status update uses `memory_order_relaxed`. This change
happens to also eliminate an undesirable sleep loop.

Test Plan: injected sleeps to verify the failure repros before this PR and does not
repro after.

### DIFF
--- a/util/thread_list_test.cc
+++ b/util/thread_list_test.cc
@@ -38,6 +38,7 @@ class SimulatedBackgroundTask {
   void Run() {
     std::unique_lock<std::mutex> l(mutex_);
     running_count_++;
+    bg_cv_.notify_all();
     Env::Default()->GetThreadStatusUpdater()->SetColumnFamilyInfoKey(cf_key_);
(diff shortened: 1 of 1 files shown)


ajkr:
The test failure is unrelated:

```
g++: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-5/README.Bugs> for instructions.
make: *** [db/db_basic_test.o] Error 4
```

jay-zhuang (approved):
LGTM 👍

ajkr:
Thanks for the review!

PR #7834, opened 2021-01-05
Add more tests to the ASC pass list 

mrambacher:
Fixed the following  to now pass ASC checks:
* `ttl_test`
* `blob_db_test`
* `backupable_db_test`,
* `delete_scheduler_test`

### DIFF
--- a/file/delete_scheduler.cc
+++ b/file/delete_scheduler.cc
@@ -51,6 +51,9 @@ DeleteScheduler::~DeleteScheduler() {
   if (bg_thread_) {
     bg_thread_->join();
   }
+  for (const auto& it : bg_errors_) {
+    it.second.PermitUncheckedError();
+  }
 }
 
 Status DeleteScheduler::DeleteFile(const std::string& file_path,
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -216,7 +216,7 @@ class BackupEngineImpl : public BackupEngine {
     ~BackupMeta() {}
 
     void RecordTimestamp() {
-      env_->GetCurrentTime(&timestamp_);
+      env_->GetCurrentTime(&timestamp_).PermitUncheckedError();
     }
     int64_t GetTimestamp() const {
       return timestamp_;
@@ -402,6 +402,7 @@ class BackupEngineImpl : public BackupEngine {
                              std::string* db_session_id);
 
   struct CopyOrCreateResult {
+    ~CopyOrCreateResult() { status.PermitUncheckedError(); }
     uint64_t size;
     std::string checksum_hex;
     std::string db_id;
@@ -670,6 +671,9 @@ BackupEngineImpl::~BackupEngineImpl() {
     t.join();
   }
   LogFlush(options_.info_log);
+  for (const auto& it : corrupt_backups_) {
+    it.second.first.PermitUncheckedError();
+  }
 }
 
 Status BackupEngineImpl::Initialize() {
@@ -783,7 +787,11 @@ Status BackupEngineImpl::Initialize() {
     for (const auto& rel_dir :
          {GetSharedFileRel(), GetSharedFileWithChecksumRel()}) {
       const auto abs_dir = GetAbsolutePath(rel_dir);
-      InsertPathnameToSizeBytes(abs_dir, backup_env_, &abs_path_to_size);
+      auto s =
+          InsertPathnameToSizeBytes(abs_dir, backup_env_, &abs_path_to_size);
+      if (!s.ok()) {
+        return s;
+      }
     }
     // load the backups if any, until valid_backups_to_open of the latest
     // non-corrupted backups have been successfully opened.
@@ -801,11 +809,13 @@ Status BackupEngineImpl::Initialize() {
 
       // Insert files and their sizes in backup sub-directories
       // (private/backup_id) to abs_path_to_size
-      InsertPathnameToSizeBytes(
+      Status s = InsertPathnameToSizeBytes(
           GetAbsolutePath(GetPrivateFileRel(backup_iter->first)), backup_env_,
           &abs_path_to_size);
-      Status s = backup_iter->second->LoadFromFile(options_.backup_dir,
-                                                   abs_path_to_size);
+      if (s.ok()) {
+        s = backup_iter->second->LoadFromFile(options_.backup_dir,
+                                              abs_path_to_size);
+      }
       if (s.IsCorruption()) {
         ROCKS_LOG_INFO(options_.info_log, "Backup %u corrupted -- %s",
                        backup_iter->first, s.ToString().c_str());
@@ -986,7 +996,7 @@ Status BackupEngineImpl::CreateNewBackupWithMetadata(
 
   std::vector<BackupAfterCopyOrCreateWorkItem> backup_items_to_finish;
   // Add a CopyOrCreateWorkItem to the channel for each live file
-  db->DisableFileDeletions();
+  db->DisableFileDeletions().PermitUncheckedError();
   if (s.ok()) {
     CheckpointImpl checkpoint(db);
     uint64_t sequence_number = 0;
@@ -1091,7 +1101,7 @@ Status BackupEngineImpl::CreateNewBackupWithMetadata(
   }
 
   // we copied all the files, enable file deletions
-  db->EnableFileDeletions(false);
+  db->EnableFileDeletions(false).PermitUncheckedError();
 
   auto backup_time = backup_env_->NowMicros() - start_backup;
 
@@ -1133,7 +1143,7 @@ Status BackupEngineImpl::CreateNewBackupWithMetadata(
                    backup_statistics_.ToString().c_str());
     // delete files that we might have already written
     might_need_garbage_collect_ = true;
-    DeleteBackup(new_backup_id);
+    DeleteBackup(new_backup_id).PermitUncheckedError();
     return s;
   }
 
@@ -1201,6 +1211,7 @@ Status BackupEngineImpl::DeleteBackup(BackupID backup_id) {
   }
 
   if (!s1.ok()) {
+    s2.PermitUncheckedError();  // What to do?
     return s1;
   } else {
     return s2;
@@ -1229,6 +1240,7 @@ Status BackupEngineImpl::DeleteBackupInternal(BackupID backup_id) {
     if (!s.ok()) {
       return s;
     }
+    corrupt->second.first.PermitUncheckedError();
     corrupt_backups_.erase(corrupt);
   }
 
@@ -1310,8 +1322,8 @@ Status BackupEngineImpl::RestoreDBFromBackup(const RestoreOptions& options,
                  static_cast<int>(options.keep_log_files));
 
   // just in case. Ignore errors
-  db_env_->CreateDirIfMissing(db_dir);
-  db_env_->CreateDirIfMissing(wal_dir);
+  db_env_->CreateDirIfMissing(db_dir).PermitUncheckedError();
+  db_env_->CreateDirIfMissing(wal_dir).PermitUncheckedError();
 
   if (options.keep_log_files) {
     // delete files in db_dir, but keep all the log files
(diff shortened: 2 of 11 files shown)


ajkr (approved):
LGTM, thanks!

file/delete_scheduler.cc
```diff
@@ -51,6 +51,9 @@ DeleteScheduler::~DeleteScheduler() {
   if (bg_thread_) {
     bg_thread_->join();
   }
+  for (const auto& it : bg_errors_) {
+    it.second.PermitUncheckedError();
```
adamretter, line 55: I am just wondering about this. Doing it in the destructor makes me think that it is not important if a background error occurs or or not? Is that the case?

utilities/backupable/backupable_db.cc
```diff
@@ -216,7 +216,7 @@ class BackupEngineImpl : public BackupEngine {
     ~BackupMeta() {}
 
     void RecordTimestamp() {
-      env_->GetCurrentTime(&timestamp_);
+      env_->GetCurrentTime(&timestamp_).PermitUncheckedError();
```
adamretter, line 219: Perhaps better to return the Status from this function? It appears to only be used in one place, which itself is inside a function that can return Status.

utilities/backupable/backupable_db.cc
```diff
@@ -670,6 +671,9 @@ BackupEngineImpl::~BackupEngineImpl() {
     t.join();
   }
   LogFlush(options_.info_log);
+  for (const auto& it : corrupt_backups_) {
+    it.second.first.PermitUncheckedError();
```
adamretter, line 675: I have the same question about this being in the destructor as before

utilities/backupable/backupable_db.cc
```diff
@@ -986,7 +996,7 @@ Status BackupEngineImpl::CreateNewBackupWithMetadata(
 
   std::vector<BackupAfterCopyOrCreateWorkItem> backup_items_to_finish;
   // Add a CopyOrCreateWorkItem to the channel for each live file
-  db->DisableFileDeletions();
+  db->DisableFileDeletions().PermitUncheckedError();
```
adamretter, line 999: I don't know the implications of Enable/DisableFileDeletions failing. If it is okay to ignore them, can you add a comment?

utilities/backupable/backupable_db.cc
```diff
@@ -1201,6 +1211,7 @@ Status BackupEngineImpl::DeleteBackup(BackupID backup_id) {
   }
 
   if (!s1.ok()) {
+    s2.PermitUncheckedError();  // What to do?
```
adamretter, line 1214: Good question! It looks to me like `GarbageCollect` is optional here? If that is the case, then I think ignoring any `s2` error is fine. If the garbage collection is optional, maybe we should refactor slightly to:

```c++
  auto s1 = DeleteBackupInternal(backup_id);

  // Clean up after any incomplete backup deletion, potentially from
  // earlier session.
  if (might_need_garbage_collect_) {
    GarbageCollect().PermitUncheckedError();  // Garbage Collection is allowed to fail
  }

  if (!s1.ok()) {
    return s1;
  }
  return Status::OK();
```

utilities/backupable/backupable_db.cc
```diff
@@ -1710,7 +1725,8 @@ Status BackupEngineImpl::AddBackupFileWorkItem(
           "overwrite the file.",
           fname.c_str());
       need_to_copy = true;
-      backup_env_->DeleteFile(final_dest_path);
+      //**TODO: What to do on error?
+      backup_env_->DeleteFile(final_dest_path).PermitUncheckedError();
```
adamretter, line 1729: I am not certain... but what is the downside of returning the error here?

utilities/backupable/backupable_db_test.cc
```diff
@@ -408,7 +408,7 @@ class FileManager : public EnvWrapper {
   Status GetRandomFileInDir(const std::string& dir, std::string* fname,
                             uint64_t* fsize) {
     std::vector<FileAttributes> children;
-    GetChildrenFileAttributes(dir, &children);
+    GetChildrenFileAttributes(dir, &children).PermitUncheckedError();
```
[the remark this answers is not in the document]
adamretter: It looks like if GetChildrenFileAttributes returns an error, you can just return it (as two lines below the same thing happens).

utilities/backupable/backupable_db_test.cc
```diff
@@ -428,7 +428,7 @@ class FileManager : public EnvWrapper {
 
   Status DeleteRandomFileInDir(const std::string& dir) {
     std::vector<std::string> children;
-    GetChildren(dir, &children);
+    GetChildren(dir, &children).PermitUncheckedError();
```
adamretter, line 431: It looks like if GetChildrenFileAttributes returns an error, you can just return it (as two lines below the same thing happens).

utilities/backupable/backupable_db_test.cc
```diff
@@ -446,7 +446,7 @@ class FileManager : public EnvWrapper {
   Status AppendToRandomFileInDir(const std::string& dir,
                                  const std::string& data) {
     std::vector<std::string> children;
-    GetChildren(dir, &children);
+    GetChildren(dir, &children).PermitUncheckedError();
```
adamretter, line 449: It looks like if GetChildrenFileAttributes returns an error, you can just return it (as two lines below the same thing happens).

utilities/backupable/backupable_db_test.cc
```diff
@@ -640,7 +640,8 @@ class BackupableDBTest : public testing::Test {
     // Create logger
     DBOptions logger_options;
     logger_options.env = db_chroot_env_.get();
-    CreateLoggerFromOptions(dbname_, logger_options, &logger_);
+    CreateLoggerFromOptions(dbname_, logger_options, &logger_)
```
adamretter, line 643: Why not `EXPECT_OK` ?

utilities/backupable/backupable_db_test.cc
```diff
@@ -748,13 +749,13 @@ class BackupableDBTest : public testing::Test {
 
   void DeleteLogFiles() {
     std::vector<std::string> delete_logs;
-    db_chroot_env_->GetChildren(dbname_, &delete_logs);
+    db_chroot_env_->GetChildren(dbname_, &delete_logs).PermitUncheckedError();
```
adamretter, line 752: I would rather see either:

1. `if (db_chroot_env_->GetChildren(...).ok()) {`
2. return Status on error from the function and call ASSERT/EXPECT on the two calls to `DeleteLogFiles`

utilities/backupable/backupable_db_test.cc
```diff
@@ -748,13 +749,13 @@ class BackupableDBTest : public testing::Test {
...
     for (auto f : delete_logs) {
       uint64_t number;
       FileType type;
       bool ok = ParseFileName(f, &number, &type);
       if (ok && type == kWalFile) {
-        db_chroot_env_->DeleteFile(dbname_ + "/" + f);
+        db_chroot_env_->DeleteFile(dbname_ + "/" + f).PermitUncheckedError();
```
adamretter, line 758: See above

utilities/backupable/backupable_db_test.cc
```diff
@@ -2814,6 +2815,7 @@ TEST_P(BackupableDBTestWithParam, BackupUsingDirectIO) {
   // We use ChrootEnv underneath so the below line checks for direct I/O support
   // in the chroot directory, not the true filesystem root.
   if (!test::IsDirectIOSupported(test_db_env_.get(), "/")) {
+    ROCKSDB_GTEST_SKIP("Test requires Direct I/O Support");
```
adamretter, line 2818: Nice!

utilities/blob_db/blob_db_impl.cc
```diff
@@ -2055,7 +2055,7 @@ Status DestroyBlobDB(const std::string& dbname, const Options& options,
                                         : bdb_options.blob_dir;
 
   std::vector<std::string> filenames;
-  env->GetChildren(blobdir, &filenames);
+  env->GetChildren(blobdir, &filenames).PermitUncheckedError();
```
adamretter, line 2058: I think you should return the error here.

utilities/blob_db/blob_db_impl.cc
```diff
@@ -2068,7 +2068,7 @@ Status DestroyBlobDB(const std::string& dbname, const Options& options,
       }
     }
   }
-  env->DeleteDir(blobdir);
+  env->DeleteDir(blobdir).PermitUncheckedError();
```
adamretter, line 2071: Perhaps the error should also be returned here... I am not certain

adamretter (commented):
In general looks good. I think there are a few cases where `PermitUnchecked` could possibly be avoided.

utilities/backupable/backupable_db.cc
```diff
@@ -783,7 +787,11 @@ Status BackupEngineImpl::Initialize() {
     for (const auto& rel_dir :
          {GetSharedFileRel(), GetSharedFileWithChecksumRel()}) {
       const auto abs_dir = GetAbsolutePath(rel_dir);
-      InsertPathnameToSizeBytes(abs_dir, backup_env_, &abs_path_to_size);
+      auto s =
+          InsertPathnameToSizeBytes(abs_dir, backup_env_, &abs_path_to_size);
+      if (!s.ok()) {
```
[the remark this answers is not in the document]
mrambacher: Are you suggesting that 1) this status should be ignored or 2) the InsertPathnameToSizeBytes should check both NotFound and PathNotFound status codes or 3) something else entirely?

If there is an Env that does not follow the documented spec and returns the wrong value, then this (and likely other) code will likely end up broken.

file/delete_scheduler.cc
mrambacher: I cannot find any code that actually checks these errors outside of the tests, so I am thinking that background errors might not be that important.  But this is new code to me, so I might be missing something...

utilities/backupable/backupable_db.cc
mrambacher: There does not appear to be code that makes sure there are no corrupt backups prior to shutting down, hence the need for a check here.  I will add a comment

utilities/backupable/backupable_db.cc
mrambacher: They never currently fail.  Restructured the code slightly

utilities/backupable/backupable_db.cc
mrambacher: My assumption was a little different.  I assumed that we always needed to do the GarbageCollect, even if the DeleteBackupInternal failed.  If the Delete fails, we return its status code and ignore the code from GarbageCollect.  If the Delete succeeds, we return the code from GarbageCollect.

utilities/backupable/backupable_db.cc
mrambacher: I do not know.  I did not return an error, as I did not want to change the underlying functionality for fear of what might break.

utilities/backupable/backupable_db_test.cc
mrambacher: At least on my Mac, this fails regularly:
 NotFound: /private/tmp/rocksdbtest-501/backupable_db_14615238825541450748/tempdb: No such file or directory
Which is why I did not use EXPECT_OK.  I think there is another bug with the test, and I did not want to dig into it.

Added a comment stating that issue.

utilities/blob_db/blob_db_impl.cc
mrambacher: It is legal for this to return an error, for example, if the directory does not exist.

Changed it to only delete the children if this returns ok.

utilities/backupable/backupable_db.cc
mrambacher: I would rather not.  The caller of that code is pretty complicated and probably over-engineered.  Most likely a lot of the pre-conditions should be refactored for Status and prevent work items from being created to begin with.  But this seems like a separate engineering effort from the Status checks.

utilities/backupable/backupable_db.cc
```diff
@@ -402,6 +402,7 @@ class BackupEngineImpl : public BackupEngine {
                              std::string* db_session_id);
 
   struct CopyOrCreateResult {
+    ~CopyOrCreateResult() { status.PermitUncheckedError(); }
```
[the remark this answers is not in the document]
mrambacher: The BackupadbleDBTest.CorruptionsTest is the only test that fails with this commented out.  I think there are two potential issues:
1.  When the BackupEngineImpl is shutting down, the backup threads are stopped/joined.  These return a std::promise/future that contains a Status that is never read.  The BackupEngineImpl would need to read/ignore the last Status values.
2. If there are still items in the "files_to_copy_or_create_ channel, then these items will have unread Status as well.

I will add a comment as to what the reasons are for requiring the PermitUncheckedError here.

ajkr (approved):
Thanks for the comments/updates. I verified the suggested change to restore does allow us to avoid `PermitUncheckedError()` on the `CopyOrCreateResult::status`, though I wouldn't block this PR on it.

PR #7836, opened 2021-01-05
Get manifest size again after getting min_log_num during checkpoint

cheng-chang:
Currently, manifest size is determined before getting min_log_num.

But between getting manifest size and getting min_log_num, concurrently, a flush might succeed, which will write new records to manifest to make some WALs become outdated, then min_log_num will be correspondingly increased, but the new records in manifest will not be copied into the checkpoint because the manifest's size is determined before them, then the newly outdated WALs will still exist in the checkpoint's manifest, but they are not linked/copied to the checkpoint because their log number is < min_log_num, so a corruption of missing WAL will be reported when restoring from the checkpoint.

The ideal solution is to get manifest size and min_log_num together atomically in GetLiveFiles. But since GetLiveFiles is a public interface, modifying its parameters might break builds of extensions of DB outside of rocksdb. The current compromised solution is to call GetLiveFiles again after getting min_log_num.

Test Plan:
make crash_test

### DIFF
--- a/utilities/checkpoint/checkpoint_impl.cc
+++ b/utilities/checkpoint/checkpoint_impl.cc
@@ -235,26 +235,18 @@ Status CheckpointImpl::CreateCustomCheckpoint(
   if (!db_->GetIntProperty(DB::Properties::kMinLogNumberToKeep, &min_log_num)) {
     return Status::InvalidArgument("cannot get the min log number to keep.");
   }
-
-  if (s.ok() && db_options.allow_2pc) {
-    // We need to refetch live files with flush to handle this case:
-    // A previous 000001.log contains the prepare record of transaction tnx1.
-    // The current log file is 000002.log, and sequence_number points to this
-    // file.
-    // After calling GetLiveFiles(), 000003.log is created.
-    // Then tnx1 is committed. The commit record is written to 000003.log.
-    // Now we fetch min_log_num, which will be 3.
-    // Then only 000002.log and 000003.log will be copied, and 000001.log will
-    // be skipped. 000003.log contains commit message of tnx1, but we don't
-    // have respective prepare record for it.
-    // In order to avoid this situation, we need to force flush to make sure
-    // all transactions committed before getting min_log_num will be flushed
-    // to SST files.
-    // We cannot get min_log_num before calling the GetLiveFiles() for the
-    // first time, because if we do that, all the logs files will be included,
-    // far more than needed.
-    s = db_->GetLiveFiles(live_files, &manifest_file_size, flush_memtable);
-  }
+  // Between GetLiveFiles and getting min_log_num, flush might happen
+  // concurrently, so new WAL deletions might be tracked in MANIFEST. If we do
+  // not get the new MANIFEST size, the deleted WALs might not be reflected in
+  // the checkpoint's MANIFEST.
+  //
+  // If we get min_log_num before the above GetLiveFiles, then there might
+  // be too many unnecessary WALs to be included in the checkpoint.
+  //
+  // Ideally, min_log_num should be got together with manifest_file_size in
+  // GetLiveFiles atomically. But that needs changes to GetLiveFiles' signature
+  // which is a public API.
+  s = db_->GetLiveFiles(live_files, &manifest_file_size, flush_memtable);
 
   TEST_SYNC_POINT("CheckpointImpl::CreateCheckpoint:SavedLiveFiles1");
   TEST_SYNC_POINT("CheckpointImpl::CreateCheckpoint:SavedLiveFiles2");


utilities/checkpoint/checkpoint_impl.cc
```diff
@@ -229,13 +229,13 @@ Status CheckpointImpl::CreateCustomCheckpoint(
...
 
-  // this will return live_files prefixed with "/"
-  s = db_->GetLiveFiles(live_files, &manifest_file_size, flush_memtable);
-
   if (!db_->GetIntProperty(DB::Properties::kMinLogNumberToKeep, &min_log_num)) {
     return Status::InvalidArgument("cannot get the min log number to keep.");
   }
 
+  // this will return live_files prefixed with "/"
+  s = db_->GetLiveFiles(live_files, &manifest_file_size, flush_memtable);
```
siying, line 237: If GetLiveFiles() is called after getting kMinLogNumberToKeep, how do we guarantee all files after min_log_num is preserved and returned by GetLiveFiles()?

cheng-chang: we do not need to make this guarantee. If some WALs after min_log_num  do not show up in GetLiveFiles(), then they won't show up in checkpoint anyway.

cheng-chang: we only make sure that the WALs that are "alive" according to MANIFEST really show up in the checkpoint.

utilities/checkpoint/checkpoint_impl.cc
```diff
@@ -253,6 +253,10 @@ Status CheckpointImpl::CreateCustomCheckpoint(
     // We cannot get min_log_num before calling the GetLiveFiles() for the
     // first time, because if we do that, all the logs files will be included,
     // far more than needed.
+    if (!db_->GetIntProperty(DB::Properties::kMinLogNumberToKeep,
+                             &min_log_num)) {
```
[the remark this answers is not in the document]
cheng-chang: it's tricky, the above existing comments explained the reasoning.
Without this, CheckpointTest.CurrentFileModifiedWhileCheckpointing2PC will fail.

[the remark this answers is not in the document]
siying: The above comment explains why GetLiveFiles() is called twice, not min_log_num is fetched twice, unless I missed anything.

utilities/checkpoint/checkpoint_impl.cc
siying: Oh GetLiveFiles() doesn't fetch log files. Ignore te question then.

utilities/checkpoint/checkpoint_impl.cc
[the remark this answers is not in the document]
cheng-chang: You are right. I proposed a new solution, PTAL.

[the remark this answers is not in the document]
cheng-chang: Since `GetLiveFiles` will do flush if flush is true, getting min_log_num before `GetLiveFiles` will cause the min_log_num to be unnecessarily small, but getting it after `GetLiveFiles` has the problem described in the PR's description. 
So I think the proper way is to get `min_log_num` and `manifest_size` while holding the mutex.

cheng-chang:
Since `GetLiveFiles` is a public interface, modifying its parameters might break builds of extensions of `DB` outside of rocksdb. I'm thinking of a better solution.

utilities/checkpoint/checkpoint_impl.cc
```diff
@@ -235,26 +235,18 @@ Status CheckpointImpl::CreateCustomCheckpoint(
...
-    // have respective prepare record for it.
-    // In order to avoid this situation, we need to force flush to make sure
-    // all transactions committed before getting min_log_num will be flushed
-    // to SST files.
-    // We cannot get min_log_num before calling the GetLiveFiles() for the
-    // first time, because if we do that, all the logs files will be included,
-    // far more than needed.
-    s = db_->GetLiveFiles(live_files, &manifest_file_size, flush_memtable);
-  }
+  // Between GetLiveFiles and getting min_log_num, flush might happen
```
cheng-chang, line 238: @siying This solution is a compromise. Actually, if we get min_log_num and manifest_file_size together atomically, we can also get rid of the original `if` block for `allow_2pc`.

utilities/checkpoint/checkpoint_impl.cc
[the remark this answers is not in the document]
siying: I think it is a good idea. We probably need to preserve the current interface and add a new one to DB. A lot of places need to change then.

Since it is getting a little bit deeper, maybe @pdillinger can take a quick look since he worked on checkpoint/backup recently and might have more opinions on it.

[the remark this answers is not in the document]
siying: @cheng-chang what would be the problem if min_log_num is smaller than needed? We just hardlink more files that might not be strictly needed, right? It shouldn't be a big problem?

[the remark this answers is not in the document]
cheng-chang: Yes, we just hardlink more files. But since it's a hardlink, it means those WALs will never be removed from disk as long as the checkpoint exists. Not sure whether this will be a concern.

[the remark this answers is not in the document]
cheng-chang: @siying  what's your opinion on the current solution of getting the live files twice?

siying:
We have always been getting live files twice, so there isn’t urgent to optimize.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

utilities/checkpoint/checkpoint_impl.cc
```diff
@@ -235,26 +235,18 @@ Status CheckpointImpl::CreateCustomCheckpoint(
...
+  // not get the new MANIFEST size, the deleted WALs might not be reflected in
+  // the checkpoint's MANIFEST.
+  //
+  // If we get min_log_num before the above GetLiveFiles, then there might
+  // be too many unnecessary WALs to be included in the checkpoint.
+  //
+  // Ideally, min_log_num should be got together with manifest_file_size in
+  // GetLiveFiles atomically. But that needs changes to GetLiveFiles' signature
+  // which is a public API.
+  s = db_->GetLiveFiles(live_files, &manifest_file_size, flush_memtable);
```
siying, line 249: It feels that the previous GetLiveFiles() is only for a flush. Can we issue a flush instead (if flush_memtable is true), and here do a non flush GetLiveFiles()?

cheng-chang: Yes, we can do a flush, but then we may need to duplicate [this code block](https://github.com/facebook/rocksdb/blob/master/db/db_filesnapshot.cc#L33), I think it's unnecessary.

utilities/checkpoint/checkpoint_impl.cc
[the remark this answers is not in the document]
ajkr: > what would be the problem if min_log_num is smaller than needed? We just hardlink more files that might not be strictly needed, right? It shouldn't be a big problem?

I thought WALs are copied.

> I think it is a good idea. We probably need to preserve the current interface and add a new one to DB. A lot of places need to change then.

Agree it is a good idea. Is the DB interface supposed to be extensible? If not I think we can add a pointer-to-int argument that defaults to nullptr, preserving API compatibility. If it is supposed to be extensible, we can instead add an overload that takes an additional pointer-to-int argument. Then the old overload can call the new overload passing `nullptr` for that argument. I may be missing the main difficulty with that approach.

utilities/checkpoint/checkpoint_impl.cc
ajkr: I wish we could do this too, but it seems there is no DB API like `FlushAll()`, and checkpoint doesn't have access to any `ColumnFamilyHandle*`s.

ajkr (approved):
LGTM.

utilities/checkpoint/checkpoint_impl.cc
cheng-chang: Maybe we can refactor out the flush logic in `GetLiveFiles` into a new `FlushAll`  API in `DB`.

siying, line 249: Sorry for being late, but I'm a little bit worried about triggering another flush soon after the previous one, causing more unexpected stalling to some users. Is it possible to change from flush_memtable to false?

PR #7837, opened 2021-01-05
 Add new API to report dummy entries size in cache in WriteBufferManager

akankshamahajan15:
Summary: Add new API WriteBufferManager::dummy_entries_in_cache_usage() which reports the dummy entries size stored in cache to account for DataBlocks in WriteBufferManager.

Test Plan: Updated test ./write_buffer_manager_test

Reviewers:

Subscribers:

Tasks:

Tags:


akankshamahajan15:
I will update the History.md once all test passes.

ajkr (approved):
LGTM, thanks! Do you want to say the improvement in "Public API Changes" section of HISTORY.md? It will be great that people can start distinguishing block cache usage for writers vs. readers

PR #7838, opened 2021-01-06
Fixed the swallowed IOStatus in Compaction Job introduced in PR 7718

zhichao-cao:
The IOStatus of TableBuilder is returned by copy the io status from builder->io_status(). pr #7718 swallowed the io status and it will cause the write IO error become non-retryable and no auto resume logic will handle it. Roll back to previous implementation.

test plan: make check

### DIFF
--- a/db/compaction/compaction_job.cc
+++ b/db/compaction/compaction_job.cc
@@ -1449,8 +1449,9 @@ Status CompactionJob::FinishCompactionOutputFile(
   } else {
     sub_compact->builder->Abandon();
   }
+  IOStatus io_s = sub_compact->builder->io_status();
   if (s.ok()) {
-    s = sub_compact->builder->io_status();
+    s = io_s;
   }
   const uint64_t current_bytes = sub_compact->builder->FileSize();
   if (s.ok()) {
@@ -1461,7 +1462,6 @@ Status CompactionJob::FinishCompactionOutputFile(
   sub_compact->total_bytes += current_bytes;
 
   // Finish and check for file errors
-  IOStatus io_s;
   if (s.ok()) {
     StopWatch sw(env_, stats_, COMPACTION_OUTFILE_SYNC_MICROS);
     io_s = sub_compact->outfile->Sync(db_options_.use_fsync);


db/compaction/compaction_job.cc
```diff
@@ -1449,8 +1449,9 @@ Status CompactionJob::FinishCompactionOutputFile(
   } else {
     sub_compact->builder->Abandon();
   }
+  IOStatus io_s = sub_compact->builder->io_status();
   if (s.ok()) {
```
ajkr, line 1453: I find it kind of confusing that this same conditional appears not far below, although it seems correct - https://github.com/facebook/rocksdb/blob/ad10d67b2853d3d31f6bc083c7c5a097363148d4/db/compaction/compaction_job.cc#L1480-L1482

It also feels confusing there's a comment later saying:

```
    // Since this error is really a copy of the
    // "normal" status, it does not also need to be checked
```

But actually `s` might be a different error altogether if it was encountered before we observed a non-ok `IOStatus`

ajkr (approved):
I couldn't really figure out the existing code so left some questions for whoever knows it well. But this change LGTM either way.

db/compaction/compaction_job.cc
zhichao-cao: Hi Andrew, thanks for the review. The logic is Status S is the overall status, IOStatus io_s only contains the IO related Status. So io_s is a sub-value/partial-value of s. What we need to ensure is 1) Status s bring the earliest error information, 2) any non-ok status in io_s should be stored in s if it is the first error. Therefore, the 
```
if (s.ok()) { 
   s = io_s; 
 } 
```

Appears in multiple locations once io_s has new value.

zhichao-cao:
> I couldn't really figure out the existing code so left some questions for whoever knows it well. But this change LGTM either way.

The final information in io_s will be stored in the subcompaction->io_status. (https://github.com/facebook/rocksdb/pull/7838/files#diff-6862dde35f8188f1c00a45681b295230026a98e313f455f8d1a67a4d792ff725R1484)

PR #7840, opened 2021-01-07
Treat File Scope Write IO Error the same as Retryable IO Error

zhichao-cao:
In RocksDB, when IO error happens, the flags of IOStatus can be set. If the IOStatus is set as "File Scope IO Error", it indicate that the error is constrained in the file level. Since RocksDB does not continues write data to a file when any IO Error happens, File Scope IO Error can be treated the same as Retryable IO Error. Adding the logic to ErrorHandler::SetBGError to include the file scope IO Error in its error handling logic, which is the same as retryable IO Error.

test plan: added new unit tests in error_handler_fs_test. make check


anand1976 (approved):
LGTM

PR #7847, opened 2021-01-09
Make ldb load column family options from OPTIONS file

anand1976:
When the --try_load_options is used in conjunction with the
--column_family option, ldb incorrectly sets the ColumnFamilyOptions for
that column family to defaults. This PR fixes that by retaining from the
OPTIONS file and applying command line overrides.

Test Plan:
Add a unit test in ldb_cmd_test

### DIFF
--- a/tools/ldb_cmd.cc
+++ b/tools/ldb_cmd.cc
@@ -1953,6 +1953,7 @@ void ReduceDBLevelsCommand::Help(std::string& ret) {
 
 void ReduceDBLevelsCommand::OverrideBaseCFOptions(
     ColumnFamilyOptions* cf_opts) {
+  LDBCommand::OverrideBaseCFOptions(cf_opts);
   cf_opts->num_levels = old_levels_;
   cf_opts->max_bytes_for_level_multiplier_additional.resize(cf_opts->num_levels,
                                                             1);
@@ -2106,6 +2107,7 @@ void ChangeCompactionStyleCommand::Help(std::string& ret) {
 
 void ChangeCompactionStyleCommand::OverrideBaseCFOptions(
     ColumnFamilyOptions* cf_opts) {
+  LDBCommand::OverrideBaseCFOptions(cf_opts);
   if (old_compaction_style_ == kCompactionStyleLevel &&
       new_compaction_style_ == kCompactionStyleUniversal) {
     // In order to convert from level compaction to universal compaction, we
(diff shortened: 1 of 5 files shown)


tools/ldb_cmd.cc
```diff
@@ -740,7 +744,7 @@ void LDBCommand::PrepareOptions() {
           "Non-existing column family " + column_family_name_);
       return;
     }
-    column_families_iter->options = options_;
```
ajkr, line 743: oops, thanks for fixing this!!

tools/ldb_cmd.cc
```diff
@@ -1947,14 +1951,14 @@ void ReduceDBLevelsCommand::Help(std::string& ret) {
   ret.append("\n");
 }
 
-void ReduceDBLevelsCommand::OverrideBaseOptions() {
-  LDBCommand::OverrideBaseOptions();
-  options_.num_levels = old_levels_;
-  options_.max_bytes_for_level_multiplier_additional.resize(options_.num_levels,
+void ReduceDBLevelsCommand::OverrideBaseCFOptions(
+    ColumnFamilyOptions* cf_opts) {
+  cf_opts->num_levels = old_levels_;
```
ajkr, line 1956: Do we want to still call the base class's `OverrideBaseCFOptions()` for option overrides common to `ldb` subcommands?

tools/ldb_cmd.cc
```diff
@@ -2100,18 +2104,17 @@ void ChangeCompactionStyleCommand::Help(std::string& ret) {
   ret.append("\n");
 }
 
-void ChangeCompactionStyleCommand::OverrideBaseOptions() {
-  LDBCommand::OverrideBaseOptions();
-
+void ChangeCompactionStyleCommand::OverrideBaseCFOptions(
+    ColumnFamilyOptions* cf_opts) {
```
ajkr, line 2108: same question as above

ajkr (approved):
LGTM

ajkr (commented):
Some day we should probably refactor even further and have options only in a `DBOptions` and in a `std::vector<ColumnFamilyDescriptor>`. When DB doesn't exist and there's no options file, we can prepopulate that vector with a default CF and modify its options there.

tools/ldb_cmd.cc
anand1976: Yes, good point!

PR #7848, opened 2021-01-09
Cleanup Travis CI config

adamretter:
We now only use Travis CI for testing RocksDB against Linux on:
* ppc64le
* arm64 (aarch64)

This is just some initial cleanup. I will add further ppc64le and arm64 jobs in a subsequent PR...


akankshamahajan15 (approved):
LGTM!

PR #7849, opened 2021-01-09
Fix checkpoint_test hang

jay-zhuang:
`CheckpointTest.CurrentFileModifiedWhileCheckpointing` could hang
because now create_checkpoint triggers flush twice. The test should wait
both flush done.

Test Plan:
`gtest-parallel ./checkpoint_test --gtest_filter=CheckpointTest.CurrentFileModifiedWhileCheckpointing -r 100`


ajkr (approved):
LGTM

PR #7851, opened 2021-01-10
Remove Legacy and Custom FileWrapper classes from header files

mrambacher:
Removed the uses of the Legacy FileWrapper classes from the source code.  The wrappers were creating an additional layer of indirection/wrapping, as the Env already has a FileSystem.

Moved the Custom FileWrapper classes into the CustomEnv, as these classes are really for the private use the the CustomEnv class.


table/#table_reader_bench.cc#
```diff
@@ -0,0 +1,346 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
```
anand1976, line 1: I guess this file should be removed from the PR?

anand1976:
LGTM. I think this can be merged once #7858 is merged. Any idea why the clang-analyze test is failing?

anand1976 (approved):

PR #7852, opened 2021-01-10
Small improvements to CircleCI Windows jobs

adamretter:
* Clearer indication of which versions of msbuild and Visual Studio is used
* Explicit naming of the build jobs within the Windows workflows

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -364,9 +364,13 @@ jobs:
       - run:
           name: "Setup VS"
           command: |
-            if [[ "${VS_YEAR}" == "2017" ]]; then
+            if [[ "${VS_YEAR}" == "2019" ]]; then
+              echo "VS2019 already present."
+            elif [[ "${VS_YEAR}" == "2017" ]]; then
+              echo "Installing VS2017..."
               powershell .circleci/vs2017_install.ps1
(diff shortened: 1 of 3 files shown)


jay-zhuang (approved):
👍

akankshamahajan15 (approved):

.circleci/config.yml
```diff
@@ -392,6 +396,7 @@ jobs:
             cd build
             ${CMAKE_BIN} -G "${CMAKE_GENERATOR}" -DCMAKE_BUILD_TYPE=Debug -DOPTDBG=1 -DPORTABLE=1 -DSNAPPY=1 -DJNI=1 << parameters.extra_cmake_opt >> ..
             cd ..
+            msbuild.exe /version
```
jay-zhuang, line 399: Seems msbuild.exe version is always the same for all builds (either 2015 or 2019, etc.), and it already displays the version information:
```
Microsoft (R) Build Engine version 16.8.2+25e4d540b for .NET Framework
```
I don't this line is necessary, and it's causing an error line:
```
MSBUILD : error MSB1009: Project file does not exist.
```
I think we should just change it to
```
echo "Building with VS version: ${CMAKE_GENERATOR}"
```

PR #7855, opened 2021-01-11
Add note for PR 7789 in history

cheng-chang:


jay-zhuang (approved):

ajkr (approved):
LGTM, thanks!

HISTORY.md
```diff
@@ -22,6 +22,7 @@
 * Fixed the logic of populating native data structure for `read_amp_bytes_per_bit` during OPTIONS file parsing on big-endian architecture. Without this fix, original code introduced in PR7659, when running on big-endian machine, can mistakenly store read_amp_bytes_per_bit (an uint32) in little endian format. Future access to `read_amp_bytes_per_bit` will give wrong values. Little endian architecture is not affected.
 * Fixed prefix extractor with timestamp issues.
 * Fixed a bug in atomic flush: in two-phase commit mode, the minimum WAL log number to keep is incorrect.
+* Fixed a bug related to checkpoint in PR7789: if there are multiple column families, and the checkpoint is not opened as read only, then in rare cases, data loss may happen in the checkpoint. Since backup engine relies on checkpoint, it may also be affected.
```
ajkr, line 25: Seems GitHub lost my comment (?). Trying again -

I didn't yet understand the safety for read-only DB users. Wouldn't they also see the hole since it was introduced during the checkpoint process? I can't think of any API they could use that'd let them only read data prior to the hole.

cheng-chang: > Seems GitHub lost my comment (?). Trying again -
> 
> I didn't yet understand the safety for read-only DB users. Wouldn't they also see the hole since it was introduced during the checkpoint process? I can't think of any API they could use that'd let them only read data prior to the hole.

I'm thinking of this case: Since the sequence number returned by CreateCheckpoint is retrieved before `GetLiveFiles`, if  the user opens the checkpoint in read-only mode, and uses that returned sequence number, then they may not see the hole.

ajkr: I can't seem to think of an API that allows reading from a sequence number earlier than the recovered sequence number.

Issue #7784, opened 2020-12-20
Rocksdb failed to recover after disk full error


deepakojha123:
Hi,

We saw an issue with rocksdb v6.4.6 where after disk full error rocksdb failed to recover and went into crashloop during WAL recovery. The error was as follows:
"Corruption: SST file is ahead of WALs"

Our use case: We open a single DB instance with multiple column-families and we do only sync writes. We use kPointInTimeRecovery mode. The rocksdb logs indicate a corrupted WAL file which was safely dropped.
2020/12/16-23:39:46.326488 52f0700 [WARN] [_impl/db_impl_open.cc:629] /home/cohesity/data/disks/nvme0n1p2_ce37060b-5269-4177-9a70-f61b46b8791d/scribe/newscribe/tablet_repo/tablet_1/db/253737849.log: dropping 266 bytes; Corruption: checksum mismatch
2020/12/16-23:39:46.326524 52f0700 [_impl/db_impl_open.cc:923] Point in time recovered to log #253737849 seq #653422636747

Listing the WAL files indicate another empty WAL file next to this corrupted file:
rw-rr-. 1 cohesity cohesity 175902 Dec 16 23:39 253737847.log
rw-rr-. 1 cohesity cohesity 0 Dec 16 23:39 253737851.log             <-----  Empty WAL file.
rw-rr-. 1 cohesity cohesity 340221 Dec 16 23:39 253737849.log

### Expected behavior

Once the disk space was cleared rocksdb should have safely dropped the truncated WAL file and recovered.

### Actual behavior

Rocksdb went into a crashloop with "Corruption: SST file is ahead of WALs"

### Steps to reproduce the behavior

I think below steps lead to the crashloop:
1. Crash rocksdb while writing the WAL file (the WAL file will be truncated). Please note another empty WAL file should have been created.
2. Upon restart during WAL recovery while doing a flush hit another crash. Please note before this point some of the column families(with no data to be recovered, would start pointing to the empty WAL file).
3. All subsequent WAL recovery would hit the error "Corruption: SST file is ahead of WALs".

I will go through the rocksdb unit test framework to create a test with above steps. Meanwhile I have attached the relevant LOG files.

[LOG.old.1608190304754106.txt](https://github.com/facebook/rocksdb/files/5720558/LOG.old.1608190304754106.txt)
[LOG.old.1608190774690063.txt](https://github.com/facebook/rocksdb/files/5720559/LOG.old.1608190774690063.txt)
[...]

PR #7859, opened 2021-01-13
Fix for a false corruption error during point-in-time WAL recovery.

deepakojha123:
Fix for a false corruption error during point-in-time WAL recovery.

The error is thrown when last-flushed-WAL file for any CF is more
recent than corrupted WAL file. The fix is to skip this error if
all WAL files more recent than the corrupted WAL file are empty.

Fixes issue #7784.

### DIFF
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -936,22 +936,6 @@ Status DBImpl::RecoverLogFiles(const std::vector<uint64_t>& wal_numbers,
         }
         if (stop_replay_for_corruption) {
           logFileDropped();
-          // Skip over any subsequent empty wal files and update corrupted wal
-          // number. This avoids false corruption error due to some column
-          // families pointing to empty wal file next to the corrupted wal
-          // file.
-          ++ii;
-          for (; ii < wal_numbers.size(); ++ii) {
-            wal_number = wal_numbers[ii];
-            std::string wal_file_name =
-                LogFileName(immutable_db_options_.wal_dir, wal_number);
-            uint64_t bytes;
-            const auto& s = env_->GetFileSize(wal_file_name, &bytes);
-            if (!s.ok() || bytes != 0) {
-              break;
-            }
-            corrupted_wal_number = wal_number;
-          }
           break;
         }
       }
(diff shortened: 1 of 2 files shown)


deepakojha123:
@ajkr, @anand1976 Any inputs/comments on this change?

db/db_impl/db_impl_open.cc
```diff
@@ -935,6 +936,22 @@ Status DBImpl::RecoverLogFiles(const std::vector<uint64_t>& wal_numbers,
         }
         if (stop_replay_for_corruption) {
           logFileDropped();
+          // Skip over any subsequent empty wal files and update corrupted wal
+          // number. This avoids false corruption error due to some column
+          // families pointing to empty wal file next to the corrupted wal
+          // file.
+          ++ii;
```
ajkr, line 943: Sorry it isn't obvious to me why we increment `ii` here. I thought we executed this logic only for WALs after the corrupted WAL, so we are already past it. Also it seems to execute for all non-empty WALs following the first corrupted WAL, in which case could this scenario increase `corrupted_wal_number` too far?

- WAL 1 is corrupt
- WAL 2 is empty
- WAL 3 is empty
- WAL 4 is non-empty
- WAL 5 is empty

I am not sure what prevents `corrupted_wal_number` from reaching 5, but feel it should be 3.

ajkr (commented):
Thanks for the PR and sorry for the delay!

ajkr (changes requested):
I think the logic would be correct if it's moved to the outer loop, around where we set `stop_replay_for_corruption = true;`. The only other thing is we need to make sure `ii` doesn't advance too far. Maybe keep `ii` trailing one behind the file whose emptiness we're checking.

deepakojha123:
@ajkr  Thanks for the review! Agreed the logic should be where we set stop_replay_for_corruption = true. As for corrupted_wal_number going too far: the loop breaks when it finds first non-empty WAL so ii should be set back to last empty WAL which was missing in my previous commit. I have a new commit, please see if this looks better?

deepakojha123:
db_wal_test failed with this change and upon debugging I think pointing corrupted_wal_number to next empty wal file has one hole, if a CF flush had happened with some keys from the original corrupted_wal_number we won't catch it?

[ RUN      ] DBWALTest.kPointInTimeRecoveryCFConsistency
db/db_wal_test.cc:1311: Failure
Value of: (TryReopenWithColumnFamilies({"default", "one", "two"}, options)).ok()
  Actual: true
Expected: false

ajkr:
> one hole, if a CF flush had happened with some keys from the original corrupted_wal_number we won't catch it?

This sounds right. Sorry I should have thought through it more carefully. The zero-byte WAL doesn't seem like the root cause (if you still have the original DB, does `Open()` still fail with the zero-byte WAL removed?). It seems more likely an SST actually could have contained data newer than the checksum error that was detected. In that case it'd be a valid failure.

deepakojha123:
@ajkr Unfortunately I don't have the original DB. Digging further, looks like 6.4.6 has a bug which was fixed in PR #5856. The condition is precisely what we are hitting: a point-in-time recovery with a corrupted wal file and a crash during recovery after flushing some CFs. Now the flushed CFs moved past the corrupted wal while remaining CFs still behind. Next Open will hit the corruption error. The fix is to update the Manifest atomically for all CFs after the flush.

deepakojha123:
I think we can drop this PR given that issue #7784 seems to be already fixed in PR #5856.

PR #7860, opened 2021-01-13
Fix Java hashCode implementation

laurentgo:
Summary:
Classes ColumnFamilyHandle and CapturingWriteBatchHandler.Event have
byte array fields as part of their identity, but they do not use the
arrays' content to compute the instance's hash, and instead rely on the
arrays' identity, causing instances to have different hashcodes
although they are equal.
The PR addresses it by using the arrays' content to compute the hash,
like the equals method does.


adamretter (approved):
LGTM

adamretter:
@akankshamahajan15 @jay-zhuang Could you merge please?

PR #7864, opened 2021-01-13
Make regression test load options from file for checkpoint

anand1976:
The regression_test.sh script checkpoints the DB directory before running db_bench on it. Specify the --try_load_options when creating the checkpoint in order to load options from the OPTIONS file.

Test:
manually run db_bench on the checkpoint dir


akankshamahajan15 (approved):
LGTM!

PR #7867, opened 2021-01-14
MergeHelper::FilterMerge() calling ElapsedNanosSafe() upon exit even …

matthewvon:
…when unused.  Causes many calls to clock_gettime, impacting performance.

Was looking for something else via Linux "perf" command when I spotted heavy usage of clock_gettime during a compaction.  Our product heavily uses the rocksdb::Options::merge_operator.  MergeHelper::FilterMerge() properly tests if timing is enabled/disabled upon entry, but not on exit.  This patch fixes the exit.

Note:  the entry test also verifies if "nullptr!=stats_".  This test is redundant to code within ShouldReportDetailedTime().  Therefore I omitted it in my change.

merge_test.cc updated with test that shows failure before merge_helper.cc change ... and fix after change.

### DIFF
--- a/db/merge_helper.cc
+++ b/db/merge_helper.cc
@@ -183,7 +183,6 @@ Status MergeHelper::MergeUntil(InternalIterator* iter,
 
     assert(IsValueType(ikey.type));
     if (ikey.type != kTypeMerge) {
-
       // hit a put/delete/single delete
       //   => merge the put value or a nullptr with operands_
       //   => store result in operands_.back() (and update keys_.back())
--- a/db/merge_helper.cc
+++ b/db/merge_helper.cc
@@ -414,7 +414,9 @@ CompactionFilter::Decision MergeHelper::FilterMerge(const Slice& user_key,
                                                        kValueTypeForSeek);
     }
   }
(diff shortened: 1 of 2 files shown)


db/merge_helper.cc
```diff
@@ -414,7 +413,9 @@ CompactionFilter::Decision MergeHelper::FilterMerge(const Slice& user_key,
                                                        kValueTypeForSeek);
     }
   }
-  total_filter_time_ += filter_timer_.ElapsedNanosSafe();
+  if (ShouldReportDetailedTime(env_, stats_)) {
```
mrambacher, line 416: I know it does not matter for execution purposes, but can you make this "ShouldReportTime" condition match the earlier one [ If (stats_ != nullptr && ... ) ]

Otherwise, someone is likely to look later and question why they are different.

mrambacher (approved):
LGTM.

db/merge_helper.cc
matthewvon: done.

ajkr (approved):
LGTM too. Not necessary to fix in this PR, but I don't see where `total_filter_time_` is reported or used.

PR #7870, opened 2021-01-15
Fix write-ahead log file size overflow

ReaLNeroM:
The WAL's file size is stored as an unsigned 64 bit integer. 

In db_info_dumper.cc, this integer gets converted to a string. Since 2^64 is approximately 10^19, we need 20 digits to represent the integer correctly. To store the decimal representation, we need 21 bytes (+1 due to the '\0' terminator at the end). The code previously used 16 bytes, which would overflow if the log is really big (>1 petabyte).

### DIFF
--- a/db/db_info_dumper.cc
+++ b/db/db_info_dumper.cc
@@ -64,9 +64,10 @@ void DumpDBFileSummary(const ImmutableDBOptions& options,
         break;
       case kWalFile:
         if (env->GetFileSize(dbname + "/" + file, &file_size).ok()) {
-          char str[21];
-          snprintf(str, sizeof(str), "%" PRIu64, file_size);
-          wal_info.append(file).append(" size: ").append(str).append(" ; ");
+          wal_info.append(file)
+              .append(" size: ")
+              .append(std::to_string(file_size))
+              .append(" ; ");
         } else {
           Error(options.info_log, "Error when reading LOG file: %s/%s\n",
                 dbname.c_str(), file.c_str());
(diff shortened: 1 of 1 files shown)


db/db_info_dumper.cc
```diff
@@ -64,7 +64,7 @@ void DumpDBFileSummary(const ImmutableDBOptions& options,
         break;
       case kWalFile:
         if (env->GetFileSize(dbname + "/" + file, &file_size).ok()) {
-          char str[16];
+          char str[21];
           snprintf(str, sizeof(str), "%" PRIu64, file_size);
           wal_info.append(file).append(" size: ").append(str).append(" ; ");
```
jay-zhuang, line 69: Thanks @ReaLNeroM for the contribution.
It would be better just to use `std::to_string()`:

```
wal_info.append(file).append(" size: ").append(std::to_string(file_size)).append(" ; ");
```

In reality, we should never have WAL size > 1PB (15 decimal number ~= 1 P) which means huge memory buffer size, and the most of filesystems don't even support 1P large single file. But it still worth to change it to to_string.

jay-zhuang (changes requested):
Thanks @ReaLNeroM for the contribution.
It would be better just to use `std::to_string()`:

```
wal_info.append(file).append(" size: ").append(std::to_string(file_size)).append(" ; ");
```

In reality, we should never have WAL size > 1P (15 decimal number ~= 1 Pb), and the most of filesystems don't support 1P single file.

db/db_info_dumper.cc
ReaLNeroM: Sounds good!

I agree with you that WAL >1PB is unlikely. I would argue that since the <1PB assumption is very hard to notice, it's still worth resolving this issue.

jay-zhuang (approved):
LGTM.

PR #7873, opened 2021-01-16
Make it able to ignore WAL related VersionEdits in older versions

cheng-chang:
Although the tags for `WalAddition`, `WalDeletion` are after `kTagSafeIgnoreMask`, to actually be able to skip these entries in older versions of RocksDB, we require that they are encoded with their encoded size as the prefix. This requirement is not met in the current codebase, so a downgraded DB may fail to open if these entries exist in the MANIFEST.

If a DB wants to downgrade, and its MANIFEST contains `WalAddition` or `WalDeletion`, it can set `track_and_verify_wals_in_manifest` to `false`, then restart twice, then downgrade. On the first restart, a new MANIFEST will be created with a `WalDeletion` indicating that all previously tracked WALs are removed from MANIFEST. On the second restart, since there is  no tracked WALs in MANIFEST now, a new MANIFEST will be created with neither `WalAddition` nor `WalDeletion`. Then the DB can downgrade.

Tags for `BlobFileAddition`, `BlobFileGarbage` also have the same problem, but this PR focuses on solving the problem for WAL edits.

Test Plan:
Added a `VersionEditTest::IgnorableTags` unit test to verify all entries with tags larger than `kTagSafeIgnoreMask` can actually be skipped and won't affect parsing of other entries.

### DIFF
--- a/db/version_edit.cc
+++ b/db/version_edit.cc
@@ -226,14 +226,14 @@ bool VersionEdit::EncodeTo(std::string* dst) const {
   }
 
   for (const auto& wal_addition : wal_additions_) {
-    PutVarint32(dst, kNewWalAddition);
+    PutVarint32(dst, kWalAddition2);
     std::string encoded;
     wal_addition.EncodeTo(&encoded);
     PutLengthPrefixedSlice(dst, encoded);
   }
 
   if (!wal_deletion_.IsEmpty()) {
-    PutVarint32(dst, kNewWalDeletion);
+    PutVarint32(dst, kWalDeletion2);
     std::string encoded;
     wal_deletion_.EncodeTo(&encoded);
     PutLengthPrefixedSlice(dst, encoded);
@@ -589,7 +589,7 @@ Status VersionEdit::DecodeFrom(const Slice& src) {
         break;
       }
 
-      case kNewWalAddition: {
+      case kWalAddition2: {
         Slice encoded;
         if (!GetLengthPrefixedSlice(&input, &encoded)) {
           msg = "WalAddition not prefixed by length";
@@ -617,7 +617,7 @@ Status VersionEdit::DecodeFrom(const Slice& src) {
         break;
       }
 
-      case kNewWalDeletion: {
+      case kWalDeletion2: {
         Slice encoded;
         if (!GetLengthPrefixedSlice(&input, &encoded)) {
           msg = "WalDeletion not prefixed by length";
--- a/db/version_edit.cc
+++ b/db/version_edit.cc
@@ -396,7 +396,6 @@ Status VersionEdit::DecodeFrom(const Slice& src) {
   while (msg == nullptr && GetVarint32(&input, &tag)) {
 #ifndef NDEBUG
     if (ignore_ignorable_tags && tag > kTagSafeIgnoreMask) {
-      printf("tag = %d\n", tag);
       tag = kTagSafeIgnoreMask;
     }
 #endif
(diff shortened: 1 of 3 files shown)


db/version_edit.cc
```diff
@@ -217,22 +217,30 @@ bool VersionEdit::EncodeTo(std::string* dst) const {
 
   for (const auto& blob_file_addition : blob_file_additions_) {
     PutVarint32(dst, kBlobFileAddition);
-    blob_file_addition.EncodeTo(dst);
+    std::string encoded;
```
cheng-chang, line 220: @ltamasi This breaks backward compatibility. But without doing so, users of BlobDB cannot downgrade. What do you think?

cheng-chang: I'll remove changes related to BlobDB from this PR to make it focus on WAL edits.

anand1976:
As @ltamasi points out for the BlobDB entries, wouldn't this change cause problems when upgrading? Maybe we could introduce new tags with the correct format, and retain the existing ones to be able to decode?

db/version_edit.cc
ltamasi: > @ltamasi This breaks backward compatibility. But without doing so, users of BlobDB cannot downgrade. What do you think?

Nice catch, @cheng-chang and thanks for letting me know! In hindsight, the blob file related tags shouldn't have been put in the "ignorable" range since if you have blob files and downgrade to a version that cannot read these, you won't be able to read your data. Fortunately, since these are related to the integrated BlobDB project (which is WIP), we can still move these to the "non-ignorable" range. I'll take care of that.

cheng-chang:
> As @ltamasi points out for the BlobDB entries, wouldn't this change cause problems when upgrading? Maybe we could introduce new tags with the correct format, and retain the existing ones to be able to decode?

Then when we do the downgrade to a version older than 6.15, it still fails to be decoded unless we restart DB twice to remove the entries.

ajkr (approved):
Should we be pessimistic and call them "kWal{Addition,Deletion}2"? It also helps since "new" is ambiguous -- in the case of `kNewFile`, "new" means the file is being added, I think.

ajkr:
Also can we have a "bug fixes" release note that mentions the downgrade instructions for users who ran 6.15.1 with WAL tracking enabled?

anand1976 (approved):

cheng-chang:
> Also can we have a "bug fixes" release note that mentions the downgrade instructions for users who ran 6.15.1 with WAL tracking enabled?

The wal tracking is only released in 6.16. In 6.15, there is a comment on the option indicating that it's not ready for use yet. MyRocks accidentally enabled it in 6.15.

ajkr:
> MyRocks accidentally enabled it in 6.15.

Got it, I missed this detail earlier, ignore my suggestion then here and questions in group chat

ajkr:
> I think we'll mention this in 6.16.1's release note.

It makes sense. I hope the release note will appear on master branch as well, so that it's clear upgrading from 6.16.0 to a newer major/minor version also gets the fix, not just upgrading from 6.16.0 to 6.16.1 or a later patch version.

ltamasi:
> > I think we'll mention this in 6.16.1's release note.
> 
> It makes sense. I hope the release note will appear on master branch as well, so that it's clear upgrading from 6.16.0 to a newer major/minor version also gets the fix, not just upgrading from 6.16.0 to 6.16.1 or a later patch version.

@cheng-chang Seems to me you've also cherry-picked this into 6.17, so I think we should also call it out in the 6.17.0 section of the changelog (on master and 6.17.fb as well).

PR #7875, opened 2021-01-16
Fix compilation against musl lib C

adamretter:
See https://github.com/percona/PerconaFT/pull/450


jay-zhuang:
Is there a way for me to verify that?

adamretter:
> Is there a way for me to verify that?

Good point @jay-zhuang! I will try and get a CircleCI setup which covers building with musl.

jay-zhuang (approved):
👍

adamretter:
@jay-zhuang I have a CircleCI job now that tests building with musl.
I think perhaps it is better to keep that in a separate PR - https://github.com/facebook/rocksdb/pull/7876

But you can see the output from trying to build (before this patch) with musl - https://app.circleci.com/pipelines/github/facebook/rocksdb/5557/workflows/90e82fca-ff41-4085-b3d2-5e66fc644581/jobs/85799

PR #7876, opened 2021-01-17
CircleCI jobs for RocksJava binaries using the normal docker build env

adamretter:


pdillinger:
Can you argue for (justify) testing different distributions? (Burning CircleCI credits)

jay-zhuang:
> Can you argue for (justify) testing different distributions? (Burning CircleCI credits)

maybe a candidate for nightly build: https://github.com/facebook/rocksdb/blob/d561432de89e684129d6db939f9a77d7ed0b0ba0/.circleci/config.yml#L816-L826

adamretter:
@pdillinger Sure of course! I often find when I test each tag that there are regressions whereby RocksDB no longer compiles on the platforms we release for, I then have to fix those for each release and send PRs. The idea of this is to test on the actual platforms we release for, and prevent regressions from occurring at their source.

pdillinger:
@adamretter I would expect pass/fail to be almost entirely dependent on platform + compiler + compiler version, not distro, glibc version, etc. What compiler+version do these use? Sell it to me that way?

@jay-zhuang Not a bad idea. I don't think I've ever seen non-shm failing uniquely, so I would nominate those to go into nightly

adamretter:
> What compiler+version do these use? Sell it to me that way?

@pdillinger these are the platforms that this would add:

* CentOS 6
    * x86_64 / gcc (GCC) 7.3.1 20180303
    * i386 / gcc (GCC) 4.8.2 20140120
* Alpine 3.8.5
    * x86_64 / gcc 6.4.0
    * i386 / gcc 6.4.0

pdillinger:
OK, if you get this into a shippable state, I'm ok with it. We can consider moving lower-value builds to "nightly" after this lands.

adamretter:
@pdillinger I believe this one is now ready (albeit it needs https://github.com/facebook/rocksdb/pull/9500 merged first and then CI re-run)

pdillinger:
Not passing

> Java version is javac 1.7.0_281, minimum required version is 8

PR #7883, opened 2021-01-20
Fix CMakeLists.txt for Apple Silicon

SeekingMeaning:
`CMAKE_SYSTEM_PROCESSOR` defaults to `arm64` on Apple Silicon starting with CMake 3.19.2; see https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#macos-platforms


adamretter (approved):
Looks ok to me.

PR #7884, opened 2021-01-20
Add full_history_ts_low option to compaction

jay-zhuang:
The full_history_ts_low is used for user-defined timestamp GC
compaction, which is introduced in #7740, #7657 and #7655.

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -806,6 +806,25 @@ Status DBImpl::CompactRange(const CompactRangeOptions& options,
                               end_with_ts);
 }
 
+Status DBImpl::IncreaseFullHistoryTsLow(ColumnFamilyData* cfd,
+                                        std::string ts_low) {
+  VersionEdit edit;
+  edit.SetColumnFamily(cfd->GetID());
+  edit.SetFullHistoryTsLow(ts_low);
+
+  InstrumentedMutexLock l(&mutex_);
+  std::string current_ts_low = cfd->GetFullHistoryTsLow();
+  const Comparator* ucmp = cfd->user_comparator();
+  if (!current_ts_low.empty() &&
+      ucmp->CompareTimestamp(ts_low, current_ts_low) < 0) {
+    return Status::InvalidArgument(
+        "Cannot decrease full_history_timestamp_low");
+  }
+
+  return versions_->LogAndApply(cfd, *cfd->GetLatestMutableCFOptions(), &edit,
+                                &mutex_);
+}
+
 Status DBImpl::CompactRangeInternal(const CompactRangeOptions& options,
                                     ColumnFamilyHandle* column_family,
                                     const Slice* begin, const Slice* end) {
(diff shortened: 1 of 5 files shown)


db/db_with_timestamp_basic_test.cc
```diff
@@ -223,6 +223,111 @@ TEST_F(DBBasicTestWithTimestamp, CompactRangeWithSpecifiedRange) {
...
+  options.comparator = &test_cmp;
+  DestroyAndReopen(options);
+
+  const std::string kKey = "test kKey";
+
+  // Test set ts_low first and flush()
+  int current_ts_low = 5;
+  std::string ts_low_str = Timestamp(current_ts_low, 0);
+  Slice ts_low = ts_low_str;
+  CompactRangeOptions comp_opts = CompactRangeOptions();
```
ltamasi, line 241: Minor: you can simply do

```
CompactRangeOptions comp_opts;
```

ltamasi (approved):
LGTM with a couple of minor comments, thanks!

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -806,6 +806,25 @@ Status DBImpl::CompactRange(const CompactRangeOptions& options,
...
+                                        std::string ts_low) {
+  VersionEdit edit;
+  edit.SetColumnFamily(cfd->GetID());
+  edit.SetFullHistoryTsLow(ts_low);
+
+  InstrumentedMutexLock l(&mutex_);
+  std::string current_ts_low = cfd->GetFullHistoryTsLow();
+  const Comparator* ucmp = cfd->user_comparator();
+  if (!current_ts_low.empty() &&
+      ucmp->CompareTimestamp(ts_low, current_ts_low) < 0) {
```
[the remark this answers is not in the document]
jay-zhuang: Sounds good. I was thinking if the compaction failed and the user want to retry with the same parameter. On the other hand, in that case, the user could retry without timestamp, as the CompactRange() call returned succeed, so they should just retry the compaction but not the setting timestamp.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -817,6 +836,23 @@ Status DBImpl::CompactRangeInternal(const CompactRangeOptions& options,
   }
 
   bool flush_needed = true;
+
+  // Update full_history_ts_low if it's set
+  if (options.full_history_ts_low != nullptr) {
+    std::string ts_low = options.full_history_ts_low->ToString();
+    if (!ts_low.empty()) {
```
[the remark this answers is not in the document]
jay-zhuang: Updated.

db/db_with_timestamp_basic_test.cc
jay-zhuang: Updated.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -806,6 +806,25 @@ Status DBImpl::CompactRange(const CompactRangeOptions& options,
...
+  InstrumentedMutexLock l(&mutex_);
+  std::string current_ts_low = cfd->GetFullHistoryTsLow();
+  const Comparator* ucmp = cfd->user_comparator();
+  if (!current_ts_low.empty() &&
+      ucmp->CompareTimestamp(ts_low, current_ts_low) < 0) {
+    return Status::InvalidArgument(
+        "Cannot decrease full_history_timestamp_low");
+  }
+
+  return versions_->LogAndApply(cfd, *cfd->GetLatestMutableCFOptions(), &edit,
```
riversand963, line 824: There is an edge case here.
Since `LogAndApply()` can release and acquire the dbmutex, multiple concurrent calls to `IncreaseFullHistoryTsLow()` may interleave. It is possible for one thread to attempt to increase the threshold to 5, while another thread increases the threshold to 10. Either thread can precede the other when writing to the MANIFEST. It's possible that `LogAndApply()` returns OK, but the actual ts_low is 10. In this case, it may be better to inform the caller that the actual ts_low is higher so that this caller will not make the wrong assumption such that data with ts between [5, 10) will be preserved.

riversand963 (commented):
Thanks @jay-zhuang for the PR, and sorry for the late review.

PR #7885, opened 2021-01-20
In IOTracing, add filename with each operation in trace file.

akankshamahajan15:
Summary: 1. In IOTracing, add filename with each IOTrace record. Filename is stored in file object (Tracing Wrappers).
         2. Change the logic of figuring out which additional information (file_size,
            length, offset etc) needs to be store with each operation
            which is different for different operations.
            When new information will be added in future (depends on operation),
            this change would make the future additions simple.
  
Logic: In IOTraceRecord, io_op_data is added and its
         bitwise positions represent which additional information need
         to added in the record from enum IOTraceOp. Values in IOTraceOp represent bitwise positions.
         So if length and offset needs to be stored (IOTraceOp::kIOLen
         is 1 and IOTraceOp::kIOOffset is 2), position 1 and 2 (from rightmost bit) will be set
         and io_op_data will contain 110.

Test Plan: Updated io_tracer_test and verified the trace file manually.

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/env/file_system_tracer.h
+++ b/env/file_system_tracer.h
@@ -168,7 +168,9 @@ class FSSequentialFilePtr {
                       const std::string& file_name)
       : fs_(std::move(fs)),
         io_tracer_(io_tracer),
-        fs_tracer_(fs_.get(), io_tracer_, file_name) {}
+        fs_tracer_(fs_.get(), io_tracer_,
+                   file_name.substr(file_name.find_last_of("/\\") +
+                                    1) /* pass file name */) {}
 
   FSSequentialFile* operator->() const {
     if (io_tracer_ && io_tracer_->is_tracing_enabled()) {
@@ -225,6 +227,7 @@ class FSRandomAccessFileTracingWrapper : public FSRandomAccessFileWrapper {
  private:
   std::shared_ptr<IOTracer> io_tracer_;
   Env* env_;
+  // Stores file name instead of full path.
   std::string file_name_;
 };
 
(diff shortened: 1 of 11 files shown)


env/file_system_tracer.h
```diff
@@ -150,6 +152,7 @@ class FSSequentialFileTracingWrapper : public FSSequentialFileWrapper {
  private:
   std::shared_ptr<IOTracer> io_tracer_;
   Env* env_;
+  std::string file_name_;
```
anand1976, line 155: Instead of saving the entire file name here and in the other wrapper classes, which could be pretty long as it would contain the entire path, we can consider parsing the file name and just storing the number and type. Then when writing the trace, reconstruct the file name. Its not necessary to write the full path.

anand1976 (commented):
Thanks for the PR!

tools/io_tracer_parser_tool.cc
```diff
@@ -42,28 +42,29 @@ void IOTraceRecordParser::PrintHumanReadableIOTraceRecord(
...
-      FALLTHROUGH_INTENDED;
-    case TraceType::kIOFileName: {
-      if (!record.file_name.empty()) {
-        ss << ", File Name: " << record.file_name.c_str();
-      }
-      break;
+     << ", IO Status: " << record.io_status.c_str()
+     << ", File Name: " << record.file_name.c_str();
+
+  /* Read remaining options based on io_op_data set by file operation */
```
zhichao-cao, line 48: Better to add a little bit more comments to explain the logic here, and also how to add other operations in the future.

anand1976 (approved):
LGTM

Issue #7881, opened 2021-01-19
Undefined behaviour in multithreaded code


aleksuss:
I decided to bump version of rocksdb up to 6.13.3 in [rust-rocksdb](https://github.com/rust-rocksdb/rust-rocksdb) but stuck with undefined behaviour while rust tests executed. I figured out that the reason of such behaviour is multithreaded execution of rust tests. Every test creates an instance of the database. Current release of `rust-rocksdb` crate uses `rocksdb` version 6.11.4. With this version such behaviour doesn't reproduce. So. I wrote c++ code which reproduces this behaviour. See below. 

### Expected behaviour
Creating an instance of the database per thread without errors.

### Actual behavior
```
libc++abi.dylib: terminating
Process 86130 stopped
* thread #2, stop reason = signal SIGABRT
    frame #0: 0x00007fff20345462 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff20345462 <+10>: jae    0x7fff2034546c            ; <+20>
    0x7fff20345464 <+12>: movq   %rax, %rdi
    0x7fff20345467 <+15>: jmp    0x7fff2033f6a1            ; cerror_nocancel
    0x7fff2034546c <+20>: retq   
Target 0: (cpp-playground) stopped.
(lldb) bt
* thread #2, stop reason = signal SIGABRT
  * frame #0: 0x00007fff20345462 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff20373610 libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007fff202c6720 libsystem_c.dylib`abort + 120
    frame #3: 0x00007fff20338418 libc++abi.dylib`abort_message + 231
    frame #4: 0x00007fff203299a3 libc++abi.dylib`demangling_terminate_handler() + 48
    frame #5: 0x00007fff20337847 libc++abi.dylib`std::__terminate(void (*)()) + 8
    frame #6: 0x00007fff203377f8 libc++abi.dylib`std::terminate() + 56
    frame #7: 0x00007fff203117ed libc++.1.dylib`std::__1::thread::~thread() + 17
    frame #8: 0x000000010026d741 librocksdb.6.dylib`std::__1::unique_ptr<std::__1::thread, std::__1::default_delete<std::__1::thread> >::reset(std::__1::thread*) + 25
[...]

PR #7888, opened 2021-01-21
workaround race conditions during `PeriodicWorkScheduler` registration

ajkr:
This provides a workaround for two race conditions that will be fixed in
a more sophisticated way later. This PR:

(1) Makes the client serialize calls to `Timer::Start()` and `Timer::Shutdown()` (see #7711). The long-term fix will be to make those functions thread-safe.
(2) Makes `PeriodicWorkScheduler` atomically add/cancel work together with starting/shutting down its `Timer`. The long-term fix will be for `Timer` API to offer more specialized APIs so the client will not need to synchronize.

Test Plan: ran the repro provided in #7881


jay-zhuang (approved):
LGTM.
Thanks for the fix.

PR #7891, opened 2021-01-22
Cleanup multiple DBs after running db_bench in multi-DB mode

anand1976:
Summary:
Currently, db_bench cleanup only deletes the main DB, if there's one.
Multiple DBs that are opened when --num_multi_db is specified are not
deleted, which can lead to crashes due to running compaction threads on
process exit.

Test plan:
Run regression test


ajkr (approved):
LGTM

PR #7896, opened 2021-01-26
Fix flaky `error_handler_fs_test.MultiDBCompactionError`

jay-zhuang:
The error recovery thread may out-live DBImpl object, which causing
access released DBImpl.mutex. Close SstFileManager before closing DB.

Test Plan: the issue can be reproduced by adding sleep in recovery code.
Pass the tests with sleep.


zhichao-cao (approved):
LGTM, thanks for the fix!

PR #7897, opened 2021-01-26
Fix deadlock in `fs_test.WALWriteRetryableErrorAutoRecover1`

jay-zhuang:
The recovery thread could hold the db.mutex, which is needed from sync
write in main thread.
Make sure the write is done before recovery thread starts.

Test Plan: `gtest-parallel ./error_handler_fs_test --gtest_filter=DBErrorHandlingFSTest.WALWriteRetryableErrorAutoRecover1 -r 10000 --workers=200`

### DIFF
--- a/db/error_handler_fs_test.cc
+++ b/db/error_handler_fs_test.cc
@@ -2248,7 +2248,8 @@ TEST_F(DBErrorHandlingFSTest, WALWriteRetryableErrorAutoRecover1) {
       ASSERT_OK(batch.Put(Key(i), rnd.RandomString(1024)));
     }
     ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->LoadDependency(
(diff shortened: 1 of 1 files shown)


zhichao-cao (commented):
So the case is, the recover thread hold the db_mutex, and is at RecoverFromRetryableBGIOError:BeforeResume0 wait for WALWriteError1:0. However, the main thread try to grab db_mutex for sync, so waiting at that location, which is ahead of WALWriteError1:0. Do I understand correctly?

jay-zhuang:
> So the case is, the recover thread hold the db_mutex, and is at RecoverFromRetryableBGIOError:BeforeResume0 wait for WALWriteError1:0. However, the main thread try to grab db_mutex for sync, so waiting at that location, which is ahead of WALWriteError1:0. Do I understand correctly?

Yes exactly.

zhichao-cao (approved):
LGTM, thanks for the fix!

PR #7898, opened 2021-01-26
Stall writes in WriteBufferManager when memory_usage exceeds buffer_size

akankshamahajan15:
Summary: When WriteBufferManager is shared across DBs and column families
to maintain memory usage under a limit, OOMs have been observed when flush cannot
finish but writes continuously insert to memtables.
In order to avoid OOMs, when memory usage goes beyond buffer_limit_ and DBs tries to write,
this change will stall incoming writers until flush is completed and memory_usage
drops.

Design: Stall condition: When total memory usage exceeds WriteBufferManager::buffer_size_
(memory_usage() >= buffer_size_) WriterBufferManager::ShouldStall() returns true.

DBImpl first block incoming/future writers by calling write_thread_.BeginWriteStall()
(which adds dummy stall object to the writer's queue).
Then DB is blocked on a state State::Blocked (current write doesn't go
through). WBStallInterface object maintained by every DB instance is added to the queue of
WriteBufferManager.

If multiple DBs tries to write during this stall, they will also be
blocked when check WriteBufferManager::ShouldStall() returns true.

End Stall condition: When flush is finished and memory usage goes down, stall will end only if memory
waiting to be flushed is less than buffer_size/2. This lower limit will give time for flush
to complete and avoid continous stalling if memory usage remains close to buffer_size.

WriterBufferManager::EndWriteStall() is called,
which removes all instances from its queue and signal them to continue.
Their state is changed to State::Running and they are unblocked. DBImpl
then signal all incoming writers of that DB to continue by calling
write_thread_.EndWriteStall() (which removes dummy stall object from the
queue).

DB instance creates WBMStallInterface which is an interface to block and
signal DBs during stall.
When DB needs to be blocked or signalled by WriteBufferManager,
state_for_wbm_ state is changed accordingly (RUNNING or BLOCKED).

Test Plan: Added a new test db/db_write_buffer_manager_test.cc

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -234,8 +234,7 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
       error_handler_(this, immutable_db_options_, &mutex_),
       atomic_flush_install_cv_(&mutex_),
       blob_callback_(immutable_db_options_.sst_file_manager.get(), &mutex_,
-                     &error_handler_),
-      wbm_stall_(nullptr) {
+                     &error_handler_) {
   // !batch_per_trx_ implies seq_per_batch_ because it is only unset for
   // WriteUnprepared, which should use seq_per_batch_.
   assert(batch_per_txn_ || seq_per_batch_);
@@ -273,7 +272,7 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
   preserve_deletes_seqnum_.store(0);
 
   if (write_buffer_manager_) {
-    wbm_stall_ = new WBMStallInterface();
+    wbm_stall_.reset(new WBMStallInterface());
   }
 }
 
@@ -666,9 +665,7 @@ Status DBImpl::CloseHelper() {
   }
 
   if (write_buffer_manager_ && wbm_stall_) {
-    write_buffer_manager_->RemoveDBFromQueue(wbm_stall_);
-    delete wbm_stall_;
-    wbm_stall_ = nullptr;
+    write_buffer_manager_->RemoveDBFromQueue(wbm_stall_.get());
   }
 
   if (ret.IsAborted()) {
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -234,7 +234,8 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
       error_handler_(this, immutable_db_options_, &mutex_),
       atomic_flush_install_cv_(&mutex_),
       blob_callback_(immutable_db_options_.sst_file_manager.get(), &mutex_,
-                     &error_handler_) {
+                     &error_handler_),
+      wbm_stall_(new WBMStallInterface()) {
   // !batch_per_trx_ implies seq_per_batch_ because it is only unset for
   // WriteUnprepared, which should use seq_per_batch_.
   assert(batch_per_txn_ || seq_per_batch_);
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -236,6 +236,7 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
       blob_callback_(immutable_db_options_.sst_file_manager.get(), &mutex_,
                      &error_handler_),
       wbm_stall_(new WBMStallInterface()) {
+      wbm_stall_(nullptr) {
   // !batch_per_trx_ implies seq_per_batch_ because it is only unset for
   // WriteUnprepared, which should use seq_per_batch_.
   assert(batch_per_txn_ || seq_per_batch_);
@@ -271,6 +272,10 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
   // we won't drop any deletion markers until SetPreserveDeletesSequenceNumber()
   // is called by client and this seqnum is advanced.
   preserve_deletes_seqnum_.store(0);
+
+  if (write_buffer_manager_) {
+    wbm_stall_ = new WBMStallInterface();
+  }
 }
 
 Status DBImpl::Resume() {
@@ -659,6 +664,11 @@ Status DBImpl::CloseHelper() {
     if (!s.ok() && !s.IsNotSupported() && ret.ok()) {
       ret = s;
     }
+  }
+
+  if (wbm_stall_) {
+    delete wbm_stall_;
+    wbm_stall_ = nullptr;
   }
 
   if (ret.IsAborted()) {
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -235,7 +235,6 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
       atomic_flush_install_cv_(&mutex_),
       blob_callback_(immutable_db_options_.sst_file_manager.get(), &mutex_,
                      &error_handler_),
-      wbm_stall_(new WBMStallInterface()) {
       wbm_stall_(nullptr) {
   // !batch_per_trx_ implies seq_per_batch_ because it is only unset for
   // WriteUnprepared, which should use seq_per_batch_.
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -665,7 +665,8 @@ Status DBImpl::CloseHelper() {
     }
   }
 
-  if (wbm_stall_) {
+  if (write_buffer_manager_ && wbm_stall_) {
+    write_buffer_manager_->RemoveDBFromQueue(wbm_stall_);
     delete wbm_stall_;
     wbm_stall_ = nullptr;
   }
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -1064,12 +1064,16 @@ class DBImpl : public DB {
       state_ = State::RUNNING;
     }
 
+    void SetState(State state) {
+      MutexLock lock(&state_mutex_);
+      state_ = state;
+    }
+
     // Change the state_ to State::BLOCKED and wait until its state is
     // changed by WriteBufferManager. When stall is cleared, Signal() is
     // called to change the state and unblock the DB.
     void Block() override {
       MutexLock lock(&state_mutex_);
-      state_ = State::BLOCKED;
       while (state_ == State::BLOCKED) {
         TEST_SYNC_POINT("WBMStallInterface::BlockDB");
         state_cv_.Wait();
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -113,6 +113,54 @@ class Directories {
   std::unique_ptr<FSDirectory> wal_dir_;
 };
 
+// Interface to block and signal the DB in case of stalling writes by
+// WriteBufferManager. Each DBImpl object contains ptr to WBMStallInterface.
+// When DB needs to be blocked or signalled by WriteBufferManager,
+// state_for_wbm_ state is changed accordingly.
+class WBMStallInterface {
+ public:
+  enum State {
+    BLOCKED = 0,
+    RUNNING,
+  };
+
+  WBMStallInterface() : state_cv_(&state_mutex_) {
+    {
+      MutexLock lock(&state_mutex_);
+      state_for_wbm_.store(State::RUNNING, std::memory_order_relaxed);
+    }
+  }
+
+  // Change the state_for_wbm_ to State::BLOCKED and wait until its state is
+  // changed by WriteBufferManager. When stall is cleared, SignalDB() is called
+  // to change the state and unblock the DB.
+  void BlockDB() {
+    MutexLock lock(&state_mutex_);
+    state_for_wbm_.store(State::BLOCKED, std::memory_order_relaxed);
+    while (state_for_wbm_.load(std::memory_order_relaxed) == State::BLOCKED) {
+      TEST_SYNC_POINT("DBImpl::BlockDB");
+      state_cv_.Wait();
+    }
+  }
+
+  // Called from WriteBufferManager. This function changes the state_for_wbm_ to
+  // State::RUNNING indicating the stall is cleared and DB can proceed.
+  void SignalDB() {
+    MutexLock lock(&state_mutex_);
+    state_for_wbm_.store(State::RUNNING, std::memory_order_relaxed);
+    state_cv_.Signal();
+  }
+
+ private:
+  // Conditional variable and mutex to block and
+  // signal the DB during stalling process.
+  port::Mutex state_mutex_;
+  port::CondVar state_cv_;
+  // state represting whether DB is running or blocked because of stall by
+  // WriteBufferManager.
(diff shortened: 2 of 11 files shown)


akankshamahajan15:
TODO: If memory stays close to the limit, will FreeMem() cause all the dbs to be added and removed from the queue frequently. For example, one write can cause all writing dbs to enqueue, while a cache eviction can cause all of them to be removed. Then another write comes and the memory usage goes above the limit and all dbs are queued again. 
Need to find a way to avoid it.

memtable/write_buffer_manager.cc
```diff
@@ -78,6 +79,34 @@ WriteBufferManager::~WriteBufferManager() {
 #endif  // ROCKSDB_LITE
 }
 
+// Should only be called from write thread
+bool WriteBufferManager::ShouldFlush() const {
```
anand1976, line 83: Any particular reason why this is here instead of the header file? ```ShouldFlush()``` is called by every write, so keeping it in the header file may make it more likely to be inlined. Maybe compiler would optimize it anyway, but just to be on the safe side.

db/db_impl/db_impl.h
```diff
@@ -113,6 +113,54 @@ class Directories {
   std::unique_ptr<FSDirectory> wal_dir_;
 };
 
+// Interface to block and signal the DB in case of stalling writes by
+// WriteBufferManager. Each DBImpl object contains ptr to WBMStallInterface.
+// When DB needs to be blocked or signalled by WriteBufferManager,
+// state_for_wbm_ state is changed accordingly.
+class WBMStallInterface {
```
anand1976, line 120: Can this be defined as a trait in ```write_buffer_manager.h``` and ```DBImpl``` implements it? Or define an abstract class in ```write_buffer_manager.h``` and derive ```WBMStallInterface``` from it? That would avoid ```static_cast```ing it.

db/db_impl/db_impl.h
```diff
@@ -113,6 +113,54 @@ class Directories {
...
+  }
+
+ private:
+  // Conditional variable and mutex to block and
+  // signal the DB during stalling process.
+  port::Mutex state_mutex_;
+  port::CondVar state_cv_;
+  // state represting whether DB is running or blocked because of stall by
+  // WriteBufferManager.
+  std::atomic<uint8_t> state_for_wbm_;
```
anand1976, line 161: Why is ```state_for_wbm_``` an atomic variable? Access is anyway protected by ```state_mutex_``` right?

db/db_impl/db_impl_write.cc
```diff
@@ -1528,6 +1542,27 @@ Status DBImpl::DelayWrite(uint64_t num_bytes,
   return s;
 }
 
+// REQUIRES: mutex_ is held
```
anand1976, line 1545: I suggest asserting that ```mutex_``` is locked at the start of the function

memtable/write_buffer_manager.cc
```diff
@@ -141,8 +188,43 @@ void WriteBufferManager::FreeMemWithCache(size_t mem) {
...
+  while (!ShouldStall() && !queue_.empty()) {
+    std::unique_lock<std::mutex> lock(mu_);
+    DB* db = queue_.front();
+    queue_.pop_front();
+    (static_cast<DBImpl*>(db))->GetWBMStallInterface()->SignalDB();
+  }
+}
+
+bool WriteBufferManager::ShouldStall() {
+  if (enabled()) {
```
anand1976, line 222: I don't think we want to impact existing behavior. Its better if the stall behavior is enabled through an option in the ```WriteBufferManager``` constructor. And if its enabled, I think the stall condition should only be ```memory_usage() >= buffer_size_```, since the idea is to have a hard limit. Also, once the stall is triggered, it might be a good idea to have a low watermark for removing the stall in order to give some time for flushes to complete.

akankshamahajan15:
> TODO: If memory stays close to the limit, will FreeMem() cause all the dbs to be added and removed from the queue frequently. For example, one write can cause all writing dbs to enqueue, while a cache eviction can cause all of them to be removed. Then another write comes and the memory usage goes above the limit and all dbs are queued again.
> Need to find a way to avoid it.
It has been addressed by adding a limit and wait until memory being flushed is less than buffer_size/2.

ajkr:
> TODO: If memory stays close to the limit, will FreeMem() cause all the dbs to be added and removed from the queue frequently. For example, one write can cause all writing dbs to enqueue, while a cache eviction can cause all of them to be removed. Then another write comes and the memory usage goes above the limit and all dbs are queued again.
> Need to find a way to avoid it.

> Also, once the stall is triggered, it might be a good idea to have a low watermark for removing the stall in order to give some time for flushes to complete.

> End Stall condition: When flush is finished and memory usage goes down, stall will end only if memory
waiting to be flushed is less than buffer_size/2.

Sorry for the late comment - was just passing by and had a question about these. Is it really better for a DB to be in stall mode for a longer time than it is to thrash in and out of stall mode? I tend to think a stalled DB is basically unavailable so we'd only want to stall when absolutely necessary to avoid crashing. If the problem is the thrashing is so crazy it slows the DB from recovering, maybe there can be a gentler low watermark, just enough to prevent a single block eviction from changing the stall mode?

akankshamahajan15:
> Sorry for the late comment - was just passing by and had a question about these. Is it really better for a DB to be in stall mode for a longer time than it is to thrash in and out of stall mode? I tend to think a stalled DB is basically unavailable so we'd only want to stall when absolutely necessary to avoid crashing. If the problem is the thrashing is so crazy it slows the DB from recovering, maybe there can be a gentler low watermark, just low enough to prevent irrelevant things like block evictions from changing the stall mode?

@ajkr, I see. If i set lower watermark to be 90% instead of 50% of buffer_size, does that seem ok? Low watermark is added because if memory stays close to buffer_size that it will not add/remove DBs from stall continously and give time for memory to go down so that stall doesn't happen frequently/continously.

akankshamahajan15:
> > Sorry for the late comment - was just passing by and had a question about these. Is it really better for a DB to be in stall mode for a longer time than it is to thrash in and out of stall mode? I tend to think a stalled DB is basically unavailable so we'd only want to stall when absolutely necessary to avoid crashing. If the problem is the thrashing is so crazy it slows the DB from recovering, maybe there can be a gentler low watermark, just low enough to prevent irrelevant things like block evictions from changing the stall mode?
> 
> @ajkr, I see. If i set lower watermark to be 90% instead of 50% of buffer_size, does that seem ok? Low watermark is added because if memory stays close to buffer_size that it will not add/remove DBs from stall continously and give time for memory to go down so that stall doesn't happen frequently/continously.

50% might be too aggressive.

ajkr:
> @ajkr, I see. If i set lower watermark to be 90% instead of 50% of buffer_size, does that seem ok?

Probably, though I am wondering about a couple things related to the approach:

- If the end of write stall mode can only be triggered when flush finishes (and not when a block is evicted from cache), would that also prevent thrashing?
- Is it guaranteed that size of memtables awaiting flush is any particular portion of buffer_size at the point stalling starts? I recall during a write it picks only the "best" CF in the current DB to flush, so we might not flush all memtables before reaching the limit. If that's the case, then no matter the low watermark we choose (50% of buffer_size, 90%, or something else), it's possible that watermark is already satisfied as soon as we enter stall mode, before any flush finishes.

akankshamahajan15:
> > @ajkr, I see. If i set lower watermark to be 90% instead of 50% of buffer_size, does that seem ok?
> 
> Probably, though I am wondering about a couple things related to the approach:
> 
> * If the end of write stall mode can only be triggered when flush finishes (and not when a block is evicted from cache), would that also prevent thrashing?
> * Is it guaranteed that size of memtables awaiting flush is any particular portion of buffer_size at the point stalling starts? I recall during a write it picks only the "best" CF in the current DB to flush, so we might not flush all memtables before reaching the limit. If that's the case, then no matter the low watermark we choose (50% of buffer_size, 90%, or something else), it's possible that watermark is already satisfied as soon as we enter stall mode, before any flush finishes.

For 2nd point: Condition I wanted to add is if flush is going on and we stall, we wait for some percentage of flush to complete before releasing the stall.  But i see your point and we cannot gurantee that with buffer_size_.  I am not sure if we can add lower watermark in that case considering in terms memory in flush as it will be keep changing as DB who will write might/might not flush and stall and if yes, what it should be.

akankshamahajan15:
> For 2nd point: Condition I wanted to add is if flush is going on and we stall, we wait for some percentage of flush to complete before releasing the stall. But i see your point and we cannot gurantee that with buffer_size_. I am not sure if we can add lower watermark in that case considering in terms memory in flush as it will be keep changing as DB who will write might/might not flush and stall and if yes, what it should be.

Removed lower watermark as of now and updated code to remove stall when buffer_size is increased dynamically.

memtable/write_buffer_manager.cc
```diff
@@ -145,4 +178,38 @@ void WriteBufferManager::FreeMemWithCache(size_t mem) {
...
+    queue_.push_back(wbm_stall);
+  }
+}
+
+// Called when memory is freed in FreeMem.
+void WriteBufferManager::EndWriteStall() {
+  if (enabled()) {
+    {
+      std::unique_lock<std::mutex> lock(mu_);
+      if (queue_.empty()) {
```
anand1976, line 194: Shouldn't ```IsStallThresholdExceeded``` be checked and ```stall_active_``` cleared anyway, even if the queue is empty? The calls to ```ShouldStall``` and ```BeginWriteStall``` are not atomic.

memtable/write_buffer_manager.cc
```diff
@@ -145,4 +178,38 @@ void WriteBufferManager::FreeMemWithCache(size_t mem) {
   (void)mem;
 #endif  // ROCKSDB_LITE
 }
+
+void WriteBufferManager::BeginWriteStall(StallInterface* wbm_stall) {
+  if (wbm_stall && ShouldStall()) {
```
anand1976, line 183: Maybe just assert ```wbm_stall != nullptr```, since passing null is invalid.

anand1976, line 183: Would it be better to call ```ShouldStall`` after enqueuing ```wbm_stall```, and then calling ```EndWriteStall``` if necessary? That would handle race conditions where memory may get freed before the thread has a chance to enqueue itself, resulting in an unnecessary stall.

db/db_impl/db_impl.h
```diff
@@ -1046,6 +1046,52 @@ class DBImpl : public DB {
...
+      MutexLock lock(&state_mutex_);
+      state_ = State::RUNNING;
+    }
+
+    // Change the state_ to State::BLOCKED and wait until its state is
+    // changed by WriteBufferManager. When stall is cleared, Signal() is
+    // called to change the state and unblock the DB.
+    void Block() override {
+      MutexLock lock(&state_mutex_);
+      state_ = State::BLOCKED;
```
anand1976, line 1070: I'm kind of concerned about missed signals. I think a better sequence would be as follows -
```
state_ = State::BLOCKED;
wbm->BeginWriteStall(this);
MutexLock lock(&state_mutex_);
while (state_ == State::BLOCKED) {
  state_cv_.Wait();
}
```

anand1976:
> 
> Probably, though I am wondering about a couple things related to the approach:
> 
> * If the end of write stall mode can only be triggered when flush finishes (and not when a block is evicted from cache), would that also prevent thrashing?

@ajkr I believe this is the case, since its a one way relationship from WBM to the block cache. WBM doesn't know about the block cache memory usage.

db/db_impl/db_impl.h
akankshamahajan15: But lock protects ```state_``` so shouldn't ```state_ = State::BLOCKED;``` be inside the lock? Also can you explain about missed signals, I didn't get it.

anand1976: The current sequence is roughly as follows -
```
wbm->BeginWriteStall(wbm_stall_);
mutex_.Unlock();
...
wbm_stall_->Block();
```
The problem is, between the first ```mutex_.Unlock()``` and the call to ```Block()```, WBM could have signalled this thread. That signal will be lost, and ```Block()``` will call ```state_cv_.Wait()``` and not get woken up.

akankshamahajan15:
Ran ./db_bench --benchmarks="fillseq,readrandom,readseq" -cost_write_buffer_to_cache=true -db_write_buffer_size=10000000 (Cache size = 8MB) to test any hang.

db/db_impl/db_impl.cc
```diff
@@ -660,6 +665,11 @@ Status DBImpl::CloseHelper() {
     }
   }
 
+  if (wbm_stall_) {
```
anand1976, line 668: What's the current behavior if writes are stalled when ```DB::Close()``` is called? Does it wait for all pending writes to finish?

akankshamahajan15: Oh.  I missed the removal of DBs from queue in case of DB::Close(). I will update the code.

akankshamahajan15: I couldn't write a test to know the exact behavior. But I have updated the code to remove DB from stall in case ```Close``` is called. Will that cover the scenario?. From the code it doesn't seem like it waits for pending writes to finish. It only waits for background jobs to finish. Do I need to handle this scenario or its upto users to not call shutdown in case there is a pending write.

db/db_impl/db_impl.cc
```diff
@@ -270,6 +271,10 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
   // we won't drop any deletion markers until SetPreserveDeletesSequenceNumber()
   // is called by client and this seqnum is advanced.
   preserve_deletes_seqnum_.store(0);
+
+  if (write_buffer_manager_) {
+    wbm_stall_ = new WBMStallInterface();
```
anand1976, line 276: Can we just use ```std::unique_ptr```?

memtable/write_buffer_manager.cc
```diff
@@ -145,4 +178,47 @@ void WriteBufferManager::FreeMemWithCache(size_t mem) {
...
+      return;
+    }
+
+    // Get the instances from the list and call WBMStallInterface::Signal to
+    // change the state to running and unblock the DB instances.
+    // Check ShouldStall() incase stall got active by other DBs.
+    while (!ShouldStall() && !queue_.empty()) {
+      StallInterface* wbm_stall = queue_.front();
+      queue_.pop_front();
+      wbm_stall->Signal();
```
anand1976, line 210: Maybe we should release the lock and reacquire it after calling ```wbm_stall->Signal()```. In general I don't think its a good idea to hold locks across modules.

anand1976 (approved):
LGTM. A couple of minor comments.

db/db_impl/db_impl.cc
anand1976: Right, I don't see the code waiting even for write controller initiated stalls. And ```Close()``` doesn't wait for in progress reads either. So it seems reasonable to assume that reads/writes must be quiesced before the DB is closed/destructed.

PR #7900, opened 2021-01-27
Add the integrated BlobDB to the stress/crash tests

ltamasi:
Summary:
The patch adds support for the options related to the new BlobDB implementation
to `db_stress`, including support for dynamically adjusting them using `SetOptions`
when `set_options_one_in` and a new flag `allow_setting_blob_options_dynamically`
are specified. (The latter is used to prevent the options from being enabled when
incompatible features are in use.)

The patch also updates the `db_stress` help messages of the existing stacked BlobDB
related options to clarify that they pertain to the old implementation. In addition, it
adds the new BlobDB to the crash test script. In order to prevent a combinatorial explosion
of jobs and still perform whitebox/blackbox testing (including under ASAN/TSAN/UBSAN),
and to also test BlobDB in conjunction with atomic flush and transactions, the script sets
the BlobDB options in 10% of normal/`cf_consistency`/`txn` crash test runs.

Test Plan:
Ran `make check` and `db_stress`/`db_crashtest.py` with various options.

### DIFF
--- a/tools/db_crashtest.py
+++ b/tools/db_crashtest.py
@@ -271,7 +271,7 @@ blob_params = {
     # enabled/disabled during the test via SetOptions
     "enable_blob_files": lambda: random.choice([0] + [1] * 3),
     "min_blob_size": lambda: random.choice([0, 16, 256]),
-    "blob_file_size": lambda: random.choice([1048576, 268435456, 1073741824]),
+    "blob_file_size": lambda: random.choice([1048576, 16777216, 268435456, 1073741824]),
     "blob_compression_type": lambda: random.choice(["none", "snappy", "lz4", "zstd"]),
     "enable_blob_garbage_collection": lambda: random.choice([0] + [1] * 3),
     "blob_garbage_collection_age_cutoff": lambda: random.choice([0.0, 0.25, 0.5, 0.75, 1.0]),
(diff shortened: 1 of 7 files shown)


jay-zhuang (approved):
👍👍👍

tools/db_crashtest.py
```diff
@@ -356,6 +373,12 @@ def gen_cmd_params(args):
     if args.test_best_efforts_recovery:
         params.update(best_efforts_recovery_params)
 
+    # Best-effort recovery and BlobDB are currently incompatible. Test BE recovery
+    # if specified on the command line; otherwise, apply BlobDB related overrides
+    # with a 10% chance.
```
jay-zhuang, line 378: Is it worth having a BlobDB only stress test job?

jay-zhuang (approved):

tools/db_crashtest.py
ltamasi: That's what I'm trying to avoid :) Note that BlobDB should also work with features like transactions and atomic flush, so testing it by itself as well as in conjunction with those features would require setting up jobs for the direct product `{no sanitizer, ASAN, TSAN, UBSAN} x {no additional features, atomic flush, transactions}`, which is 12 jobs. If you toss in `{whitebox, blackbox}` as well, it's 24 jobs (and it might be more later).

PR #7901, opened 2021-01-27
Add support for specifying working directory in lego-determinator

jay-zhuang:
So we could change the repo or repo structure.

Test plan: works with existing job: https://www.internalfb.com/intern/sandcastle/job/13510799254899745/
works with internal repo: https://www.internalfb.com/intern/sandcastle/job/27021598137143731


pdillinger (approved):
LGTM

PR #7902, opened 2021-01-27
Accurate re-generate command in TARGETS file

pdillinger:
Summary: TIL we have different versions of TARGETS file generated with
options passed to buckifier. Someone thought they were totally fine to
squash the file by re-running the command to generate (pretty reasonable
assumption) but the command was incorrect due to missing the extra
argument used to generate THAT TARGETS file.

This change includes in the command written in the TARGETS header the
extra argument passed to buckify (when used).

Test Plan: manual, as in the (now fixed) comments at the top of
buckify_rocksdb.py


ajkr (approved):
Good improvement, I also didn't know this script took an argument and wrote it in TARGETS.

PR #7903, opened 2021-01-27
Accumulate blob file additions in VersionEdit during recovery

ltamasi:
Summary:
During recovery, RocksDB performs a kind of dummy flush; namely, entries
from the WAL are added to memtables, which then get written to SSTs and
blob files (if enabled) just like during a regular flush. Note that
multiple memtables might be flushed during recovery for the same column
family, for example, if the DB is reopened with a lower write buffer size,
and therefore, we need to make sure to collect all SST and blob file
additions. The patch fixes a bug in the earlier logic which resulted in
later blob file additions overwriting earlier ones.

Test Plan:
Added a unit test and ran `db_stress`.


jay-zhuang (approved):
LGTM.
Nice test👍

Issue #7869, opened 2021-01-15
rocksdbjni: Possible NPE in RocksDB.setOptions


stefan-zobel:
Version:
**RocksJava 6.15.2**

The Java code in RocksDB.java (method setOptions() in line 3575) derefences the ColumnFamilyHandle argument to obtain the native handle without checking it for null first (the code uses **columnFamilyHandle.nativeHandle_** without checking _columnFamilyHandle_ for **null** first).

Actual:

```
  /**
   * Change the options for the column family handle.
   *
   * @param columnFamilyHandle {@link org.rocksdb.ColumnFamilyHandle}
   *     instance, or null for the default column family.
   * @param mutableColumnFamilyOptions the options.
   *
   * @throws RocksDBException if an error occurs whilst setting the options
   */
  public void setOptions(
      /* @Nullable */final ColumnFamilyHandle columnFamilyHandle,
      final MutableColumnFamilyOptions mutableColumnFamilyOptions)
      throws RocksDBException {
    setOptions(nativeHandle_, columnFamilyHandle.nativeHandle_,
        mutableColumnFamilyOptions.getKeys(),
        mutableColumnFamilyOptions.getValues());
  }
```

Expected:

```
  /**
   * Change the options for the column family handle.
   *
   * @param columnFamilyHandle {@link org.rocksdb.ColumnFamilyHandle}
   *     instance, or null for the default column family.
   * @param mutableColumnFamilyOptions the options.
   *
   * @throws RocksDBException if an error occurs whilst setting the options
   */
  public void setOptions(
      /* @Nullable */final ColumnFamilyHandle columnFamilyHandle,
      final MutableColumnFamilyOptions mutableColumnFamilyOptions)
      throws RocksDBException {
    setOptions(nativeHandle_, columnFamilyHandle == null ? 0 : columnFamilyHandle.nativeHandle_,
        mutableColumnFamilyOptions.getKeys(),
        mutableColumnFamilyOptions.getValues());
  }
```

nivekuil:
Indeed, I just ran into this.
```
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f418725f314, pid=1165516, tid=1327833
#
# JRE version: OpenJDK Runtime Environment (15.0.1+9) (build 15.0.1+9)
# Java VM: OpenJDK 64-Bit Server VM (15.0.1+9, mixed mode, tiered, z gc, linux-amd64)
[...]

PR #7909, opened 2021-01-29
rocksdbjni: Possible NPE in RocksDB.setOptions #7869

stefan-zobel:
Fix for https://github.com/facebook/rocksdb/issues/7869


ajkr (approved):
LGTM

PR #7912, opened 2021-01-30
Override DeleteRangeCF in TransactionLog visitor

### LINKED ISSUE #7846: One-off "DeleteRangeCF not implemented" error calling TransactionDb::write
Running RocksDB 6.15.2 on x86_64 Linux.

We ran into a "DeleteRangeCF not implemented" error when calling `TransactionDb::write` with a batch containing a range delete, which appears to come from the default implementation of `WriteBatch::Handler::DeleteRangeCF`. It seems that the failure is rare, since we've been making batch writes with range deletes for quite a while without issue.

I'm assuming that there's a `Handler` implementation in a cold part of the write codepath that doesn't override the handler method. Looking through the codebase one possibility is `BatchCounter` in `TransactionLogIteratorImpl::UpdateCurrentWriteBatch`, but I don't know if that's reachable from a write call.

Any ideas?

sfackler:
Closes #7846

### DIFF
--- a/db/transaction_log_impl.cc
+++ b/db/transaction_log_impl.cc
@@ -277,6 +277,11 @@ void TransactionLogIteratorImpl::UpdateCurrentWriteBatch(const Slice& record) {
     Status SingleDeleteCF(uint32_t /*cf*/, const Slice& /*key*/) override {
       return Status::OK();
     }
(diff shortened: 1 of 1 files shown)


sfackler:
This was actually not the problem.

PR #7916, opened 2021-01-31
Fix build_version.cc generation error if GIT not found

mrambacher:
(Fixes a regression introduced in the build_version generation PR #7866 )

In the Makefile case, needed to ignore stderr on the tag (everywhere else was fine).

In the CMAKE case, no GIT implies "changes" so that we use the system date rather than the empty GIT date.

Test Plan:
Built in a tree that did not contain the ".git" directory.  Validated that no errors appeared during the build process and that the build version date was not empty.


jay-zhuang (approved):
LGTM. Thanks for the quick fix.

PR #7918, opened 2021-02-02
Makefile support to statically link external plugin code

ajkr:
Added support for detecting plugins linked in the "plugin/" directory and building them from our Makefile in a standardized way. See "plugin/README.md" for details. An example of a plugin that can be built in this way can be found in https://github.com/ajkr/dedupfs.

There will be more to do in terms of making this process more convenient and adding support for CMake.

Test Plan: my own plugin (https://github.com/ajkr/dedupfs) and also heard this patch worked with ZenFS.


Makefile
```diff
@@ -269,6 +272,12 @@ dummy := $(shell (export ROCKSDB_ROOT="$(CURDIR)"; \
 # this file is generated by the previous line to set build flags and sources
 include make_config.mk
 
+ROCKSDB_PLUGIN_MKS = $(foreach plugin, $(ROCKSDB_PLUGINS), plugin/$(plugin)/*.mk)
+include $(ROCKSDB_PLUGIN_MKS)
```
mrambacher, line 276: I am a little confused as to how one specifies ROCKSDB_PLUGINS and gets the plugin/* directory in the corresponding source tree.  More documentation would be helpful.

Makefile
```diff
@@ -1437,7 +1447,7 @@ librocksdb_env_basic_test.a: $(OBJ_DIR)/env/env_basic_test.o $(LIB_OBJECTS) $(TE
 	$(AM_V_at)$(AR) $(ARFLAGS) $@ $^
 
 db_bench: $(OBJ_DIR)/tools/db_bench.o $(BENCH_OBJECTS) $(TESTUTIL) $(LIBRARY)
-	$(AM_LINK)
+	$(AM_LINK_WHOLE)
```
mrambacher, line 1450: Same question as above.  Why not have db_bench depend on LIBRARY + PLUGIN_LIBRARIES?

Makefile
```diff
@@ -2076,6 +2086,10 @@ install-headers: gen-pc
 	for header in `$(FIND) "include/rocksdb" -type f -name *.h`; do \
 		install -C -m 644 $$header $(DESTDIR)/$(PREFIX)/$$header; \
 	done
+	for header in $(ROCKSDB_PLUGIN_HEADERS); do \
+		install -d $(DESTDIR)/$(PREFIX)/include/rocksdb/`dirname $$header`; \
+		install -C -m 644 $$header $(DESTDIR)/$(PREFIX)/include/rocksdb/$$header; \
+	done
```
mrambacher, line 2092: Separate out the plugin library and install it as well?

mrambacher (commented):
I am a little confused as to how this is supposed to work.  I am not certain I understand how adding the plugin to the master library solves anything (how do any of the function in the plugin get called in something like db_bench), and I do not understand how the directory tree gets established.

Makefile
ajkr: We're not separating it.

Makefile
ajkr: It's created by the user.

ajkr (commented):
> I am not certain I understand how adding the plugin to the master library solves anything (how do any of the function in the plugin get called in something like db_bench)

This is exactly demonstrated in the example: https://github.com/ajkr/dedupfs

> I do not understand how the directory tree gets established.

By the user. See the example

Makefile
```diff
@@ -269,6 +269,25 @@ dummy := $(shell (export ROCKSDB_ROOT="$(CURDIR)"; \
 # this file is generated by the previous line to set build flags and sources
 include make_config.mk
 
+# Linking with `AM_LINK_WHOLE` prevents linker from dropping unreferenced symbols,
+# which could include static variables registering plugins in `ObjectLibrary`.
+# To achieve this on Mac OS X, the static library must be preceded by `-force_load`
+# linker flag. Elsewhere, the library must be surrounded by `--whole-archive` and
+# `--no-whole-archive`.
+$(warning "platform $(PLATFORM)")
```
pdillinger, line 277: Debug print?

pdillinger (approved):
This looks to be a great first implementation of the agreed-upon design 👍👍

Makefile
ajkr: Good catch, will remove it.

ajkr (commented):
Thanks very much for the review!

ramvadiv (approved):

mrambacher:
I agree that this is a good first pass and is something that could potentially be marked "experimental".  There are quite a few issues I see before we could mark this is "ready":
- The size of the executables grows substantially in this mode;
- There is no way to tell what plugins are installed/available in the library;
- Without CMAKE support, this will not work on Windows and potentially other platforms
- There needs to be a means to build and run the tests against the plugins
- Support for shared libraries (and potentially external libraries) is also needed

This work dovetails with some of the work I have been doing in #6591.  I will write something up to bring those two efforts together and try to address some of these issues.

mrambacher:
Should there also be a way to add/build custom tests?  For example, if one viewed the "BackupableDB" suite as a plugin, how would one build those tests using this model?

ajkr:
> I agree that this is a good first pass and is something that could potentially be marked "experimental". There are quite a few issues I see before we could mark this is "ready":

The release note claims a "mechanism for building external plugin code into the RocksDB libraries/binaries". In my view, the public interface consists of the `*.mk` files in the plugin directory, which can specify source files, header files, and linker flags. I am pretty convinced we should provide the mechanism, and haven't heard concerns about the interface as described, so do not see a reason to mark the support introduced here as experimental.

> 
> * The size of the executables grows substantially in this mode;

This is addressed by no longer adding `-Wl,--whole-archive`.

> * There is no way to tell what plugins are installed/available in the library;
> * Without CMAKE support, this will not work on Windows and potentially other platforms
> * There needs to be a means to build and run the tests against the plugins
> * Support for shared libraries (and potentially external libraries) is also needed

I agree these would be nice to have. However, the release note only claims a "mechanism for building external plugin code into the RocksDB libraries/binaries".

> This work dovetails with some of the work I have been doing in #6591. I will write something up to bring those two efforts together and try to address some of these issues.

I think it is a great long-term direction. As it relates to this PR, I expect the "building-in" mechanism introduced here will still be needed.

PR #7921, opened 2021-02-02
Fix checkpoint stuck

1996fanrui:
## 1. Bug description:

When RocksDB Checkpoint, it may be stuck in `WaitUntilFlushWouldNotStallWrites` method.

### 2.1 Configuration parameters:

```yaml
Compaction Style : Universal 

max_write_buffer_number : 4
min_write_buffer_number_to_merge : 3
```

Checkpoint is usually very fast. When the Checkpoint is executed, `WaitUntilFlushWouldNotStallWrites` is called. If there are 2 Immutable MemTables, which are less than `min_write_buffer_number_to_merge`, they will not be flushed. But will enter this code.

```c++
// method: GetWriteStallConditionAndCause
if (mutable_cf_options.max_write_buffer_number> 3 &&
              num_unflushed_memtables >=
                  mutable_cf_options.max_write_buffer_number-1) {
     return {WriteStallCondition::kDelayed, WriteStallCause::kMemtableLimit};
}
```

code link: https://github.com/facebook/rocksdb/blob/fbed72f03c3d9e4fdca3e5993587ef2559ba6ab9/db/column_family.cc#L847

Checkpoint thought there was a FlushJob, but it didn't. So will always wait.

### 2.2 solution:

Increase the restriction: the `number of Immutable MemTable` >= `min_write_buffer_number_to_merge will wait`.

If there are other better solutions, you can correct me.

### 2.3 Code that can reproduce the problem:

https://github.com/1996fanrui/fanrui-learning/blob/flink-1.12/module-java/src/main/java/com/dream/rocksdb/RocksDBCheckpointStuck.java

## 3. Interesting point

This bug will be triggered only when `the number of sorted runs >= level0_file_num_compaction_trigger`.

Because there is a break in WaitUntilFlushWouldNotStallWrites. 

```c++
if (cfd->imm()->NumNotFlushed() <
        cfd->ioptions()->min_write_buffer_number_to_merge &&
    vstorage->l0_delay_trigger_count() <
        mutable_cf_options.level0_file_num_compaction_trigger) {
  break;
}
```

code link: https://github.com/facebook/rocksdb/blob/fbed72f03c3d9e4fdca3e5993587ef2559ba6ab9/db/db_impl/db_impl_compaction_flush.cc#L1974

Universal may have `l0_delay_trigger_count() >= level0_file_num_compaction_trigger`, so this bug is triggered.


1996fanrui:
Hi @zhichao-cao @ajkr , could you have time to help review it? Thank you.

ajkr:
Sorry for the delay - we will get to this eventually. One thing I wanted to point out is RocksDB slows down writes when `max_write_buffer_number - 1` memtables are filled, as long as `max_write_buffer_number > 3`. So by setting `min_write_buffer_number_to_merge` equal to `max_write_buffer_number - 1` , does it always slow down user writes whenever flush is ongoing?

1996fanrui:
> Sorry for the delay - we will get to this eventually. One thing I wanted to point out is RocksDB slows down writes when `max_write_buffer_number - 1` memtables are filled, as long as `max_write_buffer_number > 3`. So by setting `min_write_buffer_number_to_merge` equal to `max_write_buffer_number - 1` , does it always slow down user writes whenever flush is queued or running?

@ajkr Sorry for the late reply, because I was on vacation some time ago.

The point you mentioned does exist. I think when `max_write_buffer_number> 3`, `min_write_buffer_number_to_merge` should be <= `max_write_buffer_number-2`.

We previously set performance-unfriendly parameters during use:

```
max_write_buffer_number: 4
min_write_buffer_number_to_merge: 3
```

But I think: If users accidentally configure parameters that are not performance friendly, RocksDB should not get stuck. This patch should be necessary. I don’t know if my current fix is reasonable, I hope to get expert guidance and review. Thank you.

1996fanrui:
Hello @ajkr @zhichao-cao 
Can the current patch continue forward? thanks.

ajkr (approved):
Sorry for the delay. I agree it is fair to say that if user configures `min_write_buffer_number_to_merge == max_write_buffer_number - 1`, we should not delay writes while filling the last memtable. Thanks for the fix!

PR #7928, opened 2021-02-04
Possible hang in DBImpl::CloseHelper

akankshamahajan15:
Summary: In DBImpl::CloseHelper, we wait for bg_compaction_scheduled_
and bg_flush_scheduled_ to drop to 0. Unschedule is called prior
to cancel any unscheduled flushes/compactions. It is assumed that
anything in the high priority is a flush, and anything in the low
priority pool is a compaction. This assumption, however, is broken when
the high-pri pool is full.
As a result, bg_compaction_scheduled_ can go < 0 and bg_flush_scheduled_
will remain > 0 and DB can be in hang state.

Test Plan: Added new test case which hangs without the fix.

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -190,6 +190,7 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
       bg_flush_scheduled_(0),
       num_running_flushes_(0),
       bg_purge_scheduled_(0),
+      bg_flush_scheduled_low_(0),
       disable_delete_obsolete_files_(0),
       pending_purge_obsolete_files_(0),
       delete_obsolete_files_last_run_(clock_->NowMicros()),
@@ -522,7 +523,22 @@ Status DBImpl::CloseHelper() {
   int flushes_unscheduled = env_->UnSchedule(this, Env::Priority::HIGH);
   Status ret = Status::OK();
   mutex_.Lock();
+
   bg_bottom_compaction_scheduled_ -= bottom_compactions_unscheduled;
+  if (bg_flush_scheduled_low_) {
+    // Even if flush jobs were scheduled in Env::Priority::LOW, they should be
+    // consider part of flushes_unscheduled. Otherwise, bg_compaction_scheduled_
+    // can go < 0 and bg_flush_scheduled_ will remain > 0.
+    compactions_unscheduled -= bg_flush_scheduled_low_;
+    flushes_unscheduled += bg_flush_scheduled_low_;
+    bg_flush_scheduled_low_ = 0;
+#ifndef NDEBUG
+    std::pair<int, int> unscheduled_jobs =
+        std::make_pair(compactions_unscheduled, flushes_unscheduled);
+    TEST_SYNC_POINT_CALLBACK("DBImpl::CloseHelper:UnscheduleBGJobs:0",
+                             &unscheduled_jobs);
+#endif
+  }
   bg_compaction_scheduled_ -= compactions_unscheduled;
   bg_flush_scheduled_ -= flushes_unscheduled;
 
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -2185,6 +2185,7 @@ void DBImpl::MaybeScheduleFlushOrCompaction() {
       env_->Schedule(&DBImpl::BGWorkFlush, fta, Env::Priority::LOW, this,
                      &DBImpl::UnscheduleFlushCallback);
       --unscheduled_flushes_;
+      bg_flush_scheduled_low_++;
     }
   }
 
@@ -2505,6 +2506,9 @@ void DBImpl::BackgroundCallFlush(Env::Priority thread_pri) {
     InstrumentedMutexLock l(&mutex_);
     assert(bg_flush_scheduled_);
     num_running_flushes_++;
+    if (thread_pri == Env::Priority::LOW) {
+      bg_flush_scheduled_low_--;
+    }
 
     std::unique_ptr<std::list<uint64_t>::iterator>
         pending_outputs_inserted_elem(new std::list<uint64_t>::iterator(
--- a/db/db_flush_test.cc
+++ b/db/db_flush_test.cc
@@ -181,6 +181,59 @@ TEST_F(DBFlushTest, FlushInLowPriThreadPool) {
   ASSERT_EQ(1, num_compactions);
 }
 
+// Test when flush job is submitted to low priority thread pool and when DB is
+// closed in the meanwhile, CloseHelper doesn't hang.
+TEST_F(DBFlushTest, CloseDBWhenFlushInLowPri) {
+  Options options = CurrentOptions();
+  options.max_background_flushes = 1;
+  options.max_total_wal_size = 8192;
+
+  DestroyAndReopen(options);
+  CreateColumnFamilies({"cf1", "cf2"}, options);
+
+  env_->SetBackgroundThreads(0, Env::HIGH);
+  env_->SetBackgroundThreads(1, Env::LOW);
+  test::SleepingBackgroundTask sleeping_task_low;
+  int num_flushes = 0;
+
+  SyncPoint::GetInstance()->SetCallBack("DBImpl::BGWorkFlush",
+                                        [&](void* /*arg*/) { ++num_flushes; });
+
+  SyncPoint::GetInstance()->SetCallBack(
+      "DBImpl::CloseHelper:UnscheduleBGJobs:0", [&](void* arg) {
+        std::pair<int, int>* unscheduled_jobs =
+            static_cast<std::pair<int, int>*>(arg);
+        // unscheduled compaction jobs
+        ASSERT_EQ((*unscheduled_jobs).first, 0);
+        // unscheduled flush jobs.
+        ASSERT_EQ((*unscheduled_jobs).second, 1);
+      });
+
+  SyncPoint::GetInstance()->EnableProcessing();
+
+  ASSERT_OK(Put(0, "key1", DummyString(8192)));
+  // Block thread so that flush cannot be run and can be removed from the queue
+  // when called Unschedule.
+  env_->Schedule(&test::SleepingBackgroundTask::DoSleepTask, &sleeping_task_low,
+                 Env::Priority::LOW);
+  sleeping_task_low.WaitUntilSleeping();
+
+  // Trigger flush and flush job will be scheduled to LOW priority thread.
+  ASSERT_OK(Put(0, "key2", DummyString(8192)));
+
+  // Close DB and flush job in low priority queue will be removed without
+  // running.
+  Close();
+  sleeping_task_low.WakeUp();
+  sleeping_task_low.WaitUntilDone();
+  ASSERT_EQ(0, num_flushes);
+
+  TryReopenWithColumnFamilies({"default", "cf1", "cf2"}, options);
+  ASSERT_OK(Put(0, "key3", DummyString(8192)));
+  ASSERT_OK(Flush(0));
+  ASSERT_EQ(1, num_flushes);
+}
+
 TEST_F(DBFlushTest, ManualFlushWithMinWriteBufferNumberToMerge) {
   Options options = CurrentOptions();
   options.write_buffer_size = 100;
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -2068,6 +2068,10 @@ class DBImpl : public DB {
   // number of background obsolete file purge jobs, submitted to the HIGH pool
   int bg_purge_scheduled_;
 
+  // number of background memtable flush jobs scheduled to the LOW pool. They
+  // are not running and can be unscheduled.
+  int bg_flush_scheduled_low_;
+
   std::deque<ManualCompactionState*> manual_compaction_dequeue_;
 
   // shall we disable deletion of obsolete files


db/db_impl/db_impl.cc
```diff
@@ -522,7 +523,22 @@ Status DBImpl::CloseHelper() {
   int flushes_unscheduled = env_->UnSchedule(this, Env::Priority::HIGH);
   Status ret = Status::OK();
   mutex_.Lock();
+
   bg_bottom_compaction_scheduled_ -= bottom_compactions_unscheduled;
+  if (bg_flush_scheduled_low_) {
```
sherriiiliu, line 528: I also found the same issue and attempt to fix in #8007. Here I have a doubt. It looks you have an assumption that all bg_flush_scheduled_low_ will be successfully unscheduled, but there could be flushes not in unscheduled queue yet. Otherwise, the original logic could become something as below if following this assumption 
```
  bg_compaction_scheduled_ = 0;
  bg_flush_scheduled_ = 0;
```
Actually I met store hanging in DBTest.DynamicMemtableOptions unit test, and I can repro the issue with this test case on Windows. With your fix, I have the following assert error after running many times
```
Note: Google Test filter = DBTest.DynamicMemtableOptions
[----------] 1 test from DBTest
[ RUN      ] DBTest.DynamicMemtableOptions
Assertion failed: bg_flush_scheduled_, file e:\repo\rocksdb\db\db_impl\db_impl_compaction_flush.cc, line 2434
```
If you check that line, it means `bg_flush_scheduled_` is already 0 but there is still background flush running.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -2505,6 +2506,9 @@ void DBImpl::BackgroundCallFlush(Env::Priority thread_pri) {
     InstrumentedMutexLock l(&mutex_);
     assert(bg_flush_scheduled_);
     num_running_flushes_++;
+    if (thread_pri == Env::Priority::LOW) {
+      bg_flush_scheduled_low_--;
```
sherriiiliu, line 2510: I think better decrease bg_flush_scheduled_low_ after flush is really completed, like right after `bg_flush_scheduled_--` at line 2568

db/db_impl/db_impl.cc
akankshamahajan15: @sherriiiliu  Thanks for pointing it out. I will look into it and update the fix.

akankshamahajan15: @sherriiiliu  Can you share the Assertion failure. Also, bg_flush_scheduled_low_ represents jobs that are in the queue but not running yet.  I didn't fully understand your point here:                                       
 ```bg_compaction_scheduled_ = 0; // all compactions are unscheduled in LOW pool
  bg_flush_scheduled_ = 0; // all flushes are unscheduled in HIGH pool```
Jobs which are unscheduled. they are not in the queue. They are represented by ```unscheduled_flushes_```

akankshamahajan15: One more thing you mentioned "It looks you have an assumption that all bg_flush_scheduled_low_ will be successfully unscheduled, but there could be flushes not in unscheduled queue yet." But then don't we need to consider those because they are not running/haven't be scheduled yet.``` unscheduled_flushes_``` would represent those jobs.

sherriiiliu: The assertion failure is in `DBImpl::BackgroundCallFlush(Env::Priority thread_pri)`:
```
void DBImpl::BackgroundCallFlush(Env::Priority thread_pri) {
  bool made_progress = false;
  JobContext job_context(next_job_id_.fetch_add(1), true);

  TEST_SYNC_POINT("DBImpl::BackgroundCallFlush:start");

  LogBuffer log_buffer(InfoLogLevel::INFO_LEVEL,
                       immutable_db_options_.info_log.get());
  {
    assert(bg_flush_scheduled_); // fails here, bg_flush_scheduled_=0
    ...
```
I am not sure if you can repro on Linux or MaxOS, but it happens on my Win10.
My point is, the value returned by `env_->UnSchedule(this, Env::Priority::LOW)` is not always equal to the scheduled jobs in the corresponding threadpool by my observation. I think that's the reason why the original code logic is to decrease `bg_compaction_scheduled_ ` or `bg_flush_scheduled_` by the actual number that were successfully unscheduled, and then check if `bg_compaction_scheduled_ ` or `bg_flush_scheduled_` is greater than 0 to see if we need to wait any leftover jobs to finish in the while loop. Here you assume if there is any flush job in LOW pool, `bg_flush_scheduled_low_` will become 0 after unschedule is triggered, which looks contradict the above logic.

anand1976: This is a good point. Only the portion of ```bg_flush_scheduled_low_``` that's still in the thread pool queue will be returned by ```env->UnSchedule```. Therefore, adjust ```compactions_unscheduled``` and ```flushes_unscheduled``` using only ```bg_flush_scheduled_low_```.

How about considering ```bg_compaction_scheduled_ + bg_flush_scheduled_``` as one metric? Then you just have to subtract ```compactions_unscheduled + flushes_unscheduled``` and not have to track ```bg_flush_scheduled_low_```.

db/db_impl/db_impl_compaction_flush.cc
anand1976: Not sure if we even need ```bg_flush_scheduled_low_``` if we consider flushes and compactions together.

db/db_impl/db_impl.cc
sherriiiliu: Exactly what I think. In my fix #8007 , I only set `num_low_scheduled` to track the sum of `bg_compaction_scheduled_ + bg_flush_scheduled_`

ajkr (commented):
Nice find/fix, I never noticed flushes could run in the low-pri threads.

Was wondering about a different way - not sure if practical. Do you happen to know if we can adjust the `bg_compaction_scheduled_` and `bg_flush_scheduled_` counts in `UnscheduleCompactionCallback()` and `UnscheduleFlushCallback()`?

sherriiiliu:
> Nice find/fix, I never noticed flushes could run in the low-pri threads.
> 
> Was wondering about a different way - not sure if practical. Do you happen to know if we can adjust the `bg_compaction_scheduled_` and `bg_flush_scheduled_` counts in `UnscheduleCompactionCallback()` and `UnscheduleFlushCallback()`?
> 
> I guess one issue is we don't hold the DB mutex when calling each queue entry's `UnSchedule()`. We might have to make the counts atomic or grab mutex. Not sure if that's a big deal or not. Or if there are other issues.

`UnscheduleCompactionCallback()` and `UnscheduleFlushCallback()` only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess. About the DB mutex in Unschedule(), I am also concerned whether it will bring any other issue. @anand1976  and my proposal is, what about just counting flushes and compactions together, like `num_low_scheduled` in my PR #8007 , just use this variable to track flushed and compactions in low pool when unloading

akankshamahajan15:
> `UnscheduleCompactionCallback()` and `UnscheduleFlushCallback()` only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess. About the DB mutex in Unschedule(), I am also concerned whether it will bring any other issue. @anand1976 and my proposal is, what about just counting flushes and compactions together, like `num_low_scheduled` in my PR #8007 , just use this variable to track flushed and compactions in low pool when unloading

Your solution looks good. I will import that and we can go with that. I will merge the unit test later.

ajkr:
> UnscheduleCompactionCallback() and UnscheduleFlushCallback() only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess

Why do we need priority? I thought the goal is to maintain `bg_compaction_scheduled_` and `bg_flush_scheduled_`.

ajkr:
> > UnscheduleCompactionCallback() and UnscheduleFlushCallback() only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess
> 
> Why do we need priority? I thought the goal is to maintain `bg_compaction_scheduled_` and `bg_flush_scheduled_`.

Oh maybe you mean for bg_bottom_compaction_scheduled_. That is indeed unfortunate. We could register a `UnscheduleBottomCompactionCallback()` callback for it. I think this code is getting too hard to understand; we should take a lot of time to make it simpler if there's any possible way.

sherriiiliu:
> > > UnscheduleCompactionCallback() and UnscheduleFlushCallback() only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess
> > 
> > 
> > Why do we need priority? I thought the goal is to maintain `bg_compaction_scheduled_` and `bg_flush_scheduled_`.
> 
> Oh maybe you mean for bg_bottom_compaction_scheduled_. That is indeed unfortunate. We could register a `UnscheduleBottomCompactionCallback()` callback for it. I think this code is getting too hard to understand; we should take a lot of time to make it simpler if there's any possible way.



> > > UnscheduleCompactionCallback() and UnscheduleFlushCallback() only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess
> > 
> > 
> > Why do we need priority? I thought the goal is to maintain `bg_compaction_scheduled_` and `bg_flush_scheduled_`.
> 
> Oh maybe you mean for bg_bottom_compaction_scheduled_. That is indeed unfortunate. We could register a `UnscheduleBottomCompactionCallback()` callback for it. I think this code is getting too hard to understand; we should take a lot of time to make it simpler if there's any possible way.

In this way, I am wondering if we should just update `bg_compaction_scheduled_`, `bg_bottom_compaction_scheduled_` and `bg_flush_scheduled_` in callback functions, or move all counter updates into callback (such as `num_running_compactions_`,`num_running_flushes_`,etc., which will involve more changes). We can go with simple solution first and then like you said, spend time refactoring code later and make things simpler

ajkr:
> > > > UnscheduleCompactionCallback() and UnscheduleFlushCallback() only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess
> > > 
> > > 
> > > Why do we need priority? I thought the goal is to maintain `bg_compaction_scheduled_` and `bg_flush_scheduled_`.
> > 
> > 
> > Oh maybe you mean for bg_bottom_compaction_scheduled_. That is indeed unfortunate. We could register a `UnscheduleBottomCompactionCallback()` callback for it. I think this code is getting too hard to understand; we should take a lot of time to make it simpler if there's any possible way.
> 
> In this way, I am wondering if we should just update `bg_compaction_scheduled_`, `bg_bottom_compaction_scheduled_` and `bg_flush_scheduled_` in callback functions, or move all counter updates into callback (such as `num_running_compactions_`,`num_running_flushes_`,etc., which will involve more changes). We can go with simple solution first and then like you said, spend time refactoring code later and make things simpler

For locking, would it work to move this `mutex_.Lock()` up a few lines to cover the `UnSchedule()` calls? https://github.com/facebook/rocksdb/blob/063a68b9cd9e66727bde1e59f17aba53c0ae03db/db/db_impl/db_impl.cc#L530

Actually I start to think that change may be necessary irrespective of whether the `*scheduled_` accounting is in the callbacks. The callbacks are capable of destroying `Compaction` objects. `Compaction` destructor appears to assume DB mutex is held (it does not say so explicitly, but I checked its other clients and also the functions it calls, which led me to believe holding mutex is needed).

sherriiiliu:
> > > > > UnscheduleCompactionCallback() and UnscheduleFlushCallback() only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess
> > > > 
> > > > 
> > > > Why do we need priority? I thought the goal is to maintain `bg_compaction_scheduled_` and `bg_flush_scheduled_`.
> > > 
> > > 
> > > Oh maybe you mean for bg_bottom_compaction_scheduled_. That is indeed unfortunate. We could register a `UnscheduleBottomCompactionCallback()` callback for it. I think this code is getting too hard to understand; we should take a lot of time to make it simpler if there's any possible way.
> > 
> > 
> > In this way, I am wondering if we should just update `bg_compaction_scheduled_`, `bg_bottom_compaction_scheduled_` and `bg_flush_scheduled_` in callback functions, or move all counter updates into callback (such as `num_running_compactions_`,`num_running_flushes_`,etc., which will involve more changes). We can go with simple solution first and then like you said, spend time refactoring code later and make things simpler
> 
> For locking, would it work to move this `mutex_.Lock()` up a few lines to cover the `UnSchedule()` calls?
> 
> https://github.com/facebook/rocksdb/blob/063a68b9cd9e66727bde1e59f17aba53c0ae03db/db/db_impl/db_impl.cc#L530
> 
> Actually I start to think that change may be necessary irrespective of whether the `*scheduled_` accounting is in the callbacks. The callbacks are capable of destroying `Compaction` objects. `Compaction` destructor appears to assume DB mutex is held (it does not say so explicitly, but I checked its other clients and also the functions it calls, which led me to believe holding mutex is needed).

I tried moving `mutex_.Lock()` to cover the `UnSchedule()` calls, but it still happens that the `UnSchedule()` returns 0, so there are leftover flush jobs in low pool. My proposal is still counting flushes and compactions together in low pool, and minus them by the real unscheduled job counts to see if we have unfinished jobs that we need to wait on LOW. BTW, this PR belongs to @akankshamahajan15 so needs her assistance. I am joining this conversation as I am also trying to resolve this problem in my open PR #8007 , but we should only check in one since we are doing duplicate jobs

ajkr:
> > > > > > UnscheduleCompactionCallback() and UnscheduleFlushCallback() only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess
> > > > > 
> > > > > 
> > > > > Why do we need priority? I thought the goal is to maintain `bg_compaction_scheduled_` and `bg_flush_scheduled_`.
> > > > 
> > > > 
> > > > Oh maybe you mean for bg_bottom_compaction_scheduled_. That is indeed unfortunate. We could register a `UnscheduleBottomCompactionCallback()` callback for it. I think this code is getting too hard to understand; we should take a lot of time to make it simpler if there's any possible way.
> > > 
> > > 
> > > In this way, I am wondering if we should just update `bg_compaction_scheduled_`, `bg_bottom_compaction_scheduled_` and `bg_flush_scheduled_` in callback functions, or move all counter updates into callback (such as `num_running_compactions_`,`num_running_flushes_`,etc., which will involve more changes). We can go with simple solution first and then like you said, spend time refactoring code later and make things simpler
> > 
> > 
> > For locking, would it work to move this `mutex_.Lock()` up a few lines to cover the `UnSchedule()` calls?
> > https://github.com/facebook/rocksdb/blob/063a68b9cd9e66727bde1e59f17aba53c0ae03db/db/db_impl/db_impl.cc#L530
> > 
> > Actually I start to think that change may be necessary irrespective of whether the `*scheduled_` accounting is in the callbacks. The callbacks are capable of destroying `Compaction` objects. `Compaction` destructor appears to assume DB mutex is held (it does not say so explicitly, but I checked its other clients and also the functions it calls, which led me to believe holding mutex is needed).
> 
> I tried moving `mutex_.Lock()` to cover the `UnSchedule()` calls, but it still happens that the `UnSchedule()` returns 0, so there are leftover flush jobs in low pool. My proposal is still counting flushes and compactions together in low pool, and minus them by the real unscheduled job counts to see if we have unfinished jobs that we need to wait on LOW. BTW, this PR belongs to @akankshamahajan15 so needs her assistance. I am joining this conversation as I am also trying to resolve this problem in my open PR #8007 , but we should only check in one since we are doing duplicate jobs

Sorry for the confusion. I just meant that `mutex_` rearrangement can enable us to make the actual change, which is decrementing `bg_{flush,compaction,bottom_compaction}_scheduled_` inside the `Unschedule{Flush,Compaction,BottomCompaction}Callback()`s. `UnscheduleBottomCompactionCallback()` is a proposed new callback. These callbacks can modify the `DBImpl` variables via `reinterpret_cast<...>(arg)->db->...`

sherriiiliu:
> > > > > > > UnscheduleCompactionCallback() and UnscheduleFlushCallback() only takes compaction args, so we might need to find a way to pass in priority as well, which will change several method signatures I guess
> > > > > > 
> > > > > > 
> > > > > > Why do we need priority? I thought the goal is to maintain `bg_compaction_scheduled_` and `bg_flush_scheduled_`.
> > > > > 
> > > > > 
> > > > > Oh maybe you mean for bg_bottom_compaction_scheduled_. That is indeed unfortunate. We could register a `UnscheduleBottomCompactionCallback()` callback for it. I think this code is getting too hard to understand; we should take a lot of time to make it simpler if there's any possible way.
> > > > 
> > > > 
> > > > In this way, I am wondering if we should just update `bg_compaction_scheduled_`, `bg_bottom_compaction_scheduled_` and `bg_flush_scheduled_` in callback functions, or move all counter updates into callback (such as `num_running_compactions_`,`num_running_flushes_`,etc., which will involve more changes). We can go with simple solution first and then like you said, spend time refactoring code later and make things simpler
> > > 
> > > 
> > > For locking, would it work to move this `mutex_.Lock()` up a few lines to cover the `UnSchedule()` calls?
> > > https://github.com/facebook/rocksdb/blob/063a68b9cd9e66727bde1e59f17aba53c0ae03db/db/db_impl/db_impl.cc#L530
> > > 
> > > Actually I start to think that change may be necessary irrespective of whether the `*scheduled_` accounting is in the callbacks. The callbacks are capable of destroying `Compaction` objects. `Compaction` destructor appears to assume DB mutex is held (it does not say so explicitly, but I checked its other clients and also the functions it calls, which led me to believe holding mutex is needed).
> > 
> > 
> > I tried moving `mutex_.Lock()` to cover the `UnSchedule()` calls, but it still happens that the `UnSchedule()` returns 0, so there are leftover flush jobs in low pool. My proposal is still counting flushes and compactions together in low pool, and minus them by the real unscheduled job counts to see if we have unfinished jobs that we need to wait on LOW. BTW, this PR belongs to @akankshamahajan15 so needs her assistance. I am joining this conversation as I am also trying to resolve this problem in my open PR #8007 , but we should only check in one since we are doing duplicate jobs
> 
> Sorry for the confusion. I just meant that `mutex_` rearrangement can enable us to make the actual change, which is decrementing `bg_{flush,compaction,bottom_compaction}_scheduled_` inside the `Unschedule{Flush,Compaction,BottomCompaction}Callback()`s. `UnscheduleBottomCompactionCallback()` is a proposed new callback. These callbacks can decrement the corresponding `DBImpl` variable via `reinterpret_cast<...>(arg)->db->...`

Thanks for clarification. I tried with your suggestions and it works! I think this is the most elegant solution so far. I've created another PR #8125 and integrated @akankshamahajan15 's unit test. Regarding proposed new `UnscheduleBottomCompactionCallback()`, I just added `compaction_pri_` to get the priority info, so that we can decrement corresponding compaction counters correctly. @ajkr Please help review #8125 at your convenience. Thanks again for providing help!

akankshamahajan15:
abandoning the PR as issue will be fixed in #8125

PR #7929, opened 2021-02-04
Allow range deletions in `*TransactionDB` only when safe

### LINKED ISSUE #7913: TransactionDB now explicitly rejects range deletes in 6.15.
#7572 added missing error checks to various places. However, one in particular (https://github.com/facebook/rocksdb/pull/7572/files#diff-8d4f52efb53431efba2194e523696a9e147c6b24421f12d81b7aae9678495994R513-R516) caused TransactionDBs to start explicitly rejecting range deletes. The old behavior, where the delete would commit but fail to take locks on the affected keys, seems pretty clearly wrong, but it is a change that didn't end up in the release notes and took me a while to track down.

We were relying on the old behavior in an application I'm working on, but luckily we only perform the deletes on key ranges we know aren't visible to any transactions (it's happening in a background GC job). Am I right in thinking that we should be able to just perform the deletes on the base DB (i.e. `db->GetBaseDB()->DeleteRange(...)` rather than `db->DeleteRange(...)` as long as we are confident that we won't have any concurrent use of those keys?

It may also be worth updating the release notes to include this change for future people that come across it.

ajkr:
Explicitly reject all range deletions on `TransactionDB` or `OptimisticTransactionDB`, except when the user provides sufficient promises that allow us to proceed safely. The necessary promises are described in the API doc for `TransactionDB::DeleteRange()`. There is currently no way to provide enough promises to make it safe in `OptimisticTransactionDB`.

Fixes #7913.

Test Plan: unit tests covering the cases it's permitted/rejected

### DIFF
--- a/utilities/transactions/write_prepared_txn_db.cc
+++ b/utilities/transactions/write_prepared_txn_db.cc
@@ -157,7 +157,9 @@ Status WritePreparedTxnDB::WriteInternal(const WriteOptions& write_options_orig,
     // TODO(myabandeh): add an option to allow user skipping this cost
     SubBatchCounter counter(*GetCFComparatorMap());
     auto s = batch->Iterate(&counter);
-    assert(s.ok());
+    if (!s.ok()) {
+      return s;
+    }
     batch_cnt = counter.BatchCount();
     WPRecordTick(TXN_DUPLICATE_KEY_OVERHEAD);
     ROCKS_LOG_DETAILS(info_log_, "Duplicate key overhead: %" PRIu64 " batches",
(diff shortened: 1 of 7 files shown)


ajkr:
requested oncall review. I am planning to backport this to branches 6.15 and later.

utilities/transactions/write_prepared_txn_db.cc
```diff
@@ -157,7 +157,9 @@ Status WritePreparedTxnDB::WriteInternal(const WriteOptions& write_options_orig,
     // TODO(myabandeh): add an option to allow user skipping this cost
     SubBatchCounter counter(*GetCFComparatorMap());
     auto s = batch->Iterate(&counter);
-    assert(s.ok());
+    if (!s.ok()) {
```
ajkr, line 160: The relevant case this is non-OK is when `batch` contains a range deletion. That case is possible through API misuse (the tests exercise this path) so I think it should be a runtime check instead of an assertion.

ltamasi (approved):

ajkr:
Thanks for the review!

ltamasi:
Thanks for the fix! :)

PR #7931, opened 2021-02-04
make PerfStepTimer struct smaller by reordering members

jsteemann:
On x86_64, this makes the struct 8 bytes smaller, so creating a PerfStepTimer on the stack will use slightly less stack space.


jsteemann:
The test failures that were uncovered by the CI all look unrelated to the changes in this PR. :thinking:

ajkr (approved):
LGTM

PR #7932, opened 2021-02-04
db_bench: dump cpu info for Mac.

devnexen:


zhichao-cao:
Thanks for the contribution! Can you do a "make format" to fix the format issue?

tools/db_bench_tool.cc
```diff
@@ -2576,14 +2581,15 @@ class Benchmark {
...
+#if defined(__linux__) || defined(__APPLE__)
     time_t now = time(nullptr);
     char buf[52];
     // Lint complains about ctime() usage, so replace it with ctime_r(). The
     // requirement is to provide a buffer which is at least 26 bytes.
     fprintf(stderr, "Date:       %s",
             ctime_r(&now, buf));  // ctime_r() adds newline
 
+#if defined(__linux)
     FILE* cpuinfo = fopen("/proc/cpuinfo", "r");
```
mrambacher, line 2593: The code varies between using __linux and __linux__.  Probably should be one or the other...

mrambacher (approved):
Outside of the comment, this LGTM.

I do wonder if this should be promoted (along with potentially some other stuff -- like NUMA and memory stats) to a method off Env.  Then it would be available for other tools and could possibly be involved in run-time decisions (such how many threads to use, etc).  

Not expecting that to be done as part of this PR, but it if seems like a worthwhile thing to do, perhaps another issue/suggestion should be filed to do that work in the future...

zhichao-cao:
> Outside of the comment, this LGTM.
> 
> I do wonder if this should be promoted (along with potentially some other stuff -- like NUMA and memory stats) to a method off Env. Then it would be available for other tools and could possibly be involved in run-time decisions (such how many threads to use, etc).
> 
> Not expecting that to be done as part of this PR, but it if seems like a worthwhile thing to do, perhaps another issue/suggestion should be filed to do that work in the future...

I believe it would be useful, if we can dump the cpu Information as a method of Env.

devnexen:
seems to be sort of timeouts.

devnexen:
I rebased and seems better now.

zhichao-cao:
> I rebased and seems better now.

Thanks for the check! LGTM.

zhichao-cao (commented):
LGTM

PR #7935, opened 2021-02-05
Fix db_bench_tool_test

ltamasi:
Summary:
The patch fixes the build for `db_bench_tool_test` and makes the tests pass.
Namely, it fixes the following issues:

* #7703 removed the member variable `fs_` but the test case `OptionsFileMultiLevelUniversal`
was not updated.
* #7344 fixed the `OptionsFile` test case for the case when Snappy is *not* available but at the
same time broke it for the case when it *is* available. (The test used a default-constructed
`ColumnFamilyOptions` object, and the default value of the `compression` option is either
Snappy or no compression depending on whether Snappy is supported.)
* The test used `google::ParseCommandLineFlags` instead of
`GFLAGS_NAMESPACE::ParseCommandLineFlags`.

Test Plan:
Ran the test both with and without Snappy support.

### DIFF
--- a/tools/db_bench_tool_test.cc
+++ b/tools/db_bench_tool_test.cc
@@ -69,7 +69,7 @@ class DBBenchTest : public testing::Test {
     opt.dump_malloc_stats = true;  // db_bench uses a different default
     opt.compaction_style = style;
     opt.num_levels = levels;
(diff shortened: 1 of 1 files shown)


zhichao-cao (approved):
LGTM, thanks for the fix!

PR #7936, opened 2021-02-05
Fix handling of Mutable options; Allow DB::SetOptions to update mutable TableFactory Options

### LINKED ISSUE #7385: block_size and block_restart_interval are not dynamically changeable
### Expected behavior
The comment indicates `block_size` is changeable:
https://github.com/facebook/rocksdb/blob/9a5886bd8c71aaa4466e0a8b1d7d1f04ae452b6a/include/rocksdb/table.h#L166-L167

### Actual behavior
But it's actually not:

https://github.com/facebook/rocksdb/blob/b9bb59d49df264363d7250ceba59213706b3bb61/options/cf_options.cc#L511-L514
```
db_->SetOptions({{"block_based_table_factory", "{block_size=7k;}"}});
// got error: Invalid argument: Option not changeable: block_based_table_factory
```

mrambacher:
Added a "only_mutable_options" flag to the ConfigOptions.  When set, the Configurable methods will only look at/update options that are marked as kMutable.  

Fixed DB::SetOptions to allow for the update of any mutable TableFactory options.  Fixes #7385.

Added tests for the new flag.  Updated HISTORY.md


options/configurable.cc
```diff
@@ -239,9 +241,15 @@ Status Configurable::ParseOption(const ConfigOptions& config_options,
                                  const OptionTypeInfo& opt_info,
                                  const std::string& opt_name,
                                  const std::string& opt_value, void* opt_ptr) {
-  if (opt_info.IsMutable() || opt_info.IsConfigurable()) {
-    return opt_info.Parse(config_options, opt_name, opt_value, opt_ptr);
-  } else if (prepared_) {
+  if (opt_info.IsMutable()) {
+    if (config_options.only_mutable_options) {
+      ConfigOptions copy = config_options;
+      copy.only_mutable_options = false;
```
akankshamahajan15, line 247: Can you explain this section of code. If ```config_options.only_mutable_options``` is set true why does```copy.only_mutable_options = false```.  Or may be add comments in the code.

akankshamahajan15 (approved):
LGTM

options/configurable.cc
mrambacher: I will do both (push a comment in the next update).  

There are two things of interest used repeatedly.  One is whether or not the field is marked as IsMutable.  Another is if we are operating on only mutable options or not.

If a field is marked as Mutable, then all of the fields in it should inherit its mutable property even if they are not explicitly marked as mutable.   In other words, if a field is marked as mutable, we should toggle whether or not we are examining only mutable options. 

For example, if an object "o" has a member in it "Configurable c" and "c" is marked as Mutable, then all of the elements of "c" (when accessed from "o") will also be mutable even if they are not marked as such explicitly.

akankshamahajan15: ahh.. "One is whether or not the field is marked as IsMutable. Another is if we are operating on only mutable options or not." This makes sense. Thanks for explaining it.

zhichao-cao:
Can you explain a little bit more on the motivation of adding a "only_mutable_options" flag to the ConfigOptions?

include/rocksdb/convenience.h
```diff
@@ -56,6 +56,9 @@ struct ConfigOptions {
   // Whether or not to invoke PrepareOptions after configure is called.
   bool invoke_prepare_options = true;
 
+  // When true, only options that are mutable will be touched
+  bool only_mutable_options = false;
```
zhichao-cao, line 60: How about "mutable_options_only"?

mrambacher:
> Can you explain a little bit more on the motivation of adding a "only_mutable_options" flag to the ConfigOptions?

Are you asking for an explanation here or in the code/HISTORY?

Options can be marked as Mutable or not.  When the "only_mutable_options=false", all options are evaluated -- whether they are marked as Mutable or not.  When "only_mutable_options="true", any option not marked as Mutable is either ignored (in the case of string/equals methods) or results in an error (in the case of Configure).

This flag is important for things "sub options" -- like TableFactory options.  The table is not part of options that are mutable, and which of the table options are mutable are not known.  By telling the configure methods that the configure is only to apply to the "mutable options", something like SetOptions will be able to update only the ColumnFamilyOptions that are mutable -- meaning those that came directly from MutableCFOptions, but also those that have mutable characteristics.  For example, the "block_size" table attribute is mutable, but the "no_block_cache" is not.

Note that as more features become plugins, there may be different items that are Mutable.  This setup should allow more options to become mutable without changing the infrastructure or making the infrastructure tightly coupled to the implementations.

zhichao-cao:
> > Can you explain a little bit more on the motivation of adding a "only_mutable_options" flag to the ConfigOptions?
> 
> Are you asking for an explanation here or in the code/HISTORY?
> 
> Options can be marked as Mutable or not. When the "only_mutable_options=false", all options are evaluated -- whether they are marked as Mutable or not. When "only_mutable_options="true", any option not marked as Mutable is either ignored (in the case of string/equals methods) or results in an error (in the case of Configure).
> 
> This flag is important for things "sub options" -- like TableFactory options. The table is not part of options that are mutable, and which of the table options are mutable are not known. By telling the configure methods that the configure is only to apply to the "mutable options", something like SetOptions will be able to update only the ColumnFamilyOptions that are mutable -- meaning those that came directly from MutableCFOptions, but also those that have mutable characteristics. For example, the "block_size" table attribute is mutable, but the "no_block_cache" is not.
> 
> Note that as more features become plugins, there may be different items that are Mutable. This setup should allow more options to become mutable without changing the infrastructure or making the infrastructure tightly coupled to the implementations.

Thanks for the explanation! Maybe add a summary of these as a comments for bool only_mutable_options = false in convience.h would be helpful.

zhichao-cao (commented):
Can you do a rebase again? Looks like your changes has mixed with other's commits.

zhichao-cao (approved):
LGTM, thanks for the fix!

PR #7938, opened 2021-02-06
Fix an assertion failure in range locking, locktree code.

spetrunia:
Fix this scenario:
trx1> acquire shared lock on $key
trx2> acquire shared lock on the same $key
trx1> attempt to acquire a unique lock on $key.

Lock acquisition will fail, and deadlock detection will start.
It will call iterate_and_get_overlapping_row_locks() which will
produce a list with two locks (shared locks by trx1 and trx2).

However the code in lock_request::build_wait_graph() was not prepared
to find the lock by the same transaction in the list of conflicting
locks. Fix it to ignore it.

(One may suggest to fix iterate_and_get_overlapping_row_locks() to not
include locks by trx1. This is not a good idea, because that function
is also used to report all locks currently held)

### DIFF
--- a/utilities/transactions/lock/range/range_locking_test.cc
+++ b/utilities/transactions/lock/range/range_locking_test.cc
@@ -136,6 +136,41 @@ TEST_F(RangeLockingTest, MyRocksLikeUpdate) {
   delete txn0;
 }
 
+TEST_F(RangeLockingTest, UpgradeLockAndGetConflict) {
+  WriteOptions write_options;
+  TransactionOptions txn_options;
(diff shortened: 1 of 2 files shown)


zhichao-cao:
@cheng-chang You may have more context on this.

spetrunia:
I was studying where/what kind of assertion to add, and I've found a better place where to put the fix in.  I think assertion is now un-necessary.  @cheng-chang , PTAL.

ajkr (approved):
Makes sense, thanks for the fix and the test!

PR #7942, opened 2021-02-08
Avoid self-move-assign in pop operation of binary heap.

neunhoef:
The current implementation of a binary heap in `util/heap.h` does a move-assign in the `pop` method. In the case that there is exactly one element stored in the heap, this ends up being a self-move-assign. This can cause trouble with certain classes, which are not prepared for this. Furthermore, it trips up the glibc STL debugger (`-D_GLIBCXX_DEBUG`), which produces an assertion failure in this case.

This PR addresses this problem by not doing the (unnecessary in this case) move-assign if there is only one element in the heap.

### DIFF
--- a/util/heap.h
+++ b/util/heap.h
@@ -72,7 +72,12 @@ class BinaryHeap {
 
   void pop() {
     assert(!empty());
-    data_.front() = std::move(data_.back());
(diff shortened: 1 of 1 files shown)


ajkr (approved):
LGTM

PR #7943, opened 2021-02-08
Build a full RocksDB on M1 macs

xavierd:
With M1 macs being available, it is possible that RocksDB will be built on them, without the resulting artifacts to be intended for iOS, where a non-lite RocksDB is needed.

It is not clear to me why the ROCKSDB_LITE cmake option isn't used for iOS consumer, so sending this pull request as a way to foster discussion and to find a path forward to get a full RocksDB build on M1.


adamretter:
@xavierd FYI - Using Make instead of CMake I have been able to build on the Apple DTK without issue.

xavierd:
@adamretter, yep it does build properly with CMake, but for instance EdenFS (https://github.com/facebookexperimental/eden), relies on a non-lite version of RocksDB to build with CMake :).

jay-zhuang (approved):
LGTM.
IOS user, if there's any, can still build it by passing the flags.

PR #7944, opened 2021-02-08
Update clang-format-diff.py path

riversand963:
Summary:
Recent Github actions of format checking fail due to invalid location
from where clang-format-diff.py is downloaded. Update the path to point
to a stable, archived location.

Test Plan: manually check the result of Github action.

### DIFF
--- a/.github/workflows/sanity_check.yml
+++ b/.github/workflows/sanity_check.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Download clang-format-diff.py
       uses: wei/wget@v1
       with:
-        args: https://raw.githubusercontent.com/llvm-mirror/clang/master/tools/clang-format/clang-format-diff.py
+        args: https://raw.githubusercontent.com/llvm/llvm-project/main/clang/tools/clang-format/clang-format-diff.py
 
     - name: Check format
       run: VERBOSE_CHECK=1 make check-format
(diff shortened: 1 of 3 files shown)


riversand963:
cc @mrambacher 
After #7866 , running `make check-format` and `make check-buck-targets` will cause building the source files, which is unnecessary. I haven't figured out exactly why, but this PR is able to fix.

.github/workflows/sanity_check.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     - name: Download clang-format-diff.py
       uses: wei/wget@v1
       with:
-        args: https://raw.githubusercontent.com/llvm/llvm-project/master/clang/tools/clang-format/clang-format-diff.py
+        args: https://raw.githubusercontent.com/llvm-mirror/clang/master/tools/clang-format/clang-format-diff.py
```
ltamasi, line 35: Thanks @riversand963 ! Was about to put out a PR when I noticed yours :) I believe we could stick with the official repo (i.e. no need to move to the unofficial mirror) as the script is still there under `https://raw.githubusercontent.com/llvm/llvm-project/main/clang/tools/clang-format/clang-format-diff.py`

ltamasi: The only thing that's different is the branch name.

ltamasi (approved):

.github/workflows/sanity_check.yml
riversand963: Sure, let me update.

riversand963:

PR #7951, opened 2021-02-10
Make BlockBasedTable::kMaxAutoReadAheadSize configurable

akankshamahajan15:
Summary: RocksDB does auto-readahead for iterators on noticing more
than two reads for a table file. The readahead starts at 8KB and doubles on every
additional read upto BlockBasedTable::kMaxAutoReadAheadSize which is
256*1024.
This PR adds a new option BlockBasedTableOptions::max_auto_readahead_size which
replaces BlockBasedTable::kMaxAutoReadAheadSize and the new option can be
configured.
If max_auto_readahead_size is set 0 then no implicit auto prefetching will
be done. If max_auto_readahead_size provided is less than
8KB (which is initial readahead size used by rocksdb in case of
auto-readahead), readahead size will remain same as max_auto_readahead_size.

Test Plan: Add new unit test case.

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/include/rocksdb/table.h
+++ b/include/rocksdb/table.h
@@ -457,7 +457,10 @@ struct BlockBasedTableOptions {
   // DB::SetOptions({{"block_based_table_factory",
   //                  "{max_auto_readahead_size=0;}"}}));
   //
-  //  Default: 256 KB (256 * 1024).
+  // Changing the value dynamically will only affect files opened after the
+  // change.
+  //
+  // Default: 256 KB (256 * 1024).
   size_t max_auto_readahead_size = 256 * 1024;
 };
 
--- a/include/rocksdb/table.h
+++ b/include/rocksdb/table.h
@@ -437,13 +437,18 @@ struct BlockBasedTableOptions {
       IndexShorteningMode::kShortenSeparators;
 
   // RocksDB does auto-readahead for iterators on noticing more than two reads
-  // for a table file. The readahead starts at 8KB and doubles on every
-  // additional read upto max_auto_readahead_size and max_auto_readahead_size
-  // can be configured.
-  // If max_auto_readahead_size is set 0 then no implicit auto prefetching will
-  // be done. If max_auto_readahead_size provided is less than 8KB (which is
-  // initial readahead size used by rocksdb in case of auto-readahead),
-  // readahead size will remain same as max_auto_readahead_size.
+  // for a table file if user doesn't provide readahead_size. The readahead
+  // starts at 8KB and doubles on every additional read upto
+  // max_auto_readahead_size and max_auto_readahead_size can be configured.
+  //
+  // Special Value: 0 - If max_auto_readahead_size is set 0 then no implicit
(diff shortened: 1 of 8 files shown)


akankshamahajan15:
Depends on PR https://github.com/facebook/rocksdb/pull/7936.

table/block_based/block_based_table_reader.h
```diff
@@ -65,9 +65,6 @@ class BlockBasedTable : public TableReader {
 
   // All the below fields control iterator readahead
   static const size_t kInitAutoReadaheadSize = 8 * 1024;
-  // Found that 256 KB readahead size provides the best performance, based on
-  // experiments, for auto readahead. Experiment data is in PR #3282.
-  static const size_t kMaxAutoReadaheadSize;
   static const int kMinNumFileReadsToStartAutoReadahead = 2;
```
mrambacher, line 68: Is there a reason that you want to configure the MaxAutoReadaheadSize but leave the MinNumFilesToStartAutoReadahead and InitAutoReadaheadSize as constant?

akankshamahajan15: It was a requirement from a service owner to configure kMaxAutoReadaheadSize so that they disable prefetching for L0 files and enable it for other level files by configuring kMaxAutoReadaheadSize. Users can specify initial readahead_size in ReadOptions. kInitAutoReadaheadSize is used if user didn't specify  and after few seqeuntial reads RocksDB enable automatic prefetching.

include/rocksdb/table.h
```diff
@@ -435,6 +435,23 @@ struct BlockBasedTableOptions {
...
+  // initial readahead size used by rocksdb in case of auto-readahead),
+  // readahead size will remain same as max_auto_readahead_size.
+  //
+  // Found that 256 KB readahead size provides the best performance, based on
+  // experiments, for auto readahead. Experiment data is in PR #3282.
+  //
+  // This parameter can be changed dynamically by
+  // DB::SetOptions({{"block_based_table_factory",
+  //                  "{max_auto_readahead_size=0;}"}}));
+  size_t max_auto_readahead_size = 256 * 1024;
```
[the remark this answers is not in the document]
akankshamahajan15: I will update the code and the comments.

include/rocksdb/table.h
```diff
@@ -435,6 +435,30 @@ struct BlockBasedTableOptions {
...
+  // max_auto_readahead_size.
+  //
+  // Value should be provided along with KB i.e. 256 * 1024 as it will prefetch
+  // the blocks.
+  //
+  // Found that 256 KB readahead size provides the best performance, based on
+  // experiments, for auto readahead. Experiment data is in PR #3282.
+  //
+  // This parameter can be changed dynamically by
+  // DB::SetOptions({{"block_based_table_factory",
```
anand1976, line 457: Also mention that changing the value will only affect files opened after the change.

anand1976 (approved):
LGTM. ```HISTORY.md``` needs to be updated

PR #7957, opened 2021-02-11
Update internal build script

jay-zhuang:
For internal build.

### DIFF
--- a/tools/regression_test.sh
+++ b/tools/regression_test.sh
@@ -158,7 +158,7 @@ function init_arguments {
 
   current_time=$(date +"%F-%H:%M:%S")
   RESULT_PATH=${RESULT_PATH:-"$1/results/$current_time"}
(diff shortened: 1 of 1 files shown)


tools/regression_test.sh
```diff
@@ -367,16 +367,6 @@ function exit_on_error {
...
-function checkout_rocksdb {
-  echo "Checking out commit $1 ..."
-
-  git fetch --all
-  exit_on_error $?
-
-  git checkout $1
-  exit_on_error $?
-}
-
```
ltamasi, line 379: I'm assuming this isn't called from anywhere, right? :)

ltamasi (approved):

tools/regression_test.sh
jay-zhuang: Yes, it's dead code.

PR #7958, opened 2021-02-12
Add support for IOTracing in blob files

akankshamahajan15:
Summary: Add support for IOTracing in blob files

Test Plan: Add a new test and checked manually the trace_file for blob
files being recorded during read and write.

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/blob/blob_file_builder.cc
+++ b/db/blob/blob_file_builder.cc
@@ -21,6 +21,7 @@
 #include "rocksdb/slice.h"
 #include "rocksdb/status.h"
 #include "test_util/sync_point.h"
+#include "trace_replay/io_tracer.h"
 #include "util/compression.h"
 
 namespace ROCKSDB_NAMESPACE {
@@ -33,12 +34,13 @@ BlobFileBuilder::BlobFileBuilder(
     const std::string& column_family_name, Env::IOPriority io_priority,
     Env::WriteLifeTimeHint write_hint,
     std::vector<std::string>* blob_file_paths,
-    std::vector<BlobFileAddition>* blob_file_additions)
+    std::vector<BlobFileAddition>* blob_file_additions,
+    const std::shared_ptr<IOTracer>& io_tracer)
     : BlobFileBuilder([versions]() { return versions->NewFileNumber(); }, env,
                       fs, immutable_cf_options, mutable_cf_options,
                       file_options, job_id, column_family_id,
                       column_family_name, io_priority, write_hint,
-                      blob_file_paths, blob_file_additions) {}
+                      blob_file_paths, blob_file_additions, io_tracer) {}
 
 BlobFileBuilder::BlobFileBuilder(
     std::function<uint64_t()> file_number_generator, Env* env, FileSystem* fs,
@@ -48,7 +50,8 @@ BlobFileBuilder::BlobFileBuilder(
     const std::string& column_family_name, Env::IOPriority io_priority,
     Env::WriteLifeTimeHint write_hint,
     std::vector<std::string>* blob_file_paths,
-    std::vector<BlobFileAddition>* blob_file_additions)
+    std::vector<BlobFileAddition>* blob_file_additions,
+    const std::shared_ptr<IOTracer>& io_tracer)
     : file_number_generator_(std::move(file_number_generator)),
       fs_(fs),
       immutable_cf_options_(immutable_cf_options),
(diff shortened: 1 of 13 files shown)


db/blob/blob_file_builder.cc
```diff
@@ -33,12 +34,13 @@ BlobFileBuilder::BlobFileBuilder(
     const std::string& column_family_name, Env::IOPriority io_priority,
     Env::WriteLifeTimeHint write_hint,
     std::vector<std::string>* blob_file_paths,
-    std::vector<BlobFileAddition>* blob_file_additions)
+    std::vector<BlobFileAddition>* blob_file_additions,
+    const std::shared_ptr<IOTracer>& io_tracer)
```
ltamasi, line 38: Could you move the new parameter so it is between `write_hint` and `blob_file_paths` ? (That way, input parameters would precede output parameters, i.e. `blob_file_paths` and `blob_file_additions`.)

db/blob/blob_file_builder.h
```diff
@@ -87,6 +90,7 @@ class BlobFileBuilder {
   std::unique_ptr<BlobLogWriter> writer_;
   uint64_t blob_count_;
   uint64_t blob_bytes_;
+  std::shared_ptr<IOTracer> io_tracer_;
```
ltamasi, line 93: Could you please also move this so it's between `write_hint_` and `blob_file_paths_` ?

db/blob/blob_file_builder.cc
```diff
@@ -48,7 +50,8 @@ BlobFileBuilder::BlobFileBuilder(
...
```
ltamasi, line 54: Same here.

db/blob/blob_file_cache.h
```diff
@@ -44,6 +46,7 @@ class BlobFileCache {
   HistogramImpl* blob_file_read_hist_;
 
   static constexpr size_t kNumberOfMutexStripes = 1 << 7;
+  std::shared_ptr<IOTracer> io_tracer_;
```
ltamasi, line 49: Could you please move this right below `blob_file_read_hist_` (so it precedes the class constant)?

db/blob/blob_file_reader.cc
```diff
@@ -20,11 +20,13 @@
...
-    const FileOptions& file_options, uint32_t column_family_id,
-    HistogramImpl* blob_file_read_hist, uint64_t blob_file_number,
-    std::unique_ptr<BlobFileReader>* blob_file_reader) {
+Status BlobFileReader::Create(const ImmutableCFOptions& immutable_cf_options,
+                              const FileOptions& file_options,
+                              uint32_t column_family_id,
+                              HistogramImpl* blob_file_read_hist,
+                              uint64_t blob_file_number,
+                              std::unique_ptr<BlobFileReader>* blob_file_reader,
+                              const std::shared_ptr<IOTracer>& io_tracer) {
```
ltamasi, line 29: Similarly to the above, `blob_file_reader` is the output parameter of this method; could we keep it in the last position?

db/blob/blob_file_reader.cc
```diff
@@ -69,7 +71,8 @@ Status BlobFileReader::OpenFile(
     const ImmutableCFOptions& immutable_cf_options,
     const FileOptions& file_opts, HistogramImpl* blob_file_read_hist,
     uint64_t blob_file_number, uint64_t* file_size,
-    std::unique_ptr<RandomAccessFileReader>* file_reader) {
+    std::unique_ptr<RandomAccessFileReader>* file_reader,
+    const std::shared_ptr<IOTracer>& io_tracer) {
```
ltamasi, line 75: Same applies here.

db/blob/db_blob_basic_test.cc
```diff
@@ -210,6 +210,36 @@ TEST_F(DBBlobBasicTest, GetBlob_IndexWithInvalidFileNumber) {
...
+
+  constexpr char key[] = "key";
+  constexpr char blob_value[] = "blob_value";
+
+  ASSERT_OK(Put(key, blob_value));
+  ASSERT_OK(Flush());
+  ASSERT_EQ(Get(key), blob_value);
+
+  ASSERT_OK(db_->EndIOTrace());
+  ASSERT_OK(env_->FileExists(trace_file));
```
ltamasi, line 235: Question: is there a way to make sure that blob file I/Os specifically got traced, using e.g. `IOTraceReader`? I believe this test case would pass even without the current changes b/c we also read/write an SST.

ltamasi (commented):
Thanks so much for the PR @akankshamahajan15 !

ltamasi (approved):

PR #7962, opened 2021-02-14
First pass at a CompactionService definition

mrambacher:
This is an initial pass at a definition for a pluggable CompactionService.    Users will be able to provide alternative implementations to perform compaction differently -- such as sending the job to a remote process.

This feature is completely EXPERIMENTAL and subject to change.  Currently the interface is not used, but will be in subsequent updates/PRs.


include/rocksdb/compaction_service.h
```diff
@@ -0,0 +1,96 @@
...
+  // Returns the name of this compaction service.
+  virtual const char *Name() const = 0;
+  // Starts a compaction job using the supplied parameters.
+  // This method starts or schedules the compaction and returns a "job_id"
+  // that can be used to manage the job.
+  //
+  // If an implementation does not support the specified options, it can
+  // return "Status::NotSupported", in which case the default compaction service
+  // will attempt the job.
+  virtual Status Start(const CompactionServiceOptions & compaction_options,
```
jay-zhuang, line 72: As we discussed in the chat, we would prefer a binary format to keep the compaction details internal and give us more controls.

include/rocksdb/compaction_service.h
```diff
@@ -0,0 +1,96 @@
...
+                       const Options& options,
```
jay-zhuang, line 73: could Options be combined with compaction_options? (ideally we should only send options needed by compactor worker).

include/rocksdb/compaction_service.h
```diff
@@ -0,0 +1,96 @@
...
+                                 CompactionJobInfo* compaction_job_info) = 0;
+                                 
+  // Install files that were generated by a pluggable compaction request into
+  // the local database.
+  virtual Status DownloadFile(const FileOptions& options,
+                              const std::string& remote_path,
+                              const std::string& local_path) = 0;
+  virtual std::vector<Status> DownloadFiles(const FileOptions& options,
+                                            const std::vector<std::string>& remote_paths,
+                                            const std::vector<std::string>& local_paths);
```
jay-zhuang, line 93: Should we combine these to one function?

include/rocksdb/compaction_service.h
mrambacher: The CompactionServiceOptions need to be non-binary for anyone who is implementing one of these services.  They will need to know of all of these fields.  

There should be a means of converting the CompactionServiceOptions to/from "binary/string/encoded" format so that there can be some guarantee that a remote side supports the same fields/versions/settings as a local side and to  create an easier/standard means of producing those settings.  But when they get to the other service, it will still need to know all of these fields.

include/rocksdb/compaction_service.h
mrambacher: If the Options are part of the CompactionServiceOptions, that forces another copy of the options object (this is a reference and not a copy).  By passing the options here, the implementer of the service is free to grab which options they need from the options and not store the entire object.

Additionally, there are quite a few values in Options that are required by Compaction, and it might be hard to separate them out cleanly in a way that would not be subject to change between releases.

mrambacher: Note that if it was possible to get a set of MutableOptions, this option could simply be those fields, as the other options would be passed in at CompactionService configuration time (Validate/PrepareOptions during DB startup).

Using the Options infrastructure, it would be possible to pass in a string representing the MutableOptions values, but not their structure.  Is that preferred?

include/rocksdb/compaction_service.h
```diff
@@ -0,0 +1,101 @@
...
+  // Cancels a compaction job that is currently running.
+  // This method can be used to abort a manual compaction or to
+  // indicate that the job should be terminated when the database is being
+  // shutdown.
+  virtual Status Cancel(const std::string& job_id) = 0;
+
+  // Waits for the specified compaction job to complete.  Returns the results
+  // of this compaction.
+  virtual Status WaitForComplete(const std::string& job_id,
+                                 CompactionJobInfo* compaction_job_info) = 0;
```
jay-zhuang, line 93: We might need more (or different information) for remote compaction result. But I think we can add that later.

include/rocksdb/compaction_service.h
```diff
@@ -0,0 +1,101 @@
...
+  ColumnFamilyHandle *column_family;
+
+  // The DBOptions for this compaction.
+  DBOptions db_options;
+
+  // List of existing snapshots in the db
+  std::vector<SequenceNumber> existing_snapshots;
+
+  // List of input files of the compaction
+  std::vector<std::string> input_files;
```
jay-zhuang, line 50: We might still need a blob (a binary format) parameter for compaction details. For example, manifest file needed for this compaction, for the first version, we plan to load manifest file from the remote disk. I think we could expand that later.

jay-zhuang (approved):
I think it's a good start, we'll expand and change the interface as needed during development.

ajkr (changes requested):
I don't think we should take it. Copying my old pro-binary format argument here.

The low-level details necessary to run these compaction are specific to the `DBImpl` implementation of the `DB` interface. Other `DB` subclasses will need different details in order to run a remote compaction, so we shouldn't put `DBImpl`-specific code in the interface.

My more worrisome concern though for a non-binary format is that the functionality is not actually customizable in the usual sense. The only supported compaction algorithm would be doing exactly what RocksDB would do in the version running on the leader. We typically don't expose public APIs which constrain the user to only do exactly what RocksDB would have done. Doing so would certainly require their implementation be based on internal details, and there'd be no way at all to provide any sort of upgrade/downgrade compatibility.

jay-zhuang:
> I don't think we should take it. Copying my old pro-binary format argument here.
> 
> The low-level details necessary to run these compaction are specific to the `DBImpl` implementation of the `DB` interface. Other `DB` subclasses will need different details in order to run a remote compaction, so we shouldn't put `DBImpl`-specific code in the interface.
> 
> My more worrisome concern though for a non-binary format is that the functionality is not actually customizable in the usual sense. The only supported compaction algorithm would be doing exactly what RocksDB would do in the version running on the leader. We typically don't expose public APIs which constrain the user to only do exactly what RocksDB would have done. Doing so would certainly require their implementation be based on internal details, and there'd be no way at all to provide any sort of upgrade/downgrade compatibility.

I agree. One example is passing MANIFEST information to the remote compaction (or the information of the LSM tree for the compaction), which should not be exposed directly to the user.

mrambacher:
> I don't think we should take it. Copying my old pro-binary format argument here.
> 
> The low-level details necessary to run these compaction are specific to the `DBImpl` implementation of the `DB` interface. Other `DB` subclasses will need different details in order to run a remote compaction, so we shouldn't put `DBImpl`-specific code in the interface.
> 
> My more worrisome concern though for a non-binary format is that the functionality is not actually customizable in the usual sense. The only supported compaction algorithm would be doing exactly what RocksDB would do in the version running on the leader. We typically don't expose public APIs which constrain the user to only do exactly what RocksDB would have done. Doing so would certainly require their implementation be based on internal details, and there'd be no way at all to provide any sort of upgrade/downgrade compatibility.

I do not agree with the binary argument and in fact would argue that the RocksDB Cloud shows the opposite.  Their RemoteCompactionService implementation is based on structures very much like the ones suggested here.  

If you look at the CompactionIterator, I think it needs to know very little about the DB implementation.  The only reasons I see for requiring the "Versions" is to optimize "deleted" keys out of the SST file and to find the Blob file.  I would argue we could live without the first one and come up with an alternative for the second.  

I would however argue that it would be useful for having a way of serializing the "Compaction" information in order to send this to the other side.  For example, it would be nice to be able to serialize the state of compaction, and send it to another process that can validate it and invoke the proper compaction.  The point of this serialization is that we can do more validation of the serialized information and make sure that the remote side is set up to handle the compaction properly (the same Mutable parameters for example).

mrambacher:
Again, this API is intended for the users who wish to implement their own compaction service.  It is not meant for those who wish to use a remote compaction service.  To use a remote compaction service, one will set the property in the DBOptions to use the service.  

With a little bit of surgery, I could write a CompactionService that did not use a DB on the other side but only used the CompactionIterator (in fact, this will be a good test of implementing a CompactionService).  The output files may not be identical but should be functionality equivalent.  I think I can implement this without reading the MANIFEST or using other DB functions on the "remote side".  Compaction is taking this set of input files and emitting this set of output files and should be doable without a DBImpl.

Or am I missing something?

pdillinger:
> Again, this API is intended for the users who wish to implement their own compaction service. It is not meant for those who wish to use a remote compaction service.

What's the rush to make this public, then? I suggest that our experiences with FilterPolicy, Cache, and others suggest that when we race to make something customizable before we really understand it we almost always end up wrestling with bad abstractions.

> The only reasons I see for requiring the "Versions" is to optimize "deleted" keys out of the SST file

Are you suggesting we have tombstones stick around forever with remote compaction?

riversand963:
Current `CompactionIterator` relies on a few important internal data structures, e.g. `compaction`, `snapshot_checker` to 1) save space by zeroing out sequence numbers and drop tombstones, 2) function correctly for transactions.

mrambacher:
Abandoning this change as there is an alternate circulating now.

PR #7963, opened 2021-02-15
Fix txn `MultiGet()` return un-committed data with snapshot

jay-zhuang:
TransactionDB uses read callback to filter out un-committed data before
a snapshot. But `MultiGet()` API doesn't use that at all, which causes
returning unwanted data.

Test Plan: Added unittest to reproduce


anand1976 (approved):
Thanks a lot for investigating and getting to the bottom of this!! Just a reminder, ```HISTORY.md``` needs to be updated.

ajkr:
Nice finding, fix, and test - thanks!

PR #7967, opened 2021-02-16
c:h export rocksdb_transactiondb_open_column_families

morningsend:
Hi, I noticed a bug in rocksdb C API, where a function is not exported and created a fix.


ajkr (approved):
LGTM

PR #7968, opened 2021-02-16
Make ChRootEnv, EncryptedEnv, and TimedEnv into FileSystems

mrambacher:
These classes were wraps of Env that provided only extensions to the FileSystem functionality.  Changed the classes to be FileSystems and the wraps to be of the CompositeEnvWrapper.


anand1976 (approved):
LGTM

PR #7970, opened 2021-02-16
Limit buffering for collecting samples for compression dictionary

ajkr:
For dictionary compression, we need to collect some representative samples of the data to be compressed, which we use to either generate or train (when `CompressionOptions::zstd_max_train_bytes > 0`) a dictionary. Previously, the strategy was to buffer all the data blocks during flush, and up to the target file size during compaction. That strategy allowed us to randomly pick samples from as wide a range as possible that'd be guaranteed to land in a single output file.

However, some users try to make huge files in memory-constrained environments, where this strategy can cause OOM. This PR introduces an option, `CompressionOptions::max_dict_buffer_bytes`, that limits how much data blocks are buffered before we switch to unbuffered mode (which means creating the per-SST dictionary, writing out the buffered data, and compressing/writing new blocks as soon as they are built). It is not strict as we currently buffer more than just data blocks -- also keys are buffered. But it does make a step towards giving users predictable memory usage.

Related changes include:

- Changed sampling for dictionary compression to select unique data blocks when there is limited availability of data blocks
- Made use of `BlockBuilder::SwapAndReset()` to save an allocation+memcpy when buffering data blocks for building a dictionary
- Changed `ParseBoolean()` to accept an input containing characters after the boolean. This is necessary since, with this PR, a value for `CompressionOptions::enabled` is no longer necessarily the final component in the `CompressionOptions` string.

Test Plan:

- updated `CompressionOptions` unit tests to verify limit is respected (to the extent expected in the current implementation) in various scenarios of flush/compaction to bottommost/non-bottommost level
- looked at jemalloc heap profiles right before and after switching to unbuffered mode during flush/compaction. Verified memory usage in buffering is proportional to the limit set.

### DIFF
--- a/options/options_settable_test.cc
+++ b/options/options_settable_test.cc
@@ -411,10 +411,11 @@ TEST_F(OptionsSettableTest, ColumnFamilyOptionsAllFieldsSettable) {
   FillWithSpecialChar(options_ptr, sizeof(ColumnFamilyOptions),
                       kColumnFamilyOptionsExcluded);
 
-  // It based on the behavior of compiler that padding bytes are not changed
-  // when copying the struct. It's prone to failure when compiler behavior
-  // changes. We verify there is unset bytes to detect the case.
-  *options = ColumnFamilyOptions();
+  // Invoke a user-defined constructor in the hope that it does not overwrite
+  // padding bytes. Note that previously we relied on the implicitly-defined
+  // copy-assignment operator (i.e., `*options = ColumnFamilyOptions();`) here,
+  // which did in fact modify padding bytes.
+  options = new (options_ptr) ColumnFamilyOptions();
 
   // Deprecatd option which is not initialized. Need to set it to avoid
   // Valgrind error
@@ -472,8 +473,8 @@ TEST_F(OptionsSettableTest, ColumnFamilyOptionsAllFieldsSettable) {
       "max_bytes_for_level_multiplier=60;"
       "memtable_factory=SkipListFactory;"
       "compression=kNoCompression;"
-      "compression_opts=5:6:7:8:9:true;"
-      "bottommost_compression_opts=4:5:6:7:8:true;"
+      "compression_opts=5:6:7:8:9:10:true:11;"
+      "bottommost_compression_opts=4:5:6:7:8:9:true:10;"
       "bottommost_compression=kDisableCompressionOption;"
       "level0_stop_writes_trigger=33;"
       "num_levels=99;"
--- a/options/options_settable_test.cc
+++ b/options/options_settable_test.cc
@@ -407,7 +407,6 @@ TEST_F(OptionsSettableTest, ColumnFamilyOptionsAllFieldsSettable) {
   // Count padding bytes by setting all bytes in the memory to a special char,
   // copy a well constructed struct to this memory and see how many special
   // bytes left.
-  ColumnFamilyOptions* options = new (options_ptr) ColumnFamilyOptions();
   FillWithSpecialChar(options_ptr, sizeof(ColumnFamilyOptions),
                       kColumnFamilyOptionsExcluded);
 
@@ -415,7 +414,7 @@ TEST_F(OptionsSettableTest, ColumnFamilyOptionsAllFieldsSettable) {
   // padding bytes. Note that previously we relied on the implicitly-defined
   // copy-assignment operator (i.e., `*options = ColumnFamilyOptions();`) here,
   // which did in fact modify padding bytes.
-  options = new (options_ptr) ColumnFamilyOptions();
+  ColumnFamilyOptions* options = new (options_ptr) ColumnFamilyOptions();
 
   // Deprecatd option which is not initialized. Need to set it to avoid
   // Valgrind error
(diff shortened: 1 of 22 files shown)


options/options_settable_test.cc
```diff
@@ -411,10 +411,11 @@ TEST_F(OptionsSettableTest, ColumnFamilyOptionsAllFieldsSettable) {
   FillWithSpecialChar(options_ptr, sizeof(ColumnFamilyOptions),
                       kColumnFamilyOptionsExcluded);
 
-  // It based on the behavior of compiler that padding bytes are not changed
-  // when copying the struct. It's prone to failure when compiler behavior
-  // changes. We verify there is unset bytes to detect the case.
-  *options = ColumnFamilyOptions();
+  // Invoke a user-defined constructor in the hope that it does not overwrite
```
pdillinger, line 414: Above `new (options_ptr)` now obsolete?

tools/db_crashtest.py
```diff
@@ -51,6 +51,7 @@
     # Disabled compression_parallel_threads as the feature is not stable
     # lambda: random.choice([1] * 9 + [4])
     "compression_parallel_threads": 1,
+    "compression_max_dict_buffer_bytes": lambda: 4096 * random.randint(0, 32),
```
pdillinger, line 54: There should be no reason to use only "friendly" sizes in the stress test, right?

util/string_util.cc
```diff
@@ -279,9 +279,16 @@ bool StartsWith(const std::string& string, const std::string& pattern) {
 #ifndef ROCKSDB_LITE
 
 bool ParseBoolean(const std::string& type, const std::string& value) {
-  if (value == "true" || value == "1") {
+  const static std::string kTrue = "true", kFalse = "false";
```
pdillinger, line 282: Why?

table/block_based/block_based_table_builder.cc
```diff
@@ -1630,14 +1644,37 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
...
+  std::vector<uint16_t> data_block_order;
+  size_t data_block_order_idx = 0;
+  if (r->data_block_and_keys_buffers.size() <= ((1 << 16) - 1)) {
+    data_block_order.resize(r->data_block_and_keys_buffers.size());
+    std::iota(data_block_order.begin(), data_block_order.end(),
+              static_cast<uint16_t>(0));
+    // We could be smarter and interleave the shuffling and sample appending
+    // logic. Then we could terminate as soon as `kSampleBytes` is reached,
+    // saving some shuffling computation.
+    RandomShuffle(data_block_order.begin(), data_block_order.end());
```
pdillinger, line 1660: I like how the existing code uses well-seeded but reproducible randomness. We should try to keep that here.

table/block_based/block_based_table_builder.cc
```diff
@@ -1630,14 +1644,37 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
   const size_t kSampleBytes = r->compression_opts.zstd_max_train_bytes > 0
                                   ? r->compression_opts.zstd_max_train_bytes
                                   : r->compression_opts.max_dict_bytes;
+
+  // If buffer size is reasonable, we pre-generate a permutation to enforce
+  // uniqueness. This prevents wasting samples on duplicates, which is
+  // particularly likely when not many blocks were buffered.
```
pdillinger, line 1650: Another option to consider vs. random order is some linear ordering that guarantees uniqueness (or almost).

Consider this (classic) option: pick a random starting position, and increment by a large fixed prime, mod max_val. The downside of this is that some max_vals will guarantee poor spread, e.g. if the prime is === 1 (mod number of blocks).

Another option I like (almost uniqueness; less threat of "bad" randomness): treat a w bit number (32 or 64 bits) as a fraction of 0 to max_val (see "FastRange" in our code). Pick a random starting position (a random w-bit number) and each time increment by the w-bit golden ratio. This maximizes the spatial spread of samples[0], which nearly guarantees uniqueness, but also the increased entropy of tracking w bits means there are no "misbehaving" choices of max_val and essentially no chance of bad correlations with underlying data.

[0] You can get slightly more even spread if you use the number of samples in your choice, but that is to the detriment of other properties.

pdillinger (approved):
LGTM, modulo some minor concerns and suggestions

util/string_util.cc
ajkr: There was a confusing usage in the client code (`ParseCompressionOptions()`) where it passed all remaining fields in `value`. It only worked before this PR because the boolean (`CompressionOptions::enabled`) was the final option to parse, so the remaining fields would consist of only that field. This PR exposed the issue by adding an option (`CompressionOptions::max_dict_buffer_bytes`) that should be parsed after the boolean.

Anyways, I fixed the confusing usage in `ParseCompressionOptions()` now (see reply above), and reverted this change.

tools/db_crashtest.py
ajkr: Right. Not sure of any particularly unfriendly value but we can make the range wider. I changed it to one less than powers of two from 1 -> 40.

ajkr (commented):
Thanks very much for the review! I will address the couple remaining comments soon.

table/block_based/block_based_table_builder.cc
ajkr: Done.

table/block_based/block_based_table_builder.cc
ajkr: This is great, the extra memory needed for storing the vector (or deciding not to store it in certain cases) bothered me a lot. I will probably try the first classic suggestion since it's easier for me to understand and the test cases assume samples are unique. But it'll probably need to be deferred for this to make it into the release...

PR #7971, opened 2021-02-16
malloc_usable_size api support for MacOS.

devnexen:
The platform has a similar api, thus creating a simple alias
 for simplicity.


util/compression.h
```diff
@@ -14,7 +14,10 @@
 #ifdef ROCKSDB_MALLOC_USABLE_SIZE
 #ifdef OS_FREEBSD
 #include <malloc_np.h>
-#else  // OS_FREEBSD
```
ajkr, line 17: If it's not necessary for the original code to do this, can we just `#include "port/malloc.h"` instead?

ajkr (approved):
Nice! I checked the man pages -- the "usable" aspect of `malloc_usable_size()` is not really relevant; we aren't supposed to use memory outside our allocation anyways. We just use the API to inspect internal fragmentation, which `malloc_size()` can also do. So the approach is good to me.

ajkr (commented):
> Nice! I checked the man pages -- the "usable" aspect of `malloc_usable_size()` is not really relevant; we aren't supposed to use memory outside our allocation anyways. We just use the API to inspect internal fragmentation, which `malloc_size()` can also do. So the approach is good to me.

Actually I was wrong, there are places we use the internal fragmentation: https://github.com/facebook/rocksdb/blob/5b2bbacb6fd2b49606ef2cd24e3bb2d2061239b4/include/rocksdb/table.h#L230-L234. Hm... At least we should update comments like this one since now the user would be trusting `malloc_size()`.

mrambacher (commented):
Doesn't the build_detect_platform code also need to change?  Right now it looks for the "malloc_usable_size".

The check here might be simplest to have it:
- Do the test compile with -DROCKSDB_MALLOC_USABLE_SIZE defined
- Include "port/malloc.h" (instead of malloc)

ajkr (changes requested):
> > Nice! I checked the man pages -- the "usable" aspect of `malloc_usable_size()` is not really relevant; we aren't supposed to use memory outside our allocation anyways. We just use the API to inspect internal fragmentation, which `malloc_size()` can also do. So the approach is good to me.
> 
> Actually I was wrong, there are places we use the internal fragmentation:
> 
> https://github.com/facebook/rocksdb/blob/5b2bbacb6fd2b49606ef2cd24e3bb2d2061239b4/include/rocksdb/table.h#L230-L234
> 
> . Hm... At least we should update comments like this one since now the user would be trusting `malloc_size()`.

I ended up not sure about this. The `malloc_usable_size()` manpage says: "Although the excess bytes can be overwritten by the application without ill effects, ...". But the `malloc_size()` manpage does not say that. `malloc_size()` does not seem to have any documentation indicating its use in `optimize_filters_for_memory=true` is safe.

However I do still wish we can use `malloc_size()` for the purpose of memory tracking on Mac. What about if we use `malloc_good_size()` to round-up allocation sizes and then `optimize_filters_for_memory=true` can use those extra bytes? Or is it too much work to avoid a minor risk? @pdillinger 

edit: This may be very difficult since `malloc_good_size()` appears only available on Mac...

pdillinger:
The only risk is during filter construction and only with a particular option, so I consider it low risk. I recommend just updating my comment as

    ... if you do not trust malloc_usable_size (malloc_size on MacOS).

Filters are always readable without going beyond requested allocation size.

ajkr:
We should also add an entry to HISTORY.md behavior change to mention on Mac the effect of taking internal fragmentation into account -- memory usage should be lower (closer to configured limit), flush can happen earlier, less can be cached. I can do this after it lands if @devnexen 's not available.

PR #7977, opened 2021-02-18
Introduce a new trace file format (v 0.2) for better extension

zhichao-cao:
The trace file record and payload encode is fixed, which requires complex backward compatibility resolving. This PR introduce a new trace file format, which makes it easier to add new entries to the payload and does not have backward compatible issues. V 0.1 is still supported in this PR. Added the tracing for lower_bound and upper_bound for iterator.

test plan: make check. tested with old trace file in replay and analyzing.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,7 +2,7 @@
 ## Unreleased
 ### Behavior Changes
 * When retryable IO error occurs during compaction, it is mapped to soft error and set the BG error. However, auto resume is not called to clean the soft error since compaction will reschedule by itself. In this change, When retryable IO error occurs during compaction, BG error is not set. User will be informed the error via EventHelper.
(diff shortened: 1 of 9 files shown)


HISTORY.md
```diff
@@ -2,6 +2,7 @@
 ## Unreleased
 ### Behavior Changes
 * When retryable IO error occurs during compaction, it is mapped to soft error and set the BG error. However, auto resume is not called to clean the soft error since compaction will reschedule by itself. In this change, When retryable IO error occurs during compaction, BG error is not set. User will be informed the error via EventHelper.
+* Introduce a new trace file format for query tracing and replay and trace file version is bump up to 0.2. A payload map is added as the first portion of the payload. We will not have backward compatible issues when adding new entries to trace records.
```
anand1976, line 5: Also mention iterator upper/lower bound is added to the trace

anand1976 (approved):
LGTM. This should make it easier to trace more information in the future.

HISTORY.md
zhichao-cao: Sure

PR #7979, opened 2021-02-18
Extend VerifyFileChecksums API for  blob files

akankshamahajan15:
Summary: Extend VerifyFileChecksums API to verify blob files in case of
use_file_checksum.

Test Plan: New unit test db_blob_corruption_test

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -591,8 +591,8 @@ ifdef ASSERT_STATUS_CHECKED
 		db_iter_stress_test \
 		db_log_iter_test \
 		db_bloom_filter_test \
-		db_blob_corruption_test \
 		db_blob_basic_test \
+		db_blob_corruption_test \
 		db_blob_index_test \
 		db_block_cache_test \
 		db_compaction_test \
--- a/Makefile
+++ b/Makefile
@@ -591,6 +591,7 @@ ifdef ASSERT_STATUS_CHECKED
 		db_iter_stress_test \
 		db_log_iter_test \
 		db_bloom_filter_test \
+		db_blob_corruption_test \
 		db_blob_basic_test \
 		db_blob_index_test \
 		db_block_cache_test \
@@ -2059,6 +2060,8 @@ io_tracer_parser_test: $(OBJ_DIR)/tools/io_tracer_parser_test.o $(OBJ_DIR)/tools
 io_tracer_parser: $(OBJ_DIR)/tools/io_tracer_parser.o $(TOOLS_LIBRARY) $(LIBRARY)
 	$(AM_LINK)
 
+db_blob_corruption_test: $(OBJ_DIR)/db/blob/db_blob_corruption_test.o $(TEST_LIBRARY) $(LIBRARY)
+	$(AM_LINK)
 #-------------------------------------------------
 # make install related stuff
 PREFIX ?= /usr/local
--- a/db/blob/db_blob_corruption_test.cc
+++ b/db/blob/db_blob_corruption_test.cc
@@ -6,7 +6,6 @@
 #include "db/db_test_util.h"
 #include "port/stack_trace.h"
 #include "test_util/sync_point.h"
-#include "utilities/fault_injection_env.h"
 
 namespace ROCKSDB_NAMESPACE {
 
--- a/db/blob/db_blob_corruption_test.cc
+++ b/db/blob/db_blob_corruption_test.cc
@@ -35,6 +35,7 @@ class DBBlobCorruptionTest : public DBTestBase {
   }
 };
 
+#ifndef ROCKSDB_LITE
 TEST_F(DBBlobCorruptionTest, VerifyWholeTableChecksum) {
   Options options = GetDefaultOptions();
   options.enable_blob_files = true;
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -591,6 +591,7 @@ ifdef ASSERT_STATUS_CHECKED
 		db_iter_stress_test \
 		db_log_iter_test \
 		db_bloom_filter_test \
+		db_blob_corruption_test \
```
ltamasi, line 594: Nit: could you put this in between `db_blob_basic_test` and `db_blob_index_test` to keep the alphabetical order?

db/blob/db_blob_corruption_test.cc
```diff
@@ -0,0 +1,78 @@
...
+      : DBTestBase("/db_blob_corruption_test", /* env_do_fsync */ false) {}
+
+  void Corrupt(FileType filetype, int offset, int bytes_to_corrupt) {
+    // Pick file to corrupt
+    std::vector<std::string> filenames;
+    ASSERT_OK(env_->GetChildren(dbname_, &filenames));
+    uint64_t number;
+    FileType type;
+    std::string fname;
+    int picked_number = -1;
```
ltamasi, line 24: This could be `uint64_t picked_number = 0;` or `uint64_t picked_number = kInvalidBlobFileNumber;` (which is also zero; defined in `db/blob/blob_constants.h`). This would eliminate the need to `static_cast` below.

db/blob/db_blob_corruption_test.cc
```diff
@@ -0,0 +1,78 @@
...
+        picked_number = static_cast<int>(number);
+      }
+    }
+    ASSERT_TRUE(!fname.empty()) << filetype;
+    ASSERT_OK(test::CorruptFile(env_, fname, offset, bytes_to_corrupt));
+  }
+};
+
+#ifndef ROCKSDB_LITE
+TEST_F(DBBlobCorruptionTest, VerifyWholeTableChecksum) {
```
ltamasi, line 38: How about `VerifyWholeBlobFileChecksum` ?

db/blob/db_blob_corruption_test.cc
```diff
@@ -0,0 +1,78 @@
...
+  Close();
+
+  Corrupt(kBlobFile, 0, 2);
+
+  ASSERT_OK(TryReopen(options));
+
+  int count{0};
+  SyncPoint::GetInstance()->SetCallBack(
+      "DBImpl::VerifySstFileChecksum:mismatch", [&](void* arg) {
+        auto* s = reinterpret_cast<Status*>(arg);
```
ltamasi, line 61: Minor: `static_cast` would suffice here. Also, it could be `const Status* s` since we're not modifying the value.

db/blob/db_blob_corruption_test.cc
```diff
@@ -0,0 +1,78 @@
...
+        ASSERT_NE(s, nullptr);
+        ++count;
+        ASSERT_NOK(*s);
+      });
+  SyncPoint::GetInstance()->EnableProcessing();
+
+  ASSERT_TRUE(db_->VerifyFileChecksums(ReadOptions()).IsCorruption());
+  ASSERT_EQ(1, count);
+}
```
ltamasi, line 70: Can we call `DisableProcessing` and `ClearAllCallBacks` for the `SyncPoint`s here just in case?

db/db_impl/db_impl.cc
```diff
@@ -4890,26 +4910,28 @@ Status DBImpl::VerifyChecksumInternal(const ReadOptions& read_options,
   return s;
 }
 
-Status DBImpl::VerifySstFileChecksum(const FileMetaData& fmeta,
+Status DBImpl::VerifySstFileChecksum(const std::string& file_checksum_expected,
```
ltamasi, line 4913: How about renaming this to `VerifyFullFileChecksum` to make it clear it's not solely for SST files?

src.mk
```diff
@@ -377,6 +377,7 @@ TEST_MAIN_SOURCES =                                                     \
   db/blob/blob_file_garbage_test.cc                                     \
   db/blob/blob_file_reader_test.cc                                      \
   db/blob/db_blob_basic_test.cc                                         \
+  db/blob/db_blob_corruption_test.cc                                    \
```
ltamasi, line 380: Could you please add it to `CMakeLists.txt` as well?

db/db_impl/db_impl.cc
```diff
@@ -4890,26 +4910,28 @@ Status DBImpl::VerifyChecksumInternal(const ReadOptions& read_options,
...
+                                     const std::string& func_name_expected,
                                      const std::string& fname,
                                      const ReadOptions& read_options) {
   Status s;
-  if (fmeta.file_checksum == kUnknownFileChecksum) {
+  if (file_checksum_expected == kUnknownFileChecksum) {
```
ltamasi, line 4918: I think we should also bail out here if `file_checksum_expected` is empty.

ltamasi (commented):
Thanks for the PR @akankshamahajan15 ! LGTM in general, just some minor comments (please see below).

db/db_impl/db_impl.cc
akankshamahajan15: ```kUnknownFileChecksum``` is intialized to empty string.

ltamasi (approved):

PR #7981, opened 2021-02-19
Java-API: byteCompressionType should be declared as primitive type byte

stefan-zobel:
The variable `byteCompressionType` is only assigned values of primitive type and is never 'null', but it is declared with the boxed type 'Byte'.


adamretter (approved):
LGTM :-)

adamretter:
@mrambacher If you could do the honours please?

PR #7987, opened 2021-02-22
Pick samples for compression dictionary using prime number

ajkr:
The sample selection technique taken in #7970 was problematic
because it had two code paths for sample selection depending on the
number of data blocks, and one of those code paths involved an
allocation. Using prime numbers, we can consolidate into one code path
without allocation. The downside is there will be values of N (number of
data blocks buffered) that suffer from poor spread in the selected
samples.

Test Plan: `make check -j48`


table/block_based/block_based_table_builder.cc
```diff
@@ -1644,44 +1644,40 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
...
+
+  // Abstract algebra teaches us that a finite cyclic group (such as the
+  // additive group of integers modulo N) can be generated by a number that is
+  // coprime with N. Since N is variable (number of buffered data blocks), we
+  // must then pick a prime number in order to guarantee coprimeness with any N.
+  // Picked a Mersenne prime because it's easy on the eyes. No idea about its
+  // properties.
+  //
+  // One downside of this approach is the spread will be poor when
+  // `kPrimeGeneratorRemainder` is small enough that we pick all samples before
```
ajkr, line 1657: Hm, this isn't really right. If `kPrimeGeneratorRemainder == kNumBlocksBuffered - 1` (i.e., if it is large enough), we have the same problem. I will fix this comment...

table/block_based/block_based_table_builder.cc
```diff
@@ -1644,44 +1644,40 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
...
+  // cycling through the whole array.
+  const uint64_t kPrimeGenerator = (1ull << 61) - 1;
```
ajkr, line 1659: Is this the worst choice of prime for some reason? For almost all `kNumBlocksBuffered` the test cases use (e.g., 6, 9, 31, 32, 33) I am seeing `kPrimeGeneratorRemainder == 1` or `kPrimeGeneratorRemainder == kNumBlocksBuffered - 1`. I will change this to something unrelated to powers of two...

pdillinger (approved):
LGTM

If you were feeling ambitious, there's another version of the golden ratio generator that can ensure no repeats in the sequence before all have been covered: find smallest power of two >= kNumBlocksBuffered. Multiply (inverse) golden ratio by this power of two, rounding to nearest (or easiest) odd number. Use this as your increment (mod the power of two), but skip over values where current (buffer_idx) >= kNumBlocksBuffered.

ajkr:
> LGTM
> 
> If you were feeling ambitious, there's another version of the golden ratio generator that can ensure no repeats in the sequence before all have been covered: find smallest power of two >= kNumBlocksBuffered. Multiply (inverse) golden ratio by this power of two, rounding to nearest (or easiest) odd number. Use this as your increment (mod the power of two), but skip over values where current (buffer_idx) >= kNumBlocksBuffered.

Thanks for the review! I hope to learn the math background for this and then try it some day. But probably in the near future there are too many other things needing attention, so I'll land this for now.

pdillinger:
https://www.youtube.com/watch?v=sj8Sg8qnjOg

PR #7991, opened 2021-02-23
Append all characters not captured by xsputn() in overflow() function

sherriiiliu:
In the adapter class `WritableFileStringStreamAdapter`, which wraps WritableFile to be used for std::ostream, previouly only `std::endl` is considered a special case because `endl` is written by `os.put()` directly without going through `xsputn()`. `os.put()` will call `sputc()` and if we further check the internal implementation of `sputc()`, we will see it is
```
int_type __CLR_OR_THIS_CALL sputc(_Elem _Ch) {  // put a character
    return 0 < _Pnavail() ? _Traits::to_int_type(*_Pninc() = _Ch) : overflow(_Traits::to_int_type(_Ch));
```
As we explicitly disabled buffering, _Pnavail() is always 0. Thus every write, not captured by xsputn, becomes an overflow.

When I run tests on Windows, I found not only `std::endl` will drop into this case, writing an unsigned long long will also call `os.put()` then followed by `sputc()` and eventually call `overflow()`. Therefore, instead of only checking `std::endl`, we should try to append other characters as well unless the appending operation fails.


HISTORY.md
```diff
@@ -1,5 +1,7 @@
 # Rocksdb Change Log
 ## Unreleased
+### Bug Fixes
+* In WritableFileStringStreamAdapter::overflow(), not only we need to handle `std::endl`, but also there are other cases like unsigned long long that calls `os.put()` directly then drops into sputc() function. The internal implementation of `sputc()` calls `overflow()`. As buffering is explicitly disabled in this stream wrapper, every write, not captured by xsputn, becomes an overflow here. Now it is fixed so all characters that are not captured by xsputn will be appended here, unless the file appending fails.
```
ajkr, line 4: Can we write this more in terms of user-visible APIs and what is the user impact? It appears `WritableFileStringStreamAdapter` is used only in `BlockBasedTable::DumpTable()`. Then, I believe "dumping block-based SST files in a human-readable format" would be affected. Regarding user impact, it seems the table dump would be truncated?

ajkr (changes requested):
I think it is a good fix. We should just try to make the release note friendlier to users.

HISTORY.md
sherriiiliu: Yes correct, from user side I observed the table dump was truncated in some cases. I have updated the unreleased note accordingly. Thanks for suggestion!

ajkr (approved):
Looks great, thanks!

PR #7993, opened 2021-02-23
Create DB properties for internal table cache

matthewvon:
This PR duplicates the existing two properties for BlockCacheCapacity() and BlockCacheUsage() by adding TableCacheCapacity() and TableCacheUsage().

There is currently no way to know how many .sst files are actually open within rocksdb.  The "usage" of the table cache gives that information.  It is important to know how many files are open to estimate the amount of table reader space per open file.  That estimate is then useful for memory management, including the reduction / expansion of the table_cache via SetDBOptions("max_open_files").

### DIFF
--- a/db/db_properties_test.cc
+++ b/db/db_properties_test.cc
@@ -245,7 +245,8 @@ void GetExpectedTableProperties(
   const int kDeletionCount = kTableCount * kDeletionsPerTable;
   const int kMergeCount = kTableCount * kMergeOperandsPerTable;
   const int kRangeDeletionCount = kTableCount * kRangeDeletionsPerTable;
-  const int kKeyCount = kPutCount + kDeletionCount + kMergeCount + kRangeDeletionCount;
+  const int kKeyCount =
+      kPutCount + kDeletionCount + kMergeCount + kRangeDeletionCount;
   const int kAvgSuccessorSize = kKeySize / 5;
   const int kEncodingSavePerKey = kKeySize / 4;
   expected_tp->raw_key_size = kKeyCount * (kKeySize + 8);
@@ -256,7 +257,8 @@ void GetExpectedTableProperties(
   expected_tp->num_merge_operands = kMergeCount;
   expected_tp->num_range_deletions = kRangeDeletionCount;
   expected_tp->num_data_blocks =
-      kTableCount * (kKeysPerTable * (kKeySize - kEncodingSavePerKey + kValueSize)) /
+      kTableCount *
+      (kKeysPerTable * (kKeySize - kEncodingSavePerKey + kValueSize)) /
       kBlockSize;
   expected_tp->data_size =
       kTableCount * (kKeysPerTable * (kKeySize + 8 + kValueSize));
@@ -1090,7 +1092,8 @@ class CountingUserTblPropCollector : public TablePropertiesCollector {
     std::string encoded;
     PutVarint32(&encoded, count_);
     *properties = UserCollectedProperties{
-        {"CountingUserTblPropCollector", message_}, {"Count", encoded},
+        {"CountingUserTblPropCollector", message_},
+        {"Count", encoded},
     };
     return Status::OK();
   }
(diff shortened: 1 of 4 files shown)


db/db_properties_test.cc
```diff
@@ -1713,6 +1716,44 @@ TEST_F(DBPropertiesTest, BlockCacheProperties) {
...
+  //  get default max_open_files
+  //
+  ASSERT_TRUE(db_->GetIntProperty(DB::Properties::kTableCacheCapacity, &value));
+  new_value = value / 2;
+
+  std::unordered_map<std::string, std::string> new_options;
+  new_options.insert(std::pair<std::string, std::string>(
+      "max_open_files", std::to_string(new_value)));
+  rocksdb::Status stat = db_->SetDBOptions(new_options);
+  ASSERT_TRUE(stat.ok());
```
mrambacher: This should just be ASSERT_OK(db_->SetOptions(...)

db/db_properties_test.cc
```diff
@@ -1713,6 +1716,44 @@ TEST_F(DBPropertiesTest, BlockCacheProperties) {
...
+  ASSERT_TRUE(db_->GetIntProperty(DB::Properties::kTableCacheCapacity, &value));
+  // rocksdb does a -10 to number passed in
+  ASSERT_EQ(new_value - 10, value);
+
+  //
+  // create a new table to see if usage is "live"
```
mrambacher, line 1746: I do not see where this is creating a "new table".  Can you elaborate in the comment?

db/db_properties_test.cc
```diff
@@ -1713,6 +1716,44 @@ TEST_F(DBPropertiesTest, BlockCacheProperties) {
...
 }
 
+TEST_F(DBPropertiesTest, TableCacheProperties) {
+  Options options;
+  uint64_t value, new_value;
+
+  options.env = CurrentOptions().env;
+
+  // Block cache properties are not available if block cache is not used.
+  BlockBasedTableOptions table_options;
```
mrambacher, line 1726: There is already a default options, so you should not need to create the table_options here...

db/db_properties_test.cc
```diff
@@ -1713,6 +1716,44 @@ TEST_F(DBPropertiesTest, BlockCacheProperties) {
...
+  new_value = value / 2;
+
+  std::unordered_map<std::string, std::string> new_options;
+  new_options.insert(std::pair<std::string, std::string>(
+      "max_open_files", std::to_string(new_value)));
+  rocksdb::Status stat = db_->SetDBOptions(new_options);
+  ASSERT_TRUE(stat.ok());
+  ASSERT_TRUE(db_->GetIntProperty(DB::Properties::kTableCacheCapacity, &value));
+  // rocksdb does a -10 to number passed in
+  ASSERT_EQ(new_value - 10, value);
```
mrambacher, line 1743: I do not understand how the math works in this test.  You get the value, divide it by 2, subtract 10, and add 1.  By just reading the test this appears to be some sort of odd voodoo.

Can you add some comments somewhere to explain how/why this math works and what you are measuring?

db/db_properties_test.cc
matthewvon: Already done.

db/db_properties_test.cc
matthewvon: Will add comments.  In short, there are two different properties being tested "capacity" and "usage".

db/db_properties_test.cc
matthewvon: Previous tests make changes.  Thought this was a way to guarantee the traditional .sst tables.  Because I know how to force creation of a new .sst file with the traditional.  Not sure if same rules apply with others.

db/db_properties_test.cc
matthewvon: Done.

matthewvon:
@mrambacher One of the failing tests is upset that I am forcing a compaction without there already being on .sst file.  I will work that in the morning.

db/db_properties_test.cc
matthewvon: The redundant table factory creation is now gone.

matthewvon:
@mrambacher The one failing test should succeed.  Changed CompactRange() call to Flush().  Essentially the two calls accomplish the same goal of pushing data from write buffer to a new .sst file.  But somewhere there is code that asserts() if CompactRange() is called with no existing .sst files.  Do not remember which environment.

... and removed the redundant table factory creation in same test.  All should be good now.

mrambacher (approved):
LGTM

PR #7994, opened 2021-02-23
make:Fix c header prototypes

edwinzrodriguez:


examples/Makefile
```diff
@@ -12,6 +12,8 @@ ifneq ($(USE_RTTI), 1)
 	CXXFLAGS += -fno-rtti
 endif
 
+CFLAGS += -Wstrict-prototypes
+
```
mrambacher, line 16: Should a comparable flag also be added for CMAKE?

edwinzrodriguez: I've added this flag to Makefile as well.  This should prevent future changes that break strict prototypes

edwinzrodriguez: Also added to CMakeLists.txt and verified passing with cmake -Wdev

ajkr (approved):
LGTM, test failures appear unrelated

PR #7997, opened 2021-02-24
Merge Compare's condition

darionyaphet:


ajkr (changes requested):
I don't get it. If a.seq = 10, b.seq = 20, a.type = 2, b.type = 1, then:

- `a.sequence > b.sequence || a.type > b.type` is true
- `r = -1`
- This means `a` is ordered before `b`.  But it shouldn't be. The newer key `b` should come first.

ajkr:
This PR does expose a gap in our testing. We could catch this case by running multiple versions on the same DB (data written without this change would be broken when reading with this change, or vice-versa), or with a test case that manually crafts internal keys and verifies the expected order.

darionyaphet:
Got it. I ignored the order

Issue #7815, opened 2020-12-28
Secondary Instance will lost manifest data when calling TryCatchUpWithPrimary


CoderSong2015:
### Expected behavior
Secondary Instance could read all primary's manifest data after calling TryCatchUpWithPrimary()
### Actual behavior
Secondary Instance will lost manifest data when the MANIFEST file switch twice before TryCatchUpWithPrimary() called
### Steps to reproduce the behavior
Calling TryCatchUpWithPrimary after primary instance switch MANIFEST twice.

@riversand963

riversand963:
Thanks @CoderSong2015. This is a bug due to current (hacky) way to handle the primary's MANIFEST switch.
After the primary switches to the new MANIFEST, the first certain number of `VersionEdit`s in the file is a summary of current state, while `VersionEdit`s corresponding to newer updates follow. When it calls `TryCatchUp`, the secondary has already consumed the summary edits and must skip them. We currently do not have a good way of distinguishing between them, thus used the hack to skip a certain number of edits at the beginning of the new MANIFEST. This hack breaks when switching MANIFEST multiple (>1) times.
I have been thinking of writing a special version edits to denote the end of summary edits (see #7058), but this makes it impossible to roll back to older releases (<6.0). Feel free to contribute if you are interested and we can discuss an ideal way of fixing it.

riversand963:
I will submit a fix for this. cc @CoderSong2015

CoderSong2015:
@riversand963 Thanks very much. 
I have fixed it on my own project. My way is that  loading the full new MANIFEST file when MANIFEST file changed. I think it's not very frequently for rocksdb to change the manifest file so that the secondary instance could load full manifest every time and then it will not lost manifest data.

riversand963:
This is the approach I took in #7998. Should we close this issue?

PR #7998, opened 2021-02-24
Make secondary instance use ManifestTailer

riversand963:
This PR

- adds a class `ManifestTailer` that inherits from `VersionEditHandlerPointInTime`. `ManifestTailer::Iterate()` can be called multiple times to tail the primary instance's MANIFEST and apply the changes to the secondary,
- updates the implementation of `ReactiveVersionSet::ReadAndApply` to use this class,
- removes unused code in version_set.cc,
- updates existing tests, e.g. removing deleted sync points from unit tests,
- adds a new test to address the bug in #7815.

Test plan:
make check
Existing and newly-added tests in version_set_test.cc and db_secondary_test.cc

### DIFF
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -703,7 +703,7 @@ Status ManifestTailer::Initialize() {
 
     Version* dummy_version = default_cfd->dummy_versions();
     assert(dummy_version);
-    Version* base_version = dummy_version->TEST_Next();
+    Version* base_version = dummy_version->Next();
     assert(base_version);
     base_version->Ref();
     VersionBuilderUPtr new_builder(
@@ -747,7 +747,7 @@ Status ManifestTailer::OnColumnFamilyAdd(VersionEdit& edit,
 
   Version* dummy_version = tmp_cfd->dummy_versions();
   assert(dummy_version);
-  Version* base_version = dummy_version->TEST_Next();
+  Version* base_version = dummy_version->Next();
   assert(base_version);
   base_version->Ref();
   VersionBuilderUPtr new_builder(
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -62,8 +62,6 @@ void VersionEditHandlerBase::Iterate(log::Reader& reader,
     s = *log_read_status;
   }
 
-  read_buffer_.Clear();
-
   CheckIterationResult(reader, &s);
 
   if (!s.ok()) {
@@ -129,13 +127,13 @@ Status FileChecksumRetriever::ApplyVersionEdit(VersionEdit& edit,
 }
 
 VersionEditHandler::VersionEditHandler(
-    bool read_only, const std::vector<ColumnFamilyDescriptor>& column_families,
+    bool read_only, std::vector<ColumnFamilyDescriptor> column_families,
     VersionSet* version_set, bool track_missing_files,
     bool no_error_if_table_files_missing,
     const std::shared_ptr<IOTracer>& io_tracer, bool skip_load_table_files)
     : VersionEditHandlerBase(),
       read_only_(read_only),
-      column_families_(column_families),
+      column_families_(std::move(column_families)),
       version_set_(version_set),
       track_missing_files_(track_missing_files),
       no_error_if_table_files_missing_(no_error_if_table_files_missing),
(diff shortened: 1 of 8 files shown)


riversand963:
cc @jay-zhuang

db/version_set.cc
```diff
@@ -6323,15 +5797,22 @@ Status ReactiveVersionSet::MaybeSwitchManifest(
...
+      }
     }
   } while (s.IsPathNotFound());
   return s;
 }
 
+uint64_t ReactiveVersionSet::TEST_read_edits_in_atomic_group() const {
+  assert(manifest_tailer_);
+  return manifest_tailer_->GetReadBuffer().TEST_read_edits_in_atomic_group();
+}
```
jay-zhuang, line 5811: Should this be defined within:
```
#ifndef NDEBUG
...
#endif

File: db/version_edit_handler.cc [hunk 2]
```diff
@@ -668,6 +668,106 @@ Status VersionEditHandlerPointInTime::MaybeCreateVersion(
...
+  if (!tmp_cfd) {
+    // For now, ignore new column families created after Recover() succeeds.
+    return Status::OK();
+  }
+  auto builder_iter = builders_.find(edit.GetColumnFamily());
+  assert(builder_iter != builders_.end());
+
+  Version* dummy_version = tmp_cfd->dummy_versions();
+  assert(dummy_version);
+  Version* base_version = dummy_version->TEST_Next();
```
jay-zhuang, line 736: Should `TEST_Next()` only be used for debug code?

db/version_edit_handler.cc
```diff
@@ -668,6 +668,106 @@ Status VersionEditHandlerPointInTime::MaybeCreateVersion(
...
+  ColumnFamilySet* cfd_set = version_set_->GetColumnFamilySet();
+  assert(cfd_set);
+  ColumnFamilyData* tmp_cfd = cfd_set->GetColumnFamily(edit.GetColumnFamily());
+  assert(cfd);
+  *cfd = tmp_cfd;
+  if (!tmp_cfd) {
+    // For now, ignore new column families created after Recover() succeeds.
+    return Status::OK();
+  }
+  auto builder_iter = builders_.find(edit.GetColumnFamily());
```
jay-zhuang, line 731: In which condition, will it reach here? Should not new CF only be handled in Recovery mode (which is handled by `VersionEditHanlder::OnColumnFamilyAdd()`)?

db/version_edit_handler.cc
riversand963: It should. I should probably replace `TEST_Next()` with `Next()` and see if it will cause additional irrelevant changes.

db/version_edit_handler.cc
riversand963: If the primary later creates a new column family, it will write a VersionEdit to the MANIFEST. The secondary, whether in Recovery mode or Catchup mode, will be able to see this VersionEdit.
If the column family is added by the primary after the secondary is created and the secondary does not specify it in its `Open` call, then we ignore it. Should the secondary be interested in this column family, it should be opened with this column family in the first place. This is aligned with how we call `DB::Open()` with column families.
For column families that are specified when creating the secondary, we will see `VersionEdit.is_column_family_add == true` for them multiple times, whether secondary is in Recovery or Catchup mode. In fact, the primary will write a VersionEdit to add these column families each time it switches to a new MANIFEST.
Maybe an example can help.
```
MANIFEST-1 : | 'add_cf_1' | 'cf1': 'add [2.sst]' | 'cf1': 'add [3.sst]' |
MANIFEST-4 : |'**add_cf_1**' | 'cf1': 'add [2.sst,3.sst]' |
```
The code snippet here handles the case when the `add_cf_1` in MANIFEST-4 is seen. Suppose at this time, the secondary instance sees the second version edit of MANIFEST-1 but for some reason switches to MANIFEST-4. We need to build the version storage for cf1 from the empty base version of cf1, but from the version with 2.sst, otherwise the consistency check will fail due to two occurences of 2.sst.

db/version_set.cc
riversand963: Sure.

db/version_edit_handler.cc
jay-zhuang: Thanks for the explanation.

db/version_edit_handler.cc
jay-zhuang: will it expose too much internal information of version?

jay-zhuang (approved):
LGTM.

db/version_edit_handler.cc
riversand963: It will, but I am not too worried here because `Version` itself is not public. It's defined in `db` and not exposed to application. Do you have a case in mind that will be risky?

riversand963:
Thanks @jay-zhuang for the review

PR #8000, opened 2021-02-24
Add options to enable tracing for ZippyDB Laser etc.

nmjnmjnmj:
This is part of  https://www.internalfb.com/diff/D24276354

Enable tracing for zippydb and other services(Laser, etc.) which use rocksdb.

In the past couple months, we've canaried this to zippydb.disagg/zippydb.disagg.wildcard/zippydb.shadow.wildcard.flash etc. tiers and successfully collected multiple traces.

We'd like to checkin this to to collect traces in a more scalable way by just Canary the config changes.
It will enable us to roll out the config change and enable tracing in a larger scale.


ajkr (changes requested):
Can we update the title/description to describe only what changed in RocksDB? Internal service names, tier names, and tool links are confusing to have in the public commit history.

zhichao-cao (changes requested):
I'm not comfortable to put them in db_options. They are not related to the db behavior. Also, we have query tracing, IO tracing, and block cache tracing. Those options are miss leading. We have the options specifically for tracing: struct TraceOptions. uint64_t max_trace_file_size is already there. And I believe, some of the options you introduced, can be the member of TraceOptions. Another issue is, enable_trace and trace_is_on is not the option. We can introduce the function in tracer to get the status.

include/rocksdb/options.h
```diff
@@ -375,6 +375,20 @@ struct DBOptions {
...
+  std::string trace_path = "";
+
+  // This specifies the traces file name.
+  std::string trace_name_postfix = "";
+
+  // Max trace file size in byte
+  uint64_t max_trace_file_size = 1 * 1024 * 1024 * 1024;
+
+  // If true, the database will be created if it is missing.
+  bool enable_trace = false;
```
zhichao-cao, line 388: Why the database will be created if it is missing?

include/rocksdb/options.h
```diff
@@ -375,6 +375,20 @@ struct DBOptions {
...
   DBOptions* IncreaseParallelism(int total_threads = 16);
 #endif  // ROCKSDB_LITE
+  // This specifies the traces file path.
+  std::string trace_path = "";
+
+  // This specifies the traces file name.
+  std::string trace_name_postfix = "";
+
+  // Max trace file size in byte
+  uint64_t max_trace_file_size = 1 * 1024 * 1024 * 1024;
```
riversand963, line 385: I suggest that you do `1ULL * 1024 * 1024 * 1024` or `static_cast<uint64_t>(1) * 1024 * 1024 * 10224`.
Should people in the future change the default to something like `4 * 1024 * 1024 * 1024`, this value will overflow.

riversand963:
@nmjnmjnmj do you still plan to work on this?

nmjnmjnmj:
@zhichao-cao no, moved whole thing to zippydb side, will abandon this one

PR #8003, opened 2021-02-26
Support retrieving checksums for blob files from the MANIFEST when checkpointing

akankshamahajan15:
Summary: The checkpointing logic supports passing file level checksums
to the copy_file_cb callback function which is used by the backup code
for detecting corruption during file copies.
However, this is currently implemented only for table files.

This PR extends the checksum retrieval to blob files as well.

Test Plan: Add new test units

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -111,6 +111,14 @@ Status FileChecksumRetriever::ApplyVersionEdit(VersionEdit& edit,
       return s;
     }
   }
+  for (const auto& new_blob_file : edit.GetBlobFileAdditions()) {
+    Status s = file_checksum_list_.InsertOneFileChecksum(
+        new_blob_file.GetBlobFileNumber(), new_blob_file.GetChecksumValue(),
+        new_blob_file.GetChecksumMethod());
+    if (!s.ok()) {
+      return s;
+    }
+  }
   return Status::OK();
 }
 
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -112,9 +112,15 @@ Status FileChecksumRetriever::ApplyVersionEdit(VersionEdit& edit,
     }
   }
   for (const auto& new_blob_file : edit.GetBlobFileAdditions()) {
+    std::string checksum_value = new_blob_file.GetChecksumValue();
+    std::string checksum_method = new_blob_file.GetChecksumMethod();
+    assert(checksum_value.empty() == checksum_method.empty());
+    if (checksum_method.empty()) {
+      checksum_value = kUnknownFileChecksum;
+      checksum_method = kUnknownFileChecksumFuncName;
+    }
     Status s = file_checksum_list_.InsertOneFileChecksum(
-        new_blob_file.GetBlobFileNumber(), new_blob_file.GetChecksumValue(),
-        new_blob_file.GetChecksumMethod());
+        new_blob_file.GetBlobFileNumber(), checksum_value, checksum_method);
     if (!s.ok()) {
       return s;
     }
--- a/utilities/backupable/backupable_db_test.cc
+++ b/utilities/backupable/backupable_db_test.cc
@@ -770,6 +770,18 @@ class BackupableDBTest : public testing::Test {
     return s;
   }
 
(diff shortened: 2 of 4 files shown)


db/version_edit_handler.cc
```diff
@@ -111,6 +111,14 @@ Status FileChecksumRetriever::ApplyVersionEdit(VersionEdit& edit,
       return s;
     }
   }
+  for (const auto& new_blob_file : edit.GetBlobFileAdditions()) {
+    Status s = file_checksum_list_.InsertOneFileChecksum(
+        new_blob_file.GetBlobFileNumber(), new_blob_file.GetChecksumValue(),
+        new_blob_file.GetChecksumMethod());
```
ltamasi, line 117: In general, we also add files without checksums to the list, right?

If so, there's one difference between how this case is handled for table files and blob files. IIRC for table files, `file_checksum_func_name` would contain `kUnknownFileChecksumFuncName`, while for blob files it would be empty. It would be nice to make these uniform when adding them to `file_checksum_list_` by replacing an empty checksum method with `kUnknownFileChecksumFuncName` when adding blob files.

utilities/backupable/backupable_db_test.cc
```diff
@@ -770,6 +770,18 @@ class BackupableDBTest : public testing::Test {
     return s;
   }
 
+  Status GetBlobFilesInDB(std::vector<FileAttributes>* blob_files) {
```
ltamasi, line 773: Instead of duplicating the code of these helper methods, can we generalize the existing `GetTableFilesInDB` into e.g. `GetDataFilesInDB(FileType, std::vector<FileAttributes*>)` ? (Similarly for the other two.)

Internally, `GetDataFilesInDB` could use `ParseFileName` instead of doing a string search and compare the parsed file type with the passed in one.

utilities/backupable/backupable_db_test.cc
```diff
@@ -1414,6 +1542,36 @@ TEST_P(BackupableDBTestWithParam, TableFileCorruptedBeforeBackup) {
...
+// been corrupted and the blob file checksum is stored in the DB manifest for
+// the case when backup blob files will be stored in a shared directory
+TEST_P(BackupableDBTestWithParam, BlobFileCorruptedBeforeBackup) {
+  const int keys_iteration = 50000;
+  options_.enable_blob_files = true;
+  OpenDBAndBackupEngine(true /* destroy_old_data */);
+  FillDB(db_.get(), 0, keys_iteration);
+  CloseAndReopenDB();
+  // corrupt a random blob file in the DB directory
+  ASSERT_OK(CorruptRandomTableFileInDB());
```
ltamasi, line 1555: This and the call below are supposed to corrupt a blob file I suppose :)

ltamasi (commented):
Thanks for the patch @akankshamahajan15 !

db/version_edit_handler.cc
akankshamahajan15: Oh. Thanks for pointing it out. Also is it possible for some files to have checksums and some not from testing purpose if i want to add a unit test. Is it possible through an option in general?

ltamasi: > Also is it possible for some files to have checksums and some not from testing purpose if i want to add a unit test. Is it possible through an option in general?

In general, it's possible: one can open a database without specifying `file_checksum_gen_factory`, close it, then reopen it with `file_checksum_gen_factory` set. (It cannot be adjusted on the fly though while the DB is open b/c it's an immutable option.)

ltamasi: Sorry, what I meant here is that we could translate an empty checksum method/value for blob files to `kUnknownFileChecksumFuncName` / `kUnknownFileChecksum` when we build the `FileChecksumList` in `FileChecksumRetriever`.  Changing the MANIFEST related stuff (`BlobFileAddition` and friends) does not seem backwards compatible.

akankshamahajan15: oh ok. By adding blob files I thought when blob files are added to db.  I will update the code.

ltamasi (approved):
LGTM, thanks @akankshamahajan15 !

PR #8006, opened 2021-02-26
Still use SystemClock* instead of shared_ptr in StepPerfTimer

riversand963:
This is likely a temp fix before we figure out a better way.

PerfStepTimer is used intensively in certain benchmarking/testings. #7858 stores a `shared_ptr` to system clock in PerfStepTimer which gets created each time a `PerfStepTimer` object is created. The atomic operations in `shared_ptr` may add overhead in CPU cycles. Therefore, we change it back to a raw `SystemClock*` for now.

Test plan
make check

### DIFF
--- a/monitoring/iostats_context_imp.h
+++ b/monitoring/iostats_context_imp.h
@@ -40,7 +40,7 @@ extern __thread IOStatsContext iostats_context;
 // Declare and set start time of the timer
 #define IOSTATS_CPU_TIMER_GUARD(metric, clock)         \
   PerfStepTimer iostats_step_timer_##metric(           \
-      &(iostats_context.metric), clock, true,          \
+      &(iostats_context.metric), clock.get(), true,    \
       PerfLevel::kEnableTimeAndCPUTimeExceptForMutex); \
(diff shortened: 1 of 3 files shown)


pdillinger (approved):
This leaves internal API in a misleading state: taking a shared_ptr parameter suggests it will be copied with refcount but it is not. Up to you (and other internal factors) whether you want to update all the callers to make this change, or make this a small patch like this.

riversand963:
> This leaves internal API in a misleading state: taking a shared_ptr parameter suggests it will be copied with refcount but it is not. Up to you (and other internal factors) whether you want to update all the callers to make this change, or make this a small patch like this.

This makes sense. Looks like I will need to update three macros: `IOSTATS_CPU_TIMER_GUARD`, `PERF_CPU_TIMER_GUARD` and `PERF_TIMER_GUARD_WITH_CLOCK`.

riversand963:
And thanks @pdillinger for the quick review

mrambacher (approved):
I agree with Peter.  This was on my list of things to do, but I have not made it to it yet.

Thanks for the change!

PR #8007, opened 2021-02-27
Fix stuck issue in ~DBImpl() when HIGH pool is empty

sherriiiliu:
Found db could be stuck in `DBImpl::CloseHelper()` in a rare case - can repro in DBTest.DynamicMemtableOptions test.

When high pool is empty, flushes will be scheduled in LOW pool. If db is destroyed in this case, the scheduled work counters `bg_compaction_scheduled_ ` and `bg_flush_scheduled_ ` become incorrect. 
One example: suppose when db is ready to close, we have 1 scheduled flush job and 0 compaction job in LOW pool. After unscheduling 1 flush job, we have variables `compactions_unscheduled` = 1  and `flushes_unscheduled` = 0. 
```
  int compactions_unscheduled = env_->UnSchedule(this, Env::Priority::LOW);
  int flushes_unscheduled = env_->UnSchedule(this, Env::Priority::HIGH);
  ...
  bg_compaction_scheduled_ -= compactions_unscheduled;
  bg_flush_scheduled_ -= flushes_unscheduled;
```
With current calculation, `bg_compaction_scheduled_` becomes 0-1=-1 and `bg_flush_scheduled_`  becomes 1-0=1, but the truth should be 0 and 0. Thus db will be stuck in the following while loop waiting for completion signal, which will never be triggered.
```
  // Wait for background work to finish
  while (bg_bottom_compaction_scheduled_ || bg_compaction_scheduled_ ||
         bg_flush_scheduled_ || bg_purge_scheduled_ ||
         pending_purge_obsolete_files_ ||
         error_handler_.IsRecoveryInProgress()) {
    TEST_SYNC_POINT("DBImpl::~DBImpl:WaitJob");
    bg_cv_.Wait();
  }
```

As a fix, I add another counter `num_low_scheduled` to count the total work items scheduled in LOW pool. When db is being destroyed, we handle the special case (HIGH pool is empty) independently and remain the rest of logic the same for normal cases.


mrambacher:
Is this the same as #7928?

sherriiiliu:
> Is this the same as #7928?

oh yes, looks exactly the issue I found. Have that PR been merged?

sherriiiliu:
> > Is this the same as #7928?
> 
> oh yes, looks exactly the issue I found. Have that PR been merged?

I have different solutions as #7928 , also raised a doubt there. The unit test in that PR is great and should be added.

PR #8010, opened 2021-03-01
Add support for custom file systems to ldb and sst_dump

yhr:
This PR adds support for custom file systems to ldb and sst_dump by adding command line options for specifying --fs_uri and --backup_fs uri (for ldb backup/restore commands). fs_uri is already supported in db_bench and db_stress, and there is already support in ldb and db stress for specifying customized envs.

The PR also fixes what looks like a bug in the ldb backup/restore commands. As it is right now, backups can only be made from and to the same environment/file system which does not seem to be the intended behavior. This PR makes it possible to do/restore backups between different envs/file systems.

Example:
`./ldb backup --fs_uri=zenfs://dev:nvme2n1 --backup_fs_uri=posix:// --backup_dir=/tmp/my_rocksdb_backup  --db=rocksdbtest/dbbench
`

### DIFF
--- a/tools/ldb_cmd.cc
+++ b/tools/ldb_cmd.cc
@@ -49,6 +49,7 @@ class FileChecksumGenCrc32c;
 class FileChecksumGenCrc32cFactory;
 
 const std::string LDBCommand::ARG_ENV_URI = "env_uri";
+const std::string LDBCommand::ARG_FS_URI = "fs_uri";
 const std::string LDBCommand::ARG_DB = "db";
 const std::string LDBCommand::ARG_PATH = "path";
 const std::string LDBCommand::ARG_SECONDARY_PATH = "secondary_path";
@@ -286,6 +287,16 @@ LDBCommand* LDBCommand::SelectCommand(const ParsedParams& parsed_params) {
   return nullptr;
 }
 
+static Env* GetCompositeEnv(std::shared_ptr<FileSystem> fs) {
+  static std::shared_ptr<Env> composite_env = NewCompositeEnv(fs);
+  return composite_env.get();
+}
+
+static Env* GetCompositeBackupEnv(std::shared_ptr<FileSystem> fs) {
+  static std::shared_ptr<Env> composite_backup_env = NewCompositeEnv(fs);
+  return composite_backup_env.get();
+}
+
 /* Run the command, and return the execute result. */
 void LDBCommand::Run() {
   if (!exec_state_.IsNotStarted()) {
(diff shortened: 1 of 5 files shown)


yhr:
@ajkr : Could you have a look at this? This should wrap up the custom file system support.

tools/ldb_cmd.cc
```diff
@@ -294,12 +305,33 @@ void LDBCommand::Run() {
...
+      Status s = Env::LoadEnv(env_uri_, &env, &env_guard_);
+      if (!s.ok() && !s.IsNotFound()) {
+        fprintf(stderr, "LoadEnv: %s\n", s.ToString().c_str());
+        exec_state_ = LDBCommandExecuteResult::Failed(s.ToString());
+        return;
+      }
+    } else if (!fs_uri_.empty()) {
+      std::shared_ptr<FileSystem> fs;
+      Status s = FileSystem::Load(fs_uri_, &fs);
+      if (fs == nullptr) {
```
mrambacher, line 327: Why is this fs == nullptr instead of !s.ok() ?

tools/ldb_cmd.cc
```diff
@@ -294,12 +305,33 @@ void LDBCommand::Run() {
...
+        fprintf(stderr, "error: %s\n", s.ToString().c_str());
+        exec_state_ = LDBCommandExecuteResult::Failed(s.ToString());
+        return;
+      }
+      env = GetCompositeEnv(fs);
```
mrambacher, line 332: Why GetCompositeEnv vs using the existing env_guard_ like above?

tools/ldb_cmd.cc
```diff
@@ -3087,15 +3136,26 @@ void BackupCommand::DoCommand() {
...
   }
   fprintf(stdout, "open db OK\n");
+
   Env* custom_env = nullptr;
-  Env::LoadEnv(backup_env_uri_, &custom_env, &backup_env_guard_);
-  assert(custom_env != nullptr);
+  if (!backup_fs_uri_.empty()) {
+    std::shared_ptr<FileSystem> fs;
+    Status s = FileSystem::Load(backup_fs_uri_, &fs);
+    if (fs == nullptr) {
```
mrambacher, line 3144: Same question as above on checking status

tools/ldb_cmd.cc
```diff
@@ -3087,15 +3136,26 @@ void BackupCommand::DoCommand() {
...
+      exec_state_ = LDBCommandExecuteResult::Failed(s.ToString());
+      return;
+    }
+    custom_env = GetCompositeBackupEnv(fs);
```
mrambacher, line 3148: Same question as above on using backup_env_guard_

tools/ldb_cmd.cc
```diff
@@ -3087,15 +3136,26 @@ void BackupCommand::DoCommand() {
...
+  } else {
+    Env::LoadEnv(backup_env_uri_, &custom_env, &backup_env_guard_);
```
mrambacher, line 3150: Shouldn't this check the return Status from LoadEnv?

tools/ldb_cmd.cc
yhr: It is to ensure that env::default and the custom env will always be destroyed in the right order, (same as for db_bench)

tools/ldb_cmd.cc
yhr: See above

tools/ldb_cmd.cc
yhr: It's the same check as in db_bench and db_stress, and is equivalent to checking s.

tools/ldb_cmd.cc
yhr: I did not change the existing behavior, but that might be a good idea.

PR #8012, opened 2021-03-01
Revamp check_format_compatible.sh

pdillinger:
Summary:
* Adds backup/restore forward/backward compatibility testing
* Adds forward/backward compatibility testing to sst ingestion
* More structure sharing and comments for the lists of branches
comprising each group
* Less reliant on invariants between groups with de-duplication logic
* Restructured for n+1 branch checkout+build steps rather than something
like 3n. Should be much faster despite more checks.

And to make manual runs easier

* On success, restores working trees to original working branch (aborts
early if uncommitted changes) and deletes temporary branch & remote
* Adds SHORT_TEST=1 mode that uses only the oldest version for each
* Adds USE_SSH=1 to use ssh instead of https for github
group

Test Plan: a number of manual tests, mostly with SHORT_TEST=1. Using one
version older for any of the groups (except I didn't check
db_backward_only_refs) fails. Changing default format_version to 5
(planned) without updating this script fails as it should, and passes
with appropriate update. Full local run passed (had to remove "2.7.fb.branch"
due to compiler issues, also before this change).


tools/check_format_compatible.sh
```diff
@@ -105,89 +165,185 @@ ingest_external_sst()
...
-   compare_db $test_dir/$checkout_obj $compare_base_db_dir forward_${checkout_obj}_dump.txt 0
-done
+  if member_of_array "$checkout_ref" "${ext_backward_only_refs[@]}" ||
+    member_of_array "$checkout_ref" "${ext_forward_refs[@]}"
+  then
+    rm -rf $ext_test_dir/${checkout_ref}_ingest
+    echo "== Use $current_checkout_name to ingest extern SST file from $checkout_ref"
+    ingest_external_sst $ext_test_dir/${checkout_ref}_ingest $current_ext_test_dir
+    compare_db $ext_test_dir/${checkout_ref}_ingest ${current_ext_test_dir}_ingest db_dump.txt 1 1
+    echo hi
```
riversand963, line 329: Is this intended or a temporary debug message?

riversand963 (approved):
LGTM with minor comments/questions. Thanks @pdillinger for the improvement!

tools/check_format_compatible.sh
```diff
@@ -105,89 +165,185 @@ ingest_external_sst()
...
+
+# To restore to prior branch at the end
+orig_branch=`git branch --show-current`
+tmp_branch=_tmp_format_compatible
+tmp_origin=_tmp_origin
+
+# Don't depend on what current "origin" might be
+git remote remove $tmp_origin 2>/dev/null
+if [ "$USE_SSH" ]; then
+  git remote add $tmp_origin "git@github.com:facebook/rocksdb.git"
```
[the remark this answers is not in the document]
pdillinger: This remote is only used with the hard-coded branch names in the test. The <checkout> passed on the command line can be any locally recognized ref, including HEAD.

tools/check_format_compatible.sh
pdillinger: No

pdillinger:
Running time in CircleCI went from 2h13m to 59m.

Before: https://app.circleci.com/pipelines/github/facebook/rocksdb/5990/workflows/62a1c39e-894e-4dfc-97e6-d4b7a2ac22ed/jobs/100162
After: https://app.circleci.com/pipelines/github/facebook/rocksdb/6003/workflows/d0b787c7-6687-422b-a0b4-0a7b92d1d212/jobs/100583

PR #8013, opened 2021-03-02
Break down the amount of data written during flushes/compactions per file type

ltamasi:
Summary:
The patch breaks down the "bytes written" (as well as the "number of output files")
compaction statistics into two, so the values are logged separately for table files
and blob files in the info log, and are shown in separate columns (`Write(GB)` for table
files, `Wblob(GB)` for blob files) when the compaction statistics are dumped.
This will also come in handy for fixing the write amplification statistics, which currently
do not consider the amount of data read from blob files during compaction. (This will
be fixed by an upcoming patch.)

Test Plan:
Ran `make check` and `db_bench`.


riversand963 (approved):
LGTM.

PR #8015, opened 2021-03-02
Possibly bump NUMBER_OF_RESEEKS_IN_ITERATION

riversand963:
When changing db iterator direction, we may perform a reseek.
Therefore, we should bump the NUMBER_OF_RESEEKS_IN_ITERATION counter.

Test plan
make check

### DIFF
--- a/db/db_iter.cc
+++ b/db/db_iter.cc
@@ -722,19 +722,17 @@ bool DBIter::ReverseToBackward() {
         saved_key_.GetUserKey(), kMaxSequenceNumber, kValueTypeForSeek));
     if (!expect_total_order_inner_iter()) {
       iter_.SeekForPrev(last_key.GetInternalKey());
-      RecordTick(statistics_, NUMBER_OF_RESEEKS_IN_ITERATION);
     } else {
       // Some iterators may not support SeekForPrev(), so we avoid using it
       // when prefix seek mode is disabled. This is somewhat expensive
       // (an extra Prev(), as well as an extra change of direction of iter_),
       // so we may need to reconsider it later.
       iter_.Seek(last_key.GetInternalKey());
-      RecordTick(statistics_, NUMBER_OF_RESEEKS_IN_ITERATION);
       if (!iter_.Valid() && iter_.status().ok()) {
         iter_.SeekToLast();
-        RecordTick(statistics_, NUMBER_OF_RESEEKS_IN_ITERATION);
       }
     }
+    RecordTick(statistics_, NUMBER_OF_RESEEKS_IN_ITERATION);
   }
 
   direction_ = kReverse;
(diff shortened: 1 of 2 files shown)


db/db_iter.cc
```diff
@@ -681,6 +681,7 @@ bool DBIter::ReverseToForward() {
     last_key.SetInternalKey(ParsedInternalKey(
         saved_key_.GetUserKey(), kMaxSequenceNumber, kValueTypeForSeek));
     iter_.Seek(last_key.GetInternalKey());
+    RecordTick(statistics_, NUMBER_OF_RESEEKS_IN_ITERATION);
```
ltamasi, line 684: Should we update `ReverseToBackward` as well?

ltamasi (commented):
Thanks for the patch @riversand963 !

db/db_iter.cc
riversand963: We can. I didn't not include `ReverseToBackward` because it involves Merge, and I intentionally wants to keep the scope of a single PR as small as possible. Maybe I can open a new PR for that later.

riversand963: Let me update it and see if it can remain small.

riversand963: @ltamasi , fortunately the new update does not break any existing test. Therefore, I added a few lines to test it.

db/db_iter.cc
```diff
@@ -721,14 +722,17 @@ bool DBIter::ReverseToBackward() {
...
     } else {
       // Some iterators may not support SeekForPrev(), so we avoid using it
       // when prefix seek mode is disabled. This is somewhat expensive
       // (an extra Prev(), as well as an extra change of direction of iter_),
       // so we may need to reconsider it later.
       iter_.Seek(last_key.GetInternalKey());
+      RecordTick(statistics_, NUMBER_OF_RESEEKS_IN_ITERATION);
       if (!iter_.Valid() && iter_.status().ok()) {
         iter_.SeekToLast();
+        RecordTick(statistics_, NUMBER_OF_RESEEKS_IN_ITERATION);
```
ltamasi, line 735: We have the same pattern (`Seek` potentially followed by `SeekToLast` if we went too far) in `FindValueForCurrentKeyUsingSeek`; I'm wondering if we should bump the counter twice there as well? (Or just once here too to be consistent with that.)

ltamasi (approved):
Thanks once again for the fix @riversand963 ! LGTM, just one question

db/db_iter.cc
riversand963: Good catch. I do not have a strong opinion about this. Since the counter values may be used by applications already, I suggest that I keep it consistent with existing pattern to reduce the scope of this PR. We can solve the issue you mentioned in a separate PR.

riversand963:
Thanks @ltamasi for the detailed review!

PR #8025, opened 2021-03-04
Universal Binaries for macOS

adamretter:
Support for building a static library for both x86_64 and arm64 within a single fat Universal Binary on macOS


adamretter:
This was completed elsewhere more recently.

PR #8028, opened 2021-03-04
Revert SystemClock (PR #7858 & #7851)

pdillinger:
This is the master version of reverting these two commits. shard_ptr<SystemClock> caused significant performance regression and we don't know the scope of the damage. The other commit/PR is closely intertwined so is reverted also.

Test Plan: CI


pdillinger:
By the way, I hope we can sort out the performance issues without landing this, but I thought we should have proof that the revert likely going into 6.18 could be rolled forward without extraordinary effort or damage.

pdillinger:
#8033 approved instead

PR #8031, opened 2021-03-04
Set perf_level to kEnableTimeExceptForMutex in regression script

anand1976:
Set the perf_level in ```tools/regression_test.sh``` in order to exercise ```PerfContext``` counters in regression tests.

Test:
Manually run the script

### DIFF
--- a/tools/regression_test.sh
+++ b/tools/regression_test.sh
@@ -158,7 +158,7 @@ function init_arguments {
 
   current_time=$(date +"%F-%H:%M:%S")
   RESULT_PATH=${RESULT_PATH:-"$1/results/$current_time"}
-  COMMIT_ID=`hg id -i`
+  COMMIT_ID=`git log | head -n1 | cut -c 8-`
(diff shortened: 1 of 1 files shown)


tools/regression_test.sh
```diff
@@ -222,7 +222,9 @@ function run_db_bench {
   db_bench_cmd="("'\$(which time)'" -p $DB_BENCH_DIR/db_bench \
       --benchmarks=$1 --db=$DB_PATH --wal_dir=$WAL_PATH \
       --use_existing_db=$USE_EXISTING_DB \
+      --perf_level=3 \
```
riversand963, line 225: Do we want to always enable perf_level 3, or should we also run with perf_level (default)?

anand1976: Make it an option? It can be set based on the regression test workload. Can be set to 3 for "myrocks" and default for "viewstate".

riversand963 (approved):
LGTM. No further comment. Thanks @anand1976 for the PR

ajkr:
Ping @anand1976 , needs import.

PR #8036, opened 2021-03-05
Deflake tests of compaction based on compensated file size

ajkr:
CompactionDeletionTriggerReopen was observed to be flaky recently:
https://app.circleci.com/pipelines/github/facebook/rocksdb/6030/workflows/787af4f3-b9f7-4645-8e8d-1fb0ebf05539/jobs/101451.

I went through it and the related tests and arrived at different
conclusions on what constraints we can expect on DB size. Some
constraints got looser and some got tighter. The particular constraint
that flaked got a lot looser so at least the flake linked above would have been prevented.

### DIFF
--- a/db/db_compaction_test.cc
+++ b/db/db_compaction_test.cc
@@ -206,6 +206,7 @@ Options DeletionTriggerOptions(Options options) {
       options.target_file_size_base * options.target_file_size_multiplier;
   options.max_bytes_for_level_multiplier = 2;
   options.disable_auto_compactions = false;
+  options.compaction_options_universal.max_size_amplification_percent = 100;
   return options;
 }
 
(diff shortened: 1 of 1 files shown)


db/db_compaction_test.cc
```diff
@@ -206,6 +206,7 @@ Options DeletionTriggerOptions(Options options) {
       options.target_file_size_base * options.target_file_size_multiplier;
   options.max_bytes_for_level_multiplier = 2;
   options.disable_auto_compactions = false;
+  options.compaction_options_universal.max_size_amplification_percent = 100;
```
ajkr, line 209: I changed this from the default (200) because "CompactionDeletionTrigger" was in the [190, 200) range and relying on size-ratio compaction, which just so happened to pick through to the last sorted run. Size-amp is more reliable as it's guaranteed to pick through the last sorted run.

db/db_compaction_test.cc
```diff
@@ -360,12 +361,33 @@ TEST_P(DBCompactionTestWithParam, CompactionDeletionTrigger) {
     for (int k = 0; k < kTestSize; ++k) {
       ASSERT_OK(Delete(Key(k)));
     }
-    ASSERT_OK(dbfull()->TEST_WaitForFlushMemTable());
+    Flush();
```
ajkr, line 364: I found it hard to reason about an unpredictable number of tombstones getting stuck in the memtable, so triggered a manual flush.

db/db_compaction_test.cc
```diff
@@ -767,10 +793,13 @@ TEST_F(DBCompactionTest, DisableStatsUpdateReopen) {
     if (options.skip_stats_update_on_db_open) {
       // If update stats on DB::Open is disable, we don't expect
       // deletion entries taking effect.
-      ASSERT_LT(db_size[0] / 3, db_size[2]);
+      ASSERT_LE(db_size[0], db_size[2]);
```
ajkr, line 796: This one I'm most worried about. Open still does a flush that can trigger compaction, irrespective of deletion entries.

ajkr: Fixed.

db/db_compaction_test.cc
ajkr: Done.

db/db_compaction_test.cc
```diff
@@ -360,12 +361,32 @@ TEST_P(DBCompactionTestWithParam, CompactionDeletionTrigger) {
...
+    // `db_size[0] / 16`, and so on.
+    //
+    // Claim: in universal compaction none of the original data will remain once
+    // compactions settle.
+    //
+    // Proof: The compensated size of the file containing the most tombstones is
+    // enough on its own to trigger size amp compaction. Size amp compaction is
+    // a full compaction, so all tombstones meet the obsolete keys they cover.
+
+    // The looser of the above two constraints can apply in all cases.
```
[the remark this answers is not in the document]
ajkr: Yeah, these are not great tests. I'll probably land this since having documented/flaky limits is at least better than having undocumented/flaky limits. Hopefully we still later think of a simpler way to test whether compaction picking takes into account tombstones.

db/db_compaction_test.cc
```diff
@@ -360,12 +361,32 @@ TEST_P(DBCompactionTestWithParam, CompactionDeletionTrigger) {
...
+    ASSERT_GT(db_size[0] / 2, db_size[1]);
```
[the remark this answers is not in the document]
ajkr: I guess we can assert zero for universal compaction according to the analysis. Changed it and verified it works.

db/db_compaction_test.cc
```diff
@@ -735,8 +761,15 @@ TEST_F(DBCompactionTest, DisableStatsUpdateReopen) {
       values.push_back(rnd.RandomString(kCDTValueSize));
       ASSERT_OK(Put(Key(k), values[k]));
     }
-    ASSERT_OK(dbfull()->TEST_WaitForFlushMemTable());
+    Flush();
```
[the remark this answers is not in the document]
ajkr: Done.

ajkr (commented):
Thanks for the review!

PR #8037, opened 2021-03-06
Use SST file manager to track blob files as well

akankshamahajan15:
Summary: Extend support to track blob files in SST File manager.
    This PR notifies SstFileManager whenever a new blob file is created,
    via OnAddFile and  an obsolete blob file deleted via OnDeleteFile
    and delete file via ScheduleFileDeletion.

 Test Plan: Add new unit tests

 Reviewers:

 Subscribers:

  Tasks: T85239542

  Tags:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,7 +13,7 @@
 * Clarified the required semantics of Read() functions in FileSystem and Env APIs. Please ensure any custom implementations are compliant.
 * For the new integrated BlobDB implementation, compaction statistics now include the amount of data read from blob files during compaction (due to garbage collection or compaction filters). Write amplification metrics have also been extended to account for data read from blob files.
 * Add EqualWithoutTimestamp() to Comparator.
-* Extend support to track blob files in SSTFileManager whenever a blob file is created/deleted. Blob files will be scheduled to delete via SSTFileManager and  SStFileManager will now take blob files in account while calculating size and space limits along with SST files.
+* Extend support to track blob files in SSTFileManager whenever a blob file is created/deleted. Blob files will be scheduled to delete via SSTFileManager and SStFileManager will now take blob files in account while calculating size and space limits along with SST files.
 
 ### New Features
 * Support compaction filters for the new implementation of BlobDB. Add `FilterBlobByKey()` to `CompactionFilter`. Subclasses can override this method so that compaction filters can determine whether the actual blob value has to be read during compaction. Use a new `kUndetermined` in `CompactionFilter::Decision` to indicated that further action is necessary for compaction filter to make a decision.
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,6 +13,7 @@
 * Clarified the required semantics of Read() functions in FileSystem and Env APIs. Please ensure any custom implementations are compliant.
 * For the new integrated BlobDB implementation, compaction statistics now include the amount of data read from blob files during compaction (due to garbage collection or compaction filters). Write amplification metrics have also been extended to account for data read from blob files.
 * Add EqualWithoutTimestamp() to Comparator.
+* Extend support to track blob files in SSTFileManager whenever a blob file is created/deleted. Blob files will be scheduled to delete via SSTFileManager and  SStFileManager will now take blob files in account while calculating size and space limits along with SST files.
 
 ### New Features
 * Support compaction filters for the new implementation of BlobDB. Add `FilterBlobByKey()` to `CompactionFilter`. Subclasses can override this method so that compaction filters can determine whether the actual blob value has to be read during compaction. Use a new `kUndetermined` in `CompactionFilter::Decision` to indicated that further action is necessary for compaction filter to make a decision.
--- a/db/blob/blob_file_completion_callback.h
+++ b/db/blob/blob_file_completion_callback.h
@@ -23,27 +23,30 @@ class BlobFileCompletionCallback {
         mutex_(mutex),
         error_handler_(error_handler) {}
 
-#ifndef ROCKSDB_LITE
-  Status Callback(const std::string& file_name) {
+  Status OnBlobFileCompleted(const std::string& file_name) {
     Status s;
 
+#ifndef ROCKSDB_LITE
     auto sfm = static_cast<SstFileManagerImpl*>(sst_file_manager_);
     if (sfm) {
       // Report new blob files to SstFileManagerImpl
-      s = sfm->OnAddFile(file_name);
+      // TODO (PR7798).  We should only add the file to the FileManager if it
+      // exists. Otherwise, some tests may fail.  Ignore the error in the
+      // interim.
+      sfm->OnAddFile(file_name).PermitUncheckedError();
       if (sfm->IsMaxAllowedSpaceReached()) {
         s = Status::SpaceLimit("Max allowed space was reached");
-        TEST_SYNC_POINT_CALLBACK(
-            "BlobFileCompletionCallback::CallBack::MaxAllowedSpaceReached", &s);
+        TEST_SYNC_POINT(
+            "BlobFileCompletionCallback::CallBack::MaxAllowedSpaceReached");
         InstrumentedMutexLock l(mutex_);
         error_handler_->SetBGError(s, BackgroundErrorReason::kFlush);
       }
     }
+#else
+    (void)file_name;
+#endif  // ROCKSDB_LITE
     return s;
   }
-#else
-  Status Callback(const std::string& /*file_name*/) { return Status::OK(); }
-#endif  // ROCKSDB_LITE
 
  private:
   SstFileManager* sst_file_manager_;
--- a/db/blob/blob_file_completion_callback.h
+++ b/db/blob/blob_file_completion_callback.h
@@ -30,10 +30,7 @@ class BlobFileCompletionCallback {
     auto sfm = static_cast<SstFileManagerImpl*>(sst_file_manager_);
     if (sfm) {
       // Report new blob files to SstFileManagerImpl
-      // TODO (PR7798).  We should only add the file to the FileManager if it
-      // exists. Otherwise, some tests may fail.  Ignore the error in the
-      // interim.
-      sfm->OnAddFile(file_name).PermitUncheckedError();
+      s = sfm->OnAddFile(file_name);
       if (sfm->IsMaxAllowedSpaceReached()) {
         s = Status::SpaceLimit("Max allowed space was reached");
         TEST_SYNC_POINT(
--- a/db/blob/blob_file_completion_callback.h
+++ b/db/blob/blob_file_completion_callback.h
@@ -1,0 +1,53 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+//
+// Copyright (c) 2011 The LevelDB Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file. See the AUTHORS file for names of contributors.
+#pragma once
+
+#include "db/error_handler.h"
+#include "file/sst_file_manager_impl.h"
+#include "rocksdb/status.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+class BlobFileCompletionCallback {
+ public:
+  BlobFileCompletionCallback(SstFileManager* sst_file_manager,
+                             InstrumentedMutex* mutex,
+                             ErrorHandler* error_handler)
+      : sst_file_manager_(sst_file_manager),
+        mutex_(mutex),
+        error_handler_(error_handler) {}
+
+#ifndef ROCKSDB_LITE
+  Status Callback(const std::string& file_name) {
+    Status s;
+
+    auto sfm = static_cast<SstFileManagerImpl*>(sst_file_manager_);
+    if (sfm) {
+      // Report new blob files to SstFileManagerImpl
+      s = sfm->OnAddFile(file_name);
+      if (sfm->IsMaxAllowedSpaceReached()) {
+        s = Status::SpaceLimit("Max allowed space was reached");
+        TEST_SYNC_POINT_CALLBACK(
+            "BlobFileCompletionCallback::CallBack::MaxAllowedSpaceReached", &s);
+        InstrumentedMutexLock l(mutex_);
+        error_handler_->SetBGError(s, BackgroundErrorReason::kFlush);
+      }
+    }
+    return s;
+  }
+#else
+  Status Callback(const std::string& /*file_name*/) { return Status::OK(); }
+#endif  // ROCKSDB_LITE
+
+ private:
+  SstFileManager* sst_file_manager_;
+  InstrumentedMutex* mutex_;
+  ErrorHandler* error_handler_;
+};
+}  // namespace ROCKSDB_NAMESPACE
--- a/db/builder.cc
+++ b/db/builder.cc
@@ -293,6 +293,7 @@ Status BuildTable(
             s = add_s;
           }
           if (sfm->IsMaxAllowedSpaceReached()) {
+            TEST_SYNC_POINT("BuildTable:MaxAllowedSpaceReached");
             s = Status::SpaceLimit("Max allowed space was reached");
           }
         }
--- a/db/builder.cc
+++ b/db/builder.cc
@@ -25,6 +25,7 @@
 #include "db/version_edit.h"
 #include "file/filename.h"
 #include "file/read_write_util.h"
+#include "file/sst_file_manager_impl.h"
 #include "file/writable_file_writer.h"
 #include "monitoring/iostats_context_imp.h"
 #include "monitoring/thread_status_util.h"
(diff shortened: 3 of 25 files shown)


db/compaction/compaction_job.cc
```diff
@@ -1192,6 +1192,30 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
       status = blob_file_builder->Finish();
     }
 
+#ifndef ROCKSDB_LITE
+    auto sfm =
```
riversand963, line 1196: Same question for compaction, but didn't verify the details.

db/builder.cc
```diff
@@ -281,6 +282,28 @@ Status BuildTable(
       if (s.ok()) {
         s = blob_file_builder->Finish();
       }
+      assert(blob_file_additions);
+      assert(!ioptions.cf_paths.empty());
+#ifndef ROCKSDB_LITE
```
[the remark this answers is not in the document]
akankshamahajan15: Ah ok. Thanks for pointing it out. I will check and update the both locations. I added here because Blob file builder calls Finish(). I didn't consider about failure.

ltamasi (commented):
Thanks for the PR @akankshamahajan15 !

db/db_test_util.cc
```diff
@@ -1436,26 +1462,26 @@ void DBTestBase::CopyFile(const std::string& source,
   ASSERT_OK(destfile->Close());
 }
 
-Status DBTestBase::GetAllSSTFiles(
-    std::unordered_map<std::string, uint64_t>* sst_files,
+Status DBTestBase::GetAllDataFiles(
```
ltamasi, line 1465: Nice change, thanks!

db/compaction/compaction_job.cc
```diff
@@ -1192,6 +1192,30 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
...
+        static_cast<SstFileManagerImpl*>(db_options_.sst_file_manager.get());
+    if (status.ok() && sfm) {
+      // Report new blob files to SstFileManagerImpl
+      for (auto const& blob_file : sub_compact->blob_file_additions) {
```
ltamasi, line 1200: One more thing to keep in mind here, considering that compactions do not perform cleanup: `BlobFileBuilder` also exposes `blob_file_paths`, which contains all blob files, not just the successfully closed ones like `blob_file_additions`. We might want to use `blob_file_paths` here; that would also eliminate the need to form the paths below.

db/compaction/compaction_job.cc
riversand963: Good point. I wonder why not also delete output file from a failed compaction immediately. This file may very likely be corrupted or inaccessible.

ltamasi: > Good point. I wonder why not also delete output file from a failed compaction immediately. This file may very likely be corrupted or inaccessible.

Right, and I do wonder why it's not handled uniformly across flush/compaction. On the other hand, I think it's pretty harmless since the output of failed compactions does not get installed (it's just a matter of when trash gets cleaned up).

ltamasi: @riversand963 @akankshamahajan15  In terms of reporting files to `SstFileManager`, I did some more digging and now I actually think what we do for compactions (i.e. reporting new files as we go) makes a lot of sense. Calling `OnAddFile` results in `total_file_size_` getting incremented in `SstFileManagerImpl`, which is then used for checking the "max allowed space reached" condition as well as allowing further compactions. Considering that compactions can run for a long time and generate a lot of data, I think doing it this way is better b/c it keeps `total_file_size_` more up-to-date.

In fact, we might want to report blob files as they are generated (or deleted) during flush too for the same reason. (Note that a flush job only generates a single SST; one way to look at the current code which only notifies `sfm` upon success is treating it as an optimization that eliminates a pair of `OnAddFile` / `OnDeleteFile` in failure cases.) Based on this, the right place to notify `sfm` would probably be in `BlobFileBuilder`.

ltamasi:
@akankshamahajan15 
Sorry, couple more things:
1) Could you also update the comments in `sst_file_manager_impl.{h,cc}` as needed?
2) The two `OnAddFile` overloads and `OnAddFileImpl` has a boolean parameter `compaction`, which, as far as I can tell, is always `false`. Could you confirm this? If so, we could remove this flag and clean up all the code and variables that are related to it. (For instance, seems to me that `in_progress_files_size_` / `in_progress_files_` are redundant, assuming that `compaction` is always `false`.)

db/compaction/compaction_job.cc
akankshamahajan15: @ltamasi  Thanks for looking into it. It makes sense. I will notify the sfm in BlobFileBuilder.

akankshamahajan15:
> 2. The two `OnAddFile` overloads and `OnAddFileImpl` have a boolean parameter `compaction`, which, as far as I can tell, is always `false`. Could you confirm this? If so, we could remove this flag and clean up all the code and variables that are related to it. (For instance, seems to me that `in_progress_files_size_` / `in_progress_files_` are redundant, assuming that `compaction` is indeed always `false`.)

Yes, you are right. I checked all the calls and compaction is always false. So I will update the code and remove this paramater and its related code.

HISTORY.md
```diff
@@ -13,6 +13,7 @@
 * Clarified the required semantics of Read() functions in FileSystem and Env APIs. Please ensure any custom implementations are compliant.
 * For the new integrated BlobDB implementation, compaction statistics now include the amount of data read from blob files during compaction (due to garbage collection or compaction filters). Write amplification metrics have also been extended to account for data read from blob files.
 * Add EqualWithoutTimestamp() to Comparator.
+* Extend support to track blob files in SSTFileManager whenever a blob file is created/deleted. Blob files will be scheduled to delete via SSTFileManager and  SStFileManager will now take blob files in account while calculating size and space limits along with SST files.
```
riversand963, line 16: Nit: remove extra whitespace.

db/blob/blob_file_completion_callback.h
```diff
@@ -0,0 +1,52 @@
...
+        InstrumentedMutexLock l(mutex_);
+        error_handler_->SetBGError(s, BackgroundErrorReason::kFlush);
+      }
+    }
+#endif
+    return s;
+  }
+
+ private:
+  ImmutableDBOptions immutable_db_options_;
```
riversand963, line 48: Do we need to copy here?

db/compaction/compaction_job.cc
```diff
@@ -976,14 +976,21 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
 
   std::vector<std::string> blob_file_paths;
 
+  BlobFileCompletionCallback* blob_callback =
```
riversand963, line 979: When is this object free'ed?

riversand963 (commented):
Took another pass, thanks @akankshamahajan15 for the updates.

db/blob/blob_file_completion_callback.h
```diff
@@ -0,0 +1,52 @@
...
+                             ErrorHandler* error_handler)
+      : immutable_db_options_(immutable_db_options),
+        mutex_(mutex),
+        error_handler_(error_handler) {}
+
+  Status Callback(const std::string& file_name) {
+    Status s;
+#ifndef ROCKSDB_LITE
+    auto sfm = static_cast<SstFileManagerImpl*>(
+        immutable_db_options_.sst_file_manager.get());
```
ltamasi: I think we don't need to `static_cast` here: both `OnAddFile` and `IsMaxAllowedSpaceReached` are part of the `SstFileManager` interface. If so, it would also suffice to `#include "rocksdb/sst_file_manager.h"` instead of the private header containing the `Impl` class.

db/builder.cc
```diff
@@ -14,6 +14,7 @@
 #include <vector>
 
 #include "db/blob/blob_file_builder.h"
+#include "db/blob/blob_file_completion_callback.h"
```
ltamasi, line 17: Minor: this `#include` does not seem necessary (the fwd declaration in the `builder.h` is sufficient, since we're just passing a pointer around).

On a similar note, I don't think we need to `#include "file/sst_file_manager_impl.h"` below.

db/blob/blob_file_completion_callback.h
```diff
@@ -0,0 +1,53 @@
...
+                             ErrorHandler* error_handler)
+      : sst_file_manager_(sst_file_manager),
+        mutex_(mutex),
+        error_handler_(error_handler) {}
+
+#ifndef ROCKSDB_LITE
+  Status Callback(const std::string& file_name) {
+    Status s;
+
+    auto sfm = static_cast<SstFileManagerImpl*>(sst_file_manager_);
```
ltamasi, line 30: I think we don't need to `static_cast` here: both `OnAddFile` and `IsMaxAllowedSpaceReached` are part of the `SstFileManager` interface. If so, it would also suffice to `#include "rocksdb/sst_file_manager.h"` instead of the private header containing the `Impl` class.

db/blob/blob_file_completion_callback.h
```diff
@@ -0,0 +1,53 @@
...
+ public:
+  BlobFileCompletionCallback(SstFileManager* sst_file_manager,
+                             InstrumentedMutex* mutex,
+                             ErrorHandler* error_handler)
+      : sst_file_manager_(sst_file_manager),
+        mutex_(mutex),
+        error_handler_(error_handler) {}
+
+#ifndef ROCKSDB_LITE
+  Status Callback(const std::string& file_name) {
```
ltamasi, line 27: Could we have a more specific name for the callback method, e.g. `OnBlobFileCompleted` ?

db/db_impl/db_impl_open.cc
```diff
@@ -1755,13 +1758,12 @@ Status DBImpl::Open(const DBOptions& db_options, const std::string& dbname,
         FileType file_type;
         std::string file_path = path + "/" + file_name;
         if (ParseFileName(file_name, &file_number, &file_type) &&
-            file_type == kTableFile) {
+            (file_type == kTableFile || file_type == kBlobFile)) {
```
ltamasi, line 1761: I think we should probably leave this as-is for now (i.e. don't use this branch for blob files) since we know blob files are not there in the map at the moment. (In other words, this should be updated when the `TODO` above is addressed.)

file/sst_file_manager_impl.cc
```diff
@@ -440,24 +442,15 @@ void SstFileManagerImpl::WaitForEmptyTrash() {
...
   } else {
     total_files_size_ += file_size;
-    if (compaction) {
-      // Keep track of the size of files created by in-progress compactions.
-      // When calculating whether there's enough headroom for new compactions,
-      // this will be subtracted from cur_compactions_reserved_size_.
-      // Otherwise, compactions will be double counted.
-      in_progress_files_size_ += file_size;
-      in_progress_files_.insert(file_path);
-    }
```
ltamasi, line 460: Unless I'm missing something, I think we could remove `in_progress_files_` and `in_progress_files_size_` altogether from this class (i.e. even the member variables, as well as any code that touches them). I believe this was the only place where files could get added to these structures, so if this is dead code (due to `compaction` always being `false`), so is any other chunk of code that queries the map or deletes entries etc.

ltamasi (commented):
Thanks @akankshamahajan15 ! Looks really nice in general, just some minor comments.

file/sst_file_manager_impl.cc
akankshamahajan15: ```in_progress_files_size_``` is used during compaction to check if there is enough headroom for compaction by db_impl_compaction_flush.cc. So I couldn't remove it at other places. https://github.com/facebook/rocksdb/blob/1a343bc393b5328c9825896a20fc750fb4de3e1c/file/sst_file_manager_impl.cc#L202

akankshamahajan15: But since in_progress_files_size_ is always zero, i guess we can remove this. Let me check again.

ltamasi: > `in_progress_files_size_` is used during compaction to check if there is enough headroom for compaction by db_impl_compaction_flush.cc. So I couldn't remove it at other places.

The way I read the code, `in_progress_files_size_` is always zero, since it could only be incremented if someone called `OnAddFile` with `compaction==true`, which doesn't happen, right?

If so, any code like `needed_headroom -= in_progress_files_size_` is a no-op.

ltamasi: P.S. I'm bringing this up because if for some reason it was necessary to track the "in progress" files of compactions, we would need to _extend_ this logic to cover blobs. But hopefully, we can just ditch it altogether :)

akankshamahajan15: Yes, you are right. It will always be zero and we won't add anything to the map so we can remove the variables altogether  from the code. I will update the code.

ltamasi: > But since in_progress_files_size_ is always zero, i guess we can remove this. Let me check again.

Yep, and `in_progress_files_` is supposed to be always empty, so any lookups in it would be unsuccessful.

db/db_impl/db_impl_open.cc
akankshamahajan15: Can you explain why the blob files won't be there in map?  It gets all the existing files by calling env->GetChildren.

db/blob/blob_file_completion_callback.h
akankshamahajan15: ```IsMaxAllowedSpaceReached``` is part of ```SstFileManager``` but ```OnAddFile``` is not. So I had to convert it to ```SstFileManagerImpl``` instance.

db/db_impl/db_impl_open.cc
ltamasi: Right, sorry, I misread the code

ltamasi:
One final thing: as we discussed, failed flushes immediately clean up after themselves here: https://github.com/facebook/rocksdb/blob/master/db/builder.cc#L330-L346 .

Now, when this piece of logic is triggered, we haven't yet notified `sfm` of the SST; however, we _have_ notified it of any blob files that might have been created. So here, we would need to call `OnDeleteFile` for the blob files in `blob_file_paths` as we iterate. (If the deletion was successful that is, which would also enable us to get rid of the `PermitUncheckedError` there.)

ltamasi (approved):
Awesome, thanks once again @akankshamahajan15 !

db/blob/blob_file_completion_callback.h
```diff
@@ -0,0 +1,56 @@
...
+    Status s;
+
+#ifndef ROCKSDB_LITE
+    auto sfm = static_cast<SstFileManagerImpl*>(sst_file_manager_);
+    if (sfm) {
+      // Report new blob files to SstFileManagerImpl
+      // TODO (PR7798).  We should only add the file to the FileManager if it
+      // exists. Otherwise, some tests may fail.  Ignore the error in the
+      // interim.
+      sfm->OnAddFile(file_name).PermitUncheckedError();
```
ltamasi, line 36: Just curious, under what conditions would `OnAddFile` fail here? (When would we call `OnAddFile` for a non-existent file?)

db/db_impl/db_impl.cc
```diff
@@ -511,6 +513,11 @@ Status DBImpl::CloseHelper() {
   }
   mutex_.Unlock();
 
+  // Below check is added as recovery_error_ is not checked and it causes crash
+  // in DBSSTTest.DBWithMaxSpaceAllowedWithBlobFiles when space limit is
+  // reached.
+  error_handler_.GetRecoveryError().PermitUncheckedError();
+
```
ltamasi, line 520: Could you please also elaborate on this scenario?

ltamasi (approved):

db/blob/blob_file_completion_callback.h
akankshamahajan15: I don't think we will call for non existent file but this TODO has been added at other places when call ```OnAddFile``` so to make it consistent I added here. But I can remove this TODO if it is not required here.

db/db_impl/db_impl.cc
akankshamahajan15: I was getting the error mentioned here.  https://github.com/facebook/rocksdb/pull/7715#issuecomment-742460190. After going through the comments, I added the PermitUncheckedError() here. But I am not sure if there is a better place to add the PermitUncheckedError.

db/blob/blob_file_completion_callback.h
ltamasi: I _think_ that the reason this is done elsewhere is because `BuildTable` might end up "building" an empty table, in which case it returns `Status::OK()` but there's no file to add to the `sfm`. (Checking this condition would most likely be the right fix for those TODOs BTW.)

This scenario is not possible for `BlobFileCompletionCallback` (since it's only called if a file was in the process of being written), so I think it would be better to go back to `s = sfm->OnAddFile(file_name);` here.

akankshamahajan15: Ah Makes Sense. Thanks!

db/db_impl/db_impl.cc
ltamasi: Took an admittedly brief look at the `ErrorHandler` code. I think the place that's missing a `PermitUncheckedError` call is in `RecoverFromRetryableBGIOError` when `recovery_error_` is reset to `Status::OK` (on line 653).

ltamasi: ... then again, I think that could only be triggered by using the other `SetBGError` overload that takes an `IOStatus`.

akankshamahajan15: Yesterday, I added the ```PermitUncheckedError``` at line https://github.com/facebook/rocksdb/blob/3dff28cf9b12d5bb2ac8bf339c2f635d57c77a1c/db/error_handler.cc#L242 and it ran succesfully so I added in the ```CloseHelper```.  I tried at Line 653 but that didn't work.

PR #8044, opened 2021-03-08
Deflake DBTest2.PartitionedIndexUserToInternalKey on ppc64le

ajkr:
For some reason I still cannot figure out, the manual flush in this test
was sometimes producing a third tiny file. I saw it a bunch of times on
ppc64le, but even running a qemu system with that architecture (and
playing with various other options) could not repro. However we did get
an instrumented Travis run to confirm the problem is indeed a third tiny
file - https://travis-ci.org/github/facebook/rocksdb/jobs/761986592. We
can avoid it by filling memtables less full and using manual flush.

### DIFF
--- a/db/db_test2.cc
+++ b/db/db_test2.cc
@@ -154,8 +154,14 @@ class PartitionedIndexTestListener : public EventListener {
 };
 
 TEST_F(DBTest2, PartitionedIndexUserToInternalKey) {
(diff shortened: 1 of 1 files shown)


akankshamahajan15 (approved):
Thanks a lot for debugging and fixing it!

ajkr:
Thanks for the review!

PR #8045, opened 2021-03-09
Add request_id in IODebugContext.

akankshamahajan15:
Summary: Add request_id in IODebugContext which will be populated by
    underlying FileSystem for IOTracing purposes. Update IOTracer to trace
    request_id in the tracing records. Provided API
    IODebugContext::SetRequestId which will set the request_id and enable
    tracing for request_id. The API hides the implementation and underlying
    file system needs to call this API directly.

Update DB::StartIOTrace API and remove redundant Env* from the
    argument as its not used and DB already has Env that is passed down to
    IOTracer.

Test Plan: Update unit test.

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -15,6 +15,7 @@
 
 ### Public API change
 * Added `TableProperties::slow_compression_estimated_data_size` and `TableProperties::fast_compression_estimated_data_size`. When `ColumnFamilyOptions::sample_for_compression > 0`, they estimate what `TableProperties::data_size` would have been if the "fast" or "slow" (see `ColumnFamilyOptions::sample_for_compression` API doc for definitions) compression had been used instead.
+* Update DB::StartIOTrace and remove Env object from the arguments as its redundant and DB already has Env object that is passed down to IOTracer::StartIOTrace
 
 ## 6.19.0 (03/21/2021)
 ### Bug Fixes
--- a/include/rocksdb/file_system.h
+++ b/include/rocksdb/file_system.h
@@ -132,10 +132,34 @@ struct IODebugContext {
   // To be set by the FileSystem implementation
   std::string msg;
 
+  // To be set by the underlying FileSystem implementation.
+  std::string request_id;
+
+  // In order to log required information in IO tracing for different
+  // operations, Each bit in trace_data stores which corresponding info from
+  // IODebugContext will be added in the trace. Foreg, if trace_data = 1, it
+  // means bit at position 0 is set so TraceData::kRequestID (request_id) will
+  // be logged in the trace record.
+  //
+  enum TraceData : char {
+    // The value of each enum represents the bitwise position for
+    // that information in trace_data which will be used by IOTracer for
+    // tracing. Make sure to add them sequentially.
+    kRequestID = 0,
+  };
+  uint64_t trace_data = 0;
+
   IODebugContext() {}
 
   void AddCounter(std::string& name, uint64_t value) {
     counters.emplace(name, value);
+  }
+
+  // Called by underlying file system to set request_id and log request_id in
+  // IOTracing.
+  void SetRequestId(const std::string& _request_id) {
+    request_id = _request_id;
+    trace_data |= (1 << TraceData::kRequestID);
   }
 
   std::string ToString() {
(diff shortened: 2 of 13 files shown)


include/rocksdb/file_system.h
```diff
@@ -132,6 +132,25 @@ struct IODebugContext {
...
+  // To be set by the underlying FileSystem implementation.
+  std::string request_id;
+
+  // In order to log required information in IO tracing for different
+  // operations, Each bit in trace_data stores which corresponding info from
+  // IODebugContext will be added in the trace. Foreg, if trace_data = 1, it
+  // means bit at position 0 is set so TraceOp::kRequestID (request_id) will be
+  // logged in the trace record.
+  //
+  // In order to log the required info, trace_data needs to be updated as
```
anand1976, line 144: Why not just log whatever's in request_id? If the FS implementation didn't set it, it'll be an empty string.

akankshamahajan15: Yes, I can add it like filename. It will be empty string if not set. I thought we don't want empty string in trace so I added the logic. I will make the request_id default.

anand1976: That's a  good point. What would the trace look like with and without empty string (I assume there will be some delimiters)? It might not be a bad thing to leave it there if it makes the trace easier to parse.

akankshamahajan15: It will be something like ```"RequestID:   , FileOperation: Append"```. Since its a string, I can add a check at the parser side as well to  add request id only if it's not empty, if that's fine.

akankshamahajan15: But in future, if WS want log additional information of type int then we might need to switch back to this approach.

anand1976: > It will be something like `"RequestID: , FileOperation: Append"`. Since its a string, I can add a check at the parser side as well to add request id only if it's not empty, if that's fine.

Yeah, either checking on the parser side, or providing a setter in ```IODebugContext``` might be better as it'll hide the implementation details from WS.

HISTORY.md
```diff
@@ -12,6 +12,9 @@
 ### Performance Improvements
 * On ARM platform, use `yield` instead of `wfe` to relax cpu to gain better performance.
 
+### Public API change
+* Update DB::StartIOTrace and remove Env object from the arguments as its redundant and DB already has Env object that is passed down to IOTracer::StartIOTrace
```
anand1976, line 16: The env is passed when constructing ```TraceWriter``` right?

include/rocksdb/file_system.h
```diff
@@ -132,6 +132,14 @@ struct IODebugContext {
   // To be set by the FileSystem implementation
   std::string msg;
 
+  // To be set by the underlying FileSystem implementation.
+  std::string request_id;
+
+  // To be set by Warm Storage if underlying filesystem is warm storage.
+  // This bool is used during IO tracing to add additional information in trace
+  // records.
+  bool is_ws = 0;
```
anand1976, line 141: Warm Storage is an FB intetrnal component. We don't want to pollute RocksDB code with references to such components, and even any specific file system for that matter. Is this to decide whether to add ```request_id``` to the trace? Perhaps we can call this flag ```fs_info``` or something generic, and provide a ```SetRequestID()``` function that will hide these details.

HISTORY.md
akankshamahajan15: Yes. And since DB already has ```env_``` so that is passed down to TraceWriter.

include/rocksdb/file_system.h
akankshamahajan15: @anand1976 I was wondering should I add the logic I added in first PR and hide the implementation in SetRequestID? That way underlying FileSystem had to call SetRequestId only and i can add the logic of whether to trace that data inside that function.

anand1976 (approved):
Thanks for addressing the comments. LGTM.

include/rocksdb/file_system.h
anand1976: I think the bitwise logic is fine, since we're hiding it anyway.

PR #8049, opened 2021-03-10
Separate handling of WAL Sync io error with SST flush io error

zhichao-cao:
In previous codebase, if WAL is used, all the retryable IO Error will be treated as hard error. So write is stalled. In this PR, the retryable IO error from WAL sync is separated from SST file flush io error. If WAL Sync is ok and retryable IO Error only happens during SST flush, the error is mapped to soft error. So user can continue insert to Memtable and append to WAL. 

Resolve the bug that if WAL sync fails, the memtable status does not roll back due to calling PickMemtable early than calling and checking SyncClosedLog.

test plan: added new unit test, make check

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -171,7 +171,7 @@ Status DBImpl::FlushMemTableToOutputFile(
 #endif  // ROCKSDB_LITE
 
   Status s;
-  bool pick_status = false;
+  bool need_cancel = false;
   IOStatus log_io_s = IOStatus::OK();
   if (logfile_number_ > 0 &&
       versions_->GetColumnFamilySet()->NumberOfColumnFamilies() > 1) {
@@ -196,7 +196,7 @@ Status DBImpl::FlushMemTableToOutputFile(
   TEST_SYNC_POINT("DBImpl::FlushMemTableToOutputFile:BeforePickMemtables");
   if (s.ok()) {
     flush_job.PickMemTable();
-    pick_status = true;
+    need_cancel = true;
   }
   TEST_SYNC_POINT("DBImpl::FlushMemTableToOutputFile:AfterPickMemtables");
 
@@ -208,9 +208,10 @@ Status DBImpl::FlushMemTableToOutputFile(
   // is unlocked by the current thread.
   if (s.ok()) {
     s = flush_job.Run(&logs_with_prep_tracker_, &file_meta);
+    need_cancel = false;
   }
 
-  if (!s.ok() && pick_status) {
+  if (!s.ok() && need_cancel) {
     flush_job.Cancel();
   }
   IOStatus io_s = IOStatus::OK();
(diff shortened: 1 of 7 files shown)


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -252,23 +259,25 @@ Status DBImpl::FlushMemTableToOutputFile(
       // be pessimistic and try write to a new MANIFEST.
       // TODO: distinguish between MANIFEST write and CURRENT renaming
       if (!versions_->io_status().ok()) {
-        if (total_log_size_ > 0) {
+        if (total_log_size_ > 0 || !log_io_s.ok()) {
```
anand1976, line 262: Not clear why we need to check ```log_io_s``` here and on line 270. If ```SyncClosedLogs``` failed, we shouldn't see SST or manifest write error right, since those wouldn't even be attempted?

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -169,17 +164,15 @@ Status DBImpl::FlushMemTableToOutputFile(
...
-
 #ifndef ROCKSDB_LITE
   // may temporarily unlock and lock the mutex.
   NotifyOnFlushBegin(cfd, &file_meta, mutable_cf_options, job_context->job_id);
 #endif  // ROCKSDB_LITE
 
   Status s;
-  IOStatus io_s = IOStatus::OK();
+  bool pick_status = false;
+  bool exec_status = false;
```
anand1976, line 174: It feels a bit confusing. How about just have a single ```bool need_cancel``` that's set when ```PickMemTable``` is called and reset when ```Run``` is called?

zhichao-cao: Oh sure.

db/db_impl/db_impl_compaction_flush.cc
zhichao-cao: Is it better to add an assert(log_io_s.ok()) here inside the if to make sure the logic is correct?

anand1976: Yes, I think so.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -658,23 +685,25 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
       // be pessimistic and try write to a new MANIFEST.
       // TODO: distinguish between MANIFEST write and CURRENT renaming
       if (!versions_->io_status().ok()) {
-        if (total_log_size_ > 0) {
+        if (total_log_size_ > 0 || !log_io_s.ok()) {
```
anand1976, line 688: Same question here as before - when would ```versions_->io_status()``` and ```log_io_s``` both be non-ok? If ```!log_io_s.ok()```, a background error would have already been reported on line 437.

anand1976 (approved):
LGTM. Just had one minor comment.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -474,6 +500,7 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
     s = error_status.ok() ? s : error_status;
   }
 
+  IOStatus io_s = IOStatus::OK();
```
riversand963, line 503: Maybe I am missing something.
If we have line 503 setting io_s to OK, then line 504 will always evaluate to true. Is this intended? @zhichao-cao ?

PR #8051, opened 2021-03-12
Instantiate tests DBIteratorTestForPinnedData

pdillinger:
Summary: a trial gtest upgrade discovered some parameterized tests missing instantiation. By some miracle, they still pass.

Test Plan: thisisthetest


mrambacher (approved):
LGTM.

My only question is if the "BackwardIterationOnInplaceUpdateMemtable" test should also be parameterized to do the ReadCallback.

pdillinger:
> My only question is if the "BackwardIterationOnInplaceUpdateMemtable" test should also be parameterized to do the ReadCallback.

The same question I think applies to all the other TEST_F in the file, and I don't know the answer. I don't know what the ReadCallback is doing.

PR #8061, opened 2021-03-15
Add user-defined timestamps to db_stress

riversand963:
Add some basic test for user-defined timestamp to db_stress. Currently,
read with timestamp always tries to read using the current timestamp.
Due to the per-key timestamp-sequence ordering constraint, we only add timestamp-
related tests to the `NonBatchedOpsStressTest` since this test serializes accesses
to the same key and uses a file to cross-check data correctness.
The timestamp feature is not supported in a number of components, e.g. Merge, SingleDelete,
DeleteRange, CompactionFilter, Readonly instance, secondary instance, SST file ingestion, transaction,
etc. Therefore, db_stress should exit if user enables both timestamp and these features at the same
time. The (currently) incompatible features can be found in
`CheckAndSetOptionsForUserTimestamp`.

This PR also fixes a bug triggered when timestamp is enabled together with
`index_type=kBinarySearchWithFirstKey`. This bug fix will also be in another separate PR
with more unit tests coverage. Fixing it here because I do not want to exclude the index type
from crash test.

Test plan:
make crash_test_with_ts

### DIFF
--- a/db_stress_tool/db_stress_common.cc
+++ b/db_stress_tool/db_stress_common.cc
@@ -30,7 +30,7 @@ enum ROCKSDB_NAMESPACE::ChecksumType checksum_type_e =
     ROCKSDB_NAMESPACE::kCRC32c;
 enum RepFactory FLAGS_rep_factory = kSkipList;
 std::vector<double> sum_probs(100001);
-int64_t zipf_sum_size = 100000;
+constexpr int64_t zipf_sum_size = 100000;
 
 namespace ROCKSDB_NAMESPACE {
 
@@ -232,6 +232,15 @@ size_t GenerateValue(uint32_t rand, char* v, size_t max_sz) {
   v[value_sz] = '\0';
   return value_sz;  // the size of the value set.
 }
+
+std::string NowNanosStr() {
+  uint64_t t = db_stress_env->NowNanos();
+  std::string ret;
+  PutFixed64(&ret, t);
+  return ret;
+}
+
+std::string GenerateTimestampForRead() { return NowNanosStr(); }
 
 namespace {
 
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -2606,7 +2606,7 @@ void StressTest::CheckAndSetOptionsForUserTimestamp() {
             "disabled.\n");
     exit(1);
   }
-  if (FLAGS_use_merge) {
+  if (FLAGS_use_merge || FLAGS_use_full_merge_v1) {
     fprintf(stderr, "Merge not supported.\n");
     exit(1);
   }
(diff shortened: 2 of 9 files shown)


riversand963:
There may be other incompatible features, and will be added to the check in `CheckAndSetOptions()`. Features that start to support timestamp will be removed from the list.

db_stress_tool/db_stress_test_base.cc
```diff
@@ -765,6 +776,20 @@ void StressTest::OperateDb(ThreadState* thread) {
...
+      // Assign timestamps if necessary.
+      std::string read_ts_str;
+      std::string write_ts_str;
+      Slice read_ts;
+      Slice write_ts;
+      if (ShouldAcquireMutexOnKey() && FLAGS_user_timestamp_size > 0) {
+        read_ts_str = GenerateTimestampForRead();
+        read_ts = read_ts_str;
+        read_opts.timestamp = &read_ts;
+        write_ts_str = NowNanosStr();
```
jay-zhuang, line 788: Maybe with NowSecondsStr (or even Minutes) could cover more cases with equal timestamps. Worth to have the option for that.

db_stress_tool/db_stress_common.cc
```diff
@@ -233,6 +233,15 @@ size_t GenerateValue(uint32_t rand, char* v, size_t max_sz) {
...
 }
 
+std::string NowNanosStr() {
+  uint64_t t = db_stress_env->NowNanos();
+  std::string ret;
+  PutFixed64(&ret, t);
+  return ret;
+}
+
+std::string GenerateTimestampForRead() { return NowNanosStr(); }
```
jay-zhuang, line 243: Should db_stress just use `NowNanoStr()` instead?

db_stress_tool/db_stress_test_base.cc
```diff
@@ -2491,5 +2571,72 @@ void StressTest::Reopen(ThreadState* thread) {
           clock_->TimeToString(now / 1000000).c_str(), num_times_reopened_);
   Open();
 }
+
+void StressTest::CheckAndSetOptionsForUserTimestamp() {
```
jay-zhuang, line 2575: Will it cause db_stress likely to fail? Should we skip the un-supported options in `db_crashtest.py`?

jay-zhuang (approved):
LGTM. 👍
Assuming you're going to setup the jobs to run the tests.

db_stress_tool/db_stress_test_base.cc
riversand963: This is already done.

db_stress_tool/db_stress_test_base.cc
riversand963: For single db instance running on a single host (as db_stress), it is relatively easy to guarantee monotonically increasing timestamps, assuming we 1) have a sufficiently accurate clock, e.g. one that returns nano secs. 2) no faulty hardware, 3) each db_stress process restarts and delays to wait out max possible clock skew, then I think we are fine here.
We already have 1), we assume 2) in this test, and 3) may be done later.

db_stress_tool/db_stress_common.cc
riversand963: I imagine we will in the future add an option `-read_with_latest_ts`. Currently, we always read from latest ts. In the future, this can be set to false and we will get a random ts for read. Therefore, I think it is better to put the entire logic in a function `GenerateTimestampForRead()`.

riversand963:

PR #8062, opened 2021-03-15
Fix a bug in key comparison when index type is kBinarySearchWithFirstKey

riversand963:
When timestamp is enabled, key comparison should take this into account.
In `BlockBasedTableReader::Get()`, `BlockBasedTableReader::MultiGet()`,
assume the target key is `key`, and the timestamp upper bound is `ts`.
The highest key in current block is (key, ts1), while the lowest key in next
block is (key, ts2).
If
```
ts1 > ts > ts2
```
then
```
(key, ts1) < (key, ts) < (key, ts2) 
```
It can be shown that if `Compare()` is used, then we will mistakenly skip the next
block. Instead, we should use `CompareWithoutTimestamp()`.

The majority of this PR makes some existing tests in `db_with_timestamp_basic_test.cc`
parameterized so that different index types can be tested. A new unit test is
also added for more coverage.

Test plan
make check

### DIFF
--- a/db/db_with_timestamp_basic_test.cc
+++ b/db/db_with_timestamp_basic_test.cc
@@ -504,6 +504,75 @@ INSTANTIATE_TEST_CASE_P(
         BlockBasedTableOptions::IndexType::kTwoLevelIndexSearch,
         BlockBasedTableOptions::IndexType::kBinarySearchWithFirstKey));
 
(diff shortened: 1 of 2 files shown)


ltamasi (approved):
LGTM, thanks for the fix @riversand963 !

riversand963:
> LGTM, thanks for the fix @riversand963 !

Thanks @ltamasi for the quick review!

PR #8064, opened 2021-03-15
Revamp WriteController

pdillinger:
Summary: WriteController had a number of issues:
* It could introduce a delay of 1ms even if the write rate never exceeded the
configured delayed_write_rate.
* The DB-wide delayed_write_rate could be exceeded in a number of ways
with multiple column families:
  * Wiping all pending delay "debts" when another column family joins
  the delay with GetDelayToken().
  * Resetting last_refill_time_ to (now + sleep amount) means each
  column family can write with delayed_write_rate for large writes.
  * Updating bytes_left_ for a partial refill without updating
  last_refill_time_ would essentially give out random bonuses,
  especially to medium-sized writes.

Now the code is much simpler, with these issues fixed. See comments in
the new code and new (replacement) tests.

Test Plan: new tests, better than old tests


db/write_controller.cc
```diff
@@ -62,49 +60,39 @@ uint64_t WriteController::GetDelay(SystemClock* clock, uint64_t num_bytes) {
...
+  if (next_refill_time_ == 0) {
+    // Start with an initial allotment of bytes for one interval
+    next_refill_time_ = time_now;
+  }
+  if (next_refill_time_ <= time_now) {
+    // Refill based on time interval plus any extra elapsed
+    uint64_t elapsed = time_now - next_refill_time_ + kRefillInterval;
+    bytes_left_ += static_cast<uint64_t>(
+        1.0 * elapsed / kMicrosPerSecond * delayed_write_rate_ + 0.999999);
+    next_refill_time_ = time_now + kRefillInterval;
```
mrambacher, line 75: A couple of things here look wrong to me and could use an explanation.  time_now and next_refill_time are in time units but the refill interval is in bytes (1K).  This leads to a funky value for elapsed (is it bytes or time?).

Then a few lines later next_refill_time is also updated based on 1K, which is effectively upping the time by a millisecond I think.

HISTORY.md
```diff
@@ -2,6 +2,7 @@
 ## Unreleased
 ### Bug Fixes
 * Fixed the truncation error found in APIs/tools when dumping block-based SST files in a human-readable format. After fix, the block-based table can be fully dumped as a readable file.
+* Made `delayed_write_rate` more strictly enforced, except now avoiding unnecessary initial delays. Specifically, writes are only delayed if during elapsed time t they exceed the allowance for t + 1 millisecond.
```
mrambacher, line 5: The first sentence is a little awkward, I think you mean "now avoiding" (no except).

mrambacher (commented):
This code is confusing to me.  I wonder if life would be simpler if:
-> The WriteController was created with a SystemClock
-> When a DelayToken is requested,  
 - If this is the first one, bytes_left is set to 0 and the "last_refill_time" is now
 - If this is not the first one, update the bytes left from the current rate and the "last refill time" and then update the last refill time

In GetDelay, the bytes left is the elapsed time (now - last refill) * rate.  Bytes left and last refill would also need to be updated.

db/write_controller.cc
pdillinger: kRefillInterval is time everywhere. This hasn't changed. I don't know why it was chosen to be 1.024 ms, but I didn't find a bug in that so I didn't change it. I can add a clarifying comment.

The code author is unavailable atm.

HISTORY.md
pdillinger: How about this:

> When hitting a write slowdown condition, no write delay (previously 1 millisecond) is imposed until `delayed_write_rate` is actually exceeded, with an initial burst allowance of 1 millisecond worth of bytes. Also, beyond the initial burst allowance, `delayed_write_rate` is now more strictly enforced, especially with multiple column families.

pdillinger:
> In GetDelay, the bytes left is the elapsed time (now - last refill) * rate. Bytes left and last refill would also need to be updated.

How do you track the debt accumulated by outstanding writes in multiple column families? Negative bytes left, or "last refill" in the future?

db/write_controller.cc
```diff
@@ -20,10 +21,11 @@ std::unique_ptr<WriteControllerToken> WriteController::GetStopToken() {
 
 std::unique_ptr<WriteControllerToken> WriteController::GetDelayToken(
     uint64_t write_rate) {
-  total_delayed_++;
-  // Reset counters.
-  last_refill_time_ = 0;
-  bytes_left_ = 0;
+  if (0 == total_delayed_++) {
+    // Starting delay, so reset counters.
+    next_refill_time_ = 0;
```
mrambacher, line 26: Should next_refill_time be reset every time?  Here is my logic:
- When you start the first time, there is no time (t0) and no bytes left, so both are 0 and the delay is X
- If you start another delay at t1 with rate Y, then the write rate between t0 and t1 should be X but the write rate from t1 forward should be Y.

So ideally you would like to recalculate the number of bytes left on the second delay (but cannot since you do not have a SystemClock) and tell the "GetDelay" to recalculate on the next call as well

db/write_controller.cc
```diff
@@ -62,49 +61,39 @@ uint64_t WriteController::GetDelay(SystemClock* clock, uint64_t num_bytes) {
...
-    bytes_left_ = bytes_left_ + single_refill_amount - num_bytes;
-    last_refill_time_ = time_now + kRefillInterval;
-    return kRefillInterval + sleep_debt;
-  }
+  // We need to delay to avoid exceeding write rate.
+  assert(num_bytes > bytes_left_);
+  uint64_t bytes_over_budget = num_bytes - bytes_left_;
+  uint64_t needed_delay = static_cast<uint64_t>(
+      bytes_over_budget / static_cast<long double>(delayed_write_rate_) *
+      kMicrosPerSecond);
```
mrambacher, line 90: Is the math simpler as "kMicrosPerSecond * bytes_over_budget / delated_write_rate_?  Would you still need the casts to double?

db/write_controller.h
```diff
@@ -28,7 +28,7 @@ class WriteController {
         total_delayed_(0),
         total_compaction_pressure_(0),
         bytes_left_(0),
```
mrambacher, line 30: It might just be me, but bytes_left is a bad name.  It is NOT the number of bytes_left to send but the BUDGET of bytes that can still be sent at the given rate.  At the minimum it merits a clarifying comment but a name change would also be helpful (byte_budget_left_?)

db/write_controller_test.cc
```diff
@@ -26,115 +27,206 @@ class WriteControllerTest : public testing::Test {
...
-  ASSERT_EQ(static_cast<uint64_t>(500000),
-            controller.GetDelay(clock_.get(), 20000000u));
+
+  // Verify no delay because write rate has not been exceeded within
+  // refill interval.
+  EXPECT_EQ(0U, controller.GetDelay(clock_.get(), 2000u /*bytes*/));
+  EXPECT_EQ(0U, controller.GetDelay(clock_.get(), 2000u /*bytes*/));
+
+  // Allow refill
+  clock_->now_micros_ += 1024;
```
mrambacher, line 113: Since it is an odd number, can you please comment this is the kRefillInterval?

db/write_controller_test.cc
```diff
@@ -26,115 +27,206 @@ class WriteControllerTest : public testing::Test {
...
-TEST_F(WriteControllerTest, ChangeDelayRateTest) {
-  WriteController controller(40000000u);  // also set max delayed rate
-  controller.set_delayed_write_rate(10000000u);
+// Make tests easier to read
+#define MILLION *1000000u
+#define MB MILLION
+#define MBPS MILLION
+#define SECS MILLION  // in microseconds
+
+TEST_F(WriteControllerTest, BasicAPI) {
```
mrambacher, line 36: Is/Should there be a test that:
- Validates that GetDelay returns 0 if there are stop tokens
- Validates that GetDelay returns 0 if there are no delay tokens
- Cancels a stop token and checks the delay

db/write_controller.cc
pdillinger: Arithmetic overflow on 18TB write batch. Probably OK. 👍

db/db_test.cc
```diff
@@ -247,17 +247,19 @@ TEST_F(DBTest, SkipDelay) {
       wo.sync = sync;
       wo.disableWAL = disableWAL;
       wo.no_slowdown = true;
-      dbfull()->Put(wo, "foo", "bar");
+      // Large enough to exceed allowance for one time interval
+      std::string large_value(1024, 'x');
+      dbfull()->Put(wo, "foo", large_value).PermitUncheckedError();
```
riversand963, line 252: Should we check the result of `Put()`?

db/write_controller.cc
```diff
@@ -20,10 +21,11 @@ std::unique_ptr<WriteControllerToken> WriteController::GetStopToken() {
 
 std::unique_ptr<WriteControllerToken> WriteController::GetDelayToken(
     uint64_t write_rate) {
-  total_delayed_++;
-  // Reset counters.
-  last_refill_time_ = 0;
-  bytes_left_ = 0;
+  if (0 == total_delayed_++) {
```
riversand963, line 24: Not an expert at this code, and I am also trying to reverse-engineering what it does.
One question: after a new super version is installed, we may call `SetupDelay()` in which a delay token is acquired via `GetDelayToken()` and the internal `total_delayed_` will become 1. Later, the first writer comes in. 
It will still see `NeedsDelay()` as true, and will enter `DelayWrite()`. Inside `DelayWrite()`, when it reaches
```
while (write_controller_.NeedsDelay()) {
...
// sleep for 1ms
}
```
It seems it will still sleep.
Maybe I am missing something...

riversand963 (commented):
Thanks @pdillinger for working on this.

db/write_controller.cc
pdillinger: That should never be reached if GetDelay returns 0, which we try to do. See e.g. the `StartFilled` test. :)

db/db_test.cc
pdillinger: My understanding is that it would be nice if the first large write triggered a delay, but for now it requires a second write.

pdillinger: I can check for that and make the comment more explicit what might be OK or GOOD progress in the case of failure.

db/write_controller.cc
pdillinger: I don't think it's worth trying to be that accurate about exactly when different rates are applied. Keep it simple(-ish).

db/write_controller_test.cc
pdillinger: I believe I have found a reason to change this now.

In DBImpl::DelayWrite, the delay is implemented in 1ms increments. But WriteController's minimum delay interval is 1.024ms. This is silly and seems backwards. It means we are very often going to wake up right before we've finished the delay, then sleep again, totaling almost twice as long as requested (in the minimum case).

I think I'll use 1.000 ms in WriteController and 1.001 ms in DelayWrite.

db/write_controller.cc
pdillinger: Actually, float division is faster than (large) scalar division. We don't need super accuracy float is fine.

db/write_controller.cc
riversand963: My bad. I missed the `if (delay > 0)` part just before that.

db/write_controller.cc
mrambacher: A comment about ignoring this condition might be useful for future reference.

mrambacher (approved):
LGTM.  Thanks for the cleanup!

PR #8068, opened 2021-03-16
Include C++ standard library headers instead of C compatibility headers

storagezhang:


riversand963 (commented):
Thanks @storagezhang for the PR.
Could you update the title? I find it a little confusing.
Per https://en.cppreference.com/w/cpp/header, `cstdio`, `cassert`, etc. are std C++ headers which are recommended by c++ standard. It is the C-compatible headers, such as stdio.h, stdlib.h, assert.h that we want to replace.
How about "Include C++ standard library headers instead of C compatibility headers", or something like that?

storagezhang:
@riversand963 Sorry about that. I updated the title.

riversand963 (approved):
Thanks @storagezhang

PR #8069, opened 2021-03-17
Begin forward compatibility for new backup meta schema

pdillinger:
Summary: This does not add any new public APIs or published
functionality, but adds the ability to read and use (and in tests,
write) backups with a new meta file schema, based on the old schema
but not forward-compatible (before this change). The new schema enables
some capabilities not in the old:

* Explicit versioning, so that users get clean error messages the next
time we want to break forward compatibility.
* Ignoring unrecognized fields (with warning), so that new non-critical
features can be added without breaking forward compatibility.
* Rejecting future "non-ignorable" fields, so that new features critical
to some use-cases could potentially be added outside of linear schema
versions, with broken forward compatibility.
* Fields at the end of the meta file, such as for checksum of the meta
file's contents (up to that point)
* New optional 'size' field for each file, which is checked when present
* Optionally omitting 'crc32' field, so that we aren't required to have
a crc32c checksum for files to take a backup. (E.g. to support backup
via hard links and to better support file custom checksums.)

Because we do not have a JSON parser and to share code, the new schema
is simply derived from the old schema.

BackupEngine code is updated to allow missing checksums in some places,
and to make that easier, `has_checksum` and `verify_checksum_after_work`
are eliminated. Empty `checksum_hex` indicates checksum is unknown. I'm
not too afraid of regressing on data integrity, because
(a) we have pretty good test coverage of corruption detection in backups, and
(b) we are increasingly relying on the DB itself for data integrity rather than
it being an exclusive feature of backups.

Test Plan: new unit tests, added to crash test (some local run with
boosted backup probability)

### DIFF
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -906,6 +906,7 @@ Status BackupEngineImpl::Initialize() {
                                      work_item.dst_path + ": " + checksum_info);
             }
           } else {
+            // FIXME(peterd): dead code?
             std::string checksum_function_info(
                 "Existing checksum function is " +
                 work_item.src_checksum_func_name +
@@ -2236,7 +2237,9 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
     }
   }
 
-  int schema_major_version = 1;  // implicit default
+  // If we don't read an explicit schema_version, that implies version 1,
+  // which is what we call the original backup meta schema.
+  int schema_major_version = 1;
 
   // Failures handled at the end
   std::string line;
@@ -2292,7 +2295,7 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
                                   field_name + " (from future version?)");
     } else {
       // Warn the first time we see any particular unrecognized meta field
-      if (reported_ignored_fields->insert(field_name + "1").second) {
+      if (reported_ignored_fields->insert("meta:" + field_name).second) {
         ROCKS_LOG_WARN(info_log, "Ignoring unrecognized backup meta field %s",
                        field_name.c_str());
       }
@@ -2357,7 +2360,7 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
       if (field_name == kFileCrc32cFieldName) {
         uint32_t checksum_value =
             static_cast<uint32_t>(strtoul(field_data.c_str(), nullptr, 10));
-        if (components[2] != ROCKSDB_NAMESPACE::ToString(checksum_value)) {
+        if (field_data != ROCKSDB_NAMESPACE::ToString(checksum_value)) {
           return Status::Corruption("Invalid checksum value for " + filename +
                                     " in " + meta_filename_);
         }
@@ -2375,7 +2378,7 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
                                     field_name + " (from future version?)");
       } else {
         // Warn the first time we see any particular unrecognized file field
-        if (reported_ignored_fields->insert(field_name + "2").second) {
+        if (reported_ignored_fields->insert("file:" + field_name).second) {
           ROCKS_LOG_WARN(info_log, "Ignoring unrecognized backup file field %s",
                          field_name.c_str());
         }
@@ -2400,7 +2403,8 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
       if (StartsWith(field_name, kNonIgnorableFieldPrefix)) {
         return Status::NotSupported("Unrecognized non-ignorable field " +
                                     field_name + " (from future version?)");
-      } else if (reported_ignored_fields->insert(field_name + "3").second) {
+      } else if (reported_ignored_fields->insert("footer:" + field_name)
+                     .second) {
         // Warn the first time we see any particular unrecognized footer field
         ROCKS_LOG_WARN(info_log,
                        "Ignoring unrecognized backup meta footer field %s",
--- a/utilities/backupable/backupable_db.cc
+++ b/utilities/backupable/backupable_db.cc
@@ -2147,7 +2147,26 @@ Status BackupEngineImpl::BackupMeta::Delete(bool delete_meta) {
   return s;
 }
 
-const std::string kAppMetaDataFieldName("metadata");
+// Constants for backup meta file schema (see LoadFromFile)
+namespace {
+
+const std::string kSchemaVersionPrefix{"schema_version "};
+const std::string kFooterMarker{"// FOOTER"};
+
+const std::string kAppMetaDataFieldName{"metadata"};
+
+// WART: The checksums are crc32c but named "crc32"
+const std::string kFileCrc32cFieldName{"crc32"};
+const std::string kFileSizeFieldName{"size"};
+
+// Marks a (future) field that should cause failure if not recognized.
+// Other fields are assumed to be ignorable. For example, in the future
+// we might add
+//  ni::file_name_escape uri_percent
+// to indicate all file names have had spaces and special characters
+// escaped using a URI percent encoding.
+const std::string kNonIgnorableFieldPrefix{"ni::"};
+}  // namespace
 
 // Each backup meta file is of the format (schema version 1):
 //----------------------------------------------------------
@@ -2204,7 +2223,9 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
     const std::unordered_map<std::string, uint64_t>& abs_path_to_size,
     Logger* info_log,
     std::unordered_set<std::string>* reported_ignored_fields) {
+  assert(reported_ignored_fields);
   assert(Empty());
+
   std::unique_ptr<LineFileReader> backup_meta_reader;
   {
     Status s =
@@ -2220,9 +2241,8 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
   // Failures handled at the end
   std::string line;
   if (backup_meta_reader->ReadLine(&line)) {
-    const std::string ver_prefix{"schema_version "};
-    if (StartsWith(line, ver_prefix)) {
-      std::string ver = line.substr(ver_prefix.size());
+    if (StartsWith(line, kSchemaVersionPrefix)) {
+      std::string ver = line.substr(kSchemaVersionPrefix.size());
       if (ver == "2" || StartsWith(ver, "2.")) {
         schema_major_version = 2;
       } else {
@@ -2267,10 +2287,9 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
     } else if (schema_major_version < 2) {
       return Status::Corruption("Expected number of files or \"" +
                                 kAppMetaDataFieldName + "\" field");
-    } else if (StartsWith(field_name, "ni_")) {
-      // ni_ short for "non-ignorable"
+    } else if (StartsWith(field_name, kNonIgnorableFieldPrefix)) {
       return Status::NotSupported("Unrecognized non-ignorable meta field " +
-                                  field_name);
+                                  field_name + " (from future version?)");
     } else {
       // Warn the first time we see any particular unrecognized meta field
       if (reported_ignored_fields->insert(field_name + "1").second) {
@@ -2288,7 +2307,7 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
       return Status::Corruption("Empty line instead of file entry.");
     }
     if (schema_major_version >= 2 && components.size() == 2 &&
-        components[0] == "//" && components[1] == "footer") {
+        line == kFooterMarker) {
       footer_present = true;
       break;
     }
@@ -2315,12 +2334,12 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
             "Bad number of line components for file entry.");
       }
     } else {
+      // Check restricted original schema
       if (components.size() < 3) {
         return Status::Corruption("File checksum is missing for " + filename +
                                   " in " + meta_filename_);
       }
-      // WART: The checksums are crc32c, not original crc32
-      if (components[1] != "crc32") {
+      if (components[1] != kFileCrc32cFieldName) {
         return Status::Corruption("Unknown checksum type for " + filename +
                                   " in " + meta_filename_);
       }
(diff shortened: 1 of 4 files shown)


utilities/backupable/backupable_db.cc
```diff
@@ -2159,84 +2215,182 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
     }
   }
 
+  int schema_major_version = 1;  // implicit default
+
   // Failures handled at the end
   std::string line;
   if (backup_meta_reader->ReadLine(&line)) {
+    const std::string ver_prefix{"schema_version "};
```
mrambacher, line 2223: Should this be a constexpr or a static?

utilities/backupable/backupable_db.cc
```diff
@@ -2135,19 +2147,63 @@ Status BackupEngineImpl::BackupMeta::Delete(bool delete_meta) {
...
+//   * "size" - the size of the file (new)
+// * Footer meta fields:
+//   * None yet (future use for meta file checksum anticipated)
+//
 Status BackupEngineImpl::BackupMeta::LoadFromFile(
     const std::string& backup_dir,
-    const std::unordered_map<std::string, uint64_t>& abs_path_to_size) {
+    const std::unordered_map<std::string, uint64_t>& abs_path_to_size,
+    Logger* info_log,
+    std::unordered_set<std::string>* reported_ignored_fields) {
```
mrambacher, line 2206: Do you need to assert(reported_ignored_fields)?

utilities/backupable/backupable_db.cc
```diff
@@ -2159,84 +2215,182 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
...
     }
 
-    if (components.size() < 3) {
-      return Status::Corruption("File checksum is missing for " + filename +
-                                " in " + meta_filename_);
+    if (schema_major_version >= 2) {
+      if (components.size() % 2 != 1) {
+        return Status::Corruption(
+            "Bad number of line components for file entry.");
+      }
```
mrambacher, line 2316: Is components.size() == 1 valid?

utilities/backupable/backupable_db.cc
```diff
@@ -2159,84 +2215,182 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
...
+        checksum_hex = ChecksumInt32ToHex(checksum_value);
+      } else if (field_name == "size") {
+        uint64_t ex_size =
+            std::strtoull(field_data.c_str(), nullptr, /*base*/ 10);
+        if (ex_size != actual_size) {
+          return Status::Corruption("For file " + filename + " expected size " +
+                                    ToString(ex_size) + " but found size" +
+                                    ToString(actual_size));
+        }
+      } else if (StartsWith(field_name, "ni_")) {
```
mrambacher, line 2355: I am not sure "ni_" is the best choice.  It seems a little opaque, but I guess this is internal.  If you want "ni_", would something like "ni::" be better, as it uses a (potentially) clearer delimiter

utilities/backupable/backupable_db_test.cc
```diff
@@ -2792,6 +2793,188 @@ TEST_F(BackupableDBTest, MetadataTooLarge) {
...
+
+  test_opts.file_fields["ni_123"] = "xyz";
+  TEST_EnableWriteFutureSchemaVersion2(backup_engine_.get(), test_opts);
+  ASSERT_OK(
+      backup_engine_->CreateNewBackupWithMetadata(db_.get(), app_metadata));
+  test_opts.file_fields.clear();
+
+  test_opts.footer_fields["ni_123"] = "xyz";
+  TEST_EnableWriteFutureSchemaVersion2(backup_engine_.get(), test_opts);
+  ASSERT_OK(
```
mrambacher, line 2897: Adding extra spaces (" field", "field ", " value", "value ", "ni_ field", etc) would be nice tests as well.  I do not know how they will behave...

utilities/backupable/backupable_db_impl.h
```diff
@@ -0,0 +1,29 @@
...
+  std::map<std::string, std::string> file_fields;
+  std::map<std::string, std::string> footer_fields;
+};
+
+// Modifies the BackupEngine(Impl) to write backup meta files using the
+// unpublished schema version 2, for the life of this object (not backup_dir).
+// TEST_FutureSchemaVersion2Options offers some customization for testing.
+void TEST_EnableWriteFutureSchemaVersion2(
+    BackupEngine *engine, TEST_FutureSchemaVersion2Options options);
+
```
mrambacher, line 27: The Options should be a const &

utilities/backupable/backupable_db_impl.h
```diff
@@ -0,0 +1,29 @@
...
+  bool file_sizes = true;
+  std::map<std::string, std::string> meta_fields;
+  std::map<std::string, std::string> file_fields;
+  std::map<std::string, std::string> footer_fields;
+};
+
+// Modifies the BackupEngine(Impl) to write backup meta files using the
+// unpublished schema version 2, for the life of this object (not backup_dir).
+// TEST_FutureSchemaVersion2Options offers some customization for testing.
+void TEST_EnableWriteFutureSchemaVersion2(
```
mrambacher, line 25: Is there a reason not to make this a method off of the BackupEngineImpl?

utilities/backupable/backupable_db_test.cc
```diff
@@ -2792,6 +2793,188 @@ TEST_F(BackupableDBTest, MetadataTooLarge) {
...
+  OpenDBAndBackupEngine(true);
+  // Start with supported
+  TEST_EnableWriteFutureSchemaVersion2(backup_engine_.get(), test_opts);
+  ASSERT_OK(
+      backup_engine_->CreateNewBackupWithMetadata(db_.get(), app_metadata));
+
+  // Because we are injecting badness with a TEST API, the badness is only
+  // detected on attempt to restore.
+  // NotSupported versions
+  test_opts.version = "3";
```
mrambacher, line 2872: There needs to be a test for version="2.x"

utilities/backupable/backupable_db_test.cc
```diff
@@ -2792,6 +2793,188 @@ TEST_F(BackupableDBTest, MetadataTooLarge) {
...
+  TEST_EnableWriteFutureSchemaVersion2(backup_engine_.get(), test_opts);
+  ASSERT_OK(
+      backup_engine_->CreateNewBackupWithMetadata(db_.get(), app_metadata));
+  test_opts.version = "asdf";
+  TEST_EnableWriteFutureSchemaVersion2(backup_engine_.get(), test_opts);
+  ASSERT_OK(
+      backup_engine_->CreateNewBackupWithMetadata(db_.get(), app_metadata));
```
mrambacher, line 2879: I am a little confused as to how this "bad version" is not raising a flag /failing something somewhere in the test.  Can you document where these bad versions will be caught in the test code?

mrambacher (commented):
There are a lot of constants in this code ("ni_", "size") etc.  Can you make them constants rather than literals?

utilities/backupable/backupable_db_impl.h
pdillinger: BackupEngineImpl is not in a header accessible to the test. It would be hard to see what changed if that was refactored along with this diff.

utilities/backupable/backupable_db_test.cc
pdillinger: Later in the test when go to try to restore backups, 5 of the 6 are corrupt (as expected), because of the bad meta files.

utilities/backupable/backupable_db_test.cc
pdillinger: Because this is an internal-only format (data not written by API users), I'm not really concerned about behavior outside of how we might expand use in the future. We *should* guarantee not to crash on arbitrary corruption, but this is not the highest value target for fuzz testing.

utilities/backupable/backupable_db.cc
pdillinger: Sounds reasonable.

utilities/backupable/backupable_db.cc
pdillinger: Yes, we might have just the file name and no size or checksum. I envision us always including the size when we put the new format into production, but it is useful for testing individual aspects of the old format (size not included; crc32(c) checksum included). Plus it is good to leave options open.

utilities/backupable/backupable_db_test.cc
pdillinger: In the _Restore test:

    // Minor version updates should be forward-compatible
    test_opts.version = "2.5.70";

utilities/backupable/backupable_db.cc
pdillinger: By the way, I did not elect to solve the problem of spaces in filenames. That would have been easy with JSON but we don't have an existing escape/unescape functions that escape spaces (e.g. with https://en.wikipedia.org/wiki/Percent-encoding). But this can be added later with something like

    ni::file_name_escape uri_percent

utilities/backupable/backupable_db_test.cc
pdillinger: And a checksum in the footer would obviously be better for detecting corruption

utilities/backupable/backupable_db.cc
```diff
@@ -2159,84 +2236,177 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
...
-                                " in " + meta_filename_);
+    std::string checksum_hex;
+    for (unsigned i = 1; i < components.size(); i += 2) {
+      const std::string& field_name = components[i];
+      const std::string& field_data = components[i + 1];
+
+      if (field_name == kFileCrc32cFieldName) {
+        uint32_t checksum_value =
+            static_cast<uint32_t>(strtoul(field_data.c_str(), nullptr, 10));
+        if (components[2] != ROCKSDB_NAMESPACE::ToString(checksum_value)) {
```
ajkr, line 2360: I think we need to `s/components[2]/field_data/`

ajkr (approved):
LGTM! Probably the only comment that should be addressed or replied to before landing is the one about `components[2]`.

> (b) we are increasingly relying on the DB itself for data integrity rather than
> it being an exclusive feature of backups.

Good point, the file checksums may already be in manifest and verified by user on open or restore.

utilities/backupable/backupable_db.cc
```diff
@@ -2091,10 +2097,16 @@ Status BackupEngineImpl::BackupMeta::AddFile(
...
     }
-    // Note: to save I/O, this check will pass trivially on already backed
-    // up files that don't have the checksum in their name. And it should
-    // never fail for files that do have checksum in their name.
-    if (itr->second->checksum_hex != file_info->checksum_hex) {
+    if (file_info->checksum_hex.empty()) {
+      // No checksum available to check
+    } else if (itr->second->checksum_hex.empty()) {
+      // Remember checksum if newly acquired
+      itr->second->checksum_hex = file_info->checksum_hex;
```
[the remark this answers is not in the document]
pdillinger: If we first process a backup with shared file x and unknown checksum, then process a backup with shared file x and a known checksum. We should use the best information if we have it. :)

utilities/backupable/backupable_db.cc
```diff
@@ -2159,84 +2236,177 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
     }
   }
 
+  int schema_major_version = 1;  // implicit default
+
   // Failures handled at the end
```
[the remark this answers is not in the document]
pdillinger: I don't understand the feedback. I'm calling the original schema, with no occurrence of "schema_version" version 1.

utilities/backupable/backupable_db.cc
```diff
@@ -2159,84 +2236,177 @@ Status BackupEngineImpl::BackupMeta::LoadFromFile(
...
+    } else if (schema_major_version < 2) {
+      return Status::Corruption("Expected number of files or \"" +
+                                kAppMetaDataFieldName + "\" field");
+    } else if (StartsWith(field_name, kNonIgnorableFieldPrefix)) {
+      return Status::NotSupported("Unrecognized non-ignorable meta field " +
+                                  field_name + " (from future version?)");
     } else {
-      // process the line below
+      // Warn the first time we see any particular unrecognized meta field
+      if (reported_ignored_fields->insert(field_name + "1").second) {
```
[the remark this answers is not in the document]
pdillinger: To disambiguate meta vs. file vs. footer fields, without keeping three maps. I can change it to be something like

    "file::" + field_name

utilities/backupable/backupable_db.cc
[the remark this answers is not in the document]
ajkr: Sorry just found the comment "// Failures handled at the end" amusing since there's `return`s everywhere (including before your change). It doesn't matter.

utilities/backupable/backupable_db.cc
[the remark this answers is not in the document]
ajkr: Either way is fine for me.

utilities/backupable/backupable_db.cc
```diff
@@ -884,9 +889,9 @@ Status BackupEngineImpl::Initialize() {
             work_item.progress_callback);
         result.db_id = work_item.db_id;
         result.db_session_id = work_item.db_session_id;
-        if (result.status.ok() && work_item.verify_checksum_after_work) {
```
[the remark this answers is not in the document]
pdillinger: The code is not exercised by unit tests neither before nor after this change. I'll make a note to sort that out in checksum-related follow-up.

utilities/backupable/backupable_db.cc
```diff
@@ -805,10 +809,11 @@ Status BackupEngineImpl::Initialize() {
...
           &abs_path_to_size);
       if (s.ok()) {
-        s = backup_iter->second->LoadFromFile(options_.backup_dir,
-                                              abs_path_to_size);
+        s = backup_iter->second->LoadFromFile(
+            options_.backup_dir, abs_path_to_size, options_.info_log,
+            &reported_ignored_fields_);
       }
-      if (s.IsCorruption()) {
+      if (s.IsCorruption() || s.IsNotSupported()) {
```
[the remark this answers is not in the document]
pdillinger: That is a very good point. In theory we could try to heuristically determine the files associated with a backup even when we know we cannot fully faithfully process the meta file, but that seems fragile and annoying to test.

The good news is that we only auto-GC if/when a write operation on BackupEngine is called. By implication, by invoking a write operation, you are essentially agreeing that you are OK with the set of good backups and will not be attempting some intervention for recovery of corrupted backups. Technically, this is already true, but we should probably re-emphasize this when putting schema update into production, because we will increase the likelihood of hitting this.

PR #8070, opened 2021-03-17
range_tree requires GNU libc on ppc64

adamretter:
If the platform is ppc64 and the libc is not GNU libc, then we exclude the range_tree from compilation.

See https://jira.percona.com/browse/PS-7559


adamretter:
Tests are running on pcc64le...

build_tools/build_detect_platform
```diff
@@ -668,6 +668,23 @@ else
...
+
+    int main(int argc, char *argv[]) {
+      printf("GNU libc version: %s\n", gnu_get_libc_version());
+      return 0;
+    }
+EOF
+  if [ "$?" != 0 ]; then
+      PPC_LIBC_IS_GNU=0
+  fi
+fi
```
mrambacher, line 686: Should this be LIBC_IS_GNU and be on by default?  I am missing where this value is set to 1 (non-zero).

src.mk
```diff
@@ -300,6 +286,26 @@ LIB_SOURCES_ASM =
 LIB_SOURCES_C =
 endif
 
+# range_tree is not compatible with non GNU libc on ppc64
+# see https://jira.percona.com/browse/PS-7559
+ifneq ($(PPC_LIBC_IS_GNU),0)
+LIB_SOURCES +=\
```
mrambacher, line 292: The src.mk file is included before the make_platform.mk file.  Is PPC_LIBC_IS_GNU going to be set properly before the platform file is generated?

mrambacher, line 292: Can I suggest putting these into LOCK_RANGE_SOURCES and include them in the Makefile if necessary?

mrambacher (commented):
Does the same changes also apply to CMAKE?

What about CLANG or is that not an issue?

build_tools/build_detect_platform
adamretter: Nope. It is designed so that it checks that `PPC_LIBC_IS_GNU` is not-equal-to `0` in the default case. If the flag is absent then it is not-equal-to `0`. So we only need to set it in this specific case

src.mk
adamretter: Done.

adamretter: Indeed. I have fixed that now.

adamretter:
> Does the same changes also apply to CMAKE?

@mrambacher It might need to be done for CMake, but I fixed this predominantly for ppc64le and Make. If someone needs CMake on pcc64le we could take a look later, or would you rather I attempt and fix that in this PR too?

> What about CLANG or is that not an issue?
This only checks if GNU libc is absent on ppc64le. If the libc from llvm project is used instead of gnu, then this will still correctly check that.

mrambacher (commented):
Just for clarity, can you outline how this works?  This must only work because there is no public API to use this lock manager code?  

This looks like a prime candidate for something that should be a plugin and NOT in the main source tree (not suggesting that as part of this PR but...)

adamretter:
> This must only work because there is no public API to use this lock manager code?

@mrambacher I think it was explained to me that this code is compiled but not actually used yet.

> This looks like a prime candidate for something that should be a plugin and NOT in the main source tree (not suggesting that as part of this PR but...)

I think it is anticipated that range_tree will be used from the Transactions classes, but I am not certain.

mrambacher (approved):
Works for me.

jay-zhuang (approved):
LGTM. cc. @spetrunia

PR #8071, opened 2021-03-18
Add new Append API with DataVerificationInfo to Env WritableFile

zhichao-cao:
Add the new Append and PositionedAppend API to env WritableFile. User is able to benefit from the write checksum handoff API when using the legacy Env classes. FileSystem already implemented the checksum handoff API.

test plan: make check, added new unit test.

### DIFF
--- a/include/rocksdb/env.h
+++ b/include/rocksdb/env.h
@@ -772,6 +772,13 @@ class RandomAccessFile {
   // RandomAccessFileWrapper too.
 };
 
+// A data structure brings the data verification information, which is
+// used togther with data being written to a file.
+struct DataVerificationInfo {
+  // checksum of the data being written.
+  Slice checksum;
+};
+
 // A file abstraction for sequential writing.  The implementation
 // must provide buffering since callers may append small fragments
 // at a time to the file.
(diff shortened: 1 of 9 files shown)


include/rocksdb/env.h
```diff
@@ -772,6 +772,13 @@ class RandomAccessFile {
   // RandomAccessFileWrapper too.
 };
 
+// A data structure brings the data verification information, which is
+// used togther with data being written to a file.
+struct DataVerificationInfo {
+  // checksum of the data being written.
+  Slice checksum;
+};
+
```
mrambacher, line 781: Can you just leave this in file_system and make it a forward declaration?

mrambacher:
Why do we want to add this to the legacy env rather than encouraging the new interface for this functionality?  It seems backwards to add this functionality to code that should really be deprecated.

mrambacher (commented):
I do not understand why we want this functionality in a legacy interface rather than encouraging development on the new interface instead.  Why do we want to support this in an old interface that should be deprecated?

zhichao-cao:
> I do not understand why we want this functionality in a legacy interface rather than encouraging development on the new interface instead. Why do we want to support this in an old interface that should be deprecated?

Originally, we just want to support it in File System interfaces. However, it is an urgent requests from internal products who want to use the handoff function. However, they are still based on Env and will not finish the migration to File System within 1-2 years.

include/rocksdb/env.h
zhichao-cao: Yes. I agree, it will be easier for future deprecation from Env.

include/rocksdb/env.h
```diff
@@ -801,6 +802,18 @@ class WritableFile {
...
+  // Append data with verification information.
+  // Note that this API change is experimental and it might be changed in
+  // the future. Currently, RocksDB only generates crc32c based checksum for
+  // the file writes when the checksum handoff option is set.
+  // Expected behavior: if currently ChecksumType::kCRC32C is not supported by
+  // WritableFile, the information in DataVerificationInfo can be ignored
+  // (i.e. does not perform checksum verification).
+  virtual Status Append(const Slice& data,
+                        const DataVerificationInfo& /* verification_info */) {
+    return Append(data);
```
anand1976, line 814: The ```verification_info``` will be silently ignored if the ```Env``` doesn't support it. Its the same behavior as ```FSWritableFile```, so I guess its ok.

anand1976 (approved):
LGTM. I agree its not an ideal to have these new APIs in Env, but we have to make the exception.

zhichao-cao:
> LGTM. I agree its not an ideal to have these new APIs in Env, but we have to make the exception.

Right. I think once all the services migrate to the new FileSystem APIs, we may deprecate it in the future.

PR #8075, opened 2021-03-18
Fix a test failure when built with ASSERT_STATUS_CHECKED=1

riversand963:
As title.
Test plan
ASSERT_STATUS_CHECKED=1 make -j20 backupable_db_test error_handler_fs_test
./backupable_db_test
./error_handler_fs_test

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -182,15 +182,14 @@ Status DBImpl::FlushMemTableToOutputFile(
     // other column families are missing.
     // SyncClosedLogs() may unlock and re-lock the db_mutex.
     log_io_s = SyncClosedLogs(job_context);
-    s = log_io_s;
     if (!log_io_s.ok() && !log_io_s.IsShutdownInProgress() &&
         !log_io_s.IsColumnFamilyDropped()) {
       error_handler_.SetBGError(log_io_s, BackgroundErrorReason::kFlush);
     }
   } else {
-    log_io_s.PermitUncheckedError();
     TEST_SYNC_POINT("DBImpl::SyncClosedLogs:Skip");
   }
+  s = log_io_s;
 
   // If the log sync failed, we do not need to pick memtable. Otherwise,
   // num_flush_not_started_ needs to be rollback.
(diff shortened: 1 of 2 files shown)


riversand963:
Seems related to #8049 .

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -188,6 +188,7 @@ Status DBImpl::FlushMemTableToOutputFile(
       error_handler_.SetBGError(log_io_s, BackgroundErrorReason::kFlush);
     }
   } else {
+    log_io_s.PermitUncheckedError();
```
mrambacher, line 191: Rather than PermitUncheckedError, why not make the check below on line 198 and 209 check on log_io_s?  Alternatively, move the s = log_io_s outside of the if (!ok) since the two are always effectively the same.

The log_io_s should probably be merged with the io_s a few lines below.

zhichao-cao: > Rather than PermitUncheckedError, why not make the check below on line 198 and 209 check on log_io_s? Alternatively, move the s = log_io_s outside of the if (!ok) since the two are always effectively the same.
> 
> The log_io_s should probably be merged with the io_s a few lines below.

log_io_s is intentioned to used only for io status of log, in this way, we are able to handle WAL write error and SST file write error in different logic.

zhichao-cao: > I think `log_io_s` is kept separate from the `io_s` intentionally so that the execution result of `SyncClosedLogs()` is preserved. This is used for distinguishing between different failure cases, under the current control flow: function does not return early on failure, and there is one exit point. Consequently, errors that occur in different places along the code path need to be preserved if we want to handle errors in a single place later.

You are right, thanks for the detailed explanation and thanks for the fix.

zhichao-cao (approved):
LGTM, thanks for the fix!

db/db_impl/db_impl_compaction_flush.cc
riversand963: @zhichao-cao . I do think we can improve. Currently, there are multiple such `status` or `io_status` or `log_io_status` flying around, making the already-intricate code even more complex. Would you consider using a data structure such as
```
class ErrorReportAndHandlerContext {
  Status s;
  IOStatus io_s;
  std::string path;
  FileType type;
  ...
};
```
to record the error's information? This way, we do not have to use different variables, making the code cleaner.

zhichao-cao: Yes, it is in the plan to refactor the error handling status, context, and reason.

PR #8082, opened 2021-03-20
Move compacted_db_impl.[c|h] to db/db_impl

riversand963:
As title. All core db implementations should stay in db_impl.

Test plan
make check


ajkr (approved):
LGTM

PR #8084, opened 2021-03-22
Added append with checksum handoff API to hdfs

zhichao-cao:
Added append with checksum handoff API to hdfs

test plan: make check

### DIFF
--- a/env/env_hdfs.cc
+++ b/env/env_hdfs.cc
@@ -213,6 +213,8 @@ class HdfsWritableFile: public WritableFile {
     }
   }
 
+  using WritableFile::Append;
+
   // If the file was successfully created, then this returns true.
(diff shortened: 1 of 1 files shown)


env/env_hdfs.cc
```diff
@@ -238,6 +238,11 @@ class HdfsWritableFile: public WritableFile {
     return Status::OK();
   }
 
+  virtual Status Append(const Slice& data,
```
anand1976, line 241: The default implementation is already defined in the base class, so this shouldn't be required. Can't it be solved by adding ```using WritableFile::Append```?

ajkr (approved):
LGTM

anand1976 (approved):

PR #8087, opened 2021-03-23
Use thread-safe `strerror_r()` to get error message

jay-zhuang:
`strerror()` is not thread-safe, using `strerror_r()` instead. The API could be different on the different platforms, used the code from https://github.com/facebook/folly/blob/0deef031cb8aab76dc7e736f8b7c22d701d5f36b/folly/String.cpp#L457

### DIFF
--- a/port/win/env_win.cc
+++ b/port/win/env_win.cc
@@ -62,7 +62,7 @@ typedef std::unique_ptr<void, decltype(FindCloseFunc)> UniqueFindClosePtr;
 
 void WinthreadCall(const char* label, std::error_code result) {
   if (0 != result.value()) {
-    fprintf(stderr, "pthread %s: %s\n", label,
+    fprintf(stderr, "Winthread %s: %s\n", label,
             errnoStr(result.value()).c_str());
     abort();
   }
--- a/port/win/env_win.cc
+++ b/port/win/env_win.cc
@@ -35,6 +35,7 @@
 #include "rocksdb/env.h"
 #include "rocksdb/slice.h"
 #include "strsafe.h"
+#include "util/string_util.h"
 
 // Undefine the functions  windows might use (again)...
 #undef GetCurrentTime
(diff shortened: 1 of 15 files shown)


jay-zhuang:
the vs2019-cxx20 build failure is unrelated, should be fixed by: #8090

util/string_util.h
```diff
@@ -141,4 +141,6 @@ bool SerializeIntVector(const std::vector<int>& vec, std::string* value);
 
 extern const std::string kNullptrString;
 
+extern std::string errnoStr(int err);
```
mrambacher, line 144: Please add a comment as to what this function is used for.

port/win/env_win.cc
```diff
@@ -61,7 +62,8 @@ typedef std::unique_ptr<void, decltype(FindCloseFunc)> UniqueFindClosePtr;
 
 void WinthreadCall(const char* label, std::error_code result) {
   if (0 != result.value()) {
-    fprintf(stderr, "pthread %s: %s\n", label, strerror(result.value()));
+    fprintf(stderr, "pthread %s: %s\n", label,
+            errnoStr(result.value()).c_str());
```
mrambacher, line 66: Not that it is related to this fix, but shouldn't this be WinThread, not pthread?

port/win/env_win.cc
```diff
@@ -61,7 +62,8 @@ typedef std::unique_ptr<void, decltype(FindCloseFunc)> UniqueFindClosePtr;
 
 void WinthreadCall(const char* label, std::error_code result) {
   if (0 != result.value()) {
-    fprintf(stderr, "pthread %s: %s\n", label, strerror(result.value()));
```
mrambacher, line 64: Is there a reason to go through errnoStr here rather than calling the Windows function directly?  Just curious if the redirection is worth it and does not obfuscate the code a bit.

mrambacher (approved):
Minor comments.  LGTM

port/win/env_win.cc
jay-zhuang: yeah, either way works, but as the new function also needs to allocate a buffer, it might be simpler just calling a centralized function errnoStr().

Issue #7625, opened 2020-11-02
[std:c++latest] error C2668: 'folly::bit_cast': ambiguous call to overloaded function


QuellaZhang:
Hi All,

RocksDB failed to build on MSVC under /std:c++latest mode due to "error C2668: 'folly::bit_cast': ambiguous call to overloaded function" on https://github.com/facebook/rocksdb/blob/master/third-party/folly/folly/synchronization/DistributedMutex-inl.h#L1377, we changed the "bit_cast" to "folly::bit_cast" as workaround. This is a third-party library, not sure if it needs to be modified by you, so reported to you first.

**repro steps:**

1. open VS2019 x64 tool command
2. git clone https://github.com/facebook/rocksdb.git rocksdb
3. mkdir rocksdb\build_amd64 && cd rocksdb\build_amd64
4. set _CL_=/std:c++latest
5. cmake -G "Visual Studio 16 2019" -A x64 -DCMAKE_SYSTEM_VERSION=10.0.18362.0 .
6. msbuild /m /p:Platform=x64 /p:Configuration=Release rocksdb.sln /t:Rebuild /p:BuildInParallel=true

**errors:**
DistributedMutex.cpp
F:\gitP\facebook\rocksdb\third-party\folly\folly\synchronization\DistributedMutex-inl.h(1377,1): error C2668: 'folly::bit_cast': ambiguous call to overloaded function
F:\gitP\facebook\rocksdb\third-party\folly\folly\lang\Bits.h(30,4): note: could be 'To folly::bit_cast<uintptr_t,folly::detail::distributed_mutex::Waiter<Atomic>*,0>(const From &) noexcept'
        with
        [
            To=uintptr_t,
            Atomic=std::atomic,
            From=folly::detail::distributed_mutex::Waiter<std::atomic> *
        ]
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.26.28801\include\bit(33,26): note: or       '_To std::bit_cast<uintptr_t,Waiter*,0>(const _From &) noexcept' [found using argument-dependent lookup]
        with
        [
            _To=uintptr_t,
            Waiter=folly::detail::distributed_mutex::Waiter<std::atomic>,
            _From=folly::detail::distributed_mutex::Waiter<std::atomic> *
        ]

riversand963:
Fixed #8090

PR #8090, opened 2021-03-23
Fix a compilation error in CircleCI vs2019 CXX20

riversand963:
As title.
Always specify namespace::symbol_name...
Test plan
CircleCI and other CI results


ltamasi (approved):

ltamasi:
Turns out we actually have an open issue for this (https://github.com/facebook/rocksdb/issues/7625) that we could close once this lands.

PR #8092, opened 2021-03-23
Fix Race condition in db_sst_test

akankshamahajan15:
Summary: Fix race condition in
DBSSTTest.DBWithMaxSpaceAllowedWithBlobFiles where background flush
thread updates delete_blob_file but in test thread Flush() already
completes after getting bg_error and delete_blob_file remains false.

Test Plan: Ran ASAN job few times on CircleCI

Reviewers:

Subscribers:

Tasks:

Tags:


riversand963 (approved):
Thanks @akankshamahajan15 for the fix.

PR #8093, opened 2021-03-23
Improve bloom_test bits_per_key flag

### LINKED ISSUE #7019: bloom_test fails with non-default bits_per_key

pdillinger:
Summary: Improved handling of -bits_per_key other than 10, but at least
the OptimizeForMemory test is simply not designed for generally handling
other settings. (ribbon_test does have a statistical framework for this
kind of testing, but it's not important to do that same for Bloom right
now.)

Closes #7019

Test Plan:

    for I in `seq 1 20`; do ./bloom_test --gtest_filter=-*OptimizeForMemory* --bits_per_key=$I &> /dev/null || echo FAILED; done

### DIFF
--- a/util/bloom_test.cc
+++ b/util/bloom_test.cc
@@ -30,6 +30,9 @@ int main() {
 
 using GFLAGS_NAMESPACE::ParseCommandLineFlags;
 
+// The test is not fully designed for bits_per_key other than 10, but with
+// this parameter you can easily explore the behavior of other bits_per_key.
+// See also filter_bench.
 DEFINE_int32(bits_per_key, 10, "");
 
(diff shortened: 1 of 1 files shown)


mrambacher (approved):
LGTM.

PR #8094, opened 2021-03-24
Combine condition check in track

darionyaphet:


jay-zhuang (approved):

trace_replay/trace_replay.cc
```diff
@@ -92,10 +92,8 @@ Status TracerHelper::DecodeTrace(const std::string& encoded_trace,
                                  Trace* trace) {
   assert(trace != nullptr);
   Slice enc_slice = Slice(encoded_trace);
-  if (!GetFixed64(&enc_slice, &trace->ts)) {
-    return Status::Incomplete("Decode trace string failed");
-  }
-  if (enc_slice.size() < kTraceTypeSize + kTracePayloadLengthSize) {
+  if (!GetFixed64(&enc_slice, &trace->ts) ||
+      enc_slice.size() < kTraceTypeSize + kTracePayloadLengthSize) {
```
riversand963, line 96: Should we instead report different error messages here?

riversand963 (changes requested):
See comments.

trace_replay/trace_replay.cc
darionyaphet: Actually, They're all failure about decode. I think it's fine . WDYT ~

riversand963: I prefer providing more information about the cause of failure when it's feasible and inexpensive.

jay-zhuang:
close as no response.

PR #8098, opened 2021-03-24
[Java] Fix getApproximateMemTableStats() return type

jay-zhuang:
Which should return 2 long instead of an array.


adamretter (approved):
Nice catch. LGTM :-)

PR #8099, opened 2021-03-24
Always truncate the latest WAL file on DB Open

anand1976:
Currently, we only truncate the latest alive WAL files when the DB is opened. If the latest WAL file is  empty or was flushed during Open, its not truncated since the file will be deleted later on in the Open path. However,  before deletion, a new WAL file is created, and if the process crash loops between the new WAL file creation and deletion of the old WAL file, the preallocated space will keep accumulating and eventually use up all disk space. To prevent this, always truncate  the latest WAL file, even if its empty or the data was flushed.

Tests:
Add unit tests to db_wal_test

### DIFF
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -1507,6 +1507,12 @@ class DBImpl : public DB {
   Status WriteLevel0TableForRecovery(int job_id, ColumnFamilyData* cfd,
                                      MemTable* mem, VersionEdit* edit);
 
+  // Get the size of a log file and, if truncate is true, truncate the
+  // log file to its actual size, thereby freeing preallocated space.
+  // Return success even if truncate fails
+  LogFileNumberSize GetLogSizeAndMaybeTruncate(uint64_t wal_number,
+                                               bool truncate, Status* s);
+
   // Restore alive_log_files_ and total_log_size_ after recovery.
   // It needs to run only when there's no flush during recovery
   // (e.g. avoid_flush_during_recovery=true). May also trigger flush
--- a/db/db_impl/db_impl.h
+++ b/db/db_impl/db_impl.h
@@ -1308,6 +1308,7 @@ class DBImpl : public DB {
 
   struct LogFileNumberSize {
     explicit LogFileNumberSize(uint64_t _number) : number(_number) {}
+    LogFileNumberSize() {}
     void AddSize(uint64_t new_size) { size += new_size; }
     uint64_t number;
     uint64_t size = 0;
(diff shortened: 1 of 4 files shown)


db/db_wal_test.cc
```diff
@@ -1888,6 +1888,151 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
...
+  int alloc_status = fallocate(fd, 0, 0, 1);
+  int err_number = errno;
+  close(fd);
+  ASSERT_OK(options.env->DeleteFile(fname_test_fallocate));
+  if (err_number == ENOSYS || err_number == EOPNOTSUPP) {
+    fprintf(stderr, "Skipped preallocated space check: %s\n",
+            strerror(err_number));
+    return;
+  }
+  ASSERT_EQ(0, alloc_status);
```
riversand963, line 1920: This chunk of code seems duplicate. Maybe ok for this PR, but we should consider deduping it later.

db/db_wal_test.cc
```diff
@@ -1888,6 +1888,151 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
...
+  options.env = env_;
+  options.avoid_flush_during_recovery = false;
+  if (mem_env_) {
+    ROCKSDB_GTEST_SKIP("Test requires non-mem environment");
+    return;
+  }
+  // Test fallocate support of running file system.
+  // Skip this test if fallocate is not supported.
+  std::string fname_test_fallocate = dbname_ + "/preallocate_testfile";
+  int fd = -1;
```
riversand963, line 1969: Maybe use a function to test support for fallocate.

db/db_wal_test.cc
```diff
@@ -1888,6 +1888,151 @@ TEST_F(DBWALTest, TruncateLastLogAfterRecoverWithoutFlush) {
...
+       {"DBWALTest::TruncateLastLogAfterRecoverWithFlush:AfterTruncate",
+        "DBImpl::DeleteObsoleteFileImpl::BeforeDeletion"}});
+  ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->EnableProcessing();
+  port::Thread reopen_thread([&]() { Reopen(options); });
+
+  TEST_SYNC_POINT(
+      "DBWALTest::TruncateLastLogAfterRecoverWithFlush:AfterRecover");
+  // After the flush during Open, the log file should get deleted.  However,
+  // if  the process is in a crash loop, the log file may not get
+  // deleted and thte preallocated space will keep accumulating. So we need
```
riversand963, line 1948: Nit: `s/thte/the`

db/db_impl/db_impl.h
```diff
@@ -1507,6 +1507,12 @@ class DBImpl : public DB {
   Status WriteLevel0TableForRecovery(int job_id, ColumnFamilyData* cfd,
                                      MemTable* mem, VersionEdit* edit);
 
+  // Get the size of a log file and, if truncate is true, truncate the
+  // log file to its actual size, thereby freeing preallocated space.
+  // Return success even if truncate fails
+  LogFileNumberSize GetLogSizeAndMaybeTruncate(uint64_t wal_number,
```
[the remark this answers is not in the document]
anand1976: ```LogFileNumberSize``` doesn't have a default constructor, hence did it this way.

riversand963 (approved):

PR #8103, opened 2021-03-24
Remove local static string

burtonli:
Local static string is not friendly to Jemalloc arena aware implementation, as it will be allocated on the arena of the first caller, which causes crash if the allocated arena gets refunded earlier.

P.S. A Jemalloc arena aware implementation is each rocksdb instance only use certain Jemalloc arenas, and arena will be refunded after associated DB instance is destroyed.

### DIFF
--- a/db/blob/blob_index.h
+++ b/db/blob/blob_index.h
@@ -89,7 +89,7 @@ class BlobIndex {
   }
 
   Status DecodeFrom(Slice slice) {
-    static const std::string kErrorMessage = "Error while decoding blob index";
+    const char* kErrorMessage = "Error while decoding blob index";
     assert(slice.size() > 0);
     type_ = static_cast<Type>(*slice.data());
     if (type_ >= Type::kUnknown) {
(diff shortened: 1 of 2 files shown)


ajkr:
To support this use case, it sounds like we would have to disallow local static variables that allocate memory. It sounds like an odd restriction - is this correct, and is there another way?

burtonli:
> To support this use case, it sounds like we would have to disallow local static variables that allocate memory. It sounds like an odd restriction - is this correct, and is there another way?

It's ok to have local static variable without heap memory allocation, and it's also ok to have local static variables in global instance, e.g. those singletons like the implementation for Env::Default(), SystemClock::Default(), so that user can explicitly call the constructor from the environment arena (the one last as long as the process). 
I agree we may not want to add the restriction to prohibit local static var, but I think the removed ones in the PR are not necessary.

ltamasi (approved):
Thanks for the PR @burtonli !

taoseng:
@burtonli hi~,  I recently encountered a RocksDB crash, and the issue happened to be exactly the one reported in your issue. However, what confuses me is that not only was the content of the kErrorMessage string altered, but its length was also changed. The actual crash occurred because the length was modified to an extremely large value, triggering a std::bad_alloc exception when the Status object attempted a deep copy and allocated memory. Shouldn't the length of a static local string be stored in the static area? Why was even this modified?

PR #8105, opened 2021-03-25
Apply `sample_for_compression` to all block-based tables

ajkr:
Previously it only applied to block-based tables generated by flush. This restriction
was undocumented and blocked a new use case. Now compression sampling
applies to all block-based tables we generate when it is enabled.

Test Plan: new unit test


db/db_properties_test.cc
```diff
@@ -1175,6 +1175,60 @@ class CountingDeleteTabPropCollectorFactory
...
+    return Status::OK();
+  }
+
+  Status AddUserKey(const Slice& /*user_key*/, const Slice& /*value*/,
+                    EntryType /*type*/, SequenceNumber /*seq*/,
+                    uint64_t /*file_size*/) override {
+    return Status::OK();
+  }
+
+  void BlockAdd(uint64_t /* blockRawBytes */, uint64_t blockCompressedBytesFast,
```
riversand963, line 1198: Nit: block_compressed_bytes_fast and block_compressed_bytes_slow?
To make the linter happy, we need to update the variable name in `TablePropertiesCollector::BlockAdd()` too.

riversand963 (approved):
LGTM.

db/db_properties_test.cc
ajkr: Done.

ajkr (commented):
Thanks for the review!

PR #8106, opened 2021-03-25
Adding safer permissions to PosixFilesystem::NewLogger

ImanolBarba:
We have observed rocksdb databases creating info log files with world-writeable permissions.

The reason why the file is created like so is because stdio streams opened with fopen calls use mode 0666, and while normally most systems have a umask of 022, in some occasions (for instance, while running daemons), you may find that the application is running with a less restrictive umask. The result is that when opening the DB, the LOG file would be created with world-writeable perms:

```
$ ls -lh db/
total 6.4M
-rw-r--r-- 1 ibarba users  115 Mar 24 17:41 000004.log
-rw-r--r-- 1 ibarba users   16 Mar 24 17:41 CURRENT
-rw-r--r-- 1 ibarba users   37 Mar 24 17:41 IDENTITY
-rw-r--r-- 1 ibarba users    0 Mar 24 17:41 LOCK
-rw-rw-r-- 1 ibarba users 114K Mar 24 17:41 LOG
-rw-r--r-- 1 ibarba users  514 Mar 24 17:41 MANIFEST-000003
-rw-r--r-- 1 ibarba users  31K Mar 24 17:41 OPTIONS-000018
-rw-r--r-- 1 ibarba users  31K Mar 24 17:41 OPTIONS-000020
```

This diff replaces the fopen call with a regular open() call restricting mode, and then using fdopen to associate an stdio stream with that file descriptor. Resulting in the following files being created:

```
-rw-r--r-- 1 ibarba users   58 Mar 24 18:16 000004.log
-rw-r--r-- 1 ibarba users   16 Mar 24 18:16 CURRENT
-rw-r--r-- 1 ibarba users   37 Mar 24 18:16 IDENTITY
-rw-r--r-- 1 ibarba users    0 Mar 24 18:16 LOCK
-rw-r--r-- 1 ibarba users 111K Mar 24 18:16 LOG
-rw-r--r-- 1 ibarba users  514 Mar 24 18:16 MANIFEST-000003
-rw-r--r-- 1 ibarba users  31K Mar 24 18:16 OPTIONS-000018
-rw-r--r-- 1 ibarba users  31K Mar 24 18:16 OPTIONS-000020
```

With the correct permissions


env/fs_posix.cc
```diff
@@ -554,23 +554,31 @@ class PosixFileSystem : public FileSystem {
                    std::shared_ptr<Logger>* result,
                    IODebugContext* /*dbg*/) override {
     FILE* f;
```
mrambacher, line 556: This should be initialized to nullptr

env/fs_posix.cc
```diff
@@ -554,23 +554,31 @@ class PosixFileSystem : public FileSystem {
...
-      );
+        );
+      }
+    }
+    if (fd == -1) {
+      result->reset();
+      return status_to_io_status(
+          IOError("when open a file for new logger", fname, errno));
     }
     if (f == nullptr) {
```
mrambacher, line 577: Doesn't this leak a fd (leave it open)?  If fd != -1, then the open call succeeded and f == nullptr would mean that the fdopen failed.

env/fs_posix.cc
```diff
@@ -554,23 +554,31 @@ class PosixFileSystem : public FileSystem {
                    std::shared_ptr<Logger>* result,
                    IODebugContext* /*dbg*/) override {
     FILE* f;
+    int fd;
     {
       IOSTATS_TIMER_GUARD(open_nanos);
-      f = fopen(fname.c_str(),
-                "w"
+      fd = open(fname.c_str(), O_WRONLY | O_CREAT | O_TRUNC | O_CLOEXEC, 0644);
```
mrambacher, line 560: This should be open(fname.c_str(), cloexec_flags(O_WRONLY | O_CREAT | O_TRUNC, nullptr), GetDBFileMode(allow_non_owner_access_));

env/fs_posix.cc
ImanolBarba: True! Gonna address that now

mrambacher (approved):
LGTM.

PR #8107, opened 2021-03-25
Fix flush no wal IO error bug

zhichao-cao:
There is bug in the current code base introduced in #8049 , we still set the SST file write IO Error only case as hard error. Fix it by removing the logic.

test plan: make check, error_handler_fs_test

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -260,18 +260,16 @@ Status DBImpl::FlushMemTableToOutputFile(
       // be pessimistic and try write to a new MANIFEST.
       // TODO: distinguish between MANIFEST write and CURRENT renaming
       if (!versions_->io_status().ok()) {
-        if (total_log_size_ > 0) {
-          // If the WAL is empty, we use different error reason
-          error_handler_.SetBGError(io_s,
-                                    BackgroundErrorReason::kManifestWrite);
-        } else {
-          error_handler_.SetBGError(io_s,
-                                    BackgroundErrorReason::kManifestWriteNoWAL);
-        }
-      } else if (total_log_size_ > 0 || !log_io_s.ok()) {
-        error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush);
+        // If WAL sync is successful (either WAL size is 0 or there is no IO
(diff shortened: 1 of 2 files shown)


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -268,10 +268,10 @@ Status DBImpl::FlushMemTableToOutputFile(
           error_handler_.SetBGError(io_s,
                                     BackgroundErrorReason::kManifestWriteNoWAL);
         }
-      } else if (total_log_size_ > 0 || !log_io_s.ok()) {
-        error_handler_.SetBGError(io_s, BackgroundErrorReason::kFlush);
       } else {
-        // If the WAL is empty, we use different error reason
+        // If WAL sync is successful (either WAL size is 0 or there is no IO
+        // error), all the other SST file write errors will be set as
+        // kFlushNoWAL.
```
anand1976, line 274: The ```kFlushNoWAL``` and ```kManifestWriteNoWAL``` have become a little bit of a misnomer, since there may be a non-empty WAL. It can be addressed later I think, but something to keep in mind.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -268,10 +268,10 @@ Status DBImpl::FlushMemTableToOutputFile(
           error_handler_.SetBGError(io_s,
                                     BackgroundErrorReason::kManifestWriteNoWAL);
```
anand1976, line 269: Should the same change be done here, i.e ignore ```total_log_size_``` and always set reason to ```kManifestWriteNoWAL```?

db/db_impl/db_impl_compaction_flush.cc
zhichao-cao: Yeah, I have created a task to refactor it with reason+context, then we can clearly separate them with reason==flush, context==SST file or Manifest file. Will work on it later.

db/db_impl/db_impl_compaction_flush.cc
zhichao-cao: > Should the same change be done here, i.e ignore `total_log_size_` and always set reason to `kManifestWriteNoWAL`?

Yes

anand1976 (approved):

PR #8111, opened 2021-03-25
Pass PLATFORM_FLAGS in build_detect_platform

mrambacher:
At least under MacOS, some things were excluded from the build (like Snappy) because the compilation flags were not passed in correctly.  This PR does a few things:
- Passes the EXTRA_CXX/LDFLAGS into build_detect_platform.  This means that if some tool (like TBB for example) is not installed in a standard place, it could still be detected by build_detect_platform.  In this case, the developer would invoke: "EXTRA_CXXFLAGS=<path to TBB include> EXTRA_LDFLAGS=<path to TBB library> make", and the build script would find the tools in the extra location.
- Changes the compilation tests to use PLATFORM_CXXFLAGS.  This change causes the EXTRA_FLAGS passed in to the script to be included in the compilation check.  Additionally, flags set by the script itself (like --std=c++11) will be used during the checks.  

Validated that the make_platform.mk file generated on Linux does not change with this change.  On my MacOS machine, the SNAPPY libraries are now available (they were not before as they required --std=c++11 to build).

I also verified that I can build against TBB installed on my Mac by passing in the EXTRA CXX and LD FLAGS to the location in which TBB is installed.


build_tools/build_detect_platform
```diff
@@ -177,7 +177,7 @@ case "$TARGET_OS" in
         PLATFORM_LDFLAGS="$PLATFORM_LDFLAGS -lpthread -lrt -ldl"
         if test $ROCKSDB_USE_IO_URING; then
             # check for liburing
-            $CXX $CFLAGS -x c++ - -luring -o /dev/null 2>/dev/null  <<EOF
+            $CXX $PLATFORM_CXXFLAGS -x c++ - -luring -o /dev/null 2>/dev/null  <<EOF
```
jay-zhuang, line 180: Will `$PLATFORM_CXXFLAGS` still include the `$CFLAGS`, if the user set it?

jay-zhuang (approved):
LGTM.

build_tools/build_detect_platform
mrambacher: Nothing has really changed as to what is written to the make_config.mk file.  AFAICT, CFLAGS end up in the PLATFORM_CXXFLAGS but the other options (CXXFLAGS, LDFLAGS, EXTRA_CXXFLAGS, EXTRA_LDFLAGS) do not get written to the file.  They are used however when doing the build/compile.

PR #8112, opened 2021-03-25
Introduce a ThreadGuard class and use it in ExternalSSTFileTest.PickedLevelBug

ltamasi:
Summary:
The patch adds a resource management/RAII class called `ThreadGuard`,
which can be used to ensure that the managed thread is joined when the
`ThreadGuard` is destroyed, regardless of whether it is due to the
object going out of scope, an early return, an exception etc. This is
important because if an `std::thread` object is destroyed without having
been joined (or detached) first, the process is aborted (via
`std::terminate`).

For now, `ThreadGuard` is only used in the test case
`ExternalSSTFileTest.PickedLevelBug`; however, it could come in handy
elsewhere in the codebase as well (both in test code and "real" code).
Case in point: in the `PickedLevelBug` test case, with the earlier code we
could end up in the above situation when the following assertion (which is
before the threads are joined) is triggered:

```
ASSERT_FALSE(bg_compact_started.load());
```

Test Plan:
```
make check
gtest-parallel --repeat=10000 ./external_sst_file_test --gtest_filter="*PickedLevelBug"
```

### DIFF
--- a/db/external_sst_file_test.cc
+++ b/db/external_sst_file_test.cc
@@ -1306,35 +1306,38 @@ TEST_F(ExternalSSTFileTest, PickedLevelBug) {
 
   ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->EnableProcessing();
 
-  // While writing the MANIFEST start a thread that will ask for compaction
   Status bg_compact_status;
-  ThreadGuard bg_compact(port::Thread([&]() {
-    bg_compact_status =
-        db_->CompactRange(CompactRangeOptions(), nullptr, nullptr);
-  }));
-  TEST_SYNC_POINT("ExternalSSTFileTest::PickedLevelBug:2");
-
-  // Start a thread that will ingest a new file
   Status bg_addfile_status;
-  ThreadGuard bg_addfile(port::Thread([&]() {
-    file_keys = {1, 2, 3};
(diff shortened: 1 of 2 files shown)


db/external_sst_file_test.cc
```diff
@@ -1305,38 +1306,38 @@ TEST_F(ExternalSSTFileTest, PickedLevelBug) {
...
-  // Wait for AddFile to start picking levels and writing MANIFEST
-  TEST_SYNC_POINT("ExternalSSTFileTest::PickedLevelBug:0");
-
-  TEST_SYNC_POINT("ExternalSSTFileTest::PickedLevelBug:3");
 
-  // We need to verify that no compactions can run while AddFile is
-  // ingesting the files into the levels it find suitable. So we will
-  // wait for 2 seconds to give a chance for compactions to run during
-  // this period, and then make sure that no compactions where able to run
-  env_->SleepForMicroseconds(1000000 * 2);
```
riversand963, line 1332: Testing for thread interaction should never rely on time....

ltamasi: I agree that it's not the best test case ever but in this PR, I'm just trying to prevent the test from aborting. (The test also seems flaky, although I haven't been able to reproduce that so far.)

riversand963 (approved):
LGTM.

PR #8113, opened 2021-03-25
[POC] Initial support for NVM cache in LRUCache

anand1976:
Defined the abstract interface for a NVM/persistent cache in ```include/rocksdb/nvm_cache.h```, and updated ```LRUCacheOptions``` to take a ```std::shared_ptr<NvmCache>```. An item is initially inserted into the LRU cache. When it ages out and evicted from memory, its inserted into the NVM cache. On a LRU cache miss and successful lookup in NVM, the item is promoted to the in memory cache. Only support synchronous lookup currently.

### DIFF
--- a/cache/lru_cache.cc
+++ b/cache/lru_cache.cc
@@ -97,7 +97,8 @@ void LRUHandleTable::Resize() {
 LRUCacheShard::LRUCacheShard(size_t capacity, bool strict_capacity_limit,
                              double high_pri_pool_ratio,
                              bool use_adaptive_mutex,
-                             CacheMetadataChargePolicy metadata_charge_policy)
+                             CacheMetadataChargePolicy metadata_charge_policy,
+                             const std::shared_ptr<NvmCache>& nvm_cache)
     : capacity_(0),
       high_pri_pool_usage_(0),
       strict_capacity_limit_(strict_capacity_limit),
@@ -105,7 +106,8 @@ LRUCacheShard::LRUCacheShard(size_t capacity, bool strict_capacity_limit,
       high_pri_pool_capacity_(0),
       usage_(0),
       lru_usage_(0),
-      mutex_(use_adaptive_mutex) {
+      mutex_(use_adaptive_mutex),
+      nvm_cache_(nvm_cache) {
   set_metadata_charge_policy(metadata_charge_policy);
   // Make empty circular linked list
   lru_.next = &lru_;
@@ -256,8 +258,13 @@ void LRUCacheShard::SetCapacity(size_t capacity) {
     EvictFromLRU(0, &last_reference_list);
   }
 
+  // Try to insert the evicted entries into NVM cache
   // Free the entries outside of mutex for performance reasons
   for (auto entry : last_reference_list) {
+    if (nvm_cache_ && entry->IsNvmCompatible() && !entry->IsPromoted()) {
+      nvm_cache_->Insert(entry->key(), entry->value, entry->info_.helper_cb)
+          .PermitUncheckedError();
+    }
     entry->Free();
   }
 }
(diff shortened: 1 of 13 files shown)


cache/lru_cache.h
```diff
@@ -69,6 +76,12 @@ struct LRUHandle {
     IN_HIGH_PRI_POOL = (1 << 2),
     // Whether this entry has had any lookups (hits).
     HAS_HIT = (1 << 3),
+    // Can this be inserted into the NVM cache
+    IS_NVM_COMPATIBLE = (1 << 4),
+    // Is the handle still being read from NVM
+    IS_INCOMPLETE = (1 << 5),
```
zhichao-cao, line 82: Maybe change to IS_IN_NVMREADING? Or something similar?

include/rocksdb/cache.h
```diff
@@ -137,6 +144,57 @@ class Cache {
   // likely to get evicted than low priority entries.
   enum class Priority { HIGH, LOW };
 
+  // A set of callbacks to allow objects in the volatile block cache to be
```
zhichao-cao, line 147: A general suggestion : Maybe we do not need to explicitly say it is designed for persistent cache tier usage. It is also possible that applications use block cache + another cache tier (can be either persistent or other cache designs). And we have a flag to indicate the secondary cache tier is persistent or not (to avoid cache warmup to use its persistency).

include/rocksdb/cache.h
```diff
@@ -137,6 +144,57 @@ class Cache {
...
+  // callbacks. Fill any of size_cb, saveto_cb, del_cb that is non-null
+  typedef void (*CacheItemHelperCallback)(SizeCallback* size_cb,
+                                          SaveToCallback* saveto_cb,
+                                          DeletionCallback* del_cb);
+
+  // The CreateCallback is passed by the block cache user to Lookup(). It
+  // takes in a buffer from the NVM cache and constructs an object using
+  // it. The callback doesn't have ownership of the buffer and should
+  // copy the contents into its own buffer.
+  typedef std::function<rocksdb::Status(void* buf, size_t size, void** out_obj,
```
zhichao-cao, line 194: The name "CreateCallback" is a little confusing. How about PrepareBufferCallBack or other names?

include/rocksdb/cache.h
```diff
@@ -199,6 +289,25 @@ class Cache {
   // function.
   virtual Handle* Lookup(const Slice& key, Statistics* stats = nullptr) = 0;
 
+  // Lookup the key in the volatile and NVM tiers (if one is configured).
+  // The create_cb callback function object will be used to contruct the
+  // cached object.
+  // If none of the tiers have the mapping for the key, rturns nullptr.
+  // Else, returns a handle that corresponds to the mapping.
+  //
+  // The handle returned may not be ready. The caller should call isReady()
```
zhichao-cao, line 298: Maybe an example in the comment to example implementation of how isReady, Wait, Value is used will be better.

include/rocksdb/nvm_cache.h
```diff
@@ -0,0 +1,75 @@
...
+#include "rocksdb/status.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// A handle for lookup result. The handle may not be immediately ready or
+// have a valid value. The caller must call isReady() to determine if its
+// ready, and call Wait() in order to block until it becomes ready.
+// The caller must call value() after it becomes ready to determine if the
+// handle successfullly read the item.
+class NvmCacheHandle {
```
zhichao-cao, line 24: Is it possible to introduce `class SecondaryCacheHandle {}` and make `class NvmCacheHandle : public SecondaryCacheHandle {}`. So we do not need to explicitly indicate NvmCache and let it be more extendable.

cache/lru_cache.cc
```diff
@@ -267,17 +274,126 @@ void LRUCacheShard::SetStrictCapacityLimit(bool strict_capacity_limit) {
...
+    if (nvm_cache_ && entry->IsNvmCompatible() && !entry->IsPromoted()) {
+      nvm_cache_->Insert(entry->key(), entry->value, entry->info_.helper_cb)
+          .PermitUncheckedError();
+    }
+    entry->Free();
+  }
+
+  return s;
+}
+
```
zhichao-cao, line 341: For EraseUnRefEntries() and Erase, do we need to also consider adding the entry to NVM cache?

zhichao-cao: Or if we call Erase(), we remove the entry from both block cache and NVM cache?

anand1976: I think ```Erase``` is mostly called for entries that are no longer valid (when a file is deleted for example), so we need to also erase from the NVM cache. I'll defer the implementation to a follow-on PR.

include/rocksdb/cache.h
zhichao-cao: Yeah, maybe referred to as a tiered cache will make it more general. For the flag to indicate persistent or not, I'm thinking, if it is persistent cache, we may need to check the consistency or correctness of the data in the cache when initialize the DB, such that the data in the tiered cache can be trusted. But for now, maybe we do not need it?

zhichao-cao:
Thanks for addressing the previous comments. Also, the PR description may need to be modified to reflect the tiered cache designs.

include/rocksdb/tiered_cache.h
```diff
@@ -0,0 +1,76 @@
...
+#include "rocksdb/status.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// A handle for lookup result. The handle may not be immediately ready or
+// have a valid value. The caller must call isReady() to determine if its
+// ready, and call Wait() in order to block until it becomes ready.
+// The caller must call value() after it becomes ready to determine if the
+// handle successfullly read the item.
+class TieredCacheHandle {
```
zhichao-cao, line 24: So in the regular case, use need to have a loop to check isReady() and if not, sleep the thread?

include/rocksdb/tiered_cache.h
```diff
@@ -0,0 +1,76 @@
...
+//
+// Cache interface for caching blocks on a stackable tiers (which can include
+// non-volatile mediums)
+class TieredCache {
+ public:
+  virtual ~TieredCache() {}
+
+  virtual std::string Name() = 0;
+
+  // Insert the given value into the NVM cache. The value is not written
```
zhichao-cao, line 51: May need to update the comments accordingly into Tiered Cache (instead of NVM cache) at multiple places.

cache/lru_cache.cc
```diff
@@ -267,17 +275,127 @@ void LRUCacheShard::SetStrictCapacityLimit(bool strict_capacity_limit) {
...
+        *handle = reinterpret_cast<Cache::Handle*>(e);
+      }
+    }
+  }
+
+  // Try to insert the evicted entries into NVM cache
+  // Free the entries here outside of mutex for performance reasons
+  for (auto entry : last_reference_list) {
+    if (tiered_cache_ && entry->IsTieredCacheCompatible() &&
+        !entry->IsPromoted()) {
```
zhichao-cao, line 334: I have one question here. If the entry is not promoted, means the entry is from the file reads, and we need to insert the evicted entries to tiered cache. But, if the entry is promoted, we cannot ensure that the entry is still in the tiered cache, right? It might be evicted based on the tiered cache algorithm itself.

cache/lru_cache.cc
```diff
@@ -338,81 +456,32 @@ bool LRUCacheShard::Release(Cache::Handle* handle, bool force_erase) {
...
       new char[sizeof(LRUHandle) - 1 + key.size()]);
-  Status s = Status::OK();
-  autovector<LRUHandle*> last_reference_list;
 
   e->value = value;
-  e->deleter = deleter;
+  e->flags = 0;
+  if (helper_cb) {
+    e->SetTieredCacheCompatible(true);
+    e->info_.helper_cb = helper_cb;
```
zhichao-cao, line 471: Why we don't set e->info_.deleter here through `ShardedCache::DeletionCallback del_cb; (*info_.helper_cb)(nullptr, nullptr, &del_cb);`? Does it means, the call back function can be different when we call helper_cb at different time period?

include/rocksdb/tiered_cache.h
anand1976: Not necessarily a loop, but yes, check ```isReady()``` and then call ```Wait()```.

cache/lru_cache.cc
anand1976: Inserting it here would artificially bump up the entry usage count in the tiered cache. It would have been bumped when the entry was promoted.

cache/lru_cache.cc
anand1976: The reason is we also need callbacks for demoting/promoting the entry, in addition to deleting it. Instead of keeping track of all 3, just track the ```helper_cb```. It saves some metadata in ```LRUHandle```.

cache/lru_cache.cc
zhichao-cao: I see

zhichao-cao:
Thanks for addressing the comments, the PR looks good to me.

cache/lru_cache.cc
```diff
@@ -267,17 +275,127 @@ void LRUCacheShard::SetStrictCapacityLimit(bool strict_capacity_limit) {
...
+        // The entry is in LRU since it's in hash and has no external references
+        LRU_Remove(e);
+      }
+      e->Ref();
+      e->SetHit();
+    }
+  }
+
+  // If handle table lookup failed, then allocate a handle outside the
+  // mutex if we're going to lookup in the NVM cache
```
zhichao-cao, line 365: NVM cache-> tiered cache

include/rocksdb/cache.h
```diff
@@ -137,6 +144,57 @@ class Cache {
   // likely to get evicted than low priority entries.
   enum class Priority { HIGH, LOW };
 
+  // A set of callbacks to allow objects in the volatile block cache to be
+  // be persisted in a NVM cache tier. Since the volatile cache holds C++
```
zhichao-cao, line 148: NVM cache tier -> tiered cache (e.g., NVM cache tier)

include/rocksdb/cache.h
```diff
@@ -137,6 +144,57 @@ class Cache {
...
+  // this interface, but that would create issues with managing the object
+  // lifecycle.
+  //
+  // All the callbacks are C-style function pointers in order to simplify
+  // lifecycle management. Objects in the cache can outlive the parent DB,
+  // so anything required for these operations should be contained in the
+  // object itself.
+  //
+  // The SizeCallback takes a void* pointer to the object and returns the size
+  // of the persistable data. It can be used by the NVM cache to allocate
```
zhichao-cao, line 169: Also replacing the "NVM cache" -> "tiered ache" in multiple places here

anand1976:
Cc @riversand963

include/rocksdb/cache.h
```diff
@@ -199,6 +289,25 @@ class Cache {
   // function.
   virtual Handle* Lookup(const Slice& key, Statistics* stats = nullptr) = 0;
 
+  // Lookup the key in the volatile and NVM tiers (if one is configured).
```
zhichao-cao, line 292: NVM tiers -> tiered cache layers

zhichao-cao (approved):
LGMT, thanks!

PR #8116, opened 2021-03-26
Compaction should not move data to up level

jay-zhuang:


mrambacher (approved):
LGTM.

db/compaction/compaction_picker.cc
```diff
@@ -957,6 +960,11 @@ Status CompactionPicker::SanitizeCompactionInputFilesForAllLevels(
       }
     }
   }
+  if (!unchecked_input_files.empty()) {
+    return Status::Aborted(
```
ajkr, line 964: Should it be `InvalidArgument`? Also we iterate over all the `input_files` later, is it easier to check their level against `output_level` there? https://github.com/facebook/rocksdb/blob/ec346da98c6326f65473331cec06818a6fa3d23b/db/compaction/compaction_picker.cc#L1005

ajkr (approved):
LGTM, thanks for fixing this!

ajkr:
> LGTM, thanks for fixing this!

Just remembered - can we mention this in Behavior Changes section of HISTORY.md since it feels fairly likely this will affect someone?

db/compaction/compaction_picker.cc
jay-zhuang: Yeah, thanks, that a better place to check that. Updated.

ajkr (approved):
LGTM

PR #8117, opened 2021-03-26
Disable IOStatsContext/PerfContext if no thread local

riversand963:
Before this PR, `get_iostats_context()` will silently return a nullptr if no thread_local support is detected.
This can be the result of build_detect_platform's failure to compile the simple code snippet on certain platforms, as
reported in https://github.com/facebook/mysql-5.6/issues/904.
To be safe, we should fail the compilation if user does not opt out IOStatsContext and
ROCKSDB_SUPPORT_THREAD_LOCAL is not defined.

If RocksDB relies on c++11, can we just always use thread_local? It turns out there might be
performance concerns (https://github.com/facebook/rocksdb/issues/5774),
which is beyond the scope of this PR. We can revisit this later. Here, we stick to the original impl.


mrambacher (changes requested):
This does not look correct/complete to me.  Unlike the PerfContext, the methods in the IO Context are not guarded by a define of NIO_CONTEXT.  Additionally, the PerfContext defines a static one in none-thread-local mode.

riversand963:
>  Additionally, the PerfContext defines a static one in none-thread-local mode.

This PR also adds a static one if ROCKSDB_SUPPORT_THREAD_LOCAL is undefined.

Thanks @mrambacher for the review.

mrambacher (commented):
Can you explain why the IO Stats Context should do something different than the PerfContext?  If the PerfContext is used (NPERF_CONTEXT is not defined), then we either return a static one or a thread-local one.  

I understand there may be issues with it not behaving correctly if thread locals are not supported, but do not understand why the two should behave differently.

I would rather not see the error compiler directive here.  There is no comment in the get_iostats_context() method that this might return null or something that is not "good" if thread locals are not supported.

riversand963:
> Can you explain why the IO Stats Context should do something different than the PerfContext? If the PerfContext is used (NPERF_CONTEXT is not defined), then we either return a static one or a thread-local one.
> 
If NIOSTATS_CONTEXT is not defined and RocksDB does not support thread local, the iostats is incorrect when there are multiple threads. This PR does not necessarily have to make IOContext behave the same as PerfContext, because I prefer to keep the a PR small. In fact, I believe the behavior of PerfCotext should be the same as IOStatsContext when user does not opt out perf context and lacks thread local support. When there are multiple threads, concurrent access to the static PerfContext object is not thread-safe. I can't find a mutex for this object, please let me know if I am wrong.

> 
> I would rather not see the error compiler directive here. There is no comment in the get_iostats_context() method that this might return null or something that is not "good" if thread locals are not supported.
>
We should have mentioned that this function could return nullptr. But the fact that it can return null has caused issues, which leads us to believe that it is better to fail the compilation in the first place.

riversand963:
Maybe let the Makefile / build_detect_platform report the error instead of using directives? @mrambacher

ajkr:
Should "monitoring/iostats_context_imp.h" also omit the macros when `NIOSTATS_CONTEXT` is enabled, even if thread-local is supported?

ajkr:
Also to be pedantic about API - `get_iostats_context()` says it returns a pointer to thread-local `IOStatsContext`. But now it can return pointer to a global instance. I think we should make it return nullptr or update the doc.

riversand963:
> Also to be pedantic about API - `get_iostats_context()` says it returns a pointer to thread-local `IOStatsContext`. But now it can return pointer to a global instance. I think we should make it return nullptr or update the doc.

Originally, it can return nullptr or pointer to a thread-local object. Returning nullptr silently can be problematic because it requires application developer to write their code to check for it at runtime. If we just fail the compilation, it would be simpler and safer.

mrambacher (commented):
I am not a huge fan of the error in the Makefile if local threads are not supported, but can probably live with it.  

Can we at least make the IOSTATS and PerfContext behave the same?  The PerfContext does exactly the same sort of thing with thread locals.   If one works/errors, the other should as well.

riversand963:
> 
> Can we at least make the IOSTATS and PerfContext behave the same? The PerfContext does exactly the same sort of thing with thread locals. If one works/errors, the other should as well.
>
Okay, let me update the PR and rename it.

include/rocksdb/iostats_context.h
```diff
@@ -50,7 +50,11 @@ struct IOStatsContext {
   uint64_t cpu_read_nanos;
 };
 
-// Get Thread-local IOStatsContext object pointer
+// Get Thread-local IOStatsContext object pointer if supported.
```
mrambacher, line 53: Is there a reason this method definition cannot be conditionally defined?  If NIOSTATS_CONTEXT is defined, this method does not exist, and attempts to use it will fail at compile time.

mrambacher (commented):
The comment/code around get_io_context should match that of get_perf_context.  Whatever the final form should take...

include/rocksdb/iostats_context.h
riversand963: I think the reason is we do not have to complicate the API with `ifdef`. Even if `NIOSTATS_CONTEXT` is defined, this method can still exist and return a pointer, as in the case of `PerfContext` (`get_perf_context()`). See the comment of `get_iostats_context()` and `get_perf_context()`. Keeping the definition will be safe in all combinations of thread-local support and `-DNIOSTATS_CONTEXT` / `-DNPERF_CONTEXT`.

Makefile
```diff
@@ -489,6 +489,19 @@ endif
 CFLAGS += $(C_WARNING_FLAGS) $(WARNING_FLAGS) -I. -I./include $(PLATFORM_CCFLAGS) $(OPT)
 CXXFLAGS += $(WARNING_FLAGS) -I. -I./include $(PLATFORM_CXXFLAGS) $(OPT) -Woverloaded-virtual -Wnon-virtual-dtor -Wno-missing-field-initializers
 
+TARGETS_NO_BUILD=tags tags0 clean format check-format check-buck-targets
```
mrambacher, line 492: watch-log, uninstall, install-headers...

monitoring/iostats_context.cc
```diff
@@ -11,17 +11,18 @@ namespace ROCKSDB_NAMESPACE {
...
+static IOStatsContext iostats_context;
+#endif  // !ROCKSDB_SUPPORT_THREAD_LOCAL
 
 IOStatsContext* get_iostats_context() {
-#ifdef ROCKSDB_SUPPORT_THREAD_LOCAL
   return &iostats_context;
-#else
-  return nullptr;
-#endif
 }
```
mrambacher, line 22: Should this return nullptr if NIOSTATS_CONTEXT is defined?  Should it assert?  Or can we make the function not exist at all of NIOSTATS_CONTEXT is defined?

include/rocksdb/iostats_context.h
```diff
@@ -50,7 +50,15 @@ struct IOStatsContext {
   uint64_t cpu_read_nanos;
 };
 
-// Get Thread-local IOStatsContext object pointer
+// If RocksDB is compiled with -DNIOSTATS_CONTEXT, then a pointer to a global,
+// non-thread-local IOStatsContext object will be returned. Attempts to update
+// this object will be ignored, and reading from it will also be no-op.
```
mrambacher, line 55: Is the "ignored and reading" true or is it "undefined"?

Makefile
riversand963: Will add more to the list.

ajkr: Is it common to provide invalid options with non-build targets and require the target to succeed? This feels like a detour.

ajkr (approved):
LGTM because this appears to solve both problems mentioned in https://github.com/facebook/mysql-5.6/issues/904#issuecomment-439909250.

Makefile
ajkr: On second thought, maybe it is common, since people likely provide -DNIOSTATS_CONTEXT through EXTRA_CXXFLAGS which wouldn't be provided to `make clean` etc. Do we have a standard way of detecting whether compilation is not going to happen at all?

monitoring/iostats_context.cc
```diff
@@ -11,14 +11,19 @@ namespace ROCKSDB_NAMESPACE {
 
 #ifdef ROCKSDB_SUPPORT_THREAD_LOCAL
 __thread IOStatsContext iostats_context;
-#endif
+#else
+#ifndef NIOSTATS_CONTEXT
+#error \
```
ajkr: The old approach seemed effective at only failing the make command if it involved compilation.

monitoring/iostats_context.cc
riversand963: - It does not have to and should not nullptr. If user defines `NIOSTATS_CONTEXT` explicitly (off by default), it indicates the user knows what to expect. It is better than returning a nullptr without user knowing about it, as reported in https://github.com/facebook/mysql-5.6/issues/904. If we allow returning nullptr, then user has to do nullptr-check every time, just to be safe. This is more expensive, especially for profiling code.

- It should not assert either, because assertion is no-op in opt mode.

- We do not have to make the function not exist. If we should, then I will argue that the entire file be made non-existent if `NIOSTATS_CONTEXT` is defined. So does this apply to perf context. However, this is not necessary.

include/rocksdb/iostats_context.h
riversand963: The method will be no-op, with deterministic execution result.

riversand963:
Thanks @ajkr . Somehow, Github does not allow me to reply to one of your original comments, so I'll just post here:

> Is it common to provide invalid options with non-build targets and require the target to succeed? This feels like a detour.

Here, the problem is that if the platform does not support thread-local, then even if the user types `make clean`, it will complain that `-DNIOSTATS_CONTEXT` must be defined. Adding this list so that the check is not performed for non-build targets.

Makefile
riversand963: @ajkr, I am not aware of a way to detect whether compilation will happen. However, I am not a Makefile expert, and would love to learn if there is a way of doing it.

monitoring/iostats_context.cc
riversand963: Yeah, that's right, assuming you are referring to the `#error` approach.

riversand963: It looks like we already use `#error` directives in the codebase. Now, it looks better than the current approach, since we do not have to make these changes to the Makefile and CMakelists.txt

ajkr (approved):
The updates LGTM too.

Makefile
mrambacher: @riversand963 @ajkr There is an attempt in the Makefile to determine when compilation is not going to occur.  The rules around generating %.d (ROCKSD_DEP_RULES) do this check.  But some of the checks here are still compilation.  But it might be a good start...

mrambacher (approved):
This LGTM.  Thanks for the changes!

PR #8119, opened 2021-03-26
Avoid checking errno on success call

jay-zhuang:


env/fs_posix.cc
```diff
@@ -620,9 +620,10 @@ class PosixFileSystem : public FileSystem {
       }
     }
 
-    const auto pre_read_errno = errno;  // errno may be modified by readdir
+    // reset errno before calling readdir()
+    errno = 0;
```
riversand963, line 624: At this point, we know:
1) `d` is not null, otherwise we must have returned from this function already.
2) If `d` is not null here, should we also check the errno here, or can we assume it's 0?

jay-zhuang: I think we cannot assume it's 0, as success lib function call could still change `errno`: https://stackoverflow.com/questions/45103482/can-errno-be-set-on-success
here we just reset it 0, as previous function call (`opendir()`) succeed, so we don't care what `errno` it set.

riversand963 (approved):

env/fs_posix.cc
```diff
@@ -620,9 +620,10 @@ class PosixFileSystem : public FileSystem {
...
     struct dirent* entry;
-    while ((entry = readdir(d)) != nullptr && errno == pre_read_errno) {
```
sushilpa, line 625: Since `readdir` may get called multiple times, you will need to reset `errno` every time before you call `readdir`.

sushilpa (changes requested):

env/fs_posix.cc
riversand963: Just curious: is it possible that return value of readdir is not null and error is set? Considering errno is thread local.

sushilpa: It may be possible. However, if the return value of `readdir` is not `NULL` then we should treat `errno` value as `undefined`.

ajkr (approved):
LGTM

PR #8120, opened 2021-03-26
Fix spelling in comments in include/rocksdb/

mrambacher:
Ran a spell check over the comments in the include/rocksdb directory and fixed any mis-spellings.

There are still some variable names that are spelled incorrectly (like SizeApproximationOptions::include_memtabtles, SstFileMetaData::oldest_ancester_time) that were not fixed, as those would break compilation.


include/rocksdb/table.h
```diff
@@ -555,7 +555,7 @@ struct PlainTableOptions {
 
   // @store_index_in_file: compute plain table index and bloom filter during
   //                       file building and store it in file. When reading
-  //                       file, index will be mmaped instead of recomputation.
+  //                       file, index will be mapped instead of recomputation.
```
zhichao-cao, line 558: I think they intend to use mmap here?

zhichao-cao (approved):
LGTM, thanks!

include/rocksdb/table.h
mrambacher: It is not actually mmap.  If this variable is true, the index is stored in the file and read back out (as a MetaBlock).  If the variable is false, no MetaBlock is written to the file and the index will be computed when the file is read.

PR #8122, opened 2021-03-26
Always truncate the latest WAL file on DB Open

anand1976:
Currently, we only truncate the latest alive WAL files when the DB is opened. If the latest WAL file is empty or was flushed during Open, its not truncated since the file will be deleted later on in the Open path. However, before deletion, a new WAL file is created, and if the process crash loops between the new WAL file creation and deletion of the old WAL file, the preallocated space will keep accumulating and eventually use up all disk space. To prevent this, always truncate the latest WAL file, even if its empty or the data was flushed.

Tests:
Add unit tests to db_wal_test


anand1976:
@riversand963 This is the same as #8099 which got deleted by mistake.

riversand963 (approved):

PR #8124, opened 2021-03-26
[Flaky test] Unittest uses unique test db name

jay-zhuang:
thread_id is only unique within a process. If we run the same test-set with multiple processes, it could cause db path collision between 2 runs, error message will be like:
```
...
IO error: While lock file: /tmp/rocksdbtest-501//deletefile_test_8093137327721791717/LOCK: Resource temporarily unavailable
...
```
This is could be likely reproduced by:
```
gtest-parallel ./deletefile_test --gtest_filter=DeleteFileTest.BackgroundPurgeCFDropTest -r 1000 -w 1000
```

### DIFF
--- a/test_util/testharness.cc
+++ b/test_util/testharness.cc
@@ -20,11 +20,11 @@ namespace test {
 std::string GetPidStr() {
   return std::to_string(GetCurrentProcessId());
 }
-#endif
-
+#else
 std::string GetPidStr() {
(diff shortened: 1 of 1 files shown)


ajkr (commented):
There was a request once to use `gettid()` for our system-wide unique thread ID needs (internal task T22421600). It would at least be shorter than `getpid()` + `pthread_self()`.

PR #8125, opened 2021-03-27
Fix possible hang issue in ~DBImpl() when flush is scheduled in LOW pool

sherriiiliu:
Summary: In DBImpl::CloseHelper, we wait for bg_compaction_scheduled_
and bg_flush_scheduled_ to drop to 0. Unschedule is called prior
to cancel any unscheduled flushes/compactions. It is assumed that
anything in the high priority is a flush, and anything in the low
priority pool is a compaction. This assumption, however, is broken when
the high-pri pool is full.
As a result, bg_compaction_scheduled_ can go < 0 and bg_flush_scheduled_
will remain > 0 and DB can be in hang state.
The fix is, we decrement the `bg_{flush,compaction,bottom_compaction}_scheduled_` 
inside the `Unschedule{Flush,Compaction,BottomCompaction}Callback()`s. DB 
`mutex_` will make the counts atomic in `Unschedule`.
Related discussion: #7928

Test Plan: Added new test case which hangs without the fix.

Reviewers:

Subscribers:

Tasks:

Tags:


HISTORY.md
```diff
@@ -5,6 +5,7 @@
 
 ### Bug Fixes
 * Use thread-safe `strerror_r()` to get error messages.
+* Fixed the potential hang issue when store closes in a special case. When store is unloading while there are scheduled compaction/flush, if HIGH priority pool is empty, both compactions and flushes are scheduled in LOW pool and the work counters were incorrect during unscheduling. After fix, the counters are now decremented in corresponding unschedule callbacks.
```
ajkr, line 8: It's a lot of details. How about "Fixed a potential hang in shutdown for a DB whose `Env` has high-pri thread pool disabled (`GetBackgroundThreads(Env::Priority::HIGH) == 0`)" ?

ajkr (approved):
LGTM! Thanks for trying out the suggestion!

HISTORY.md
sherriiiliu: Sure. I have updated history. Thanks for approving!

PR #8126, opened 2021-03-28
Add Internal TEST_Initialize method to Configurable

mrambacher:
Some classes/implementations of Configurable may want to initialize themselves differently when running in TEST mode.  For example, in test mode an EncryptionProvider may wish to set a default encryption key or an Env may wish to specify certain settings or its parent.

When TEST=value is specified as an option, the TEST_Initialize method will be invoked first with "value" for the Configurable object.  All Configurable objects will understand TEST and be able to initialize themselves differently based on this value.

If the TEST_Initialize method fails, configuration is aborted and the status returned.


mrambacher:
Abandoning this PR as alternatives are under investigation.

PR #8127, opened 2021-03-28
Disable partitioned filters in ts stress test

riversand963:
Currently, partitioned filter does not support user-defined timestamp. Disable it for now in ts stress test so that
the contrun jobs can proceed.

Test plan:
make crash_test_with_ts

### DIFF
--- a/tools/db_crashtest.py
+++ b/tools/db_crashtest.py
@@ -299,6 +299,7 @@ ts_params = {
     "use_blob_db": 0,
     "enable_compaction_filter": 0,
     "ingest_external_file_one_in": 0,
+    "partition_filters": 0,
 }
 
 def finalize_and_sanitize(src_params):


ajkr (approved):
Is the combination (partition filter + user ts) explicitly prevented? If there's a long way to go for fixing/preventing feature interactions, maybe we should strengthen the warning on https://github.com/facebook/rocksdb/blob/master/include/rocksdb/options.h#L1506 to say correctness is not yet guaranteed.

riversand963:
> Is the combination (partition filter + user ts) explicitly prevented? If there's a long way to go for fixing/preventing feature interactions, maybe we should strengthen the warning on https://github.com/facebook/rocksdb/blob/master/include/rocksdb/options.h#L1506 to say correctness is not yet guaranteed.

Thanks @ajkr. It's not disabled explicitly now.
I am currently doing the investigation, and I feel it should not be too hard.
Some initial debugging suggests that something is wrong during SeekForPrev() when we try to locate the correct partition filter.
When we try to locate the partition filter, we seek in the top-level index. We will be seeking (a, ts_min) because ts is set to ts_min in DBIter::SeekForPrev(). The PrefixMayMatch() will return false because the partition on the right will be used.
```
[       ]       [    ]
(a,100)  (a,90)  (b,100)
```
Will spend more time on this next week.

PR #8130, opened 2021-03-30
Replace Status with IOStatus for block fetcher IO function

zhichao-cao:
To propagate the IOStatus from file reads to RocksDB read logic, some of the existing status needs to be replaced by IOStatus.

test plan: make check

### DIFF
--- a/file/random_access_file_reader.cc
+++ b/file/random_access_file_reader.cc
@@ -22,26 +22,26 @@
 #include "util/rate_limiter.h"
 
 namespace ROCKSDB_NAMESPACE {
-Status RandomAccessFileReader::Create(
+IOStatus RandomAccessFileReader::Create(
     const std::shared_ptr<FileSystem>& fs, const std::string& fname,
     const FileOptions& file_opts,
     std::unique_ptr<RandomAccessFileReader>* reader, IODebugContext* dbg) {
   std::unique_ptr<FSRandomAccessFile> file;
-  Status s = fs->NewRandomAccessFile(fname, file_opts, &file, dbg);
-  if (s.ok()) {
+  IOStatus io_s = fs->NewRandomAccessFile(fname, file_opts, &file, dbg);
+  if (io_s.ok()) {
     reader->reset(new RandomAccessFileReader(std::move(file), fname));
   }
-  return s;
+  return io_s;
 }
 
-Status RandomAccessFileReader::Read(const IOOptions& opts, uint64_t offset,
-                                    size_t n, Slice* result, char* scratch,
-                                    AlignedBuf* aligned_buf,
-                                    bool for_compaction) const {
+IOStatus RandomAccessFileReader::Read(const IOOptions& opts, uint64_t offset,
+                                      size_t n, Slice* result, char* scratch,
+                                      AlignedBuf* aligned_buf,
+                                      bool for_compaction) const {
   (void)aligned_buf;
 
   TEST_SYNC_POINT_CALLBACK("RandomAccessFileReader::Read", nullptr);
-  Status s;
+  IOStatus io_s;
   uint64_t elapsed = 0;
   {
     StopWatch sw(clock_, stats_, hist_type_,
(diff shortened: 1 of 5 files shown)


mrambacher (commented):
This change feels off to me.  Is there a reason we cannot replace the status_ completely with io_status_ and have only one status (instead of status and io_status) in the class?  The fact that every use of the class must ignore the io status is suspect.

With this change, there are two places that update status_ left that are not io_status -- in CheckBlockChecksum and UncompressBlockContents.  (I think the call in TryGetCompressedBlockFromPersistentCache should be a local variable, like TryGetUncompressed.

Is there a reason not to convert these two remaining status into an io_status?

zhichao-cao:
> This change feels off to me. Is there a reason we cannot replace the status_ completely with io_status_ and have only one status (instead of status and io_status) in the class? The fact that every use of the class must ignore the io status is suspect.
> 
> With this change, there are two places that update status_ left that are not io_status -- in CheckBlockChecksum and UncompressBlockContents. (I think the call in TryGetCompressedBlockFromPersistentCache should be a local variable, like TryGetUncompressed.
> 
> Is there a reason not to convert these two remaining status into an io_status?

Actually, we hope that io_status_ only carries the IO related error to the upper layer. Its status can be overwritten by other status as you mentioned like TryGetUncompressed. The follow up PR's will check the io_status_ or pass it higher. Currently, we just ignore it.

mrambacher:
> > This change feels off to me. Is there a reason we cannot replace the status_ completely with io_status_ and have only one status (instead of status and io_status) in the class? The fact that every use of the class must ignore the io status is suspect.
> > With this change, there are two places that update status_ left that are not io_status -- in CheckBlockChecksum and UncompressBlockContents. (I think the call in TryGetCompressedBlockFromPersistentCache should be a local variable, like TryGetUncompressed.
> > Is there a reason not to convert these two remaining status into an io_status?
> 
> Actually, we hope that io_status_ only carries the IO related error to the upper layer. Its status can be overwritten by other status as you mentioned like TryGetUncompressed. The follow up PR's will check the io_status_ or pass it higher. Currently, we just ignore it.

(I would argue that the TryGetUncompressed code is wrong and should use a local status). 

My concern is all of the places that now need to ignore the IO Status.  That feels like a bad programming practice to me.  If there is an error, it should not matter if it is an IO or Status.  I understand that the IO status carries more information, but am not clear why you need to maintain both in this class.  Are you suggesting that a different action would be taken on a Status::NotFound then an IOStatus::NotFound (for example)?

zhichao-cao:
> > > This change feels off to me. Is there a reason we cannot replace the status_ completely with io_status_ and have only one status (instead of status and io_status) in the class? The fact that every use of the class must ignore the io status is suspect.
> > > With this change, there are two places that update status_ left that are not io_status -- in CheckBlockChecksum and UncompressBlockContents. (I think the call in TryGetCompressedBlockFromPersistentCache should be a local variable, like TryGetUncompressed.
> > > Is there a reason not to convert these two remaining status into an io_status?
> > 
> > 
> > Actually, we hope that io_status_ only carries the IO related error to the upper layer. Its status can be overwritten by other status as you mentioned like TryGetUncompressed. The follow up PR's will check the io_status_ or pass it higher. Currently, we just ignore it.
> 
> (I would argue that the TryGetUncompressed code is wrong and should use a local status).
> 
> My concern is all of the places that now need to ignore the IO Status. That feels like a bad programming practice to me. If there is an error, it should not matter if it is an IO or Status. I understand that the IO status carries more information, but am not clear why you need to maintain both in this class. Are you suggesting that a different action would be taken on a Status::NotFound then an IOStatus::NotFound (for example)?

For example, if we have retryable read error, the information can only be stored in IOStatus, then another non IO related error happens in the block_fetcher logic and override the previous one. In this case, IOStatus has retryable flag, but Status show other error. We will not do retry for this read. But if IOStatus and Status matches, we can re-read the data.

anand1976:
As @mrambacher said, it would be cleaner to just store and return ```IOStatus``` throughout ```BlockFetcher```. I don't think we need to worry too much about the ```TryGet*BlockFromPersistentCache```, as the persistent cache is not used much and will likely be deprecated at some point. The caller of ```ReadBlockContents``` can decide whether to act on the ```IOStatus``` or not, right?

zhichao-cao:
> As @mrambacher said, it would be cleaner to just store and return `IOStatus` throughout `BlockFetcher`. I don't think we need to worry too much about the `TryGet*BlockFromPersistentCache`, as the persistent cache is not used much and will likely be deprecated at some point. The caller of `ReadBlockContents` can decide whether to act on the `IOStatus` or not, right?

Right. I did not consider that the persistent cache will be deprecated. Will make the change accordingly.

anand1976 (approved):
LGTM!

zhichao-cao:
> LGTM!

Thanks for the review!

PR #8131, opened 2021-03-30
Vulnerability issue in  kramdown dependency

akankshamahajan15:
Summary: GitHub has detected that a package defined in the
docs/Gemfile.lock file of the facebook/rocksdb repository contains a
security vulnerability.
This patch fixes it by upgrading the version of kramdown to 2.3.1

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


jay-zhuang (approved):
LGTM

PR #8133, opened 2021-03-30
Move the check when verify checksums

darionyaphet:


riversand963 (commented):
The original code does almost the same thing, and I do not see an obvious advantage in one or the other.

darionyaphet:
> The original code does almost the same thing, and I do not see an obvious advantage in one or the other.

Actually, it have checked at Line:532 and only recheck after `VerifyChecksum`.

Issue #7370, opened 2020-09-10
Java UX: using an iterator from WriteBatchWithIndex can segfault


adammitchelldev:
Just a UX issue with the JNI and `WriteBatchWithIndex` that caused us a headache; RocksDB causing avoidable segfaults is annoying to debug in Java.

In this case, creating an iterator with base for a WBWI is not supported in the default configuration (`overwriteKey = false`). The C++ behaviour appears to be to return a `nullptr` when this happens, so an acceptable Java fix may be to return `null` and avoid adding a checked exception to the signature.

Another issue is that the Javadoc does not mention this limitation (where the C++ comments do), so the user is left to figure out why this segfault happens themselves.

### Expected behavior

`WriteBatchWithIndex#newIteratorWithBase(RocksIterator)` either succeeds or throws an exception (or returns null).

### Actual behavior

The method appears to succeed, but the returned object has a 0 nativeHandle, so any successive calls segfault.

### Steps to reproduce the behavior

```java
try (Options options = new Options().setCreateIfMissing(true);
     RocksDB db = RocksDB.open(options, "test");
     WriteBatchWithIndex wbwi = new WriteBatchWithIndex();
     RocksIterator base = db.newIterator();
     RocksIterator iterator = wbwi.newIteratorWithBase(base)) {
    iterator.status();
}
```

The user fix is to use `new WriteBatchWithIndex(true)`.

riversand963:
Thanks for reporting @adammitchelldev . The behavior is indeed different, and we should fix it. Would you like to contribute a PR for this?

adammitchelldev:
@riversand963 No problem, looks like @adamretter is already working on this area and should fix this on the way?

riversand963:
Looks like it since @adamretter has self-assigned this issue.

PR #8134, opened 2021-03-30
Return nullptr from WriteBatchWithIndex::NewIteratorWithBase when overwrite_key=false

adamretter:
1. Previously the behaviour of ` WriteBatchWithIndex::NewIteratorWithBase` when `WriteBatchWithIndex::overwrite_key=false` was different in debug and release builds. It is now consistent, and returns `nullptr`.
2. Tests for this functionality were missing, so I have added them.
3. I also updated the behaviour to be consistent in the Rocks Java API.

Closes https://github.com/facebook/rocksdb/issues/7370

### DIFF
--- a/include/rocksdb/utilities/write_batch_with_index.h
+++ b/include/rocksdb/utilities/write_batch_with_index.h
@@ -162,7 +162,8 @@ class WriteBatchWithIndex : public WriteBatchBase {
   // base_iterator as base.
   //
   // This function is only supported if the WriteBatchWithIndex was
-  // constructed with overwrite_key=true.
+  // constructed with overwrite_key=true, otherwise nullptr
+  // is returned.
   //
   // The returned iterator should be deleted by the caller.
   // The base_iterator is now 'owned' by the returned iterator. Deleting the
--- a/java/src/main/java/org/rocksdb/WriteBatchWithIndex.java
+++ b/java/src/main/java/org/rocksdb/WriteBatchWithIndex.java
@@ -20,6 +20,8 @@ import java.nio.ByteBuffer;
  * to get an iterator for the database with Read-Your-Own-Writes like capability
  */
 public class WriteBatchWithIndex extends AbstractWriteBatch {
+  private final boolean overwriteKey;
+
   /**
    * Creates a WriteBatchWithIndex where no bytes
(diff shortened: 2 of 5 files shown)


include/rocksdb/utilities/write_batch_with_index.h
```diff
@@ -162,7 +162,8 @@ class WriteBatchWithIndex : public WriteBatchBase {
   // base_iterator as base.
   //
   // This function is only supported if the WriteBatchWithIndex was
-  // constructed with overwrite_key=true.
+  // constructed with overwrite_key=true, otherwise nullptr
+  // is returned.
```
mrambacher, line 166: If nullptr is returned, the input iterator is still owned by the caller and must be deleted by them, not by the returned iterator.  Can you please clarify this is the documentation.

java/src/main/java/org/rocksdb/WriteBatchWithIndex.java
```diff
@@ -152,8 +169,13 @@ public RocksIterator newIteratorWithBase(
...
    * point-in-time from baseIterator and modifications made in this write batch.
    */
-  public RocksIterator newIteratorWithBase(final ColumnFamilyHandle columnFamilyHandle,
-      final RocksIterator baseIterator, /* @Nullable */ final ReadOptions readOptions) {
+  public /* @Nullable */ RocksIterator newIteratorWithBase(
+      final ColumnFamilyHandle columnFamilyHandle, final RocksIterator baseIterator,
+      /* @Nullable */ final ReadOptions readOptions) {
+    if (!overwriteKey) {
+      return null;
+    }
```
mrambacher, line 177: Instead of this, why not call iteratorWithBase and return null if the JNI method returned 0 (effectively null).  That way, if the underlying C++ code changes, the Java should "just work"

adamretter: Every call from Java into C++ is expensive, so we try and avoid them wherever possible.

jay-zhuang (approved):
LGTM.

As @mrambacher mentioned, it worth updating the doc about the ownership of base iterator when it returns nullptr.

adamretter:
Superseded by https://github.com/facebook/rocksdb/pull/9107

PR #8135, opened 2021-03-30
 Add Merge Operator support to WriteBatchWithIndex

mrambacher:
The WBWI has two differing modes of operation dependent on the value
of the constructor parameter `overwrite_key`.
Currently, regardless of the parameter, neither mode performs as
expected when using Merge. This PR remedies this by correctly invoking
the appropriate Merge Operator before returning results from the WBWI.

Examples of issues that exist which are solved by this PR:

## Example 1 with `overwrite_key=false`
Currently, from an empty database, the following sequence:
```
Put('k1', 'v1')
Merge('k1', 'v2')
Get('k1')
```
Incorrectly yields `v2`, that is to say that the Merge behaves like a Put.

## Example 2 with o`verwrite_key=true`
Currently, from an empty database, the following sequence:
```
Put('k1', 'v1')
Merge('k1', 'v2')
Get('k1')
```
Incorrectly yields `ERROR: kMergeInProgress`.

## Example 3 with `overwrite_key=false`
Currently, with a database containing `('k1' -> 'v1')`, the following sequence:
```
Merge('k1', 'v2')
GetFromBatchAndDB('k1')
```
Incorrectly yields `v1,v2`

## Example 4 with `overwrite_key=true`
Currently, with a database containing `('k1' -> 'v1')`, the following sequence:
```
Merge('k1', 'v1')
GetFromBatchAndDB('k1')
```
Incorrectly yields `ERROR: kMergeInProgress`.

## Example 5 with `overwrite_key=false`
Currently, from an empty database, the following sequence:
```
Put('k1', 'v1')
Merge('k1', 'v2')
GetFromBatchAndDB('k1')
```
Incorrectly yields `v1,v2`

## Example 6 with `overwrite_key=true`
Currently, from an empty database, `('k1' -> 'v1')`, the following sequence:
```
Put('k1', 'v1')
Merge('k1', 'v2')
GetFromBatchAndDB('k1')
```
Incorrectly yields `ERROR: kMergeInProgress`.


include/rocksdb/utilities/write_batch_with_index.h
```diff
@@ -176,6 +176,11 @@ class WriteBatchWithIndex : public WriteBatchBase {
   Iterator* NewIteratorWithBase(ColumnFamilyHandle* column_family,
                                 Iterator* base_iterator,
                                 const ReadOptions* opts = nullptr);
+  Iterator* NewIteratorWithBase(DB* db, Iterator* base_iterator,
```
pdillinger, line 179: When is the DB parameter needed? (This is public API.)

pdillinger (commented):
Also update HISTORY.md

include/rocksdb/utilities/write_batch_with_index.h
mrambacher: The DB option is used similarly to the DB option for "GetFromBatchAndBase".  Without the DB option, if you see a Merge without a preceding Put/Delete, you cannot get the "initial" value for the key without the DB.  If the other signatures are used (without the DB) and a Merge is encountered without a root, then a "MergeInProgress" error is returned.

I can update the comment appropriately when I figure out the wording...

pdillinger: Although I don't know this area well, the rest of the PR LGTM, seems like a clear improvement even if it's not perfect.

Someone questioned whether it's a good idea for WBWI to effectively re-implement all the DB logic (without persistence, etc.). I believe it's potentially useful for finding bugs on both sides if we compare them.

db/merge_context.h
```diff
@@ -68,21 +68,21 @@ class MergeContext {
...
   // Same as GetOperandsDirectionForward
-  const std::vector<Slice>& GetOperands() {
+  const std::vector<Slice>& GetOperands() const {
     return GetOperandsDirectionForward();
   }
 
   // Return all the operands in the order as they were merged (passed to
   // FullMerge or FullMergeV2)
-  const std::vector<Slice>& GetOperandsDirectionForward() {
+  const std::vector<Slice>& GetOperandsDirectionForward() const {
```
pdillinger, line 85: I think it's worth mentioning here and below that the returned vector is only good until another call to this MergeContext, including potentially a 'const' call.

pdillinger (approved):
OK with me!

db/merge_context.h
mrambacher: Fixed.

PR #8136, opened 2021-03-30
Fix error_handler_fs_test failure due to statistics

zhichao-cao:
Fix error_handler_fs_test failure due to statistics, it will fails due to multi-thread running and resume is different.

test plan: make check


akankshamahajan15 (approved):
LGTM

PR #8137, opened 2021-03-31
Fix a bug for SeekForPrev with partitioned filter and prefix

riversand963:
According to #5907, each filter partition "should include the bloom of the prefix of the last
key in the previous partition" so that SeekForPrev() in prefix mode can return correct result.
The prefix of the last key in the previous partition does not necessarily have the same prefix
as the first key in the current partition. Regardless of the first key in current partition, the
prefix of the last key in the previous partition should be added. The existing code, however,
does not follow this. Furthermore, there is another issue: when finishing current filter partition,
`FullFilterBlockBuilder::AddPrefix()` is called for the first key in next filter partition, which effectively
overwrites `last_prefix_str_` prematurely. Consequently, when the filter block builder proceeds
to the next partition, `last_prefix_str_` will be the prefix of its first key, leaving no way of adding
the bloom of the prefix of the last key of the previous partition.

Prefix extractor is FixedLength.2.
```
[  filter part 1   ]    [  filter part 2    ]
                  abc    d
```
When SeekForPrev("abcd"), checking the filter partition will land on filter part 2 because "abcd" > "abc"
but smaller than "d".
If the filter in filter part 2 happens to return false for the test for "ab", then SeekForPrev("abcd") will build
incorrect iterator tree in non-total-order mode.

Also fix a unit test which starts to fail following this PR. `InDomain` should not fail due to assertion
error when checking on an arbitrary key.

Test plan
```
make check
```

Without this fix, the following command will fail pretty soon.
```
./db_stress --acquire_snapshot_one_in=10000 --avoid_flush_during_recovery=0 \
--avoid_unnecessary_blocking_io=0 --backup_max_size=104857600 --backup_one_in=0 \
--batch_protection_bytes_per_key=0 --block_size=16384 --bloom_bits=17 \
--bottommost_compression_type=disable --cache_index_and_filter_blocks=1 --cache_size=1048576 \
--checkpoint_one_in=0 --checksum_type=kxxHash64 --clear_column_family_one_in=0 \
--compact_files_one_in=1000000 --compact_range_one_in=1000000 --compaction_ttl=0 \
--compression_max_dict_buffer_bytes=0 --compression_max_dict_bytes=0 \
--compression_parallel_threads=1 --compression_type=zstd --compression_zstd_max_train_bytes=0 \
--continuous_verification_interval=0 --db=/dev/shm/rocksdb/rocksdb_crashtest_whitebox \
--db_write_buffer_size=8388608 --delpercent=5 --delrangepercent=0 --destroy_db_initially=0 --enable_blob_files=0 \
--enable_compaction_filter=0 --enable_pipelined_write=1 --file_checksum_impl=big --flush_one_in=1000000 \
--format_version=5 --get_current_wal_file_one_in=0 --get_live_files_one_in=1000000 --get_property_one_in=1000000 \
--get_sorted_wal_files_one_in=0 --index_block_restart_interval=4 --index_type=2 --ingest_external_file_one_in=0 \
--iterpercent=10 --key_len_percent_dist=1,30,69 --level_compaction_dynamic_level_bytes=True \
--log2_keys_per_lock=10 --long_running_snapshots=1 --mark_for_compaction_one_file_in=0 \
--max_background_compactions=20 --max_bytes_for_level_base=10485760 --max_key=100000000 --max_key_len=3 \
--max_manifest_file_size=1073741824 --max_write_batch_group_size_bytes=16777216 --max_write_buffer_number=3 \
--max_write_buffer_size_to_maintain=8388608 --memtablerep=skip_list --mmap_read=1 --mock_direct_io=False \
--nooverwritepercent=0 --open_files=500000 --ops_per_thread=20000000 --optimize_filters_for_memory=0 --paranoid_file_checks=1 --partition_filters=1 --partition_pinning=0 --pause_background_one_in=1000000 \
--periodic_compaction_seconds=0 --prefixpercent=5 --progress_reports=0 --read_fault_one_in=0 --read_only=0 \
--readpercent=45 --recycle_log_file_num=0 --reopen=20 --secondary_catch_up_one_in=0 \
--snapshot_hold_ops=100000 --sst_file_manager_bytes_per_sec=104857600 \
--sst_file_manager_bytes_per_truncate=0 --subcompactions=2 --sync=0 --sync_fault_injection=False \
--target_file_size_base=2097152 --target_file_size_multiplier=2 --test_batches_snapshots=0 --test_cf_consistency=0 \
--top_level_index_pinning=0 --unpartitioned_pinning=1 --use_blob_db=0 --use_block_based_filter=0 \
--use_direct_io_for_flush_and_compaction=0 --use_direct_reads=0 --use_full_merge_v1=0 --use_merge=0 \
--use_multiget=0 --use_ribbon_filter=0 --use_txn=0 --user_timestamp_size=8 --verify_checksum=1 \
--verify_checksum_one_in=1000000 --verify_db_one_in=100000 --write_buffer_size=4194304 \
--write_dbid_to_manifest=1 --writepercent=35
```

### DIFF
--- a/db/db_bloom_filter_test.cc
+++ b/db/db_bloom_filter_test.cc
@@ -7,6 +7,9 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
+#include <iomanip>
+#include <sstream>
+
 #include "db/db_test_util.h"
 #include "options/options_helper.h"
 #include "port/stack_trace.h"
@@ -2142,20 +2145,23 @@ TEST_F(DBBloomFilterTest, SeekForPrevWithPartitionedFilters) {
   WriteOptions write_opts;
   write_opts.disableWAL = true;
   for (size_t i = 0; i < kNumKeys; ++i) {
-    char buf[5] = {0};
-    snprintf(buf, sizeof(buf), "%04d", static_cast<int>(i));
-    ASSERT_OK(db_->Put(write_opts, Slice(buf, sizeof(buf)), value));
+    std::ostringstream oss;
+    oss << std::setfill('0') << std::setw(4) << std::fixed << i;
+    ASSERT_OK(db_->Put(write_opts, oss.str(), value));
   }
   ASSERT_OK(Flush());
 
   ReadOptions read_opts;
+  // Use legacy, implicit prefix seek
   read_opts.total_order_seek = false;
   read_opts.auto_prefix_mode = false;
   std::unique_ptr<Iterator> it(db_->NewIterator(read_opts));
   for (size_t i = 0; i < kNumKeys; ++i) {
-    char buf[6] = {0};
-    snprintf(buf, sizeof(buf), "%04da", static_cast<int>(i));
-    it->SeekForPrev(Slice(buf, sizeof(buf)));
+    // Seek with a key after each one added but with same prefix. One will
+    // surely cross a partition boundary.
+    std::ostringstream oss;
+    oss << std::setfill('0') << std::setw(4) << std::fixed << i << "a";
+    it->SeekForPrev(oss.str());
     ASSERT_OK(it->status());
     ASSERT_TRUE(it->Valid());
   }
--- a/db/db_bloom_filter_test.cc
+++ b/db/db_bloom_filter_test.cc
@@ -2117,6 +2117,51 @@ TEST_F(DBBloomFilterTest, DynamicBloomFilterOptions) {
   }
 }
 
+TEST_F(DBBloomFilterTest, SeekForPrevWithPartitionedFilters) {
+  Options options = CurrentOptions();
+  constexpr size_t kNumKeys = 10000;
+  static_assert(kNumKeys <= 10000, "kNumKeys have to be <= 10000");
+  options.memtable_factory.reset(new SpecialSkipListFactory(kNumKeys + 10));
+  options.create_if_missing = true;
+  constexpr size_t kPrefixLength = 4;
+  options.prefix_extractor.reset(NewFixedPrefixTransform(kPrefixLength));
+  options.compression = kNoCompression;
+  BlockBasedTableOptions bbto;
+  bbto.filter_policy.reset(NewBloomFilterPolicy(50));
+  bbto.index_shortening =
+      BlockBasedTableOptions::IndexShorteningMode::kNoShortening;
+  bbto.block_size = 128;
+  bbto.metadata_block_size = 128;
+  bbto.partition_filters = true;
+  bbto.index_type = BlockBasedTableOptions::IndexType::kTwoLevelIndexSearch;
+  options.table_factory.reset(NewBlockBasedTableFactory(bbto));
+  DestroyAndReopen(options);
+
+  const std::string value(64, '\0');
+
+  WriteOptions write_opts;
+  write_opts.disableWAL = true;
+  for (size_t i = 0; i < kNumKeys; ++i) {
+    char buf[5] = {0};
+    snprintf(buf, sizeof(buf), "%04d", static_cast<int>(i));
+    ASSERT_OK(db_->Put(write_opts, Slice(buf, sizeof(buf)), value));
+  }
+  ASSERT_OK(Flush());
+
+  ReadOptions read_opts;
+  read_opts.total_order_seek = false;
+  read_opts.auto_prefix_mode = false;
+  std::unique_ptr<Iterator> it(db_->NewIterator(read_opts));
+  for (size_t i = 0; i < kNumKeys; ++i) {
+    char buf[6] = {0};
+    snprintf(buf, sizeof(buf), "%04da", static_cast<int>(i));
+    it->SeekForPrev(Slice(buf, sizeof(buf)));
+    ASSERT_OK(it->status());
+    ASSERT_TRUE(it->Valid());
+  }
+  it.reset();
+}
+
 #endif  // ROCKSDB_LITE
 
 }  // namespace ROCKSDB_NAMESPACE
--- a/table/block_based/full_filter_block.cc
+++ b/table/block_based/full_filter_block.cc
@@ -22,6 +22,7 @@ FullFilterBlockBuilder::FullFilterBlockBuilder(
       whole_key_filtering_(whole_key_filtering),
       last_whole_key_recorded_(false),
       last_prefix_recorded_(false),
+      last_key_in_domain_(false),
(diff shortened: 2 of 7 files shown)


pdillinger:
Can you just add key0001 ... key9999 and repeatedly SeekForPrev(key9998a) ... SeekForPrev(key0001a)?

riversand963:
> Can you just add key0001 ... key9999 and repeatedly SeekForPrev(key9998a) ... SeekForPrev(key0001a)?

I can try that. Are we counting on there being one prefix for which the filter partition does not false-positive?

pdillinger:
> I can try that. Are we counting on there being one prefix for which the filter partition does not false-positive?

I can't think that through at the moment, but you can always set like 50 bits per key to essentially ensure no false positives.

db/db_bloom_filter_test.cc
```diff
@@ -2117,6 +2117,51 @@ TEST_F(DBBloomFilterTest, DynamicBloomFilterOptions) {
...
+  write_opts.disableWAL = true;
+  for (size_t i = 0; i < kNumKeys; ++i) {
+    char buf[5] = {0};
+    snprintf(buf, sizeof(buf), "%04d", static_cast<int>(i));
+    ASSERT_OK(db_->Put(write_opts, Slice(buf, sizeof(buf)), value));
+  }
+  ASSERT_OK(Flush());
+
+  ReadOptions read_opts;
+  read_opts.total_order_seek = false;
```
pdillinger, line 2152: Suggest a comment like

    // Use legacy, implicit prefix seek

db/db_bloom_filter_test.cc
```diff
@@ -2117,6 +2117,51 @@ TEST_F(DBBloomFilterTest, DynamicBloomFilterOptions) {
...
+  bbto.index_type = BlockBasedTableOptions::IndexType::kTwoLevelIndexSearch;
+  options.table_factory.reset(NewBlockBasedTableFactory(bbto));
+  DestroyAndReopen(options);
+
+  const std::string value(64, '\0');
+
+  WriteOptions write_opts;
+  write_opts.disableWAL = true;
+  for (size_t i = 0; i < kNumKeys; ++i) {
+    char buf[5] = {0};
```
pdillinger, line 2145: Nit: I would prefer us to get away from C arrays and snprintf. Here your keys oddly include the null byte. And initializing only the first element to the null byte doesn't accomplish anything because if snprintf fails to overwrite, we still have uninitialized bytes in our key, because it is not interpreted as a C array.

I known C++ doesn't provide convenient fixed-width printing of numbers into strings, but perhaps we can add something to string_util.h.

db/db_bloom_filter_test.cc
```diff
@@ -2117,6 +2117,51 @@ TEST_F(DBBloomFilterTest, DynamicBloomFilterOptions) {
...
+  }
+  ASSERT_OK(Flush());
+
+  ReadOptions read_opts;
+  read_opts.total_order_seek = false;
+  read_opts.auto_prefix_mode = false;
+  std::unique_ptr<Iterator> it(db_->NewIterator(read_opts));
+  for (size_t i = 0; i < kNumKeys; ++i) {
+    char buf[6] = {0};
+    snprintf(buf, sizeof(buf), "%04da", static_cast<int>(i));
```
pdillinger, line 2157: Suggest a comment like

    // Seek with a key after each one added but with same prefix. One will surely cross
    // a partition boundary.

table/block_based/full_filter_block.cc
```diff
@@ -30,6 +31,15 @@ FullFilterBlockBuilder::FullFilterBlockBuilder(
 void FullFilterBlockBuilder::Add(const Slice& key_without_ts) {
   const bool add_prefix =
       prefix_extractor_ && prefix_extractor_->InDomain(key_without_ts);
+
+  if (!last_prefix_recorded_ && last_key_in_domain_) {
+    // We can reach here when a new filter partition starts in partitioned
+    // filter. The last prefix in the previous partition should be added if
+    // necessary regardless of key_without_ts.
```
pdillinger, line 38: Perhaps add

    // necessary regardless of key_without_ts, to support prefix SeekForPrev.

table/block_based/partitioned_filter_block.cc
```diff
@@ -79,7 +79,7 @@ void PartitionedFilterBlockBuilder::MaybeCutAFilterBlock(
   const bool add_prefix =
       next_key && prefix_extractor() && prefix_extractor()->InDomain(*next_key);
   if (add_prefix) {
-    FullFilterBlockBuilder::AddPrefix(*next_key);
+    AddKey(prefix_extractor()->Transform(*next_key));
```
pdillinger, line 82: Why this change? It seems like we want to skip adding and `num_added_++` if redundant.

pdillinger (commented):
Nice fix. Some nits and a question.

db/db_bloom_filter_test.cc
riversand963: Let me try to update string_util.h

table/block_based/partitioned_filter_block.cc
riversand963: Good finding.
We cannot call `AddPrefix` because it may update `lasts_prefix_str_` so that it stores the prefix of `*next_key`. However, when we build the next filter partition, we need the prefix of the last key of the previous filter partition.
Originally I thought adding the same string to the filter is OK, but maybe it's not. Maybe let me do the checking here before calling `AddKey()`.

db/db_bloom_filter_test.cc
riversand963: Actually we can just use #include <iomanip>

table/block_based/partitioned_filter_block.cc
pdillinger: Oh right. It should be "safe" to add redundantly, though the filter implementation will allocate space for one more key than necessary.

table/block_based/partitioned_filter_block.cc
```diff
@@ -76,10 +76,13 @@ void PartitionedFilterBlockBuilder::MaybeCutAFilterBlock(
   // Add the prefix of the next key before finishing the partition. This hack,
```
pdillinger, line 76: Suggest

    ... before finishing the partition, without updating last_prefix_str.

pdillinger (approved):
LGTM. Thanks for the find & fix!

riversand963:

ajkr (commented):
Great finding and fix! Interesting to learn #8127's cause was different from expected.

PR #8140, opened 2021-03-31
Improve performance of MultiRead by doing more work async

micvbang:
Hello,

I'm currently working on implementing a new environment for RocksDB that abstracts the file API so that the same environment can be used for multiple platforms (Windows, FreeBSD, and Linux).
In my work, I stumbled upon some surprising performance numbers; my new environment showed performance improvements compared to the existing PosixEnv on the order of \~5-15% for async workloads utilizing direct io, although this new environment uses liburing just like the existing implementation. The performance seems to improve with increased batch size.

Performance improvements

The ops reported on the graph above are the ops/sec reported by `db_bench fillrandom,multireadrandom`'s, averaged over 5 runs with the stddev as error bars.

The tests were run on a box with:
```
CPU: AMD Ryzen 5 5600X 6-Core Processor
RAM: 32GB DDR4-3200
Disk: Samsung 860 EVO SSD
```
See last section for more details.

# Technical details

Both `PosixEnv` and the new environment I'm working on uses liburing - I'm using liburing through [xNVMe](https://xnvme.io) though. The difference in performance comes from the _way_ liburing is used.

In the existing implementation of PosixRandomAccessFile, MultiRead submits and waits for all IO using a single call to `io_uring_submit_and_wait()`. Here, it seems there's a missed opportunity to do work while waiting for other IOs to complete, i.e. preparing IOs to be submitted or processing completed IOs.

This commit introduces the following changes:

- submits IOs individually using `io_uring_submit`, allowing IO prep work to be done between submissions

- uses `io_uring_peek_batch_cqe` to poll for CQEs, allowing RocksDB's CQE completion work to be done when some IOs have completed, but other IOs are still pending

- replace the usage of `io_uring_prep_readv` with `io_uring_prep_read` which was not available at the time the existing implementation was introduced

Note: `io_uring_prep_read` was introduced in `liburing-0.4` and thus potentially is not available on all existing systems. The change from `io_uring_prep_readv` to `io_uring_prep_read` is not strictly necessary as the bulk of the performance improvement comes from doing while other IOs are pending. If preferred, this commit can just be left out

# Performance evaluation

As described earlier, my performance numbers were measured using `db_bench`'s benchmarks `fillrandom` and `multireadrandom`. The `ops` I report in my graph are the output from `multireadrandom`, e.g. `multireadrandom :      22.475 micros/op 44492 ops/sec; (1000000 of 1000000 found)` would be reported as 44,492 ops/sec.

It would be great if you could run and verify the measurements on your own hardware.

The following bash script can be used to run the tests that I've run:

```bash
for BATCH_SIZE in 64 32 16 8 4 2
do
	db_bench -benchmarks=fillrandom,multireadrandom --num=1000000 -use_direct_reads=True -use_direct_io_for_flush_and_compaction=True -multiread_batched=true -batch-size=$BATCH_SIZE
done
```

I used a much more elaborate setup than the above script for running the tests, but it's not easily shareable. Sorry.

Please let me know if you have any questions or comments


micvbang:
Hey @zhichao-cao, @anand1976! Just thought I'd ping you on this one :) Let me know if there's anything that should be clarified or done in order to make this easier to review. I'm available for any questions or comments you might have.
Thanks!

micvbang:
Hey @zhichao-cao, @anand1976, just thought I'd ping you again :) Is there something I can do on this one to help a review in any way?

Thank you!

anand1976:
@micvbang My apologies. Will get back to you in a couple of days.

micvbang:
@anand1976 not a problem. Thank you!

env/io_posix.cc
```diff
@@ -663,82 +664,81 @@ IOStatus PosixRandomAccessFile::MultiRead(FSReadRequest* reqs,
...
 
       struct io_uring_sqe* sqe;
       sqe = io_uring_get_sqe(iu);
-      io_uring_prep_readv(
-          sqe, fd_, &rep_to_submit->iov, 1,
+      io_uring_prep_read(
+          sqe, fd_, rep_to_submit->buf, rep_to_submit->len,
           rep_to_submit->req->offset + rep_to_submit->finished_len);
       io_uring_sqe_set_data(sqe, rep_to_submit);
+      io_uring_submit(iu);
```
anand1976, line 678: I'm not that familiar with the IO uring internals. Does each call to ```io_uring_submit``` result in a system call?

env/io_posix.cc
```diff
@@ -663,82 +664,81 @@ IOStatus PosixRandomAccessFile::MultiRead(FSReadRequest* reqs,
...
-    ssize_t ret =
-        io_uring_submit_and_wait(iu, static_cast<unsigned int>(this_reqs));
-    if (static_cast<size_t>(ret) != this_reqs) {
-      fprintf(stderr, "ret = %ld this_reqs: %ld\n", (long)ret, (long)this_reqs);
-    }
-    assert(static_cast<size_t>(ret) == this_reqs);
+    size_t cqes_left = this_reqs;
+    while (cqes_left > 0) {
+      struct io_uring_cqe* cqes[cqes_left];
+      size_t completed = io_uring_peek_batch_cqe(iu, cqes, cqes_left);
```
anand1976, line 685: IIUC, ```io_uring_peek_batch_cqe``` doesn't wait for any completions, right? So is this effectively a busy wait loop?

anand1976 (commented):
Thanks for the PR! This is an interesting approach. I have some concerns about CPU utilization though. Can you measure the CPU util while running the benchmarks for comparison?

env/io_posix.cc
micvbang: Yep, correct :)

micvbang:
Hello @anand1976,

I've done some CPU utilization measurements and they are like you would expect; the busy waiting loop indeed grabs a core while waiting for IO.

Below I've plotted the CPU utilization of all cores on my system during the performance evaluation described in the original post with a batch size of 32 (`db_bench`'s benchmarks `fillrandom` and `multireadrandom`, on the same hardware mentioned earlier.) These plots are representative for all batch sizes so I'll the number of figures and just post the following two.
 
| CPU utilization master |
|:--:| 
| *CPU utilization on master*|

| CPU utilization PR branch |
|:--:| 
| *CPU utilization on the PR branch* |

As you hinted at in your comment about CPU utilization, this is a trade-off between CPU utilization and performance. Having a core polling the ring for completions decreases the latency of each IO and improves IO throughput.

I guess the question then becomes if this trade-off is something that the wider RocksDB community would like enough that it becomes the default. I can certainly understand if it isn't :)

In case it isn't great as the default, I would suggest making this optimization available using a flag. In this case I think it would also be natural to investigate using io uring's kernel polling feature to avoid the overhead of syscalls during sustained IO.

What do you think?

micvbang:
@anand1976  Just thought I'd ping you on this :)

micvbang:
Hey @zhichao-cao, @anand1976,

I've now put this feature behind a flag so that users that want can trade CPU utilization for improved IO latency and overall performance, while users that don't can keep using the existing implementation.

Please let me know if there's anything I can do to make this easier to review!

anand1976:
@micvbang Thanks for updating the PR. Putting the feature behind the flag is a good idea as IO throughput may not be the dominant factor for most users. I'll review and get back to you in a couple of days.

micvbang:
> @micvbang Thanks for updating the PR. Putting the feature behind the flag is a good idea as IO throughput may not be the dominant factor for most users. I'll review and get back to you in a couple of days.

@anand1976 Thank you. Just thought I'd ping you on this :)

micvbang:
Hey @zhichao-cao, @anand1976, just thought I'd ping you again :) Is there something I can do on this one to help a review in any way?

Thank you!

env/io_posix.cc
anand1976: Can we check ```allow_io_polling``` here and , if false, use the wait variant?

anand1976 (commented):
@micvbang Thanks for addressing the previous comments! Apologies for not getting to this sooner.

env/io_posix.cc
micvbang: Yeah, definitely. But it shouldn't be necessary - since `io_uring_submit_and_wait()` is called in case `allow_io_polling` is set, all IOs will be ready to be fetched when we call `io_uring_peek_batch_cqe()` here

micvbang:
@anand1976 thanks for getting back to me!

FYI I'm on vacation this and next week, so that's why I'll be slow/not respond during that time :)

env/io_posix.cc
micvbang: Just pinging you @anand1976 - the thread was marked as resolved when I commented, so I'm not sure if you got a notification.

micvbang:
@anand1976 Just thought I'd ping you on this :)

micvbang:
Hey @anand1976, just thought I'd ping you again :) Is there something I can do on this one to help a review in any way?

Thank you!

PR #8141, opened 2021-04-01
Fix compression dictionary sampling with dedicated range tombstone SSTs

ajkr:
Return early in case there are zero data blocks when
`BlockBasedTableBuilder::EnterUnbuffered()` is called. This crash can
only be triggered by applying dictionary compression to SST files that
contain only range tombstones. It cannot be triggered by a low buffer
limit alone since we only consider entering unbuffered mode after
buffering a data block causing the limit to be breached, or `Finish()`ing the file. It also cannot
be triggered by a totally empty file because those go through
`Abandon()` rather than `Finish()` so unbuffered mode is never entered.

Test Plan: added a unit test that repro'd the "Floating point exception"


riversand963 (approved):
Thanks @ajkr for the fix.

PR #8142, opened 2021-04-01
Make backups openable as read-only DBs

pdillinger:
Summary: A current limitation of backups is that you don't know the
exact database state of when the backup was taken. With this new
feature, you can at least inspect the backup's DB state without
restoring it by opening it as a read-only DB.

Rather than add something like OpenAsReadOnlyDB to the BackupEngine API,
which would inhibit opening stackable DB implementations read-only
(if/when their APIs support it), we instead provide a DB name and Env
that can be used to open as a read-only DB.

Possible follow-up work:

* Add a version of GetBackupInfo for a single backup.
* Let CreateNewBackup return the BackupID of the newly-created backup.

Implementation details:

Refactored ChrootFileSystem to split off new base class RemapFileSystem,
which allows more general remapping of files. We use this base class to
implement BackupEngineImpl::RemapSharedFileSystem.

To minimize API impact, I decided to just add these fields `name_for_open`
and `env_for_open` to those set by GetBackupInfo when
include_file_details=true. Creating the RemapSharedFileSystem adds a bit
to the memory consumption, perhaps unnecessarily in some cases, but this
has been mitigated by (a) only initialize the RemapSharedFileSystem
lazily when GetBackupInfo with include_file_details=true is called, and
(b) using the existing `shared_ptr<FileInfo>` objects to hold most of the
mapping data.

To enhance API safety, RemapSharedFileSystem is wrapped by new
ReadOnlyFileSystem which rejects any attempts to write. This uncovered a
couple of places in which DB::OpenForReadOnly would write to the
filesystem, so I fixed these. Added a release note because this affects
logging.

Additional minor refactoring in backupable_db.cc to support the new
functionality.

Test Plan: new test (run with ASAN and UBSAN), added to stress test and
ran it for a while with amplified backup_one_in


env/fs_remap.h
```diff
@@ -0,0 +1,209 @@
...
+                           IODebugContext* dbg) override;
+};
+
+// An abstract FileSystem wrapper that only allows read-only operation.
+//
+// This class has not been fully analyzed for providing strong security
+// guarantees.
+class ReadOnlyFileSystem : public FileSystemWrapper {
+  static inline IOStatus FailReadOnly() {
+    return IOStatus::IOError("Attempted write to ReadOnlyFileSystem");
```
mrambacher, line 135: Why an IOError and not NotSupported?  Seems like an IOError could lead to situations where the operation might be retried when it will always fail.

include/rocksdb/utilities/backupable_db.h
```diff
@@ -280,24 +280,43 @@ struct BackupFileInfo {
...
+  // include_file_details=true. (The FileSystem in this Env takes care
+  // of making shared backup files openable from the `name_for_open` DB
+  // directory.) See also name_for_open.
+  //
+  // This Env might or might not be shared with other backups. To work
+  // around DBOptions::env being a raw pointer, this is a shared_ptr so
+  // that keeping either this BackupInfo, the BackupEngine, or a copy of
+  // this shared_ptr alive is sufficient to keep the Env alive for use by
+  // a read-only DB.
+  std::shared_ptr<Env> env_for_open;
```
mrambacher, line 318: This seems very dangerous to me.  If env_for_open is set to Env::Default (for example), then won't the default env be deleted once this is out of scope?

Cn this be a FileSystem instead of an Env?  Are there operations in the Env that you require that are not in the FS?

utilities/backupable/backupable_db.cc
```diff
@@ -65,6 +66,16 @@ inline std::string ChecksumInt32ToHex(const uint32_t& checksum_value) {
...
+
+const std::string kPrivateDirName = "private";
+const std::string kMetaDirName = "meta";
+const std::string kSharedDirName = "shared";
+const std::string kSharedChecksumDirName = "shared_checksum";
+const std::string kPrivateDirSlash = kPrivateDirName + "/";
+const std::string kMetaDirSlash = kMetaDirName + "/";
+const std::string kSharedDirSlash = kSharedDirName + "/";
+const std::string kSharedChecksumDirSlash = kSharedChecksumDirName + "/";
+
```
mrambacher, line 78: Can/Should these be static const?

utilities/backupable/backupable_db.cc
```diff
@@ -195,6 +206,133 @@ class BackupEngineImpl {
...
+    }
+  }
+
+  static inline std::string WithTrailingSlash(const std::string& path) {
+    if (path.empty() || path.back() != '/') {
+      return path + '/';
+    } else {
+      return path;
+    }
+  }
```
mrambacher, line 244: Would NormalizePath meet the needs here?

db/db_impl/db_impl_files.cc
```diff
@@ -865,9 +865,14 @@ Status DBImpl::SetDBId() {
     // it is no longer available then at this point DB ID is not in Identity
     // file or Manifest.
     if (s.IsNotFound()) {
-      s = SetIdentityFile(env_, dbname_);
-      if (!s.ok()) {
-        return s;
+      if (read_only) {
+        db_id_ = env_->GenerateUniqueId();
+        return Status::OK();
```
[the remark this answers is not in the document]
pdillinger: I don't know the history on why, but backups do not save the IDENTITY file. If you restore to a fresh dir, you get a new db ID, which is what this replicates.

env/fs_remap.h
```diff
@@ -0,0 +1,209 @@
...
+  IOStatus NewLogger(const std::string& fname, const IOOptions& options,
+                     std::shared_ptr<Logger>* result,
+                     IODebugContext* dbg) override;
+
+  IOStatus GetAbsolutePath(const std::string& db_path, const IOOptions& options,
+                           std::string* output_path,
+                           IODebugContext* dbg) override;
+};
+
+// An abstract FileSystem wrapper that only allows read-only operation.
```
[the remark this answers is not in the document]
pdillinger: Thoughtless copy-paste error

env/fs_remap.h
pdillinger: I attempted to replicate the same error you would get by trying to write to a read-only filesystem. I believe this construction sets retryable to false.

include/rocksdb/utilities/backupable_db.h
pdillinger: env_for_open is a field, so for an API user to reset it to some raw ptr they would have to use .reset() on a raw ptr they don't control the lifetime of. What's the new hazard here? Also this is an "output" struct for our API. Why would an API user be writing to these fields at all?

> Cn this be a FileSystem instead of an Env?

Then the caller has unnecessary complexity of creating their own temporary Env. What's the benefit?

utilities/backupable/backupable_db.cc
pdillinger: They're in an anonymous namespace. I didn't think it was important to put them in the class.

mrambacher: I was not suggesting in a class, just static to guarantee scoping and creation constraints.

utilities/backupable/backupable_db.cc
pdillinger: It does not remove trailing /

ajkr (approved):
LGTM! Glad restore validation can now be done natively. Seems good to push users through the `DB` open functions.

PR #8144, opened 2021-04-02
stack trace freebsd update. using native api to get the process

devnexen:
full name.
Summary: /proc/ filesystem is not necessarily present on this platform thus relying more on the sysctl/IOD api to get the process path.

Test Plan: Provoking purposely a segfault.


riversand963 (approved):
Thanks @devnexen. Could you update the PR description?

PR #8145, opened 2021-04-02
Fix db_test2 parallelism

pdillinger:
Summary: With thread/process-specific dirs. (Errors seen in FB infra.)

Test Plan: see in FB infra tests


pdillinger:
Internal example: https://fburl.com/sandcastle/r4sw93w0

riversand963 (approved):

PR #8146, opened 2021-04-02
Make tests "parallel" and "passing ASC" by default

pdillinger:
Summary: New tests should by default be expected to be parallelizeable
and passing with ASSERT_STATUS_CHECKED. Thus, I'm changing those two
lists to exclusions rather than inclusions.

For the set of exclusions, I only listed things that currently failed
for me when attempting not to exclude, or had some other documented
reason. This marks many more tests as "parallel," which will potentially
cause some failures from self-interference, but we can address those as
they are discovered.

Also changed CircleCI ASC test to be parallelized; the easy way to do
that is to exclude building tests that don't pass ASC, which is now a
small set.

Test Plan: Watch CI, etc.


pdillinger:
Looking into

    Makefile:812: target 'parallel_file_reader_writer_test' given more than once in the same rule

And linux-non-shm-X builds

riversand963 (approved):

PR #8147, opened 2021-04-03
Use separate db dir for different tests hoping to remove flakiness

riversand963:
DBWALTestWithParam relies on `SstFileManager` to have the expected behavior. However, if this test shares
db directories with other DBSSTTest, then the SstFileManager may see non-empty data, thus will change its
behavior to be different from expectation, introducing flakiness.

Test plan
make check

### DIFF
--- a/db/db_sst_test.cc
+++ b/db/db_sst_test.cc
@@ -751,10 +751,11 @@ TEST_F(DBSSTTest, RateLimitedWALDelete) {
 }
 
 class DBWALTestWithParam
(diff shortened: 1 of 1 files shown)


jay-zhuang (approved):
LGTM.

Should the test also cleanup DB path before finish?

riversand963:
> LGTM.
> 
> Should the test also cleanup DB path before finish?

It is cleaned up in the destructor of base class, unless told otherwise.

PR #8148, opened 2021-04-03
Add Blob Options to C API

mrambacher:
Added the Blob option settings from the AdvancedColmnFamilyOptions to the C API.

There are no tests for getting/setting options in the C API currently, hence no specific test plans.  Should there be a some?


ltamasi:
> There are no tests for getting/setting options in the C API currently, hence no specific test plans. Should there be a some?

We do have a C API test suite hidden under https://github.com/facebook/rocksdb/blob/master/db/c_test.c that we could extend.

mrambacher:
> > There are no tests for getting/setting options in the C API currently, hence no specific test plans. Should there be a some?
> 
> We do have a C API test suite hidden under https://github.com/facebook/rocksdb/blob/master/db/c_test.c that we could extend.

I can add some tests of the Blob code there, but not sure how I can validate the functionality from the DB API.  If you point me at an example/sample test that could validate Blobs in C, I will add it to this PR (or a new one if you prefer).

ltamasi:
> I can add some tests of the Blob code there, but not sure how I can validate the functionality from the DB API. If you point me at an example/sample test that could validate Blobs in C, I will add it to this PR (or a new one if you prefer).

I think there are two kinds of tests that we can add here:
1) Simply setting then getting the options to make sure the API works. This could be done in the `options` phase in `c_test` (look for `StartPhase("options");`).
2) Some sort of functional testing. Technically most of the stuff except for backup/restore (which is in progress) and `Merge` should also work with BlobDB but I don't think testing everything is necessary since we have the C++ unit tests for that. Maybe we could do something like open the database with `enable_blob_files` set to `true`, and do some `Put`s followed by `Get`s / iteration? (I.e. just make sure we can read back what we wrote through the C API.)

db/c.cc
```diff
@@ -2682,6 +2682,59 @@ unsigned char rocksdb_options_get_skip_checking_sst_file_sizes_on_db_open(
   return opt->rep.skip_checking_sst_file_sizes_on_db_open;
 }
 
+/* Blob Options Settings */
+void rocksdb_options_set_enable_blob_files(rocksdb_options_t* opt,
+                                           unsigned char val) {
+  opt->rep.enable_blob_files = val;
+}
+extern ROCKSDB_LIBRARY_API unsigned char rocksdb_options_get_enable_blob_files(
```
ltamasi, line 2690: `ROCKSDB_LIBRARY_API` might be redundant here.

ltamasi (approved):
Thanks @mrambacher ! Could you remove that `ROCKSDB_LIBRARY_API` from the .cc file and add an entry to the changelog? Other than that, LGTM.

PR #8150, opened 2021-04-05
Fix flush reason attribution

ot:
Current flush reason attribution is misleading or incorrect (depending on what the original intention was):

- Flush due to WAL reaching its maximum size is attributed to `kWriteBufferManager`
- Flushes due to full write buffer and write buffer manager are not distinguishable, both are attributed to `kWriteBufferFull`

This changes the first to a new flush reason `kWALFull`, and splits the second between `kWriteBufferManager` and `kWriteBufferFull`.

### DIFF
--- a/db/db_impl/db_impl_write.cc
+++ b/db/db_impl/db_impl_write.cc
@@ -937,7 +937,7 @@ Status DBImpl::PreprocessWrite(const WriteOptions& write_options,
     // be flushed. We may end up with flushing much more DBs than needed. It's
     // suboptimal but still correct.
     WaitForPendingWrites();
-    status = HandleWriteBufferFull(write_context);
+    status = HandleWriteBufferManagerFlush(write_context);
   }
 
   if (UNLIKELY(status.ok() && !trim_history_scheduler_.Empty())) {
@@ -1348,20 +1348,20 @@ Status DBImpl::SwitchWAL(WriteContext* write_context) {
       if (!immutable_db_options_.atomic_flush) {
         FlushRequest flush_req;
         GenerateFlushRequest({cfd}, &flush_req);
-        SchedulePendingFlush(flush_req, FlushReason::kWriteBufferManager);
+        SchedulePendingFlush(flush_req, FlushReason::kWalFull);
       }
     }
     if (immutable_db_options_.atomic_flush) {
       FlushRequest flush_req;
       GenerateFlushRequest(cfds, &flush_req);
-      SchedulePendingFlush(flush_req, FlushReason::kWriteBufferManager);
+      SchedulePendingFlush(flush_req, FlushReason::kWalFull);
     }
     MaybeScheduleFlushOrCompaction();
   }
   return status;
 }
 
-Status DBImpl::HandleWriteBufferFull(WriteContext* write_context) {
+Status DBImpl::HandleWriteBufferManagerFlush(WriteContext* write_context) {
   mutex_.AssertHeld();
   assert(write_context != nullptr);
   Status status;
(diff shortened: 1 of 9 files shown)


include/rocksdb/listener.h
```diff
@@ -118,6 +118,7 @@ enum class FlushReason : int {
   // When set the flush reason to kErrorRecoveryRetryFlush, SwitchMemtable
   // will not be called to avoid many small immutable memtables.
   kErrorRecoveryRetryFlush = 0xc,
+  kWALFull = 0xd,
```
pdillinger, line 121: Google C++ style calls for `kWalFull`

ot: I spelled it consistently with other names in the file, for example

https://github.com/facebook/rocksdb/blob/a951b38e9de962eb9e9a41bdedbbe8a8743c6519/include/rocksdb/listener.h#L134

But I can change it if there's a preference.

pdillinger: Yes, perhaps the most egregious violator is class `DB`. I think we're trying to follow the Google style on new code though.

    $ git grep '[^a-zA-Z]kWal' | wc -l
    78
    $ git grep '[^a-zA-Z]kWAL' | wc -l
    17

riversand963:
The HISTORY.md has to be updated since it's change to public API.

riversand963:
> @riversand963 has imported this pull request. If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D27569645).

Oh, just realize you can also import it. Feel free to re-import and commandeer the resulting diff. @ot

ot:
Sounds good, I'll make the requested changes and re-import

pdillinger (commented):
I think this is correct, but I think that tests checking the flush reason are needed to be confident.

Are you willing to add the checks (should be addable to existing tests) or want to hand this off to us?

db/db_impl/db_impl_write.cc
```diff
@@ -1434,13 +1434,13 @@ Status DBImpl::HandleWriteBufferFull(WriteContext* write_context) {
       if (!immutable_db_options_.atomic_flush) {
         FlushRequest flush_req;
         GenerateFlushRequest({cfd}, &flush_req);
-        SchedulePendingFlush(flush_req, FlushReason::kWriteBufferFull);
+        SchedulePendingFlush(flush_req, FlushReason::kWriteBufferManager);
```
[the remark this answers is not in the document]
ot: Yes I hadn't seen this test, should be easy to plug the check there. Doing it now. I'll see if we can also cover kWriteBufferFull somehow.

Note that the existing test for full WAL was just sanctioning the existing behavior, I'm curious how this wasn't noticed when that check was added.

I'll also rename the function (I assume that name is there for legacy reasons, and that's how the wrong flush reason ended up there). I can also fix the wording in the logs, is there any concern about anything parsing those logs?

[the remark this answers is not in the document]
ot: @pdillinger do you want to give another look?

db/db_test_util.h
```diff
@@ -761,6 +761,17 @@ class OnFileDeletionListener : public EventListener {
   size_t matched_count_;
   std::string expected_file_name_;
 };
+
+class FlushCounterListener : public EventListener {
+ public:
+  std::atomic<int> count{0};
```
riversand963, line 767: Very minor: these can be non-public, and naming convention is `count_` and `expected_flush_reason_`.

riversand963 (approved):
Thanks @ot for the fix. LGTM with a minor comment. Will defer to @pdillinger to accept.

db/db_test_util.h
ot: They need to be public, the tests access them directly. I can add setters/getters but it seems overkill.

db/db_impl/db_impl_write.cc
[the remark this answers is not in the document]
pdillinger: Not concerned about log parsing. It's better to have language that is not misleading.

pdillinger:
LGTM :)  Thanks for noticing and fixing.

I will trust your test additions fail if any of the (non-atomic-flush) reasons is changed.

PR #8151, opened 2021-04-05
Likely fix flaky TableFileCorruptedBeforeBackup

pdillinger:
Summary: Before corrupting a file in the DB and expecting corruption to
be detected, open DB read-only to ensure file is not made obsolete by
compaction. Also, to avoid obsolete files not yet deleted, only select
live files to corrupt.

Test Plan: watch CI


akankshamahajan15 (approved):
LGTM

PR #8152, opened 2021-04-06
Local compaction service

mrambacher:
This is a first pass at introducing a local compaction service.  This change adds a LocalCompactionService to the db_impl.  This LocalCompactionService contains all of the options to the compaction job which do not change from one invocation to the next.  The CompactionJob now uses this "service" for to obtain access to the variables that were previously passed to its constructor.

The CompactionProxy class used by the CompactionIterator now has methods to deal with Blobs.  This eliminates the need for the "Version" to be part of the API of the Proxy.  Note that a "Fake" CompactionIterator cannot deal with Blobs currently.

The "compact_" member variable inside of CompactionJob was renamed to "compaction_state".  The variable is an instance of a CompactionState and there is another Compact class, leading to confusion as to what this variable represents.

Note that the LocalCompactionService is not currently a service but a container for options.  Future changes can make the LocalCompactionService into a worker service.


mrambacher:
Abandoning this PR.  Was implemented differently by Jay.  This was a prototype.

PR #8155, opened 2021-04-06
Update ZStd. Fixes an issue with Make 3.82

adamretter:
The previous version of ZStd doesn't build correctly with Make 3.82. Updating it resolves the issue.

@jay-zhuang This also needs to be cherry-picked to:
1. 6.17.fb
2. 6.18.fb
3. 6.19.fb


ajkr (approved):
LGTM

PR #8156, opened 2021-04-06
Remove check for status returned by `InvalidatePageCache`

### LINKED ISSUE #7600: A failed `SstFileWriter::Add()` should not apply any internal state changes
#7448 accidentally changed the API contract of `SstFileWriter::Add()` and dependent functions (`SstFileWriter::Put()`, etc.). Now there is a case where the KV may be added to the table builder _and_ the returned `Status` is non-OK, i.e., when the `InvalidatePageCache()` fails. If there are clients that retry, they may now attempt to add an already existing key to a table builder. Ideally, clients shouldn't have been retrying since the pre-existing non-OK `Status`es were all `InvalidArgument`, on which retry would never succeed. However, in practice we have found some clients had implemented retry logic anyways, and after this change, the following sequence is possible:

1. `Add()` successfully adds a kv to the table builder.
2. `Add()` fails on `InvalidatePageCache()` and returns non-OK `Status`.
3. The retrying client calls `Add()` again with the same kv
4. `Add()` fails on validation of arguments (user key not strictly ascending)

Prior to this change, a failure in `InvalidatePageCache()` could not fail `Add()`, so the client would just move on to the next key.

sunby:
Summary:
Failures in `InvalidatePageCache` will change the API contract. So we remove the status check for `InvalidatePageCache` in `SstFileWriter::Add()`, `SstFileWriter::Finish` and `Rep::DeleteRange`

Fixes #7600.

### DIFF
--- a/table/sst_file_writer.cc
+++ b/table/sst_file_writer.cc
@@ -99,7 +99,8 @@ struct SstFileWriter::Rep {
     file_info.largest_key.assign(user_key.data(), user_key.size());
     file_info.file_size = builder->FileSize();
 
-    return InvalidatePageCache(false /* closing */);
+    InvalidatePageCache(false /* closing */).PermitUncheckedError();
+    return Status::OK();
   }
 
   Status DeleteRange(const Slice& begin_key, const Slice& end_key) {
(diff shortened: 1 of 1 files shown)


riversand963 (changes requested):
I believe errors should be checked and handled. Could you point out which contract is broken? I am looking at the callers of `InvalidatePageCache`, and I do not seem to find a comment about the return status.

riversand963:
okay, I saw the discussion thread in #7600. Will defer to @ajkr for further review.
Curious whether it's possible to differentiate the error here from others.
Or at least call `PermitUnchecked` as `InvalidatePageCache(..).PermitUncheckedError()`.

riversand963 (commented):
Thanks @sunby for the fix

ajkr (approved):
LGTM, unless anyone has other ideas how to handle this case.

riversand963 (approved):

PR #8158, opened 2021-04-06
Update installation instructions

shoda-tibco:
Updated instructions for installing zstd on CentOS and note about clang-format dependency

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -96,9 +96,13 @@ to build a portable binary, add `PORTABLE=1` before your make commands, like thi
               sudo yum install libasan
 
     * Install zstandard:
-        * With [EPEL](https://fedoraproject.org/wiki/EPEL) or CentOS 8:
+        * With [EPEL](https://fedoraproject.org/wiki/EPEL):
 
               sudo yum install libzstd-devel
+
+        * With CentOS 8:
+
+              sudo dnf install libzstd-devel
 
         * From source:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -43,6 +43,8 @@ to build a portable binary, add `PORTABLE=1` before your make commands, like thi
       command line flags processing. You can compile rocksdb library even
       if you don't have gflags installed.
 
+* `make check` will also check code formatting, which requires [clang-format](https://clang.llvm.org/docs/ClangFormat.html)
+
 * If you wish to build the RocksJava static target, then cmake is required for building Snappy.
 
 ## Supported platforms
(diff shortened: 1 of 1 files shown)


ajkr (approved):
LGTM

INSTALL.md
```diff
@@ -94,12 +96,17 @@ to build a portable binary, add `PORTABLE=1` before your make commands, like thi
               sudo yum install libasan
 
     * Install zstandard:
+        * With [EPEL](https://fedoraproject.org/wiki/EPEL) or CentOS 8:
+
+              sudo yum install libzstd-devel
```
riversand963, line 101: I am under the impression that `dnf` is preferred on CentOS 8. https://centos.pkgs.org/8/elrepo-testing-x86_64/zstd-1.3.4-1.el8.elrepo.x86_64.rpm.html

riversand963: Tested on my CentOS 8 box with the following
```
dnf install libzstd-devel
```

shoda-tibco: `yum` works as well with CentOS 8, and `dnf` is not mentioned elsewhere in this doc for CentOS/RHEL. Also it appears `yum` is symlinked to `dnf` on CentOS 8:

```
[root@6ae6dc102717 /]# ls -laF $(command -v yum)
lrwxrwxrwx 1 root root 5 Aug  4  2020 /usr/bin/yum -> dnf-3*
[root@6ae6dc102717 /]# ls -laF $(command -v dnf)
lrwxrwxrwx 1 root root 5 Aug  4  2020 /usr/bin/dnf -> dnf-3*
```

adamretter: I would suggest to stick with Yum as it is probably the most used and can is operable on all versions of CentOS.

riversand963: > Also it appears `yum` is symlinked to `dnf` on CentOS 8:

Not on one of my machines, but true on the other. So I think this assumption does not hold.
On the first CentOS 8 machine, I tried to use yum install, but I got error "yum not found".

FWIW, I think the doc should try to be up-to-date with what the OS recommends. dnf is going to replace yum, making this doc out-dated again.

riversand963 (changes requested):
After repeating the "yum" command to install libzstd-devel on my CentOS 8 machine, I think we can not assume yum is available on CentOS 8. It may be available on some for compatibility reasons, but a counter example clearly exists.

INSTALL.md
riversand963: > I would suggest to stick with Yum as it is probably the most used and is operable on all versions of CentOS.

Unless the CentOS 8 in the above line is removed.

shoda-tibco: Ok broke out CentOS 8 on its own here with `dnf`.

riversand963 (approved):

riversand963:
Thanks @shoda-tibco for improving the documentation!

INSTALL.md
ajkr: [unrelated to this specific PR] Speaking of out-dated, our suggestion to install ZSTD 1.1.3 on non-CentOS 8 could be updated as well. I wonder if there's a URL for the latest release code.

shoda-tibco: I don't know of any direct link to the latest release source on github, partly because zstd puts the release tag in the asset name. There are lots of examples on scripted solutions that use curl to query the github api and parse out the latest release tag using jq/sed/grep whatever, and then create a download link from that, e.g.: https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c

That could be an option, assuming the latest-in-time zstd is always what folks should use.

PR #8161, opened 2021-04-07
Fix crash test with backup as read-only DB

pdillinger:
Summary: Forgot to re-test crash test after adding read-only filesystem
enforcement to #8142. The problem is ReadOnlyFileSystem would reject
CreateDirIfMissing whenever DBOptions::create_if_missing=true. The fix
that is better for users is to allow CreateDirIfMissing in
ReadOnlyFileSystem if the directory exists, so that they don't cause a
failure on using create_if_missing with opening backups as read-only
DBs. Added this option test to the unit test (in addition to being in the
crash test).

Also fixed a couple of lints.

And some better messaging from 'make format' so that when you run it
with uncommitted changes, it's clear that it's only checking the
uncommitted changes.

Test Plan: local blackbox_crash_test with amplified backup_one_in

### DIFF
--- a/build_tools/format-diff.sh
+++ b/build_tools/format-diff.sh
@@ -136,9 +136,11 @@ then
   FORMAT_UPSTREAM_MERGE_BASE="$(git merge-base "$FORMAT_UPSTREAM" HEAD)"
   # Get the differences
   diffs=$(git diff -U0 "$FORMAT_UPSTREAM_MERGE_BASE" | $CLANG_FORMAT_DIFF -p 1)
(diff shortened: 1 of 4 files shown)


pdillinger:
```
$ make format
$DEBUG_LEVEL is 1
Makefile:203: Warning: Compiling in debug mode. Don't use the resulting binary in production
build_tools/format-diff.sh
Nothing needs to be reformatted!
```

ajkr (approved):
LGTM

zhichao-cao (approved):
Thanks for the fix. LGTM.

PR #8162, opened 2021-04-07
Better check should merge

darionyaphet:


ajkr (commented):
It looks too similar to me. Do you want to try searching the "up-for-grabs" label in our GitHub issues and see if there's any problem that interests you?

darionyaphet:
> It looks too similar to me. Do you want to try searching the "up-for-grabs" label in our GitHub issues and see if there's any problem that interests you?

Hello, thank you for your reply.  It is my pleasure and you mean I can try to fix some problems?

ajkr:
Yes, that'd be great if you have time and find something of interest in the up-for-grabs issues list.

PR #8163, opened 2021-04-07
Event listener overwrite enumerate class

darionyaphet:


mrambacher (commented):
A couple of general comments:
- I think a ToString(enum) method would be better than this operator.  The ToString has potentially more uses
- There are a couple of other examples of how RocksDB converts enums to/from Strings.  The Options code has one mechanism and another was proposed in #7081.  It might be nice to have some standard in RocksDB on enum/string conversions

PR #8164, opened 2021-04-07
Fix read-only DB writing to filesystem with write_dbid_to_manifest

pdillinger:
Summary: Fixing another crash test failure in the case of
write_dbid_to_manifest=true and reading a backup as read-only DB.

Test Plan: enhanced unit test for backup as read-only DB, ran
blackbox_crash_test more with elevated backup_one_in


zhichao-cao (approved):
LGTM, thanks for the fix!

PR #8168, opened 2021-04-08
Add util/crc32c_arm64.cc to TARGETS

xavierd:
When compiling RocksDB with Buck for ARM64, the linker complains about missing crc32 symbols that are defined in the crc32c_arm64.cc file. Since this file wasn't included in the build this is totally expected


zhichao-cao (approved):
LGTM

PR #8170, opened 2021-04-09
Misc Backup API enhancements

pdillinger:
Summary:
* CreateNewBackup(WithMetadata) returning the BackupID of new backup
through optional new output param. This is especially useful with the
new mutithreading support, so that you can transactionally determine the
ID of a backup you create.
* GetBackupInfo / GetLatestBackupInfo for individual backups, so that
you don't have to comb through a vector of backups if you don't want to.

Updated HISTORY.md (including re: BlobDB support as new feature)

Test Plan: Added test logic to existing tests, to minimize increase in
cost of running tests


akankshamahajan15 (approved):
LGTM!

ajkr (approved):
LGTM too!

PR #8171, opened 2021-04-09
Add internal compaction API for Secondary instance

jay-zhuang:
Add compaction API for secondary instance, which compact the files to a secondary DB path without installing to the LSM tree.
The API will be used to remote compaction.

Test plan: `make check`


db/db_impl/db_impl_secondary.cc
```diff
@@ -17,8 +17,9 @@ namespace ROCKSDB_NAMESPACE {
 
 #ifndef ROCKSDB_LITE
 DBImplSecondary::DBImplSecondary(const DBOptions& db_options,
-                                 const std::string& dbname)
-    : DBImpl(db_options, dbname) {
+                                 const std::string& dbname,
+                                 const std::string& secondary_path)
+    : DBImpl(db_options, dbname), secondary_path_(secondary_path) {
```
ajkr, line 22: (unrelated to this PR) I think `DBImpl` should be passed `read_only=true`. That's a new flag introduced in #8142 to prevent `SanitizeOptions()` from creating an info logger when none is provided (i.e., `info_log == nullptr`). I'd guess in a secondary DB, a `SanitizeOptions()`-created info logger could conflict with the primary DB info logger. It looks like `OpenAsSecondary()` tries to avoid opening the DB with option `info_log == nullptr` -- except when  L616 is executed.

db/db_impl/db_impl_secondary.cc
```diff
@@ -663,6 +664,103 @@ Status DB::OpenAsSecondary(
...
+  c->SetInputVersion(version);
+
+  // Create output directory if it's not existed yet
+  std::unique_ptr<FSDirectory> output_dir;
+  s = CreateAndNewDirectory(fs_.get(), secondary_path_, &output_dir);
+  if (!s.ok()) {
+    return s;
+  }
+
+  LogBuffer log_buffer(InfoLogLevel::INFO_LEVEL,
```
ajkr, line 722: Does this need to be flushed somewhere? I don't know why `CompactionJob` requires this, but it feels surprising if the client is expected to own it but doesn't need to do anything with it after the compaction finishes.

ajkr (approved):
LGTM, good to see `CompactionJob` can support local execution and remote execution. What will the primary side look like during remote execution? Do we also create a `CompactionJob` there and invoke some subset of its operations?

db/db_impl/db_impl_secondary.cc
```diff
@@ -17,8 +17,9 @@ namespace ROCKSDB_NAMESPACE {
 
 #ifndef ROCKSDB_LITE
 DBImplSecondary::DBImplSecondary(const DBOptions& db_options,
-                                 const std::string& dbname)
-    : DBImpl(db_options, dbname) {
+                                 const std::string& dbname,
+                                 const std::string& secondary_path)
```
riversand963, line 21: Nit: since we are going to store `secondary_path_` as a copy, we can make `secondary_path` of type `std::string`, and do `secondary_path_(std::move(secondary_path))`.

db/db_impl/db_impl_secondary.cc
```diff
@@ -663,6 +664,103 @@ Status DB::OpenAsSecondary(
...
+  if (!s.ok()) {
+    return s;
+  }
+
+  LogBuffer log_buffer(InfoLogLevel::INFO_LEVEL,
+                       immutable_db_options_.info_log.get());
+
+  // TODO: there's only 1 job could be run per DB secondary instance, add
+  // validation for that.
+  const int job_id = 1;
```
riversand963, line 727: Can we still use `next_job_id_`?
> TODO: there's only 1 job could be run per DB secondary instance

I am not sure whether you mean "only 1 job ID" or "only 1 job", because they do not seem to contradict with each other.

riversand963 (approved):
LGTM. Thanks @jay-zhuang for the PR

ajkr:
> What will the primary side look like during remote execution? Do we also create a CompactionJob there and invoke some subset of its operations?

Overall I'm finding it difficult to wrap my head around the different modes of `CompactionJob` (on primary w/ local execution, on primary w/ remote execution, and remote execution). Unless I'm missing something, it seems both the `CompactionJob` client and the `CompactionJob` itself need to behave differently in these different modes. The client needs to call the right subset of `CompactionJob` functions according to the mode, and the functions themselves need to behave differently according to the mode.

Have we tried prototyping `ProcessKeyValueCompaction()` as the unit of work that can be relocated to a remote worker? I still suspect keeping control logic all on the primary can help. I recall we found execution logic residing outside of `ProcessKeyValueCompaction()` (i.e., fsync'ing parent directory and verifying output tables). After thinking through it more, it makes sense to me to do the table verification in `ProcessKeyValueCompaction()`, and that function can fsync the secondary_path directory too when run remotely.

db/compaction/compaction_job.cc
```diff
@@ -899,6 +910,35 @@ Status CompactionJob::Install(const MutableCFOptions& mutable_cf_options) {
...
+  compaction_result->output_level = compact_->compaction->output_level();
+  compaction_result->output_path = output_path_;
+  // Compaction service should not have sub compaction
+  assert(compact_->sub_compact_states.size() == 1);
+  SubcompactionState& subcompaction_state = compact_->sub_compact_states[0];
+
+  for (auto& output_file : subcompaction_state.outputs) {
+    auto& meta = output_file.meta;
+    compaction_result->output_files.emplace_back(
+        MakeTableFileName(".", meta.fd.GetNumber()), meta.fd.smallest_seqno,
```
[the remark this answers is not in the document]
jay-zhuang: I guess we don't need the "." here.

db/compaction/compaction_job.h
```diff
@@ -57,6 +57,81 @@ class Version;
...
 class VersionSet;
 
+// CompactionServiceInput is used the pass compaction information between two
+// db instances. It contains the information needed to do a compaction. It
+// doesn't contain the LSM tree information, which is passed though MANIFEST
+// file.
+struct CompactionServiceInput {
+  std::string column_family_name;
+
+  DBOptions db_options;
```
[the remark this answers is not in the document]
jay-zhuang: It has column_family_name and `ColumnFamilyOptions`. I'll combine them with `ColumnFamilyDescriptor`.
I don't think we need vector here, as compaction is only for 1 column family.

db/db_impl/db_impl_secondary.cc
jay-zhuang: Thanks for pointing it out, I need look into it to understand the problem, filed an issue to myself: #8193

db/db_impl/db_impl_secondary.cc
jay-zhuang: Isn't it flushed in `CompactionJob::Run()`: https://github.com/facebook/rocksdb/blob/d89483098ff4a77bd0ef12454a64554805984cc5/db/compaction/compaction_job.cc#L586

jay-zhuang:
> What will the primary side look like during remote execution? Do we also create a CompactionJob there and invoke some subset of its operations?

Yes, the CompactionJob will be created and then `Prepare()` -> `Run()` -> `Install()`. The remote_compaction part will be handled inside of `Run()`, instead of calling `ProcessKeyValueCompaction()`, it will call the CompactionService callback.

> Overall I'm finding it difficult to wrap my head around the different modes of `CompactionJob` (on primary w/ local execution, on primary w/ remote execution, and remote execution). Unless I'm missing something, it seems both the `CompactionJob` client and the `CompactionJob` itself need to behave differently in these different modes. The client needs to call the right subset of `CompactionJob` functions according to the mode, and the functions themselves need to behave differently according to the mode.

Yeah, I agree, I updated the patch to have a `CompactionServiceCompactionJob` for remote side, which differentiate itself from primary's `CompactionJob` and also group things together.

> Have we tried prototyping `ProcessKeyValueCompaction()` as the unit of work that can be relocated to a remote worker? I still suspect keeping control logic all on the primary can help. I recall we found execution logic residing outside of `ProcessKeyValueCompaction()` (i.e., fsync'ing parent directory and verifying output tables). After thinking through it more, it makes sense to me to do the table verification in `ProcessKeyValueCompaction()`, and that function can fsync the secondary_path directory too when run remotely.

I updated it to have `CompactionServiceCompactionJob` to call `ProcessKeyValueCompaction()` directly instead of using the same `Prepare()` and `Run()`. It does make it cleaner.

db/compaction/compaction_job.h
```diff
@@ -194,19 +202,130 @@ class CompactionJob {
...
+  // location of the output files
+  std::string output_path;
+
+  // some statistics about the compaction
+  uint64_t num_output_records;
+  uint64_t total_bytes;
+  uint64_t approx_size;
+  uint64_t bytes_read;
+  uint64_t bytes_written;
+  CompactionJobStats stats;
```
ajkr, line 289: Is it used?

ajkr (approved):
LGTM! Nice separation of the remote compaction work into a dedicated class with a more targeted API.

PR #8172, opened 2021-04-10
Assert unlimited max_open_files for FIFO compaction.

### LINKED ISSUE #8014: creation time can not be read

### Expected behavior

i am using JNI the FIFO compaction and snapshot sst files to hdfs and recover it from previous run, looks like the SST file lost the creation time table property, can any body suggest how can i fix this?

2021/03/01-18:51:01.202049 7f59042fc700 (Original Log Time 2021/03/01-18:51:01.200883) [/compaction/compaction_picker_fifo.cc:107] [_timer_state/processing_user-timers] FIFO compaction: picking file 1710 with creation time 0 for deletion

 

the configuration is like 

currentOptions.setCompactionStyle(getCompactionStyle());
currentOptions.setLevel0FileNumCompactionTrigger(8);
CompactionOptionsFIFO compactionOptionsFIFO = new CompactionOptionsFIFO();
compactionOptionsFIFO.setMaxTableFilesSize(MemorySize.parse("8gb").getBytes());
compactionOptionsFIFO.setAllowCompaction(true);

 

the rocksdb version is 

<dependency>
<groupId>io.github.myasuka</groupId>
<artifactId>frocksdbjni</artifactId>
<version>6.10.2-ververica-3.0</version>
</dependency>

thejchap:
Resolves #8014

- Add an assertion on `DB::Open` to ensure `db_options.max_open_files` is unlimited if FIFO Compaction is being used.
- This is to align with what the docs mention and to prevent premature data deletion.
- Update tests to work with this assertion.

Test Plan:
```bash
$ make check -j$(nproc)

Generated TARGETS Summary:
- 6 libs
- 0 binarys
- 180 tests
```

### DIFF
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -1355,6 +1355,12 @@ Status ColumnFamilyData::ValidateOptions(
         "[0.0, 1.0].");
   }
 
+  if (cf_options.compaction_style == kCompactionStyleFIFO &&
+      db_options.max_open_files != -1 && cf_options.ttl > 0) {
+    return Status::NotSupported(
+        "FIFO compaction only supported with max_open_files = -1.");
+  }
+
   return s;
 }
 
--- a/utilities/option_change_migration/option_change_migration_test.cc
+++ b/utilities/option_change_migration/option_change_migration_test.cc
@@ -54,7 +54,9 @@ TEST_P(DBOptionChangeMigrationTests, Migrate1) {
   if (old_options.compaction_style == CompactionStyle::kCompactionStyleLevel) {
     old_options.level_compaction_dynamic_level_bytes = is_dynamic1_;
   }
-
+  if (old_options.compaction_style == CompactionStyle::kCompactionStyleFIFO) {
+    old_options.max_open_files = -1;
+  }
   old_options.level0_file_num_compaction_trigger = 3;
   old_options.write_buffer_size = 64 * 1024;
   old_options.target_file_size_base = 128 * 1024;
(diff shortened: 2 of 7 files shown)


ajkr (commented):
Looks good so far. Do you mind also mentioning this under "Bug Fixes" section in the "HISTORY.md" file?

It looks like some tests relied on the now unsupported combination of options. You can try running `make check -j$(nproc)` locally (if that fails and it's not obvious what caused it, try `make watch-log`) to find all such cases.

utilities/option_change_migration/option_change_migration_test.cc
```diff
@@ -49,6 +49,7 @@ class DBOptionChangeMigrationTests
 #ifndef ROCKSDB_LITE
 TEST_P(DBOptionChangeMigrationTests, Migrate1) {
   Options old_options = CurrentOptions();
+  old_options.max_open_files = -1;
```
ajkr, line 52: Does it work to only make this change when `compaction_style == CompactionStyle::kCompactionStyleFIFO`? We'd probably also need to conditionally modify `new_options` below if we conditionally modified `old_options` here.

ajkr (approved):
Looks great!

utilities/option_change_migration/option_change_migration_test.cc
thejchap: @ajkr yep! will do. i planned to ping you on this one as i wasn't super confident that this particular change was in the right place heh

db/column_family.cc
```diff
@@ -1355,6 +1355,12 @@ Status ColumnFamilyData::ValidateOptions(
         "[0.0, 1.0].");
   }
 
+  if (cf_options.compaction_style == kCompactionStyleFIFO &&
+      db_options.max_open_files != -1 && cf_options.ttl > 0) {
```
mrambacher, line 1359: Can you explain the ttl > 0?  If ttl == 0, is it okay if max_open_files != -1 ?

ajkr: As far as we know, the `max_open_files == -1` is only needed to ensure the files' creation_time properties are read, which are used for deciding to trigger TTL compaction. When `ttl == 0`, TTL compaction is disabled.

PR #8173, opened 2021-04-11
FlushMemTable return ok but memtable does not synchronize flush

### LINKED ISSUE #8046: When I set flush_options.wait = 1, FlushMemTable return ok but memtable does not synchronize flush

### Expected behavior
FlushMemTable return failed 
### Actual behavior
FlushMemTable return ok
1、In my environment ：db_options_.paranoid_checks = ture and auto_recovery_ = true
2、One compaction thread encountered a problem of insufficient disk space, so it set SetBGError. The bg_error_ will be set kSoftError.
3、Another compaction thread also encountered a problem of insufficient disk space, so it set SetBGError . The recovery_error will be set.
4、My code set flush_options.wait = 1  then call  DBImpl::Flush .
5、FlushMemTable  return ok but flush memtable thread failed with an insufficient disk space error.

### Where is the bug I guess
image

zhuchong0329:
Fix #8046 : FlushMemTable return ok but memtable does not synchronize flush. The way to fix it is to expose RecoveryError.


mrambacher (changes requested):
I think this change is going to trigger a lot of failures in the "ASSERT_STATUS_CHECKED" cases.

A better approach may be to leave the code around line 2127 alone and add/copy line 2130-31 to around 2165.

zhuchong0329:
@mrambacher , Thanks for the reminder. I did ignore ROCKSDB_ASSERT_STATUS_CHECKED.

ajkr:
@zhuchong0329 this seems like a useful fix. Do you plan to try making it work with ASSERT_STATUS_CHECKED? I think you can get rid of all the early returns (e.g., L2125) and instead make them set `s` to a non-OK Status. Then add `s.ok()` to all the conditionals/loop conditions to make sure nothing meaningful executes and nothing overwrites `s` for the rest of the function.

riversand963:
What's our plan with this PR? @jay-zhuang

ajkr (approved):
LGTM too.

PR #8174, opened 2021-04-11
Add CreateFrom methods to Env/FileSystem

mrambacher:
- Added CreateFromString method to Env and FilesSystem to replace LoadEnv/Load.  This method/signature is a precursor to making these classes extend Customizable.

- Added CreateFromSystem to Env.  This method standardizes creating an Env from the environment variables.  Previously, some places would check TEST_ENV_URI and others would also check TEST_FS_URI.  Now the code is more command/standardized.

- Added CreateFromFlags to Env.  These method allows Env to be create from string options (such as GFLAGS options) in a more standard way.


env/env.cc
```diff
@@ -597,18 +606,29 @@ Status Env::LoadEnv(const std::string& value, Env** result) {
...
   std::unique_ptr<Env> uniq_guard;
   std::string err_msg;
   assert(guard != nullptr);
+  (void)config_options;
   env = ObjectRegistry::NewInstance()->NewObject<Env>(value, &uniq_guard,
                                                       &err_msg);
   if (!env) {
-    s = Status::NotFound(std::string("Cannot load ") + Env::Type() + ": " +
-                         value);
+    s = Status::NotSupported(std::string("Cannot load ") + Env::Type() + ": " +
```
zhichao-cao, line 630: Do we need to change the returned status from NotFound to NotSpported？Or you want to align with the other classes' CreateFromString return status?

env/env.cc
```diff
@@ -618,13 +638,63 @@ Status Env::LoadEnv(const std::string& value, Env** result,
...
 #else
+  (void)config_options;
   (void)result;
   (void)guard;
   s = Status::NotSupported("Cannot load environment in LITE mode", value);
 #endif
   return s;
 }
 
+Status Env::CreateFromSystem(const ConfigOptions& config_options, Env** result,
```
zhichao-cao, line 649: CreateFromSystem is a hard to understand. Maybe use a better name?

env/env.cc
```diff
@@ -618,13 +638,63 @@ Status Env::LoadEnv(const std::string& value, Env** result,
...
+    return Status::OK();
+  }
+}
+
+Status Env::CreateFromFlags(const ConfigOptions& config_options,
+                            const std::string& env_uri,
+                            const std::string& fs_uri, Env** result,
+                            std::shared_ptr<Env>* guard) {
+  *result = config_options.env;
+  if (env_uri.empty() && fs_uri.empty()) {
```
zhichao-cao, line 678: Can we reuse the code of CreateFromSystem. The logic is almost the same.

table/sst_file_reader_test.cc
```diff
@@ -37,14 +39,7 @@ class SstFileReaderTest : public testing::Test {
     sst_name_ = test::PerThreadDBPath("sst_file");
 
     Env* base_env = Env::Default();
-    const char* test_env_uri = getenv("TEST_ENV_URI");
```
zhichao-cao, line 40: I think we should use CreateFromFlags here and specify the input. Instead of give CreateFromSystem. We may not need CreateFromSystem as the API.

env/env.cc
mrambacher: I wanted to make it align with the other methods that return NotSupported if an instance could not be created.

env/env.cc
mrambacher: I am open to a better name.  Got any suggestions?  I was going to call it CreateFromEnv but that is a little misleading as well...

env/env.cc
```diff
@@ -618,13 +638,63 @@ Status Env::LoadEnv(const std::string& value, Env** result,
...
+    }
+    return s;
+  } else {
+    // Neither specified.  Use the default
+    guard->reset();
+    return Status::OK();
+  }
+}
+
+Status Env::CreateFromFlags(const ConfigOptions& config_options,
```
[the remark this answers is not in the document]
mrambacher: If you like that name better, sure.  I used "Flags" in deference to "GFLAGS" or OptionFlags, but was also debating about "CreateFromArgv" or something like that. 

I am open to renaming both of the CreateFrom methods...

env/env.cc
mrambacher: Done.

env/env.cc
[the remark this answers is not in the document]
zhichao-cao: Maybe "CreateFromArgv" is better.

env/env.cc
zhichao-cao: I mean, remove the CreateFromSystem method. Using CreateFromArgv or others to achieve the same function. I cannot see the essential of introducing CreateFromSystem.

mrambacher: Moved it to test::CreateEnvFromSystem and out of the public API

tools/db_bench_tool.cc
```diff
@@ -7666,23 +7658,20 @@ int db_bench_tool(int argc, char** argv) {
...
+                                  &FLAGS_env, &env_guard);
+    if (!s.ok()) {
+      fprintf(stderr, "Failed creating env: %s\n", s.ToString().c_str());
       exit(1);
     }
-    FLAGS_env = GetCompositeEnv(fs);
   } else if (FLAGS_simulate_hybrid_fs_file != "") {
-    FLAGS_env = GetCompositeEnv(std::make_shared<SimulatedHybridFileSystem>(
-        FileSystem::Default(), FLAGS_simulate_hybrid_fs_file));
+    //**TODO: Make the siimulated fs something that can be loaded
```
zhichao-cao, line 7669: nit: siimulated->simulate

tools/sst_dump_tool.cc
```diff
@@ -354,32 +346,17 @@ int SSTDumpTool::Run(int argc, char const* const* argv, Options options) {
...
-        exit(1);
-      }
-      env = GetCompositeEnv(fs);
+  {
+    ConfigOptions config_options;
+    config_options.env = options.env;
+    Status s = Env::CreateFromUri(config_options, env_uri, fs_uri, &options.env,
+                                  &env_guard);
+    if (!s.ok()) {
+      fprintf(stderr, "CreateEnv: %s\n", s.ToString().c_str());
```
zhichao-cao, line 355: CreateEnv: ->CreateFromUri:

zhichao-cao (approved):
Thanks for addressing the review comments! Just left some minor suggestions. Good to go.

PR #8179, opened 2021-04-12
Extend file_checksum_dump ldb command and DB::GetLiveFilesChecksumInfo to blob files

akankshamahajan15:
Summary: Extend the DB::GetLiveFilesChecksumInfo API to blob files.
This API is also used by the file_checksum_dump ldb command to dump checksum
of SST files which now also dumps blob files checksum.

Test Plan: Add new unit test

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -4954,6 +4954,7 @@ Status VersionSet::GetLiveFilesChecksumInfo(FileChecksumList* checksum_list) {
     if (cfd->IsDropped() || !cfd->initialized()) {
       continue;
     }
+    /* SST files */
     for (int level = 0; level < cfd->NumberLevels(); level++) {
       for (const auto& file :
            cfd->current()->storage_info()->LevelFiles(level)) {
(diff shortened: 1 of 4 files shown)


db/version_set.cc
```diff
@@ -4954,24 +4954,44 @@ Status VersionSet::GetLiveFilesChecksumInfo(FileChecksumList* checksum_list) {
...
+    /* Blob files */
+    const auto& blob_files = cfd->current()->storage_info()->GetBlobFiles();
+    for (const auto& pair : blob_files) {
+      const uint64_t blob_file_number = pair.first;
+      const auto& meta = pair.second;
+
+      assert(meta);
+      assert(blob_file_number == meta->GetBlobFileNumber());
+
+      std::string checksum_value = meta->GetChecksumValue();
```
riversand963, line 4979: Minor: can we avoid creating `std::string` here? We will create a copy anyway in `InsertOneFileChecksum()`.

tools/ldb_cmd_test.cc
```diff
@@ -349,6 +349,83 @@ TEST_F(LdbCmdTest, DumpFileChecksumNoChecksum) {
...
+  std::string dbname = test::PerThreadDBPath(env.get(), "ldb_cmd_test");
+  ASSERT_OK(DB::Open(opts, dbname, &db));
+
+  WriteOptions wopts;
+  FlushOptions fopts;
+  fopts.wait = true;
+  Random rnd(test::RandomSeed());
+  for (int i = 0; i < 200; i++) {
+    char buf[16];
+    snprintf(buf, sizeof(buf), "%08d", i);
```
riversand963, line 370: Nit: maybe we can use the formatting functionality available in `iomanip`. We have some discussion about migrating away from c-array and snprintf (https://github.com/facebook/rocksdb/pull/8137#discussion_r607324936). For an example, you can refer to https://github.com/facebook/rocksdb/blob/master/db/db_bloom_filter_test.cc#L2149.

riversand963 (approved):
Thanks @akankshamahajan15 for the PR

zhichao-cao (approved):
LGTM, thanks

db/version_set.cc
akankshamahajan15: I added the string because ```meta->GetChecksumValue();``` can be empty whereas SST files stores ```Unknown``` instead of empty. So to make behavior consistent I replace empty string with ```Unknown``` in case of BlobFiles.

Issue #6995, opened 2020-06-17
options_helper.cc fails to compile with clang10


kryton:
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://groups.google.com/forum/#!forum/rocksdb or https://www.facebook.com/groups/rocksdb.dev

### Actual behavior
```
[ 13%] Building CXX object CMakeFiles/rocksdb-shared.dir/options/options_helper.cc.o
/home/iholsman/src/rocksdb/options/options_helper.cc:936:19: error: loop variable 'iter' of type 'const std::pair<const std::__cxx11::basic_string<char>, rocksdb::OptionTypeInfo>' creates a copy from type 'const std::pair<const std::__cxx11::basic_string<char>, rocksdb::OptionTypeInfo>' [-Werror,-Wrange-loop-construct]
  for (const auto iter : type_info) {
                  ^
/home/iholsman/src/rocksdb/options/options_helper.cc:936:8: note: use reference type 'const std::pair<const std::__cxx11::basic_string<char>, rocksdb::OptionTypeInfo> &' to prevent copying
  for (const auto iter : type_info) {
       ^~~~~~~~~~~~~~~~~
                  &
1 error generated.
make[2]: *** [CMakeFiles/rocksdb-shared.dir/build.make:1532: CMakeFiles/rocksdb-shared.dir/options/options_helper.cc.o] Error 1
```
### Steps to reproduce the behavior
on Ubuntu 20.04.
using
/usr/bin/clang++-10
$ cmake --version
cmake version 3.16.3

feliwir:
I can't compile either with latest clang on Ubuntu. Can the PR please be merged or looked at?

mrambacher:
In what version of RocksDB are you seeing this issue?  I am attempting to reproduce it in master, but do not see the failure.

adamretter:
This not the same issue, but I had problems also with Clang 11 on Ubuntu - https://github.com/facebook/rocksdb/issues/7736

mrambacher:
I installed clang via "sudo apt install clang" on my ubuntu box, which installed clang v10:
clang --version
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

I did not see the same build issue, but saw an issue with atomic functions:

[...]

PR #8183, opened 2021-04-13
Add check to cmake to see if we need to link against -latomic

mrambacher:
For some compilers/environments (e.g. Clang, riscv64), we need to link against -latomic.  Check if this is a requirement and add the library to the third-party libs if it is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,13 +312,32 @@ int main() {
   auto d = _mm_cvtsi128_si64(c);
 }
 " HAVE_SSE42)
-unset(CMAKE_REQUIRED_FLAGS)
 if(HAVE_SSE42)
   add_definitions(-DHAVE_SSE42)
   add_definitions(-DHAVE_PCLMUL)
 elseif(FORCE_SSE42)
   message(FATAL_ERROR "FORCE_SSE42=ON but unable to compile with SSE4.2 enabled")
 endif()
+
+# Check if -latomic is required or not
+if (NOT MSVC)
+  set(CMAKE_REQUIRED_FLAGS "--std=c++11")
+  CHECK_CXX_SOURCE_COMPILES("
+#include <atomic>
+std::atomic<uint64_t> x(0);
+int main() {
+  uint64_t i = x.load(std::memory_order_relaxed);
+  bool b = x.is_lock_free();
+  return 0;
+}
+" BUILTIN_ATOMIC)
+if (NOT BUILTIN_ATOMIC)
+  #TODO: Check if -latomic exists 
+  list(APPEND THIRDPARTY_LIBS atomic)
+endif()
+endif()
+
+unset(CMAKE_REQUIRED_FLAGS)
 
 CHECK_CXX_SOURCE_COMPILES("
 #if defined(_MSC_VER) && !defined(__thread)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,6 +296,7 @@ else()
 endif()
 
 include(CheckCXXSourceCompiles)
+set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
 if(NOT MSVC)
   set(CMAKE_REQUIRED_FLAGS "-msse4.2 -mpclmul")
 endif()
@@ -337,7 +338,8 @@ if (NOT BUILTIN_ATOMIC)
 endif()
 endif()
 
-unset(CMAKE_REQUIRED_FLAGS)
+# Reset the required flags
+set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
 
 CHECK_CXX_SOURCE_COMPILES("
 #if defined(_MSC_VER) && !defined(__thread)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -333,7 +333,7 @@ int main() {
 }
 " BUILTIN_ATOMIC)
 if (NOT BUILTIN_ATOMIC)
-  #TODO: Check if -latomic exists 
+  #TODO: Check if -latomic exists
   list(APPEND THIRDPARTY_LIBS atomic)
 endif()
 endif()


mrambacher:
This should fix #7051 and #6995.

grooverdan:
Thanks for writing this.

This starts ok, however the logic of failure to compile and `-latomic` is the solution is a little weak.

```
set(ATOMIC_TEST_PROGRAM ".....")
CHECK_CXX_SOURCE_COMPILES(${ATOMIC_TEST_PROGRAM}  BUILTIN_ATOMIC)
if (NOT BUILTIN_ATOMIC)
   set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} atomic)
   CHECK_CXX_SOURCE_COMPILES(${ATOMIC_TEST_PROGRAM} WITH_ATOMIC_LIB)
   if(NOT WITH_ATOMIC_LIB)
      message(FATAL ...
   endif()
endif()
```

Is unconditionally clearing `CMAKE_REQUIRED_FLAGS` right, or should there be a save/restore?

mrambacher:
> Thanks for writing this.
> 
> This starts ok, however the logic of failure to compile and `-latomic` is the solution is a little weak.

I am not sure what you mean by a "little weak".  There is a "CheckAtomic.cmake" that can be found as part of clang but my lack of knowledge of cmake and how to find this file stymied me but this solution is very similar to what they do (they do a better job of checking if the atomic library is present).  

> ```
> set(ATOMIC_TEST_PROGRAM ".....")
> CHECK_CXX_SOURCE_COMPILES(${ATOMIC_TEST_PROGRAM}  BUILTIN_ATOMIC)
> if (NOT BUILTIN_ATOMIC)
>    set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} atomic)
>    CHECK_CXX_SOURCE_COMPILES(${ATOMIC_TEST_PROGRAM} WITH_ATOMIC_LIB)
>    if(NOT WITH_ATOMIC_LIB)
>       message(FATAL ...
>    endif()
> endif()
> ```
> 
> Is unconditionally clearing `CMAKE_REQUIRED_FLAGS` right, or should there be a save/restore?

I did not add the clearing of CMAKE_REQUIRED_FLAGS, only moved it.  If it should be a save/restore, that is fine.

pdillinger (approved):
¯\\_(ツ)_/¯ seems fine

I wish we would unify detection of stuff like this between the cmake and Makefile builds, but that's on the wishlist.

mrambacher:
> ¯\_(ツ)_/¯ seems fine
> 
> I wish we would unify detection of stuff like this between the cmake and Makefile builds, but that's on the wishlist.

Thanks for the review!  I fully agree that it would be nice if we could unify the cmake and make systems.  Trying to keep the two systems in sync is problematic and error prone to say the least.

alexfanqi:
Hi,
Sorry for replying to this closed issue. But the cpp code piece used in this pr for testing libatomic seems not adequate. 

I tested  with gcc 11.2 and on Sifive Unmatched (riscv64). The code piece passed without linker error and I still get ```undefined reference to `__atomic_compare_exchange_1'``` later in compiling the whole program.

This snippet from WebKit will work. 
https://github.com/WebKit/WebKit/blob/87d88025bf5a3089e68b4466a12dd1cb2584d557/Source/cmake/WebKitCompilerFlags.cmake#L282

original
```
unmatch2 /tmp # cat test.cpp 
#include <atomic>
std::atomic<uint64_t> x(0);
int main() {
uint64_t i = x.load(std::memory_order_relaxed);
bool b = x.is_lock_free();
return 0;
}
unmatch2 /tmp # gcc -Wall test.cpp 
test.cpp: In function ‘int main()’:
test.cpp:4:10: warning: unused variable ‘i’ [-Wunused-variable]
    4 | uint64_t i = x.load(std::memory_order_relaxed);
      |          ^
test.cpp:5:6: warning: unused variable ‘b’ [-Wunused-variable]
    5 | bool b = x.is_lock_free();
      |      ^
"
```

new from Webkit
```
unmatch2 /tmp # cat test2.cpp 
#include <atomic>
int main() {
std::atomic<bool> y (false);
std::atomic<uint64_t> x (0);
bool expected = true;
bool j = y.compare_exchange_weak(expected,false);
x++;
return 0;
}
unmatch2 /tmp # gcc test2.cpp 
/usr/lib/gcc/riscv64-unknown-linux-gnu/11.2.0/../../../../riscv64-unknown-linux-gnu/bin/ld: /tmp/.private/root/cc45pcng.o: in function `.L0 ':
test2.cpp:(.text._ZNSt6atomicIbE21compare_exchange_weakERbbSt12memory_order[_ZNSt6atomicIbE21compare_exchange_weakERbbSt12memory_order]+0xb2): undefined reference to `__atomic_compare_exchange_1'
collect2: error: ld returned 1 exit status
unmatch2 /tmp # 

```

gcc -v
```
Using built-in specs.
COLLECT_GCC=gcc    
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/riscv64-unknown-linux-gnu/11.2.0/lto-wrapper
Target: riscv64-unknown-linux-gnu     
Configured with: /var/tmp/portage/sys-devel/gcc-11.2.0/work/gcc-11.2.0/configure --host=riscv64-unknown-linux-gnu --
build=riscv64-unknown-linux-gnu --prefix=/usr --bindir=/usr/riscv64-unknown-linux-gnu/gcc-bin/11.2.0 --includedir=/u
sr/lib/gcc/riscv64-unknown-linux-gnu/11.2.0/include --datadir=/usr/share/gcc-data/riscv64-unknown-linux-gnu/11.2.0 -
-mandir=/usr/share/gcc-data/riscv64-unknown-linux-gnu/11.2.0/man --infodir=/usr/share/gcc-data/riscv64-unknown-linux
-gnu/11.2.0/info --with-gxx-include-dir=/usr/lib/gcc/riscv64-unknown-linux-gnu/11.2.0/include/g++-v11 --with-python-
dir=/share/gcc-data/riscv64-unknown-linux-gnu/11.2.0/python --enable-languages=c,c++,fortran --enable-obsolete --ena
ble-secureplt --disable-werror --with-system-zlib --enable-nls --without-included-gettext --disable-libunwind-except
ions --enable-checking=release --with-bugurl=https://bugs.gentoo.org/ --with-pkgversion='Gentoo 11.2.0 p1' --disable
-esp --enable-libstdcxx-time --enable-shared --enable-threads=posix --enable-__cxa_atexit --enable-clocale=gnu --dis
able-multilib --disable-fixed-point --with-abi=lp64d --enable-libgomp --disable-libssp --disable-libada --disable-sy
stemtap --disable-valgrind-annotations --disable-vtable-verify --disable-libvtv --withou
t-zstd --enable-lto --withou
t-isl --disable-libsanitizer --enable-default-pie --enable-default-ssp
Thread model: posix
Supported LTO compression algorithms: zlib
gcc version 11.2.0 (Gentoo 11.2.0 p1) 

```

PR #8189, opened 2021-04-14
Fix for LITE mode failure on MacOS

akankshamahajan15:
Summary: Fix for failure to build in LITE mode on MacOs from
BlobFileCompletionCallback unused private fields.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


mrambacher (approved):
Personally, I would rather see the OnBlobFileCompleted be defined and a no-op in LITE mode rather than not present.  This would limit your changes to only the blob_file_completion_callback.h file. 

Having said that, this change does work and I am not sure which "style" is preferred.

PR #8190, opened 2021-04-14
Fix Windows strcmp for Unicode

adamretter:
The code for strcmp that was present does work when compiled for Windows unicode file paths.

Needs backporting to:
* 6.17.fb
* 6.18.fb
* 6.19.fb

### DIFF
--- a/port/win/env_win.cc
+++ b/port/win/env_win.cc
@@ -682,7 +682,8 @@ IOStatus WinFileSystem::GetChildren(const std::string& dir,
     // which appear only on some platforms
     const bool ignore =
         ((data.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) != 0) &&
-        (strcmp(data.cFileName, ".") == 0 || strcmp(data.cFileName, "..") == 0);
+        (RX_FNCMP(data.cFileName, ".") == 0 ||
+         RX_FNCMP(data.cFileName, "..") == 0);
     if (!ignore) {
       auto x = RX_FILESTRING(data.cFileName, RX_FNLEN(data.cFileName));
       result->push_back(FN_TO_RX(x));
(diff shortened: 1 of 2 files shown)


mrambacher (commented):
I am a little suspect of this code.  The fact that the RX_FNCMP method treats the first and second arguments differently bothers me a bit.

What would happen if you used std::string methods instead?  What if "." and ".." were store in static std::string and compared that way?  Would it just "work" for both character cases or would there be something else required?

adamretter:
> The fact that the RX_FNCMP method treats the first and second arguments differently bothers me a bit.

That was done for convenience.

> What would happen if you used std::string methods instead? What if "." and ".." were store in static std::string and compared that way? Would it just "work" for both character cases or would there be something else required?

You cannot use std::string as an argument to wcscmp

mrambacher:
> > The fact that the RX_FNCMP method treats the first and second arguments differently bothers me a bit.
> 
> That was done for convenience.
> 
> > What would happen if you used std::string methods instead? What if "." and ".." were store in static std::string and compared that way? Would it just "work" for both character cases or would there be something else required?
> 
> You cannot use std::string as an argument to wcscmp

I meant what if you did something like:
static std::string dot = ".";
...
if (dot == data.cFileName ...  )

<Obviously the same sort of code for dotdot>.  Does the std::string class do the right thing for char vs wchar comparisons for "free"?

adamretter:
@mrambacher I don't think std::string overloads `==` - http://www.cplusplus.com/reference/string/string/

jay-zhuang (approved):
LGTM.

@mrambacher any other concern?

jay-zhuang:
@adamretter would you please do `make format`? Or please give us permission to edit the file.

mrambacher (approved):
LGTM.

adamretter:
@jay-zhuang I ran make format as requested and rebased.

PR #8191, opened 2021-04-15
Adding the callback and test case

zhichao-cao:
Add the callback functions to block_based_table_reader such that it can use tiered cache.

test plan: added the testing case


table/block_based/block_based_table_reader.cc
```diff
@@ -1182,11 +1187,61 @@ Status BlockBasedTable::GetDataBlockFromCache(
   Status s;
   BlockContents* compressed_block = nullptr;
   Cache::Handle* block_cache_compressed_handle = nullptr;
+  Cache::CreateCallback create_cb;
+  Cache::CacheItemHelperCallback helper_cb;
+  if (std::is_same<TBlocklike, BlockContents>::value) {
```
anand1976, line 1192: An alternative approach could be to add  ```CacheItemHelperCallback``` to ```BlocklikeTraits```, and have a common templatized implementation of ```CreateCallback``` that calls ```BlocklikeTraits::Create<TBlocklike>```.  It seems cleaner to have all cache related functionality for a block type in one place like ```BlocklikeTraits```. What do you think?

zhichao-cao: Sure, it would be easier to extend in the future. Otherwise, people may forget to add the std::is_same<TBlocklike, "">::value branch to the two places, or even more places in the future.

table/block_based/block_based_table_reader.cc
```diff
@@ -1204,9 +1271,8 @@ Status BlockBasedTable::GetDataBlockFromCache(
 
   assert(!compressed_block_cache_key.empty());
   block_cache_compressed_handle =
-      block_cache_compressed->Lookup(compressed_block_cache_key);
-
-  Statistics* statistics = rep_->ioptions.statistics;
+      block_cache_compressed->Lookup(compressed_block_cache_key, helper_cb,
+                                     create_cb, Cache::Priority::LOW, true);
```
anand1976, line 1275: Should priority be HIGH or LOW based on the block type?

anand1976 (approved):
Thanks for addressing the comments. I had one more comment, but overall LGTM

anand1976:
@zhichao-cao Maybe moving the functions to ```block_based_table_reader.cc``` is better, since ```util/ ``` is generic and doesn't have anything specific to the block based table format.

ajkr:
Can the title be more specific?

zhichao-cao:
> Can the title be more specific?

Sure

table/block_based/block_based_table_reader.cc
```diff
@@ -1203,10 +1373,8 @@ Status BlockBasedTable::GetDataBlockFromCache(
   }
 
   assert(!compressed_block_cache_key.empty());
-  block_cache_compressed_handle =
-      block_cache_compressed->Lookup(compressed_block_cache_key);
-
-  Statistics* statistics = rep_->ioptions.statistics;
+  block_cache_compressed_handle = block_cache_compressed->Lookup(
```
anand1976, line 1376: We might need to explicitly use the ```BlockContents``` helper here too.

PR #8192, opened 2021-04-15
Handle rename() failure in non-local FS

riversand963:
In a distributed environment, a file `rename()` operation can succeed on server (remote)
side, but the client can somehow return non-ok status to RocksDB. Possible reasons include
network partition, connection issue, etc. This happens in `rocksdb::SetCurrentFile()`, which
can be called in `LogAndApply() -> ProcessManifestWrites()` if RocksDB tries to switch to a
new MANIFEST. We currently always delete the new MANIFEST if an error occurs.

This is problematic in distributed world. If the server-side successfully updates the CURRENT
file via renaming, then a subsequent `DB::Open()` will try to look for the new MANIFEST and fail.

As a fix, we can track the execution result of IO operations on the new MANIFEST.
- If IO operations on the new MANIFEST fail, then we know the CURRENT must point to the original
  MANIFEST. Therefore, it is safe to remove the new MANIFEST.
- If IO operations on the new MANIFEST all succeed, but somehow we end up in the clean up
  code block, then we do not know whether CURRENT points to the new or old MANIFEST. (For local
  POSIX-compliant FS, it should still point to old MANIFEST, but it does not matter if we keep the
  new MANIFEST.) Therefore, we keep the new MANIFEST.
    - Any future `LogAndApply()` will switch to a new MANIFEST and update CURRENT.
    - If process reopens the db immediately after the failure, then the CURRENT file can point
      to either the new MANIFEST or the old one, both of which exist. Therefore, recovery can
      succeed and ignore the other.

Test plan:
make check

### DIFF
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -313,8 +313,11 @@ Status DBImpl::NewDB(std::vector<std::string>* new_filenames) {
       new_filenames->emplace_back(
           manifest.substr(manifest.find_last_of("/\\") + 1));
     }
-  } else {
-    fs_->DeleteFile(manifest, IOOptions(), nullptr);
+  }
+  if (!s.ok()) {
+    fs_->DeleteFile(manifest, IOOptions(), nullptr).PermitUncheckedError();
+    fs_->DeleteFile(CurrentFileName(dbname_), IOOptions(), nullptr)
+        .PermitUncheckedError();
   }
   return s;
 }
--- a/db/db_impl/db_impl_open.cc
+++ b/db/db_impl/db_impl_open.cc
@@ -285,6 +285,9 @@ Status DBImpl::NewDB(std::vector<std::string>* new_filenames) {
   ROCKS_LOG_INFO(immutable_db_options_.info_log, "Creating manifest 1 \n");
   const std::string manifest = DescriptorFileName(dbname_, 1);
   {
+    if (fs_->FileExists(manifest, IOOptions(), nullptr).ok()) {
+      fs_->DeleteFile(manifest, IOOptions(), nullptr).PermitUncheckedError();
+    }
     std::unique_ptr<FSWritableFile> file;
     FileOptions file_options = fs_->OptimizeForManifestWrite(file_options_);
     s = NewWritableFile(fs_.get(), manifest, &file, file_options);
@@ -313,11 +316,8 @@ Status DBImpl::NewDB(std::vector<std::string>* new_filenames) {
       new_filenames->emplace_back(
           manifest.substr(manifest.find_last_of("/\\") + 1));
     }
-  }
-  if (!s.ok()) {
+  } else {
     fs_->DeleteFile(manifest, IOOptions(), nullptr).PermitUncheckedError();
-    fs_->DeleteFile(CurrentFileName(dbname_), IOOptions(), nullptr)
-        .PermitUncheckedError();
   }
   return s;
 }
--- a/db/db_test2.cc
+++ b/db/db_test2.cc
@@ -5457,7 +5457,10 @@ class RenameCurrentTest : public DBTestBase,
  public:
   RenameCurrentTest()
       : DBTestBase("rename_current_test", /*env_do_fsync=*/true),
-        sync_point_(GetParam()) {}
+        sync_point_(GetParam()),
+        test_env_(new TestEnvWrapper(env_)) {}
+
+  ~RenameCurrentTest() override { Close(); }
 
   void SetupSyncPoints() {
     SyncPoint::GetInstance()->DisableProcessing();
@@ -5469,6 +5472,7 @@ class RenameCurrentTest : public DBTestBase,
   }
 
   const std::string sync_point_;
+  std::unique_ptr<Env> test_env_;
 };
 
 INSTANTIATE_TEST_CASE_P(DistributedFS, RenameCurrentTest,
@@ -5478,8 +5482,7 @@ INSTANTIATE_TEST_CASE_P(DistributedFS, RenameCurrentTest,
 TEST_P(RenameCurrentTest, Open) {
   Destroy(last_options_);
   Options options = CurrentOptions();
-  TestEnvWrapper env_wrapper(env_);
-  options.env = static_cast_with_check<Env>(&env_wrapper);
+  options.env = test_env_.get();
   SetupSyncPoints();
   SyncPoint::GetInstance()->EnableProcessing();
   Status s = TryReopen(options);
@@ -5487,15 +5490,13 @@ TEST_P(RenameCurrentTest, Open) {
 
   SyncPoint::GetInstance()->DisableProcessing();
   Reopen(options);
-  Close();
 }
 
 TEST_P(RenameCurrentTest, Flush) {
   Destroy(last_options_);
   Options options = CurrentOptions();
   options.max_manifest_file_size = 1;
-  TestEnvWrapper env_wrapper(env_);
-  options.env = static_cast_with_check<Env>(&env_wrapper);
+  options.env = test_env_.get();
   Reopen(options);
   ASSERT_OK(Put("key", "value"));
   SetupSyncPoints();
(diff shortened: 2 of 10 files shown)


db/db_impl/db_impl_open.cc
```diff
@@ -313,8 +313,11 @@ Status DBImpl::NewDB(std::vector<std::string>* new_filenames) {
       new_filenames->emplace_back(
           manifest.substr(manifest.find_last_of("/\\") + 1));
     }
-  } else {
-    fs_->DeleteFile(manifest, IOOptions(), nullptr);
+  }
+  if (!s.ok()) {
+    fs_->DeleteFile(manifest, IOOptions(), nullptr).PermitUncheckedError();
+    fs_->DeleteFile(CurrentFileName(dbname_), IOOptions(), nullptr)
```
zhichao-cao, line 319: I'm not sure the behavior here. Before we do the renaming, we have xxx.dbtmp and CURRENT. If rename failed, we may keep the same status xxx.dbtmp and CURRENT or we only have the CURRENT (new one). If it is the previous case, we will have not the CURRENT file at all?

riversand963: This is `NewDB()`. If `NewDB()` failed, then the `DB::Open()` call fails too. I think it's OK to delete all generated files here.

zhichao-cao: I see. For newly created DB, it is fine to delete all of them, which will not create any lose.

riversand963: Furthermore, if we do not delete them. A subsequent attempt to create new db may fail because some FS does not support overwriting an existing file.

zhichao-cao: > Furthermore, if we do not delete them. A subsequent attempt to create new db may fail because some FS does not support overwriting an existing file.

I did not read the code or subsequent attempt. In that case, does the sequence number increases? If so, we aways create a Manifest file with a new name and so the xxxx.dbtmp.

mrambacher: I am a little confused here after reading the code.  The SetCurrentFile already deletes the temporary file if the rename failed so I am not sure why that is needed here.  I can see where the Manifest could be left around if the SetCurrentFile failed, but am not sure why we need both DeleteFile calls here.

riversand963: Thinking more. I think original code of deleting only the MANIFEST here is correct, but the current one has issues. If we delete the MANIFEST, we should also make sure the CURRENT file is deleted in this case. Otherwise, a future call to `DB::Open()` will find the CURRENT file and think the DB exists. If the MANIFEST is deleted, then open will fail. The difficulty here is that we cannot make sure CURRENT is deleted if we already end up at this point.
1. If the rename succeeded on server side, but client returns error. Then we cannot rely on the client to delete the CURRENT file.
2. If rename failed on remote, then we know CURRENT does not exist, and we are fine.

The original motivation for deleting the MANIFEST is to avoid future `NewDB()` call overwriting an existing file. Now I think this should not be done in the clean-up-after-error phase. Instead, this should be done at the beginning of `NewDB()`. Assuming no data loss. If a previous call to `NewDB()` failed, then the caller may retry `DB::Open()`. There are several possibilities.
1. Both CURRENT and MANIFEST exist. `NewDB()` won't be called, and the db is empty. This is OK.
2. Only the MANIFEST exists (maybe there is a .tmp file that failed to have been deleted). `NewDB()` will be called. In this case, we should avoid creating the same MANIFEST and tmp file.

db/version_set.cc
```diff
@@ -4303,11 +4307,41 @@ Status VersionSet::ProcessManifestWrites(
...
+    if (manifest_io_status.ok()) {
+      manifest_file_number_ = pending_manifest_file_number_;
+      manifest_file_size_ = new_manifest_file_size;
+    }
     // If manifest append failed for whatever reason, the file could be
     // corrupted. So we need to force the next version update to start a
     // new manifest file.
     descriptor_log_.reset();
-    if (new_descriptor_log) {
+    // If manifest operations failed, then we know the CURRENT file still
```
jay-zhuang, line 4318: If `rename()` failed, should it just keep retrying? I guess we will retry `ProcessManifestWrites()` later by re-writing the manifest file and call `rename()` again, but should it just retry the single failed operation `rename()`?

db/db_test2.cc
```diff
@@ -5439,6 +5439,108 @@ TEST_F(DBTest2, AutoPrefixMode1) {
...
+  }
+};
+
+class RenameCurrentTest : public DBTestBase,
+                          public testing::WithParamInterface<std::string> {
+ public:
+  RenameCurrentTest()
+      : DBTestBase("rename_current_test", /*env_do_fsync=*/true),
+        sync_point_(GetParam()) {}
+
```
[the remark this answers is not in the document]
riversand963: Thanks. I think it may be more complex than this.
`~DBTestBase()` will be called later which will `delete` the `db` pointer. At that time, the `TestEnv` object created in `RenameCurrentTest` is already destroyed, even if it is a `std::unique_ptr` in `RenameCurrentTest`. Therefore, we have to close the db anyway.

jay-zhuang (commented):
Is there other operation (failed but could be success in the backend) that could cause the similar problem?

db/version_set.cc
riversand963: We do not have a good retry policy, so it is better to return the error and let application decide. We cannot keep retrying because we do not know how long it takes.
In the meantime, RocksDB should make sure there is no data loss. As you mentioned, the first subsequent `LogAndApply()` will do retry as necessary.

db/version_set.cc
```diff
@@ -4303,11 +4307,41 @@ Status VersionSet::ProcessManifestWrites(
...
+    // successfully, a subsequent recovery attempt will likely see the CURRENT
+    // pointing to the new MANIFEST, thus fail. We will not be able to open the
+    // DB again. Therefore, if manifest operations succeed, we should keep the
+    // the new MANIFEST. If the process proceeds, any future LogAndApply() call
+    // will switch to a new MANIFEST and update CURRENT. If user tries to
+    // re-open the DB,
+    // a) CURRENT points to the new MANIFEST, and the new MANIFEST is present.
+    // b) CURRENT points to the original MANIFEST, and the original MANIFEST
+    //    also exists.
+    if (new_descriptor_log && !manifest_io_status.ok()) {
```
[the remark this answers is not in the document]
riversand963: I think a subsequent db closing will delete it (https://github.com/facebook/rocksdb/blob/6.19.fb/db/db_impl/db_impl.cc#L585).

db/db_impl/db_impl_open.cc
riversand963: > > Furthermore, if we do not delete them. A subsequent attempt to create new db may fail because some FS does not support overwriting an existing file.
> 
> I did not read the code or subsequent attempt. In that case, does the sequence number increases? If so, we aways create a Manifest file with a new name and so the xxxx.dbtmp.

In `DeleteUnreferencedSst()`, we will bump the next_file_number_ to be the current largest file number + 1

db/db_test2.cc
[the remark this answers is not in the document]
mrambacher: You could address issue by doing a Close() in the Rename test destructor.  It might be a minor issue, but I can see someone adding a test in the future that fails in a strange way because they forgot to call Close() at the end.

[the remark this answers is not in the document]
riversand963: Make sense, will update.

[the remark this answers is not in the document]
riversand963: Done

db/db_test2.cc
```diff
@@ -5439,6 +5439,104 @@ TEST_F(DBTest2, AutoPrefixMode1) {
...
 }
+
+class TestEnvWrapper : public EnvWrapper {
+ public:
+  TestEnvWrapper(Env* _target) : EnvWrapper(_target) {}
+  Status NewWritableFile(const std::string& fname,
+                         std::unique_ptr<WritableFile>* result,
+                         const EnvOptions& env_opts) override {
+    Status s = target()->FileExists(fname);
+    EXPECT_TRUE(s.IsNotFound()) << fname << " already exists.";
```
siying, line 5450: Nit: can this logic live in SpecialEnv and guard by a variable?

siying (commented):
LGTM

db/db_test2.cc
riversand963: Sure. Let me update

siying (commented):
LGTM

db/db_test2.cc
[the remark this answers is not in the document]
riversand963: We ended up moving this logic to `SpecialEnv`.

zhichao-cao (approved):
LGTM, thanks for the fix!

PR #8195, opened 2021-04-15
Fix Makefile when multiple targets are invoked

mrambacher:
- Fixes the makefile to do the right thing when invoking multiple targets (e.g. make shared_lib install-shared).

- Fixes the building of db_stress in shared lib mode.


pdillinger (approved):
👍

PR #8196, opened 2021-04-16
Fix flaky failure in DBSSTest.DBWithSstFileManagerForBlobFilesWithGC

akankshamahajan15:
Summary: Updated the test to wait until all trash files are deleted by
SSTFileManager in the background. Since deletion runs in background so
number of files deleted might not always be as expected.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


jay-zhuang (approved):
LGTM. Thanks for fixing the flaky test.🥳

PR #8197, opened 2021-04-16
Fix flaky test BackupableDBTest.FileSizeForIncremental

akankshamahajan15:
Summary: Test was flaky because for kUseDbSessionId naming, blob files use
naming scheme kLegacyCrc32cAndFileSize. So expected number of files
because of collision can vary. So disabling blobdb for this test case.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


pdillinger (approved):
LGTM but maybe at least drop in a comment to re-enable for blob files once the db session id support is added 👍

PR #8198, opened 2021-04-16
Ribbon long-term support, starting level support

pdillinger:
Summary: Since the Ribbon filter schema seems good (compatible back to
6.15.0), this change commits to long term support of the SST schema,
even though we expect the API for enabling Ribbon to change (still
called NewExperimentalRibbonFilterPolicy).

This also adds support for "hybrid" configuration in which some levels
use Bloom (higher levels, lower numbered) for speed and the rest use
Ribbon (lower levels, higher numbered) for memory space efficiency.

Test Plan: unit test added, crash test support


pdillinger:
Can't make server side formatter happy it seems

include/rocksdb/filter_policy.h
```diff
@@ -216,23 +216,33 @@ class FilterPolicy {
...
+// Ribbon filters are compatible with RocksDB >= 6.15.0. Earlier
+// versions reading the data will behave as if no filter was used
+// (degraded performance until compaction rebuilds filters).
+//
+// Note: even with ribbon_starting_level=0, this policy can generate
+// Bloom filters in some cases. For very small filters (well under 1KB),
+// Bloom fallback is by design, as the current Ribbon schema is not
+// optimized to save vs. Bloom for such small filters. Other cases of
+// Bloom fallback should be exceptional and log an appropriate warning.
 extern const FilterPolicy* NewExperimentalRibbonFilterPolicy(
```
jay-zhuang, line 245: should `Experiment` be removed for the class name too?

include/rocksdb/filter_policy.h
```diff
@@ -216,23 +216,33 @@ class FilterPolicy {
...
-    double bloom_equivalent_bits_per_key);
+    double bloom_equivalent_bits_per_key, int ribbon_starting_level = 1);
```
jay-zhuang, line 246: I think the level specific information should be outside of a specific filter configuration if it's possible.
And I'm not sure if it's exposing too much information about levels.
Could it be like compression that has CompressionType and Bottommost CompressionType:
https://github.com/facebook/rocksdb/blob/524b10bd6ee5fbea8ec9e90dd668c07d0c653371/include/rocksdb/options.h#L200-L206

pdillinger: The intention with this transitional API is to provide some manual controls to play with to help inform what the next generation API should look like

pdillinger: And it should be easy to maintain support for this transitional API for a while.

include/rocksdb/filter_policy.h
pdillinger: Hmm. I want to convey that the API will likely go away some day but the schema should not. Somebody could be annoyed with a change in name, but that's probably not a big concern. I guess since we want to deprecate the NewBloomFilterPolicy API also, it wouldn't be a big deal to have NewRibbonFilterPolicy also.

I can make that change.

include/rocksdb/filter_policy.h
jay-zhuang: If it's going to be deprecated, then we should keep the `Experimental` name.
But what would be the future API look like and should we considering the migrating path? For example, if the future API is like compression, which only has bottommost config, then it would be difficult if we already provide level. (For filter, by the way, I think an upmost config may make more sense.)

jay-zhuang (approved):
LGTM.

I think we should keep the `Experimental` name, so the interface and feature could be changed in the future.

include/rocksdb/filter_policy.h
pdillinger: I agree topmost is likely to deserve special treatment but it's easier to try some different things with customers if the API is there.  The new level threshold functionality is already implementable with public APIs, btw. This change just makes it easier. So I'm not too concerned about migration. So maybe keep the Experimental name of you're ok with it

PR #8200, opened 2021-04-17
Return error if trying to open secondary on missing or inaccessible primary

riversand963:
If the primary's CURRENT file is missing or inaccessible, the secondary should not hang
trying repeatedly to switch to the next MANIFEST.

Test plan:
make check

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -5760,49 +5760,55 @@ Status ReactiveVersionSet::MaybeSwitchManifest(
     std::unique_ptr<log::FragmentBufferedReader>* manifest_reader) {
   assert(manifest_reader != nullptr);
   Status s;
-  do {
-    std::string manifest_path;
-    s = GetCurrentManifestPath(dbname_, fs_.get(), &manifest_path,
-                               &manifest_file_number_);
-    std::unique_ptr<FSSequentialFile> manifest_file;
-    if (s.ok()) {
-      if (nullptr == manifest_reader->get() ||
-          manifest_reader->get()->file()->file_name() != manifest_path) {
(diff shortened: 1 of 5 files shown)


db/version_set.cc
```diff
@@ -5757,43 +5759,62 @@ Status ReactiveVersionSet::MaybeSwitchManifest(
...
+    if (manifest_tailer_) {
+      manifest_tailer_->PrepareToReadNewManifest();
+    }
+  } else if (s.IsPathNotFound()) {
+    // This can happen if the primary switches to a new MANIFEST after the
+    // secondary reads the CURRENT file but before the secondary actually tries
+    // to open the MANIFEST.
+    s = Status::TryAgain(
+        "The primary may have switched to a new MANIFEST and deleted the old "
+        "one.");
```
jay-zhuang, line 5816: Should we retry it, instead of reporting to user? Maybe the secondary should take a snapshot of manifest/CURRENT file at the beginning.

jay-zhuang (approved):
LGTM.

db/version_set.cc
riversand963: If the secondary checks for the presence of MANIFEST via `FileExists()` and then tries to open it, there is always a possibility that the primary deletes the MANIFEST in between, thus the secondary will always likely hit a missing MANIFEST file. Similarly, after checking the presence of MANIFEST but before taking a hard-link, it is possible that the MANIFEST is deleted. We should always handle this situation.
Retry is possible, but I think we should have an interface for application to set retry policy. Since we do not have the interface yet, and it is beyond the scope of this PR. Future change of adding retry logic will be a behavior/contract change for `TryCatchUpWithPrimary()`.

Issue #6245, opened 2019-12-24
global seqnum should be applied before sorting a level's files


ajkr:
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://www.facebook.com/groups/rocksdb.dev

Commit hash: 2d1670948752a54c9b67797c83e9fdb563ebd0b1 (Dec 17 2019)

### Steps to reproduce the behavior

```
$ rm -rf ./tmp && for i in $(seq 3) ; do ./ldb put k a --create_if_missing --db=./tmp ; done
$ echo "k ==> b" | ./ldb write_extern_sst ./tmp.sst --db=./tmp/
external SST file written to ./tmp.sst
$ ./ldb ingest_extern_sst ./tmp.sst --db=./tmp
external SST files ingested
$ ./ldb get k --db=./tmp
a
```

### Expected behavior

The repro sequence of commands should lead to `ldb get k` returning "b", i.e., the latest value written for key "k" via ingestion.

### Actual behavior

The repro sequence led to `ldb get k` returning "a", which is incorrect. If we iterate over the internal keys we can see the key with largest seqnum comes last for some reason.

```
$ ./ldb idump ./tmp.sst --db=./tmp
'k' seq:3, type:1 => a
'k' seq:2, type:1 => a
'k' seq:1, type:1 => a
'k' seq:4, type:1 => b
Internal keys in range: 4
```

Examining the manifest shows us why. The file containing `k@4` comes last, whereas it should come first. I believe this is due to not applying the global seqnum before sorting the files within a level.

```
./ldb manifest_dump --db=./tmp | head

--------------- Column family "default"  (ID 0) --------------
log number: 16
comparator: leveldb.BytewiseComparator
--- level 0 --- version# 1 ---
--- level 1 --- version# 1 ---
 10:794[3 .. 3]['k' seq:3, type:1 .. 'k' seq:3, type:1]
 7:794[2 .. 2]['k' seq:2, type:1 .. 'k' seq:2, type:1]
 4:794[1 .. 1]['k' seq:1, type:1 .. 'k' seq:1, type:1]
 21:845[4 .. 4]['k' seq:0, type:1 .. 'k' seq:0, type:1]
--- level 2 --- version# 1 ---
```

yiwu-arbug:
@ajkr seems a serious issue. have you root-caused this one? thanks.

yiwu-arbug:
Anyone have progress on fixing this?

[...]

PR #8201, opened 2021-04-17
add assigned seqno before sorting.

jijiew:
Summary:
bug fix for the issue/6245
https://github.com/facebook/rocksdb/issues/6245

Test Plan:
test added in test file

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/external_sst_file_basic_test.cc
+++ b/db/external_sst_file_basic_test.cc
@@ -1542,6 +1542,43 @@ TEST_F(ExternalSSTFileBasicTest, OverlappingFiles) {
   ASSERT_EQ(2, NumTableFilesAtLevel(0));
 }
 
+TEST_F(ExternalSSTFileBasicTest, IngestFileAfterDBPut) {
+
+  Options options = CurrentOptions();
+
+  ASSERT_OK(Put("k", "a"));
+  Flush();
+  ASSERT_OK(Put("k", "a"));
+  Flush();
+  ASSERT_OK(Put("k", "a"));
+  Flush();
+  SstFileWriter sst_file_writer(EnvOptions(), options);
+
+  // Current file size should be 0 after sst_file_writer init and before open a
+  // file.
+  ASSERT_EQ(sst_file_writer.FileSize(), 0);
+
+  std::string file1 = sst_files_dir_ + "file1.sst";
+  ASSERT_OK(sst_file_writer.Open(file1));
+  ASSERT_OK(sst_file_writer.Put("k", "b"));
+
+  ExternalSstFileInfo file1_info;
+  Status s = sst_file_writer.Finish(&file1_info);
+  ASSERT_OK(s) << s.ToString();
+
+  // Current file size should be non-zero after success write.
+  ASSERT_GT(sst_file_writer.FileSize(), 0);
+
+  IngestExternalFileOptions ifo;
+  ifo.allow_global_seqno = true;
+  ifo.write_global_seqno = true;
+  s = db_->IngestExternalFile({file1}, ifo);
+  ASSERT_OK(dbfull()->TEST_WaitForCompact());
+
+  ASSERT_EQ(Get("k"), "b");
+
+}
+
 INSTANTIATE_TEST_CASE_P(ExternalSSTFileBasicTest, ExternalSSTFileBasicTest,
                         testing::Values(std::make_tuple(true, true),
                                         std::make_tuple(true, false),
--- a/db/external_sst_file_basic_test.cc
+++ b/db/external_sst_file_basic_test.cc
@@ -1543,7 +1543,6 @@ TEST_F(ExternalSSTFileBasicTest, OverlappingFiles) {
 }
 
 TEST_F(ExternalSSTFileBasicTest, IngestFileAfterDBPut) {
-
   Options options = CurrentOptions();
 
   ASSERT_OK(Put("k", "a"));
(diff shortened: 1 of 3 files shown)


db/external_sst_file_basic_test.cc
```diff
@@ -1542,6 +1542,43 @@ TEST_F(ExternalSSTFileBasicTest, OverlappingFiles) {
...
+  ExternalSstFileInfo file1_info;
+  Status s = sst_file_writer.Finish(&file1_info);
+  ASSERT_OK(s) << s.ToString();
+
+  // Current file size should be non-zero after success write.
+  ASSERT_GT(sst_file_writer.FileSize(), 0);
+
+  IngestExternalFileOptions ifo;
+  ifo.allow_global_seqno = true;
+  ifo.write_global_seqno = true;
```
mrambacher, line 1574: Is there a reason you need "write_global_seqno=true".  Is it part of this fix/required?  Comments might help.

db/external_sst_file_basic_test.cc
```diff
@@ -1542,6 +1542,43 @@ TEST_F(ExternalSSTFileBasicTest, OverlappingFiles) {
...
+  s = db_->IngestExternalFile({file1}, ifo);
```
mrambacher, line 1575: Do you need to check s?  Should there be an ASSERT_OK(s) here?

db/external_sst_file_ingestion_job.cc
```diff
@@ -622,7 +622,10 @@ Status ExternalSstFileIngestionJob::GetIngestedFileInfo(
     if (key.sequence != 0) {
       return Status::Corruption("External file has non zero sequence number");
     }
-    file_to_ingest->smallest_internal_key.SetFrom(key);
+
+    SequenceNumber assigned_seqno = versions_->LastSequence();
+
+    file_to_ingest->smallest_internal_key.Set(key.user_key, assigned_seqno, ValueType::kTypeValue);
```
mrambacher, line 628: Shouldn't the last arg be key.type and not ValueType::kTypeValue??

db/external_sst_file_ingestion_job.cc
```diff
@@ -633,7 +636,8 @@ Status ExternalSstFileIngestionJob::GetIngestedFileInfo(
     if (key.sequence != 0) {
       return Status::Corruption("External file has non zero sequence number");
     }
-    file_to_ingest->largest_internal_key.SetFrom(key);
+
+    file_to_ingest->largest_internal_key.Set(key.user_key, assigned_seqno, ValueType::kTypeValue);
```
mrambacher, line 640: Same as above.

mrambacher (commented):
I think a few more test cases might be in order:
- What happens if you ingest when data is not flushed?
- What happens if you Get after ingest without the compact?
- What happens if there are more keys in the ingested file (e.g. {j, k}, {k, l}, {j,k,l} where the key in question is not always the smallest and/or largest

Generally, this looks okay to me, but I am not an expert in this area....

db/external_sst_file_ingestion_job.cc
```diff
@@ -622,7 +622,10 @@ Status ExternalSstFileIngestionJob::GetIngestedFileInfo(
     if (key.sequence != 0) {
       return Status::Corruption("External file has non zero sequence number");
     }
-    file_to_ingest->smallest_internal_key.SetFrom(key);
+
+    SequenceNumber assigned_seqno = versions_->LastSequence();
```
riversand963, line 626: Maybe I am missing some context from the original bug, but I think we may later call `AssignLevelAndSeqnoForIngestedFile(...)` to compute `assigned_seqno` which does not have to be the current latest seqno in the db.

jijiew:
> I think a few more test cases might be in order:
> 
> * What happens if you ingest when data is not flushed?
> * What happens if you Get after ingest without the compact?
> * What happens if there are more keys in the ingested file (e.g. {j, k}, {k, l}, {j,k,l} where the key in question is not always the smallest and/or largest
> 
> Generally, this looks okay to me, but I am not an expert in this area....

thanks for the review, more test added.

db/external_sst_file_basic_test.cc
jijiew: removed, not necessary.

db/external_sst_file_basic_test.cc
jijiew: Done.

db/external_sst_file_ingestion_job.cc
jijiew: hmm, I Could not find the key.type, the value type is defined here:
https://github.com/facebook/rocksdb/blob/711881bc257d761fd44924686069cb5a0cea580b/db/dbformat.h

db/external_sst_file_ingestion_job.cc
jijiew: same above, I could not find the key.type.

db/external_sst_file_ingestion_job.cc
jijiew: just did a bit of refactoring to include that part.

ajkr (commented):
A couple things to do:

- run `make format`
- do you mind mentioning this under the unreleased bug fixes section in the HISTORY.md file?

A general comment on the approach - it prevents newly ingested files from being written with incorrect smallest/largest key metadata, which is a good step forward. Reading from a DB with pre-existing ingested files could still be problematic. That appears difficult to fix. Perhaps we can have a consistency check that `FileMetaData::smallest` and `FileMetaData::largest` have seqnums in [`FileMetaData::smallest_seqno`, `FileMetaData::largest_seqno`]. This can be done later.

PR #8204, opened 2021-04-19
Fix unittest no space issue

jay-zhuang:
Unittest reports no space from time to time, which can be reproduced on a small memory machine with SHM. It's caused by large WAL files generated during the test, which is preallocated, but didn't truncate during close(). Adding the missing APIs to set preallocation.
It added arm test as nightly build, as the test runs more than 1 hour.

Test Plan: test on small memory arm machine

### DIFF
--- a/env/env_encryption.cc
+++ b/env/env_encryption.cc
@@ -366,6 +366,11 @@ void EncryptedWritableFile::SetPreallocationBlockSize(size_t size) {
   file_->SetPreallocationBlockSize(size);
 }
 
+void EncryptedWritableFile::GetPreallocationStatus(size_t* block_size,
+                                                   size_t* last_allocated_block) {
+  file_->GetPreallocationStatus(block_size, last_allocated_block);
+}
+
 // Pre-allocates space for a file.
 IOStatus EncryptedWritableFile::Allocate(uint64_t offset, uint64_t len,
                                          const IOOptions& options,
--- a/env/env_encryption.cc
+++ b/env/env_encryption.cc
@@ -363,6 +363,8 @@ void EncryptedWritableFile::PrepareWrite(size_t offset, size_t len,
 }
 
 void EncryptedWritableFile::SetPreallocationBlockSize(size_t size) {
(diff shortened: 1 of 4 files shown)


env/env_encryption.cc
```diff
@@ -362,6 +362,15 @@ void EncryptedWritableFile::PrepareWrite(size_t offset, size_t len,
   file_->PrepareWrite(offset + prefixLength_, len, options, dbg);
 }
 
+void EncryptedWritableFile::SetPreallocationBlockSize(size_t size) {
+  file_->SetPreallocationBlockSize(size);
+}
```
mrambacher, line 367: Is this correct?  I thought all of the encrypted files were written in more certain block sizes with potentially a long header.  Is this keeping with that?  It might be correct, but a comment/explanation would be helpful...

env/env_encryption.cc
```diff
@@ -362,6 +362,15 @@ void EncryptedWritableFile::PrepareWrite(size_t offset, size_t len,
...
+
+void EncryptedWritableFile::GetPreallocationStatus(
+    size_t* block_size, size_t* last_allocated_block) {
+  file_->GetPreallocationStatus(block_size, last_allocated_block);
```
mrambacher, line 371: Aren't these methods already there for "free" from the WritableFileWrapper?

mrambacher (approved):
The changes to the encryption classes appear superfluous to me but do no harm.  Otherwise LGTM.

env/env_encryption.cc
jay-zhuang: Added comment for that.
Size here is the preallocation size, there's no need to add or remove `prefixLength_`. `PrepareWrite()` will use `preallocation size` to preallocate if needed, which is including the `prefixLength_`.

env/env_encryption.cc
jay-zhuang: `EncryptedWritableFile` is kind of another wrapper here right? Which overrides some functions, but just forward other functions.

PR #8205, opened 2021-04-19
Update HISTORY.md on bug fix for * Fix on issue#6245

jijiew:
Fix a bug that prevents newly ingested files from being written with incorrect smallest/largest key metadata, which is a good step forward. Reading from a DB with pre-existing ingested files could still be problematic. That appears difficult to fix. Perhaps we can have a consistency check that FileMetaData::smallest and FileMetaData::largest have seqnums in [FileMetaData::smallest_seqno, FileMetaData::largest_seqno]

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,6 +13,7 @@
 * Fixed crash (divide by zero) when compression dictionary is applied to a file containing only range tombstones.
 * Fixed a backward iteration bug with partitioned filter enabled: not including the prefix of the last key of the previous filter partition in current filter partition can cause wrong iteration result.
 * Fixed a bug that allowed `DBOptions::max_open_files` to be set with a non-negative integer with `ColumnFamilyOptions::compaction_style = kCompactionStyleFIFO`.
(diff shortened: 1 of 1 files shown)


riversand963:
Is this part of #8201?

riversand963:
Thanks @jijiew for the PR. I think this should be part of #8201, and can be closed.

Furthermore, it looks like you are pushing from local working directory directly into upstream: `github.com/facebook/rocksdb/`. For PRs, especially for bug fixes, we create a branch in our own fork, and create pull request from there.
Can you check the output of executing the following:
```
$git remote -v
```
The result should be something like the following:
```
origin	git@github.com:<your git username>/rocksdb.git (fetch)
origin	git@github.com:<your git username>/rocksdb.git (push)
upstream	https://github.com/facebook/rocksdb.git (fetch)
upstream	https://github.com/facebook/rocksdb.git (push)
```
Your local branch is jijiew-patch-1.
When you push, please push to origin, assuming origin is "git@github.com:<your git username>/rocksdb.git"
```
git push origin jijiew-patch-1:jijiew-patch-1
```

PR #8212, opened 2021-04-20
Revert Ribbon starting level support from #8198

pdillinger:
Summary:
This partially reverts commit 10196d7edc2fc5c03553c76acaf1337b5c7c1718.

The problem with this change is because of important filter use cases:
FIFO compaction and SST writer. FIFO "compaction" always uses level 0 so
would only use Ribbon filters if specifically including level 0 for the
Ribbon filter policy. SST writer sets level_at_creation=-1 to indicate
unknown level, and this would be treated the same as level 0 unless
fixed.

We are keeping the part about committing to permanent schema, which is
only changes to API comments and HISTORY.md.

Test Plan: CI


pdillinger:
This (partial) revert would be cherry-picked into 6.20 before release.

jay-zhuang (approved):

PR #8213, opened 2021-04-21
Allow cache_bench and db_bench to use 3rd party tiered cache

anand1976:
This diff contains the following changes to allow versions of cache_bench and db_bench that can use a 3rd party tiered cache allocated using the object registry -
1. Provide a public wrapper around ```ConfigurableHelper::RegisterOptions()``` so a 3rd party implementation can register an ```OptionTypeInfo``` map for easy parsing of option strings
2. Add the ```tiered_cache_uri``` gflag in cache_bench and db_bench, and allocate  a ```TieredCache``` object using the object registry
3. Make ```cache_bench_tool.cc``` a library function


include/rocksdb/configurable.h
```diff
@@ -356,4 +356,8 @@ class Configurable {
   // Configurable option via
   std::vector<RegisteredOptions> options_;
 };
+
+extern void RegisterConfigurableOptions(
+    Configurable& configurable, const std::string& name, void* opt_ptr,
+    const std::unordered_map<std::string, OptionTypeInfo>* opt_map);
```
mrambacher, line 362: I am not sure what the purpose of exposing this method is.  What are you hoping to accomplish?

I do not see where this method is used anywhere in this check-in.  Perhaps this should be moved into a separate PR?

include/rocksdb/cache.h
```diff
@@ -417,6 +345,78 @@ class Cache {
...
+  // cache may or may not write it to NVM depending on its admission
+  // policy.
+  //
+  // When the inserted entry is no longer needed, the key and
+  // value will be passed to "deleter".
+  virtual Status Insert(const Slice& key, void* value,
+                        CacheItemHelperCallback helper_cb, size_t charge,
+                        Handle** handle = nullptr,
+                        Priority priority = Priority::LOW) {
+    DeletionCallback delete_cb = nullptr;
```
mrambacher, line 375: assert helper_cb != nullptr?

include/rocksdb/configurable.h
anand1976: The purpose of this PR is to enable 3rd party implementations of ```TieredCache```. The URI will contain the type, as well as any options required to instantiate the object. For example, you could have "cachelib://path_to_cache_file/size=1G;buffers=10M;io_threads=16". We currently allow 3rd party implementations of Env etc. to register a factory function with ```ObjectLibrary``` for the URI pattern, but I couldn't find any support for passing option name-value pairs.

Let me know if there's a better way. I'm open to any suggestions.

include/rocksdb/configurable.h
```diff
@@ -356,4 +356,8 @@ class Configurable {
   // Configurable option via
   std::vector<RegisteredOptions> options_;
 };
+
+extern void RegisterConfigurableOptions(
```
zhichao-cao, line 360: Maybe a comment to explain it will be used by the third party to create their own OptionTypeInfo map for the plugin object configuration options.

include/rocksdb/configurable.h
zhichao-cao: @mrambacher do your future PRs on Configurable be able to handle the configuration from the URI string?

mrambacher: @zhichao-cao @anand1976 Yes, it would be a straightforward exercise to make TieredCache into a Customizable object that can be configured from an input string.  The string would look slightly different than what may be proposed here (id=TieredCacheClass; prop1=value1; etc), but it should be easy to implement.

cache/cache_bench_tool.cc
```diff
@@ -0,0 +1,506 @@
...
+  void AddLookupUs(uint64_t lookup) { lookup_us += lookup; }
+  void AddInsertUs(uint64_t insert) { insert_us += insert; }
+  void Merge(const Stats& other) {
+    hits += other.hits;
+    misses += other.misses;
+    lookup_us += other.lookup_us;
+    insert_us += other.insert_us;
+  }
+  void Report() {
+    fprintf(stdout, "%lu hits, %lu misses, %lu lookup_ns, %lu insert_ns\n",
```
zhichao-cao, line 140: What's the lookup_ns and insert_ns meanse? us or ns?

cache/cache_bench_tool.cc
```diff
@@ -0,0 +1,506 @@
...
+  ThreadState(uint32_t index, SharedState* _shared)
+      : tid(index), rnd(1000 + index), shared(_shared) {}
+};
+
+struct KeyGen {
+  char key_data[27];
+
+  Slice GetRand(Random64& rnd, uint64_t max_key, int max_log) {
+    uint64_t key = 0;
+    if (!FLAGS_skewed) {
```
zhichao-cao, line 162: If !FLAGS_skewed, should we directly use random instead of using FLAGS_skew?

zhichao-cao (approved):
LGTM

cache/cache_bench_tool.cc
zhichao-cao: > The original algorithm used `FLAGS_skew` to generate the random key. So just retaining that behavior if `FLAGS_skewed` is not specified.

Yeah, after compared the original code yesterday, we can just follow the original behavior.

PR #8216, opened 2021-04-21
Ignore comparator name mismatch in ldb manifest dump

riversand963:
RocksDB allows user-specified custom comparators which may not be known to `ldb`,
a built-in tool for checking/mutating the database. Therefore, column family comparator
names mismatch encountered during manifest dump should not prevent the dumping from
proceeding.

Test plan:
```
make check
```

Also manually do the following
```
KEEP_DB=1 ./db_with_timestamp_basic_test
./ldb --db=<db> manifest_dump --verbose
```
The ldb should succeed and print something like:
```
...
--------------- Column family "default"  (ID 0) --------------
log number: 6
comparator: <TestComparator>, but the comparator object is not available.
...
```

### DIFF
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -903,7 +903,7 @@ void DumpManifestHandler::CheckIterationResult(const log::Reader& reader,
     auto it = cf_to_cmp_names_->find(cfd->GetID());
     if (it != cf_to_cmp_names_->end()) {
       fprintf(stdout,
-              "comparator: <%s>, but the comparator object is not avaiable.\n",
+              "comparator: <%s>, but the comparator object is not available.\n",
               it->second.c_str());
     } else {
       fprintf(stdout, "comparator: %s\n", cfd->user_comparator()->Name());
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -553,9 +553,13 @@ Status VersionEditHandler::ExtractInfoFromVersionEdit(ColumnFamilyData* cfd,
     }
     if (edit.has_comparator_ &&
         edit.comparator_ != cfd->user_comparator()->Name()) {
-      s = Status::InvalidArgument(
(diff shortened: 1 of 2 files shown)


ltamasi (approved):

db/version_edit_handler.cc
```diff
@@ -889,13 +893,21 @@ void DumpManifestHandler::CheckIterationResult(const log::Reader& reader,
...
+  assert(cf_to_cmp_names_);
   for (auto* cfd : *(version_set_->column_family_set_)) {
     fprintf(stdout,
             "--------------- Column family \"%s\"  (ID %" PRIu32
             ") --------------\n",
             cfd->GetName().c_str(), cfd->GetID());
     fprintf(stdout, "log number: %" PRIu64 "\n", cfd->GetLogNumber());
-    fprintf(stdout, "comparator: %s\n", cfd->user_comparator()->Name());
+    auto it = cf_to_cmp_names_->find(cfd->GetID());
+    if (it != cf_to_cmp_names_->end()) {
```
mrambacher, line 904: This is not exactly the same test as in the Extract case, since the extract case depended on has_comparator_.  Does that matter or not for these purposes?

riversand963: Sorry missed this.
The check in `ExtractInfoFromVersionEdit` is performed while reading MANIFEST. On seeing a version edit that contains comparator name (has_comparator_), ldb compares the comparator name with the one specified via `options` with the name recorded in the MANIFEST. If they do not match, it may just be because the comparator is a custom comparator which `ldb` does not know. In order to be able to print its name later, we save the `cf_id->comparator` mapping.
The two lines you quoted here are executed after reading the MANIFEST. At this point, the column family's comparator name is either equal to `cfd->user_comparator()->Name()`, or saved in the map. There must have been a version edit that has comparator, due to how we create column family in `CreateColumnFamilyImpl` and `WriteCurrentStateToManifest`.

PR #8219, opened 2021-04-22
Eliminate double-buffering of keys in block_based_table_builder

saketh-are:
The block_based_table_builder buffers some blocks in memory to construct a good compression dictionary. Before this commit, the keys from each block were buffered separately for convenience. However, the buffered block data implicitly contains all keys. This commit eliminates the redundant key buffers and reduces memory usage.

### DIFF
--- a/table/block_based/block_based_table_builder.cc
+++ b/table/block_based/block_based_table_builder.cc
@@ -254,13 +254,10 @@ struct BlockBasedTableBuilder::Rep {
   std::atomic<uint64_t> offset;
   size_t alignment;
   BlockBuilder data_block;
-  // Buffers uncompressed data blocks and keys to replay later. Needed when
+  // Buffers uncompressed data blocks to replay later. Needed when
   // compression dictionary is enabled so we can finalize the dictionary before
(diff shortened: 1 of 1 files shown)


table/block_based/block_based_table_builder.cc
```diff
@@ -1754,49 +1742,84 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
...
+      for (; iter->Valid(); iter->Next()) {
+        Slice key = iter->key();
         if (r->filter_builder != nullptr) {
           size_t ts_sz =
               r->internal_comparator.user_comparator()->timestamp_size();
           r->filter_builder->Add(ExtractUserKeyAndStripTimestamp(key, ts_sz));
         }
         r->index_builder->OnKeyAdded(key);
+
+        last_key = key.ToString();
```
ajkr, line 1805: This could produce a lot of small/odd-sized allocations. What about calling `SeekToLast()` after this loop to find `last_key`?

ajkr (approved):
LGTM! Thanks for saving dictionary compression users memory and allocations!

PR #8220, opened 2021-04-22
Change CircleCI Windows to previous known good image

anand1976:
This is to try to resolve the VS2015 install failure in CircleCI Windows builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -113,7 +113,7 @@ commands:
 executors:
   windows-2xlarge:
     machine:
-      image: 'windows-server-2019-vs2019:previous'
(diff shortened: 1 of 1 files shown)


jay-zhuang:
Seems VS 2015 installation is still failing, the error code is 1603, which means:
```
it could mean a pending reboot is necessary prior to install or something else (like the same version is already installed)
```
seems choco has the same problem:
https://app.circleci.com/pipelines/github/facebook/rocksdb/6758/workflows/4886b474-baff-4541-9ada-46994e3dd4e1/jobs/125082
I filed a ticket to CircleCI.

jay-zhuang (approved):

PR #8223, opened 2021-04-22
Move RegisterOptions into the Configurable API

mrambacher:
As previously coded, a Configurable extension would need access to code not in the public API.  This change moves RegisterOptions into the Configurable class and therefore available to public extensions.


anand1976 (approved):
Thanks for the PR! This will allow code outside the repo to leverage the options framework.

PR #8225, opened 2021-04-23
New Cache API for gathering statistics

pdillinger:
Summary: Adds a new Cache::ApplyToAllEntries API that we expect to use
(in follow-up PRs) for efficiently gathering block cache statistics.
Notable features vs. old ApplyToAllCacheEntries:

* Includes key and deleter (in addition to value and charge). We could
have passed in a Handle but then more virtual function calls would be
needed to get the "fields" of each entry. We expect to use the 'deleter'
to identify the origin of entries, perhaps even more.
* Heavily tuned to minimize latency impact on operating cache. It
does this by iterating over small sections of each cache shard while
cycling through the shards.
* Supports tuning roughly how many entries to operate on for each
lock acquire and release, to control the impact on the latency of other
operations without excessive lock acquire & release. The right balance
can depend on the cost of the callback. Good default seems to be
around 256.
* There should be no need to disable thread safety. (I would expect
uncontended locks to be sufficiently fast.)

I have enhanced cache_bench to validate this approach:

* Reports a histogram of ns per operation, so we can look at the
ditribution of times, not just throughput (average).
* Can add a thread for simulated "gather stats" which calls
ApplyToAllEntries at a specified interval. We also generate a histogram
of time to run ApplyToAllEntries.

To make the iteration over some entries of each shard work as cleanly as
possible, even with resize between next set of entries, I have
re-arranged which hash bits are used for sharding and which for indexing
within a shard.

Test Plan: A couple of unit tests are added, but primary validation is manual, as
the primary risk is to performance.

The primary validation is using cache_bench to ensure that neither
the minor hashing changes nor the simulated stats gathering
significantly impact QPS or latency distribution. Note that adding op
latency histogram seriously impacts the benchmark QPS, so for a
fair baseline, we need the cache_bench changes (except remove simulated
stat gathering to make it compile). In short, we don't see any
reproducible difference in ops/sec or op latency unless we are gathering
stats nearly continuously. Test uses 10GB block cache with
8KB values to be somewhat realistic in the number of items to iterate
over.

Baseline typical output:

```
Complete in 92.017 s; Rough parallel ops/sec = 869401
Thread ops/sec = 54662

Operation latency (ns):
Count: 80000000 Average: 11223.9494  StdDev: 29.61
Min: 0  Median: 7759.3973  Max: 9620500
Percentiles: P50: 7759.40 P75: 14190.73 P99: 46922.75 P99.9: 77509.84 P99.99: 217030.58
------------------------------------------------------
[       0,       1 ]       68   0.000%   0.000% 
(    2900,    4400 ]       89   0.000%   0.000% 
(    4400,    6600 ] 33630240  42.038%  42.038% ########
(    6600,    9900 ] 18129842  22.662%  64.700% #####
(    9900,   14000 ]  7877533   9.847%  74.547% ##
(   14000,   22000 ] 15193238  18.992%  93.539% ####
(   22000,   33000 ]  3037061   3.796%  97.335% #
(   33000,   50000 ]  1626316   2.033%  99.368% 
(   50000,   75000 ]   421532   0.527%  99.895% 
(   75000,  110000 ]    56910   0.071%  99.966% 
(  110000,  170000 ]    16134   0.020%  99.986% 
(  170000,  250000 ]     5166   0.006%  99.993% 
(  250000,  380000 ]     3017   0.004%  99.996% 
(  380000,  570000 ]     1337   0.002%  99.998% 
(  570000,  860000 ]      805   0.001%  99.999% 
(  860000, 1200000 ]      319   0.000% 100.000% 
( 1200000, 1900000 ]      231   0.000% 100.000% 
( 1900000, 2900000 ]      100   0.000% 100.000% 
( 2900000, 4300000 ]       39   0.000% 100.000% 
( 4300000, 6500000 ]       16   0.000% 100.000% 
( 6500000, 9800000 ]        7   0.000% 100.000% 
```

New, gather_stats=false. Median thread ops/sec of 5 runs:

```
Complete in 92.030 s; Rough parallel ops/sec = 869285
Thread ops/sec = 54458

Operation latency (ns):
Count: 80000000 Average: 11298.1027  StdDev: 42.18
Min: 0  Median: 7722.0822  Max: 6398720
Percentiles: P50: 7722.08 P75: 14294.68 P99: 47522.95 P99.9: 85292.16 P99.99: 228077.78
------------------------------------------------------
[       0,       1 ]      109   0.000%   0.000% 
(    2900,    4400 ]      793   0.001%   0.001% 
(    4400,    6600 ] 34054563  42.568%  42.569% #########
(    6600,    9900 ] 17482646  21.853%  64.423% ####
(    9900,   14000 ]  7908180   9.885%  74.308% ##
(   14000,   22000 ] 15032072  18.790%  93.098% ####
(   22000,   33000 ]  3237834   4.047%  97.145% #
(   33000,   50000 ]  1736882   2.171%  99.316% 
(   50000,   75000 ]   446851   0.559%  99.875% 
(   75000,  110000 ]    68251   0.085%  99.960% 
(  110000,  170000 ]    18592   0.023%  99.983% 
(  170000,  250000 ]     7200   0.009%  99.992% 
(  250000,  380000 ]     3334   0.004%  99.997% 
(  380000,  570000 ]     1393   0.002%  99.998% 
(  570000,  860000 ]      700   0.001%  99.999% 
(  860000, 1200000 ]      293   0.000% 100.000% 
( 1200000, 1900000 ]      196   0.000% 100.000% 
( 1900000, 2900000 ]       69   0.000% 100.000% 
( 2900000, 4300000 ]       32   0.000% 100.000% 
( 4300000, 6500000 ]       10   0.000% 100.000% 
```

New, gather_stats=true, 1 second delay between scans. Scans take about
1 second here so it's spending about 50% time scanning. Still the effect on
ops/sec and latency seems to be in the noise. Median thread ops/sec of 5 runs:

```
Complete in 91.890 s; Rough parallel ops/sec = 870608
Thread ops/sec = 54551

Operation latency (ns):
Count: 80000000 Average: 11311.2629  StdDev: 45.28
Min: 0  Median: 7686.5458  Max: 10018340
Percentiles: P50: 7686.55 P75: 14481.95 P99: 47232.60 P99.9: 79230.18 P99.99: 232998.86
------------------------------------------------------
[       0,       1 ]       71   0.000%   0.000% 
(    2900,    4400 ]      291   0.000%   0.000% 
(    4400,    6600 ] 34492060  43.115%  43.116% #########
(    6600,    9900 ] 16727328  20.909%  64.025% ####
(    9900,   14000 ]  7845828   9.807%  73.832% ##
(   14000,   22000 ] 15510654  19.388%  93.220% ####
(   22000,   33000 ]  3216533   4.021%  97.241% #
(   33000,   50000 ]  1680859   2.101%  99.342% 
(   50000,   75000 ]   439059   0.549%  99.891% 
(   75000,  110000 ]    60540   0.076%  99.967% 
(  110000,  170000 ]    14649   0.018%  99.985% 
(  170000,  250000 ]     5242   0.007%  99.991% 
(  250000,  380000 ]     3260   0.004%  99.995% 
(  380000,  570000 ]     1599   0.002%  99.997% 
(  570000,  860000 ]     1043   0.001%  99.999% 
(  860000, 1200000 ]      471   0.001%  99.999% 
( 1200000, 1900000 ]      275   0.000% 100.000% 
( 1900000, 2900000 ]      143   0.000% 100.000% 
( 2900000, 4300000 ]       60   0.000% 100.000% 
( 4300000, 6500000 ]       27   0.000% 100.000% 
( 6500000, 9800000 ]        7   0.000% 100.000% 
( 9800000, 14000000 ]        1   0.000% 100.000% 

Gather stats latency (us):
Count: 46 Average: 980387.5870  StdDev: 60911.18
Min: 879155  Median: 1033777.7778  Max: 1261431
Percentiles: P50: 1033777.78 P75: 1120666.67 P99: 1261431.00 P99.9: 1261431.00 P99.99: 1261431.00
------------------------------------------------------
(  860000, 1200000 ]       45  97.826%  97.826% ####################
( 1200000, 1900000 ]        1   2.174% 100.000% 

Most recent cache entry stats:
Number of entries: 1295133
Total charge: 9.88 GB
Average key size: 23.4982
Average charge: 8.00 KB
Unique deleters: 3
```


cache/lru_cache.cc
```diff
@@ -61,27 +65,24 @@ LRUHandle* LRUHandleTable::Remove(const Slice& key, uint32_t hash) {
...
+  LRUHandle** new_list = new LRUHandle* [1U << new_length_bits] {};
   uint32_t count = 0;
-  for (uint32_t i = 0; i < length_; i++) {
+  for (uint32_t i = 0; (i >> old_length_bits) == 0; i++) {
     LRUHandle* h = list_[i];
     while (h != nullptr) {
       LRUHandle* next = h->next_hash;
       uint32_t hash = h->hash;
-      LRUHandle** ptr = &new_list[hash & (new_length - 1)];
+      LRUHandle** ptr = &new_list[hash >> (32 - new_length_bits)];
```
mrambacher, line 85: Can you comment on this?  Is the >> meant to be more efficient than the old way or something else?  

It seems like you will be using different bits from the hash at least for the mapping.  Will this break anything with persistent hashes or the like?

cache/lru_cache.h
```diff
@@ -181,10 +182,15 @@ class LRUHandleTable {
 
   void Resize();
 
+  // Number of hash bits (upper because lower bits used for sharding)
+  // used for table index. Length == 1 << length_bits_
+  int length_bits_;
```
mrambacher, line 187: uint32_t?

cache/lru_cache.cc
```diff
@@ -137,19 +138,37 @@ void LRUCacheShard::EraseUnRefEntries() {
...
+                             DeleterFn deleter)>& callback,
+    uint32_t average_entries_per_lock, uint32_t* state) {
+  // The state is essentially going to be the starting hash, which works
+  // nicely even if we resize between calls because we use upper-most
+  // hash bits for table indexes.
+  MutexLock l(&mutex_);
+  uint32_t length_bits = table_.GetLengthBits();
+  uint32_t length = uint32_t{1} << length_bits;
+
+  assert(average_entries_per_lock > 0);
```
mrambacher, line 152: Why must average_entries_per_lock > 0?  Would a value of 0 allow "unlocked" access like there was before?

include/rocksdb/cache.h
```diff
@@ -273,11 +281,27 @@ class Cache {
...
+  // also. If the Cache uses locks, setting `average_entries_per_lock` to
+  // a higher value suggests iterating over more entries each time a lock
+  // is acquired, likely reducing the time for ApplyToAllEntries but
+  // increasing latency for concurrent users of the Cache. Setting
+  // `average_entries_per_lock` to a smaller value could be helpful if
+  // callback is relatively expensive, such as using large data structures.
+  virtual void ApplyToAllEntries(
+      const std::function<void(const Slice& key, void* value, size_t charge,
+                               DeleterFn deleter)>& callback,
+      size_t average_entries_per_lock = 256) = 0;
```
mrambacher, line 296: Is this an incompatible change and is there any way around it?  As it stands now, any existing Cache implementation will not work/compile.  

Are we okay with forcing code updates?

cache/cache_bench.cc
```diff
@@ -206,28 +225,33 @@ class CacheBench {
...
-    const auto& clock = env->GetSystemClock();
+    const auto clock = SystemClock::Default().get();
 
     PrintEnv();
     SharedState shared(this);
     std::vector<std::unique_ptr<ThreadState> > threads(FLAGS_threads);
     for (uint32_t i = 0; i < FLAGS_threads; i++) {
       threads[i].reset(new ThreadState(i, &shared));
-      env->StartThread(ThreadBody, threads[i].get());
+      std::thread(ThreadBody, threads[i].get()).detach();
```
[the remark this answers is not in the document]
pdillinger: Env threads don't use template parameter pack, so ugly to use.

cache/cache_bench.cc
```diff
@@ -288,9 +389,11 @@ class CacheBench {
     // To hold handles for a non-trivial amount of time
     Cache::Handle* handle = nullptr;
     KeyGen gen;
+    StopWatchNano timer(SystemClock::Default().get());
     for (uint64_t i = 0; i < FLAGS_ops_per_thread; i++) {
       Slice key = gen.GetRand(thread->rnd, max_key_);
       uint64_t random_op = thread->rnd.Next();
+      timer.Start();
```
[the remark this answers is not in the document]
pdillinger: I'm trying to measure the latency of Cache operations, so the less "other stuff" that is included, the more accurate that is. I guess I would need to exclude timing the "do something with the data" sections to be more accurate.

cache/lru_cache.cc
pdillinger: From the description:

To make the iteration over some entries of each shard work as cleanly as
possible, even with resize between next set of entries, I have
re-arranged which hash bits are used for sharding and which for indexing
within a shard.

cache/lru_cache.h
pdillinger: For a shift amount, there's no signed/unsigned pollution (with decent compilers, `1U << some_int` is unsigned) and the values are small, so under Google C++ style, the default choice is `int`.

include/rocksdb/cache.h
pdillinger: I think custom Cache implementations are rare enough, and the likely benefit high enough to impose this. I tried other approaches to gathering the statistics but they hit critical areas of performance. Benchmark testing on this approach looks like it will be almost free even if we gathered the stats at intervals of seconds rather than minutes or hours.

Rules (e.g. "no breaking API changes") are negotiable, especially at Facebook. ;)

cache/cache_bench.cc
```diff
@@ -160,7 +173,13 @@ char* createValue(Random64& rnd) {
...
 }
 
-void deleter(const Slice& /*key*/, void* value) {
+void deleter1(const Slice& /*key*/, void* value) {
+  delete[] static_cast<char*>(value);
+}
+void deleter2(const Slice& /*key*/, void* value) {
+  delete[] static_cast<char*>(value);
+}
+void deleter3(const Slice& /*key*/, void* value) {
```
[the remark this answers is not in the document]
pdillinger: This is my planned backdoor way of identifying the origin and kind of cache entries. Unlike adding a tag to keys or values, there's no space/time impact to the critical performance path. It gives us an option to turn on "type checking" of cache entries if we want extra verification against memory corruption, which is non-trivial impact on block cache lookups. But optional as I said.

cache/lru_cache.cc
pdillinger: Because it's more efficient to sanitize the option only once, in the caller. This part of the implementation would make no progress with average_entries_per_lock=0.

cache/lru_cache.cc
```diff
@@ -61,27 +65,26 @@ LRUHandle* LRUHandleTable::Remove(const Slice& key, uint32_t hash) {
...
-  LRUHandle** new_list = new LRUHandle*[new_length];
-  memset(new_list, 0, sizeof(new_list[0]) * new_length);
+  // FIXME: don't resize beyond shard_bits + length_bits == 32, unless
+  // we upgrade hash (storage) to 64 bits.
+  int old_length_bits = length_bits_;
+  int new_length_bits = length_bits_ + 1;
+  LRUHandle** new_list = new LRUHandle* [size_t{1} << new_length_bits] {};
   uint32_t count = 0;
-  for (uint32_t i = 0; i < length_; i++) {
+  for (uint32_t i = 0; (i >> old_length_bits) == 0; i++) {
```
mrambacher, line 82: Is this more efficient than having a temporary variable "old_length = 1 << old_length_bits;"  and use old_length as your loop termination condition?

cache/lru_cache.h
```diff
@@ -173,6 +172,8 @@ class LRUHandleTable {
     }
   }
 
+  int GetLengthBits() { return length_bits_; }
```
mrambacher, line 175: mark as const

mrambacher (commented):
Please update HISTORY.md with the changes to the public API.

cache/lru_cache.cc
```diff
@@ -10,23 +10,27 @@
...
 #include "util/mutexlock.h"
 
 namespace ROCKSDB_NAMESPACE {
 
-LRUHandleTable::LRUHandleTable() : list_(nullptr), length_(0), elems_(0) {
-  Resize();
-}
+LRUHandleTable::LRUHandleTable()
+    : length_bits_(4),
+      list_(new LRUHandle* [size_t{1} << length_bits_] {}),
```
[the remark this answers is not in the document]
pdillinger: Yes. https://en.cppreference.com/w/cpp/language/value_initialization

cache/cache_bench.cc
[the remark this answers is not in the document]
mrambacher: As an aside, can you please file an issue as to what the Env threads should do to meet the needs here?  It would be good to fix them before this sort of thing propagates too far if it is easy to do.

include/rocksdb/cache.h
mrambacher: Nit: The google coding standard does not allow default arguments on virtual functions.  Can we split this into two methods, one (not virtual) with no "average_entries_per_lock" and a virtual one without the default

mrambacher (approved):
Be nice to fix the virtual functions with default args, but it seems like an existing problem with this class already...

PR #8226, opened 2021-04-23
Add WAL flush API to C client

duarten:
The C client is missing the`manual_wal_flush` option and the `flush_wal` API.

### DIFF
--- a/db/c_test.c
+++ b/db/c_test.c
@@ -1280,7 +1280,7 @@ int main(int argc, char** argv) {
     rocksdb_writebatch_destroy(wb);
 
     rocksdb_flush_wal(db, 1, &err);
-    CheckNoError(err)
+    CheckNoError(err);
 
     const char* keys[3] = { "box", "box", "barfooxx" };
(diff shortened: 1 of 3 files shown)


db/c_test.c
```diff
@@ -1276,6 +1276,9 @@ int main(int argc, char** argv) {
     CheckPinGetCF(db, roptions, handles[1], "box", "c");
     rocksdb_writebatch_destroy(wb);
 
+    rocksdb_flush_wal(db, 1, &err);
+    CheckNoError(err)
```
mrambacher, line 1280: Nit: Please put a ";" at the end of line

mrambacher (approved):
LGTM.

There is no change in the test behavior by leaving this option set, correct?  Just curious...

db/c_test.c
duarten: Do you mean the `manual_wal_flush`? If so, I think that test only tests the setters and getters; the options object is destroyed afterwards.

PR #8227, opened 2021-04-24
Rename variables in ImmutableCFOptions to avoid conflicts with ImmutableDBOptions

mrambacher:
Renaming ImmutableCFOptions::info_log and statistics to logger and stats.  This is stage 2 in creating an ImmutableOptions class.  It is necessary because the names match those in ImmutableOptions and have different types.


jay-zhuang (approved):
As `ImmutableCFOptions` is internal structure, I think renaming `info_log` and `statistics` should be fine.

PR #8228, opened 2021-04-24
Fix variable shadowing

adamretter:
Fixes an issue with compilation on macOS.


adamretter:
Looks like these were already fixed and I may have been working from an older branch

PR #8243, opened 2021-04-29
Allow applying `CompactionFilter` outside of compaction

ajkr:
From HISTORY.md release note:

- Allow `CompactionFilter`s to apply in more table file creation scenarios such as flush and recovery. For compatibility, `CompactionFilter`s by default apply during compaction. Users can customize this behavior by overriding `CompactionFilterFactory::ShouldFilterTableFileCreation()`.
- Removed unused structure `CompactionFilterContext`

Test Plan: added unit tests


db/compaction/compaction_iterator.cc
```diff
@@ -251,6 +251,8 @@ bool CompactionIterator::InvokeFilterIfNeeded(bool* need_skip,
           valid_ = false;
           return false;
         }
+        // TODO(ajkr): What happens when this runs in flush/recovery, i.e., when
```
ajkr, line 254: @ltamasi would be good to get your input here. I can investigate a bit more but am leaning toward setting a non-ok status and returning false in this case, unless anyone has a different suggestion.

db/db_compaction_filter_test.cc
```diff
@@ -192,16 +197,28 @@ class KeepFilterFactory : public CompactionFilterFactory {
 
 class DeleteFilterFactory : public CompactionFilterFactory {
  public:
+  DeleteFilterFactory(TableFileCreationReason reason) : reason_(reason) {}
+
```
mrambacher, line 201: I am having a little trouble following this class.  It looks like you can create it with a reason="Flush" but that option won't create a CompactionFilter?

Adding some comments as to why this is true to this class or what the intent would be helpful.

include/rocksdb/compaction_filter.h
```diff
@@ -13,24 +13,16 @@
 #include <string>
 #include <vector>
 
+#include "rocksdb/listener.h"
```
mrambacher, line 16: Rather than including listener, can we include "types.h" and move the enum you are using into there?  The incestual nature of the includes makes the builds take forever...

include/rocksdb/compaction_filter.h
```diff
@@ -13,24 +13,16 @@
...
 class SliceTransform;
 
-// Context information of a compaction run
-struct CompactionFilterContext {
-  // Does this compaction run include all data files
-  bool is_full_compaction;
-  // Is this compaction requested by the client (true),
-  // or is it occurring as an automatic compaction process
-  bool is_manual_compaction;
-};
```
mrambacher, line 30: Just curious on if removing this struct will cause any compatibility issues with existing deployments.  It is in the public API...

mrambacher (commented):
Are there any tests that run the CompactionFilter outside of the normal Compaction process?  I see the NotSupported that check early, but did not see any that invoke the CF (though I might have missed them).

db/compaction/compaction_iterator.cc
ltamasi: This logic pertains to the integrated BlobDB implementation and can be only hit during compaction, since large values are not yet written to blob files at this point during flush/recovery. (That happens in `ExtractLargeValueIfNeeded`, which is called from `PrepareOutput`.) So I think what you suggested would work :)

include/rocksdb/compaction_filter.h
ajkr: Maybe. It's unused in RocksDB for a long time though. It looks like removing this was forgotten in 9803e0d813104112f05e2dcea26058cb7cf1e408 (2015). I can add a release note.

db/compaction/compaction_iterator.cc
ajkr: Thanks!

ajkr (commented):
> Are there any tests that run the CompactionFilter outside of the normal Compaction process? I see the NotSupported that check early, but did not see any that invoke the CF (though I might have missed them).

Yes, the test names are "CompactionFilterFlush" and "CompactionFilterRecovery".

mrambacher:
What is the (expected) behavior of the flush and recover cases when the ColumnFamily has a CompactionFilterFactory but not MergeOperator?  Can tests be added for those cases as well?

db/db_compaction_filter_test.cc
ajkr: > Adding some comments

Done.

include/rocksdb/compaction_filter.h
ajkr: Good point, done.

include/rocksdb/compaction_filter.h
ajkr: > add a release note

Done.

db/compaction/compaction_iterator.cc
ajkr: > setting a non-ok status and returning false in this case

Done.

include/rocksdb/compaction_filter.h
```diff
@@ -214,16 +210,28 @@ class CompactionFilter {
...
  public:
   virtual ~CompactionFilterFactory() {}
 
+  // Returns whether a thread creating table files for the specified `reason`
+  // should invoke `CreateCompactionFilter()` and pass KVs through the returned
+  // filter.
+  virtual bool ShouldFilterTableFileCreation(
+      TableFileCreationReason reason) const {
+    // Default implementation only applies `CompactionFilter` to files generated
+    // by compaction, as the name implies.
```
pdillinger, line 227: I would say or add "for backward compatibility"

ajkr: Done.

ajkr:
> What is the (expected) behavior of the flush and recover cases when the ColumnFamily has a CompactionFilterFactory but not MergeOperator? Can tests be added for those cases as well?

We would need to remove the `Merge()` calls as well (merging without a merge operator is prohibited now). The expected behavior is for the PUTs to be dropped in the same situations as in the existing tests. Idk, I feel this behavior is covered by the existing tests.

pdillinger (approved):
LGTM

PR #8246, opened 2021-04-29
Add more LSM info to FilterBuildingContext

pdillinger:
Summary: Add `num_levels`, `is_bottommost`, and table file creation
`reason` to `FilterBuildingContext`, in anticipation of more powerful
Bloom-like filter support.

To support this, added `is_bottommost` and `reason` to
`TableBuilderOptions`, which allowed removing `reason` parameter from
`rocksdb::BuildTable`.

I attempted to remove `skip_filters` from `TableBuilderOptions`, because
filter construction decisions should arise from options, not one-off
parameters. I could not completely remove it because the public API for
SstFileWriter takes a `skip_filters` parameter, and translating this
into an option change would mean awkwardly replacing the table_factory
if it is BlockBasedTableFactory with new filter_policy=nullptr option.
I marked this public skip_filters option as deprecated because of this
oddity. (skip_filters on the read side probably makes sense.)

At least `skip_filters` is now largely hidden for users of
`TableBuilderOptions` and is no longer used for implementing the
optimize_filters_for_hits option. Bringing the logic for that option
closer to handling of FilterBuildingContext makes it more obvious that
hese two are using the same notion of "bottommost." (Planned:
configuration options for Bloom-like filters that generalize
`optimize_filters_for_hits`)

Recommended follow-up: Try to get away from "bottommost level" naming of
things, which is inaccurate (see
VersionStorageInfo::RangeMightExistAfterSortedRun), and move to
"bottommost run" or just "bottommost."

Test Plan: extended an existing unit test to exercise and check various
filter building contexts. Also, existing tests for
optimize_filters_for_hits validate some of the "bottommost" handling,
which is now closely connected to FilterBuildingContext::is_bottommost
through TableBuilderOptions::is_bottommost


include/rocksdb/filter_policy.h
```diff
@@ -28,6 +28,7 @@
 #include <vector>
 
 #include "rocksdb/advanced_options.h"
+#include "rocksdb/listener.h"
```
mrambacher, line 31: I made this same request in #8243, but can we move the TableFileCreationReason out of listener into somewhere else (like types.h) so that you do not need listener (and all its baggage here).  The header files in RocksDB depend too much on one another and it makes the builds slower and makes it more confusing to find the classes you really want.

include/rocksdb/filter_policy.h
```diff
@@ -99,19 +100,32 @@ struct FilterBuildingContext {
...
   const BlockBasedTableOptions& table_options;
 
+  // BEGIN from (DB|ColumnFamily)Options in effect at table creation time
+  CompactionStyle compaction_style = kCompactionStyleLevel;
+
+  // Number of LSM levels, or -1 if unknown
+  int num_levels = -1;
+
+  // An optional logger for reporting errors, warnings, etc.
+  Logger* info_log = nullptr;
```
mrambacher, line 110: Please forward declare Logger.  I know it is not pbly not needed based on the complicated includes, but it would be nice to one day clean those up...

table/sst_file_writer.cc
```diff
@@ -253,10 +254,11 @@ Status SstFileWriter::Open(const std::string& file_path) {
...
       &int_tbl_prop_collector_factories, compression_type, compression_opts,
-      r->skip_filters, cf_id, r->column_family_name, unknown_level,
-      0 /* creation_time */, 0 /* oldest_key_time */,
-      0 /* file_creation_time */, "SST Writer" /* db_id */, db_session_id,
-      0 /* target_file_size */);
+      cf_id, r->column_family_name, unknown_level, false /* is_bottommost */,
+      TableFileCreationReason::kMisc, 0 /* creation_time */,
+      0 /* oldest_key_time */, 0 /* file_creation_time */,
+      "SST Writer" /* db_id */, db_session_id, 0 /* target_file_size */);
+  table_builder_options.skip_filters = r->skip_filters;
```
mrambacher, line 261: Please add a comment why you are setting skip_filters here even tho you hope to deprecate it (for historical/cleanup purposes later)/

table/block_based/block_based_table_builder.cc
```diff
@@ -473,15 +474,29 @@ struct BlockBasedTableBuilder::Rep {
           &this->internal_prefix_transform, use_delta_encoding_for_index_values,
           table_options));
     }
-    if (tbo.skip_filters) {
+    if (ioptions.optimize_filters_for_hits && tbo.is_bottommost) {
+      // Apply optimize_filters_for_hits setting here when applicable by
+      // skipping filter generation
+      filter_builder = nullptr;
+    } else if (tbo.skip_filters) {
```
mrambacher, line 481: Can we also skip the "CreateFilterBlockBuilder" here if the filter_policy is nullptr?  Otherwise we end up creating a FilterBuildingContext for no good reason...

include/rocksdb/filter_policy.h
pdillinger: Not too concerned about this. Google style recommends against forward declarations, though we sometimes (maybe) need them to separate user-public from internal-only APIs. https://google.github.io/styleguide/cppguide.html#Forward_Declarations

mrambacher (approved):
LGTM!

PR #8249, opened 2021-04-30
Added static methods for simple types to OptionTypeInfo

mrambacher:
Added ParseType, SerializeType, and TypesAreEqual methods to OptionTypeInfo.  These methods can be used for serialization and deserialization of basic types.

Change the MutableCF/DB Options to use this format.


options/cf_options.cc
```diff
@@ -1026,4 +1026,29 @@ void MutableCFOptions::Dump(Logger* log) const {
...
+  ConfigOptions config_options;
+  Status s = OptionTypeInfo::ParseType(
+      config_options, options_map, cf_mutable_options_type_info, new_options);
+  if (!s.ok()) {
+    *new_options = base_options;
+  }
+  return s;
+}
+
+Status GetStringFromMutableCFOptions(const ConfigOptions& config_options,
```
jay-zhuang, line 1045: The functions are moved from `options_helper.cc` to `cf_options.cc`, but the header file is still `options_helper.h`.

jay-zhuang (changes requested):

PR #8250, opened 2021-04-30
Support skipping filters in custom FilterPolicy, near-zero bpk

pdillinger:
Summary: I am in the process of building the next generation of APIs for
configuring filters. To get there, we need to increase the power of
custom filter policies, in the information they have available (#8246)
and in the actions they can take in generating filters (this change).

Here we give custom FilterPolicy the power to specify that generating
filters should be skipped in some contexts. Previously, a non-null
FilterPolicy would have to build some filter in all cases (with
FilterBitsBuilder), except when optimize_filters_for_hits applies, which
is out of control of the FilterPolicy. With this change, the
FilterPolicy has the same power to disable filters completely in some
contexts as optimize_filters_for_hits does. With #8246 and this change,
a FilterPolicy can implement the same behavior as
optimize_filters_for_hits, obviously with more freedom to tweak the
behavior.

The new API support for this is a bit ugly for API compatibility reasons
(see comment on FilterBitsBuilder::ShouldSkipFilter) but it should work
fine in practice.

Here we also change the behavior of passing < 0.5 bits_per_key to
NewBloomFilterPolicy (or NewExperimentalRibbonFilterPolicy), to skip
generating a filter instead of rounding up to 1 bit per key. The
motivation doesn't really have to do with the current API but imagines a
future API in which some math determines a "best" bits_per_key
for each level, perhaps based on a min and max range. In such a
scenario, to avoid filter lookup overhead (e.g. block cache CPU
overhead) in cases where it would be unproductive, we round bpk < 0.5
down to 0 (and SKIP generating a filter) and round 0.5 <= bpk < 1.0 up
to 1.0. (We can tweak these details later.) To make the current (soon to
be "old") API a simple instantiation of the new API, which simplifies
testing considerably, we preliminarily apply this same strategy to the
current API. Thus the new API will be a pure generalization.

Also included:

I noticed that the FilterBitsBuilder API docs do not mention re-use but
re-use is required for building partitioned filters. API docs updated to
call out that requirement.

Test Plan: For the behavior change on < 0.5 bits per key,
FullBloomTest::FilterSize is updated. New test
DBBloomFilterTestWithParam::SkipFilterOnEssentiallyZeroBpk uses this
setting and custom API to ensure that filter construction and query is
actually skipped, vs. using a trivial filter (which is tested for
comparison).


pdillinger:
> With #8246 and this change, a FilterPolicy can implement the same behavior as
> optimize_filters_for_hits, obviously with more freedom to tweak the behavior.

It is not so simple. `optimize_filters_for_hits` also operates on the table read side, which this cannot duplicate unless we start consulting FilterPolicy for whether to skip filters. That might be a good idea, except that logic is checked every time Get queries a table file. More customizability likely means more performance hit, unless we can make these determinations on a less critical path.

db/db_bloom_filter_test.cc
```diff
@@ -511,6 +511,116 @@ TEST_P(DBBloomFilterTestWithParam, BloomFilter) {
...
+  table_options.format_version = format_version_;
+
+  // Test 1: bits per key < 0.5 means skip filters -> no filter
+  // constructed or read.
+  table_options.filter_policy.reset(new BFP(0.4, bfp_impl_));
+  options.table_factory.reset(NewBlockBasedTableFactory(table_options));
+  DestroyAndReopen(options);
+  PutAndGetFn();
+
+  // Verify no filter access nor contruction
```
mrambacher, line 578: "Construction".

How is this validating not constructed?

db/db_bloom_filter_test.cc
```diff
@@ -511,6 +511,116 @@ TEST_P(DBBloomFilterTestWithParam, BloomFilter) {
...
+  EXPECT_EQ(TestGetTickerCount(options, BLOOM_FILTER_FULL_POSITIVE), 0);
+  EXPECT_EQ(TestGetTickerCount(options, BLOOM_FILTER_FULL_POSITIVE), 0);
```
mrambacher, line 580: Is there a reason for the same test 2x?

db/db_bloom_filter_test.cc
```diff
@@ -511,6 +511,116 @@ TEST_P(DBBloomFilterTestWithParam, BloomFilter) {
...
+  // or read.
+  table_options.filter_policy.reset(
+      new AlwaysTrueFilterPolicy(/* skip */ true));
+  options.table_factory.reset(NewBlockBasedTableFactory(table_options));
+  DestroyAndReopen(options);
+  PutAndGetFn();
+
+  // Verify no filter access nor construction
+  EXPECT_EQ(TestGetTickerCount(options, BLOOM_FILTER_FULL_POSITIVE), 0);
+  EXPECT_EQ(TestGetTickerCount(options, BLOOM_FILTER_FULL_POSITIVE), 0);
```
mrambacher, line 597: Same test twice?

table/block_based/block_based_table_builder.cc
```diff
@@ -89,12 +92,12 @@ FilterBlockBuilder* CreateFilterBlockBuilder(
...
       return new PartitionedFilterBlockBuilder(
           mopt.prefix_extractor.get(), table_opt.whole_key_filtering,
-          filter_bits_builder, table_opt.index_block_restart_interval,
+          filter_bits_builder.release(), table_opt.index_block_restart_interval,
           use_delta_encoding_for_index_values, p_index_builder, partition_size);
     } else {
       return new FullFilterBlockBuilder(mopt.prefix_extractor.get(),
                                         table_opt.whole_key_filtering,
-                                        filter_bits_builder);
+                                        filter_bits_builder.release());
```
mrambacher, line 100: Rather than the released pointer, maybe have a std::unique_ptr<>&& passed in and use std::move?  That is what is done in most places in the code.

table/block_based/filter_policy_internal.h
```diff
@@ -38,6 +39,31 @@ class BuiltinFilterBitsBuilder : public FilterBitsBuilder {
...
+ public:
+  size_t CalculateSpace(size_t) override { return 6U; }
+  double EstimatedFpRate(size_t, size_t) override { return 1.0; }
+  size_t ApproximateNumEntries(size_t) override { return SIZE_MAX; }
+  void AddKey(const Slice&) override {}
+  Slice Finish(std::unique_ptr<const char[]>* buf) override {
+    // Interpreted as "always true" filter (0 probes over 1 byte of
+    // payload, 5 bytes metadata)
+    buf->reset(new char[6]{});
+    return Slice(buf->get(), 6);
```
jay-zhuang, line 53: nit: define size 6.

table/block_based/filter_policy_internal.h
```diff
@@ -38,6 +39,31 @@ class BuiltinFilterBitsBuilder : public FilterBitsBuilder {
...
+  }
+};
+
+// See FilterBitsBuilder::ShouldSkipFilter
+class SkipFilterBitsBuilder : public AlwaysTrueBitsBuilder {
```
jay-zhuang, line 58: Should `SkipFilterBitsBuilder` based on `AlwaysTrueBitsBuilder` or it could inherit from `BuiltinFilterBitsBuilder` directly?

pdillinger:
Obsolete with 9501

PR #8253, opened 2021-05-04
Fix ConcurrentTaskLimiter token release for shutdown

ajkr:
Previously the shutdown process did not properly wait for all
`compaction_thread_limiter` tokens to be released before proceeding to
delete the DB's C++ objects. When this happened, we saw tests like
"DBCompactionTest.CompactionLimiter" flake with the following error:

```
virtual
rocksdb::ConcurrentTaskLimiterImpl::~ConcurrentTaskLimiterImpl():
Assertion `outstanding_tasks_ == 0' failed.
```

There is a case where a token can still be alive even after the shutdown
process has waited for BG work to complete. In particular, this happens
because the shutdown process only waits for flush/compaction scheduled/unscheduled counters to all
reach zero. These counters are decremented in `BackgroundCallCompaction()`
functions. However, tokens are released in `BGWork*Compaction()` functions, which
actually wrap the `BackgroundCallCompaction()` function.

A simple sleep could repro the race condition:

```
$ diff --git a/db/db_impl/db_impl_compaction_flush.cc
b/db/db_impl/db_impl_compaction_flush.cc
index 806bc548a..ba59efa89 100644
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -2442,6 +2442,7 @@ void DBImpl::BGWorkCompaction(void* arg) {
       static_cast<PrepickedCompaction*>(ca.prepicked_compaction);
   static_cast_with_check<DBImpl>(ca.db)->BackgroundCallCompaction(
       prepicked_compaction, Env::Priority::LOW);
+  sleep(1);
   delete prepicked_compaction;
 }

$ ./db_compaction_test --gtest_filter=DBCompactionTest.CompactionLimiter
db_compaction_test: util/concurrent_task_limiter_impl.cc:24: virtual rocksdb::ConcurrentTaskLimiterImpl::~ConcurrentTaskLimiterImpl(): Assertion `outstanding_tasks_ == 0' failed.
Received signal 6 (Aborted)
#0   /usr/local/fbcode/platform007/lib/libc.so.6(gsignal+0xcf) [0x7f02673c30ff] ??      ??:0
#1   /usr/local/fbcode/platform007/lib/libc.so.6(abort+0x134) [0x7f02673ac934] ??       ??:0
...
```

Test Plan: sleeps to expose race conditions

### DIFF
--- a/util/concurrent_task_limiter_impl.cc
+++ b/util/concurrent_task_limiter_impl.cc
@@ -67,8 +67,6 @@ void TaskLimiterToken::ReleaseOnce() {
   assert(limiter_->outstanding_tasks_ >= 0);
 }
 
-TaskLimiterToken::~TaskLimiterToken() {
-  ReleaseOnce();
-}
+TaskLimiterToken::~TaskLimiterToken() { ReleaseOnce(); }
 
 }  // namespace ROCKSDB_NAMESPACE
--- a/util/concurrent_task_limiter_impl.cc
+++ b/util/concurrent_task_limiter_impl.cc
@@ -59,9 +59,16 @@ ConcurrentTaskLimiter* NewConcurrentTaskLimiter(
   return new ConcurrentTaskLimiterImpl(name, limit);
 }
 
-TaskLimiterToken::~TaskLimiterToken() {
-  --limiter_->outstanding_tasks_;
+void TaskLimiterToken::ReleaseOnce() {
+  if (!released_) {
+    --limiter_->outstanding_tasks_;
+    released_ = true;
+  }
   assert(limiter_->outstanding_tasks_ >= 0);
+}
+
+TaskLimiterToken::~TaskLimiterToken() {
+  ReleaseOnce();
 }
 
 }  // namespace ROCKSDB_NAMESPACE
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -2759,6 +2759,14 @@ void DBImpl::BackgroundCallCompaction(PrepickedCompaction* prepicked_compaction,
 
     // See if there's more work to be done
     MaybeScheduleFlushOrCompaction();
+
+    if (prepicked_compaction != nullptr &&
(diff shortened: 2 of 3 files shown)


util/concurrent_task_limiter_impl.cc
```diff
@@ -59,9 +59,14 @@ ConcurrentTaskLimiter* NewConcurrentTaskLimiter(
...
 }
 
-TaskLimiterToken::~TaskLimiterToken() {
-  --limiter_->outstanding_tasks_;
+void TaskLimiterToken::ReleaseOnce() {
+  if (!released_) {
+    --limiter_->outstanding_tasks_;
+    released_ = true;
+  }
   assert(limiter_->outstanding_tasks_ >= 0);
```
mrambacher, line 67: Why not also move the assert into ReleaseOnce()

mrambacher (commented):
Should other places check/assert that a token is not used after it is released?

Can a SyncPoint or something be added here to validate the fix going forward?  The test plan introduced temporary code that is not part of this checkin to reproduce the problem, but I do not see how we know the problem does not come back...

ajkr (commented):
> Should other places check/assert that a token is not used after it is released?

I think we don't check that yet today. The worker doesn't use the token for anything while doing its work.

 > Can a SyncPoint or something be added here to validate the fix going forward? The test plan introduced temporary code that is not part of this checkin to reproduce the problem, but I do not see how we know the problem does not come back...

I have some idea to sync DB destruction immediately following the first DB mutex release after the bg work counters becoming zero. Haven't tried it yet.

akankshamahajan15 (approved):
LGTM! Thanks for the fix.

anand1976 (approved):
LGTM

ajkr:
> I have some idea to sync DB destruction immediately following the first DB mutex release after the bg work counters becoming zero. Haven't tried it yet.

Will keep this in mind for later to move on to other things.

Thanks for the reviews!

PR #8254, opened 2021-05-04
Upgrade MacOS xcode 9.4.1->10.1.0

ajkr:
Lately homebrew steps started failing with:

```
Warning: You are using macOS 10.13.
We (and Apple) do not provide support for this old version.
...
Warning: Your Xcode (9.4.1) is outdated.
Please update to Xcode 10.1 (or delete it).
```

Trying the minimum of the suggested upgrades (xcode 10.1.0) to see if it solves it. That would still leave us on macOS 10.13.6 according to this table - https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,7 +120,7 @@ executors:
 jobs:
   build-macos:
     macos:
-      xcode: 9.4.1
+      xcode: 10.1.0
     steps:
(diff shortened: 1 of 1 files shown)


ajkr:
Didn't work. Try #8255 instead.

PR #8256, opened 2021-05-04
Update CircleCI MacOS Xcode version to 11.3.0

jay-zhuang:
To fix CircleCI pyenv installation failure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -463,7 +463,7 @@ jobs:
       xcode: 11.3.0
     resource_class: medium
     environment:
-      JAVA_HOME: /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home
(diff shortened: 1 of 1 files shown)


ajkr (approved):
LGTM, I prefer this solution over #8255.

jay-zhuang:
> LGTM, I prefer this solution over #8255.

If needed, we can setup a nightly build with older xcode. cc. @adamretter

ajkr:
I liked picking an xcode 11.3 version because it works on macOS 10.14 (according to https://xcodereleases.com/). And according to wikipedia, macOS 10.14 is still supported by Apple and can't be upgraded to 10.15 on all hardware.

jay-zhuang:
> I liked picking an xcode 11.3 version because it works on macOS 10.14 (according to [xcodereleases.com](https://xcodereleases.com/)). And according to wikipedia, macOS 10.14 is still supported by Apple and can't be upgraded to 10.15 on all hardware.

Sounds good, another reason is: 12.4 image only has jdk 11, instead of 8 which I think we need to support.

PR #8258, opened 2021-05-04
Deflake TransactionStressTest.ExpiredTransactionDataRace1

ajkr:
We saw the `Commit()` fail with "Operation expired" so apparently the
expiration time is too short. Increased the magnitude of the times in
this test to make flakiness less likely.


jay-zhuang (approved):
LGTM

PR #8259, opened 2021-05-04
Deflake DBTest.L0L1L2AndUpHitCounter

ajkr:
Previously we saw flakes on platforms like arm on CircleCI, such as the following:

```
Note: Google Test filter = DBTest.L0L1L2AndUpHitCounter
[----------] 1 test from DBTest
[ RUN      ] DBTest.L0L1L2AndUpHitCounter
db/db_test.cc:5345: Failure
Expected: (TestGetTickerCount(options, GET_HIT_L0)) > (100), actual: 30 vs 100
[  FAILED  ] DBTest.L0L1L2AndUpHitCounter (150 ms)
[----------] 1 test from DBTest (150 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (150 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] DBTest.L0L1L2AndUpHitCounter
```

The test was totally non-deterministic, e.g., flush/compaction timing would affect how many files on each level. Furthermore, it depended heavily on platform-specific details, e.g., by having a 32KB memtable, it could become full with a very different number of entries depending on the platform. 

This PR rewrites the test to build a deterministic LSM with one file per level.

### DIFF
--- a/db/db_test.cc
+++ b/db/db_test.cc
@@ -5314,24 +5314,29 @@ TEST_F(DBTest, DynamicMiscOptions) {
 #endif  // ROCKSDB_LITE
 
 TEST_F(DBTest, L0L1L2AndUpHitCounter) {
-  const int kNumKeysPerDb = 30000;
   const int kNumLevels = 3;
-  const int kNumKeysPerLevel = kNumKeysPerDb / kNumLevels;
+  const int kNumKeysPerLevel = 10000;
+  const int kNumKeysPerDb = kNumLevels * kNumKeysPerLevel;
 
   Options options = CurrentOptions();
   options.statistics = ROCKSDB_NAMESPACE::CreateDBStatistics();
   Reopen(options);
 
-  // Place one file on each of L0, L1, and L2.
-  for (int i = 0; i < kNumLevels; ++i) {
-    for (int j = 0; j < kNumKeysPerLevel; ++j) {
-      ASSERT_OK(Put(Key(i * kNumKeysPerLevel + j), "val"));
+  // After the below loop there will be one file on each of L0, L1, and L2.
+  int key = 0;
+  for (int output_level = kNumLevels - 1; output_level >= 0; --output_level) {
+    for (int i = 0; i < kNumKeysPerLevel; ++i) {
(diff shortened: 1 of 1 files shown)


db/db_test.cc
```diff
@@ -5314,41 +5314,40 @@ TEST_F(DBTest, DynamicMiscOptions) {
 #endif  // ROCKSDB_LITE
 
 TEST_F(DBTest, L0L1L2AndUpHitCounter) {
+  const int kNumKeysPerDb = 30000;
+  const int kNumLevels = 3;
+  const int kNumKeysPerLevel = kNumKeysPerDb / kNumLevels;
```
mrambacher, line 5319: Nit: It might be better to defined kNumKeysPerDb in terms of kNumKeysPerLevel/Levels.  The division requires integer math, whereas multiplication would not.

db/db_test.cc
```diff
@@ -5314,41 +5314,40 @@ TEST_F(DBTest, DynamicMiscOptions) {
...
-  options.max_background_compactions = 8;
-  options.max_background_flushes = 8;
   options.statistics = ROCKSDB_NAMESPACE::CreateDBStatistics();
-  CreateAndReopenWithCF({"mypikachu"}, options);
+  Reopen(options);
 
-  int numkeys = 20000;
-  for (int i = 0; i < numkeys; i++) {
-    ASSERT_OK(Put(1, Key(i), "val"));
+  // Place one file on each of L0, L1, and L2.
```
mrambacher, line 5325: How are you guaranteed this is "one file" on each level?  Or are you suggesting there will be one file after the flush/compact?

mrambacher (approved):
LGTM.

db/db_test.cc
ajkr: Done.

db/db_test.cc
ajkr: Yes after the flush/compact below. I rewrote the loops in a clearer way.

ajkr (commented):
Thanks for the review!

Issue #4812, opened 2018-12-21
Support DeleteRange in transactions


abhimadan:
Now that DeleteRange is performant, it would be nice to support it in transactions as well.

maysamyabandeh:
Since DeleteRange and WritePrepared Transactions are developed in parallel we need to make sure that they are consistent with each other. One difference to watch out for is that in WritePrepared the prepare and commit sequence numbers are separate: the prepare sequence number is stored in DB with the key/value and commit is stored outside.
I went through the code to figure the potential places that needs careful consideration:
1. There is a TODO here for duplicate key detection: https://github.com/facebook/rocksdb/blob/52e6404e0f72b6f994a442084415f518f9ec9f6c/db/write_batch.cc#L1435
to include the end key but I think it is fine since the end key is encoded in the value, thus not causing duplicate key issue.
2. In Memtable::Get there is a comparison of seq to max_covering_tombstone_seq which is obtained from DeleteRange tombstone.
https://github.com/facebook/rocksdb/blob/52e6404e0f72b6f994a442084415f518f9ec9f6c/db/memtable.cc#L648
I read the code a bit but could not figure what is the significance of this sequence number and whether it is consistent with separation of prepare and commit sequence number introduced in WritePrepared transactions.
Same goes for GetContext: https://github.com/facebook/rocksdb/blob/52e6404e0f72b6f994a442084415f518f9ec9f6c/table/get_context.cc#L205
3. WritePrepared tests assumed the key cannot be RangeDelete. The tests must be updated to add RangeDelete type anywhere Delete and SingleDelete is used: https://github.com/facebook/rocksdb/blob/master/utilities/transactions/write_prepared_transaction_test.cc#L513
4. ::ShouldDelete that checks against the deleted ranges, should only care about user key and ignore the sequence number. However it uses internal key comparator instead of user comparator: https://github.com/facebook/rocksdb/blob/3c5eed5ebee8eb06425e8716e1a56fb9782ce9a9/db/range_del_aggregator.cc#L210
We need to clarify why so that the sequence number, that is interpreted differently in write prepared transactions, is not used meaningfully here.
5. Finally the existing tests for range delete should be extended to include transactions (extended with ::DeleteRange) API as well.
We can discuss these items here, see which one could be potential issue, and specify what exactly needs to be done to extend transactions with DeleteRange.

@ajkr  What do you think?

ajkr:
[...]

PR #8260, opened 2021-05-04
Wbwi deleterange alan001

alanpaxton:
Add DeleteRange functionality to write_batch_with_index. Functionality is restricted to the overwrite_key=true mode on WBWI.
Deleted ranges for the WBWI are maintained in a deleted range map, which is based on the existing skiplist implementation.
Limited removal functionality is added to the skiplist implementation, in order to allow removal of deleted entries from the existing index when a range deletion is performed, and to allow updates to the deleted range map. Removed entries are not de-allocated until the skiplist is de-allocated.


ryoqun:
> It is great to have this ready as it is a necessary step towards transaction support for DeleteRange some day. However the demand for it inside FB is not there yet. I am thankful this code is ready so we can take it when we have a foreseeable customer inside FB, but that means it may end up sitting a while longer. Sorry we did not discuss this point at the outset.

@ajkr so, it looks like this no delete range with transaction is like a current implementation limitation rather then design choice? is there any update on this priority? context: https://github.com/rust-rocksdb/rust-rocksdb/pull/565

ajkr:
> > It is great to have this ready as it is a necessary step towards transaction support for DeleteRange some day. However the demand for it inside FB is not there yet. I am thankful this code is ready so we can take it when we have a foreseeable customer inside FB, but that means it may end up sitting a while longer. Sorry we did not discuss this point at the outset.
> 
> @ajkr so, it looks like this no delete range with transaction is like a current implementation limitation rather then design choice? is there any update on this priority? context: https://github.com/rust-rocksdb/rust-rocksdb/pull/565/files#r758859523

Right. Beyond this PR there are more implementation limitations for transactional DeleteRange support, like we need to lock a range for writes. Transactional DeleteRange is not on next year's roadmap :/.

adamretter:
I think the most important question to ask at this immediate time is... does this make this product better?

Of course there are Architectural concerns about how it fits with our long goals. However that is so undecided and fragrant, we could just hold off forever!

IMHO opinion, this is a great addition, which does not stop us from re-enginering things later. Or... Am I mistaken?

If it's not a breaking change... let's accept it.. no?

Zelldon:
@ajkr @adamretter would be great to get this merged or reawoken, we would like to use DeleteRange on Transaction, and realized that it is not supported (see related https://github.com/facebook/rocksdb/issues/4812). We thought about migrating to WriteBatchWithIndex but realized that this also doesn't support DeleteRange either. Would be really great to have this at least on the WriteBatch :) 

See related discussion https://groups.google.com/g/rocksdb/c/psEhYNnFnB8/m/nBL4pJ_4GgAJ

PR #8261, opened 2021-05-04
Fix use-after-free threading bug in ClockCache

pdillinger:
Summary: In testing for #8225 I found cache_bench would crash with
-use_clock_cache, as well as db_bench -use_clock_cache, but not
single-threaded. Smaller cache size hits failure much faster. ASAN
reported the failuer as calling malloc_usable_size on the `key` pointer
of a ClockCache handle after it was reportedly freed. On detailed
inspection I found this bad sequence of operations for a cache entry:

state=InCache=1,refs=1
[thread 1] Start ClockCacheShard::Unref (from Release, no mutex)
[thread 1] Decrement ref count
state=InCache=1,refs=0
[thread 1] Suspend before CalcTotalCharge (no mutex)

[thread 2] Start UnsetInCache (from Insert, mutex held)
[thread 2] clear InCache bit
state=InCache=0,refs=0
[thread 2] Calls RecycleHandle (based on pre-updated state)
[thread 2] Returns to Insert which calls Cleanup which deletes `key`

[thread 1] Resume ClockCacheShard::Unref
[thread 1] Read `key` in CalcTotalCharge

To fix this, I've added a field to the handle to store the metadata
charge so that we can efficiently remember everything we need from
the handle in Unref. We must not read from the handle again if we
decrement the count to zero with InCache=1, which means we don't own
the entry and someone else could eject/overwrite it immediately.

Note before this change, on amd64 sizeof(Handle) == 56 even though there
are only 48 bytes of data. Grouping together the uint32_t fields would
cut it down to 48, but I've added another uint32_t, which takes it
back up to 56. Not a big deal.

Also fixed DisownData to cooperate with ASAN as in LRUCache.

Test Plan: Manual + adding use_clock_cache to db_crashtest.py

Base performance
./cache_bench -use_clock_cache
Complete in 17.060 s; QPS = 2458513
New performance
./cache_bench -use_clock_cache
Complete in 17.052 s; QPS = 2459695

Any difference is easily buried in small noise.

Crash test shows still more bug(s) in ClockCache, so I'm expecting to
disable ClockCache from production code in a follow-up PR (if we
can't find and fix the bug(s))

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -19,6 +19,7 @@
 ### Public API change
 * Removed a parameter from TableFactory::NewTableBuilder, which should not be called by user code because TableBuilder is not a public API.
 * The `skip_filters` parameter to SstFileWriter is now considered deprecated. Use `BlockBasedTableOptions::filter_policy` to control generation of filters.
+* ClockCache is known to have bugs that could lead to crash or corruption, so should not be used until fixed. Use NewLRUCache instead.
 
 ## 6.20.0 (04/16/2021)
 ### Behavior Changes
--- a/cache/clock_cache.cc
+++ b/cache/clock_cache.cc
@@ -600,6 +600,15 @@ CacheHandle* ClockCacheShard::Insert(
   handle->meta_charge = meta_charge;
   handle->deleter = deleter;
   uint32_t flags = hold_reference ? kInCacheBit + kOneRef : kInCacheBit;
+
+  // TODO investigate+fix suspected race condition:
+  // [thread 1] Lookup starts, up to Ref()
+  // [thread 2] Erase/evict the entry just looked up
+  // [thread 1] Ref() the handle, even though it's in the recycle bin
+  // [thread 2] Insert with recycling that handle
+  // Here we obliterate the other thread's Ref
+  // Possible fix: never blindly overwrite the flags, but only make
+  // relative updates (fetch_add, etc).
   handle->flags.store(flags, std::memory_order_relaxed);
   HashTable::accessor accessor;
   if (table_.find(accessor, CacheKey(key, hash))) {
--- a/cache/clock_cache.cc
+++ b/cache/clock_cache.cc
@@ -209,8 +209,7 @@ struct CacheHandle {
   }
 
   inline static uint32_t CalcMetadataCharge(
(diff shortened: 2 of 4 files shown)


pdillinger:
Unfortunately, there is at least one more bug in ClockCache that shows up in the crash test:

    ***db_stress: cache/clock_cache.cc:486: bool rocksdb::{anonymous}::ClockCacheShard::Unref(rocksdb::{anonymous}::CacheHandle*, bool, rocksdb::{anonymous}::CleanupContext*): Assertion `CountRefs(flags) > 0' failed.***

I tried a few possible fixes, but none worked. At least cache_bench seems to work well enough to unblock me on #8225.

I think I will update this PR to disable the ClockCache as a feature, keeping it for testing only when/if someone fixes it.

pdillinger:
I suspect this race:

[thread 1] Lookup starts, up to Ref()
[thread 2] Erase/evict the entry just looked up
[thread 1] Ref() the handle, even though it's in the recycle bin
[thread 2] Insert with recycling that handle, handle->flags.store obliterates the other thread's Ref

Which would be fixable by never blindly overwriting flags, only making relative updates. But I am not going to invest more into this to try to reproduce to validate the fix.

pdillinger:
Actually, I want to do the disabling ClockCache in another PR for cleaner git history. I'm not adding to HISTORY.md for this one because I'm expecting this bug fix to be irrelevant for production users (in the near term at least).

mrambacher (approved):
LGTM.

Do you want to put a TODO comment about the race condition in the code or enter that as a new issue to track elsewhere?

mrambacher (approved):
LGTM.

pdillinger:
I added some notes about ClockCache being broken in case we forget to follow-up with any sort of functionality disabling.

ZacAttack:
It seems this bug is still referenced in the documentation: https://github.com/facebook/rocksdb/wiki/Block-Cache#clock-cache

> WARNING: The ClockCache implementation has at least one remaining bug that could lead to crash or data corruption. Please do not use ClockCache until this is fixed.

Does the above statement still stand in latest release, or should the documentation be updated?  Hoping to use clock cache in production.

PR #8262, opened 2021-05-05
Make ImmutableOptions struct that inherits from ImmutableCFOptions and ImmutableDBOptions

mrambacher:
The ImmutableCFOptions contained a bunch of fields that belonged to the ImmutableDBOptions.  This change cleans that up by introducing an ImmutableOptions struct.  Following the pattern of Options struct, this class inherits from the DB and CFOption structs (of the Immutable form).

Only one structural change (the ImmutableCFOptions::fs was changed to a shared_ptr from a raw one) is in this PR.  All of the other changes involve moving the member variables from the ImmutableCFOptions into the ImmutableOptions and changing member variables or function parameters as required for compilation purposes.

Follow-on PRs may do a further clean-up of the code, such as renaming variables (such as "ImmutableOptions cf_options") and potentially eliminating un-needed function parameters (there is no longer a need to pass both an ImmutableDBOptions and an ImmutableOptions to a function).


options/cf_options.h
```diff
@@ -54,86 +51,49 @@ struct ImmutableCFOptions {
...
+struct ImmutableOptions : public ImmutableDBOptions, public ImmutableCFOptions {
+  explicit ImmutableOptions();
+  explicit ImmutableOptions(const Options& options);
 
-  std::string db_host_id;  // ImmutableDBOptions
+  ImmutableOptions(const DBOptions& db_options,
+                   const ColumnFamilyOptions& cf_options);
 
-  FileTypeSet checksum_handoff_file_types;  // ImmutableDBOptions
+  //**TODO: Should be ImmutableCFOptions...
```
pdillinger, line 94: Not sure what exactly you mean, but OK

pdillinger (approved):
A solid step toward better code :-D

pdillinger:
Looks like fix needed for ColumnFamilyOptionsAllFieldsSettable - OptionsSettableTest (CircleCI)

PR #8269, opened 2021-05-05
Fix manual compaction `max_compaction_bytes` under-calculated issue

jay-zhuang:
Fix a bug that for manual compaction, `max_compaction_bytes` is only
limit the SST files from input level, but not overlapped files on output
level.

Test Plan: `make check`


ajkr (approved):
Sorry I forgot. Did you still want to try landing this before the branch cut?

PR #8268, opened 2021-05-05
Fix `GetLiveFiles()` returning OPTIONS-000000

ajkr:
See release note in HISTORY.md.

Test Plan: unit test repro

PR #8272, opened 2021-05-05
Permit stdout "fail"/"error" in whitebox crash test

ajkr:
In #8268, the `db_stress` stdout began containing both the strings
"fail" and "error" (case-insensitive). The whitebox crash test
failed upon seeing either of those strings.

I checked that all other occurrences of "fail" and "error"
(case-insensitive) that `db_stress` produces are printed to `stderr`. So
this PR separates the handling of `db_stress`'s stdout and stderr, and
only fails when one those bad strings are found in stderr.

The downside of this PR is `db_stress`'s original interleaving of stdout/stderr is not preserved in `db_crashtest.py`'s output.

Test Plan: run it; see it succeeds for several runs until encountering a real error

```
$ python3 tools/db_crashtest.py whitebox --simple --random_kill_odd=8887 --max_key=1000000 --value_size_mult=33
...
db_stress: cache/clock_cache.cc:483: bool rocksdb::{anonymous}::ClockCacheShard::Unref(rocksdb::{anonymous}::CacheHandle*, bool, rocksdb::{anonymous}::CleanupContext*): Assertion `CountRefs(flags) > 0' failed.

TEST FAILED. Output has 'fail'!!!
```

### DIFF
--- a/tools/db_crashtest.py
+++ b/tools/db_crashtest.py
@@ -615,7 +615,7 @@ def whitebox_crash_main(args, unknown_args):
         print("Running:" + ' '.join(cmd) + "\n")  # noqa: E999 T25377293 Grandfathered in
 
         popen = subprocess.Popen(cmd, stdout=subprocess.PIPE,
-                                 stderr=subprocess.STDOUT)
+                                 stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


zhichao-cao (approved):
LGTM！

PR #8274, opened 2021-05-06
Deprecate obsolete "backupable db" from public APIs

pdillinger:
Summary: An early design of BackupEngine used stackable DB, so I guess a
DB had to opt-in to being backupable. Unfortunately the naming of that
obsolete design still infects our public API and implementation.

This change fixes the public API, with a deprecated
backward-compatibility header. `BackupableDBOptions` is renamed to
`BackupEngineOptions` (copy-replace in the public header) and
backup_engine.h replaces backupable_db.h (present for backward
compatibility). The only other change in backupable_db.h ->
backup_engine.h is cleaning up headers.

Later changes will fix the internal implementation.

Test Plan: The internal implementation of BackupEngine uses the name
BackupEngineOptions, while the unit tests use the old name
BackupableDBOptions. This gives me confidence that both still work.


pdillinger:
The super annoying thing about this change is that it seems I have to break `git blame include/rocksdb/utilities/backup_engine.h`. Rename detection doesn't seem to work because I'm not deleting the source file but replacing it will much smaller contents. Because of Facebook's github PR infrastructure, I can't break this into two commits without it being two PRs. I suppose I could create one PR where I move backupable_db.h, update all the references to it, and before the next release re-instate it for backward compatibility, but that's weird.

pdillinger:
> The super annoying thing about this change is ...

I'm going to "let it go" ;)

mrambacher (approved):
Shouldn't the include files (at least backupable_db_impl.h) be changed to reflect the obsolescence of backupable_db.h?  If you want to keep the test including the old one, that might make sense but not sure on the functional code itself...

pdillinger:
> Shouldn't the include files (at least backupable_db_impl.h) be changed to reflect the obsolescence of backupable_db.h? If you want to keep the test including the old one, that might make sense but not sure on the functional code itself...

I can fix that in next PR.

PR #8275, opened 2021-05-06
flip default `DBOptions::fail_if_options_file_error`

ajkr:
Changed `DBOptions::fail_if_options_file_error` default from `false` to
`true`. It is safer to fail an operation by default when it encounters
an error.

Also changed the API doc of that option to not mention failure to detect
an OPTIONS file. I found no usage of it related to detecting OPTIONS
files.

Test Plan: rely on CI


ajkr:
Changed my mind after learning OPTIONS file writing fails in scenarios like custom TableFactory that's not registered.

ajkr:
Need to recreate targeting main branch

PR #8289, opened 2021-05-12
Fix cmake failed to build db_bench

jay-zhuang:
And change the cmake build on macos with GFLAGS on to cover more cases.


mrambacher (approved):
LGTM.

PR #8292, opened 2021-05-13
Support for Merge in Integrated BlobDB with base values

akankshamahajan15:
Summary: This PR add support for Merge operation in Integrated BlobDB with base values(i.e DB::Put). Merged values can be retrieved through  DB::Get, DB::MultiGet and Iterator operation.

Test Plan: Add new unit tests

Reviewers:

Subscribers:

Tasks:

Tags:


db/db_iter.cc
```diff
@@ -1095,25 +1120,32 @@ bool DBIter::FindValueForCurrentKeyUsingSeek() {
...
+        status_ =
+            Status::NotSupported("BlobDB does not support merge operator.");
+        valid_ = false;
+        return false;
+      }
+      if (!SetBlobValueIfNeeded(ikey.user_key, iter_.value())) {
+        return false;
+      }
+      const Slice blob_value = value();
+      Status s = Merge(&blob_value, ikey.user_key);
```
ltamasi, line 1143: Should use `saved_key_.GetUserKey()` here I believe. Also, I think we don't have a unit test that covers `FindValueForCurrentKeyUsingSeek`; would be nice to add one.

db/version_set.cc
```diff
@@ -1875,13 +1875,25 @@ void Version::Get(const ReadOptions& read_options, const LookupKey& k,
...
+                                    Slice& blob_value) -> Status {
+    Status s;
+    if (is_blob_index) {
+      constexpr uint64_t* bytes_read = nullptr;
+      s = GetBlob(read_options, user_key, blob_index, value, bytes_read);
+      blob_value = static_cast<Slice>(*value);
+    }
+    return s;
+  };
+
```
ltamasi, line 1889: Love the callback idea but instead of using `std::function` and a lambda, can we use a concrete class? (Rationale: `std::function` does type erasure, which has similar runtime overhead as virtuality.)

table/get_context.cc
```diff
@@ -273,19 +280,23 @@ bool GetContext::SaveValue(const ParsedInternalKey& parsed_key,
...
               }
+              state_ = kCorrupt;
+              return false;
             }
+            Merge(&blob_value);
+            *is_blob_index_ = false;
           } else {
+            Merge(&value);
+          }
+          if (do_merge_ == false) {
```
ltamasi, line 299: If we're doing `GetMergeOperands` and it's a blob index, we would want to save the blob we retrieved as an operand, right?

ltamasi (commented):
Thanks for the PR @akankshamahajan15! This is awesome

db/db_iter.cc
akankshamahajan15: Thanks for pointing it out. I missed changing  it to saved_key_.GetUserKey().

table/get_context.cc
akankshamahajan15: Oh, I forgot about ```DB::GetMergeOperands``` which covers ```do_merge = false```. I will add the test and fix the code.

db/version_set.cc
```diff
@@ -1875,14 +1876,16 @@ void Version::Get(const ReadOptions& read_options, const LookupKey& k,
   // need to provide it here.
   bool is_blob_index = false;
   bool* const is_blob_to_use = is_blob ? is_blob : &is_blob_index;
+  std::unique_ptr<BlobFileMergeCallback> blob_file_merge_callback(
```
ltamasi, line 1879: You could simply allocate this on the stack. No need to use `unique_ptr`; you can simply pass it to `GetContext` via a raw (non-owning) pointer.

db/version_set.cc
```diff
@@ -2031,15 +2034,19 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
   // use autovector in order to avoid unnecessary construction of GetContext
   // objects, which is expensive
   autovector<GetContext, 16> get_ctx;
+
   for (auto iter = range->begin(); iter != range->end(); ++iter) {
+    std::unique_ptr<BlobFileMergeCallback> blob_file_merge_callback(
```
ltamasi, line 2039: Same here. Also, I think one object would be sufficient (which you could create outside the loop).

table/get_context.cc
```diff
@@ -269,27 +275,40 @@ bool GetContext::SaveValue(const ParsedInternalKey& parsed_key,
...
+              return false;
+            }
+            Slice blob_value(pin_val);
+            Merge(&blob_value);
+            if (do_merge_ == false) {
+              // It means this function is called as part of DB GetMergeOperands
+              // API and the current value should be part of
+              // merge_context_->operand_list
+              push_operand(blob_value, nullptr);
             }
```
ltamasi, line 303: I think this would be easier to understand if we kept `state_ = kFound;` where it is (instead of putting it in the `Merge` helper method) and used the following pattern (same below):

```
if (do_merge_) {
   Merge(...);
} else {
   push_operand(...);
}
```

ltamasi (commented):
Thanks again @akankshamahajan15 ! The patch looks good to me, just some minor comments

table/get_context.h
```diff
@@ -106,8 +106,7 @@ class GetContext {
              PinnedIteratorsManager* _pinned_iters_mgr = nullptr,
              ReadCallback* callback = nullptr, bool* is_blob_index = nullptr,
              uint64_t tracing_get_id = 0,
-             std::unique_ptr<BlobFileMergeCallback>&& blob_file_merge_callback =
-                 nullptr);
+             BlobFetcher* blob_file_merge_callback = nullptr);
```
ltamasi, line 109: Could you also rename the constructor parameter?

ltamasi (approved):

PR #8295, opened 2021-05-13
Refactor Option obj address from char* to void*

jay-zhuang:
And replace `reinterpret_cast` with `static_cast` or no cast.

Test Plan: `make check`


mrambacher (approved):
This looks great!  Thanks for doing this.

As I was reviewing the changes, I realized why I had left it as a char* originally (rather than making the changes as you have here).  Originally the offset_ was a public member of the OptionTypeInfo and the code to calculate the new offset (addr + offset_) was scattered throughout the code base.  Rather than having to cast addr everywhere back to a char* for pointer math, I left it alone.

This is a much cleaner solution!

PR #8298, opened 2021-05-14
Make it possible to apply only a subrange of table property collectors

ltamasi:
Summary:
This patch does two things:
1) Introduces some aliases in order to eliminate/prevent long-winded type names
w/r/t the internal table property collectors (see e.g.
`std::vector<std::unique_ptr<IntTblPropCollectorFactory>>`).
2) Makes it possible to apply only a subrange of table property collectors during
table building by turning `TableBuilderOptions::int_tbl_prop_collector_factories`
from a pointer to a `vector` into a range (i.e. a pair of iterators).

Rationale: I plan to introduce a BlobDB related table property collector, which
should only be applied during table creation if blob storage is enabled at the moment
(which can be changed dynamically). This change will make it possible to include/
exclude the BlobDB related collector as needed without having to introduce
a second `vector` of collectors in `ColumnFamilyData` with pretty much the same
contents.

Test Plan:
`make check`

### DIFF
--- a/db/table_properties_collector.h
+++ b/db/table_properties_collector.h
@@ -48,6 +48,13 @@ class IntTblPropCollectorFactory {
   virtual const char* Name() const = 0;
 };
 
+using IntTblPropCollectorFactories =
+    std::vector<std::unique_ptr<IntTblPropCollectorFactory>>;
+using IntTblPropCollectorFactoryIter =
+    IntTblPropCollectorFactories::const_iterator;
+using IntTblPropCollectorFactoryRange =
+    std::pair<IntTblPropCollectorFactoryIter, IntTblPropCollectorFactoryIter>;
+
 // When rocksdb creates a new table, it will encode all "user keys" into
 // "internal keys", which contains meta information of a given entry.
 //
--- a/table/block_based/block_based_table_builder.cc
+++ b/table/block_based/block_based_table_builder.cc
@@ -505,8 +505,10 @@ struct BlockBasedTableBuilder::Rep {
           use_delta_encoding_for_index_values, p_index_builder_));
     }
 
-    for (auto it = tbo.int_tbl_prop_collector_factories.first;
-         it != tbo.int_tbl_prop_collector_factories.second; ++it) {
+    const auto& factory_range = tbo.int_tbl_prop_collector_factories;
+    for (auto it = factory_range.first; it != factory_range.second; ++it) {
+      assert(*it);
+
       table_properties_collectors.emplace_back(
           (*it)->CreateIntTblPropCollector(column_family_id));
     }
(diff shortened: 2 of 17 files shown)


table/block_based/block_based_table_builder.cc
```diff
@@ -505,9 +505,12 @@ struct BlockBasedTableBuilder::Rep {
...
     }
 
-    for (auto& collector_factories : *tbo.int_tbl_prop_collector_factories) {
+    const auto& factory_range = tbo.int_tbl_prop_collector_factories;
+    for (auto it = factory_range.first; it != factory_range.second; ++it) {
+      assert(*it);
+
       table_properties_collectors.emplace_back(
-          collector_factories->CreateIntTblPropCollector(column_family_id));
+          (*it)->CreateIntTblPropCollector(column_family_id));
```
jay-zhuang, line 513: Would it be overdesign to have a factory_range class that has an `begin/end`? which basically has the factories and the range in the same place.

table/block_based/block_based_table_builder.cc
```diff
@@ -505,9 +505,12 @@ struct BlockBasedTableBuilder::Rep {
           use_delta_encoding_for_index_values, p_index_builder_));
     }
 
-    for (auto& collector_factories : *tbo.int_tbl_prop_collector_factories) {
+    const auto& factory_range = tbo.int_tbl_prop_collector_factories;
+    for (auto it = factory_range.first; it != factory_range.second; ++it) {
```
jay-zhuang, line 509: Just a question for the future usecase, will the needed properties are all in one range?

jay-zhuang (approved):
LGTM, just a few questions.

db/table_properties_collector.h
```diff
@@ -48,6 +48,13 @@ class IntTblPropCollectorFactory {
   virtual const char* Name() const = 0;
 };
 
+using IntTblPropCollectorFactories =
+    std::vector<std::unique_ptr<IntTblPropCollectorFactory>>;
+using IntTblPropCollectorFactoryIter =
+    IntTblPropCollectorFactories::const_iterator;
+using IntTblPropCollectorFactoryRange =
+    std::pair<IntTblPropCollectorFactoryIter, IntTblPropCollectorFactoryIter>;
```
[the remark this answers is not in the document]
ltamasi: We don't need it per se but it eliminates the duplication of `IntTblPropCollectorFactories::const_iterator` in the type alias for the range.

table/block_based/block_based_table_builder.cc
ltamasi: Do you mean putting the vector of factories and the range (i.e. the two iterators) in the same class? That wouldn't work since the vector has unique ownership of the factories; what we need instead is a way to refer to either the entire range or just a subrange of the same vector of factories.

table/block_based/block_based_table_builder.cc
ltamasi: Yes. For now, we actually just need a way to skip (or not skip) a single (BlobDB specific) factory.

PR #8299, opened 2021-05-14
Fix MultiGet with PinnableSlices and Merge for WBWI

mrambacher:
The MultiGetFromBatchAndDB would fail if the PinnableSlice value being returned was pinned.  This could happen if the value was retrieved from the DB (not memtable) or potentially if the values were reused (and a previous iteration returned a slice that was pinned).

This change resets the pinnable value to clear it prior to attempting to use it, thereby eliminating the problem with the value already being pinned.

### DIFF
--- a/utilities/write_batch_with_index/write_batch_with_index.cc
+++ b/utilities/write_batch_with_index/write_batch_with_index.cc
@@ -539,13 +539,15 @@ void WriteBatchWithIndex::MultiGetFromBatchAndDB(
   // after the transaction finishes.
   for (size_t i = 0; i < num_keys; ++i) {
     MergeContext merge_context;
-    PinnableSlice* pinnable_val = &values[i];
-    std::string& batch_value = *pinnable_val->GetSelf();
+    std::string batch_value;
     Status* s = &statuses[i];
+    PinnableSlice* pinnable_val = &values[i];
+    pinnable_val->Reset();
     auto result =
         wbwii.GetFromBatch(this, keys[i], &merge_context, &batch_value, s);
 
     if (result == WBWIIteratorImpl::kFound) {
+      *pinnable_val->GetSelf() = std::move(batch_value);
       pinnable_val->PinSelf();
       continue;
     }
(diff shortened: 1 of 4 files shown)


zhichao-cao:
You may need to explain if the fix is related to #8135 . If it is related, why 8135 triggers "rocksdb::PinnableSlice::PinSelf(): Assertion `!pinned_' failed." and how this PR fix them.

mrambacher:
> You may need to explain if the fix is related to #8135 . If it is related, why 8135 triggers "rocksdb::PinnableSlice::PinSelf(): Assertion `!pinned_' failed." and how this PR fix them.

#8135 did not introduce this issue, but made it more prevalent/likely (there were more cases in which a MultiGet could handle a merge rather than returning an error).

This fix addresses the "Assertion !pinned_ failed by resetting the PinnableSlice before calling PinSelf on the slice.  The reset is necessary because the PinnableSlice may be pinned (if the value came from an SST file or if the same variable is used for multiple calls to MultiGet).

The change used by this fix uses the same logic for MultiGetFromBatchAndDB as from GetFromBatchAndDB for handling/resetting the Pinnable-ness of the value parameter.

utilities/write_batch_with_index/write_batch_with_index.cc
```diff
@@ -539,13 +539,15 @@ void WriteBatchWithIndex::MultiGetFromBatchAndDB(
   // after the transaction finishes.
   for (size_t i = 0; i < num_keys; ++i) {
     MergeContext merge_context;
-    PinnableSlice* pinnable_val = &values[i];
-    std::string& batch_value = *pinnable_val->GetSelf();
+    std::string batch_value;
     Status* s = &statuses[i];
+    PinnableSlice* pinnable_val = &values[i];
```
jay-zhuang, line 544: Should we expect the user input values are all not pinned? I think it should be user's responsibility to reset the value if needed.

mrambacher: Is it the user's responsibility to guarantee that all values are not pinned?  The db methods that return a PinnableSlice say nothing about that property.

jay-zhuang: Without break the compatibility, I guess it's hard to restrict that now.

jay-zhuang (approved):
LGTM

PR #8301, opened 2021-05-14
Add customizable_util.h to the public API

mrambacher:
Useful for allowing new classes to create and manage Customizable objects without using internal APIs.


include/rocksdb/utilities/customizable_util.h
```diff
@@ -37,20 +44,19 @@ using StaticFactoryFunc = std::function<bool(const std::string&, T**)>;
...
+// pairs are specified, there should be an "id=value" pairing or an error may
 // result.
 //
 // The config_options parameter controls the process and how errors are
 // returned. If ignore_unknown_options=true, unknown values are ignored during
-// the configuration If ignore_unsupported_options=true, unknown instance types
-// are ignored If invoke_prepare_options=true, the resulting instance wll be
-// initialized (via PrepareOptions
+// the configuration. If ignore_unsupported_options=true, unknown instance types
+// are ignored. If invoke_prepare_options=true, the resulting instance wll be
```
zhichao-cao, line 53: wll-> will

include/rocksdb/utilities/customizable_util.h
```diff
@@ -15,6 +15,13 @@
 #include "rocksdb/utilities/object_registry.h"
 
 namespace ROCKSDB_NAMESPACE {
+// The FactoryFunc functions are used to create a new customizable object
```
zhichao-cao, line 18: For this file, can we create a customizable_util.cc in the /util and here, we only keep the declarations here？

include/rocksdb/utilities/object_registry.h
```diff
@@ -116,11 +116,31 @@ class ObjectLibrary {
   std::string id_;
 };
 
+#endif  // ROCKSDB_LITE
 // The ObjectRegistry is used to register objects that can be created by a
 // name/pattern at run-time where the specific implementation of the object may
 // not be known in advance.
 class ObjectRegistry {
+ private:
```
zhichao-cao, line 124: I may prefer put the changes of object_registry.h in another PR. It is not related to the changes in customziable_utl.h right?

include/rocksdb/utilities/customizable_util.h
```diff
@@ -0,0 +1,293 @@
...
+
+  Status status = Customizable::GetOptionsMap(config_options, result->get(),
+                                              value, &id, &opt_map);
+  if (!status.ok()) {  // GetOptionsMap failed
+    return status;
+  } else if (func == nullptr ||
+             !func(id, result)) {  // No factory, or it failed
+    if (value.empty()) {           // No Id and no options.  Clear the object
+      *result = nullptr;
+      return Status::OK();
```
zhichao-cao, line 125: If the opt_value is empty, should we return a non-ok status?

mrambacher: We could do that, but some of the existing classes use an empty value to mean "reset".  If you want it to mean error, I can make that change and pick up the pieces.

include/rocksdb/utilities/customizable_util.h
```diff
@@ -0,0 +1,293 @@
...
+// @param config_options Controls how the instance is created and errors are
+// handled
+// @param id The identifier of the new object being created.  This string
+// will be used by the object registry to locate the appropriate object to
+// create.
+// @param opt_map Optional name-value pairs of properties to set for the newly
+// created object
+// @param result The newly created and configured instance.
+template <typename T>
+static Status NewStaticObject(
```
zhichao-cao, line 231: Just a general question, can we replace some of the current static object function with this interface? For example, when we new a default env, FS, rate-limiter?

zhichao-cao (approved):
LGTM, please update the HISTORY.md before merge, since we introduce a set of public APIs.

include/rocksdb/utilities/customizable_util.h
zhichao-cao: > We could do that, but some of the existing classes use an empty value to mean "reset". If you want it to mean error, I can make that change and pick up the pieces.

I see, it would be ok to return Status::OK()

include/rocksdb/utilities/customizable_util.h
mrambacher: I think the CreateFromString methods can eventually many of the extern constructors.  The methods in this file are meant more for use by developers of classes rather than users

PR #8302, opened 2021-05-15
Expose CompressionOptions::parallel_threads through C API

stanislav-tkach:


ajkr (approved):
LGTM, thanks for adding the missing dictionary compression option getters too.

PR #8305, opened 2021-05-17
Implementation of Crc32c combine function

zhichao-cao:
Implement a function to generate the crc32c of two combined strings. Suppose we have the string 1 (s1) with crc32c checksum crc32c_1 and string 2 (s2) with crc32c checksum crc32c_2, the new string is s1+s2 and its checksum is crc32c_new=Crc32cCombine(crc32c_1, crc32c_2, s2.size).

test plan: make check, added new testing case

### DIFF
--- a/util/crc32c.cc
+++ b/util/crc32c.cc
@@ -1295,8 +1295,8 @@ static constexpr uint32_t gf_multiply_sw_1(
   // clang-format off
   return i == 32 ? p : gf_multiply_sw_1(
       /* i = */ i + 1,
-      /* p = */ p ^ (-((b >> 31) & 1) & a),
-      /* a = */ (a >> 1) ^ (-(a & 1) & m),
+      /* p = */ p ^ ((0u-((b >> 31) & 1)) & a),
+      /* a = */ (a >> 1) ^ ((0u-(a & 1)) & m),
       /* b = */ b << 1,
       /* m = */ m);
   // clang-format on
--- a/util/crc32c.cc
+++ b/util/crc32c.cc
@@ -19,10 +19,11 @@
 #include <nmmintrin.h>
 #include <wmmintrin.h>
 #endif
+
 #include "port/lang.h"
 #include "util/coding.h"
-
 #include "util/crc32c_arm64.h"
+#include "util/math.h"
 
 #ifdef __powerpc64__
 #include "util/crc32c_ppc.h"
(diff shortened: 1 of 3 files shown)


util/crc32c_test.cc
```diff
@@ -137,6 +137,22 @@ TEST(CRC, Mask) {
   ASSERT_EQ(crc, Unmask(Unmask(Mask(Mask(crc)))));
 }
 
+TEST(CRC, Crc32cCombineTest1) {
```
pdillinger, line 140: Please add more tests. We should probably cover all sizes of crc2 from 0 to 1000. Perhaps also one big size like 2^24 - 1 (for good coverage of the modular exponentiation).

util/crc32c_test.cc
```diff
@@ -137,6 +137,22 @@ TEST(CRC, Mask) {
...
+  uint32_t crc1 = Value("hello ", 6);
+  uint32_t crc2 = Value("world", 5);
+  uint32_t crc3 = Value("hello world", 11);
+  uint32_t crc1_2_combine = Crc32cCombine(crc1, crc2, 5);
+  ASSERT_EQ(crc3, crc1_2_combine);
+}
+
+TEST(CRC, Crc32cCombineTest2) {
```
pdillinger, line 148: Perhaps call it `Crc32cCombineOrderMatters`

pdillinger (changes requested):
I'm also going to send over a commented/simplified version of Crc32cCombine (which I helped to adapt to our crc32c implementation)

util/crc32c_test.cc
zhichao-cao: Sure, will add the testing

util/crc32c_test.cc
```diff
@@ -137,6 +139,55 @@ TEST(CRC, Mask) {
...
+TEST(CRC, Crc32cCombineFullCoverTest) {
+  int scale = 4 * 1024;
+  Random rnd(301);
+  Random r1(47);
+  int size_1 = static_cast<int>(r1.Next());
+  std::string s1 = rnd.RandomString(size_1);
+  uint32_t crc1 = Value(s1.c_str(), size_1);
+  for (int i = 0; i < scale; i++) {
+    int size_2 = i;
+    std::string s2 = rnd.RandomString(size_2);
```
pdillinger, line 167: `RandomBinaryString` should have better coverage here.

pdillinger (changes requested):

util/crc32c.cc
```diff
@@ -1279,5 +1284,164 @@ uint32_t Extend(uint32_t crc, const char* buf, size_t size) {
...
+// Standard galois-field multiply.  The only modification is that a,
+// b, m, and p are all bit-reflected.
+//
+// https://en.wikipedia.org/wiki/Finite_field_arithmetic
+static constexpr uint32_t gf_multiply_sw_1(
+    size_t i, uint32_t p, uint32_t a, uint32_t b, uint32_t m) {
+  // clang-format off
+  return i == 32 ? p : gf_multiply_sw_1(
+      /* i = */ i + 1,
+      /* p = */ static_cast<uint32_t>(p ^ (static_cast<int64_t>(-((b >> 31) & 1) & a))),
```
pdillinger, line 1298: I guess we have to polish this line from folly to make Microsoft happy. I'm not sure what the casts are supposed to be for. It looks like it computes `/* p = */ p ^ ((0U - ((b >> 31) & 1U)) & a)`

pdillinger (approved):
LGTM once msvc is happy

zhichao-cao:
> LGTM once msvc is happy

Thanks for the help and review. Will address the windows build failu.

rockeet:
Can Crc32cCombine be optimized? I have filed an issue: https://github.com/facebook/rocksdb/issues/13535, would any one take a look at this issue?

PR #8307, opened 2021-05-17
Deflake ExternalSSTFileTest.PickedLevelBug

jay-zhuang:
The test want to make sure these's no compaction during `AddFile`
(between `DBImpl::AddFile:MutexLock` and `DBImpl::AddFile:MutexUnlock`)
but the mutex could be unlocked by `EnterUnbatched()`.
Move the lock start point after bumping the ingest file number.

Also fix the dead lock when ASSERT fails.


ajkr (approved):
LGTM

PR #8313, opened 2021-05-18
Do not truncate WAL if in read_only mode

thatsafunnyname:
I noticed ```openat``` system call with ```O_WRONLY``` flag and ```sync_file_range``` and ```truncate``` on WAL file when using ```rocksdb::DB::OpenForReadOnly``` by way of ```db_bench --readonly=true --benchmarks=readseq --use_existing_db=1 --num=1 ...```

Noticed in ```strace``` after seeing the last modification time of the WAL file change after each run (with ```--readonly=true```).

  I think introduced by https://github.com/facebook/rocksdb/commit/7d7f14480e135a4939ed6903f46b3f7056aa837a from https://github.com/facebook/rocksdb/pull/8122
 
I added a test to catch the WAL file being truncated and the modification time on it changing.
I am not sure if a mock filesystem with mock clock could be used to avoid having to sleep 1.1s.
The test could also check the set of files is the same and that the sizes are also unchanged.

Before:
 
```
[ RUN      ] DBBasicTest.ReadOnlyReopenMtimeUnchanged
db/db_basic_test.cc:182: Failure
  file_mtime_after_readonly_reopen
    Which is: 1621611136
  file_mtime_before_readonly_reopen
    Which is: 1621611135
  file is: 000010.log
[  FAILED  ] DBBasicTest.ReadOnlyReopenMtimeUnchanged (1108 ms)
```

After:
 
```
[ RUN      ] DBBasicTest.ReadOnlyReopenMtimeUnchanged
[       OK ] DBBasicTest.ReadOnlyReopenMtimeUnchanged (1108 ms)
```


jay-zhuang:
Thanks @thatsafunnyname for the contribution. Would you please add an unittest for this?

thatsafunnyname:
The ci/circleci: build-linux-clang10-ubsan tests failed with "No space left on device".
I will trigger another run with a whitespace change.

jay-zhuang:
> The ci/circleci: build-linux-clang10-ubsan tests failed with "No space left on device".
> I will trigger another run with a whitespace change.

No worry, it's a known issue, not related to this PR.

jay-zhuang:
I updated test to be more specify about WAL file truncate.

jay-zhuang (approved):
LGTM

PR #8315, opened 2021-05-18
Use new Insert and Lookup APIs in table reader to support secondary cache

zhichao-cao:
Secondary cache is implemented to achieve the secondary cache tier for block cache. New Insert and Lookup APIs are introduced in #8271  . To support and use the secondary cache in block based table reader, this PR introduces the corresponding callback functions that will be used in secondary cache, and update the Insert and Lookup APIs accordingly.

benchmarking:
./db_bench --benchmarks="fillrandom" -num=1000000 -key_size=32 -value_size=256 -use_direct_io_for_flush_and_compaction=true -db=/tmp/rocks_t/db -partition_index_and_filters=true

./db_bench -db=/tmp/rocks_t/db -use_existing_db=true -benchmarks=readrandom -num=1000000 -key_size=32 -value_size=256 -use_direct_reads=true -cache_size=1073741824 -cache_numshardbits=5 -cache_index_and_filter_blocks=true -read_random_exp_range=17 -statistics -partition_index_and_filters=true -stats_dump_period_sec=30 -reads=50000000

master benchmarking results: 
readrandom   :       3.923 micros/op 254881 ops/sec;   33.4 MB/s (23849796 of 50000000 found)
rocksdb.db.get.micros P50 : 2.820992 P95 : 5.636716 P99 : 16.450553 P100 : 8396.000000 COUNT : 50000000 SUM : 179947064

Current PR benchmarking results
readrandom   :       4.083 micros/op 244925 ops/sec;   32.1 MB/s (23849796 of 50000000 found)
rocksdb.db.get.micros P50 : 2.967687 P95 : 5.754916 P99 : 15.665912 P100 : 8213.000000 COUNT : 50000000 SUM : 187250053

About 3.8% throughput reduction.
P50: 5.2% increasing, P95, 2.09% increasing, P99 4.77% improvement

test plan: added the testing case

### DIFF
--- a/cache/lru_cache_test.cc
+++ b/cache/lru_cache_test.cc
@@ -204,7 +204,7 @@ TEST_F(LRUCacheTest, EntriesWithPriority) {
 class TestSecondaryCache : public SecondaryCache {
  public:
   explicit TestSecondaryCache(size_t capacity)
-      : num_inserts_(0), num_lookups_(0) {
+      : num_inserts_(0), num_lookups_(0), inject_failure_(false) {
     cache_ = NewLRUCache(capacity, 0, false, 0.5, nullptr,
                          kDefaultToAdaptiveMutex, kDontChargeCacheMetadata);
   }
@@ -212,8 +212,15 @@ class TestSecondaryCache : public SecondaryCache {
 
   std::string Name() override { return "TestSecondaryCache"; }
 
+  void InjectFailure() { inject_failure_ = true; }
+
+  void ResetInjectFailure() { inject_failure_ = false; }
+
   Status Insert(const Slice& key, void* value,
                 const Cache::CacheItemHelper* helper) override {
+    if (inject_failure_) {
+      return Status::Corruption("Insertion Data Corrupted");
+    }
     size_t size;
     char* buf;
     Status s;
@@ -292,6 +299,7 @@ class TestSecondaryCache : public SecondaryCache {
   std::shared_ptr<Cache> cache_;
   uint32_t num_inserts_;
   uint32_t num_lookups_;
+  bool inject_failure_;
 };
 
 class DBSecondaryCacheTest : public DBTestBase {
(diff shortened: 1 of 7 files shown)


anand1976:
This may require some coordination with #8297

zhichao-cao:
> This may require some coordination with #8297

Oh, yes. You are right. I'm not aware of this PR. I may need to wait for @pdillinger to merge the PR and update the CacheItemHelper accordingly in block_based/block_like_traits.h?

pdillinger:
> > This may require some coordination with #8297
> 
> Oh, yes. You are right. I'm not aware of this PR. I may need to wait for @pdillinger to merge the PR and update the CacheItemHelper accordingly in block_based/block_like_traits.h?

Thanks. I'll try to land soon.

table/block_based/block_based_table_reader.cc
```diff
@@ -1172,9 +1193,10 @@ Status BlockBasedTable::GetDataBlockFromCache(
         read_options.fill_cache) {
       size_t charge = block_holder->ApproximateMemoryUsage();
       Cache::Handle* cache_handle = nullptr;
-      auto deleter = BlocklikeTraits<TBlocklike>::GetDeleter(block_type);
-      s = block_cache->Insert(block_cache_key, block_holder.get(), charge,
-                              deleter, &cache_handle);
+      s = block_cache->Insert(
+          block_cache_key, block_holder.get(),
+          BlocklikeTraits<TBlocklike>::GetCacheItemHelper(block_type), charge,
+          &cache_handle, priority);
```
anand1976, line 1199: Interesting. This is actually an improvement from before. Previously, any block copied from the compressed to the uncompressed cache was low priority.

cache/lru_cache_test.cc
```diff
@@ -547,6 +566,195 @@ TEST_F(LRUSecondaryCacheTest, FullCapacityTest) {
...
+  const int N = 6;
+  for (int i = 0; i < N; i++) {
+    std::string p_v = rnd.RandomString(1000);
+    ASSERT_OK(Put(Key(i), p_v));
+  }
+  ASSERT_OK(Flush());
+  Compact("a", "z");
+  // After compaction, a iterator is created to test if the file is correct, so
+  // one block read is triggered. Since there is no data in the block cache,
+  // block cache Lookup will also check the tiered cache and trigger one read.
```
anand1976, line 594: tiered -> secondary

cache/lru_cache_test.cc
```diff
@@ -547,6 +566,195 @@ TEST_F(LRUSecondaryCacheTest, FullCapacityTest) {
...
+  DestroyAndReopen(options);
+  Random rnd(301);
+  const int N = 6;
+  for (int i = 0; i < N; i++) {
+    std::string p_v = rnd.RandomString(1000);
+    ASSERT_OK(Put(Key(i), p_v));
+  }
+  ASSERT_OK(Flush());
+  Compact("a", "z");
+  // After compaction, a iterator is created to test if the file is correct, so
```
anand1976, line 592: Does compaction also create an iterator to test the file? I thought only flush did that.

anand1976 (approved):
LGTM. Just some minor comments.

table/block_based/block_like_traits.h
```diff
@@ -32,14 +55,32 @@ class BlocklikeTraits<BlockContents> {
     return 0;
   }
 
-  static Cache::DeleterFn GetDeleter(BlockType block_type) {
+  static size_t SizeCallback(void* obj) {
+    assert(obj != nullptr);
+    BlockContents* ptr = reinterpret_cast<BlockContents*>(obj);
```
pdillinger, line 60: Please use `static_cast` to cast from `void*`

table/block_based/block_like_traits.h
```diff
@@ -32,14 +55,32 @@ class BlocklikeTraits<BlockContents> {
...
+    return ptr->data.size();
+  }
+
+  static Status SaveToCallback(void* obj, size_t offset, size_t size,
```
pdillinger, line 64: Surely we can share most of the code between the SaveToCallback implementations. In fact, the appropriate "trait" seems to be getting a Slice of underlying data from a `void* obj`. From that trait, you can have unified (template) implementations of SizeCallback and SaveToCallback.

table/block_based/block_like_traits.h
```diff
@@ -32,14 +55,32 @@ class BlocklikeTraits<BlockContents> {
...
+                               void* out) {
+    assert(obj != nullptr);
+    BlockContents* ptr = reinterpret_cast<BlockContents*>(obj);
+    const char* buf = ptr->data.data();
+    assert(size == ptr->data.size());
+    assert(offset == 0);
```
pdillinger, line 70: Also, why not just implement the `offset` functionality now? I find this `assert(offset = 0)` to be confusing.

table/block_based/block_like_traits.h
```diff
@@ -32,14 +55,32 @@ class BlocklikeTraits<BlockContents> {
...
+    BlockContents* ptr = reinterpret_cast<BlockContents*>(obj);
+    return ptr->data.size();
+  }
+
+  static Status SaveToCallback(void* obj, size_t offset, size_t size,
+                               void* out) {
+    assert(obj != nullptr);
+    BlockContents* ptr = reinterpret_cast<BlockContents*>(obj);
+    const char* buf = ptr->data.data();
+    assert(size == ptr->data.size());
```
pdillinger, line 69: I also recommend renaming `size` to `length` here and in cache.h. (We are closer to "What length of rope would you like?" than "What size of rope would you like?")

It's also not clear whether `offset` refers to input or output. Perhaps `from_offset` and even `from_obj`.

table/block_based/block_like_traits.h
zhichao-cao: One case I'm not sure is ParsedFullFilterBlock. Its data is in the block_content_ as a private member

pdillinger (approved):
Thanks for some clean-up.

I recommend running a performance test, for example with 100+ MB db, cache_index_and_filter_blocks=1, and generous block cache big enough for whole DB. This should stress the block-based table reader caching layer.

zhichao-cao:
> Thanks for some clean-up.
> 
> I recommend running a performance test, for example with 100+ MB db, cache_index_and_filter_blocks=1, and generous block cache big enough for whole DB. This should stress the block-based table reader caching layer.

Thanks for the review! Good point, will use db_bench for the testing as you suggested and see if the change will cause any performance regression.

PR #8316, opened 2021-05-18
Add a BlobDB-specific table property collector

ltamasi:
Summary:
The patch adds a BlobDB-specific table property collector that processes
any blob indexes added to an SST and keeps track of the total number and
size of blobs referenced by the SST on a per-blob file basis. This aggregated
information is then persisted as a new table property `rocksdb.blob.file.mapping`.

This information will be used to calculate the amount of garbage generated by
compactions. Namely, the amount of additional garbage for any given blob file
can be computed by subtracting the "outflow" (total number/size of blobs in the
output SSTs) from the "inflow" (total number/size of blobs in the input SSTs).
Tracking the amount of garbage in blob files in turn will allow us to optimize GC
performance. (Note: one might consider counting blobs e.g. in
`CompactionIterator` as we process key-values during compaction;
however, this would be very intrusive, and there are actually cases when
`CompactionIterator` does not process all keys individually, for instance
with deletion/TTL compactions or when a compaction filter returns
`kRemoveAndSkipUntil`.)

Test Plan:
`make check`


db/column_family.h
```diff
@@ -427,8 +428,19 @@ class ColumnFamilyData {
     return internal_comparator_;
   }
 
-  const IntTblPropCollectorFactories* int_tbl_prop_collector_factories() const {
-    return &int_tbl_prop_collector_factories_;
+  IntTblPropCollectorFactoryRange int_tbl_prop_collector_factories(
+      bool enable_blob_property_collector) const {
+    assert(!int_tbl_prop_collector_factories_.empty());
+    assert(dynamic_cast<const BlobTablePropertiesCollectorFactory*>(
+        int_tbl_prop_collector_factories_.front().get()));
```
jay-zhuang, line 435: I feel the restriction that BlobTablePropCollectorFactory has to be the first one is not so obvious.

db/blob/blob_stats_record.cc
```diff
@@ -0,0 +1,37 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#include "db/blob/blob_stats_record.h"
```
jay-zhuang, line 6: Just general code style question, should for example, `blob_stats.h`, `blob_stats_record.h` and  `blob_stats_collection.h` be combined into one file? And the same for *.cc files.

I have the impression that our codebase has lots of related structures defined in one file.

db/db_compaction_test.cc
```diff
@@ -5974,6 +5974,18 @@ TEST_F(DBCompactionTest, CompactionWithBlob) {
...
+  ASSERT_OK(db_->GetPropertiesOfAllTables(&all_table_props));
+
+  ASSERT_EQ(all_table_props.size(), 1);
+
+  const auto& table_props = all_table_props.begin()->second;
+  ASSERT_NE(table_props, nullptr);
+
+  const auto& user_props = table_props->user_collected_properties;
+  ASSERT_NE(user_props.find(TablePropertiesNames::kBlobFileMapping),
+            user_props.end());
```
jay-zhuang, line 5988: Why don't also check the value of property?

jay-zhuang (approved):
Should such aggregated information be put into a meta block, instead of a table property?

ltamasi:
Thanks for the review @jay-zhuang !

ltamasi:
> Should such aggregated information be put into a meta block, instead of a table property?

You mean a dedicated metablock of its own? (Table properties are also stored in a metablock.) I feel that would be overkill; it would involve a lot of changes for the sake of a single property. The table property collector framework gives us exactly what we need in this case.

db/column_family.h
ltamasi: It's added before the user-specified ones (see lines 558-559 in `column_family.cc`) so it's guaranteed to be at the head of the list. I can add a comment though to point this out.

db/blob/blob_stats_record.cc
ltamasi: I don't think we have a hard and fast rule for this; I personally prefer to have separate source files because they can reduce dependencies. (For example, `blob_table_properties_collector.h` only includes `blob_stats.h` but not the other two.) As for the pre-existing stuff, there has been some cleanup work done recently to split up large source files to make file sizes more manageable.

db/db_compaction_test.cc
ltamasi: I felt that would be redundant since we have dedicated unit tests for that. Here, I'm just trying to make sure BlobDB's property collector was active as expected.

jay-zhuang:
> > Should such aggregated information be put into a meta block, instead of a table property?
> 
> You mean a dedicated metablock of its own? (Table properties are also stored in a metablock.) I feel that would be overkill; it would involve a lot of changes for the sake of a single property. The table property collector framework gives us exactly what we need in this case.

Yeah, a dedicated metablock, which maybe only for blobDB, I think it will be more customizable for more information in the future. But anyway, a property here would be easier to implement and use I guess.

ammarfaizi2 (approved):

ltamasi:
Closing since we ended up going for a different approach (see #8450 ).

PR #8320, opened 2021-05-20
[CI] Use large macos instance

jay-zhuang:
Macos build is taking more than 1 hour, bump the instance type from the
default medium to large (large macos instance was not available before).

Test Plan: watch CI pass


ajkr (approved):
LGTM.

PR #8324, opened 2021-05-21
Fix cmake build failure with gflags

jay-zhuang:
- Fix cmake build failure with gflags.
- Add CI tests for both gflags 2.1 and 2.2.
- Fix ctest config with gtest.
- Add CI to run test with ctest.

One benefit of ctest is it support timeout, it's set to 5min in our CI, so we will know which test is hang.

Test Plan: CI pass


CMakeLists.txt
```diff
@@ -118,16 +125,16 @@ else()
     option(GFLAGS_USE_TARGET_NAMESPACE "Use gflags import target with namespace." ON)
     find_package(gflags CONFIG)
     if(gflags_FOUND)
-      if(TARGET ${GFLAGS_TARGET})
+      if(TARGET ${gflags_TARGET})
```
ajkr, line 128: This seems to disagree with this comment that `gflags_TARGET` never existed - https://github.com/facebook/rocksdb/pull/6771#issuecomment-846539339. Is this PR still under development or do you mind explaining why it still works?

ajkr (approved):
LGTM

CMakeLists.txt
jay-zhuang: `gflags_TARGET` does exist in 2.2 based on my test on both linux and macos. But as it's only introduced in 2.2, I think it make sense to use the upper case one.

Seems gflags 2.1 is using lower case ones, like `gflags_LIBRARIES`, the new one 2.2, is using both upper case and lower case ones like `GFLAGS_LIBRARIES` and `gflags_LIBRARIES`, (I guess it's for backward compatibility). So I was using the lower case ones which should work for both. but as TARGET is introduced in 2.2, I guess it also make sense to use the upper case one, updated.

PR #8326, opened 2021-05-22
fix lru caching test and fix reference binding to null pointer

zhichao-cao:
Fix for #8315. Inhe lru caching test, 5100 is not enough to hold meta block and first block in some random case, increase to 6100. Fix the reference binding to null pointer, use template.

test plan: make check

### DIFF
--- a/table/block_based/block_like_traits.h
+++ b/table/block_based/block_like_traits.h
@@ -23,8 +23,7 @@ Cache::CacheItemHelper* GetCacheItemHelperForRole();
 template <typename TBlocklike>
 Cache::CreateCallback GetCreateCallback(size_t read_amp_bytes_per_bit,
                                         Statistics* statistics, bool using_zstd,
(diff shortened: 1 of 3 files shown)


table/block_based/block_like_traits.h
```diff
@@ -21,10 +21,10 @@ template <typename T, CacheEntryRole R>
 Cache::CacheItemHelper* GetCacheItemHelperForRole();
 
 template <typename TBlocklike>
-Cache::CreateCallback GetCreateCallback(size_t read_amp_bytes_per_bit,
-                                        Statistics* statistics, bool using_zstd,
-                                        const FilterPolicy* filter_policy,
-                                        const TBlocklike& /*block*/) {
```
pdillinger, line 27: Were you just using this block parameter to automatically instantiate the template? You can explicitly provide a type parameter on call without this block parameter. Wouldn't that avoid the extra stuff added in this PR?

PR #8329, opened 2021-05-25
Warning about incompatible options with level_compaction_dynamic_level_bytes

asmolik:
This change introduces warnings instead of a silent override when trying to use level_compaction_dynamic_level_bytes with multiple cf_paths/db_paths.
I have completed the CLA.


ajkr (approved):
LGTM

Issue #8239, opened 2021-04-28
Call to DeleteRange with Invalid argument breaks subsequent call with valid argument


alanpaxton:
> Note: Please use Issues only for bug reports. For questions, discussions, feature requests, etc. post to dev group: https://groups.google.com/forum/#!forum/rocksdb or https://www.facebook.com/groups/rocksdb.dev

### Expected behavior

Attempt to delete a valid key range in the DB. Succeeds, correctly.
Attempt to delete an invalid key range (end before start) Fails, correctly.
(*) Attempt to delete another valid key range in the DB. Should succeed.

### Actual behavior

(*) fails, with ""Invalid argument: end key comes before start key"
See included test method for example (this can be pasted into `db_range_del_test.cc` and run.

### Steps to reproduce the behavior

```
TEST_F(DBRangeDelTest, DBInvalidRangeCascadingProblem) {
  DB* db;

  Options options;
  options.create_if_missing = true;
  options.merge_operator = MergeOperators::CreateFromStringId("stringappend");
  std::string dbname = test::PerThreadDBPath("db_write_batch_problem");

  ASSERT_OK(DestroyDB(dbname, options));
  Status s = DB::Open(options, dbname, &db);
  ASSERT_OK(s);

  WriteOptions write_options;

  Status s1 = db->DeleteRange(write_options, nullptr, "Y", "Z");
  ASSERT_OK(s1);
  Status s2 = db->DeleteRange(write_options, nullptr, "Z", "Y");
  ASSERT_TRUE(s2.IsInvalidArgument());
  Status s3 = db->DeleteRange(write_options, nullptr, "A", "B");
  ASSERT_OK(s3);  // This fails, "end key comes before start key" !!
}
```

ajkr:
For `DeleteRange()` validation we need to access the user's configured comparator. There's currently an implementation constraint that we don't have convenient access to the user's options in early stages of some write APIs. This constraint also causes the same problem for users who call `Merge()` on a DB with no `merge_operator` configured.

abbccdda:
This looks like a good starter project, could I have a try?

ajkr:
@abbccdda Sure, thanks for your interest!

I have an idea about how to proactively validate the range in almost every case. There is one final case (probably rare) I cannot think of how to solve.

[...]

PR #8340, opened 2021-05-27
Issue 8239: Add early column family check for DeleteRange 

abbccdda:
This PR tries to address the [issue](https://github.com/facebook/rocksdb/issues/8239) of getting the invalid range delete to the later stage which brings down the entire database, by always providing a valid column family and the sanity check by using `GetComparator`.

### DIFF
--- a/db/db_range_del_test.cc
+++ b/db/db_range_del_test.cc
@@ -110,6 +110,28 @@ TEST_F(DBRangeDelTest, CompactionOutputHasOnlyRangeTombstone) {
                          kSkipFIFOCompaction));
 }
 
+TEST_F(DBRangeDelTest, DBInvalidRangeCascadingProblem) {
+  DB* db;
+
+  Options options;
+  options.create_if_missing = true;
+  options.merge_operator = MergeOperators::CreateFromStringId("stringappend");
+  std::string dbname = test::PerThreadDBPath("db_write_batch_problem");
+
+  ASSERT_OK(DestroyDB(dbname, options));
+  Status s = DB::Open(options, dbname, &db);
+  ASSERT_OK(s);
+
+  WriteOptions write_options;
+
+  Status s1 = db->DeleteRange(write_options, nullptr, "Y", "Z");
+  ASSERT_OK(s1);
+  Status s2 = db->DeleteRange(write_options, nullptr, "Z", "Y");
+  ASSERT_TRUE(s2.IsInvalidArgument());
+  Status s3 = db->DeleteRange(write_options, nullptr, "A", "B");
+  ASSERT_OK(s3);  // This fails, "end key comes before start key" !!
+}
+
 TEST_F(DBRangeDelTest, CompactionOutputFilesExactlyFilled) {
   // regression test for exactly filled compaction output files. Previously
   // another file would be generated containing all range deletions, which
--- a/db/db_range_del_test.cc
+++ b/db/db_range_del_test.cc
@@ -129,7 +129,7 @@ TEST_F(DBRangeDelTest, DBInvalidRangeCascadingProblem) {
   Status s2 = db->DeleteRange(write_options, nullptr, "Z", "Y");
   ASSERT_TRUE(s2.IsInvalidArgument());
   Status s3 = db->DeleteRange(write_options, nullptr, "A", "B");
-  ASSERT_OK(s3);  // This fails, "end key comes before start key" !!
+  ASSERT_OK(s3);  // This should work as the db is still alive.
 }
 
 TEST_F(DBRangeDelTest, CompactionOutputFilesExactlyFilled) {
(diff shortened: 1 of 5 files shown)


abbccdda:
@ajkr I looked at the WriteBatch constructor and see there were quite some references to it. Since we are just trying to get a default column family, do you think we need to change every single constructor to allow the passing of CF, which seems to be a major change. Do you think we could have a better chance to address the only empty case for the API:

```
Status WriteBatchInternal::DeleteRange(WriteBatch* b,
                                       uint32_t column_family_id,
                                       const Slice& begin_key,
                                       const Slice& end_key) {
  return WriteBatchInternal::DeleteRange(b, nullptr, column_family_id, begin_key, end_key);
}
```
So that we could always pass a valid column family, instead of just getting a column family id? If we got this case covered, no need to alter the constructor.

db/write_batch.cc
```diff
@@ -1112,9 +1112,23 @@ Status WriteBatch::SingleDelete(ColumnFamilyHandle* column_family,
...
+Status WriteBatchInternal::DeleteRange(WriteBatch* b,
+                                       uint32_t column_family_id,
+                                       const Slice& begin_key,
+                                       const Slice& end_key) {
+  return WriteBatchInternal::DeleteRange(b, nullptr, column_family_id, begin_key, end_key);
+}
+
+Status WriteBatchInternal::DeleteRange(WriteBatch* b,
+                                       ColumnFamilyHandle* handle,
+                                       uint32_t column_family_id,
```
mrambacher, line 1124: Rather than taking in both a handle and an ID, can you make a public API that takes in the handle and a private one that takes the ID and have the public one call the private?  Or are there still place you will not have a handle?

db/write_batch.cc
```diff
@@ -1112,9 +1112,23 @@ Status WriteBatch::SingleDelete(ColumnFamilyHandle* column_family,
...
                                        const Slice& begin_key,
                                        const Slice& end_key) {
+  assert(handle->GetComparator() != nullptr);
+  if (handle->GetComparator()->Compare(begin_key, end_key) > 0) {
```
mrambacher, line 1128: I believe there are also types of tables that do not support delete range.  Can/Should those also be checked here?

db/write_batch.cc
mrambacher: If not, can you please assert that the input handle has the same ID?

db/db_range_del_test.cc
```diff
@@ -110,6 +110,28 @@ TEST_F(DBRangeDelTest, CompactionOutputHasOnlyRangeTombstone) {
                          kSkipFIFOCompaction));
 }
 
+TEST_F(DBRangeDelTest, DBInvalidRangeCascadingProblem) {
+  DB* db;
+
+  Options options;
+  options.create_if_missing = true;
+  options.merge_operator = MergeOperators::CreateFromStringId("stringappend");
```
mrambacher, line 118: Does the merge_operator come into play here at all?

db/db_range_del_test.cc
```diff
@@ -110,6 +110,28 @@ TEST_F(DBRangeDelTest, CompactionOutputHasOnlyRangeTombstone) {
...
+
+  WriteOptions write_options;
+
+  Status s1 = db->DeleteRange(write_options, nullptr, "Y", "Z");
+  ASSERT_OK(s1);
+  Status s2 = db->DeleteRange(write_options, nullptr, "Z", "Y");
+  ASSERT_TRUE(s2.IsInvalidArgument());
+  Status s3 = db->DeleteRange(write_options, nullptr, "A", "B");
+  ASSERT_OK(s3);  // This should work as the db is still alive.
+}
```
mrambacher, line 133: Should you close/destroy the DB when the test completes or (preferably) can that be done in the destructor?

db/write_batch.cc
```diff
@@ -1112,9 +1112,23 @@ Status WriteBatch::SingleDelete(ColumnFamilyHandle* column_family,
       this, GetColumnFamilyID(column_family), key);
 }
 
-Status WriteBatchInternal::DeleteRange(WriteBatch* b, uint32_t column_family_id,
+Status WriteBatchInternal::DeleteRange(WriteBatch* b,
+                                       uint32_t column_family_id,
+                                       const Slice& begin_key,
+                                       const Slice& end_key) {
+  return WriteBatchInternal::DeleteRange(b, nullptr, column_family_id, begin_key, end_key);
+}
```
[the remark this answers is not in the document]
abbccdda: Yes, I haven't figured out a good way to inject a column family ptr here

db/write_batch.cc
abbccdda: @mrambacher One question, what's the relation between column family handle and column family id? Could we reconstruct CF handle based off CF id?

abbccdda:
@ajkr Could you share some thoughts on this one? Right now I'm a bit concerned about the scope of changing WriteBatch constructor, let me know if you want me to proceed here.

javeme:
@abbccdda Will you continue to update this PR in the future? I also hope to fix this issue, maybe I can help to try it together.

abbccdda:
@javeme feel free to take it over

PR #8346, opened 2021-05-28
Improvements to benchmark.sh script

mrambacher:
1.  Fix printing of stats when there are no writes (wamp=0).  Previously had a div0 error

2.  Added multireadrandom command as a valid target

3.  Added ability to pass additional command line options to db_bench.  Now can say things like benchmark.sh readrandom --mmap_read and the option will be passed to db_bench.


zhichao-cao (approved):
LGTM, thanks!

PR #8351, opened 2021-06-01
Cancel compact range

ddevec:
Added the ability to cancel an in-progress range compaction by storing to an atomic "canceled" variable pointed to within the CompactRangeOptions structure.

Tested via two tests added to db_tests2.cc.

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -1257,7 +1257,7 @@ Status DBImpl::CompactFilesImpl(
       c->mutable_cf_options()->paranoid_file_checks,
       c->mutable_cf_options()->report_bg_io_stats, dbname_,
       &compaction_job_stats, Env::Priority::USER, io_tracer_,
-      &manual_compaction_paused_, db_id_, db_session_id_,
+      &manual_compaction_paused_, nullptr, db_id_, db_session_id_,
       c->column_family_data()->GetFullHistoryTsLow());
 
   // Creating a compaction influences the compaction score because the score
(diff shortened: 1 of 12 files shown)


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3140,7 +3151,8 @@ Status DBImpl::BackgroundCompaction(bool* made_progress,
         c->mutable_cf_options()->paranoid_file_checks,
         c->mutable_cf_options()->report_bg_io_stats, dbname_,
         &compaction_job_stats, thread_pri, io_tracer_,
-        is_manual ? &manual_compaction_paused_ : nullptr, db_id_,
+        is_manual ? &manual_compaction_paused_ : nullptr,
+        is_manual ? manual_compaction->canceled : nullptr, db_id_,
```
ajkr, line 3155: I don't have an opinion yet on whether consolidation is worthwhile, but it does feel we have three checks (`shutting_down_`, `manual_compaction_paused_`, `manual_compaction_canceled_`) that could be reduced.

One idea (not fully thought through) to eliminate  `manual_compaction_paused_`: we could make the `std::atomic<bool>*` in `ManualCompactionState` point to something managed by us when not provided by user. Then in `DisableManualCompaction()` we would iterate through the `ManualCompactionState`s in `manual_compaction_dequeue_` and set each of their `*canceled = true`

ajkr (approved):
LGTM, thanks! Wrote down an idea for consolidating the checks but we don't need to do it now, IMO.

db/db_impl/db_impl_compaction_flush.cc
ddevec: I considered this, but worried about the multi-threaded implications.

Currently `DisableManualCompaction()` atomically shuts down all compactions, if we were to set a series of variables, then we wouldn't be atomically disabling all compactions without adding additional locking, which would be more nuanced and could potentially break some use-cases.  Without a deeper understanding of how the system works and how `DisableManualCompaction` was used, I didn't feel fully comfortable merging the two behaviors together.

ajkr (commented):
Sorry I forgot but we should update HISTORY.md "Public API" section to mention this. It is fine to be a separate PR or delegated to me if you want to work on other things.

PR #8354, opened 2021-06-02
Make EncryptionProvider and BlockCipher into Customizable objects

mrambacher:
Made the EncryptionProvider and BlockCipher classes inherit from Customizable.  Added/fixed the CreateFromString method to these classes to create instances from builtin or registered classes.  Added tests to verify that instances can be registered and retrieved as appropriate.

Added the ability to configure the builtin (CTR, ROT13) classes from configurable properties.  Added the appropriate tests.


env/env_encryption_ctr.h
```diff
@@ -10,31 +10,6 @@
 #include "rocksdb/env_encryption.h"
 
 namespace ROCKSDB_NAMESPACE {
-
-// Implements a BlockCipher using ROT13.
-//
-// Note: This is a sample implementation of BlockCipher,
-// it is NOT considered safe and should NOT be used in production.
-class ROT13BlockCipher : public BlockCipher {
```
zhichao-cao, line 18: why we move it out from the header file?

zhichao-cao (approved):
LGTM

env/env_encryption_ctr.h
mrambacher: There is no need for it in the header file.  The class does not need to be public

env/env_basic_test.cc
```diff
@@ -17,15 +17,81 @@
 #include "test_util/testharness.h"
 
 namespace ROCKSDB_NAMESPACE {
+typedef Env* CreateEnvFunc();
```
zhichao-cao, line 20: The changes here to avoid the early initialization before main is ok to me. I think it is ready to merge.

PR #4948, opened 2019-02-05
Drop `level_and_file.second->num_deletions > 1` in `ComputeBottommostFilesMarkedForCompaction`.

mzhaom:
This is a follow-up to #4927, which made this possible by allowing tombstone dropping/seqnum zeroing optimizations on the last key in the compaction. Now the `largest_seqno != 0` condition suffices to prevent snapshot release triggered compaction from entering an infinite loop.

The issues caused by the extraneous condition `level_and_file.second->num_deletions > 1` instead are:

- it doesn't trigger compaction when there's one deletion, which could be a range deletion covering a lot of keys in the file
- it doesn't trigger compaction when there are many overwritten keys. Some MyRocks use case actually doesn't use Delete but instead calls Put with empty value to "delete" keys, so we'd like to be able to trigger compaction in this case too.

Add flag to prevent double calling the callback in test.

Test plan:
 - make check
 - ./db_bench

PR #4927, opened 2019-01-28
Zero seqnum of final key / drop final tombstone when compacting to bottommost level

mzhaom:

PR #8357, opened 2021-06-03
Snapshot release triggered compaction without multiple tombstones

ajkr:
This is a duplicate of #4948 by @mzhaom to fix tests after rebase.

This change is a follow-up to #4927, which made this possible by allowing tombstone dropping/seqnum zeroing optimizations on the last key in the compaction. Now the `largest_seqno != 0` condition suffices to prevent snapshot release triggered compaction from entering an infinite loop.

The issues caused by the extraneous condition `level_and_file.second->num_deletions > 1` are:

- files could have `largest_seqno > 0` forever making it impossible to tell they cannot contain any covering keys
- it doesn't trigger compaction when there are many overwritten keys. Some MyRocks use case actually doesn't use Delete but instead calls Put with empty value to "delete" keys, so we'd like to be able to trigger compaction in this case too.

Test plan:
 - make check

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,7 @@
 ## Unreleased
 ### Behavior Changes
 * Added API comments clarifying safe usage of Disable/EnableManualCompaction and EventListener callbacks for compaction.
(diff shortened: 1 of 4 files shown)


jay-zhuang (approved):
LGTM.
I think it worth adding a note in HISTORY.

ajkr:
Thanks for the review!

> I think it worth adding a note in HISTORY.

Done.

Issue #7405, opened 2020-09-17
SST files overlap at a level and keys are not ordered in a SST file


kevinliuca20:
We are upgrading our rocksdb storage from 5.7.2 to 6.4.6. After upgrading to the 6.4.6 version, we noticed some shards returned unexpected keys in the iteration. Then we dumped the manifest and sst files, and found the (smallest key, greatest key) of some SST files at the bottommost level overlapped,  and some SST files contained multiple sorted ranges but these ranges are overlapped. 

For example, an sst file contains 4 segments of sorted keys, these are the ranges showing the first 4 bytes of the keys in hex format(the common prefix is not shown here to make it concise) and the key counts:
```
[65 66 39 61 - 66 32 39 33] : 6816 keys
[65 63 36 63 - 66 31 36 36] : 4864 keys
[65 66 32 65 - 65 66 37 61] : 1401 keys
[65 61 30 39 - 65 63 65 65] : 740 keys
```
Then we deployed a detector for misordered keys in the compaction and iteration and found 4 shards having the misordered keys among 400k shards.

We also checked the shards still running v5.7.2, and have found a shard with similar key misordering issue. So this problem already existed in the old version. But we did not get the issue reported from the iteration in the v5.7.2 as in the v6.4.6 - that is why we just noticed it after upgrading to v6.4.6.

Then we want to ask:
1. Do some versions of rocksdb have issues potentially causing key misordering in sst files?
2. Is there a change in iterate() behavior from v5.7.2 to v6.4.6?
3. What is the suggestion to avoid the issue in future?

Thanks. We are still collecting more data from our storage, please let us know if you need additional data.

ajkr:
One known possibility is block cache key conflict leading to an old block in cache being returned instead of a new block. That's aligned with seeing an ordered sequence of keys injected in the middle of another ordered sequence of keys. Another known possibility is faulty hardware.

More details on the key conflict possibility: block cache keys are composed of (inode number, inode generation (4 bytes), file offset). In older filesystems (even older versions of ext4) the inode generation is randomly generated. Then if we recycle inode number a bunch of times, which file systems tend to do, we can end up reading a stale block from the cache.

In terms of probability of this happening, imagine there's 2^32 days in a year. The birthday paradox math says the class needs to have ~70K students in order for a birthday conflict to be 50% likely. Similarly, we need to recycle a inode number roughly ~70K times to see an (inode number, inode generation) conflict with 50% probability.

[...]

PR #8360, opened 2021-06-03
Use DbSessionId as cache key prefix when secondary cache is enabled

zhichao-cao:
Currently, we either use the file system inode or a monotonically incrementing runtime ID as the block cache key prefix. However, if we use a monotonically incrementing runtime ID (in the case that the file system does not support inode id generation), in some cases, it cannot ensure uniqueness (e.g., we have secondary cache migrated from host to host). We use DbSessionID (20 bytes) + current file number (at most 10 bytes) as the new cache block key prefix when the secondary cache is enabled. So can accommodate scenarios such as transfer of cache state across hosts.

test plan: add the test to lru_cache_test

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -251,16 +251,18 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
   co.num_shard_bits = immutable_db_options_.table_cache_numshardbits;
   co.metadata_charge_policy = kDontChargeCacheMetadata;
   table_cache_ = NewLRUCache(co);
+  SetDbSessionId();
+  std::string db_session_id;
+  GetDbSessionId(db_session_id);
 
   versions_.reset(new VersionSet(dbname_, &immutable_db_options_, file_options_,
                                  table_cache_.get(), write_buffer_manager_,
                                  &write_controller_, &block_cache_tracer_,
-                                 io_tracer_));
+                                 io_tracer_, db_session_id));
   column_family_memtables_.reset(
       new ColumnFamilyMemTablesImpl(versions_->GetColumnFamilySet()));
 
   DumpRocksDBBuildVersion(immutable_db_options_.info_log.get());
-  SetDbSessionId();
   DumpDBFileSummary(immutable_db_options_, dbname_, db_session_id_);
   immutable_db_options_.Dump(immutable_db_options_.info_log.get());
   mutable_db_options_.Dump(immutable_db_options_.info_log.get());
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -252,13 +252,12 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
   co.metadata_charge_policy = kDontChargeCacheMetadata;
   table_cache_ = NewLRUCache(co);
   SetDbSessionId();
-  std::string db_session_id;
-  GetDbSessionId(db_session_id);
+  assert(!db_session_id_.empty());
 
   versions_.reset(new VersionSet(dbname_, &immutable_db_options_, file_options_,
                                  table_cache_.get(), write_buffer_manager_,
                                  &write_controller_, &block_cache_tracer_,
-                                 io_tracer_, db_session_id));
+                                 io_tracer_, db_session_id_));
   column_family_memtables_.reset(
       new ColumnFamilyMemTablesImpl(versions_->GetColumnFamilySet()));
 
--- a/db/repair.cc
+++ b/db/repair.cc
@@ -109,14 +109,16 @@ class Repairer {
             // TableCache can be small since we expect each table to be opened
             // once.
             NewLRUCache(10, db_options_.table_cache_numshardbits)),
-        table_cache_(new TableCache(
-            default_iopts_, env_options_, raw_table_cache_.get(),
-            /*block_cache_tracer=*/nullptr, /*io_tracer=*/nullptr)),
+        table_cache_(
+            new TableCache(default_iopts_, env_options_, raw_table_cache_.get(),
+                           /*block_cache_tracer=*/nullptr,
+                           /*io_tracer=*/nullptr, /*db_session_id*/ "")),
         wb_(db_options_.db_write_buffer_size),
         wc_(db_options_.delayed_write_rate),
         vset_(dbname_, &immutable_db_options_, env_options_,
               raw_table_cache_.get(), &wb_, &wc_,
-              /*block_cache_tracer=*/nullptr, /*io_tracer=*/nullptr),
+              /*block_cache_tracer=*/nullptr, /*io_tracer=*/nullptr,
+              /*db_session_id*/ ""),
         next_file_number_(1),
         db_lock_(nullptr),
         closed_(false) {
--- a/db/repair.cc
+++ b/db/repair.cc
@@ -110,11 +110,15 @@ class Repairer {
             // once.
             NewLRUCache(10, db_options_.table_cache_numshardbits)),
         table_cache_(
+            // TODO: db_session_id for TableCache should be initialized after
+            // db_session_id_ is set.
             new TableCache(default_iopts_, env_options_, raw_table_cache_.get(),
                            /*block_cache_tracer=*/nullptr,
                            /*io_tracer=*/nullptr, /*db_session_id*/ "")),
         wb_(db_options_.db_write_buffer_size),
         wc_(db_options_.delayed_write_rate),
+        // TODO: db_session_id for VersionSet should be initialized after
+        // db_session_id_ is set and use it for initialization.
         vset_(dbname_, &immutable_db_options_, env_options_,
               raw_table_cache_.get(), &wb_, &wc_,
               /*block_cache_tracer=*/nullptr, /*io_tracer=*/nullptr,
@@ -450,7 +454,8 @@ class Repairer {
           -1 /* level */, false /* is_bottommost */,
           TableFileCreationReason::kRecovery, current_time,
           0 /* oldest_key_time */, 0 /* file_creation_time */,
-          "DB Repairer" /* db_id */, db_session_id_, 0 /*target_file_size*/);
+          "DB Repairer" /* db_id */, db_session_id_, 0 /*target_file_size*/,
+          meta.fd.GetNumber());
       status = BuildTable(
           dbname_, /* versions */ nullptr, immutable_db_options_, tboptions,
           env_options_, table_cache_.get(), iter.get(),
--- a/include/rocksdb/cache.h
+++ b/include/rocksdb/cache.h
@@ -485,6 +485,9 @@ class Cache {
   // should check the Value() of each handle for nullptr
   virtual void WaitAll(std::vector<Handle*>& /*handles*/) {}
 
+  // Determines if if the secondary cache is enabled.
+  virtual bool IsSecondaryCacheEnabled() { return false; }
+
  private:
   std::shared_ptr<MemoryAllocator> memory_allocator_;
 };
--- a/include/rocksdb/cache.h
+++ b/include/rocksdb/cache.h
@@ -485,9 +485,6 @@ class Cache {
   // should check the Value() of each handle for nullptr
   virtual void WaitAll(std::vector<Handle*>& /*handles*/) {}
 
-  // Determines if if the secondary cache is enabled.
-  virtual bool IsSecondaryCacheEnabled() { return false; }
-
  private:
   std::shared_ptr<MemoryAllocator> memory_allocator_;
 };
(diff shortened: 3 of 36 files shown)


pdillinger:
This seems like a temporary change (unnecessary?) while working toward the more permanent solution of unique identifiers for SST files. See https://github.com/facebook/rocksdb/issues/7405#issuecomment-698567230

By the way, you want "monotonically" increasing/incrementing not "monotonously" (easy mistake I've seen others make)

pdillinger:
Actually, I have a proposal to make this fit in nicely with my plan. The simple version of my plan is to combine DB session ID and file number into a single 128-bit unique identifier for an SST file. (Some supporting work is in #8362.) One complication is that not all SST files have embedded DB session IDs. A bigger complication is that the SST file itself doesn't know its own *original* file number. The file name has the file number, but (IIRC) that is changed if the file is ingested to another DB. Thus, for my plan to work for *stable* unique SST identifiers, we need to add the original file number to table properties.

But also we need some safe identifier for SST files that don't have the necessary table properties for computing the stable unique identifier. That's essentially what this PR is (close to) solving. And I propose that the safe identifier be the combination of *current* DB session ID and *current* file number. This is nice because it means we have just one schema for unique SST identifiers (combine session ID and file number), but we safely use "current" values if "original" are unavailable.

That proposed approach fixes at least one problem with this PR. Some configurations will re-open table files during their lifetime and within a single DB::Open session, and we want to allow block cache hits between table file re-open. (The current use of NewId() in this PR breaks that.) However, compared to existing cache keys, this approach still regresses in being able to share cache entries between DB instances sharing table files (including close+re-Open). I think we can accept that for now when secondary cache is enabled. 

So my proposal for this PR is to plumb through the current file number into the table reader so that it can be used in place of NewId() in building the cache key (for now, when secondary cache is enabled). Then in later work we can
* Optimize the encoding down to 16 bytes for combine(Session ID, file number), down from ~23 used in this PR.
* Include original file number in table properties
* Use original Session ID and file number (instead of current) when both are available in table properties, but still only use this schema when secondary cache is enabled.
* After a few releases with original file numbers in table properties, make this the new standard cache key schema, because almost all table files will have a stable unique identifier.

pdillinger:
> * Optimize the encoding down to 16 bytes for combine(Session ID, file number)

One integration detail I haven't discussed: block cache users other than table readers. We could have a reserved value for one of the 64-bit pieces, such as all zeros, that is never used in the combine() encoding. This leaves the other 64-bit piece for populating with NewId().

As for the remaining details of this combine() part, I'd prefer to explain it in code (soon).

pdillinger (changes requested):
See proposal in comment ("plumb through the current file number into the table reader so that it can be used in place of NewId() in building the cache key")

db/db_impl/db_impl.cc
```diff
@@ -251,16 +251,18 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
   co.num_shard_bits = immutable_db_options_.table_cache_numshardbits;
   co.metadata_charge_policy = kDontChargeCacheMetadata;
   table_cache_ = NewLRUCache(co);
+  SetDbSessionId();
+  std::string db_session_id;
```
pdillinger, line 255: There should be no need to GetDbSessionId. You can just use `db_session_id_`. Assert not empty if you want.

db/repair.cc
```diff
@@ -109,14 +109,16 @@ class Repairer {
             // TableCache can be small since we expect each table to be opened
             // once.
             NewLRUCache(10, db_options_.table_cache_numshardbits)),
-        table_cache_(new TableCache(
-            default_iopts_, env_options_, raw_table_cache_.get(),
-            /*block_cache_tracer=*/nullptr, /*io_tracer=*/nullptr)),
+        table_cache_(
+            new TableCache(default_iopts_, env_options_, raw_table_cache_.get(),
+                           /*block_cache_tracer=*/nullptr,
+                           /*io_tracer=*/nullptr, /*db_session_id*/ "")),
```
pdillinger, line 115: I think for this table_cache_ and vset_ we should be using the same db_session_id_ used elsewhere in Repairer. I'm not sure what could go wrong, but it seems like the right thing to do.

table_cache_ could probably be initialized after db_session_id_ is set, but vset_ might need to become a pointer for that to work. Alternatively, you can put a TODO item here, and I will fix it in #8362 when it will be easier to generate a new session ID without opening a DBImpl.

table/block_based/block_based_table_reader.h
```diff
@@ -446,12 +448,25 @@ class BlockBasedTable : public TableReader {
...
+    if (cc != nullptr && cc->IsSecondaryCacheEnabled() &&
+        db_session_id.size() != 0) {
+      // db_session_id is 20 bytes as defined.
+      assert(db_session_id.size() == 20);
+      memcpy(buffer, db_session_id.c_str(), 20);
+      char* end = EncodeVarint64(buffer + 20, cur_file_num);
+      // kMaxVarint64Length is 10 therefore, the prefix is at most 30 bytes.
+      *size = static_cast<size_t>(end - buffer);
+      return;
+    }
```
pdillinger, line 469: Better to use 'else' here rather than early 'return'. I only recommend early return for handling trivial or exceptional cases.

table/table_builder.h
```diff
@@ -37,12 +37,14 @@ struct TableReaderOptions {
...
                      bool _force_direct_prefetch = false, int _level = -1,
                      BlockCacheTracer* const _block_cache_tracer = nullptr,
-                     size_t _max_file_size_for_l0_meta_pin = 0)
-      : TableReaderOptions(_ioptions, _prefix_extractor, _env_options,
-                           _internal_comparator, _skip_filters, _immortal,
-                           _force_direct_prefetch, _level,
-                           0 /* _largest_seqno */, _block_cache_tracer,
-                           _max_file_size_for_l0_meta_pin) {}
+                     size_t _max_file_size_for_l0_meta_pin = 0,
+                     const std::string& _db_session_id = "",
```
pdillinger, line 41: When we get to a context associated with a single table file (such as this whole file), we should call it `cur_db_session_id` rather than `db_session_id` because it's the id of the running DB, not the id in the table file properties from the DB that generated it, which might be different.

pdillinger (approved):
Some minor recommendations

include/rocksdb/cache.h
```diff
@@ -485,6 +485,9 @@ class Cache {
   // should check the Value() of each handle for nullptr
   virtual void WaitAll(std::vector<Handle*>& /*handles*/) {}
 
+  // Determines if if the secondary cache is enabled.
+  virtual bool IsSecondaryCacheEnabled() { return false; }
```
anand1976, line 489: This could be a problem if we make the secondary cache configuration dynamic, i.e allow the user to add/remove a secondary cache. In the future, we might at least want to allow the user to remove it in order to ensure a clean shutdown of the secondary cache.

Maybe we can have an explicit option in LRUCacheOptions to allow the user to request that stronger cache keys be used.

table/block_based/block_based_table_builder.cc
```diff
@@ -873,7 +873,7 @@ BlockBasedTableBuilder::BlockBasedTableBuilder(
     BlockBasedTable::GenerateCachePrefix<Cache, FSWritableFile>(
         table_options.block_cache_compressed.get(), file->writable_file(),
         &rep_->compressed_cache_key_prefix[0],
-        &rep_->compressed_cache_key_prefix_size);
+        &rep_->compressed_cache_key_prefix_size, "", 0);
```
anand1976, line 876: This could be a problem for cache warming. Maybe not in this PR, but we should address it soon. I believe the SST file number is assigned by the CompactionJob when opening the file for writing, so shouldn't be too difficult.

db/repair.cc
zhichao-cao: TODO is added.

table/table_builder.h
zhichao-cao: Sure, will update it where it is associated with the table file.

table/block_based/block_based_table_builder.cc
zhichao-cao: Good point! Will add it. I think I missed the cache warming case.

include/rocksdb/cache.h
pdillinger: If you want to get rid of this function (good idea I think), then I think the proposed option needs to be at the DB or CF level, because the Cache might not be LRUCache. I would suggest that the option be something like `use_persistable_cache_keys` which would mean that at least for secondary cache keys applicable to secondary cache, use keys that will not cause correctness issue across DB close+re-open, including if the DB and/or secondary cache are physically migrated.

pdillinger: And if `use_persistable_cache_keys` is set to false, then we should probably disable secondary cache (in how we call into Cache object).

zhichao-cao: > And if `use_persistable_cache_keys` is set to false, then we should probably disable secondary cache (in how we call into Cache object).

It is not straight forward for user to understand the behavior: if `use_persistable_cache_keys` is set to false, then secondary cache is not applicable. How about directly have an option in DBOption `use_secondary_cache`? @anand1976 what do you think?

pdillinger: Would it solve the immediate problem to simply use the new kind of cache key when file->GetUniqueId returns empty?

pdillinger: I believe the hypothetical problem of a file unique id collision (DB close; files moved around; DB re-open) leading to a bad cache key collision is similar to the existing problem of a file unique id collision between a deleted file with stale cache entries and a new file. We know we need to solve the latter by completely migrating away from file->GetUniqueId. So IMHO the situation is not substantially worse in the interim by introducing more file unique id collision possibilities (on FileSystems providing it) through a persistent secondary cache.

zhichao-cao: Should work for secondary cache if we just replace NewId() with the DbSessionId+cur_file_num

pdillinger (approved):
Thanks! LGTM

ajkr:
@pdillinger This is a great plan! I had no idea how we were going to fix the cache key conflict problem while preserving the ability to continue using cached blocks post-ingestion and post-DB reopen.

> Actually, I have a proposal to make this fit in nicely with my plan. The simple version of my plan is to combine DB session ID and file number into a single 128-bit unique identifier for an SST file. (Some supporting work is in #8362.) One complication is that not all SST files have embedded DB session IDs. A bigger complication is that the SST file itself doesn't know its own _original_ file number. The file name has the file number, but (IIRC) that is changed if the file is ingested to another DB. Thus, for my plan to work for _stable_ unique SST identifiers, we need to add the original file number to table properties.
> 

I thought there is no file number for `SstFileWriter`-written files. Is the idea that it doesn't matter - we can just write zero in the original file number property - since the session ID alone gives enough uniqueness for such files?

> But also we need some safe identifier for SST files that don't have the necessary table properties for computing the stable unique identifier. That's essentially what this PR is (close to) solving. And I propose that the safe identifier be the combination of _current_ DB session ID and _current_ file number. This is nice because it means we have just one schema for unique SST identifiers (combine session ID and file number), but we safely use "current" values if "original" are unavailable.
> 
> That proposed approach fixes at least one problem with this PR. Some configurations will re-open table files during their lifetime and within a single DB::Open session, and we want to allow block cache hits between table file re-open. (The current use of NewId() in this PR breaks that.) However, compared to existing cache keys, this approach still regresses in being able to share cache entries between DB instances sharing table files (including close+re-Open). I think we can accept that for now when secondary cache is enabled.
> 

Random observation - the case of cached blocks surviving in block cache through a table reopen seems odd. In terms of block read count it feels like we'd be better off if the properties/footer survived, while I think those aren't cached.

> So my proposal for this PR is to plumb through the current file number into the table reader so that it can be used in place of NewId() in building the cache key (for now, when secondary cache is enabled). Then in later work we can
> 
> * Optimize the encoding down to 16 bytes for combine(Session ID, file number), down from ~23 used in this PR.
> * Include original file number in table properties
> * Use original Session ID and file number (instead of current) when both are available in table properties, but still only use this schema when secondary cache is enabled.
> * After a few releases with original file numbers in table properties, make this the new standard cache key schema, because almost all table files will have a stable unique identifier.

pdillinger:
> I thought there is no file number for `SstFileWriter`-written files. Is the idea that it doesn't matter - we can just write zero in the original file number property - since the session ID alone gives enough uniqueness for such files?

Yes we can do that or it's slightly better to use a session for each writer and incrementing file numbers.

But I wonder if we need to rewrite these in the file if the SST file is modified on ingestion. You know more about that than I do.

> Random observation - the case of cached blocks surviving in block cache through a table reopen seems odd. In terms of block read count it feels like we'd be better off if the properties/footer survived, while I think those aren't cached.

We support limited max open files, and I am thinking we might soon want to force reopen on trivial move to update state of the table reader without it being mutable (inefficient) state or recomputed on each Get.

PR #8361, opened 2021-06-03
Re-enable recycle_log_file_num with kPointInTimeRecovery.

shoda-tibco:
Add new unit test to verify a recovery hole is not created with
recycle_log_file_num and kPointInTimeRecovery.

Re-enable FullPurge unit tests.

Add caching of synced_file_bytes to Writer to prevent null pointer
access when track_and_verify_wals_in_manifest and recycle_log_file_num
are enabled.


db/db_wal_test.cc
```diff
@@ -1565,6 +1561,65 @@ TEST_P(DBWALTestWithParams, kSkipAnyCorruptedRecords) {
...
+  std::string fname = LogFileName(dbname_, wal_file_id);
+  uint64_t offset_to_corrupt;
+  ASSERT_OK(env->GetFileSize(fname, &offset_to_corrupt));
+  ASSERT_GT(offset_to_corrupt, 0);
+  ASSERT_OK(Put(0, "key1", "wal1"));
+  ASSERT_OK(Flush(1));
+  ASSERT_OK(Put(0, "key2", "wal2"));
+
+  // Corrupt WAL1 at location of key1
+  ASSERT_OK(test::CorruptFile(env, fname, static_cast<int>(offset_to_corrupt),
```
ajkr, line 1603: This is probably corrupting the WAL fragment header for key1. What if we corrupt one byte of the value? For example take the file length after putting "key1" -> "wal1" and corrupt the final byte. I think there's no WAL fragment trailer.

shoda-tibco: I updated the unit test to corrupt the end byte and a middle byte in addition to the first byte of the record. Still, it does not seem to permit a hole in recovery.

shoda-tibco:
Rebased and fixed TSAN errors, please have a look @ajkr.

ajkr:
Sorry for the delay. I do plan to have another look; it'll just be a while (1+ months) since this is unused in FB and has relatively high risk for subtle bugs.

PR #8362, opened 2021-06-03
Add API for UUIDs, minor fix

pdillinger:
Summary: Our current support for unique IDs is limited and not well
defined. This change primarily adds some public (and internal) APIs
for more detailed and expanded support for unique IDs.

Bug fix is for Env::GenerateUniqueId() on posix environments without
/proc/sys/kernel/random/uuid. In such cases, it was previously
relatively easy to get duplicate ids if calling from multiple threads
around the same time. The added unit test would often fail on the old
implementation if the /proc-based code was disabled. Also, based on
unit test, the new implementation is much faster than the /proc-based
implementation, at least after first call for a given thread.

New public API structs:

`RawUuid`: a holder for 128-bit identifiers. I plan to use this (or some
derived struct) to assign unique ids to SST files. There is no
standardized text representation, as that is left to derived classes.
A reason for exposing this in the public API is to have a general way
for Env to generate 128-bit unique identifiers, from which other
specific kinds can be derived.

`RfcUuid`: an RFC 4122 UUID, derived from RawUuid. This is what we currently
use for DB IDs, and can be generated from a RawUuid by dropping six bits
of entropy. One reason justifying the existence of RawUuid is to avoid
dropping entropy unnecessarily AND to decorate types with the kind of
data they hold.

New internal API struct:

`RocksMuid`: an identifier representable with 20 base-36 digits, derived
from RawUuid and used for DB session IDs.

New Env functions:
virtual RawUuid GenerateRawUuid(): This is basically asking for a random
128 bits from Env. I have added a standard implementation based on other
C++ and Env APIs, with enough redundancy to still provide high quality
results even when some APIs are of degraded or zero quality. I claim
that, "The possibility of duplicates is generally lower than the
possibility of computing the wrong thing due to random hardware errors."

virtual RfcUuid GenerateRfcUuid(): Although you could just call
`RfcUuid::FromRawLoseData(GenerateRawUuid())`, GenerateRfcUuid() is
separate in Env because some might want access to a native
UUID-providing API (e.g. on Windows) that might have stronger
properties (which should not matter for our purposes, but OK).

In the API doc for existing Env::GenerateUniqueId(), clarified that
it currently returns an RFC 4122 UUID string, with typical
implementation returning GenerateRfcUuid().ToString(). IMHO, this
function is only kept for compatibility.

Follow-up work:
* Unique ids for SST files, first internally then exposing in public
API.
* Consider making DB IDs and session IDs use specific uid type rather
than std::string.

Test Plan: unit tests added for Env and other key functionality.
manually tested that each entropy "track" in GenerateRawUuid() suffices
to pass tests.


include/rocksdb/uuid.h
```diff
@@ -0,0 +1,117 @@
...
+  inline bool IsEmpty() const { return lower64 == 0 && upper64 == 0; }
+
+  // For saving into a machine-independent 16-byte string. The written
+  // characters can contain nul bytes.
+  std::string ToRawBytes() const;
+  // For saving into a machine-independent 16-byte string, into an
+  // existing buffer. No trailing nul is written or expected, but the
+  // written characters can contain nul bytes. Returns a Slice for the
+  // buffer and 16 written characters.
+  Slice PutRawBytes(char *buf_of_16) const;
```
mrambacher, line 37: Please better explain the arguments and return value.  From the signature, I presume the Slice is mapped over buf_of_16, but it would be nice to explicitly state that.

include/rocksdb/uuid.h
```diff
@@ -0,0 +1,117 @@
...
+// empty value when a value is actually generated.
+struct RawUuid {
+  //== Data
+  uint64_t lower64 = 0;
+  uint64_t upper64 = 0;
+
+  //== Construction
+  RawUuid() { assert(IsEmpty()); }
+
+  static RawUuid FromRawBytes(const char *buf_of_16);
```
mrambacher, line 25: Why is this not the same signature as RfcUuid::Parse?  It seems like they should be equivalent.

env/env.cc
```diff
@@ -693,6 +696,81 @@ Status Env::GetHostNameString(std::string* result) {
   return s;
 }
 
+namespace {
+
+// For GenerateRawUuid()
+struct RawUuidHelper {
+  // All the data we want to go into the initial random seed. For safety in
```
mrambacher, line 703: This is not really related to the UUID but is more a random number generator.  Can we somehow unify this with the Random class and create one implementation of Random (or make it a virtual method off Env)?

env/env.cc
```diff
@@ -693,6 +696,81 @@ Status Env::GetHostNameString(std::string* result) {
...
+    // Mix it all together (nom nom)
+    state = XXH128(&init, sizeof(init), /*seed*/ 123U);
+  }
+
+  XXH128_hash_t state;
+};
+
+}  // namespace
+
+RawUuid Env::GenerateRawUuid() {
```
mrambacher, line 746: I am a little confused as to if this code belongs here or is part of the Env Posix class (it looks like Windows may have a different implementation)?

env/generate_uuid.cc
```diff
@@ -0,0 +1,129 @@
...
+  XXH128_hash_t state = XXH128(&init, sizeof(init), /*seed*/ 123U);
+
+  return (Unsigned128{state.high64} << 64) | state.low64;
+}
+
+const RawUuid& GetRawUuidForCurrentProcess() {
+  // Lazy initialized into static
+  // BUG/FIXME: does not change on fork()
+  static const RawUuid uuid = GenerateRandomRawUuidExpensive(Env::Default());
+  return uuid;
```
mrambacher, line 72: Is there a race condition here if two threads attempt this call simultaneously?

The name of the API is kind of off/confusing as well, as none of the entropy is related to the "process" per-se (except perhaps the thread id).

env/generate_uuid.cc
```diff
@@ -0,0 +1,129 @@
...
+}
+
+RawUuid GenerateRawUuid() {
+  static std::atomic<uint64_t> counter{0};
+
+  return AddInCounterA(GetRawUuidForCurrentProcess(), ++counter);
+}
+
+Status GenerateRfcUuidFromPlatform(RfcUuid* out) {
```
mrambacher, line 81: This would be better as a member of the Env or FileSystem class...

util/uuid.h
```diff
@@ -0,0 +1,150 @@
...
+  return lhs.scaled_data == rhs.scaled_data;
+}
+
+RawUuid AddInCounterA(const RawUuid &u, uint64_t counter);
+RawUuid AddInCounterB(const RawUuid &u, uint64_t counter);
+RawUuid AddInCounterC(const RawUuid &u, uint64_t counter);
+
+RawUuid SubtractOffCounterA(const RawUuid &u, uint64_t counter);
+RawUuid SubtractOffCounterB(const RawUuid &u, uint64_t counter);
+RawUuid SubtractOffCounterC(const RawUuid &u, uint64_t counter);
```
mrambacher, line 132: I know these are private and in an internal, non-public class but I have no idea when I would use the CounterA vs CounterB/C methods.  Perhaps some comments somewhere?

util/uuid.h
```diff
@@ -0,0 +1,150 @@
...
+// random RocksMuid every minute for a year, or one computer generates a
+// million every second for a week, the chance of a RocksMuid being repeated
+// in that time is 1 in millions. See Env::GenerateRawUuid() and
+// FromRawLoseData().
+//
+// In text form, the digits are in standard written order: lowest-order last,
+// except each piece (lower64 and upper64) only holds 10 base-36 digits.
+// Thus, each uint64_t piece is < kPieceMod. FIXME
+struct RocksMuid {
+  RawUuid scaled_data = 0;
```
mrambacher, line 87: Do you mean the scaled_data to be public?  Should users have free access to it and be able to change it?

util/uuid.h
```diff
@@ -0,0 +1,150 @@
...
+
+// An RFC 4122 Universally Unique IDentifier. Variant 1 version 4 is the
+// most common, which offers 122 bits of entropy. The text form is 32
+// hexadecimal digits and four hyphens, for a total of 36 characters.
+// https://en.wikipedia.org/wiki/Universally_unique_identifier
+//
+// This is considered suitable for identifying medium-high frequency events.
+// For example, if 1 million computers each generate 10 random RfcUuids every
+// second for a year, the chance of one being repeated in that time is 1 in
+// millions. See Env::GenerateRfcUuid().
```
mrambacher, line 24: I do not think there is such a method as Env::GenerateRfcUuid

util/uuid.h
```diff
@@ -0,0 +1,150 @@
...
+// This provides a compact, portable text representation with 103.4 bits of
+// entropy. This was originally developed for DB Session IDs (see
+// DB::GetDbSessionId(). All zeros means unset or empty.
+//
+// This is considered suitable for identifying medium-frequency events or
+// narrow scope events. For example, if 1 million computers each generate a
+// random RocksMuid every minute for a year, or one computer generates a
+// million every second for a week, the chance of a RocksMuid being repeated
+// in that time is 1 in millions. See Env::GenerateRawUuid() and
+// FromRawLoseData().
```
mrambacher, line 81: Wrong method comment?

util/uuid.h
```diff
@@ -0,0 +1,150 @@
...
+struct RocksMuid {
+  RawUuid scaled_data = 0;
+
+  // Parses from the 20-digit base-36 text form. Returns Corruption if
+  // malformed.
+  static Status Parse(const Slice &input, RocksMuid *out);
+
+  // This converts a RawUuid (also OK: RfcUuid) to a RocksMuid by
+  // FIXME on both 64-bit pieces. Assuming the input is random with
+  // 128 bits of entropy, about 24.6 bits of entropy are lost in the
```
mrambacher, line 95: Update the comment?

util/uuid.h
```diff
@@ -0,0 +1,150 @@
...
+// This is considered suitable for identifying medium-frequency events or
+// narrow scope events. For example, if 1 million computers each generate a
+// random RocksMuid every minute for a year, or one computer generates a
+// million every second for a week, the chance of a RocksMuid being repeated
+// in that time is 1 in millions. See Env::GenerateRawUuid() and
+// FromRawLoseData().
+//
+// In text form, the digits are in standard written order: lowest-order last,
+// except each piece (lower64 and upper64) only holds 10 base-36 digits.
+// Thus, each uint64_t piece is < kPieceMod. FIXME
```
mrambacher, line 85: FIXME?

util/hash_test.cc
```diff
@@ -584,6 +585,65 @@ TEST(MathTest, CodingGeneric) {
...
+        }
+        for (size_t i = 0; i < count; ++i) {
+          size_t pos;
+          if (approach == kAddition) {
+            pos = (start + i) & kMask;
+          } else {
+            pos = start ^ i;
+          }
+          if (v[pos]) {
+            goto next_iter;
```
mrambacher, line 626: Can we do this test without a goto?  I have not seen a goto since Fortran and would rather keep it that way...

pdillinger:
I didn't mean to push this for review, sorry

PR #8364, opened 2021-06-04
Add sub-compaction support for RemoteCompaction

jay-zhuang:
Change the job_id for remote compaction interface, which will include
both internal compaction job_id, also a sub_compaction_job_id. It is not
a backward compatible change. The user needs to update interface during
upgrade. (We will avoid backward incompatible change after the feature is
not experimental.)


ajkr (approved):
LGTM, sorry for the delay.

PR #8369, opened 2021-06-08
Fix a major performance bug in 6.21 for cache entry stats

pdillinger:
Summary: In final polishing of #8297 (after most manual testing), I
broke my own caching layer by sanitizing an input parameter with
std::min(0, x) instead of std::max(0, x). I resisted unit testing the
timing part of the result caching because historically, these test
are either flaky or difficult to write, and this was not a correctness
issue. This bug is essentially unnoticeable with a small number
of column families but can explode background work with a
large number of column families.

This change fixes the logical error, removes some unnecessary related
optimization, and adds mock time/sleeps to the unit test to ensure we
can cache hit within the age limit.

Test Plan: added time testing logic to existing unit test

### DIFF
--- a/cache/cache_entry_stats.h
+++ b/cache/cache_entry_stats.h
@@ -56,15 +56,11 @@ class CacheEntryStatsCollector {
     // Waits for any pending reader or writer (collector)
     std::lock_guard<std::mutex> lock(mutex_);
 
-    // Maximum allowed age is nominally given by the parameter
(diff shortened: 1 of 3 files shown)


pdillinger:
I'm assuming this doesn't need a HISTORY.md entry in master since it will be back-ported to 6.21, where the bug was introduced.

ajkr (approved):
LGTM.

PR #8373, opened 2021-06-08
Add Inner method and support for wrapped Customizable

mrambacher:
Inner method added for classes to override to return has-a relationship.  CheckedCast expanded to use Inner to return has-a relationship.

Future Customizable classes (Env, FileSystem, Statistics, etc) will use this feature


pdillinger (approved):
LGTM

PR #8376, opened 2021-06-09
All the NoSpace() errors will be handled by regular SetBGError and RecoverFromNoSpace()

zhichao-cao:
In the current logic, any IO Error with retryable flag == true will be handled by the special logic and in most cases, StartRecoverFromRetryableBGIOError will be called to do the auto resume. If the NoSpace error with retryable flag is set during WAL write, it is mapped as a hard error, which will trigger the auto recovery. During the recover process, if write continues and append to the WAL, the write process sees that bg_error is set to HardError and it calls WriteStatusCheck(), which calls SetBGError() with Status (not IOStatus). This will redirect to the regular SetBGError interface, in which recovery_error_ will be set to the corresponding error. With the recovery_error_ set, the auto resume thread created in StartRecoverFromRetryableBGIOError will keep failing as long as user keeps trying to write.

To fix this issue. All the NoSpace error (no matter retryable flag is set or not) will be redirect to the regular SetBGError, and RecoverFromNoSpace() will do the recovery job which calls SstFileManager::StartErrorRecovery().

test plan: make check and added the new testing case

### DIFF
--- a/db/error_handler.cc
+++ b/db/error_handler.cc
@@ -418,15 +418,18 @@ const Status& ErrorHandler::SetBGError(const IOStatus& bg_io_err,
                                           &bg_err, db_mutex_, &auto_recovery);
     recover_context_ = context;
     return bg_error_;
-  } else if (bg_io_err.GetScope() ==
-                 IOStatus::IOErrorScope::kIOErrorScopeFile ||
-             bg_io_err.GetRetryable()) {
+  } else if (bg_io_err.subcode() != IOStatus::SubCode::kNoSpace &&
+             (bg_io_err.GetScope() ==
+                  IOStatus::IOErrorScope::kIOErrorScopeFile ||
+              bg_io_err.GetRetryable())) {
     // Second, check if the error is a retryable IO error (file scope IO error
     // is also treated as retryable IO error in RocksDB write path). if it is
     // retryable error and its severity is higher than bg_error_, overwrite the
     // bg_error_ with new error. In current stage, for retryable IO error of
     // compaction, treat it as soft error. In other cases, treat the retryable
-    // IO error as hard error.
+    // IO error as hard error. Note that, all the NoSpace error should be
+    // handled by the SstFileManager::StartErrorRecovery(). Therefore, no matter
+    // it is retryable or file scope, this logic will be bypassed.
     bool auto_recovery = false;
     EventHelpers::NotifyOnBackgroundError(db_options_.listeners, reason,
                                           &new_bg_io_err, db_mutex_,
@@ -498,8 +501,7 @@ Status ErrorHandler::OverrideNoSpaceError(const Status& bg_error,
     return bg_error;
   }
 
-  if (db_options_.sst_file_manager.get() == nullptr ||
-      db_->env_ == Env::Default()) {
+  if (db_options_.sst_file_manager.get() == nullptr) {
     // We rely on SFM to poll for enough disk space and recover
     *auto_recovery = false;
     return bg_error;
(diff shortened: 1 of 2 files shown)


mrambacher:
I am not sure I understand this change.  What if the EncryptedEnv is used, which wraps a Default one.  Why would it be different?  What about other Env/FileSystems that use/wrap the base ones?

Also, the title says "PosixEnv" but this also applies to WinEnv.  Is it intentional?

Can you also add a comment in the code as to why this change is necessary and what problem it is addressing?

anand1976:
@zhichao-cao The issue is Status::NoSpace() errors from PosixEnv are currently handled in https://github.com/facebook/rocksdb/blob/master/db/error_handler.cc#L380 because it has the retryable flag set in IOStatus, rather than the old SetBGError() https://github.com/facebook/rocksdb/blob/master/db/error_handler.cc#L272, which has some special case handling for out of space errors.

zhichao-cao:
> @zhichao-cao The issue is Status::NoSpace() errors from PosixEnv are currently handled in https://github.com/facebook/rocksdb/blob/master/db/error_handler.cc#L380 because it has the retryable flag set in IOStatus, rather than the old SetBGError() https://github.com/facebook/rocksdb/blob/master/db/error_handler.cc#L272, which has some special case handling for out of space errors.

I see. Will change accordingly.

zhichao-cao:
> I am not sure I understand this change. What if the EncryptedEnv is used, which wraps a Default one. Why would it be different? What about other Env/FileSystems that use/wrap the base ones?
> 
> Also, the title says "PosixEnv" but this also applies to WinEnv. Is it intentional?
> 
> Can you also add a comment in the code as to why this change is necessary and what problem it is addressing?

The PR is updated, the previous change was a wrong fix remove all the title, descriptions, and the code.

db/error_handler.cc
```diff
@@ -418,15 +418,18 @@ const Status& ErrorHandler::SetBGError(const IOStatus& bg_io_err,
                                           &bg_err, db_mutex_, &auto_recovery);
     recover_context_ = context;
     return bg_error_;
-  } else if (bg_io_err.GetScope() ==
-                 IOStatus::IOErrorScope::kIOErrorScopeFile ||
-             bg_io_err.GetRetryable()) {
+  } else if (bg_io_err.subcode() != IOStatus::SubCode::kNoSpace &&
```
anand1976, line 421: Interesting. Just realized that the ```==``` operator of ```Status``` and ```IOStatus``` only checks code and not subcode. So is there a potential bug in line 330?

anand1976 (commented):
Thanks for the fix

anand1976 (approved):
The change in this PR looks good. The potential bug I mentioned in the comment needs to be looked at, but its most likely benign since NoSpace() is the only error that's marked as a soft error and retried in the old ```SetBGError```.

db/error_handler.cc
zhichao-cao: I will check line 330. I just realized the == operator only compares the code not subcode when did the implementation

zhichao-cao:
> The change in this PR looks good. The potential bug I mentioned in the comment needs to be looked at, but its most likely benign since NoSpace() is the only error that's marked as a soft error and retried in the old `SetBGError`.

I will create another PR to fix it. Some of the testing cases needs to be rewrite.

ajkr:
It's an interesting fix. Just a general question - why can foreground writes set recovery_error_? Can other hard/retryable/file-scoped errors run into the same problem?

PR #8378, opened 2021-06-09
Fixed manifest_dump issues when printing keys and values containing null characters

bjlemaire:
Changed fprintf function to fputc in ApplyVersionEdit, and replaced null characters with whitespaces. 
Added unit test in ldb_test.py - verifies that manifest_dump --verbose output is correct when keys and values containing null characters are inserted.

### DIFF
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -911,7 +911,7 @@ void DumpManifestHandler::CheckIterationResult(const log::Reader& reader,
     assert(cfd->current());
 
     // Print out DebugStrings. Can include non-terminating null characters.
-    fwrite(cfd->current()->DebugString(hex_).c_str(), sizeof(char),
+    fwrite(cfd->current()->DebugString(hex_).data(), sizeof(char),
            cfd->current()->DebugString(hex_).size(), stdout);
   }
   fprintf(stdout,
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -910,14 +910,9 @@ void DumpManifestHandler::CheckIterationResult(const log::Reader& reader,
     }
     assert(cfd->current());
 
-    // Print out DebugStrings. Null characters are replaced with whitespaces.
-    for (const char& x : cfd->current()->DebugString(hex_)) {
-      if (x == '\0') {
-        fputc(' ', stdout);
-      } else {
-        fputc(x, stdout);
-      }
-    }
+    // Print out DebugStrings. Can include non-terminating null characters.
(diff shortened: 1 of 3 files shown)


db/version_edit_handler.cc
```diff
@@ -909,7 +909,15 @@ void DumpManifestHandler::CheckIterationResult(const log::Reader& reader,
       fprintf(stdout, "comparator: %s\n", cfd->user_comparator()->Name());
     }
     assert(cfd->current());
-    fprintf(stdout, "%s \n", cfd->current()->DebugString(hex_).c_str());
+
+    // Print out DebugStrings. Null characters are replaced with whitespaces.
```
pdillinger, line 913: Hmm. If I redirect the output to a file and open it in a hex editor, I expect to see the raw data, not null bytes replaced with spaces.

pdillinger:
You can see CircleCI failed on your test. (The regex does seem so specific that I was afraid it would be fragile.)

pdillinger:
The latest CircleCI failures are unrelated to this change

db/version_edit_handler.cc
```diff
@@ -909,7 +909,10 @@ void DumpManifestHandler::CheckIterationResult(const log::Reader& reader,
       fprintf(stdout, "comparator: %s\n", cfd->user_comparator()->Name());
     }
     assert(cfd->current());
-    fprintf(stdout, "%s \n", cfd->current()->DebugString(hex_).c_str());
+
+    // Print out DebugStrings. Can include non-terminating null characters.
+    fwrite(cfd->current()->DebugString(hex_).c_str(), sizeof(char),
```
pdillinger, line 914: Nit: better to use `.data()` instead of `.c_str()` when you're not using it as a null-terminated string

(obviously 2 more like this)

tools/ldb_test.py
```diff
@@ -473,6 +473,26 @@ def testManifestDump(self):
...
+        self.assertRunOK("put --hex 0xa000000a 0xf000000f", "OK")
+        self.assertRunOK("put a3 b3", "OK")
+        self.assertRunOK("put a4 b4", "OK")
+
+        # Verifies that all "levels" are printed out.
+        # There should be 66 mentions of levels.
+        expected_verbose_output = re.compile("66")
+        # Test manifest_dump verbose and count within the
+        # Terminal how many levels are printed out.
+        cmd_verbose = "manifest_dump --verbose --db=%s | grep -c -E 'level'" %dbPath
```
pdillinger, line 490: Add a comment to mention *why* we resorted to grep

pdillinger (approved):
The test is slightly obscure but gets the job done. 👍

PR #8379, opened 2021-06-09
Clarify some DB::Open,OpenForReadOnly semantics

pdillinger:
Summary: Longstanding tech debt

Test Plan: Better than not having an API contract


jay-zhuang (approved):
👍

PR #8380, opened 2021-06-10
Fix double-dumping CF stats to log

pdillinger:
Summary:
DBImpl::DumpStats is supposed to do this:
Dump DB stats to LOG
For each CF, dump CFStatsNoFileHistogram to LOG
For each CF, dump CFFileHistogram to LOG

Instead, due to a longstanding bug from 2017 (#2126), it would dump
CFStats, which includes both CFStatsNoFileHistogram and CFFileHistogram,
in both loops, resulting in near-duplicate output.

This fixes the bug.

Test Plan: Manual inspection of LOG after db_bench


jay-zhuang (approved):
LGTM👍

PR #8384, opened 2021-06-10
Add DeteleRange support for DBWithTTL

### LINKED ISSUE #7218: Using TtlDB and deleteRange together
I tried to use `deleteRange` for rocksdb that opened with `TtlDB.open`, but it does not work for me. Is there any limitation for using them together?

### Expected behavior
`deleteRange` should delete data

### Actual behavior
`deleteRange` does not delete data when using `TtlDB.open` for opening rocksdb.

Run the following code:

```
public void test() throws RocksDBException {
    Options dbOptions = new Options();
    dbOptions.setCreateIfMissing(true);
    try (RocksDB db = TtlDB.open(dbOptions, "rocksdb")) {
        db.put("1".getBytes(), "value1".getBytes());
        db.put("2".getBytes(), "value2".getBytes());
        db.put("3".getBytes(), "value3".getBytes());
        db.put("4".getBytes(), "value4".getBytes());

        print(db.get("1".getBytes()));
        print(db.get("2".getBytes()));
        print(db.get("3".getBytes()));
        print(db.get("4".getBytes()));

        db.deleteRange("2".getBytes(), "4".getBytes());

        print(db.get("1".getBytes()));
        print(db.get("2".getBytes()));
        print(db.get("3".getBytes()));
        print(db.get("4".getBytes()));
    }
}

private static void print(byte[] bytes) {
    System.out.println(bytes == null ? "null" : new String(bytes));
}
```
Expected output:
```
value1
value2
value3
value4
value1
null
null
value4
```
Actual output:
```
value1
value2
value3
value4
value1
value2
value3
value4
```

KNOEEE:
Summary:
This commit is for enabling `DBWithTTL` to use `DeteleRange` which it cannot before.
As (int32_t)Timestamp is suffixed to values in `DBWithTTL`, there is no reason that it
cannot use the common used api. I added `DeleteRangeCF` in `DBWithTTLImpl::Write` 
so that we can use `DeteleRange` normally. When we run code like 
`dbWithTtl->DeleteRange(start, end)`, it executes`WriteBatchInternal::DeleteRange` 
internally. Intended to fix #7218 

Test Plan: added corresponded testing logic to existing unit test

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,9 @@
 * Added API comments clarifying safe usage of Disable/EnableManualCompaction and EventListener callbacks for compaction.
 * Obsolete keys in the bottommost level that were preserved for a snapshot will now be cleaned upon snapshot release in all cases. This form of compaction (snapshot release triggered compaction) previously had an artificial limitation that multiple tombstones needed to be present.
 
+### New Features
+* Allow `DBWithTTL` to use `DeleteRange` api just like other DBs. `DeleteRangeCF()` which executes `WriteBatchInternal::DeleteRange()` has been added to the handler in `DBWithTTLImpl::Write()` to implement it.
+
 ## 6.21.0 (2021-05-21)
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


KNOEEE:
fix #7218

siying (approved):

PR #8385, opened 2021-06-10
Pin CacheEntryStatsCollector to fix performance bug

pdillinger:
Summary: If the block Cache is full with strict_capacity_limit=false,
then our CacheEntryStatsCollector could be immediately evicted on
release, so iterating through column families with shared block cache
could trigger re-scan for each CF. This change fixes that problem by
pinning the CacheEntryStatsCollector from InternalStats so that it's not
evicted.

I had originally thought that this object could participate in LRU like
everything else, but even though a re-load+re-scan only touches memory,
it can be orders of magnitude more expensive than other cache misses.
One service in Facebook has scans that take ~20s over 100GB block cache
that is mostly 4KB entries. (The up-side of this bug and #8369 is that
we had a natural experiment on the effect on some service metrics even
with block cache scans running continuously in the background--a kind
of worst case scenario. Metrics like latency were not affected enough
to trigger warnings.)

Other smaller fixes:

20s is already a sizable portion of 600s stats dump period, or 180s
default max age to force re-scan, so added logic to ensure that (for
each block cache) we don't spend more than 0.2% of our background thread
time scanning it. Nevertheless, "foreground" requests for cache entry
stats (calls to `db->GetMapProperty(DB::Properties::kBlockCacheEntryStats)`)
are permitted to consume more CPU.

Renamed field to cache_entry_stats_ to match code style.

This change is intended for patching in 6.21 release.

Test Plan: unit test expanded to cover new logic (detect regression),
some manual testing with db_bench

### DIFF
--- a/db/internal_stats.cc
+++ b/db/internal_stats.cc
@@ -528,7 +528,7 @@ InternalStats::InternalStats(int num_levels, SystemClock* clock,
       cfd_(cfd),
       started_at_(clock->NowMicros()) {}
 
-Status InternalStats::CollectCacheEntryStats() {
+Status InternalStats::CollectCacheEntryStats(bool foreground) {
   // Lazy initialize/reference the collector. It is pinned in cache (through
   // a shared_ptr) so that it does not get immediately ejected from a full
   // cache, which would force a re-scan on the next GetStats.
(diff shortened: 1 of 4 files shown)


db/internal_stats.cc
```diff
@@ -528,26 +528,37 @@ InternalStats::InternalStats(int num_levels, SystemClock* clock,
...
-  if (ok) {
-    // Extract or create stats collector.
-    std::shared_ptr<Collector> collector;
-    Status s = Collector::GetShared(block_cache, clock_, &collector);
-    if (s.ok()) {
-      // TODO: use a max age like stats_dump_period_sec / 2, but it's
-      // difficult to access that setting from here with just cfd_
-      collector->GetStats(&cache_entry_stats);
+Status InternalStats::CollectCacheEntryStats(bool foreground) {
+  // Lazy initialize/reference the collector. It is pinned in cache (through
```
ajkr, line 532: Assuming you mean pinned in memory, please correct if I'm mistaken...

I wonder if different `InternalStats`es could have different collectors forever if the initial collector each one creates is quickly ejected.

ajkr (approved):
LGTM, adding Cache::Priority::HIGH and the percentage-based gap between scans seem like improvements. 

I didn't quite understand if the goal is to entirely prevent edge cases of high collection frequency or make them less likely. My current understanding is it does the latter.

ajkr:
Damn, right after submitting my review I remembered about MakeSharedCacheHandleGuard. Ignore my doubts, I agree it should entirely prevent edge cases of high collection frequency.

PR #8393, opened 2021-06-11
Disable subcompactions for user-defined timestamps

ltamasi:
Summary:
The subcompaction boundary picking logic does not currently guarantee
that all user keys that differ only by timestamp get processed by the same
subcompaction. This can cause issues with the `CompactionIterator` state
machine: for instance, one subcompaction that processes a subset of such KVs
might drop a tombstone based on the KVs it sees, while in reality the
tombstone might not have been eligible to be optimized out.
(See also #6645, which adjusted the way compaction inputs are picked for the
same reason.)

Test Plan:
Ran `make check` and the crash test script with timestamps enabled.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,7 @@
 * Obsolete keys in the bottommost level that were preserved for a snapshot will now be cleaned upon snapshot release in all cases. This form of compaction (snapshot release triggered compaction) previously had an artificial limitation that multiple tombstones needed to be present.
 ### Bug Fixes
 * fs_posix.cc GetFreeSpace() always report disk space available to root even when running as non-root.  Linux defaults often have disk mounts with 5 to 10 percent of total space reserved only for root.  Out of space could result for non-root users.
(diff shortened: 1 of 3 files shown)


jay-zhuang (approved):
I think it worth documenting that, maybe here: https://github.com/facebook/rocksdb/blob/80a59a03a7d63a913b7a5558cf6d342533517a33/include/rocksdb/options.h#L615-L621

Or maybe we should just explicitly return unsupported if the user is setting subcompaction number > 1 and timestamp is set.

ltamasi:
> I think it worth documenting that, maybe here:
> 
> https://github.com/facebook/rocksdb/blob/80a59a03a7d63a913b7a5558cf6d342533517a33/include/rocksdb/options.h#L615-L621
> 
> Or maybe we should just explicitly return unsupported if the user is setting subcompaction number > 1 and timestamp is set.

I consider this just a temporary measure since it *is* possible to support subcompactions for user-defined timestamps with some work. So I don't think we should update the API description; we should probably call this out in `HISTORY.md` though.

riversand963:
Thanks @ltamasi for the fix. Could you open a task so that we can track it?

PR #8395, opened 2021-06-11
[RemoteCompaction] Make CompactionService derived from Customizable

hx235:
Summary:
(1)Make CompactionService derived from Customizable by defining two extra functions that are needed, as described in customizable.h comment section
(2)Revise the MyTestCompactionService class in compaction_service_test.cc to satisfy the class inheritance requirement
(3)Specify namespace of ToString() in compaction_service_test.cc to avoid function collision with CompactionService's ancestor classes

Test did:
make -j24 compaction_service_test
./compaction_service_test


db/compaction/compaction_service_test.cc
```diff
@@ -86,7 +90,8 @@ class CompactionServiceTest : public DBTestBase {
     for (int i = 0; i < 20; i++) {
       for (int j = 0; j < 10; j++) {
         int key_id = i * 10 + j;
-        ASSERT_OK(Put(Key(key_id), "value" + ToString(key_id)));
+        ASSERT_OK(
+            Put(Key(key_id), "value" + ROCKSDB_NAMESPACE::ToString(key_id)));
```
jay-zhuang, line 94: We don't need to specify the namespace here. Only one needed is in CompactionService derived classes, which has name conflict for ToString().

hx235: Got you!

jay-zhuang (approved):
👍

PR #8396, opened 2021-06-12
Fix flaky ManualCompactionMax test

jay-zhuang:
Recalculate the total size after generate new sst files.
New generated files might have different size as the previous time which
could cause the test failed.

Test Plan: ```
gtest-parallel ./db_compaction_test
--gtest_filter=DBCompactionTest.ManualCompactionMax -r 1000 -w 100
```


akankshamahajan15 (approved):
LGTM. Thanks for the fix!

PR #8402, opened 2021-06-14
Fix runtime linkage with libasan in Facebook platform009

pdillinger:
Summary:

Was seeing

    ./cache_test: error while loading shared libraries: libasan.so.5: cannot open shared object file: No such file or directory

etc. using COMPILE_WITH_ASAN=1 without USE_CLANG=1

Now including compiler libs in runtime ld path.

Test Plan: reproduced with local builds


akankshamahajan15 (approved):
LGTM. Thanks a lot for fixing it.

PR #8403, opened 2021-06-14
[CI] Increase cmake test DISCOVERY_TIMEOUT from 5 to 120 seconds

jay-zhuang:
Summary: cmake test discovery may timeout especially on Windows
platform. Increase it from default 5 seconds to 120 seconds.

Test Plan: Run Windows build 10 times without issue


akankshamahajan15 (approved):
Thanks Jay for the fix!

PR #8405, opened 2021-06-15
Parallelize secondary cache lookup in MultiGet

anand1976:
Implement the ```WaitAll()``` interface in ```LRUCache``` to allow callers to issue multiple lookups in parallel and wait for all of them to complete. Modify ```MultiGet``` to use this to parallelize the secondary cache lookups in order to reduce the overall latency. A call to ```cache->Lookup()``` returns a handle that has an incomplete value (nullptr), and the caller can call ```cache->IsReady()``` to check whether the lookup is complete, and pass a vector of handles to ```WaitAll``` to wait for completion. If any of the lookups fail, ```MultiGet``` will read the block from the SST file.

Another change in this PR is to rename ```SecondaryCacheHandle``` to ```SecondaryCacheResultHandle``` as it more accurately describes the return result of the secondary cache lookup, which is more like a future. 

Tests:
1. Add unit tests in lru_cache_test
2. Benchmark results with no secondary cache configured
Master -
```
readrandom   :      41.175 micros/op 388562 ops/sec;  106.7 MB/s (7277999 of 7277999 found)
readrandom   :      41.217 micros/op 388160 ops/sec;  106.6 MB/s (7274999 of 7274999 found)
multireadrandom :      10.309 micros/op 1552082 ops/sec; (28908992 of 28908992 found)
multireadrandom :      10.321 micros/op 1550218 ops/sec; (29081984 of 29081984 found)
```

This PR -
```
readrandom   :      41.158 micros/op 388723 ops/sec;  106.8 MB/s (7290999 of 7290999 found)
readrandom   :      41.185 micros/op 388463 ops/sec;  106.7 MB/s (7287999 of 7287999 found)
multireadrandom :      10.277 micros/op 1556801 ops/sec; (29346944 of 29346944 found)
multireadrandom :      10.253 micros/op 1560539 ops/sec; (29274944 of 29274944 found) 
```

### DIFF
--- a/cache/lru_cache.cc
+++ b/cache/lru_cache.cc
@@ -393,11 +393,9 @@ void LRUCacheShard::Promote(LRUHandle* e) {
   // strict_capacity_limit_ is true. In such a case, we don't want
   // InsertItem() to free the handle, since the item is already in memory
   // and the caller will most likely just read from disk if we erase it here.
-  Status s;
   if (e->value) {
     Cache::Handle* handle = reinterpret_cast<Cache::Handle*>(e);
-    ;
-    s = InsertItem(e, &handle, /*free_handle_on_fail=*/false);
+    Status s = InsertItem(e, &handle, /*free_handle_on_fail=*/false);
     if (s.ok()) {
       // InsertItem would have taken a reference on the item, so decrement it
       // here as we expect the caller to already hold a reference
--- a/cache/lru_cache.cc
+++ b/cache/lru_cache.cc
@@ -309,7 +309,8 @@ void LRUCacheShard::SetStrictCapacityLimit(bool strict_capacity_limit) {
   strict_capacity_limit_ = strict_capacity_limit;
 }
 
-Status LRUCacheShard::InsertItem(LRUHandle* e, Cache::Handle** handle) {
+Status LRUCacheShard::InsertItem(LRUHandle* e, Cache::Handle** handle,
+                                 bool free_handle_on_fail) {
   Status s = Status::OK();
   autovector<LRUHandle*> last_reference_list;
   size_t total_charge = e->CalcTotalCharge(metadata_charge_policy_);
@@ -323,14 +324,16 @@ Status LRUCacheShard::InsertItem(LRUHandle* e, Cache::Handle** handle) {
 
     if ((usage_ + total_charge) > capacity_ &&
         (strict_capacity_limit_ || handle == nullptr)) {
+      e->SetInCache(false);
       if (handle == nullptr) {
         // Don't insert the entry but still return ok, as if the entry inserted
         // into cache and get evicted immediately.
-        e->SetInCache(false);
         last_reference_list.push_back(e);
       } else {
-        delete[] reinterpret_cast<char*>(e);
-        *handle = nullptr;
+        if (free_handle_on_fail) {
+          delete[] reinterpret_cast<char*>(e);
+          *handle = nullptr;
+        }
         s = Status::Incomplete("Insert failed due to LRU cache being full.");
       }
     } else {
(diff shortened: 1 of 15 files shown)


cache/lru_cache.cc
```diff
@@ -375,6 +378,43 @@ Status LRUCacheShard::InsertItem(LRUHandle* e, Cache::Handle** handle) {
   return s;
 }
 
+void LRUCacheShard::Promote(LRUHandle* e) {
```
zhichao-cao, line 381: So, after promotion, we do not need to deal with the data in the secondary cache? For example, erase it, or  lower its priority?

cache/lru_cache.h
```diff
@@ -326,10 +340,20 @@ class ALIGN_AS(CACHE_LINE_SIZE) LRUCacheShard final : public CacheShard {
   double GetHighPriPoolRatio();
 
  private:
-  Status InsertItem(LRUHandle* item, Cache::Handle** handle);
+  friend class LRUCache;
+  // Insert an item into the hash table and, if handle is null, into the LRU
```
zhichao-cao, line 344: if handle is null, into the LRU -> if handle is null, insert into the LRU

cache/lru_cache.cc
```diff
@@ -557,6 +595,20 @@ void LRUCacheShard::Erase(const Slice& key, uint32_t hash) {
   }
 }
 
+bool LRUCacheShard::IsReady(Cache::Handle* handle) {
+  LRUHandle* e = reinterpret_cast<LRUHandle*>(handle);
+  bool ready = true;
+  if (e->IsPending()) {
+    assert(secondary_cache_);
+    if (e->sec_handle->IsReady()) {
```
zhichao-cao, line 603: Should we check if (e->sec_handle != nulllptr)?

cache/lru_cache.cc
```diff
@@ -662,6 +715,36 @@ double LRUCache::GetHighPriPoolRatio() {
...
+      }
+      sec_handles.emplace_back(lru_handle->sec_handle);
+    }
+    secondary_cache_->WaitAll(sec_handles);
+    for (Handle* handle : handles) {
+      if (!handle) {
+        continue;
+      }
+      LRUHandle* lru_handle = reinterpret_cast<LRUHandle*>(handle);
+      if (!lru_handle->IsPending()) {
```
zhichao-cao, line 738: Is it possible that, between L727 and L738, some of the flags of the handle in the handles are changed? So here we only promote the one that is still pending.

cache/lru_cache.cc
```diff
@@ -309,7 +309,8 @@ void LRUCacheShard::SetStrictCapacityLimit(bool strict_capacity_limit) {
   strict_capacity_limit_ = strict_capacity_limit;
 }
 
-Status LRUCacheShard::InsertItem(LRUHandle* e, Cache::Handle** handle) {
+Status LRUCacheShard::InsertItem(LRUHandle* e, Cache::Handle** handle,
```
zhichao-cao, line 312: Can we just give the default bool free_handle_on_fail = true here? So in other places, we do not need to change the code, only in Promote, we explicitly give it as false.

cache/lru_cache.cc
```diff
@@ -399,47 +439,45 @@ Cache::Handle* LRUCacheShard::Lookup(
...
-      // Being conservative here since there could be lookups that are
-      // actually ok to fail rather than succeed and bloat up the memory
-      // usage (preloading partitioned index blocks, for example).
-      Status s = InsertItem(e, reinterpret_cast<Cache::Handle**>(&e));
-      if (!s.ok()) {
-        assert(e == nullptr);
-        (*helper->del_cb)(key, value);
+      e->value = nullptr;
+      e->sec_handle = secondary_handle.release();
+
```
zhichao-cao, line 466: Can we put e->Ref() here and remove L469 to L471, L480, add if (!e->value)

cache/lru_cache.cc
anand1976: The coding guidelines we follow generally discourage default parameter values.

cache/lru_cache.cc
anand1976: We could. If ```IsPending()``` is true, it means we're waiting on the secondary cache and ```e->sec_handle``` should be valid, so it might be good to assert it.

cache/lru_cache.cc
anand1976: Its not possible, because at this point, none of the handles have been inserted in the ```LRUHandleTable```, so no other lookups can get to it. Only ```WaitAll``` will change the handle from pending to not pending. I can add some comments to make it explicit that the handle is not thread safe.

cache/lru_cache.cc
zhichao-cao: I see.

cache/lru_cache.h
```diff
@@ -168,7 +173,16 @@ struct LRUHandle {
     if (!IsSecondaryCacheCompatible() && info_.deleter) {
       (*info_.deleter)(key(), value);
     } else if (IsSecondaryCacheCompatible()) {
-      (*info_.helper->del_cb)(key(), value);
+      if (IsPending()) {
+        assert(sec_handle != nullptr);
+        SecondaryCacheResultHandle* s_handle = sec_handle;
```
zhichao-cao, line 178: Maybe change s_handle to tmp_sec_handle?

cache/lru_cache.h
```diff
@@ -326,10 +340,20 @@ class ALIGN_AS(CACHE_LINE_SIZE) LRUCacheShard final : public CacheShard {
...
+  // the LRU list. Older items are evicted as necessary. If the cache is full
+  // and free_handle_on_fail is true, the item is deleted and handle is set to.
+  Status InsertItem(LRUHandle* item, Cache::Handle** handle,
+                    bool free_handle_on_fail);
   Status Insert(const Slice& key, uint32_t hash, void* value, size_t charge,
                 DeleterFn deleter, const Cache::CacheItemHelper* helper,
                 Cache::Handle** handle, Cache::Priority priority);
+  // Promote an item looked up from the secondary cache to the LRU cache. The
+  // item is only inserted into the hash table and not the LRU list, and only
+  // if the cache is not at full capacity. The caller should hold a reference
```
zhichao-cao, line 354: Maybe make the comment here more clear. Why we do not insert it into the LRU list? What's the priority?

cache/lru_cache_test.cc
```diff
@@ -985,6 +1039,146 @@ TEST_F(DBSecondaryCacheTest, SecondaryCacheFailureTest) {
...
+      ASSERT_EQ(item->ToString(), values[i]);
+    }
+    cache->Release(results[i]);
+  }
+
+  cache.reset();
+  secondary_cache.reset();
+}
+
+// In this test, the block cache size is set to 6100, after insert 6 KV-pairs
```
zhichao-cao, line 1093: Remove the comment here, the block cache size is MB level already.

zhichao-cao (approved):
Thanks for addressing the review comments. It is a very useful feature. LGTM!

PR #8409, opened 2021-06-15
Rename ImmutableOptions variables 

mrambacher:
This is the next part of the ImmutableOptions cleanup.  After changing the use of ImmutableCFOptions to ImmutableOptions, there were places in the code that had did something like "ImmutableOptions* immutable_cf_options", where "cf" referred to the "old" type.  

This change simply renames the variables to match the current type.  No new functionality is introduced.


db/#flush_job.cc#
```diff
@@ -0,0 +1,530 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
```
pdillinger, line 1: Accidentally added temporary file

pdillinger (approved):
Remove the errant file and we're good!

PR #8411, opened 2021-06-15
Added memtable garbage statistics

bjlemaire:
**Summary**:
2 new statistics counters are added to RocksDB: `MEMTABLE_PAYLOAD_BYTES_AT_FLUSH` and `MEMTABLE_GARBAGE_BYTES_AT_FLUSH`. The former tracks how many raw bytes of useful data are present on the memtable at flush time, whereas the latter is tracks how many of these raw bytes are considered garbage, meaning that they ended up not being imported on the SSTables resulting from the flush operations.

**Unit test**: run `make db_flush_test -j$(nproc); ./db_flush_test` to run the unit test.
This executable includes 3 tests, that test support and correct stat calculations for workloads with inserts, deletes, and DeleteRanges. The parameters are set such that the workloads are performed on a single memtable, and a single SSTable is created as a result of the flush operation. The flush operation is manually called in the test file. The tests verify that the values of these 2 statistics counters introduced in this PR  can be exactly predicted, showing that we have a full understanding of the underlying operations.

**Performance testing**:
`./db_bench -statistics -benchmarks=fillrandom -num=10000000` repeated 10 times.
Timing done using "date" function in a bash script.
_Results_:
Original Rocksdb fork: mean 66.6 sec, std 1.18 sec.
This feature branch: mean 67.4 sec, std 1.35 sec.


memtable/memtable_garbage_test.cc
```diff
@@ -0,0 +1,282 @@
...
+  uint64_t mem_garbage_bytes =
+      options.statistics->getTickerCount(MEMTABLE_GARBAGE_BYTES);
+
+  EXPECT_EQ(mem_data_bytes, EXPECTED_MEMTABLE_DATA_BYTES);
+  EXPECT_EQ(mem_garbage_bytes, EXPECTED_MEMTABLE_GARBAGE_BYTES);
+
+  Close();
+}
+
+TEST_P(MemtableGarbageTest, DifferentKeyValuePairs) {
```
pdillinger, line 155: This nor `DifferentKeyValuePairsAndNumRepeat` is not sufficiently different to justify another test case, especially not copy-paste (annoying and hazardous to keep everything in sync). Matching exact byte counts gives us high confidence, and besides the accuracy of these stats are not correctness-critical.

What we should be testing also (if we have time) is more ways of creating garbage:
* Delete
* DeleteRange
* CompactionFilter (configured to filter on flush)

Or ways of creating "partial" or "almost" garbage:
* Merge
* GetSnapshot before all writes finish

I don't need to see all of these, but at least some work in this direction. Perhaps we can plan a follow-up PR for testing the remaining items.

memtable/memtable_garbage_test.cc
```diff
@@ -0,0 +1,282 @@
...
+
+  // Insertion of of K-V pairs, multiple times.
+  for (size_t i = 0; i < NUM_REPEAT; i++) {
+    ASSERT_OK(db_->Put(WriteOptions(), KEY1, VALUE1));
+    ASSERT_OK(db_->Put(WriteOptions(), KEY2, VALUE2));
+    ASSERT_OK(db_->Put(WriteOptions(), KEY3, VALUE3));
+  }
+  // We assert that the K-V pairs have been successfully inserted.
+  std::string value;
+  ASSERT_OK(db_->Get(ReadOptions(), KEY1, &value));
```
pdillinger, line 129: DBTestBase offers concise versions of Get, Put, and Flush

memtable/memtable_garbage_test.cc
```diff
@@ -0,0 +1,282 @@
...
+  ASSERT_EQ(value, VALUE1);
+  ASSERT_OK(db_->Get(ReadOptions(), KEY2, &value));
+  ASSERT_EQ(value, VALUE2);
+  ASSERT_OK(db_->Get(ReadOptions(), KEY3, &value));
+  ASSERT_EQ(value, VALUE3);
+
+  // Force flush to SST. Increments the statistics counter.
+  if (db_ != nullptr) {
+    FlushOptions flush_opt;
+    flush_opt.wait = true;  // function returns once the flush is over.
```
pdillinger, line 139: This is default and we make that assumption all over the place, so Flush() (from DBTestBase) should be fine.

memtable/memtable_garbage_test.cc
```diff
@@ -0,0 +1,282 @@
...
+  const uint64_t EXPECTED_MEMTABLE_DATA_BYTES =
+      NUM_REPEAT * USEFUL_PAYLOAD_BYTES;
+  const uint64_t EXPECTED_MEMTABLE_GARBAGE_BYTES =
+      (NUM_REPEAT - 1) * USEFUL_PAYLOAD_BYTES;
+
+  // Insertion of of K-V pairs, multiple times.
+  for (size_t i = 0; i < NUM_REPEAT; i++) {
+    ASSERT_OK(db_->Put(WriteOptions(), KEY1, VALUE1));
+    ASSERT_OK(db_->Put(WriteOptions(), KEY2, VALUE2));
+    ASSERT_OK(db_->Put(WriteOptions(), KEY3, VALUE3));
```
pdillinger, line 125: Overwriting with the same value is arguably not overwriting at all. It's only by a quirk of the implementation that this approach works.

pdillinger (changes requested):

memtable/memtable_garbage_test.cc
pdillinger: Excessive critique of this one aspect of an otherwise good test:

Among the ways a test can be bad, aside from failure to detect regression, is punishing or inhibiting progress. In this case, if we make an optimization not to insert redundant writes into the memtable, your test will be broken, and it will require some rework to get it into a good state. That either punishes such progress or creates the temptation to short-circuit your test in some way, such as naively updating the "expected" numbers without fully thinking through the consequences (no longer detecting regressions). Sometimes it's extremely hard to write a test that isn't somewhat "fragile." If it's reasonably clear and easy to update a test for conceivable future changes, it's not a big deal, but we should avoid punishing progress.

include/rocksdb/statistics.h
```diff
@@ -204,6 +204,9 @@ enum Tickers : uint32_t {
   COMPACT_READ_BYTES,   // Bytes read during compaction
   COMPACT_WRITE_BYTES,  // Bytes written during compaction
   FLUSH_WRITE_BYTES,    // Bytes written during flush
+  MEMTABLE_DATA_BYTES,  // Raw bytes of data present on memtable at flush time.
```
pdillinger, line 207: We should also be planning for a (possible / likely?) future when flush could go to the next memtable rather than to storage. How do these tickers fit into that picture, and are the names appropriate?

bjlemaire: A possible change in terminology: a flush to the next emmtable could be called "a purge" (as we are purging the garbage out of the memtable). The tickers have been renamed MEMTABLE_PAYLOAD_BYTES_AT_FLUSH and MEMTABLE_GARBAGE_BYTES_AT_FLUSH respectively (names still subject to changes). In the future, these statistics will not be incremented at purge time, only at flush time, to better reflect the impact of the purging policy on the ratio of garbage in the memtable.

db/builder.cc
```diff
@@ -254,6 +258,25 @@ Status BuildTable(
...
+          memtable_garbage_bytes != nullptr) {
+        const CompactionIterationStats& ci_stats = c_iter.iter_stats();
+        uint64_t total_payload_bytes = ci_stats.total_input_raw_key_bytes +
+                                       ci_stats.total_input_raw_value_bytes +
+                                       total_tombstone_payload_bytes;
+        uint64_t total_payload_bytes_written =
+            (tp.raw_key_size + tp.raw_value_size);
+        // Prevent underflow, which may still happen at this point
+        // since we only support inserts, deletes, and deleteRanges.
+        if (total_payload_bytes_written < total_payload_bytes) {
```
akankshamahajan15, line 271: Can there be case  where memtable_garbage_bytes = 0 and memtable_payload_bytes is nonzero? Will that be covered?

java/src/main/java/org/rocksdb/TickerType.java
```diff
@@ -752,6 +752,15 @@
     ERROR_HANDLER_AUTORESUME_RETRY_TOTAL_COUNT((byte) -0x1A),
     ERROR_HANDLER_AUTORESUME_SUCCESS_COUNT((byte) -0x1B),
 
+    /**
+     * Bytes of raw data (payload) found on memtable at flush time.
```
akankshamahajan15, line 756: Can we mention in comment that ```MEMTABLE_PAYLOAD_BYTES_AT_FLUSH``` means total data found including outdated data.

include/rocksdb/statistics.h
```diff
@@ -383,6 +383,12 @@ enum Tickers : uint32_t {
   ERROR_HANDLER_AUTORESUME_RETRY_TOTAL_COUNT,
   ERROR_HANDLER_AUTORESUME_SUCCESS_COUNT,
 
+  // Statistics for memtable garbage collection:
```
akankshamahajan15, line 386: These newly added stats can be added in History.md.

db/db_flush_test.cc
```diff
@@ -285,6 +285,379 @@ TEST_F(DBFlushTest, ScheduleOnlyOneBgThread) {
...
+  //  allow_concurrent_memtable_write is set to true by default )
+  options.inplace_update_support = false;
+  options.allow_concurrent_memtable_write = true;
+
+  // Enforce size of a single MemTable to 64MB (64MB = 67108864 bytes).
+  options.write_buffer_size = 64 << 20;
+
+  ASSERT_OK(TryReopen(options));
+
+  // Put multiple times the same key-values.
```
akankshamahajan15, line 344: Thanks for detailed explanation in the test.

akankshamahajan15 (approved):
Added some minor comments. Rest LGTM!. Thanks for the PR.

db/builder.cc
bjlemaire: Oh this is a great catch! I am changing the "<" for a "<=". And yes this is typically the case for example for purely sequential write workload (no overwrites, just writes). in that situation, total_payload_bytes_written==total_payload_bytes, which must give you memtable_garbage_bytes = 0 and memtable_payload_bytes>0.

PR #8412, opened 2021-06-16
Using existing crc32c checksum in checksum handoff for Manifest and WAL

zhichao-cao:
In PR #7523 , checksum handoff is introduced in RocksDB for WAL, Manifest, and SST files. When user enable checksum handoff for a certain type of file, before the data is written to the lower layer storage system, we calculate the checksum (crc32c) of each piece of data and pass the checksum down with the data, such that data verification can be down by the lower layer storage system if it has the capability. However, it cannot cover the whole lifetime of the data in the memory and also it potentially introduces extra checksum calculation overhead.

In this PR, we introduce a new interface in WritableFileWriter::Append, which allows the caller be able to pass the data and the checksum (crc32c) together. In this way, WritableFileWriter can directly use the pass-in checksum (crc32c) to generate the checksum of data being passed down to the storage system. It saves the calculation overhead and achieves higher protection coverage. When a new checksum is added with the data, we use Crc32cCombine #8305 to combine the existing checksum and the new checksum. To avoid the segmenting of data by rate-limiter before it is stored, rate-limiter is called enough times to accumulate enough credits for a certain write. This design only support Manifest and WAL which use log_writer in the current stage.

test plan: make check, add new testing cases.


file/writable_file_writer.cc
```diff
@@ -100,7 +104,104 @@ IOStatus WritableFileWriter::Append(const Slice& data) {
...
+  return s;
+}
+
+IOStatus WritableFileWriter::Append(const Slice& data,
+                                    uint32_t crc32c_checksum) {
+  // id we do not need to perform checksum handoff or this type of file will
+  // calcualte the checksum later after rate-limiter, we will use the legacy
+  // Append.
+  if (!perform_data_verification_ || !buffered_data_with_checksum_) {
+    (void)crc32c_checksum;
```
anand1976, line 128: Why is this needed?

db/log_writer.cc
```diff
@@ -154,7 +155,7 @@ IOStatus Writer::EmitPhysicalRecord(RecordType t, const char* ptr, size_t n) {
   // Write the header and the payload
   IOStatus s = dest_->Append(Slice(buf, header_size));
```
anand1976, line 156: Do we not have to pass the header checksum here?

file/writable_file_writer.cc
```diff
@@ -100,7 +104,104 @@ IOStatus WritableFileWriter::Append(const Slice& data) {
...
+  }
+
+  TEST_KILL_RANDOM("WritableFileWriter::Append:1");
+  if (s.ok()) {
+    filesize_ += data.size();
+  }
+  return s;
+}
+
+IOStatus WritableFileWriter::Append(const Slice& data,
```
anand1976, line 122: Can we reuse code from ```WritableFileWriter::Append```, since most of the logic is the same?

file/writable_file_writer.cc
```diff
@@ -451,6 +564,73 @@ IOStatus WritableFileWriter::WriteBuffered(const char* data, size_t size) {
...
+  assert(perform_data_verification_ && buffered_data_with_checksum_);
+  const char* src = data;
+  size_t left = size;
+  DataVerificationInfo v_info;
+  char checksum_buf[sizeof(uint32_t)];
+
+  // Check how much is allowed
+  size_t data_size = left;
+  if (rate_limiter_ != nullptr) {
+    while (data_size > 0) {
```
anand1976, line 584: I guess we loop here until the rate limiter allows to write the entire buffer since the checksum is calculated on the whole buffer. Maybe we can revisit at some later point, since it sort of defeats the purpose of the rate limiter.

anand1976 (commented):
Awesome! This adds a lot of value to the checksum handoff feature.

zhichao-cao:
> Awesome! This adds a lot of value to the checksum handoff feature.

Thanks for the review! Will address the comments.

db/log_writer.cc
zhichao-cao: @anand1976 So you mean we can directly use crc32c::Value(buf, header_size) to calculate the checksum here and pass down to writable file writer, instead of wait Append calculate internally?

file/writable_file_writer.cc
zhichao-cao: I will try to combine in the same Append with default checksum = 0.

file/writable_file_writer.cc
zhichao-cao: Sure. I will add a TODO to indicate the potential improvement.

db/log_writer.cc
zhichao-cao: @anand1976 If checksum handoff is not enabled, the calculated checksum of the header is useless. Therefore, it would be more efficient to call without the checksum and calculate the checksum internally when it is needed.

file/writable_file_writer.cc
```diff
@@ -132,6 +182,9 @@ IOStatus WritableFileWriter::Pad(const size_t pad_bytes) {
   }
   pending_sync_ = true;
   filesize_ += pad_bytes;
+  buffered_data_crc32c_checksum_ =
```
anand1976, line 185: Should we check for ```perform_data_verification_``` here?

file/writable_file_writer.cc
```diff
@@ -565,5 +696,99 @@ IOStatus WritableFileWriter::WriteDirect() {
...
+  // Calculate whole page final file advance if all writes succeed
+  size_t file_advance = TruncateToPageBoundary(alignment, buf_.CurrentSize());
+
+  // Calculate the leftover tail, we write it here padded with zeros BUT we
+  // will write
+  // it again in the future either on Close() OR when the current whole page
+  // fills out
+  size_t leftover_tail = buf_.CurrentSize() - file_advance;
+
+  // Round up, padd, and combine the checksum.
```
anand1976, line 716: padd -> pad

anand1976 (approved):
LGTM. Thanks!

file/writable_file_writer.cc
zhichao-cao: Good catch! I will add the check.

file/writable_file_writer.cc
```diff
@@ -451,6 +514,76 @@ IOStatus WritableFileWriter::WriteBuffered(const char* data, size_t size) {
...
+  assert(perform_data_verification_ && buffered_data_with_checksum_);
+  const char* src = data;
+  size_t left = size;
+  DataVerificationInfo v_info;
+  char checksum_buf[sizeof(uint32_t)];
+
+  // Check how much is allowed. Here, we loop until the rate limiter allows to
+  // write the entire buffer.
+  // TODO: need to be improved since it sort of defeats the purpose of the rate
+  // limiter
```
pdillinger, line 534: The long term rate limit is respected, but the burst limit can be violated with sufficiently large Appends. ("sort of defeats the purpose" is vague.)

If we imagine a scenario with a large Append and want to minimize the damage from writing it all at once, across different scenarios with saturated or unsaturated I/O buffers/queues/etc., I can imagine the most harm coming from  one of these depending on the scenario:
* Execute the write after the full size has been accepted by the rate limiter (current). Under low demand, this could lead to a quiet period followed by a write burst. 
* Execute the write before waiting on the rate limiter. Under high demand, this could excessively burst/queue writes in the I/O system.

Perhaps a compromise implementation would execute the write after at least half of the full size has been accepted by the rate limiter. Just an idea but would add extra implementation complexity.

db/blob/blob_file_builder.cc
```diff
@@ -191,7 +191,7 @@ Status BlobFileBuilder::OpenBlobFileIfNeeded() {
       immutable_options_->clock, io_tracer_, statistics,
       immutable_options_->listeners,
       immutable_options_->file_checksum_gen_factory.get(),
-      tmp_set.Contains(FileType::kBlobFile)));
+      tmp_set.Contains(FileType::kBlobFile), false));
```
pdillinger, line 194: https://google.github.io/styleguide/cppguide.html#Function_Argument_Comments

pdillinger (commented):
Some suggested clean-up

file/writable_file_writer.cc
```diff
@@ -81,26 +82,74 @@ IOStatus WritableFileWriter::Append(const Slice& data) {
...
-      size_t appended = buf_.Append(src, left);
-      left -= appended;
-      src += appended;
-
-      if (left > 0) {
-        s = Flush();
-        if (!s.ok()) {
-          break;
+  if (perform_data_verification_ && buffered_data_with_checksum_ &&
+      crc32c_checksum != 0) {
```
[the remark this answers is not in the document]
zhichao-cao: Right, in the current case, if the actual checksum is 0, it will not be checked. Append is widely used nearly everywhere for write and in most cases checksum is not provided. So explicitly use a new API for non-checksum case might not be a efficient engineering change. How about we require a flag in the API: Append(const Slice& data, bool checksum_provided = false, uint32_t crc32c_checksum = 0)? When the flag is set, we trust whatever is provided in the crc32c_checksum, otherwise, checksum must be re-calcualted?

file/writable_file_writer.cc
zhichao-cao: > Perhaps a compromise implementation would execute the write after at least half of the full size has been accepted by the rate limiter. Just an idea but would add extra implementation complexity.

The size of the data being written is hard to decide since it is controlled by when "WriteBuffered" is called. It is possible that, as you suggested, we set a hard limit, when the buffered data size is larger than a certain size, we have the second checksum for the rest of the data. Then, with some identification and control logic, in WriteBuffered, we will call 2 WritableFile::Append with their own checksum (first half and second half). So we will accumulate the credit from rate-limiter twice and relief the burst write in some degree. 

Or, we explicitly have an option to limit the write buffer size when checksum handoff is enabled, so the very large writes can be avoided. What do you think?

PR #8419, opened 2021-06-17
Make MemTableRepFactory into a Customizable class

mrambacher:
This PR does the following:
-> Makes the MemTableRepFactory into a Customizable class and creatable/configurable via CreateFromString
-> Makes the existing implementations compatible with configurations
-> Moves the "SpecialRepFactory" test class into testutil, accessible via the ObjectRegistry or a NewSpecial API

New tests were added to validate the functionality and all existing tests pass.  db_bench and memtablerep_bench were hand-tested to verify the functionality in those tools.


memtable/skiplistrep.cc
```diff
@@ -271,6 +273,34 @@ class SkipListRep : public MemTableRep {
...
+      OptionTypeFlags::kDontSerialize /*Since it is part of the ID*/, 0}},
+#endif
+};
+
+SkipListFactory::SkipListFactory(size_t lookahead) : lookahead_(lookahead) {
+  RegisterOptions("SkipListFactoryOptions", &lookahead_,
+                  &skiplist_factory_info);
+}
+
+std::string SkipListFactory::GetId() const {
```
zhichao-cao, line 289: Any rule here to create the ID based on name:lookahead_value?

table/plain/plain_table_factory.cc
```diff
@@ -151,73 +151,159 @@ Status GetPlainTableOptionsFromString(const ConfigOptions& config_options,
...
+    // No Id and no options.  Clear the object
+    result->reset();
+    return Status::OK();
+  } else if (id.empty()) {  // We have no Id but have options.  Not good
+    return Status::NotSupported("Cannot reset object ", id);
+  } else {
+#ifndef ROCKSDB_LITE
+    status = config_options.registry->NewUniqueObject(id, result);
+#else
+    std::vector<std::string> opts_list = StringSplit(id, ':');
```
zhichao-cao, line 264: Based on the previous code, id=name:lookahead_. But it is still strange to me to use it from the ID. not options in the "value"

tools/db_bench_tool.cc
```diff
@@ -1451,8 +1427,31 @@ static const bool FLAGS_table_cache_numshardbits_dummy __attribute__((__unused__
                           &ValidateTableCacheNumshardbits);
 
 namespace ROCKSDB_NAMESPACE {
-
 namespace {
+static Status CreateRepFactory(const ConfigOptions config_options,
```
zhichao-cao, line 1431: CreateRepFactory -> CreateMemTableRepFactory

table/plain/plain_table_factory.cc
mrambacher: The old serialized options could contain things like memtable=VectorRepFactory.  The new one will say memtable=VectorRepFactory:size.  The latter form (in the new code) should be parsed just fine by the old code (I will run a compatibility test against old versions soon).

If I do not use id=name:lookahead_, then the format will be memtable={id=VectoryRepFactory; lookahead=#}.  This format will not work for the old code.

If you prefer to not use the "id", I can make that work but am not sure there is any real advantage one way or another.

zhichao-cao: I see. Then it is fine here. But maybe add comment to make it clear why.

zhichao-cao (approved):
LGTM, thanks!

PR #8421, opened 2021-06-18
Trace MultiGet Keys and CF_IDs to the trace file

zhichao-cao:
Tracing the MultiGet information including timestamp, keys, and CF_IDs to the trace file for analyzing and replay.

test plan: make check, add test to trace_analyzer_test

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -2201,6 +2201,16 @@ void DBImpl::MultiGet(const ReadOptions& read_options, const size_t num_keys,
   }
 #endif  // NDEBUG
 
+  if (tracer_) {
+    // TODO: This mutex should be removed later, to improve performance when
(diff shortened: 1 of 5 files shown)


db/db_impl/db_impl.cc
```diff
@@ -1880,6 +1880,16 @@ std::vector<Status> DBImpl::MultiGet(
   }
 #endif  // NDEBUG
 
+  if (tracer_) {
```
anand1976, line 1883: There are a few different MultiGet overloaded APIs. We should trace all of them.

anand1976 (approved):
LGTM! Can you update the test to also call the other 2 MultiGet APIs?

zhichao-cao:
> LGTM! Can you update the test to also call the other 2 MultiGet APIs?

Sure, will update soon.

PR #8426, opened 2021-06-18
Add a class for measuring the amount of garbage generated during compaction

ltamasi:
Summary:
This is part of an alternative approach to #8316.
Unlike that approach, this one relies on key-values getting processed one by one
during compaction, and does not involve persistence.

Specifically, the patch adds a class `BlobGarbageMeter` that can track the number
and total size of blobs in a (sub)compaction's input and output on a per-blob file
basis. This information can then be used to compute the amount of additional
garbage generated by the compaction for any given blob file by subtracting the
"outflow" from the "inflow."

Note: this patch only adds `BlobGarbageMeter` and associated unit tests. I plan to
hook up this class to the input and output of `CompactionIterator` in a subsequent PR.

Test Plan:
`make check`


db/blob/blob_garbage_meter.h
```diff
@@ -0,0 +1,98 @@
...
+    BlobStats in_flow_;
+    BlobStats out_flow_;
+  };
+
+  Status ProcessInFlow(const Slice& key, const Slice& value);
+  Status ProcessOutFlow(const Slice& key, const Slice& value);
+
+  const std::unordered_map<uint64_t, BlobInOutFlow>& flows() const {
+    return flows_;
+  }
```
jay-zhuang, line 89: What's our plan on how `flows()` will be used? Feels like it's exposing too much internal information. Will it worth defining some functions for the user?

ltamasi: In order to log the amount of additional garbage for all blob files, we will need to iterate through the entire `flows_` map (i.e. it's not just lookup by file number), along these lines:

```
     const auto& flows = blob_garbage_meter.flows();

      for (const auto& pair : flows) {
        const uint64_t blob_file_number = pair.first;
        const BlobGarbageMeter::BlobInOutFlow& flow = pair.second;

        assert(flow.IsValid());
        if (flow.HasGarbage()) {
            // ... process flow, log garbage
        }
    }
```

jay-zhuang: Okay, then we might need some comments about `BlobInOutFlow` and `BlobStats`.

db/blob/blob_garbage_meter.h
```diff
@@ -0,0 +1,98 @@
...
+      in_flow_.Add(bytes);
+      assert(IsValid());
+    }
+    void AddOutFlow(uint64_t bytes) {
+      out_flow_.Add(bytes);
+      assert(IsValid());
+    }
+
+    const BlobStats& GetInFlow() const { return in_flow_; }
+    const BlobStats& GetOutFlow() const { return out_flow_; }
```
jay-zhuang, line 58: The same question as `flows()`, do we really need to give the caller for that?

jay-zhuang (approved):
LGTM.

I would prefer hiding more information as a general rule of class design, but on the other hand, it's a relatively simple internal classes (and maybe also for performance reason), it looks good to me.

db/blob/blob_garbage_meter.h
ltamasi: These are used in unit tests; otherwise, they're not necessary strictly speaking.

ltamasi:
> LGTM.
> 
> I would prefer hiding more information as a general rule of class design, but on the other hand, it's a relatively simple internal classes (and maybe also for performance reason), it looks good to me.

Agree about hiding as much as possible as a general rule. In this particular case though, we do need to be able to both iterate through the entire collection and do lookups by key (i.e. map/dictionary functionality). Theoretically we could hide the map behind an interface but like you said, that would introduce a performance penalty.

PR #8428, opened 2021-06-19
[ResourceManagement] Account for dictionary-building buffer in global memory limit

hx235:
Context:
Some data blocks are temporarily buffered in memory in BlockBasedTableBuilder for building compression dictionary used in data block compression. Currently this memory usage is not counted toward our global memory usage utilizing block cache capacity. To improve that, this PR charges that memory usage into the block cache to achieve better memory tracking and limiting.

Summary: 
- Reserve memory in block cache for buffered data blocks that are used to build a compression dictionary
- Release all the memory associated with buffering the data blocks mentioned above in EnterUnbuffered(), which is called when (a) buffer limit is exceeded after buffering OR (b) the block cache becomes full after reservation OR (c) BlockBasedTableBuilder calls Finish()

Test Plan:
- Passing existing unit tests
- Passing new unit tests

### DIFF
--- a/cache/cache_entry_roles.h
+++ b/cache/cache_entry_roles.h
@@ -31,8 +31,8 @@ enum class CacheEntryRole {
   kOtherBlock,
   // WriteBufferManager reservations to account for memtable usage
   kWriteBuffer,
-  // Compression dictionary building buffer reservations to account for memtable
-  // usage
+  // BlockBasedTableBuilder reservations to account for
+  // compression dictionary building buffer's memory usage
   kCompressionDictionaryBuildingBuffer,
   // Default bucket, for miscellaneous cache entries. Do not use for
   // entries that could potentially add up to large usage.
--- a/cache/cache_entry_roles.h
+++ b/cache/cache_entry_roles.h
@@ -14,6 +14,8 @@
 namespace ROCKSDB_NAMESPACE {
 
 // Classifications of block cache entries, for reporting statistics
+// Adding new enum to this class requires corresponding updates to
+// kCacheEntryRoleToCamelString and kCacheEntryRoleToHyphenString
 enum class CacheEntryRole {
   // Block-based table data block
   kDataBlock,
--- a/cache/cache_entry_roles.h
+++ b/cache/cache_entry_roles.h
@@ -29,6 +29,9 @@ enum class CacheEntryRole {
   kOtherBlock,
   // WriteBufferManager reservations to account for memtable usage
   kWriteBuffer,
+  // Compression dictionary building buffer reservations to account for memtable
+  // usage
+  kCompressionDictionaryBuildingBuffer,
   // Default bucket, for miscellaneous cache entries. Do not use for
   // entries that could potentially add up to large usage.
   kMisc,
--- a/table/block_based/block_based_table_builder.cc
+++ b/table/block_based/block_based_table_builder.cc
@@ -21,6 +21,7 @@
 #include <unordered_map>
 #include <utility>
 
+#include "cache/cache_entry_roles.h"
 #include "db/dbformat.h"
 #include "index_builder.h"
 #include "memory/memory_allocator.h"
@@ -312,7 +313,10 @@ struct BlockBasedTableBuilder::Rep {
   // `kBuffered` state is allowed only as long as the buffering of uncompressed
   // data blocks (see `data_block_buffers`) does not exceed `buffer_limit`.
   uint64_t buffer_limit;
-
+  bool buffer_exceeds_global_memory_limit;
+  // TODO-t CacheRep is the refactored CacheRep class from WriteBufferManager
+  // TODO-t figure out a better name for cache_rep in the context of this class
+  CacheRep* cache_rep;
   const bool use_delta_encoding_for_index_values;
   std::unique_ptr<FilterBlockBuilder> filter_builder;
   char cache_key_prefix[BlockBasedTable::kMaxCacheKeyPrefixSize];
@@ -444,6 +448,12 @@ struct BlockBasedTableBuilder::Rep {
       buffer_limit = std::min(tbo.target_file_size,
                               compression_opts.max_dict_buffer_bytes);
     }
+    // TODO figure out c++ member initialization syntax
+    buffer_exceeds_global_memory_limit = false;
+    cache_rep = (table_options.block_cache != nullptr
+                     ? new CacheRep(table_options.block_cache.get())
+                     : nullptr);
+
     for (uint32_t i = 0; i < compression_opts.parallel_threads; i++) {
       compression_ctxs[i].reset(new CompressionContext(compression_type));
     }
@@ -897,9 +907,11 @@ void BlockBasedTableBuilder::Add(const Slice& key, const Slice& value) {
       r->first_key_in_next_block = &key;
       Flush();
 
-      if (r->state == Rep::State::kBuffered && r->buffer_limit != 0 &&
-          r->data_begin_offset > r->buffer_limit) {
-        EnterUnbuffered();
+      if (r->state == Rep::State::kBuffered) {
+        if ((r->buffer_limit != 0 && r->data_begin_offset > r->buffer_limit) ||
+            r->buffer_exceeds_global_memory_limit) {
+          EnterUnbuffered();
+        }
       }
 
       // Add item to index block.
@@ -1001,6 +1013,7 @@ void BlockBasedTableBuilder::WriteBlock(BlockBuilder* block,
     assert(block_type == BlockType::kData);
     rep_->data_block_buffers.emplace_back(std::move(raw_block_contents));
     rep_->data_begin_offset += rep_->data_block_buffers.back().size();
+    ReserveLastBufferedDataBlockMemInCache();
     return;
   }
   WriteBlock(raw_block_contents, handle, block_type);
@@ -1910,10 +1923,76 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
                                         r->pending_handle);
       }
     }
-
+    // TODO: figure out when we can ensure all the buffered data blocks are
+    // flushed and call this function
+    // TODO: figure out what is the right status message for successfully
+    // flushing all the buffered data blocks
+    if (ok()) {
+      FreeAllBufferedDataBlocksMemFromCache();
+    }
     std::swap(iter, next_block_iter);
   }
   r->data_block_buffers.clear();
+  r->buffer_exceeds_global_memory_limit = false;
+}
+
+// TODO figure out a better name for this function to reflect what it does in
+// the context of this class
+void BlockBasedTableBuilder::ReserveLastBufferedDataBlockMemInCache() {
+  // TODO-t figure out a better way to represent there is no need to reserve mem
+  // due to no cache
+  CacheRep* cache_rep = rep_->cache_rep;
+  if (cache_rep == nullptr) {
+    return;
+  }
+
+  // TODO figure out whether lock is needed for the cache
+
+  size_t kSizeDummyEntry = rep_->data_block_buffers.back().size();
+  Cache::Handle* handle = nullptr;
+  // TODO-l learn how cache key works
+  // TODO-l learn how (dummy) entries are deleted in cache
+  Status status = cache_rep->cache_->Insert(
+      cache_rep->GetNextCacheKey(), nullptr, kSizeDummyEntry,
+      GetNoopDeleterForRole<
+          CacheEntryRole::kCompressionDictionaryBuildingBuffer>(),
+      &handle);
+  // TODO figure out if we still need to push back a handle when insertion fails
+  cache_rep->dummy_handles_.push_back(handle);
+
+  // TODO-q ask about how to identify the specific insertion error due to cache
+  // full other than comparing the full msg string
+  // TODO figure out if there is any other insertion error we need to pay
+  // attention to
+  if (status.IsIncomplete()) {
+    rep_->buffer_exceeds_global_memory_limit = true;
+  }
+}
+
+// TODO figure out a better name for this function to reflect what it does in
+// the context of this class
+void BlockBasedTableBuilder::FreeAllBufferedDataBlocksMemFromCache() {
+  // TODO-t figure out a better way to represent there is no need to free mem
+  // due to no cache
+  CacheRep* cache_rep = rep_->cache_rep;
+  if (cache_rep == nullptr) {
+    return;
+  }
+
+  // TODO figure out whether lock is needed for the cache
+
+  // TODO figure out c++ syntax for referencing the same object for
+  // cache_rep->dummy_handles_
+  while (!cache_rep->dummy_handles_.empty()) {
+    // TODO figure out c++ syntax for temporary variable inside the loop
+    auto* handle = cache_rep->dummy_handles_.back();
+    // TODO figure out a better way to handle empty handle/represent there is no
+    // need to release the handle if it is null
+    if (handle != nullptr) {
+      cache_rep->cache_->Release(handle, true);
+    }
+    cache_rep->dummy_handles_.pop_back();
+  }
 }
 
 Status BlockBasedTableBuilder::Finish() {
--- a/table/block_based/block_based_table_builder.cc
+++ b/table/block_based/block_based_table_builder.cc
@@ -905,11 +905,12 @@ void BlockBasedTableBuilder::Add(const Slice& key, const Slice& value) {
       r->first_key_in_next_block = &key;
       Flush();
       if (r->state == Rep::State::kBuffered) {
-        bool exceeds_buffer_limit = (r->buffer_limit != 0 && r->data_begin_offset > r->buffer_limit);
+        bool exceeds_buffer_limit =
+            (r->buffer_limit != 0 && r->data_begin_offset > r->buffer_limit);
         bool is_cache_full = false;
 
         // Increase cache reservation for the last buffered data block
-        // only if the block is not going to be unbuffered immediately 
+        // only if the block is not going to be unbuffered immediately
         // and there exists a cache reservation manager
         if (!exceeds_buffer_limit && r->cache_rev_mng != nullptr) {
           Status s = r->cache_rev_mng->UpdateCacheReservation<
@@ -1937,7 +1938,8 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
   r->data_begin_offset = 0;
   if (r->cache_rev_mng != nullptr) {
     Status s = r->cache_rev_mng->UpdateCacheReservation<
-        CacheEntryRole::kCompressionDictionaryBuildingBuffer>(r->data_begin_offset);
+        CacheEntryRole::kCompressionDictionaryBuildingBuffer>(
+        r->data_begin_offset);
     s.PermitUncheckedError();
   }
 }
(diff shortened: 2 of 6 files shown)


table/block_based/block_based_table_builder.cc
```diff
@@ -461,6 +465,10 @@ struct BlockBasedTableBuilder::Rep {
       buffer_limit = std::min(tbo.target_file_size,
                               compression_opts.max_dict_buffer_bytes);
     }
+    // TODO figure out c++ member initialization syntax
+    buffer_exceeds_global_memory_limit = false;
+    cache_rep = (table_options.block_cache != nullptr ? new CacheRep(tbo.block_cache.get()) : nullptr);
```
hx235, line 470: Question 1:
While the block cache in the WriteBufferManager's case is passed in on WBM's creation, for BlockBasedTableBuilder, I believe we are using the block cache passed in the table_options. Am I correct?

table/block_based/block_based_table_builder.cc
```diff
@@ -904,9 +912,11 @@ void BlockBasedTableBuilder::Add(const Slice& key, const Slice& value) {
       r->first_key_in_next_block = &key;
       Flush();
 
-      if (r->state == Rep::State::kBuffered && r->buffer_limit != 0 &&
-          r->data_begin_offset > r->buffer_limit) {
-        EnterUnbuffered();
+      if (r->state == Rep::State::kBuffered) {
+        if ((r->buffer_limit != 0 && r->data_begin_offset > r->buffer_limit) ||
+            r->buffer_exceeds_global_memory_limit) {
```
hx235, line 917: Question 2:
I decided to do the checking and switching to unbuffer here instead of in WriteBlock() right after we add the data block into the buffer. My motivation is to (1)keep WriteBlock() simple and Flush() easy to resonate (2)align with what "checking against user-set buffer_limit" does since they all follow unbuffering-policy-by-size.

Downside though, is that we need to keep a shared bool variable "buffer_exceeds_global_memory_limit" around and we need to remember to manipulate that. 

Let me know if you see we should do the checking and switching somewhere else.

table/block_based/block_based_table_builder.cc
```diff
@@ -1886,10 +1897,72 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
...
   r->data_block_buffers.clear();
+  r->buffer_exceeds_global_memory_limit = false;
+}
+
+// TODO figure out a better name for this function to reflect what it does in
+// the context of this class
+void BlockBasedTableBuilder::ReserveLastBufferedDataBlockMemWithCache() {
+  // TODO-t figure out a better way to represent there is no need to reserve mem due to no cache
+  CacheRep* cache_rep = rep_->cache_rep;
+  if (cache_rep == nullptr) {
```
hx235, line 1916: Question 3:
I've seen different practice of condition checking in the codebase. For the scenario here, I can think of three ways of checking against no caching (i.e, when "no_block_cache" is true in the "BlockBasedTableOptions").
(a)check that before calling ReserveLastBufferedDataBlockMemWithCache and use assertion within the function
(b)check inside the function and use early return - like what I am doing here
(c)check inside the function and use "late return" - if (cache_rep != nullptr) {do useful stuff}

I've seen all three practices in the codebase so I figured I should ask if there is a general practice that I should follow in scenario like this.

table/block_based/block_based_table_builder.cc
```diff
@@ -1886,10 +1897,72 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
...
+      &handle);
+  // TODO figure out if we still need to push back a handle when insertion fails
+  cache_rep->dummy_handles_.push_back(handle);
+
+  // TODO-q ask about how to identify the specific insertion error due to cache
+  // full other than comparing the full msg string
+  // TODO figure out if there is any other insertion error we need to pay
+  // attention to
+  if (status.IsIncomplete() &&
+      strcmp(status.getState(), "Insert failed due to LRU cache being full.")) {
```
hx235, line 1939: Question 4:
Just by tracing down to what are the possible statuses returned by insertion to cache, it seems that the status.IsIncomplete() is equivalent to "insertion failure due to cache being full". Am I correct?

table/block_based/block_based_table_builder.cc
```diff
@@ -1008,6 +1018,7 @@ void BlockBasedTableBuilder::WriteBlock(BlockBuilder* block,
     assert(is_data_block);
     rep_->data_block_buffers.emplace_back(std::move(raw_block_contents));
     rep_->data_begin_offset += rep_->data_block_buffers.back().size();
+    ReserveOneBufferedDataBlockMemWithCache();
```
hx235, line 1021: Question 5:
As for the practice of using compiler flag such NDEBUG, ROCKSDB_LITE for new feature
(1)how to tell whether I need to use NDEBUG for a section of code?
(2)for ROCKSDB_LITE, I read through https://github.com/facebook/rocksdb/blob/master/ROCKSDB_LITE.md but I still can't tell whether I need to use ROCKSDB_LITE or not - need some help here

table/block_based/block_based_table_builder.cc
```diff
@@ -1886,10 +1899,71 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
...
+          CacheEntryRole::kCompressionDictionaryBuildingBuffer>(),
+      &handle);
+  // TODO figure out if we still need to push back a handle when insertion fails
+  cache_rep->dummy_handles_.push_back(handle);
+
+  // TODO-q ask about how to identify the specific insertion error due to cache
+  // full other than comparing the full msg string
+  // TODO figure out if there is any other insertion error we need to pay
+  // attention to
+  if (status.IsIncomplete()) {
```
hx235, line 1940: Question 4:
Just by tracing down and see what are possible statuses returned by cache insertion and it seems that "status.IsIncomplete())" is equivalent to insertion failure due to cache full. Did I miss anything?

table/block_based/block_based_table_builder.cc
```diff
@@ -461,6 +465,12 @@ struct BlockBasedTableBuilder::Rep {
       buffer_limit = std::min(tbo.target_file_size,
                               compression_opts.max_dict_buffer_bytes);
     }
+    // TODO figure out c++ member initialization syntax
+    buffer_exceeds_global_memory_limit = false;
+    cache_rep = (table_options.block_cache != nullptr
+                     ? new CacheRep(table_options.block_cache.get())
```
hx235, line 471: Question 1:
While the block_cache is WriteBufferManager is passed in on WBM's creation, I believe we should the account memory in the block cache set in the table_options. Did I understand it correctly?

table/block_based/block_based_table_builder.cc
```diff
@@ -1008,6 +1020,7 @@ void BlockBasedTableBuilder::WriteBlock(BlockBuilder* block,
     assert(is_data_block);
     rep_->data_block_buffers.emplace_back(std::move(raw_block_contents));
     rep_->data_begin_offset += rep_->data_block_buffers.back().size();
+    ReserveLastBufferedDataBlockMemInCache();
```
hx235, line 1023: Question 5:
As for the practice of using compiler flag such NDEBUG, ROCKSDB_LITE for new feature
(1)how to tell whether I need to use NDEBUG for a section of code?
(2)for ROCKSDB_LITE, I read through https://github.com/facebook/rocksdb/blob/master/ROCKSDB_LITE.md but I still can't tell whether I need to use ROCKSDB_LITE or not - need some help here

table/block_based/block_based_table_builder.cc
hx235: Just realized that on https://github.com/facebook/rocksdb/wiki/Block-Cache#usage, it says "RocksDB will create the default block cache if block_cache is set to nullptr. To disable block cache completely: ...". 

So it seems that there are cases where (1)user completely disabled block cache - in this case, costing buffer memory is not needed (2)user didn't disable block cache and block_cache == nullptr, RocksDB provides a default block cache (3)user passed in its block cache. 

I will temporarily close this question until I read more codes to figure out how to include case (1), (2) in my implementation.

table/block_based/block_based_table_builder.cc
```diff
@@ -461,6 +465,12 @@ struct BlockBasedTableBuilder::Rep {
       buffer_limit = std::min(tbo.target_file_size,
                               compression_opts.max_dict_buffer_bytes);
     }
+    // TODO figure out c++ member initialization syntax
+    buffer_exceeds_global_memory_limit = false;
+    cache_rep = (table_options_.no_block_cache
+                     ? nullptr)
+                     : new CacheRep(table_options.block_cache.get());
```
hx235, line 472: Question 1:
While the block_cache in the case of WriteBufferManager is passed in on WBM's creation, in this case, I believe we should account memory in the block cache that is set in the table_options by the block_based_table_factory. Did I understand it correctly?

table/block_based/block_based_table_builder.cc
```diff
@@ -311,7 +312,10 @@ struct BlockBasedTableBuilder::Rep {
   // `kBuffered` state is allowed only as long as the buffering of uncompressed
   // data blocks (see `data_block_buffers`) does not exceed `buffer_limit`.
   uint64_t buffer_limit;
-
+  bool buffer_exceeds_global_memory_limit;
+  // TODO-t CacheRep is the refactored CacheRep class from WriteBufferManager
+  // TODO-t figure out a better name for cache_rep in the context of this class
+  CacheRep* cache_rep;
```
hx235, line 318: Note[Implementation blocker]: 
we need to refactor/improve CacheRep from WriteBufferManager before we can progress from this line in this PR.

table/block_based/block_based_table_builder.cc
ajkr: > (1)how to tell whether I need to use NDEBUG for a section of code?

It is similar to `assert()`, which uses `NDEBUG` ("non-debug"). Code in `assert()` statements or `NDEBUG` sections will be excluded from compilation in production builds. It's useful for code whose only purpose is to validate some other code/assumptions are correct.

> (2)for ROCKSDB_LITE, I read through https://github.com/facebook/rocksdb/blob/master/ROCKSDB_LITE.md but I still can't tell whether I need to use ROCKSDB_LITE or not - need some help here

Some features are only supported in lite mode. Dictionary compression is not such a feature so you don't need to use it.

table/block_based/block_based_table_builder.cc
ajkr: They are all similarly good. I slightly prefer (a). For (b) and (c) the function would more accurately be called `MaybeReserveLastBufferedDataBlockMemInCache()`, with the downside that the reader needs some slight extra effort to learn what cases "maybe" means yes vs. no.

table/block_based/block_based_table_builder.cc
ajkr: I didn't read the code, but did read the public API doc, and it aligns with your findings - https://github.com/facebook/rocksdb/blob/be8199cdb99b2de17a7d1479a16e9944c278e51f/include/rocksdb/cache.h#L246-L247.

table/block_based/block_based_table_builder.cc
```diff
@@ -1886,10 +1899,76 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
...
+  size_t kSizeDummyEntry = rep_->data_block_buffers.back().size();
+  Cache::Handle* handle = nullptr;
+  // TODO-l learn how cache key works
+  // TODO-l learn how (dummy) entries are deleted in cache
+  Status status = cache_rep->cache_->Insert(
+      cache_rep->GetNextCacheKey(), nullptr, kSizeDummyEntry,
+      GetNoopDeleterForRole<
+          CacheEntryRole::kCompressionDictionaryBuildingBuffer>(),
+      &handle);
+  // TODO figure out if we still need to push back a handle when insertion fails
```
ajkr, line 1936: The API doc only says deleter needs to be called by client on failure, which is irrelevant to us since the value is nullptr. So I assume we don't need to do anything else like call `Release()`.

table/block_based/block_based_table_builder.cc
```diff
@@ -1886,10 +1899,76 @@ void BlockBasedTableBuilder::EnterUnbuffered() {
                                         r->pending_handle);
       }
     }
-
+    // TODO: figure out when we can ensure all the buffered data blocks are
+    // flushed and call this function
+    // TODO: figure out what is the right status message for successfully
+    // flushing all the buffered data blocks
+    if (ok()) {
+      FreeAllBufferedDataBlocksMemFromCache();
```
ajkr, line 1907: This appears to be inside the loop over `r->data_block_buffers` so it'll execute a lot of times.

Also we need to find a way to call this even when there's an error (might be called in multiple places).

hx235:
Update: 
- Addressed PR comments on () and two previous impl bugs ()
- Added new unit tests
- Clarified a subtle detail about adding a new CacheEntryRole enum
- Ready for review after the signals are clear

table/block_based/block_based_table_builder.cc
hx235: > Also we need to make sure to call this in the error case so a failed table building releases its dummy entry handles.

I changed the code to release memory only after the line `r->data_block_buffers.clear();` and removed my previous buggy `ok()` check since the memory reserved should stay as close as the total bytes in `r->data_block_buffers` (which is `r->data_begin_offset` I believe) regardless of any failure that can cause a non-okay status up to this point. Not sure if this is what your comment was suggesting :)

table/block_based/block_based_table_builder.cc
```diff
@@ -896,10 +904,24 @@ void BlockBasedTableBuilder::Add(const Slice& key, const Slice& value) {
...
+      if (r->state == Rep::State::kBuffered) {
+        bool exceeds_buffer_limit =
+            (r->buffer_limit != 0 && r->data_begin_offset > r->buffer_limit);
+        bool is_cache_full = false;
+
+        // Increase cache reservation for the last buffered data block
+        // only if the block is not going to be unbuffered immediately
+        // and there exists a cache reservation manager
+        if (!exceeds_buffer_limit && r->cache_rev_mng != nullptr) {
+          Status s = r->cache_rev_mng->UpdateCacheReservation<
```
hx235, line 916: [Question]
It seems that there isn't any synchronization inside of BlockBasedTableBuilder for data structure like data_blocks_buffer nor did I see multi-threaded usage of one BlockBasedTableBuilder object other than parallel compression in the test, which seems be irrelevant to our data buffering (although I am having a bit difficulties in understanding the code of how parallel compression play its part in BlockBasedBuilder)

I wonder if we can skip the external synchronization of cache_rev_mng inside of BlockBasedTableBuilder (which I feel we can)

cache/cache_entry_roles.h
```diff
@@ -14,6 +14,8 @@
 namespace ROCKSDB_NAMESPACE {
 
 // Classifications of block cache entries, for reporting statistics
+// Adding new enum to this class requires corresponding updates to
+// kCacheEntryRoleToCamelString and kCacheEntryRoleToHyphenString
```
hx235, line 18: [Note]
I tried also adding a `static_assert(kCacheEntryRoleToCamelString.size(), kNumCacheEntryRoles)` in   cache/cache_entry_roles.c but the assertion doesn't work and I haven't figured out the reason why ...

table/block_based/block_based_table_builder.cc
pdillinger: BlockBasedTableBuilder is generally single-threaded (except the experimental parallel compression parts, probably not relevant here), so having a separate cache_rev_mng per builder should suffice. 👍

cache/cache_entry_roles.h
pdillinger: std::array has size specified in template parameter, and you're allowed to initialize fewer elements. I wanted to add a similar static assertion, but C++ doesn't make it easy.

table/block_based/block_based_table_builder.cc
hx235: Thanks for jumping in and answering my question!

cache/cache_entry_roles.h
hx235: > std::array has size specified in template parameter, and you're allowed to initialize fewer elements.

That's exactly why my static_assert does not work! The only thing I can think of right now is to do some stripped string match between the "kMisc" and "kNumCacheEntryRoles-1"-th element of kCacheEntryRoleToCamelString but accessing the uninitialized element of an array (i.e, "kNumCacheEntryRoles-1"-th element of kCacheEntryRoleToCamelString if we forgot to update it) can give undefined behavior.

So now I can see more why previous attempt to add such a static assertion failed and hope that future developers find my comment useful!

table/block_based/block_based_table_builder.cc
ajkr: Right, the background threads will not be involved in `BlockBasedTableBuilder::Add()`. They only do compressing/writing, and also they only start doing work once we `EnterUnbuffered()`.

ajkr (approved):
LGTM!

table/table_test.cc
```diff
@@ -4746,6 +4747,242 @@ TEST_P(BlockBasedTableTest, OutOfBoundOnNext) {
...
+  // Adding the first key won't trigger a flush by FlushBlockEveryKeyPolicy
+  // therefore won't trigger any data block's buffering
+  builder->Add(ik1.Encode(), value1);
+  ASSERT_EQ(cache->GetPinnedUsage(), 0 * kSizeDummyEntry);
+
+  std::string key2 = "key2";
+  std::string value2 = "val2";
+  InternalKey ik2(key2, 1 /* sequnce number */, kTypeValue);
+  // Adding the second key will trigger a flush of the last data block (the one
+  // containing key1 and value1) by FlushBlockEveryKeyPolicy and hence trigger
```
ajkr, line 4804: Great work finding test utils like `FlushBlockEveryKeyPolicy` to write clean/clear/simple unit tests.

table/table_test.cc
```diff
@@ -4746,6 +4747,242 @@ TEST_P(BlockBasedTableTest, OutOfBoundOnNext) {
...
+    BlockBasedTableTest,
+    IncreaseCacheReservationForCompressDictBuildingBufferOnBuilderAddAndDecreaseOnBufferLimitExceed) {
+  constexpr std::size_t kSizeDummyEntry = 256 * 1024;
+  constexpr std::size_t kMetaDataChargeOverhead = 10000;
+  constexpr std::size_t kCacheCapacity = 8 * 1024 * 1024;
+  constexpr std::size_t kMaxDictBytes = 1024;
+  // A small positive kMaxDictBufferBytes leads to a small buffer limit for our
+  // testing purpose. 27 is chosen so that buffering two data blocks, each
+  // containing key1, value1 and key2, value2 will exceed buffer limit
+  constexpr std::size_t kMaxDictBufferBytes = 27;
```
ajkr, line 4828: 27 is kind of an odd number when KVs appear to be six bytes (three byte key, three byte value). I guess it incorporates metadata overhead. If you want to hide detail about how large metadata is, perhaps make KVs larger (like 1KB), and kMaxDictBufferBytes a multiple of it (like 2KB).

ajkr (commented):
I think the simplicity of this diff is a nice payoff for the investment in writing CacheReservationManager.

table/table_test.cc
hx235: Good point - will update that!

hx235:
Update:
- Adopted suggestion on adjusting test value to hide impl details of metadata
- Refined some comments' wording
- Will merge after signals are cleared

cache/cache_entry_roles.h
pdillinger: Fortunately, UBSAN (and friends?) are pretty good at picking up on using uninitialized memory.

PR #8430, opened 2021-06-21
Skip c_test and env_test when ASSERT_STATUS_CHECKED=1

ajkr:
- `c_test` fails because `rocksdb_compact_range()` swallows a `Status`.
- `env_test` fails because `ReadRequest`s to `MultiRead()` do not have their `Status`es checked.

Test Plan: `ASSERT_STATUS_CHECKED=1 make -j48 check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -516,8 +516,10 @@ endif
 ifdef ASSERT_STATUS_CHECKED
 	# TODO: finish fixing all tests to pass this check
 	TESTS_FAILING_ASC = \
+		c_test \
 		db_test \
 		db_test2 \
+		env_test \
 		range_locking_test \
 		testutil_test \
 


mrambacher:
Why skip the tests rather than fixing the checks?  I thought that at least env_test was at one point "cleared" of ASC errors.

ajkr:
Fixing the checks is ideal but I didn't have time. In the meantime it'd be nice if the test suite (`ASSERT_STATUS_CHECKED=1 make check`) and CI didn't fail on known issues, which might cover up new/unknown issues.

ajkr:
> env_test was at one point "cleared" of ASC errors.

Yeah, this one is closer. Only the tests requiring -DROCKSDB_IOURING_PRESENT fail.

The c_test on the other hand looks like it needs new APIs to return CompactRange() errors. It feels surprising that one ever passed.

jay-zhuang (approved):
LGTM.

cc. oncall for investigation and ideally fix them later @zhichao-cao

zhichao-cao (approved):
LGTM!

PR #8436, opened 2021-06-21
Fix the potential Segment Fault in RecoverFromBGError when db_ is nullptr

zhichao-cao:
RecoverFromBGError can be called by a separate thread (e.g., the rocksdb::SstFileManagerImpl::ClearError()). It is possible that application has already taken the action to close the db or restart the db. Therefore, when the recover thread calls RecoverFromBGError, db_ pointer can be nullptr in some cases and causes segment fault. Fix it by checking the db_ pointer. If it is nullptr, directly returns the bg_error_

test plan: make check


anand1976 (commented):
Where is ```db_``` being set to nullptr? And shouldn't ```DBImpl::CloseHelper``` wait for any background recovery thread to exit before completing the close?

PR #8437, opened 2021-06-22
Print blob file checksums as hex

ltamasi:
Summary:
Currently, blob file checksums are incorrectly dumped as raw bytes
in the `ldb manifest_dump` output (i.e. they are not printed as hex).
The patch fixes this and also updates some test cases to reflect that
the checksum value field in `BlobFileAddition` and `SharedBlobFileMetaData`
contains the raw checksum and not a hex string.

Test Plan:
`make check`
Tested using `ldb manifest_dump`


akankshamahajan15 (approved):
LGTM!

PR #8438, opened 2021-06-22
Revert "Revert "Snapshot release triggered compaction without multiple tombstones (#8357)" (#8410)"

jay-zhuang:
This reverts commit 25be1ed66a354ee1d665d7a28eb20d36afc75e90.

Test Plan: Run the impacted mysql test 40 times

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,9 +1,12 @@
 # Rocksdb Change Log
+## Unreleased
+### Behavior Changes
+* Obsolete keys in the bottommost level that were preserved for a snapshot will now be cleaned upon snapshot release in all cases. This form of compaction (snapshot release triggered compaction) previously had an artificial limitation that multiple tombstones needed to be present.
+
 ## 6.22.0 (2021-06-18)
 ### Behavior Changes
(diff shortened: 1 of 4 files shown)


HISTORY.md
```diff
@@ -3,7 +3,7 @@
 ### Behavior Changes
 * Added two additional tickers, MEMTABLE_PAYLOAD_BYTES_AT_FLUSH and MEMTABLE_GARBAGE_BYTES_AT_FLUSH. These stats can be used to estimate the ratio of "garbage" (outdated) bytes in the memtable that are discarded at flush time.
 * Added API comments clarifying safe usage of Disable/EnableManualCompaction and EventListener callbacks for compaction.
-
+* Obsolete keys in the bottommost level that were preserved for a snapshot will now be cleaned upon snapshot release in all cases. This form of compaction (snapshot release triggered compaction) previously had an artificial limitation that multiple tombstones needed to be present.
```
ajkr, line 6: I think we should put this under "Unreleased" "Behavior Changes" since it doesn't easily meet the criteria for backporting.

ajkr (approved):
LGTM, thanks for completing this.

igorcanadi:
This change triggers the immediate compaction on the first L0 file created in a new database if any snapshot was held during the flush. This is a fairly common scenario and leads to wasted work because that L0 file will be compacted soon anyway, assuming constant stream of updates.

Would it be sane to avoid triggering the bottommost compaction if only a single file exists in the LSM tree? Any other way to avoid this pathological case?

PR #8439, opened 2021-06-22
Fix double slashes in user-provided db path.

bjlemaire:
At the moment, the following command : "`./ --db=mypath/ dump_file_files`" returns a series of erronous names with double slashes, ie: "`mypath//000xxx.sst`", including manifest file names with double slashes "`mypath//MANIFEST-00XXX`", whereas "`./ --db=mypath dump_file_files`" correctly returns "`mypath/000xxx.sst`" and "`mypath/MANIFEST-00XXX`".

This (very short) PR simply checks if there is a need to add or remove any '`/`' character when the `db_path` and `manifest_filename`/sst `filenames` are concatenated.

### DIFF
--- a/tools/ldb_cmd.cc
+++ b/tools/ldb_cmd.cc
@@ -3351,7 +3351,18 @@ void DBFileDumperCommand::DoCommand() {
   }
   // remove the trailing '\n'
   manifest_filename.resize(manifest_filename.size() - 1);
-  std::string manifest_filepath = db_->GetName() + "/" + manifest_filename;
+  std::string manifest_filepath = db_->GetName();
+  // Correct concatenation of filepath and filename:
+  // Add slash '/' in between if necessary, or check that
+  // there is no double slashes when concatenation happens.
+  if (manifest_filepath.size() > 0 && manifest_filepath.back() != '/' &&
+      manifest_filename.front() != '/') {
+    manifest_filepath += "/";
+  } else if (manifest_filepath.size() > 0 && manifest_filepath.back() == '/' &&
+             manifest_filename.front() == '/') {
+    manifest_filepath.pop_back();
+  }
+  manifest_filepath += manifest_filename;
   std::cout << manifest_filepath << std::endl;
   DumpManifestFile(options_, manifest_filepath, false, false, false);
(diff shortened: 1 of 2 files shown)


tools/ldb_cmd.cc
```diff
@@ -3351,7 +3351,18 @@ void DBFileDumperCommand::DoCommand() {
...
+  // Add slash '/' in between if necessary, or check that
+  // there is no double slashes when concatenation happens.
+  if (manifest_filepath.size() > 0 && manifest_filepath.back() != '/' &&
+      manifest_filename.front() != '/') {
+    manifest_filepath += "/";
+  } else if (manifest_filepath.size() > 0 && manifest_filepath.back() == '/' &&
+             manifest_filename.front() == '/') {
+    manifest_filepath.pop_back();
+  }
+  manifest_filepath += manifest_filename;
```
mrambacher, line 3365: Is this different than NormalizePath?  file/filename.{cc,h}

bjlemaire: Thanks this is very useful I wasn't aware of this function - I made an edit such that all the situations are handled, including if neither manifest_filepath nor manifest_filename have '/' (because then the NormalizePath function wouldn't be able to add a missing '/').

mrambacher: As it is coded (without NormalizePath), I am not sure this will necessarily work on Windows.

mrambacher (approved):
LGTM.

Is there any testing for this change (other than observation)?

tools/ldb_cmd.cc
bjlemaire: Apologies, it took me a couple minutes to test my edit locally before committing - I just pushed the update a minute ago if you want to take a look at the edit (with NormalizePath).

bjlemaire:
Just crafter a short unit test to check that when the user specifies a dbpath with a terminating '/', there is no double (or more) '/' characters in the manifest and sst file names.

akankshamahajan15 (approved):
LGTM! thank you for fixing it.

mrambacher (approved):
LGTM.  Thanks for the fix!

PR #8443, opened 2021-06-22
Add an internal iterator that can measure the inflow of blobs

ltamasi:
Summary:
Follow-up to #8426 .

The patch adds a new kind of `InternalIterator` that wraps another one and
passes each key-value encountered to `BlobGarbageMeter` as inflow.
This iterator will be used as an input iterator for compactions when the input
SSTs reference blob files.

Test Plan:
`make check`

### DIFF
--- a/db/blob/blob_counting_iterator.h
+++ b/db/blob/blob_counting_iterator.h
@@ -1,0 +1,129 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#pragma once
+
+#include <cassert>
+
+#include "table/internal_iterator.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+class BlobCountingIterator : public InternalIterator {
+ public:
+  explicit BlobCountingIterator(InternalIterator* iter) : iter_(iter) {
+    assert(iter_);
+  }
+
+  bool Valid() const override {
+    assert(iter_);
+    return iter_->Valid();
+  }
+
+  void SeekToFirst() override {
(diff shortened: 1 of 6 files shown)


db/blob/blob_counting_iterator.h
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  void SeekToLast() override {
+    iter_->SeekToLast();
+    UpdateAndCountBlobIfNeeded();
+  }
+
+  void Seek(const Slice& target) override {
+    iter_->Seek(target);
+    UpdateAndCountBlobIfNeeded();
```
jay-zhuang, line 45: If we call `Seek()` then `Next()`, does it mean anything before that target won't be counted?

db/blob/blob_counting_iterator.h
```diff
@@ -0,0 +1,142 @@
...
+    const bool res = iter_->NextAndGetResult(result);
+    UpdateAndCountBlobIfNeeded();
+    return res;
+  }
+
+  void Prev() override {
+    assert(Valid());
+
+    iter_->Prev();
+    UpdateAndCountBlobIfNeeded();
```
jay-zhuang, line 72: Seems like it could be double counted if a key is iterated twice, like: `Next()`, `Prev()`, `Next()`, is that case?

jay-zhuang (approved):
LGTM.

db/blob/blob_counting_iterator.h
ltamasi: Yes, and this is the expected/correct behavior; e.g. subcompactions do exactly what you described (they `Seek` to the first key then iterate forward from there).

db/blob/blob_counting_iterator.h
ltamasi: Well, this doesn't happen in practice because during compactions we never iterate backward but if we did, it would be the correct behavior from the POV of this class.

PR #8444, opened 2021-06-22
Deflake BlobDBTest.SnapshotAndGarbageCollection

ltamasi:
Summary:
This test case has been failing occasionally due to automatic
compactions kicking in, resulting in GC generating additional
blob files that the test did not expect. Disabling automatic
compactions to get rid of this flakiness.

Test Plan:
`gtest-parallel --repeat=1000 ./blob_db_test --gtest_filter="BlobDBTest.SnapshotAndGarbageCollection"`


jay-zhuang (approved):
LGTM

PR #8445, opened 2021-06-23
Add  support for Merge with base value during Compaction in IntegratedBlobDB

akankshamahajan15:
Summary: Provide support for Merge operation with base values during
Compaction in IntegratedBlobDB.

Test Plan: Add new unit test

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/blob/db_blob_compaction_test.cc
+++ b/db/blob/db_blob_compaction_test.cc
@@ -566,30 +566,28 @@ TEST_F(DBBlobCompactionTest, TrackGarbage) {
 
   TEST_F(DBBlobCompactionTest, MergeBlobWithBase) {
   Options options = GetDefaultOptions();
-  options.create_if_missing = true;
   options.enable_blob_files = true;
   options.min_blob_size = 0;
   options.merge_operator = MergeOperators::CreateStringAppendOperator();
   options.disable_auto_compactions = true;
 
-  DestroyAndReopen(options);
+  Reopen(options);
+  ASSERT_OK(Put("Key1", "v1_1"));
+  ASSERT_OK(Put("Key2", "v2_1"));
+  ASSERT_OK(Flush());
 
-  ASSERT_OK(Put("Key1", "v1"));
+  ASSERT_OK(Merge("Key1", "v1_2"));
+  ASSERT_OK(Merge("Key2", "v2_2"));
   ASSERT_OK(Flush());
-  MoveFilesToLevel(2);
-  ASSERT_OK(Merge("Key1", "v2"));
+
+  ASSERT_OK(Merge("Key1", "v1_3"));
   ASSERT_OK(Flush());
-  MoveFilesToLevel(2);
-  ASSERT_OK(Merge("Key1", "v3"));
-  ASSERT_OK(Flush());
-  MoveFilesToLevel(2);
 
   ASSERT_OK(db_->CompactRange(CompactRangeOptions(), /*begin=*/nullptr,
                               /*end=*/nullptr));
-
-  std::string value;
-  ASSERT_OK(db_->Get(ReadOptions(), "Key1", &value));
-  ASSERT_EQ(value, "v1,v2,v3");
+  ASSERT_EQ(Get("Key1"), "v1_1,v1_2,v1_3");
+  ASSERT_EQ(Get("Key2"), "v2_1,v2_2");
+  Close();
 }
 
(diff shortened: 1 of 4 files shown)


akankshamahajan15:
Planning to update the db_stress, db_bench and crash tests in separate PR.

ltamasi:
Thanks @akankshamahajan15 ! Will take a look.

Could you update the PR title and summary to make it clear we're talking about *blob* base values?

akankshamahajan15:
> Could you update the PR title and summary to make it clear we're talking about _blob_ base values?
Sure. Also should I add unit test for CompactionFilter with Merge also?

ltamasi:
> Sure. Also should I add unit test for CompactionFilter with Merge also?

Hm, are you referring to the chunk of code around line 290 in `merge_helper.cc`? Maybe I'm missing something but I don't think that case is possible; that branch is for merge operands (`kTypeMerge != kTypeBlobIndex`) which are never stored in blob files...

db/blob/db_blob_compaction_test.cc
```diff
@@ -390,6 +390,34 @@ TEST_F(DBBlobCompactionTest, CompactionFilterReadBlobAndKeep) {
...
 
+TEST_F(DBBlobCompactionTest, MergeBlobWithBase) {
+  Options options = GetDefaultOptions();
+  options.create_if_missing = true;
+  options.enable_blob_files = true;
+  options.min_blob_size = 0;
+  options.merge_operator = MergeOperators::CreateStringAppendOperator();
+  options.disable_auto_compactions = true;
+
+  DestroyAndReopen(options);
```
ltamasi, line 401: Minor but I think you could simply call `Reopen(options)` here and remove `options.create_if_missing = true;` above.

db/blob/db_blob_compaction_test.cc
```diff
@@ -390,6 +390,34 @@ TEST_F(DBBlobCompactionTest, CompactionFilterReadBlobAndKeep) {
...
+
+  ASSERT_OK(Put("Key1", "v1"));
+  ASSERT_OK(Flush());
+  MoveFilesToLevel(2);
```
ltamasi, line 405: Are these `MoveFilesToLevel` calls necessary? I think we could simply create a bunch of L0 files and then compact them together with a single `CompactRange` call (like the one you have below)...

db/compaction/compaction_iterator.cc
```diff
@@ -756,13 +756,19 @@ void CompactionIterator::NextFromInput() {
       }
 
       pinned_iters_mgr_.StartPinning();
+      Version* version = nullptr;
+      if (compaction_) {
+        version = compaction_->input_version();
+      }
+      BlobFetcher blob_fetcher(version, ReadOptions());
+
```
ltamasi, line 764: I feel it would be nicer to just pass the `Version` to `MergeHelper::MergeUntil`. `MergeUntil` could then instantiate `BlobFetcher` internally if needed. (`MergeUntil` returns when it hits a blob, so any `BlobFetcher` object here is single-use.)

ltamasi (commented):
Thanks @akankshamahajan15 ! LGTM in general, just a few minor comments.

db/blob/db_blob_compaction_test.cc
akankshamahajan15: I tried initially `Flush()` after each `put/merge` and then `CompactRange` in the end but then it doesn't call the required code that we want to test in `merge_helper.cc`. I didn't look in depth why so. May be I am missing some option to be set in the test. I will remove the last call  `CompactRange` as it doesn't call the required code (MoveFilesToLevel calls the code)

ltamasi: Just taking a stab in the dark here but the issue you describe might be caused by trivial moves. You could try using two keys instead of one (i.e. putting the same two keys in each file).

ltamasi: P.S. You could try running the test with `KEEP_DB=1` to confirm this.

akankshamahajan15: Ah ok. I will try this  with `KEEP_DB=1`. Thanks!

akankshamahajan15: Yes, It was because of trivial move.

ltamasi (approved):

PR #8447, opened 2021-06-23
Make the benchmark scripts support integrated blobdb

zaorangyang:
Summary:
Make the benchmark scripts support integrated blobdb.


zaorangyang:
> Thanks for the patch @zaorangyang and sorry about the delay; I missed this when you first opened the PR. I believe these issues have been addressed in the meantime (in slightly different ways) by other patches. Namely, #8915 fixed the write amp issue, and #8346 made it possible to pass arbitrary `db_bench` options to `benchmark.sh`. (The multi-phase BlobDB benchmark script I've added recently in #9015 also relies on the latter functionality.) Given the above, would you be OK with closing this?

No problem. I will close it.

PR #8449, opened 2021-06-23
Remove obsolete method VersionSet::VerifyCompactionFileConsistency

ltamasi:
Summary:
`VersionSet::VerifyCompactionFileConsistency` was superseded by the LSM tree
consistency checks introduced in https://github.com/facebook/rocksdb/pull/6901,
which are more comprehensive, more efficient, and are performed unconditionally
even in release builds.

Test Plan:
`make check`


ajkr (approved):
LGTM. `VerifyCompactionFileConsistency()` looks like it was being called too early anyways since we release mutex during the log phase of LogAndApply(). Thanks for cleaning it up.

PR #8450, opened 2021-06-23
Log the amount of blob garbage generated by compactions in the MANIFEST

ltamasi:
Summary:
The patch builds on `BlobGarbageMeter` and `BlobCountingIterator`
(introduced in #8426 and
#8443 respectively)
and ties it all together. It measures the amount of garbage
generated by a compaction and logs the corresponding `BlobFileGarbage`
records as part of the compaction job's `VersionEdit`. Note: in order
to have accurate results, `kRemoveAndSkipUntil` for compaction filters
is implemented using iteration.

Test Plan:
Ran `make check` and the crash test script.

### DIFF
--- a/db/compaction/compaction_job.cc
+++ b/db/compaction/compaction_job.cc
@@ -20,6 +20,7 @@
 #include <utility>
 #include <vector>
 
+#include "db/blob/blob_counting_iterator.h"
 #include "db/blob/blob_file_addition.h"
 #include "db/blob/blob_file_builder.h"
 #include "db/builder.h"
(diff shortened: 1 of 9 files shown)


db/compaction/compaction_job.cc
```diff
@@ -1136,6 +1147,15 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
     input = clip.get();
   }
 
+  std::unique_ptr<InternalIterator> blob_counter;
+
+  if (sub_compact->compaction->DoesInputReferenceBlobFiles()) {
```
jay-zhuang, line 1152: cool that we check it first to avoid unnecessary counting.👍

jay-zhuang (approved):
LGTM.

PR #8452, opened 2021-06-24
Performance fix pass string by ref

rockeet:


rockeet:
Is there any update?

mrambacher:
> Is there any update?

I started to look at this PR.  I am generally supportive of these changes, but am also not the final arbiter on its acceptance.

Having said that, I think this PR would be easier to ingest if it was broken into smaller chunks (89 files!!).  I would suggest making this probably 4 PRs:
- One PR for the changes that affect the public API (changes to include/rocksdb)
-  One PR for the "core" RocksDB directories (db, etc)
- One PR for the "tools"
- One PR to clean up the tests.

I understand the change is fairly minor, but it does affect a large number of files.  Splitting it up might make it easier to review and accept.

ajkr (changes requested):
It isn't generally better to pass by reference. For example, the client can invoke the function with `std::move()` or a temporary object, and the callee can further `std::move()` that object into its final destination. `CompactionServiceOutputFile` is one example where this PR breaks this pattern. Plus, it contains incompatible API changes, and no mention of motivation or explanation of performance improvement/regression.

Edit: Btw, we try to maintain the "up-for-grabs" label on issues we need help with; I'd encourage checking if any of those are of interest.

rockeet:
> It isn't generally better to pass by reference. For example, the client can invoke the function with `std::move()` or a temporary object, and the callee can further `std::move()` that object into its final destination. `CompactionServiceOutputFile` is one example where this PR breaks this pattern. Plus, it contains incompatible API changes, and no mention of motivation or explanation of performance improvement/regression.
> 
> Edit: Btw, we try to maintain the "up-for-grabs" label on issues we need help with; I'd encourage checking if any of those are of interest.

std::move is not always appropriate:
1. It is appropriate just for functions just forwarding params to another function or such param just simply be used once.
2. std::move was not really be used on appropriate cases.
3. std::move value param is not as cheap as reference, for example std::string in gcc is 32 bytes, and move ctor + dtor is not as cheap as we considered.
4. although performance has hotspots, performance also has scattered spots, we are easy to notice and optimize hotspots and easy to ignore scattered spots.

rockeet:
> > Is there any update?
> 
> I started to look at this PR. I am generally supportive of these changes, but am also not the final arbiter on its acceptance.
> 
> Having said that, I think this PR would be easier to ingest if it was broken into smaller chunks (89 files!!). I would suggest making this probably 4 PRs:
> 
> * One PR for the changes that affect the public API (changes to include/rocksdb)
> * One PR for the "core" RocksDB directories (db, etc)
> * One PR for the "tools"
> * One PR to clean up the tests.
> 
> I understand the change is fairly minor, but it does affect a large number of files. Splitting it up might make it easier to review and accept.

Seperate to multiple PRs is feasible, I'm happy to do it if such PRs are planed to be accepted.

mrambacher:
> > It isn't generally better to pass by reference. For example, the client can invoke the function with `std::move()` or a temporary object, and the callee can further `std::move()` that object into its final destination. `CompactionServiceOutputFile` is one example where this PR breaks this pattern. Plus, it contains incompatible API changes, and no mention of motivation or explanation of performance improvement/regression.
> > Edit: Btw, we try to maintain the "up-for-grabs" label on issues we need help with; I'd encourage checking if any of those are of interest.
> 
> std::move is not always appropriate:
> 
> 1. It is appropriate just for functions just forwarding params to another function or such param just simply be used once and its value was really be `moved` to somewhere.
> 2. std::move was not really be used on appropriate cases.
> 3. after calling `std::move(src)` in a context, `src` may be really cleared, or not really be cleared, `src` can be legally used after `std::move(src)`, and c++ compiler is not mandated to check such use cases(`-Wextra` does not issue warn --- may be some static analysis tools can check it), if there is a such bug(src is conditionally `really` moved), it is hard to be located.
> 4. `std::move` `pass by value` param is not as cheap as reference, for example std::string in gcc is 32 bytes, with sso, move ctor + dtor is not as cheap as we considered.
> 5. although performance has hotspots, performance also has scattered spots, we are easy to notice and optimize hotspots and likely to ignore scattered spots.
> 
> for `CompactionServiceOutputFile`, the best param types should be `std::string&&`, I have fixed it.

I tend to agree that const& should be better than std::move in general.  It makes a cleaner interface and is simpler and easier to enforce/implement.  I prefer it if for nothing else "code hygiene".  

Having said that, if this is really a performance change, are there any metrics you can generate that show any of these changes improve performance?  Maybe a very targeted case with a higher perf level measurement or something?

ajkr:
> > > It isn't generally better to pass by reference. For example, the client can invoke the function with `std::move()` or a temporary object, and the callee can further `std::move()` that object into its final destination. `CompactionServiceOutputFile` is one example where this PR breaks this pattern. Plus, it contains incompatible API changes, and no mention of motivation or explanation of performance improvement/regression.
> > > Edit: Btw, we try to maintain the "up-for-grabs" label on issues we need help with; I'd encourage checking if any of those are of interest.
> > 
> > 
> > std::move is not always appropriate:
> > 
> > 1. It is appropriate just for functions just forwarding params to another function or such param just simply be used once and its value was really be `moved` to somewhere.
> > 2. std::move was not really be used on appropriate cases.
> > 3. after calling `std::move(src)` in a context, `src` may be really cleared, or not really be cleared, `src` can be legally used after `std::move(src)`, and c++ compiler is not mandated to check such use cases(`-Wextra` does not issue warn --- may be some static analysis tools can check it), if there is a such bug(src is conditionally `really` moved), it is hard to be located.
> > 4. `std::move` `pass by value` param is not as cheap as reference, for example std::string in gcc is 32 bytes, with sso, move ctor + dtor is not as cheap as we considered.
> > 5. although performance has hotspots, performance also has scattered spots, we are easy to notice and optimize hotspots and likely to ignore scattered spots.
> > 
> > for `CompactionServiceOutputFile`, the best param types should be `std::string&&`, I have fixed it.
> 
> I tend to agree that const& should be better than std::move in general. It makes a cleaner interface and is simpler and easier to enforce/implement. I prefer it if for nothing else "code hygiene".

I am not sure what "in general" means here (on average or in every case). Clearly it can't be better in every case since we just pointed out a case where const& is worse. I meant it in the latter way.

The burden to prove the change is an improvement should be the PR author's. I just picked an example from recent memory to disprove this change is generally beneficial. That doesn't mean the rest of the changes are good or that we want to spend our time re-examining them (for no apparent motivation).

Issue #109, opened 2014-04-07
official node.js binding


matthewmueller:
would love to see an official rockdb binding for node.js :-)

right now our only option is leveldb

PR #8454, opened 2021-06-24
Memtable "MemPurge" prototype

bjlemaire:
Implement an experimental feature called "MemPurge", which consists in purging "garbage" bytes out of a memtable and reuse the memtable struct instead of making it immutable and eventually flushing its content to storage.
The prototype is by default deactivated and is not intended for use. It is intended for correctness and validation testing. At the moment, the "MemPurge" feature can be switched on by using the `options.experimental_allow_mempurge` flag. For this early stage, when the allow_mempurge flag is set to `true`, all the flush operations will be rerouted to perform a MemPurge. This is a temporary design decision that will give us the time to explore meaningful heuristics to use MemPurge at the right time for relevant workloads . Moreover, the current MemPurge operation only supports `Puts`, `Deletes`, `DeleteRange` operations, and handles `Iterators` as well as `CompactionFilter`s that are invoked at flush time .
Three unit tests are added to `db_flush_test.cc` to test if MemPurge works correctly (and checks that the previously mentioned operations are fully supported thoroughly tested). 
One noticeable design decision is the timing of the MemPurge operation in the memtable workflow: for this prototype, the mempurge happens when the memtable is switched (and usually made immutable). This is an inefficient process because it implies that the entirety of the MemPurge operation happens while holding the db_mutex. Future commits will make the MemPurge operation a background task (akin to the regular flush operation) and aim at drastically enhancing the performance of this operation. The MemPurge is also not fully "WAL-compatible" yet, but when the WAL is full, or when the regular MemPurge operation fails (or when the purged memtable still needs to be flushed), a regular flush operation takes place. Later commits will also correct these behaviors.


db/db_impl/db_impl_write.cc
```diff
@@ -1933,11 +2098,43 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
   }
 
   cfd->mem()->SetNextLogNumber(logfile_number_);
-  cfd->imm()->Add(cfd->mem(), &context->memtables_to_free_);
+  // If MemPurge activated, purge and delete current memtable.
+  if (immutable_db_options_.experimental_allow_mempurge) {
+    s = MemPurge(cfd, new_mem);
+    if (s.ok()) {
```
mrambacher, line 2104: If MemPurge fails, are you intending to return that error up the chain?  Or should s be a temporary/local variable?

If it is meant to be returned, please add a comment as such -- since there is code below that looks like it is meant to fallback, it looks as if the error is local. 

If the result is not meant to be returned, use a variable other than "s".

bjlemaire: Good catch - the status was not meant to be returned. I will create another Status variable instead in the next commit.

db/db_flush_test.cc
```diff
@@ -658,6 +662,382 @@ TEST_F(DBFlushTest, StatisticsGarbageRangeDeletes) {
...
+  std::string value;
+  const std::string NOT_FOUND = "NOT_FOUND";
+
+  // Check simple operations (put-delete).
+  ASSERT_OK(Put(KEY1, VALUE1));
+  ASSERT_OK(Put(KEY2, VALUE2));
+  ASSERT_OK(Delete(KEY1));
+  ASSERT_OK(Put(KEY2, VALUE1));
+  ASSERT_OK(Put(KEY1, VALUE2));
+  ASSERT_OK(Flush());
```
akankshamahajan15, line 716: If user calls manual ```Flush()```, will the flush leads to Memtable purge or to storage?

db/db_flush_test.cc
```diff
@@ -658,6 +662,382 @@ TEST_F(DBFlushTest, StatisticsGarbageRangeDeletes) {
...
+    p_v3 = rnd.RandomString(RAND_VALUES_LENGTH);
+    p_v4 = rnd.RandomString(RAND_VALUES_LENGTH);
+    p_v5 = rnd.RandomString(RAND_VALUES_LENGTH);
+
+    ASSERT_OK(Put(KEY1, p_v1));
+    ASSERT_OK(Put(KEY2, p_v2));
+    ASSERT_OK(Put(KEY3, p_v3));
+    ASSERT_OK(Put(KEY4, p_v4));
+    ASSERT_OK(Put(KEY5, p_v5));
+    value = Get(KEY1);
```
akankshamahajan15, line 751: Any reason to store the ```Get``` value in variable? You can directly call ```ASSERT_EQ(Get(KEY1), p_v1);```. It will reduce lines of code (Same in other places).

db/db_impl/db_impl.h
```diff
@@ -1607,6 +1609,8 @@ class DBImpl : public DB {
 
   Status SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context);
 
+  Status MemPurge(ColumnFamilyData* cfd, MemTable* new_mem);
```
akankshamahajan15, line 1612: Its good to add a little description here about the API.

db/db_impl/db_impl_write.cc
```diff
@@ -1737,6 +1737,178 @@ void DBImpl::NotifyOnMemTableSealed(ColumnFamilyData* /*cfd*/,
...
+      CompactionFilter::Context ctx;
+      ctx.is_full_compaction = false;
+      ctx.is_manual_compaction = false;
+      ctx.column_family_id = cfd->GetID();
+      ctx.reason = TableFileCreationReason::kFlush;
+      compaction_filter =
+          ioptions->compaction_filter_factory->CreateCompactionFilter(ctx);
+      if (compaction_filter != nullptr &&
+          !compaction_filter->IgnoreSnapshots()) {
+        s.PermitUncheckedError();
```
akankshamahajan15, line 1801: you can assign the ```NotSupported``` status to ```s``` and ```return s```. That way you don't need to add ```s.PermitUncheckedError();```. PermitUncheckedError is added if we are sure the status is not checked.

include/rocksdb/statistics.h
```diff
@@ -388,6 +388,11 @@ enum Tickers : uint32_t {
   MEMTABLE_PAYLOAD_BYTES_AT_FLUSH,
   // Outdated bytes of data present on memtable at flush time.
   MEMTABLE_GARBAGE_BYTES_AT_FLUSH,
+  // Counter for how many times the MemPurge operation is performed.
```
akankshamahajan15, line 391: If users cannot use this its better to add experimental otherwise it's fine.

akankshamahajan15 (commented):
I have added few comments. Rest leave it to @anand1976  for the review.

db/column_family.cc
```diff
@@ -443,14 +443,23 @@ bool SuperVersion::Unref() {
   return previous_refs == 1;
 }
 
-void SuperVersion::Cleanup() {
+void SuperVersion::Cleanup(bool fromMemPurge) {
```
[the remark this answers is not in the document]
bjlemaire: Good catch, I can add the "const" keyword here.

db/db_flush_test.cc
bjlemaire: @anand1976  also brought this up during our meeting last time - I intend to create a MemFlush API (Milestone 1B) so that if a user calls Flush() manually, the flush will lead to storage.

db/db_flush_test.cc
bjlemaire: Thanks for the tip! I'll edit these lines of code right now.

include/rocksdb/statistics.h
bjlemaire: Great idea - done

db/db_impl/db_impl_write.cc
```diff
@@ -1737,6 +1737,178 @@ void DBImpl::NotifyOnMemTableSealed(ColumnFamilyData* /*cfd*/,
...
+    // Key transfer
+    for (; c_iter.Valid(); c_iter.Next()) {
+      const ParsedInternalKey ikey = c_iter.ikey();
+      const Slice value = c_iter.value();
+      new_earliest_seqno = ikey.sequence < new_earliest_seqno
+                               ? ikey.sequence
+                               : new_earliest_seqno;
+      new_first_seqno =
+          ikey.sequence < new_first_seqno ? ikey.sequence : new_first_seqno;
+      new_mem->SetFirstSequenceNumber(new_first_seqno);
```
anand1976, line 1848: Can we track ```new_first_seqno``` and call ```SetFirstSequenceNumber``` in the end since its an atomic store?

db/db_impl/db_impl_write.cc
```diff
@@ -1933,11 +2107,44 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
   }
 
   cfd->mem()->SetNextLogNumber(logfile_number_);
-  cfd->imm()->Add(cfd->mem(), &context->memtables_to_free_);
+  // If MemPurge activated, purge and delete current memtable.
+  if (immutable_db_options_.experimental_allow_mempurge &&
```
anand1976, line 2111: I think this PR doesn't handle the WAL full case right?  When WAL is full, we may want to skip the purge and flush the memtable anyway.

db/db_impl/db_impl_write.cc
```diff
@@ -1737,6 +1737,178 @@ void DBImpl::NotifyOnMemTableSealed(ColumnFamilyData* /*cfd*/,
...
+        if (!s.ok()) {
+          break;
+        }
+      }
+    }
+  }
+  // Note: if the mempurge was ineffective, meaning that there was no
+  // garbage to remove, and this new_mem needs to be flushed again,
+  // the new_mem->Add would have updated the flush status when it
+  // called "UpdateFlushState()" internally at the last Add() call.
```
anand1976, line 1908: I'm not clear on the flow if the mempurge was ineffective. How do we prevent an endless loop of purge attempts of the same memtable data?

db/db_impl/db_impl_write.cc
bjlemaire: Correct, I left all WAL handlings for later (Milestone 1C, "Getting MemPurge to work with WAL"). However following your comment I just added a check on the FlushReason so that if the flush operation happens because of a WAL full or else, the MemPurge operation is not triggered and a regular flush operation will take place instead.

db/db_impl/db_impl_write.cc
bjlemaire: Good call, I just made the edit - thanks for the catch

db/db_impl/db_impl_write.cc
bjlemaire: You are right, I forgot to add an error status. 
What is happening now is at the end of the MemPurge function I check if the new memtable needs to be flushed again. If this is the case, I change the return status such that when the MemPurge function returns, the error status is noticed, the mempurge is aborted, and the switchMemtable process is rerouted to a regular flush operation (see db/db_impl/db_impl_write.cc:2119).

db/column_family.cc
```diff
@@ -443,14 +443,23 @@ bool SuperVersion::Unref() {
...
   MemTable* m = mem->Unref();
   if (m != nullptr) {
-    auto* memory_usage = current->cfd()->imm()->current_memory_usage();
-    assert(*memory_usage >= m->ApproximateMemoryUsage());
-    *memory_usage -= m->ApproximateMemoryUsage();
+    // In MemPurge, the memtable is not made immutable.
+    // Therefore it does not contribute to the imm
+    // memory usage (and actually is not part of
+    // the 'imm' MemtableList at all).
+    if (!fromMemPurge) {
```
anand1976, line 458: I'm finding this a bit hard to reason about. At some point, the mutable memtable will be added to the immutable list (if mempurge fails or is ineffective in reclaiming space), right? Can we fix the memory usage accounting in ```SwitchMemtable()``` so this code is not impacted?

db/column_family.h
```diff
@@ -222,7 +222,7 @@ struct SuperVersion {
   // Cleanup unrefs mem, imm and current. Also, it stores all memtables
   // that needs to be deleted in to_delete vector. Unrefing those
   // objects needs to be done in the mutex
-  void Cleanup();
+  void Cleanup(const bool fromMemPurge = false);
```
pdillinger, line 225: From Google C++ style:

  * For a function parameter passed by value, const has no effect on the caller, thus is not recommended in function declarations. See [TotW \#109](https://abseil.io/tips/109).

db/db_impl/db_impl_write.cc
```diff
@@ -1933,11 +2113,46 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
   }
 
   cfd->mem()->SetNextLogNumber(logfile_number_);
-  cfd->imm()->Add(cfd->mem(), &context->memtables_to_free_);
+  // If MemPurge activated, purge and delete current memtable.
+  if (immutable_db_options_.experimental_allow_mempurge &&
+      (new_mem != nullptr) &&
+      ((cfd->GetFlushReason() == FlushReason::kOthers) ||
+       (cfd->GetFlushReason() == FlushReason::kManualFlush))) {
+    Status mempurge_s = MemPurge(cfd, new_mem);
```
pdillinger, line 2121: This is walking the memtable while holding the DB mutex. In production this would lead to many DB operations blocking for potentially a second or more during purge. This is why it needs to be a variant of (background) flush. At least eventually

pdillinger: Offline discussion: This limitation will be noted in the PR summary, indicating the intended direction of future work.

db/column_family.h
bjlemaire: This website is a gold mine, thanks for the reference. I will remove the const keyword from the function declaration in the next commit.

db/column_family.cc
bjlemaire: You are 100% correct, I will fix this by setting the "fromMemPurge" as false if the mempurge fails or is ineffective in reclaiming space. I am also renaming this bool as "noImmMemoryContribution" to make this variable more versatile and explicit.

bjlemaire: For future reference, I am realizing now that I can remove this flag once the `MemPurge` is made a background task for better performance (next PR), so this is a temporary edit to make the Cleanup code compatible with this MemPurge prototype.

anand1976: ```SuperVersion::Cleanup``` is called in the context of the last deref, which may not necessarily be the mempurge context. Can we instead increment ```current->cfd()->imm()->current_memory_usage()``` by the purged memtable size in ```SwitchMemtable```, so the last deref of that memtable doesn't have to worry about the purge status?

bjlemaire: There is no easy way to increment the `imm()->current_memory_usage()` without adding to the `imm` MemtableList. One option would be to create a new method in the `MemTableList` class, e.g "add_memory_usage", but it is not clear to me that this is a more viable option.

anand1976 (approved):
LGTM!

include/rocksdb/statistics.h
pdillinger: Except this is among the most public of public APIs. Because of lots of downstream infrastructure built on these statistics, AFAIK additions are quite permanent. I'm not sure of all the consequences of changing a name.

pdillinger:
Annoyingly msvc requires us to fix signed/unsigned mismatch (e.g. int vs. uint32_t) in ASSERT_EQ, EXPECT_EQ, EXPECT_GE, etc. without actually pointing to the code using the macro in a problematic way. Otherwise looks good.

PR #8456, opened 2021-06-24
Implement missing Handler methods in ColumnFamilyCollector.

pydima:
When db is open as secondary, there are basically 2 step process:
1) Collect column families from wal log
2) Apply changes to Memtable
In case primary db is TransactionDB instance, wal log will contain some additional data, like noop, etc. ColumnFamilyCollector doesn't implement methods to handle these, so it fails to open a wal log written by TransactionDB. (Everything works fine with standard DB::Open).
Memtable recovery process knows how to handle such wal logs, so only missing piece seems to be ColumnFamilyCollector.


pydima:
I appreciate that everyone is super busy (and thanks a lot for working on this awesome project), but could anyone take a look at this PR, please. We really miss this functionality. Thanks.
@zhichao-cao @jay-zhuang @ajkr (just pinging the most active contributors, sorry to disturb).

mrambacher (approved):
LGTM.  Thanks for the fix!

pydima:
Hi @mrambacher, sorry to disturb, just wanted to check this PR status. I see that there are some internal tests failure. Is there anything I can help with? Thanks.

pydima:
I've fixed the format issues, which seems to cause fb internal test failures. @mrambacher

pydima:
ping @mrambacher

pydima:
I'm really sorry for the noise, but can anybody merge or decline this PR, it feels like no progress and it's stuck. @siying @ajkr @jay-zhuang maybe someone can help here. Thanks.

ajkr (approved):
Seems right. Looks like secondary DB would still fail with prepared transactions (as it should).

PR #8457, opened 2021-06-24
Fix Immutable Customizable Serialization

mrambacher:
If a Customizable option was not mutable, it would still appear in the list of mutable options when serialized. This meant that when the immutable options were used to configure another immutable object, an "option not changeable" status would be returned.


zhichao-cao (approved):
LGTM

PR #8465, opened 2021-06-25
Partially revert the "apply subrange of table property collectors" change

ltamasi:
Summary:
We ended up using a different approach for tracking the amount of
garbage in blob files (see e.g. https://github.com/facebook/rocksdb/pull/8450),
so the ability to apply only a range of table property collectors is
now unnecessary. The patch reverts this part of
https://github.com/facebook/rocksdb/pull/8298 while keeping the cleanup done
in that PR.

Test Plan:
`make check`


jay-zhuang (approved):
LGTM

PR #8468, opened 2021-06-29
Fix PrepareOptions for Customizable Classes

mrambacher:
Added the Customizable::ConfigureNewObject method.  The method will configure the object if options are found and invoke PrepareOptions if the flag is set properly.

Added tests to test that PrepareOptions is properly called and to test if PrepareOptions fails.


zhichao-cao (approved):
LGTM, add a sentence to HISTORY.md to remind adding ConfigueNewObject.

PR #8469, opened 2021-06-29
Call OnCompactionCompleted API in case of DisableManualCompaction

akankshamahajan15:
Summary: Call OnCompactionCompleted API in case of 
DisableManualCompaction() with updated Status::Incomplete

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -1395,6 +1395,8 @@ void DBImpl::NotifyOnCompactionBegin(ColumnFamilyData* cfd, Compaction* c,
       manual_compaction_paused_.load(std::memory_order_acquire) > 0) {
     return;
   }
+
+  c->SetNotifyOnCompactionCompleted();
   Version* current = cfd->current();
   current->Ref();
   // release lock while notifying events
@@ -1430,10 +1432,8 @@ void DBImpl::NotifyOnCompactionCompleted(
   if (shutting_down_.load(std::memory_order_acquire)) {
     return;
   }
-  // TODO: Should disabling manual compaction squash compaction completed
-  //   notifications that aren't the result of a shutdown?
-  if (c->is_manual_compaction() &&
-      manual_compaction_paused_.load(std::memory_order_acquire) > 0) {
+
+  if (c->ShouldNotifyOnCompactionCompleted() == false) {
     return;
   }
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,7 @@
 ### Bug Fixes
 * Blob file checksums are now printed in hexadecimal format when using the `manifest_dump` `ldb` command.
 * `GetLiveFilesMetaData()` now populates the `temperature`, `oldest_ancester_time`, and `file_creation_time` fields of its `LiveFileMetaData` results when the information is available. Previously these fields always contained zero indicating unknown.
+* Fix mismatches of OnCompaction{Begin,Completed} in case of DisableManualCompaction().
 
 ### New Features
 * ldb has a new feature, `list_live_files_metadata`, that shows the live SST files, as well as their LSM storage level and the column family they belong to.
--- a/db/compaction/compaction.cc
+++ b/db/compaction/compaction.cc
@@ -237,7 +237,8 @@ Compaction::Compaction(
       is_full_compaction_(IsFullCompaction(vstorage, inputs_)),
       is_manual_compaction_(_manual_compaction),
       is_trivial_move_(false),
-      compaction_reason_(_compaction_reason) {
+      compaction_reason_(_compaction_reason),
+      notify_on_compaction_completion_(false) {
   MarkFilesBeingCompacted(true);
   if (is_manual_compaction_) {
     compaction_reason_ = CompactionReason::kManualCompaction;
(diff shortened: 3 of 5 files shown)


db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1444,7 +1438,12 @@ void DBImpl::NotifyOnCompactionCompleted(
   TEST_SYNC_POINT("DBImpl::NotifyOnCompactionCompleted::UnlockMutex");
   {
     CompactionJobInfo info{};
-    BuildCompactionJobInfo(cfd, c, st, compaction_job_stats, job_id, current,
+    Status s = st;
+    if (s.ok() && c->is_manual_compaction() &&
```
ajkr, line 1442: Hm, I checked the client code and it looks like `st` should already contain some kind of `Status::Incomplete` when it fails due to disabled manual compaction. It's also a bit late here to check since `DisableManualCompaction()` might have been called after a compaction succeeded but before this point.

ajkr (commented):
Also do you mind adding a "Bug Fixes" item to HISTORY.md?

Another thing I thought about in the distant past is making `db_stress` have a custom `EventListener` that checks OnCompaction{Begin,End}() match. A difficulty was that program usually ends by intentionally crashing at which point it's hard to say what compactions had their `OnCompactionEnd()` missed (some compactions may truly still be running). 

A workaround I just thought of is set a thread-local flag in `OnCompactionBegin()`, and reset it in `OnCompactionEnd()`. If the flag is found to be already set when `OnCompactionBegin()` runs, that means the previous compaction in that thread forgot its `OnCompactionEnd()`, and we should report a failure/terminate early. To be really advanced we could check all the flags are reset after clean shutdown.

akankshamahajan15:
> Also do you mind adding a "Bug Fixes" item to HISTORY.md?
> 
> Another thing I thought about in the distant past is making `db_stress` have a custom `EventListener` that checks OnCompaction{Begin,End}() match. A difficulty was that program usually ends by intentionally crashing at which point it's hard to say what compactions had their `OnCompactionEnd()` missed (some compactions may truly still be running).
> 
> A workaround I just thought of is set a thread-local flag in `OnCompactionBegin()`, and reset it in `OnCompactionEnd()`. If the flag is found to be already set when `OnCompactionBegin()` runs, that means the previous compaction in that thread forgot its `OnCompactionEnd()`, and we should report a failure/terminate early. To be really advanced we could check all the flags are reset after clean shutdown.

Sure SG. I will add it in the db_stress.

ajkr:
> > Also do you mind adding a "Bug Fixes" item to HISTORY.md?
> > Another thing I thought about in the distant past is making `db_stress` have a custom `EventListener` that checks OnCompaction{Begin,End}() match. A difficulty was that program usually ends by intentionally crashing at which point it's hard to say what compactions had their `OnCompactionEnd()` missed (some compactions may truly still be running).
> > A workaround I just thought of is set a thread-local flag in `OnCompactionBegin()`, and reset it in `OnCompactionEnd()`. If the flag is found to be already set when `OnCompactionBegin()` runs, that means the previous compaction in that thread forgot its `OnCompactionEnd()`, and we should report a failure/terminate early. To be really advanced we could check all the flags are reset after clean shutdown.
> 
> Sure SG. I will add it in the db_stress.

Great! I can accept this PR independently of that with the other small suggestions if you prefer.

akankshamahajan15:
> > > Also do you mind adding a "Bug Fixes" item to HISTORY.md?
> > > Another thing I thought about in the distant past is making `db_stress` have a custom `EventListener` that checks OnCompaction{Begin,End}() match. A difficulty was that program usually ends by intentionally crashing at which point it's hard to say what compactions had their `OnCompactionEnd()` missed (some compactions may truly still be running).
> > > A workaround I just thought of is set a thread-local flag in `OnCompactionBegin()`, and reset it in `OnCompactionEnd()`. If the flag is found to be already set when `OnCompactionBegin()` runs, that means the previous compaction in that thread forgot its `OnCompactionEnd()`, and we should report a failure/terminate early. To be really advanced we could check all the flags are reset after clean shutdown.
> > 
> > 
> > Sure SG. I will add it in the db_stress.
> 
> Great! I can accept this PR independently of the db_stress change with the other small suggestions if you prefer.

Yeah I think that would be better. In case I mess up the stress tests, it's easy to revert that change.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1430,10 +1432,8 @@ void DBImpl::NotifyOnCompactionCompleted(
   if (shutting_down_.load(std::memory_order_acquire)) {
     return;
   }
-  // TODO: Should disabling manual compaction squash compaction completed
-  //   notifications that aren't the result of a shutdown?
-  if (c->is_manual_compaction() &&
-      manual_compaction_paused_.load(std::memory_order_acquire) > 0) {
+
+  if (c->ShouldNotifyOnCompactionCompleted() == false) {
```
ajkr, line 1436: Is this ever the case? Or is this left over from when I (incorrectly) claimed there is a bug where bottom-pri compactions see an extra `OnCompactionCompleted()`?

ajkr (approved):
LGTM, thanks very much, especially for the test cases!

db/db_impl/db_impl_compaction_flush.cc
ajkr: Never mind this feels like Case II in the tests.

PR #8470, opened 2021-06-29
Fix assertion failure when releasing a handle after secondary cache lookup fails

anand1976:
When the secondary cache lookup fails, we may still allocate a handle and charge the cache for metadata usage. If the cache is full, this can cause the usage to go over capacity. Later, when a (unrelated) handle is released, it trips up an assertion that checks that usage is less than capacity. To prevent this assertion failure, don't charge the cache for a failed secondary cache lookup.

Tests:
Run crash_test

### DIFF
--- a/cache/lru_cache.cc
+++ b/cache/lru_cache.cc
@@ -407,11 +407,10 @@ void LRUCacheShard::Promote(LRUHandle* e) {
     }
   } else {
     // Since the secondary cache lookup failed, mark the item as not in cache
-    // and charge the cache only for metadata usage, i.e handle, key etc
+    // Don't charge the cache as its only metadata that'll shortly be released
     MutexLock l(&mutex_);
     e->charge = 0;
     e->SetInCache(false);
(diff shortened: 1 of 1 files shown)


cache/lru_cache.cc
```diff
@@ -407,11 +407,10 @@ void LRUCacheShard::Promote(LRUHandle* e) {
     }
   } else {
     // Since the secondary cache lookup failed, mark the item as not in cache
-    // and charge the cache only for metadata usage, i.e handle, key etc
+    // Don't charge the cache as its only metadata that'll shortly be released
```
zhichao-cao, line 410: Is this implementation specific ? Maybe other secondary cache implementation will not allocate a handle and charge the cache for metadata usage when it fails. This can be a short term fix. Maybe later on, we can make some change to the secondary interface and at least the secondary cache can tell RocksDB if it charged or not.

zhichao-cao (approved):
LGTM, thanks for the fix.

PR #8471, opened 2021-06-29
Enable crash test to run using fbcode components

anand1976:
Add a new test ```fbcode_crash_test``` to rocksdb-lego-determinator. This test allows the crash test to be run on Facebook Sandcastle infra using fbcode components. Also use the default Env in db_stress to access the expected values path as it requires a memory mapped file and may not work with custom Envs.

### DIFF
--- a/build_tools/rocksdb-lego-determinator
+++ b/build_tools/rocksdb-lego-determinator
@@ -1195,6 +1195,40 @@ JAVA_BUILD_TEST_COMMANDS="[
     }
 ]"
 
+#
+# RocksDB fbcode stress/crash test
+#
+FBCODE_STRESS_CRASH_TEST_COMMANDS="[
+    {
+        'name':'Rocksdb Fbcode Stress and Crash Test',
+        'oncall':'$ONCALL',
+        'executeLocal': 'true',
+        'timeout': 86400,
+        'steps': [
+            {
+                'name':'Copy RocksDB code to fbcode repo',
+                'shell':'cd $WORKING_DIR; cp -R internal_repo_rocksdb/repo/* rocksdb/src && cp -R rocksdb/src/include/rocksdb/* rocksdb  || $CONTRUN_NAME=db_stress_fbcode $TASK_CREATION_TOOL',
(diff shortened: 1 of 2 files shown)


db_stress_tool/db_stress_shared_state.h
```diff
@@ -58,7 +58,7 @@ class SharedState {
   static bool ignore_read_error;
 #endif // ROCKSDB_SUPPORT_THREAD_LOCAL
 
-  SharedState(Env* env, StressTest* stress_test)
+  SharedState(Env* /*env*/, StressTest* stress_test)
```
zhichao-cao, line 61: Can we give a flag to decide whether the customized env will be used or not?

anand1976: We could do that, but I wonder if there's any advantage in storing the expected values file using the customized env? Why not always use the default env? Its unlikely we'll want to do fault injection or anything like that on the expected values file.

ajkr (approved):
LGTM

zhichao-cao (approved):
LGTM

build_tools/rocksdb-lego-determinator
```diff
@@ -1195,6 +1195,40 @@ JAVA_BUILD_TEST_COMMANDS="[
...
+FBCODE_STRESS_CRASH_TEST_COMMANDS="[
+    {
+        'name':'Rocksdb Fbcode Stress and Crash Test',
+        'oncall':'$ONCALL',
+        'executeLocal': 'true',
+        'timeout': 86400,
+        'steps': [
+            {
+                'name':'Copy RocksDB code to fbcode repo',
+                'shell':'cd $WORKING_DIR; cp -R internal_repo_rocksdb/repo/* rocksdb/src && cp -R rocksdb/src/include/rocksdb/* rocksdb  || $CONTRUN_NAME=db_stress_fbcode $TASK_CREATION_TOOL',
```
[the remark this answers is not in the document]
anand1976: Yes, that's a good idea.

PR #8473, opened 2021-06-30
Make EventListener into a Customizable Class

mrambacher:
- Added Type/CreateFromString
- Added ability to load EventListeners to DBOptions
- Since EventListeners did not previously have a Name(), defaulted to "".  If there is no name, the listener cannot be loaded from the ObjectRegistry.


options/db_options.cc
```diff
@@ -433,6 +433,61 @@ static std::unordered_map<std::string, OptionTypeInfo>
          {offsetof(struct ImmutableDBOptions, allow_data_in_errors),
           OptionType::kBoolean, OptionVerificationType::kNormal,
           OptionTypeFlags::kNone}},
+        {"listeners",
```
zhichao-cao, line 436: Add a comment here to indicate that different EventListener can be create from an option string and also we can serialize it. It would be better to give an example of the string input and the serialized output in the comment.

zhichao-cao (approved):
LGTM, may be mention that make EventListener in the Customizable class in the HISTORY.md.

PR #8475, opened 2021-06-30
Move slow valgrind tests behind -DROCKSDB_FULL_VALGRIND_RUN

ajkr:
Various tests had disabled valgrind due to it slowing down and timing
out (as is the case right now) the CI runs. Where a test was disabled with no comment,
I assumed slowness was the cause. For these tests that were slow under
valgrind, as well as the ones identified in #8352, this PR moves them
behind the compiler flag `-DROCKSDB_FULL_VALGRIND_RUN`.

Test Plan: running `make full_valgrind_test`, `make valgrind_test`, `make check`; will verify they appear working correctly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -344,6 +344,12 @@ ifdef ROCKSDB_VALGRIND_RUN
 	PLATFORM_CCFLAGS += -DROCKSDB_VALGRIND_RUN
 	PLATFORM_CXXFLAGS += -DROCKSDB_VALGRIND_RUN
 endif
+ifdef ROCKSDB_FULL_VALGRIND_RUN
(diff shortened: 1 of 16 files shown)


jay-zhuang (approved):
LGTM.
Just curious long does full valgrind test take?

ajkr:
Good question, I forget if it ran to completion. Running again now.

ajkr:
@jay-zhuang 64 minutes at -j24. Our sandcastle contrun uses -j1 however. Timeout is 6 hours. Maybe we should try -j8 although idk why -j1 was chosen in the first place.

PR #8478, opened 2021-07-01
remove redundant std::move

rockeet:


zhichao-cao (commented):
It is not redundant, we need it to avoid the status check

rockeet:
> It is not redundant, we need it to avoid the status check

It should be expressed explicitly, I've fixed it

zhichao-cao:
@ajkr you may take a look?

ajkr:
Doesn't matter to me. It appears to do the same work in a slightly more roundabout and slightly less suspicious-looking way.

include/rocksdb/status.h
```diff
@@ -457,10 +457,10 @@ class Status {
   Code code_;
   SubCode subcode_;
   Severity sev_;
-  const char* state_;
 #ifdef ROCKSDB_ASSERT_STATUS_CHECKED
   mutable bool checked_ = false;
 #endif  // ROCKSDB_ASSERT_STATUS_CHECKED
+  const char* state_;
```
zhichao-cao, line 463: Why we move the line here?

rockeet: move the line here eliminate paddings and reduces sizeof(Status) by 8 bytes when ROCKSDB_ASSERT_STATUS_CHECKED is defined.

include/rocksdb/io_status.h
```diff
@@ -266,4 +266,8 @@ inline IOStatus status_to_io_status(Status&& status) {
   }
 }
 
+inline Status io_status_to_status(IOStatus& io_s) {
```
zhichao-cao, line 269: @anand1976 Are you ok this this change?

anand1976: Seems ok to me. Can we make it ```io_statuc_to_status(IOStatus&& io_s)```? I think doing ```std::move``` on a lvalue reference is a bit of an anti-pattern.

db/db_impl/db_impl.cc
```diff
@@ -1348,7 +1348,7 @@ Status DBImpl::LockWAL() {
     // future writes
     WriteStatusCheck(status);
   }
-  return std::move(status);
+  return io_status_to_status(status);
```
anand1976, line 1351: Can we make the type of ```status``` explicit in line 1343?

include/rocksdb/io_status.h
pdillinger: This seems unnecessary (and violating Google C++ style on names). What's wrong with

    Status(io_s)

or

    static_cast<const Status&>(io_s);

? If you're going to motivate use of move semantics, I wouldn't use a case where it's eligible for copy elision. https://en.cppreference.com/w/cpp/language/copy_elision

PR #8479, opened 2021-07-01
 Inject fatal write failures to db_stress when DB is running

zhichao-cao:
add the injest_error_severity to control if it is a retryable IO Error or a fatal or unrecoverable error. Use a flag to indicate, if fatal error comes, the flag is set and db is stopped (but not corrupted).

test plan: run  ./db_stress --reopen=0 --read_fault_one_in=1000 --write_fault_one_in=5 --disable_wal=true --write_buffer_size=3000000 -writepercent=5 -readpercent=50 --injest_error_severity=2 --column_families=1, make check

### DIFF
--- a/db_stress_tool/db_stress_gflags.cc
+++ b/db_stress_tool/db_stress_gflags.cc
@@ -827,10 +827,9 @@ DEFINE_string(secondary_cache_uri, "",
 DEFINE_int32(open_write_fault_one_in, 0,
              "On non-zero, enables fault injection on file write "
              "during DB reopen.");
-
 DEFINE_int32(injest_error_severity, 1,
              "The severity of the injested IO Error. 1 is soft error (e.g. "
-             "retryable error), 2 is fatal error, and the defulat is "
+             "retryable error), 2 is fatal error, and the default is "
              "retryable error.");
 
 #endif  // GFLAGS
--- a/db_stress_tool/db_stress_gflags.cc
+++ b/db_stress_tool/db_stress_gflags.cc
@@ -828,4 +828,9 @@ DEFINE_int32(open_write_fault_one_in, 0,
              "On non-zero, enables fault injection on file write "
              "during DB reopen.");
 
+DEFINE_int32(injest_error_severity, 1,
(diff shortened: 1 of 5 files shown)


db_stress_tool/db_stress_gflags.cc
```diff
@@ -825,4 +825,9 @@ DEFINE_string(secondary_cache_uri, "",
               "Full URI for creating a customized secondary cache object");
 #endif  // ROCKSDB_LITE
 
+DEFINE_int32(injest_error_severity, 1,
+             "The severity of the injested IO Error. 1 is soft error (e.g. "
+             "retryable error), 2 is fatal error, and the defulat is "
```
anand1976, line 830: defulat -> default

db_stress_tool/no_batched_ops_stress.cc
```diff
@@ -552,8 +552,18 @@ class NonBatchedOpsStressTest : public StressTest {
...
     shared->Put(rand_column_family, rand_key, value_base, false /* pending */);
     if (!s.ok()) {
-      fprintf(stderr, "put or merge error: %s\n", s.ToString().c_str());
-      std::terminate();
+      if (FLAGS_injest_error_severity >= 2) {
+        if (!is_db_stopped_ && s.severity() >= Status::Severity::kFatalError) {
+          is_db_stopped_ = true;
+        } else if (!is_db_stopped_ ||
+                   s.severity() < Status::Severity::kFatalError) {
+          fprintf(stderr, "put or merge error: %s\n", s.ToString().c_str());
```
anand1976, line 560: If its a retryable error, we always expect the Put/Delete to succeed. Is that because the fault injection FS never injects errors on WAL writes?

zhichao-cao: > If its a retryable error, we always expect the Put/Delete to succeed. Is that because the fault injection FS never injects errors on WAL writes?

Yes, in the current fault injection, we set for SST, Manifest, and Current. https://github.com/facebook/rocksdb/pull/8479#discussion_r662412921

anand1976 (approved):

PR #8481, opened 2021-07-02
Make MergeOperator+CompactionFilter/Factory into Customizable Classes

mrambacher:
- Changed MergeOperator, CompactionFilter, and CompactionFilterFactory into Customizable classes.
 - Added Options/Configurable/Object Registration for TTL and Cassandra variants
 - Changed the StringAppend MergeOperators to accept a string delimiter rather than a simple char.  Made the delimiter into a configurable option
 - Added tests for new functionality


zhichao-cao:
Thanks for the changes. One question, for StringAppend, this PR makes it be able to accept a string delimiter. Is this change requested by a user or it is an issue reported at Github?

mrambacher:
> Thanks for the changes. One question, for StringAppend, this PR makes it be able to accept a string delimiter. Is this change requested by a user or it is an issue reported at Github?

The StringAppend with a String delimiter has been requested by another use and there is another PR against it (#8536).  Additionally, using a String (rather than a char) kept me from having to add a new OptionType to deal specifically for a char.  Win-win.

utilities/compaction_filters.cc
```diff
@@ -0,0 +1,59 @@
...
+  }
+  return status;
+}
+
+Status CompactionFilterFactory::CreateFromString(
+    const ConfigOptions& config_options, const std::string& value,
+    std::shared_ptr<CompactionFilterFactory>* result) {
+#ifndef ROCKSDB_LITE
+  std::call_once(load_once, [&]() {
+    RegisterBuiltinCompactionFilters(*(ObjectLibrary::Default().get()), "");
```
zhichao-cao, line 52: Why we call RegisterBuiltinCompactionFilters here? Is this function used for CompactionFileter?

utilities/merge_operators.cc
```diff
@@ -0,0 +1,117 @@
...
+#include "utilities/merge_operators.h"
+#include "utilities/merge_operators/bytesxor.h"
+#include "utilities/merge_operators/sortlist.h"
+#include "utilities/merge_operators/string_append/stringappend.h"
+#include "utilities/merge_operators/string_append/stringappend2.h"
+
+namespace ROCKSDB_NAMESPACE {
+std::shared_ptr<MergeOperator> MergeOperators::CreateFromStringId(
+    const std::string& id) {
+  if (id == "put" || id == "PutOperator") {
```
zhichao-cao, line 23: Is it possible to move those branches to CreateFromString. Then, we call CreateFromString inside MergeOperators::CreateFromStringId. It might be easier to deprecate CreateFromStringId in the future. What do you think?

utilities/merge_operators/string_append/stringappend.cc
```diff
@@ -6,18 +6,35 @@
...
+
 #include "rocksdb/merge_operator.h"
+#include "rocksdb/slice.h"
+#include "rocksdb/utilities/options_type.h"
 #include "utilities/merge_operators.h"
 
 namespace ROCKSDB_NAMESPACE {
-
+namespace {
+static std::unordered_map<std::string, OptionTypeInfo>
```
zhichao-cao, line 20: We should mention the change of stringappend merge operator in HISTORY.md. There are many users use this one.

utilities/ttl/db_ttl_impl.cc
```diff
@@ -13,9 +13,146 @@
 #include "rocksdb/iterator.h"
 #include "rocksdb/system_clock.h"
 #include "rocksdb/utilities/db_ttl.h"
+#include "rocksdb/utilities/object_registry.h"
+#include "rocksdb/utilities/options_type.h"
 #include "util/coding.h"
 
 namespace ROCKSDB_NAMESPACE {
+static std::unordered_map<std::string, OptionTypeInfo> ttl_mo_type_info = {
```
zhichao-cao, line 21: ttl_mo_type_info - >ttl_merge_op_type_info

utilities/ttl/db_ttl_impl.cc
```diff
@@ -13,9 +13,146 @@
...
+     OptionTypeInfo::AsCustomSharedPtr<MergeOperator>(
+         0, OptionVerificationType::kByName, OptionTypeFlags::kNone)}};
+
+TtlMergeOperator::TtlMergeOperator(
+    const std::shared_ptr<MergeOperator>& merge_op, SystemClock* clock)
+    : user_merge_op_(merge_op), clock_(clock) {
+  RegisterOptions("TtlMergeOptions", &user_merge_op_, &ttl_mo_type_info);
+}
+
+bool TtlMergeOperator::FullMergeV2(const MergeOperationInput& merge_in,
```
zhichao-cao, line 32: merge_in - >merge_op_input

utilities/ttl/db_ttl_impl.cc
```diff
@@ -34,6 +171,139 @@ void DBWithTTLImpl::SanitizeOptions(int32_t ttl, ColumnFamilyOptions* options,
...
+
+TtlCompactionFilter::TtlCompactionFilter(
+    int32_t ttl, SystemClock* clock, const CompactionFilter* _user_comp_filter,
+    std::unique_ptr<const CompactionFilter> _user_comp_filter_from_factory)
+    : LayeredCompactionFilterBase(_user_comp_filter,
+                                  std::move(_user_comp_filter_from_factory)),
+      ttl_(ttl),
+      clock_(clock) {
+  RegisterOptions("TTL", &ttl_, &ttl_type_info);
+  RegisterOptions("UserFilter", &user_comp_filter_, &user_cf_type_info);
```
zhichao-cao, line 196: Should we check if user_comp_filter_ is nullptr or not and then call RegisterOptions?

include/rocksdb/compaction_filter.h
```diff
@@ -59,6 +60,10 @@ class CompactionFilter {
   };
 
   virtual ~CompactionFilter() {}
+  static const char* Type() { return "CompactionFilter"; }
+  static Status CreateFromString(const ConfigOptions& config_options,
+                                 const std::string& name,
+                                 const CompactionFilter** result);
```
zhichao-cao, line 66: Can we use std::shared_ptr<CompactionFilter>* result here?

utilities/cassandra/cassandra_options.h
```diff
@@ -0,0 +1,31 @@
...
+//  (found in the LICENSE.Apache file in the root directory).
+
+#pragma once
+#include <cinttypes>
+
+#include "rocksdb/rocksdb_namespace.h"
+
+namespace ROCKSDB_NAMESPACE {
+class ObjectLibrary;
+namespace cassandra {
```
zhichao-cao, line 13: Add some comments to the CassandraOptions and its member variables.

options/options_test.cc
```diff
@@ -4114,6 +4102,79 @@ TEST_F(ConfigOptionsTest, EnvFromConfigOptions) {
...
+  delimiter = merge_op->GetOptions<std::string>("Delimiter");
+  ASSERT_NE(delimiter, nullptr);
+  ASSERT_EQ(*delimiter, "||");
+
+  ASSERT_OK(MergeOperator::CreateFromString(
+      config_options, "id=stringappendtest; delimiter=&&", &merge_op));
+  ASSERT_NE(merge_op, nullptr);
+  ASSERT_TRUE(merge_op->IsInstanceOf("stringappendtest"));
+  ASSERT_STREQ(merge_op->Name(), StringAppendTESTOperator::kClassName());
+  delimiter = merge_op->GetOptions<std::string>("Delimiter");
```
zhichao-cao, line 4174: Can you add the check for serialize the option string after create?

utilities/cassandra/cassandra_functional_test.cc
```diff
@@ -318,6 +319,85 @@ TEST_F(CassandraFunctionalTest, CompactionShouldRemoveTombstoneFromPut) {
   ASSERT_FALSE(std::get<0>(store.Get("k1")));
 }
 
+#ifndef ROCKSDB_LITE
+TEST_F(CassandraFunctionalTest, LoadMergeOperator) {
+  ConfigOptions config_options;
+  std::shared_ptr<MergeOperator> mo;
+  config_options.registry->AddLibrary("cassandra", RegisterCassandraObjects,
```
zhichao-cao, line 326: Add a check when config_options.registry->AddLibrary is not called and then call CreateFromString.

utilities/cassandra/merge_operator.cc
```diff
@@ -5,16 +5,35 @@
...
 #include "rocksdb/merge_operator.h"
-#include "utilities/merge_operators.h"
+#include "rocksdb/slice.h"
+#include "rocksdb/utilities/options_type.h"
 #include "utilities/cassandra/format.h"
+#include "utilities/merge_operators.h"
 
 namespace ROCKSDB_NAMESPACE {
 namespace cassandra {
+static std::unordered_map<std::string, OptionTypeInfo> merge_options_info = {
```
zhichao-cao, line 20: nit: merge_options_info-> merge_operator_options_info

utilities/merge_operators.cc
```diff
@@ -0,0 +1,117 @@
...
+  }
+  if (opt_map.empty()) {
+    merge_op = MergeOperators::CreateFromStringId(id);
+    if (merge_op) {
+      *result = merge_op;
+      return Status::OK();
+    }
+  }
+  if (value.empty()) {
+    // No Id and no options.  Clear the object
```
zhichao-cao, line 106: minor: double space before "Clear"

include/rocksdb/compaction_filter.h
mrambacher: A CompactionFilter (unlike a CompactionFilterFactory) is stored as a raw pointer in the Options infrastructure.  This signature matches that (singleton) usage pattern.

utilities/compaction_filters.cc
mrambacher: At the moment, there is only one builtin filter and it is a CompactionFilter.  If more builtins were added, they could be CompactionFilter or CompactionFilterFactory implementations.  This method was meant to be able to register either.

At one point I had written a "RemoveValueCompactionFilterFactory" but dropped it.

If you rather separate the two (CompactionFilter from CompactionFilterFactory), I can do that.

utilities/ttl/db_ttl_impl.cc
mrambacher: We always want to register the address of user_comp_filter_, even if it points to nullptr.  This allows us to later change what is pointed to.

include/rocksdb/customizable.h
```diff
@@ -87,7 +87,18 @@ class Customizable : public Configurable {
   // @param name The name of the instance to find.
   // Returns true if the class is an instance of the input name.
   virtual bool IsInstanceOf(const std::string& name) const {
-    return name == Name();
+    if (name.empty()) {
+      return false;
+    } else if (name == Name()) {
+      return true;
+    } else {
+      auto sname = ShortName();
```
zhichao-cao, line 95: Can we avoid using auto here, it is better to give clear type in the public portion.

include/rocksdb/customizable.h
```diff
@@ -179,6 +190,10 @@ class Customizable : public Configurable {
   virtual const Customizable* Inner() const { return nullptr; }
 
  protected:
+  // Some classes have both a class/long name (e.g. PutOperator) and a short
+  // name (e.g. put). Classes can override this method to return a
+  // short/alternate name.  These names are used for object
+  virtual const char* ShortName() const { return ""; }
```
zhichao-cao, line 196: This is newly added I think. Can you explain why we need another name? What is this used for? Can be confusing for users when they use Name or ShortName

utilities/compaction_filters.cc
zhichao-cao: Maybe better to separate CompactionFilter from CompactionFilterFactory. Easy to understand.

utilities/ttl/db_ttl_impl.cc
zhichao-cao: We should mention it in the comment, otherwise, it is a little misleading.

include/rocksdb/customizable.h
mrambacher: The ShortName provides a means to define an alternative name for a class.  For example, the PutOperator is known as a "PutOperator" but also via "put".  This method allows a class to define both the class name ("PutOperator") and a short/alternate/nick name ("put").

By making it part of the base class, I do not have to repeat the same code over and over again for IsInstanceOf, and can define constants associated with the class for comparison/construction.

If you have an alternate method name you prefer (NickName? AlternativeName, ByName?), I can rename it.

zhichao-cao: It is more like a historical naming. Maybe HistoricalName would be better? But for other classes, actually, we only have the kClassName. So, if a class does not have a name before, when we migrate it to Customizable, it should only have kClassName, am I right?

mrambacher: I am not adding these names to classes that did not have them already, so in that sense they are historical.  All classes should have  Name/kClassName.  Only a select few have this alternative name.

I prefer "Byname" or "NickName" to "HistoricalName", as Historical makes it sound like we will never write another class again with these alternative names.  But if you prefer Historical, I will go with that.

zhichao-cao: So, maybe NickName. At lease, user know that they should use Name if they can.

zhichao-cao (approved):
LGTM, thanks for addressing all the comments!

Issue #8453, opened 2021-06-24
Get with kBlockCacheTier return Status::OK() when block cache is disabled


helius-isaac:
When opening a database with `no_block_cache = true`, trying to read with `read_tier = kBlockCacheTier` should return `Status::Incomplete()`. However, the return value is `Status::OK()` and the value is empty.

### Expected behavior

The `Get()` call should return `Status::Incomplete()`.

### Actual behavior

The call returns with `Status::OK()` and an empty value.

### Steps to reproduce the behavior

```c++
#include <cassert>
#include <string>
#include <rocksdb/db.h>
#include <rocksdb/table.h>

using namespace rocksdb;

int main(int argc, char const *argv[])
{
	BlockBasedTableOptions table_options;
	table_options.no_block_cache = true;
	Options options;
	options.create_if_missing = true;
	options.table_factory.reset(NewBlockBasedTableFactory(table_options));

	DB* db = nullptr;
	Status s = DB::Open(options, "./cache_tier_db", &db);
	assert(s.ok());

	const std::string key = "key";
	const std::string value = "value";

	assert(db->Put(WriteOptions(), key, value).ok());

	assert(db->Flush(FlushOptions()).ok());

	{
		PinnableSlice result;
		assert(db->Get(ReadOptions(), db->DefaultColumnFamily(), key, &result).ok());
		assert(result == value);
	}

	// Disallow I/O
	ReadOptions read_options;
	read_options.read_tier = kBlockCacheTier;

	PinnableSlice result;
	s = db->Get(read_options, db->DefaultColumnFamily(), key, &result);
	assert(s.IsIncomplete());
	assert(result.empty());

	return 0;
}
```

hongrubb:
I will work on this and try to fix it.

helius-isaac:
@ajkr I think you meant to assign @hongrubb, considering the comment above.

hongrubb:
I think I have found the problem. I will submit a PR soon.

hongrubb:
@ajkr Hi, I create a pr about this issue. Please take a look. #8485

PR #8485, opened 2021-07-06
Fix Get() return status when block cache is disabled

hongrubb:
This PR is for #8453 

We need to update `s = biter.status();`  when `biter.status().IsIncomplete()` is true. By doing this, can fix the problem in issue.
Besides, we still need to update `db_statistics`  in `get_context.ReportCounters()` before return back.


table/block_based/block_based_table_reader.cc
```diff
@@ -2341,6 +2341,7 @@ Status BlockBasedTable::Get(const ReadOptions& read_options, const Slice& key,
         // Update Saver.state to Found because we are only looking for
         // whether we can guarantee the key is not there when "no_io" is set
         get_context->MarkKeyMayExist();
+        s = biter.status();
```
helius-isaac, line 2344: Wouldn't it be better to avoid the duplication with line 2346 below by moving this entire `if` block into the `if` block there?

```suggestion
      if (!biter.status().ok()) {
        if (no_io && biter.status().IsIncomplete()) {
          // couldn't get block from block_cache
          // Update Saver.state to Found because we are only looking for
          // whether we can guarantee the key is not there when "no_io" is set
          get_context->MarkKeyMayExist();
        }
        s = biter.status();
        break;
      }
```

table/block_based/block_based_table_reader.cc
```diff
@@ -2341,6 +2341,7 @@ Status BlockBasedTable::Get(const ReadOptions& read_options, const Slice& key,
...
         break;
       }
       if (!biter.status().ok()) {
```
helius-isaac, line 2347: Wouldn't it be better to avoid the duplication with line 2347 below and instead move this `if` into the `if` block there?

```suggestion
      if (!biter.status().ok()) {
        if (no_io && biter.status().IsIncomplete()) {
          // couldn't get block from block_cache
          // Update Saver.state to Found because we are only looking for
          // whether we can guarantee the key is not there when "no_io" is set
          get_context->MarkKeyMayExist();
        }
        s = biter.status();
        break;
      }
```

ajkr (approved):
Thanks for finding/fixing the problem! This feels like something that's reasonably straightforward to cover in a test case, if you have time.

ajkr (commented):
Also do you mind updating HISTORY.md with a "Bug Fixes" item for preventing empty value returned for non-existent key, and a separate item for fixing dropped stat block cache counter updates for failed queries?

db/version_set.cc
```diff
@@ -1938,16 +1938,16 @@ void Version::Get(const ReadOptions& read_options, const LookupKey& k,
       PERF_COUNTER_BY_LEVEL_ADD(get_from_table_nanos, timer.ElapsedNanos(),
                                 fp.GetHitFileLevel());
     }
-    if (!status->ok()) {
```
ajkr, line 1941: Actually is it better to call `ReportCounters()` in this old `!status->ok()` conditional body? Seems like they are supposed to be updated before return as long as `db_statistics_ != nullptr`.

hongrubb:
> Thanks for finding/fixing the problem! This feels like something that's reasonably straightforward to cover in a test case, if you have time.

Sorry, I don't have time to add test cases right now. But I'd like to contribute test cases in later weeks. Is it ok to submit another PR for test cases?

db/version_set.cc
hongrubb: I don't know if I understand you correctly.
You mean the code should look like this?
```
    if (!status->ok()) {
      if (db_statistics_ != nullptr) {
        get_context.ReportCounters();
      }
      return;
    }
```
If that, I will change the code soon.

hongrubb:
> Also do you mind updating HISTORY.md with a "Bug Fixes" item for preventing empty value returned for non-existent key, and a separate item for fixing dropped stat block cache counter updates for failed queries?

I'd like to do that, I will update it soon.

ajkr:
> > Thanks for finding/fixing the problem! This feels like something that's reasonably straightforward to cover in a test case, if you have time.
> 
> Sorry, I don't have time to add test cases right now. But I'd like to contribute test cases in later weeks. Is it ok to submit another PR for test cases?

Sure, sounds good.

db/version_set.cc
ajkr: Right.

ajkr (approved):
Looks great, thank you very much!

ajkr:
Do you mind checking this test?

```
stdout: Note: Google Test filter = DBBlockCacheTest.TestWithCompressedBlockCache
[==========] Running 1 test from 1 test suite.
[----------] 1 test from DBBlockCacheTest
[ RUN      ] DBBlockCacheTest.TestWithCompressedBlockCache
internal_repo_rocksdb/repo/db/db_block_cache_test.cc:99: Failure
  miss_count_ + expected_misses
    Which is: 9
  new_miss_count
    Which is: 10
internal_repo_rocksdb/repo/db/db_block_cache_test.cc:99: Failure
  miss_count_ + expected_misses
    Which is: 10
  new_miss_count
```

This is from our internal pre-land test run; I don't see the error in any test run here somehow...

hongrubb:
> Do you mind checking this test?
> 
> ```
> stdout: Note: Google Test filter = DBBlockCacheTest.TestWithCompressedBlockCache
> [==========] Running 1 test from 1 test suite.
> [----------] Global test environment set-up.
> [----------] 1 test from DBBlockCacheTest
> [ RUN      ] DBBlockCacheTest.TestWithCompressedBlockCache
> internal_repo_rocksdb/repo/db/db_block_cache_test.cc:99: Failure
> Expected equality of these values:
>   miss_count_ + expected_misses
>     Which is: 9
>   new_miss_count
>     Which is: 10
> internal_repo_rocksdb/repo/db/db_block_cache_test.cc:99: Failure
> Expected equality of these values:
>   miss_count_ + expected_misses
>     Which is: 10
>   new_miss_count
> ```
> 
> This is from our internal pre-land test run; I don't see the error in any test run here somehow...

Sure.

hongrubb:
@ajkr Hi, I think I have found the problem. It's very interesting!

I found the failed test was added by #8242  25days ago.
The test failed at CheckCacheCounters(), I think we should update the counter even though the Get() is failed.
```
  ASSERT_EQ("Result incomplete: Insert failed due to LRU cache being full.",
            Get(ToString(kNumBlocks - 1)));
  // Failure won't record the miss counter.
  CheckCacheCounters(options, 0, 0, 0, 1);
```

@akankshamahajan15  thanks for the test and detailed comment in code, it helped me a lot. And please take a look at this problem, maybe we should change this test.

akankshamahajan15:
> The test failed at CheckCacheCounters(), but I think we should update the counter even though the Get() is failed.
> 
> ```
>   ASSERT_EQ("Result incomplete: Insert failed due to LRU cache being full.",
>             Get(ToString(kNumBlocks - 1)));
>   // Failure won't record the miss counter.
>   CheckCacheCounters(options, 0, 0, 0, 1);
> ```
> 
> @akankshamahajan15 thanks for the test and detailed comment in code, it helped me a lot. And please take a look at this problem, maybe we should change this test.

Since you fixed the bug of adding statistics when query is failed", this test should be updated to take that into account. Please feel free to update the test. Thanks for fixing the bug.

PR #8490, opened 2021-07-06
Bypass buffer in TestFSWritableFile if direct IO is enabled

anand1976:
```TestFSWritableFile``` buffers data in ```Append``` in order to simulate unsynced data loss on crash. This is only required for buffered IO and should be disabled for direct IO. Otherwise, it causes crash tests to assert on the buffer address alignment - ```db_stress: env/io_posix.cc:1194: virtual rocksdb::IOStatus rocksdb::PosixWritableFile::Append(const rocksdb::Slice&, const rocksdb::IOOptions&, rocksdb::IODebugContext*): Assertion `IsSectorAligned(data.data(), GetRequiredBufferAlignment())' failed.```.

### DIFF
--- a/utilities/fault_injection_fs.cc
+++ b/utilities/fault_injection_fs.cc
@@ -124,15 +124,19 @@ TestFSWritableFile::~TestFSWritableFile() {
   }
 }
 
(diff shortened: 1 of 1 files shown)


zhichao-cao (approved):
LGTM, thanks for fixing it

PR #8493, opened 2021-07-07
Add micro-benchmark support

jay-zhuang:
Summary: Add google benchmark for microbench.
Add ribbon_bench for benchmark ribbon filter vs. other filters.

Test Plan: added test to CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,6 +91,13 @@ commands:
           command: |
             sudo apt-get update -y && sudo apt-get install -y libgflags-dev
 
+  install-benchmark:
+    steps:
+      - run: # currently doesn't support ubuntu-1604 which doesn't have libbenchmark package, user can still install by building it youself
+          name: Install benchmark
+          command: |
+            sudo apt-get update -y && sudo apt-get install -y libbenchmark-dev
+
   upgrade-cmake:
     steps:
       - run:
@@ -317,7 +324,8 @@ jobs:
     steps:
       - checkout # check out the code in the project directory
       - install-gflags
-      - run: (mkdir build && cd build && cmake -DWITH_GFLAGS=1 .. && make V=1 -j20 && ctest -j20) | .circleci/cat_ignore_eagain
+      - install-benchmark
+      - run: (mkdir build && cd build && cmake -DWITH_GFLAGS=1 -DWITH_BENCHMARK=1 .. && make V=1 -j20 && ctest -j20 && make microbench) | .circleci/cat_ignore_eagain
       - post-steps
 
   build-linux-unity:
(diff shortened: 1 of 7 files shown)


microbench/ribbon_bench.cc
```diff
@@ -0,0 +1,154 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+
+#include <benchmark/benchmark.h>
+
```
pdillinger, line 7: This seems like a reasonable starting point, but filter_bench is more representative of how filters are used in RocksDB, e.g. by queries hitting different filters occupying a large memory space. We should express here that the goal is to eventually get those kinds of results into the reusable benchmarking framework. One could get the impression that this file is exclusively focused on "more micro" benchmarking of filter behavior than filter_bench does, but I do not believe that is the goal. Others could assume this file is "authoritative" and no other benchmarking for filters exists.

In broader terms, I fear that even using the term "microbenchmark" might distract from the goal (as I see it) of taking meaningful measurements of subsystem performance to catch regressions (and track improvements) by focusing too much on individual components (e.g. querying a filter) rather than treating them as subsystems (e.g. querying one of many filters).

pdillinger (approved):
Seems fine, with some questions / critiques

microbench/ribbon_bench.cc
jay-zhuang: I agree both.
1. this file is just the first benchmark that more for demonstrating the integration. Compare to `filter_bench`, it's missing lots tests, but I guess if we were having this microbenchmark framework before, we might just use it instead of creating stand-alone `filter_bench`. I'll add comment to refer to `filter_bench` for better filter benchmark.
2. For `microbenchmark` focusing too much on small component, yes, I think it should not replace db_bench, it just gives extra, targeted information. it's also up to the test designer, so when writing future benchmark tests, we should have a more meaningful benchmark and metrics. Some metrics might be useful during development (like this one) but maybe not useful (or likely change) after feature is done, which we could exclude in the run or just ignore the result.

.circleci/config.yml
```diff
@@ -370,6 +378,16 @@ jobs:
       - run: CC=gcc-10 CXX=g++-10 V=1 SKIP_LINK=1 ROCKSDB_CXX_STANDARD=c++20 make -j16 all | .circleci/cat_ignore_eagain # Linking broken because libgflags compiled with newer ABI
       - post-steps
 
+  build-linux-microbench:
```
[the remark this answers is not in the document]
jay-zhuang: Yes, just to make sure the tests pass.

mrambacher:
This change will now generate build errors/warnings on any platform/host that does not have the benchmark installed:

make dbg
$DEBUG_LEVEL is 2
Makefile:176: Warning: Compiling in debug mode. Don't use the resulting binary in production
microbench/ribbon_bench.cc:8:10: fatal error: 'benchmark/benchmark.h' file not
      found
#include <benchmark/benchmark.h>
         ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.

I believe this is part of the dependency generation step.

PR #7924, opened 2021-02-02
Wal recovery failure with encryption due to zero bytes WAL size.

deepakojha123:
Fixes issue #7410

PR #8499, opened 2021-07-07
make encrypted_env open empty files

jsteemann:
When the encrypted_env opens a file (e.g. a log file), it will try
to read the initial few bytes from the file to build up decryption
metadata.
There is a check in place that prevented opening files that were
shorter than the required metadata prefix length.

However, RocksDB can create empty log files. This can happen if
there is a crash/power loss directly after log file creation.
In this case there is only an empty log file, which then cannot be
read by the encrypted_env. The env will report a Corruption error,
and things are stuck.
The fix here is to make the encrypted_env gracefully ignore all
completely empty files, so that it will return Status::Ok() for
then.

I couldn't find any particular unit tests for encrypted_env in
RocksDB, but the fix has been tried and is covered by tests in
the ArangoDB repository.

### DIFF
--- a/env/env_encryption.cc
+++ b/env/env_encryption.cc
@@ -1215,19 +1215,24 @@ Status CTREncryptionProvider::CreateCipherStream(
   Slice iv;
   decodeCTRParameters(prefix.data(), blockSize, initialCounter, iv);
 
-  // If the prefix is smaller than twice the block size, we would below read a
-  // very large chunk of the file (and very likely read over the bounds)
-  assert(prefix.size() >= 2 * blockSize);
-  if (prefix.size() < 2 * blockSize) {
-    return Status::Corruption("Unable to read from file " + fname +
-                              ": read attempt would read beyond file bounds");
-  }
-
   // Decrypt the encrypted part of the prefix, starting from block 2 (block 0, 1
   // with initial counter & IV are unencrypted)
   CTRCipherStream cipherStream(cipher_, iv.data(), initialCounter);
   Status status;
-  {
+ 
+  // A completely empty prefix is tolerated here, which allows to open
+  // encrypted log files that were only created but not filled with any
+  // data. This can happen if there is a crash/power loss directly after
+  // log file creation.
+  if (!prefix.empty()) {
+    // If the prefix is smaller than twice the block size, we would below read a
+    // very large chunk of the file (and very likely read over the bounds)
+    assert(prefix.size() >= 2 * blockSize);
+    if (prefix.size() < 2 * blockSize) {
+      return Status::Corruption("Unable to read from file " + fname +
+                                ": read attempt would read beyond file bounds");
+    }
+
     PERF_TIMER_GUARD(decrypt_data_nanos);
     status = cipherStream.Decrypt(0, (char*)prefix.data() + (2 * blockSize),
                                   prefix.size() - (2 * blockSize));


env/env_encryption.cc
```diff
@@ -1215,19 +1215,24 @@ Status CTREncryptionProvider::CreateCipherStream(
   Slice iv;
   decodeCTRParameters(prefix.data(), blockSize, initialCounter, iv);
```
mrambacher, line 1216: Won't this crash/error if the prefix is empty?

mrambacher (commented):
I do not have time (on vacation) to do a more thorough review.  But it seems like this should be at a higher layer to make sure that the prefix is read at the proper length.  

If the read prefix is empty, perhaps the code should simply return the underlying reader rather than the wrapped one?

A test with different empty file types (sequential, random) might help.

env/env_encryption.cc
jsteemann: @mrambacher : I guess you are right, this doesn't look safe to me too.
I think right now it won't crash because for `prefix` there will always be a buffer allocated which is 16 bytes long. So the `prefix` pointer currently seems to point into a 16 bytes buffer, which isn't filled if the input file is empty.
That's potentially also the reason why it worked just fine when I tried it.

jsteemann: @mrambacher 

> If the read prefix is empty, perhaps the code should simply return the underlying reader rather than the wrapped one?

The `EncryptedFileSystemImpl`'s methods, e.g. `CreateSequentialCipherStream()` know only about the `provider_`, which is a generic `shared_ptr<EncryptionProvider>`. 
Right now they will call `CreateCipherStream()`, which is a public method on the `EncryptionProvider` interface. The concrete `CTREncryptionProvider` will then internally call its `CreateCipherStreamFromPrefix()` method, which will finally create a `BlockAccessCipherStream` of specialized type `CTRCipherStream`.

If I now wanted to add some special empty file handling in `CreateSequentialCipherStream()`, all I can do is call into the general `EncryptionProvider` interface, which only provides `CreateCipherStream()`. The methods I would acutally need are not on the generic `EncryptionProvider` interface.
So I think it would still be simplest to have the `CTREncryptionProvider` handle empty prefixes itself. This can be made safe by also bypassing the call to `decodeCTRParameters()` in case the prefix is fully empty.

So I have adjusted the PR accordingly, as I don't know how to return some `CTREncryptionProvider` internals when inside the `EncryptedFileSystemImpl`.
The adjusted version still fixes the problem of not reporting corruption for 0 byte files, as I have just verified when calling it from our application.

I also added a unit for reading an empty file with the default env and the encrypted env.
The unit test will fail for the encrypted env without the changes in this PR, and will pass with them.

ajkr:
Have we considered making `EncryptedFileSystemImpl::NewWritableFile()` do the dance of creating a tempfile (".xxxxxxx.log.tmp"), populating the prefix, fsyncing the contents, then renaming it to the final filename? I believe the caller then fsyncs the directory to complete the sequence. Then there is no special case for unprefixed encrypted WAL. It also prevents the case where a whole WAL looks like garbage (causing kTolerateCorruptTailRecords to fail recovery) because the prefix wasn't synced.

jsteemann:
@ajkr: I haven't considered that yet, but I guess it makes a lot of sense too.
However, I also need to address the problem of _existing_ empty WAL files, so I would like to have a solution for any installation that confronts `env_encryption` with an empty WAL file.
I am happy to try your suggestion on top of this PR if you agree it's a good way forward.

ajkr:
> @ajkr: I haven't considered that yet, but I guess it makes a lot of sense too.
> However, I also need to address the problem of _existing_ empty WAL files, so I would like to have a solution for any installation that confronts `env_encryption` with an empty WAL file.
> I am happy to try your suggestion on top of this PR if you agree it's a good way forward.

Does your version of RocksDB include https://github.com/facebook/rocksdb/pull/7924? It feels like that is doing what @mrambacher suggested for NewSequentialFile(), which I'd think is what recovery uses.

jsteemann:
@ajkr: ah, thanks for pointing me at that PR.
Our version of RocksDB is based on 6.8 and doesn't include that commit yet. Unfortunately it's not trivial for us to upgrade due to other backwards-incompatible changes.
I believe the fix you mentioned will work, so I am happy to fix the problem only in our bundled version of RocksDB, and abandon my PR for good.

matthewvon:
The code goes to a lot of trouble to track the prefix_length_ in each of the file types.  How about biting the bullet now and tracking the provider shared_ptr instead?  Then have a default provider that is for not encrypted files?  Covers the original problem nicely and establishes a beginning more than one provider simultaneously (including more than one key simultaneously with same provider).

PR #8503, opened 2021-07-07
Improve support for valgrind error on reachable

pdillinger:
Summary: MyRocks apparently uses valgrind to check for unreachable
unfreed data, which is stricter than our valgrind checks. Internal ref:
D29257815

This patch adds valgrind support to STATIC_AVOID_DESTRUCTION so that it's
not reported with those stricter checks.

Test Plan: make valgrind_test
Also, with modified VALGRIND_OPTS (see Makefile), more kinds of
failures seen before than after this commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -488,6 +488,7 @@ VALGRIND_ERROR = 2
 VALGRIND_VER := $(join $(VALGRIND_VER),valgrind)
 
 VALGRIND_OPTS = --error-exitcode=$(VALGRIND_ERROR) --leak-check=full
+# Not yet supported: --show-leak-kinds=definite,possible,reachable --errors-for-leak-kinds=definite,possible,reachable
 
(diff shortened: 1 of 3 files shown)


ajkr (approved):
LGTM.

> Also uses
STATIC_AVOID_DESTRUCTION for default PosixEnv to keep it from being
reported.

I didn't quite understand. Why would it be reported before that change since the destructor should still have run back when it was a regular `static` variable?

pdillinger:
> I didn't quite understand. Why would it be reported before that change since the destructor should still have run back when it was a regular `static` variable?

Good question. I don't exactly remember what I saw, and I'm too lazy right now to reproduce, so I'll just weaken the summary and ship it. ;) (We agree that avoiding the destructor on a static is generally safer.)

pdillinger:
Correction: ~PosixEnv must be called. (TASN flagged)

PR #8504, opened 2021-07-07
Fix incorrect Status::NoSpace() status check

zhichao-cao:
If we want to check whether a Status s is NoSpace() or not, we should check the subcode instread of using s==Status::NoSpace(). Fix some of the incorrect check in the ErrorHandler.

test plan: make check

### DIFF
--- a/db/error_handler.cc
+++ b/db/error_handler.cc
@@ -327,7 +327,8 @@ const Status& ErrorHandler::SetBGError(const Status& bg_err,
   }
 
   // Allow some error specific overrides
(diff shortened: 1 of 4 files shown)


zhichao-cao:
Looks like it causes the testing issue in db_sst_test-DBSSTTest.DBWithMaxSpaceAllowedRandomized, need to be investigated

anand1976 (approved):
LGTM

anand1976:
Nit: Update HISTORY.md

PR #8505, opened 2021-07-08
Make mempurge a background process (equivalent to in-memory compaction).

bjlemaire:
In #8454, I introduced a new process baptized `MemPurge` (memtable garbage collection). This new PR is built upon this past mempurge prototype.
In this PR, I made the `mempurge` process a background task, which provides superior performance since the mempurge process does not cling on the db_mutex anymore, and addresses severe restrictions from the past iteration (including a scenario where the past mempurge was failling, when a memtable was mempurged but was still referred to by an iterator/snapshot/...).
Now the mempurge process ressembles an in-memory compaction process: the stack of immutable memtables is filtered out, and the useful payload is used to populate an output memtable. If the output memtable is filled at more than 60% capacity (arbitrary heuristic) the mempurge process is aborted and a regular flush process takes place, else the output memtable is kept in the immutable memtable stack. Note that adding this output memtable to the `imm()` memtable stack does not trigger another flush process, so that the flush thread can go to sleep at the end of a successful mempurge.
MemPurge is activated by making the `experimental_allow_mempurge` flag `true`. When activated, the `MemPurge` process will always happen when the flush reason is `kWriteBufferFull`.
The 3 unit tests confirm that this process supports `Put`, `Get`, `Delete`, `DeleteRange` operators and is compatible with `Iterators` and `CompactionFilters`.


db/column_family.cc
```diff
@@ -443,26 +443,17 @@ bool SuperVersion::Unref() {
...
   MemTable* m = mem->Unref();
   if (m != nullptr) {
-    // Typically, if the m memtable was not made
-    // immutable, and therefore was not added to the
-    // imm list, it does not contribute to the imm
-    // memory footprint (and actually is not part of
-    // the 'imm' MemtableList at all).
-    // At the moment, noImmMemoryContribution is only
-    // used by the experimental 'MemPurge' prototype.
-    if (!noImmMemoryContribution) {
```
bjlemaire, line 461: Remove changes made during #8454 when the mempurge process was happening in the middle of the SwitchMemtable process.

db/column_family.cc
```diff
@@ -1272,7 +1263,7 @@ void ColumnFamilyData::InstallSuperVersion(
 
 void ColumnFamilyData::InstallSuperVersion(
     SuperVersionContext* sv_context, InstrumentedMutex* db_mutex,
-    const MutableCFOptions& mutable_cf_options, bool noImmMemoryContribution) {
```
bjlemaire, line 1275: Remove changes made during #8454 when the mempurge process was happening in the middle of the SwitchMemtable process.

db/column_family.cc
```diff
@@ -1302,7 +1293,7 @@ void ColumnFamilyData::InstallSuperVersion(
           new_superversion->write_stall_condition, GetName(), ioptions());
     }
     if (old_superversion->Unref()) {
-      old_superversion->Cleanup(noImmMemoryContribution);
```
bjlemaire, line 1305: Remove changes made during #8454 when the mempurge process was happening in the middle of the SwitchMemtable process.

db/column_family.h
```diff
@@ -222,10 +222,7 @@ struct SuperVersion {
   // Cleanup unrefs mem, imm and current. Also, it stores all memtables
   // that needs to be deleted in to_delete vector. Unrefing those
   // objects needs to be done in the mutex
-  // The 'noImmMemoryContribution' is set to true if the memtable being
-  // dereferenced in this SuperVersion was not added to the Immutable
-  // memtable list.
-  void Cleanup(bool noImmMemoryContribution = false);
```
bjlemaire, line 228: Remove changes made during #8454 when the mempurge process was happening in the middle of the SwitchMemtable process.

db/column_family.h
```diff
@@ -457,8 +454,7 @@ class ColumnFamilyData {
   // IMPORTANT: Only call this from DBImpl::InstallSuperVersion()
   void InstallSuperVersion(SuperVersionContext* sv_context,
                            InstrumentedMutex* db_mutex,
-                           const MutableCFOptions& mutable_cf_options,
-                           bool noImmMemoryContribution = false);
```
bjlemaire, line 461: Remove changes made during #8454 when the mempurge process was happening in the middle of the SwitchMemtable process.

db/db_flush_test.cc
```diff
@@ -697,74 +697,133 @@ TEST_F(DBFlushTest, MemPurgeBasic) {
...
+  std::string RNDKEY1, RNDKEY2, RNDKEY3;
   const std::string NOT_FOUND = "NOT_FOUND";
 
-  // Check simple operations (put-delete).
-  ASSERT_OK(Put(KEY1, VALUE1));
-  ASSERT_OK(Put(KEY2, VALUE2));
-  ASSERT_OK(Delete(KEY1));
-  ASSERT_OK(Put(KEY2, VALUE1));
-  ASSERT_OK(Put(KEY1, VALUE2));
-  ASSERT_OK(Flush());
```
bjlemaire, line 722: Now a manual flush like this one doesn't trigger a MemPurge anymore.

db/db_flush_test.cc
```diff
@@ -801,9 +861,9 @@ TEST_F(DBFlushTest, MemPurgeDeleteAndDeleteRange) {
   const std::string NOT_FOUND = "NOT_FOUND";
 
   Random rnd(117);
-  const size_t NUM_REPEAT = 200;
-  const size_t RAND_VALUES_LENGTH = 512;
-  bool atLeastOneFlush = false;
+  const size_t NUM_REPEAT = 1000;
+  const size_t RAND_VALUES_LENGTH = 20480;
```
bjlemaire, line 865: Another option to lower these values (and speed up the test) would be to change the memtable size (right now: uses the default 64MB size)

db/db_impl/db_impl.cc
```diff
@@ -548,12 +548,39 @@ Status DBImpl::CloseHelper() {
   flush_scheduler_.Clear();
   trim_history_scheduler_.Clear();
 
+  // For now, simply trigger a manual flush at close time
+  // on all the column families.
+  if (immutable_db_options_.experimental_allow_mempurge) {
```
bjlemaire, line 553: At the moment, automagically flush all column families.  In the future I can do a more fine-grained flushing by first checking if there is a need for flushing (but need to implement something else than imm()->IsFlushPending() because the output memtables added to imm() dont trigger flushes).

db/db_impl/db_impl_write.cc
```diff
@@ -2114,48 +1936,11 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
...
-      new_mem = cfd->ConstructNewMemtable(mutable_cf_options, seq);
-      assert(new_mem != nullptr);
-      context->superversion_context.NewSuperVersion();
-      cfd->imm()->Add(cfd->mem(), &context->memtables_to_free_);
-    }
-  } else {
-    // Else make the memtable immutable and proceed as usual.
-    cfd->imm()->Add(cfd->mem(), &context->memtables_to_free_);
-  }
+  cfd->imm()->Add(cfd->mem(), &context->memtables_to_free_);
```
bjlemaire, line 1939: Remove changes made during #8454 when the mempurge process was happening in the middle of the SwitchMemtable process.

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+    if (s.ok() && (new_first_seqno != kMaxSequenceNumber)) {
+      // Rectify the first sequence number, which (unlike the earliest seq
+      // number) needs to be present in the new memtable.
+      new_mem->SetFirstSequenceNumber(new_first_seqno);
+
+      // The new_mem is added to the list of immutable memtables
+      // only if it filled at less than 10% capacity (arbitrary heuristic).
+      if (new_mem->ApproximateMemoryUsage() <
+          static_cast<size_t>(
+              ceil(0.1 * mutable_cf_options_.write_buffer_size))) {
```
bjlemaire, line 535: Arbitrary metric: add to imm() only if memtable at less then 10% capacity. Reason for this: minimize Get/Read overheads that come from storing an extra Imm memtable, while giving us a change to perform in memory compaction.

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+  db_mutex_->Lock();
+
+  // If mempurge successful, don't write input tables to level0,
+  // but write any full output table to level0.
+  if (s.ok()) {
+    TEST_SYNC_POINT("DBImpl::FlushJob:MemPurgeSuccessful");
+    for (MemTable* m : mems_) {
+      if (m != nullptr) {
+        TEST_SYNC_POINT("DBImpl::FlushJob:MemPurgeSuccessful");
+        m->SetMempurged(true);
```
bjlemaire, line 562: Because these memtables have already been filtered out, avoid doing the work twice.

db/flush_job.cc
```diff
@@ -335,20 +612,23 @@ Status FlushJob::WriteLevel0Table() {
...
-      ROCKS_LOG_INFO(
-          db_options_.info_log,
-          "[%s] [JOB %d] Flushing memtable with next log file: %" PRIu64 "\n",
-          cfd_->GetName().c_str(), job_context_->job_id, m->GetNextLogNumber());
-      memtables.push_back(m->NewIterator(ro, &arena));
-      auto* range_del_iter =
-          m->NewRangeTombstoneIterator(ro, kMaxSequenceNumber);
-      if (range_del_iter != nullptr) {
-        range_del_iters.emplace_back(range_del_iter);
+      if (!(m->GetMempurged())) {
```
bjlemaire, line 615: See https://github.com/facebook/rocksdb/pull/8505/files#r667019976.

db/flush_job.cc
```diff
@@ -360,9 +640,12 @@ Status FlushJob::WriteLevel0Table() {
...
                          << GetFlushReasonString(cfd_->GetFlushReason());
 
-    {
+    // In certain situations, a non-empty mems_ can still lead
+    // to memtables.empty(). For example, if mems_ is made of
+    // memtables that have been mempurged.
+    if (!memtables.empty()) {
       ScopedArenaIterator iter(
-          NewMergingIterator(&cfd_->internal_comparator(), &memtables[0],
+          NewMergingIterator(&cfd_->internal_comparator(), memtables.data(),
```
bjlemaire, line 648: Use of vector.data() is preferable (allowed in C++11), because &memtables[0] leads to an undefined behavior if memtables is empty (I also added an if-statment, pure paranoia).

db/memtable_list.cc
```diff
@@ -527,7 +528,8 @@ void MemTableList::Add(MemTable* m, autovector<MemTable*>* to_delete) {
   current_->Add(m, to_delete);
   m->MarkImmutable();
   num_flush_not_started_++;
-  if (num_flush_not_started_ == 1) {
+
+  if (num_flush_not_started_ > 0 && trigger_flush) {
```
bjlemaire, line 532: This is one thing that could potentially lead to issues, maybe? Before, they stored "flush_needed" as soon as num_flush_not_started reached exactly 1. Why didnt they use num_flush_not_started>0, which should have been strictly equivalent? Either way, here we still need to increment num_flush_not_started but we dont want to trigger flush, otherwise the flush will keep spinning.

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+      }
+
+      // If new_mem is full, add it to purged_mems, and allocate new
+      // memtable.
+      if (new_mem->ShouldScheduleFlush()) {
+        // Rectify the first sequence number, which (unlike the earliest seq
+        // number) needs to be present in the new memtable.
+        new_mem->SetFirstSequenceNumber(new_first_seqno);
+        purged_mems.push_back(new_mem);
+        new_mem =
```
bjlemaire, line 464: This strategy is also probably something we want to discuss: should we create new memtables, or simply abort if we end up in this situation?
Basically: pros of creating new mems: we dont waste the time (possibly?) spent in mempurge.
Cons: spikes in memory usage by memtables.

db/flush_job.cc
pdillinger: Btw, could avoid floating point with

    (mutable_cf_options_.write_buffer_size + 9) / 10

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+        break;
+      }
+
+      // If new_mem is full, add it to purged_mems, and allocate new
+      // memtable.
+      if (new_mem->ShouldScheduleFlush()) {
+        // Rectify the first sequence number, which (unlike the earliest seq
+        // number) needs to be present in the new memtable.
+        new_mem->SetFirstSequenceNumber(new_first_seqno);
+        purged_mems.push_back(new_mem);
```
pdillinger, line 463: I thought purged_mems was for obsolete memtables. Don't we need to keep this one?

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+          nullptr,   // get_post_process_info(m) must be nullptr
+                     // when concurrent_memtable_writes is switched off.
+          nullptr);  // hint, only used when concurrent_memtable_writes_
+                     // is switched on.
+      if (!s.ok()) {
+        break;
+      }
+
+      // If new_mem is full, add it to purged_mems, and allocate new
+      // memtable.
```
pdillinger, line 458: The other option here is to abort mempurge. I suspect that's going to be better in practice to avoid continuing to run up our memory usage simply to finish our (apparently under-productive) purge right before flush. This could be a TODO to revisit, but if this code is not covered by a test, that is suspicious.

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+  db_mutex_->AssertHeld();
+  db_mutex_->Unlock();
+  assert(!mems_.empty());
+
+  // Store the full output memtables in
+  // autovector "purged_mems".
+  // autovector<MemTable*> purged_mems = {};
+  purged_mems = {};
+
+  MemTable* new_mem = nullptr;
```
pdillinger, line 331: Prefer `unique_ptr` to naked pointer when reasonably possible

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+Status FlushJob::MemPurge(autovector<MemTable*>& purged_mems) {
+  Status s;
+  db_mutex_->AssertHeld();
+  db_mutex_->Unlock();
+  assert(!mems_.empty());
+
+  // Store the full output memtables in
+  // autovector "purged_mems".
+  // autovector<MemTable*> purged_mems = {};
+  purged_mems = {};
```
pdillinger, line 329: Nit: for an accumulator output parameter, you have essentially three options:

* assert it's already empty
* just add to what's already there
* blindly replace anything that might be there

The last of these is my least favorite because it seems inherently unsafe from a maintenance perspective.

db/flush_job.cc
```diff
@@ -297,6 +303,11 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker,
            << (IOSTATS(cpu_read_nanos) - prev_cpu_read_nanos);
   }
 
+  // Clean up mempurge output memtables flushed to SST.
```
pdillinger, line 306: flushed to SST? I am confused about what `purged_mems` holds

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+    }
+  } else {
+    // If mempurge fails, simply delete newly created memtables
+    // and leave the mems_ memtables unmarked.
+    // Note that if mempurge fails, no memtable is added
+    // to cfd_->imm() so there is no need to edit cfd_->imm().
+    for (auto newm : purged_mems) {
+      // Paranoia
+      if (newm != nullptr) {
+        delete newm;
```
pdillinger, line 578: Double delete between here and caller?

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+  }
+
+  // Reacquire the mutex for WriteLevel0 function.
+  db_mutex_->Lock();
+
+  // If mempurge successful, don't write input tables to level0,
+  // but write any full output table to level0.
+  if (s.ok()) {
+    TEST_SYNC_POINT("DBImpl::FlushJob:MemPurgeSuccessful");
+    for (MemTable* m : mems_) {
```
pdillinger, line 559: How are we sure that mems_ here is the same as above? Does the flush thread effectively take ownership of mems_? Can we make that an assertion, or simply not rely on that?

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
...
+      assert(new_mem != nullptr);
+      delete new_mem;
+    }
+  }
+
+  // Reacquire the mutex for WriteLevel0 function.
+  db_mutex_->Lock();
+
+  // If mempurge successful, don't write input tables to level0,
+  // but write any full output table to level0.
```
pdillinger, line 556: I guess the track you're taking here is keeping some data in memory only to maximize the chances that it becomes garbage and never has to be flushed. My inclination is (per column family) to flush all or nothing, for the benefit of write amplification in compaction. When we create an L0 file, it should be as big as possible. This should also make it easier to purge obsolete WAL files in due course.

Doesn't necessarily have to be fixed now, but something to be noted for reconsideration.

pdillinger (changes requested):

db/db_flush_test.cc
bjlemaire: The tests are definitely on the slower end of the spectrum (all 3 tests take about 3 seconds each with a regular "make db_flush_test"). I can look at bringing them under 1sec each.

db/flush_job.cc
bjlemaire: Nice trick, I'll go ahead and edit that (even though long term there is a chance we need the flexibility a double would bring).

db/flush_job.cc
```diff
@@ -306,6 +317,272 @@ void FlushJob::Cancel() {
   base_->Unref();
 }
 
+Status FlushJob::MemPurge(autovector<MemTable*>& purged_mems) {
```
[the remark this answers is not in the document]
bjlemaire: Updated - thanks for the note!

db/flush_job.cc
```diff
@@ -227,7 +227,13 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker,
     prev_cpu_write_nanos = IOSTATS(cpu_write_nanos);
     prev_cpu_read_nanos = IOSTATS(cpu_read_nanos);
   }
+  autovector<MemTable*> purged_mems = {};
 
+  if (db_options_.experimental_allow_mempurge &&
+      (cfd_->GetFlushReason() == FlushReason::kWriteBufferFull) &&
+      (!mems_.empty())) {
+    Status mempurge_s = MemPurge(purged_mems);
```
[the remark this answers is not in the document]
bjlemaire: Updated - I cannot believe the ASSERT_CHECK_STATUS test didn't catch that.

db/flush_job.cc
```diff
@@ -227,7 +227,13 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker,
...
+  }
   // This will release and re-acquire the mutex.
   Status s = WriteLevel0Table();
```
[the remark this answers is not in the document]
bjlemaire: In the current iteration, a successful mempurge doesnt necessarily imply that there is no write-to-storage. Certain tables can be written to storage (not very obvious when the threshold to add the output table to imm is 10%, but more liekely to happen when this threshold is 50% and the average garbage ratio is >=50%) - they are then piggybacked onto the mems_ array. 
The GetMempurged is checked inside the writelevel0table function (https://github.com/facebook/rocksdb/pull/8505/files#r667020615). We then need to differentiate which tables have already been purged from the tables that we actually need to flush to storage.

db/flush_job.cc
bjlemaire: My bad, the name is misleading here. purged_mems are the memtables that "have been purged", meaning they dont contain any garbage. I am changing the name to "full_output_mems" (I make a distinction between input tables, containing garbage, and output tables, that contain "clean/valid" data). 
To answer your question, these tables are first added to this temporary vector, and then if the entire mempurge was seamless, they are piggybacked onto the mems_ memtable. The advantage of using this temp vector here is that rollbacking is very easy.
If we choose to abort mempurge if we fill more than one memtable, then we wont even need this vector anymore.

db/flush_job.cc
bjlemaire: Correct, that is another option. And this other option is actually a much easier implementation as well, because dealing with the full output memtables was delicate. 
And yes this is tested in MemPurge Basic.

db/flush_job.cc
bjlemaire: Agreed. I tried and this is just compliacted here, because you inherently need to lose ownership of the pointer to non-unique_ptr compatible objects. E.g: when adding the imm(), which only accepts raw pointers, we need to push back the pointer and not free the memory.

db/flush_job.cc
bjlemaire: Very true - I just updated the code with with the "assert it's already empty" option.

db/flush_job.cc
bjlemaire: My bad, the name `purged_mems` is visibly confusing. I've replaced it with "`full_output_mems`". It contains the memtables resulting from the mempurge that are filled up to capacity and need to be flushed to storage.

db/flush_job.cc
bjlemaire: Ouch good catch.

db/flush_job.cc
bjlemaire: Short answer: we are guaranteed that mems_ here is the same as above - no other flush thread can edit these memtables at the same time, and even additional operations like the `DB::Close` wait for all the flushes to happen before interfering with any of this. 
Long answer: `mems_` is a variable that belongs to the `FlushJob` object, and the `FlushJob` object is created inside each flush thread (details at db_impl_compaction_flush.cc:154). `mems_` simply contains the pointer addresses of the `imm()` memtables.
`mems_` is populated by the column family data object when the db_mutex is held (`FlushJob::PickMemTable`). Upon populating `mems_`, the `cfd` object mark that these memtables are being flushed by a flush thread somewhere ("`flush_in_progress_ = true`", memtable_list.cc:357), which guarantees that nothing happens to the memtables contained in the `mems_` object while the flush job is handling them.
Therefore we can be sure that there is no race condition thanks to the `db_mutex`, and since the memtables are marked with "`flush_in_progress=true`", we know that these tables wont be edited while we are working on them in the flush process.

db/flush_job.cc
bjlemaire: Agreed, provided we're allowed to go beyond the size of a regular SST file (so that we dont accidentally create 2 sst files, one of them being full, the other one filled at xx% capacity). According to the builder.cc:BuildTable code, it looks like we would indeed create a single large SST file, which would be ideal.

bjlemaire: Before this comment, i actually thought that would lead to the creation of more than 1 sst file because i was convinced the SST file size was enforced. But after code inspection it looks like we would just write a single big SST file.

db/db_flush_test.cc
```diff
@@ -697,74 +697,133 @@ TEST_F(DBFlushTest, MemPurgeBasic) {
...
+  // Check that there was no SST files created during flush.
+  const uint32_t EXPECTED_SST_COUNT = 0;
 
   EXPECT_GE(mempurge_count, EXPECTED_MIN_MEMPURGE_COUNT);
-  EXPECT_EQ(flush_count, EXPECTED_FLUSH_COUNT);
+  EXPECT_EQ(sst_count, EXPECTED_SST_COUNT);
+
+  const uint32_t mempurge_count_record = mempurge_count;
+
+  // Insertion of of K-V pairs, multiple times.
```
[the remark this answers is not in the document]
bjlemaire: Fixed

db/flush_job.cc
pdillinger: The SST file size limit is for breaking up sorted runs, primarily so that leveled compaction in L1 and above can operate on parts of sorted runs with reasonable granularity. L0 is special because (AFAIK) we assume each SST file is its own sorted run. :)

db/flush_job.cc
pdillinger: Can use `.release()`

db/flush_job.cc
```diff
@@ -227,12 +227,14 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker,
...
   }
-  autovector<MemTable*> purged_mems = {};
 
   if (db_options_.experimental_allow_mempurge &&
       (cfd_->GetFlushReason() == FlushReason::kWriteBufferFull) &&
       (!mems_.empty())) {
-    Status mempurge_s = MemPurge(purged_mems);
+    Status mempurge_s = MemPurge();
+    if (!mempurge_s.ok()) {
+      ROCKS_LOG_INFO(db_options_.info_log, "Mempurge process unsuccessful.");
```
pdillinger, line 236: Please include status details, e.g. from ToString(). I'm sure you can find an example to copy.

It seems like an Aborted status should just be INFO but any other status is maybe WARN.

db/flush_job.cc
```diff
@@ -370,6 +362,8 @@ Status FlushJob::MemPurge(autovector<MemTable*>& purged_mems) {
   // or at least range tombstones, copy over the info
   // to the new memtable.
   if (iter->Valid() || !range_del_agg->IsEmpty()) {
+    // Arbitrary heuristic: maxSize is 60% cpacity.
+    size_t maxSize = ((mutable_cf_options_.write_buffer_size + 6ULL) / 10ULL);
```
pdillinger, line 366: `* 6U) / 10U`

Also, `42ULL` probably won't fit in size_t on 32-bit. Just be consistent about unsigned vs. signed in arithmetic, and promotion will generally take care of type sizes.

db/flush_job.cc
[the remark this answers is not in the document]
pdillinger: A smart person just suggested this "purged" state might be converted to temporary state here in FlushJob::Run ;)

db/flush_job.cc
```diff
@@ -454,18 +448,12 @@ Status FlushJob::MemPurge(autovector<MemTable*>& purged_mems) {
...
-        new_mem =
-            new MemTable((cfd_->internal_comparator()), *(cfd_->ioptions()),
-                         mutable_cf_options_, cfd_->write_buffer_mgr(),
-                         mems_[0]->GetEarliestSequenceNumber(), cfd_->GetID());
-        new_first_seqno = kMaxSequenceNumber;
+      // If new_mem has size greater than maxSize,
+      // then rollback to regular flush operation,
+      // and destroy new_mem.
+      if (new_mem->ApproximateMemoryUsage() > maxSize) {
+        s = Status::Aborted(Slice("Mempurge filled more than one memtable."));
```
pdillinger, line 455: I don't think you need `Slice(` due to implicit conversions char[] -> char* -> std::string -> Slice

pdillinger (approved):
Some minor suggestions / fixes remain. Otherwise looking good :)

db/flush_job.cc
bjlemaire: Fixed!

db/flush_job.cc
bjlemaire: Fixed

PR #8506, opened 2021-07-08
[ResourceManagement] Refactor WriteBufferManager::CacheRep into CacheReservationManager

hx235:
Context:
To help cap various memory usage by a single limit of the block cache capacity, we charge the memory usage through inserting/releasing dummy entries in the block cache. CacheReservationManager is such a class (non thread-safe) responsible for  inserting/removing dummy entries to reserve cache space for memory used by the class user. 

Summary: 
- Refactored the inner private class CacheRep of WriteBufferManager into public CacheReservationManager class for reusability such as for https://github.com/facebook/rocksdb/pull/8428

- Encapsulated implementation details of cache key generation and dummy entries insertion/release in cache reservation as discussed in https://github.com/facebook/rocksdb/pull/8506#discussion_r666550838

- Consolidated increase/decrease cache reservation into one API - UpdateCacheReservation. 

- Adjusted the previous dummy entry release algorithm in decreasing cache reservation to be loop-releasing dummy entries to stay symmetric to dummy entry insertion algorithm
 
- Made the previous dummy entry release algorithm in delayed decrease mode more aggressive for better decreasing cache reservation when memory used is less likely to increase back.

  Previously, the algorithms only release 1 dummy entries when new_mem_used < 3/4 * cache_allocated_size_ and cache_allocated_size_ - kSizeDummyEntry > new_mem_used. 
Now, the algorithms loop-releases as many dummy entries as possible when new_mem_used < 3/4 * cache_allocated_size_.

- Updated WriteBufferManager's test cases to adapt to changes on the release algorithm mentioned above and left comment for some test cases for clarity

- Replaced the previous cache key prefix generation (utilizing object address related to the cache client) with one that utilizes Cache->NewID() to prevent cache-key collision among dummy entry clients sharing the same cache. 

  The specific collision we are preventing happens when the object address is reused for a new cache-key prefix while the old cache-key using that same object address in its prefix still exists in the cache. This could happen due to that, under LRU cache policy, there is a possible delay in releasing a cache entry after the cache client object owning that cache entry get deallocated. In this case, the object address related to the cache client object can get reused for other client object to generate a new cache-key prefix.

  This prefix generation can be made obsolete after Peter's unification of all the code generating cache key, mentioned in https://github.com/facebook/rocksdb/pull/8506#discussion_r667265255

Test plan:
- Passing the added unit tests cache_reservation_manager_test.cc
- Passing existing and adjusted write_buffer_manager_test.cc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1897,7 +1897,9 @@ clipping_iterator_test: $(OBJ_DIR)/db/compaction/clipping_iterator_test.o $(TEST
 
 ribbon_bench: $(OBJ_DIR)/microbench/ribbon_bench.o $(LIBRARY)
 	$(AM_LINK)
-
+	
+cache_reservation_manager_test: $(OBJ_DIR)/cache/cache_reservation_manager_test.o $(TEST_LIBRARY) $(LIBRARY)
+	$(AM_LINK)
 #-------------------------------------------------
 # make install related stuff
 PREFIX ?= /usr/local
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -43,12 +43,7 @@ Status CacheReservationManager::UpdateCacheReservation(
     std::size_t new_mem_used) {
   std::size_t cur_cache_allocated_size =
       cache_allocated_size_.load(std::memory_order_relaxed);
-  if (new_mem_used == cur_cache_allocated_size ||
-      (cur_cache_allocated_size > 0 &&
-       new_mem_used > cur_cache_allocated_size &&
-       (new_mem_used - cur_cache_allocated_size) < kSizeDummyEntry) ||
-      (new_mem_used < cur_cache_allocated_size &&
-       (cur_cache_allocated_size - new_mem_used) < kSizeDummyEntry)) {
+  if (new_mem_used == cur_cache_allocated_size) {
     return Status::OK();
   } else if (new_mem_used > cur_cache_allocated_size) {
     Status s = IncreaseCacheReservation<R>(new_mem_used);
@@ -116,7 +111,10 @@ Status CacheReservationManager::IncreaseCacheReservation(
 Status CacheReservationManager::DecreaseCacheReservation(
     std::size_t new_mem_used) {
   Status return_status = Status::OK();
-  while (new_mem_used < cache_allocated_size_.load(std::memory_order_relaxed)) {
+
+  // Decrease to the smallest multiple of kSizeDummyEntry that is greater than or equal to new_mem_used
+  // We do addition instead of new_mem_used <= cache_allocated_size_.load(std::memory_order_relaxed) - kSizeDummyEntry to avoid underflow of size_t when cache_allocated_size_ = 0
+  while (new_mem_used + kSizeDummyEntry <= cache_allocated_size_.load(std::memory_order_relaxed)) {
     assert(!dummy_handles_.empty());
     auto* handle = dummy_handles_.back();
     // If insert failed, handle is null so we should not release.
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -1,0 +1,144 @@
+//  Copyright (c) 2011-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under both the GPLv2 (found in the
+//  COPYING file in the root directory) and Apache 2.0 License
+//  (found in the LICENSE.Apache file in the root directory).
+//
+// Copyright (c) 2011 The LevelDB Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file. See the AUTHORS file for names of contributors.
+#include "cache/cache_reservation_manager.h"
+
+#include <cassert>
+#include <cstddef>
+#include <cstring>
+#include <memory>
+
+#include "cache/cache_entry_roles.h"
+#include "rocksdb/cache.h"
+#include "rocksdb/slice.h"
+#include "rocksdb/status.h"
+#include "table/block_based/block_based_table_reader.h"
+#include "util/coding.h"
+
+namespace ROCKSDB_NAMESPACE {
+CacheReservationManager::CacheReservationManager(std::shared_ptr<Cache> cache,
+                                                 bool delayed_decrease)
+    : delayed_decrease_(delayed_decrease), cache_allocated_size_(0) {
+  assert(cache != nullptr);
+  cache_ = cache;
+  std::memset(cache_key_, 0, kCacheKeyPrefixSize + kMaxVarint64Length);
+  EncodeVarint64(cache_key_, cache_->NewId());
+}
+
+CacheReservationManager::~CacheReservationManager() {
+  for (auto* handle : dummy_handles_) {
+    if (handle != nullptr) {
+      cache_->Release(handle, true);
+    }
+  }
+}
+
+template <CacheEntryRole R>
+Status CacheReservationManager::UpdateCacheReservation(
+    std::size_t new_mem_used) {
+  std::size_t cur_cache_allocated_size =
+      cache_allocated_size_.load(std::memory_order_relaxed);
+  if (new_mem_used == cur_cache_allocated_size ||
+      (new_mem_used > cur_cache_allocated_size &&
+       (new_mem_used - cur_cache_allocated_size) < kSizeDummyEntry) ||
+      (new_mem_used < cur_cache_allocated_size &&
+       (cur_cache_allocated_size - new_mem_used) < kSizeDummyEntry)) {
+    return Status::OK();
+  } else if (new_mem_used > cur_cache_allocated_size) {
+    Status s = IncreaseCacheReservation<R>(new_mem_used);
+    return s;
+  } else {
+    // In delayed decrease mode, we don't decrease cache reservation
+    // untill the memory usage is less than 3/4 of what we reserve
+    // in the cache.
+    // We do this because
+    // (1) Dummy entry insertion is expensive in block cache
+    // (2) Delayed releasing previously inserted dummy entries can save such
+    // expensive dummy entry insertion on memory increase in the near future,
+    // which is likely to happen when the memory usage is greater than or equal
+    // to 3/4 of what we reserve
+    if (delayed_decrease_ && new_mem_used >= cur_cache_allocated_size / 4 * 3) {
+      return Status::OK();
+    } else {
+      Status s = DecreaseCacheReservation(new_mem_used);
+      return s;
+    }
+  }
+}
+
+// Explicitly instantiate templates for "CacheEntryRole" values we use.
+// This makes it possible to keep the template definitions in the .cc file.
+template Status CacheReservationManager::UpdateCacheReservation<
+    CacheEntryRole::kWriteBuffer>(std::size_t new_mem_used);
+
+template <CacheEntryRole R>
+Status CacheReservationManager::IncreaseCacheReservation(
+    std::size_t new_mem_used) {
+  Status return_status = Status::OK();
+  while (new_mem_used > cache_allocated_size_.load(std::memory_order_relaxed)) {
+    Cache::Handle* handle = nullptr;
+    Status s = cache_->Insert(GetNextCacheKey(), nullptr, kSizeDummyEntry,
+                              GetNoopDeleterForRole<R>(), &handle);
+
+    // We do not return immeditately when encountering dummy entry insertion
+    // error to keep cache_allocated_size_ easy to reasonate. In other words, we
+    // still account those dummy entries that fail insertion in
+    // cache_allocation_size_. As a consequence, there will exist null dummy
+    // handles in dummy_handles_ if insertion error occurs.
+
+    // Ideallly we should prevent this allocation from happening if
+    // this insertion fails. However, the callers to this code path
+    // might not be able to handle failures properly. [TODO] We'll need to
+    // improve it in the future and figure out what to do on error
+    s.PermitUncheckedError();
+
+    // We do not overwrite any non-OK status from dummy entry insertion.
+    // This is to prevent possible subsequent successful dummy entry insertion
+    // (due to newly available cache space resulting from manipulation in other
+    // threads) from overriding the non-OK status
+    if (return_status == Status::OK()) {
+      return_status = s;
+    }
+
+    dummy_handles_.push_back(handle);
+    cache_allocated_size_ += kSizeDummyEntry;
+  }
+  return return_status;
+}
+
+Status CacheReservationManager::DecreaseCacheReservation(
+    std::size_t new_mem_used) {
+  Status return_status = Status::OK();
+  while (new_mem_used < cache_allocated_size_.load(std::memory_order_relaxed)) {
+    assert(!dummy_handles_.empty());
+    auto* handle = dummy_handles_.back();
+    // If insert failed, handle is null so we should not release.
+    if (handle != nullptr) {
+      cache_->Release(handle, true);
+    }
+    dummy_handles_.pop_back();
+    cache_allocated_size_ -= kSizeDummyEntry;
+  }
+  return return_status;
+}
+
+std::size_t CacheReservationManager::GetTotalReservedCacheSize() {
+  return cache_allocated_size_.load(std::memory_order_relaxed);
+}
+
+Slice CacheReservationManager::GetNextCacheKey() {
+  // Calling this function will have the side-effect of changing the
+  // underlying cache_key_ that is shared among other keys generated from this
+  // fucntion. Therefore please make sure the previous keys are saved/copied
+  // before calling this function.
+  std::memset(cache_key_ + kCacheKeyPrefixSize, 0, kMaxVarint64Length);
+  char* end =
+      EncodeVarint64(cache_key_ + kCacheKeyPrefixSize, next_cache_key_id_++);
+  return Slice(cache_key_, static_cast<std::size_t>(end - cache_key_));
+}
+}  // namespace ROCKSDB_NAMESPACE
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -20,7 +20,7 @@
 #include "table/block_based/block_based_table_reader.h"
 #include "util/coding.h"
 
-#include <iostream> //for debugging, to be removed
+
 
 namespace ROCKSDB_NAMESPACE {
 CacheReservationManager::CacheReservationManager(std::shared_ptr<Cache> cache,
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -80,9 +80,9 @@ Status CacheReservationManager::IncreaseCacheReservation(
   while (new_mem_used > cache_allocated_size_.load(std::memory_order_relaxed)) {
     Cache::Handle* handle = nullptr;
     return_status = cache_->Insert(GetNextCacheKey(), nullptr, kSizeDummyEntry,
-                               GetNoopDeleterForRole<R>(), &handle);
+                                   GetNoopDeleterForRole<R>(), &handle);
 
-    if (return_status != Status::OK()){
+    if (return_status != Status::OK()) {
       return return_status;
     }
 
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -20,6 +20,8 @@
 #include "table/block_based/block_based_table_reader.h"
 #include "util/coding.h"
 
+#include <iostream> //for debugging, to be removed
+
 namespace ROCKSDB_NAMESPACE {
 CacheReservationManager::CacheReservationManager(std::shared_ptr<Cache> cache,
                                                  bool delayed_decrease)
@@ -44,7 +46,7 @@ Status CacheReservationManager::UpdateCacheReservation(
   std::size_t cur_cache_allocated_size =
       cache_allocated_size_.load(std::memory_order_relaxed);
   if (new_mem_used == cur_cache_allocated_size ||
-      (new_mem_used > cur_cache_allocated_size &&
+      (cur_cache_allocated_size > 0 && new_mem_used > cur_cache_allocated_size &&
        (new_mem_used - cur_cache_allocated_size) < kSizeDummyEntry) ||
       (new_mem_used < cur_cache_allocated_size &&
        (cur_cache_allocated_size - new_mem_used) < kSizeDummyEntry)) {
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -112,9 +112,12 @@ Status CacheReservationManager::DecreaseCacheReservation(
     std::size_t new_mem_used) {
   Status return_status = Status::OK();
 
-  // Decrease to the smallest multiple of kSizeDummyEntry that is greater than or equal to new_mem_used
-  // We do addition instead of new_mem_used <= cache_allocated_size_.load(std::memory_order_relaxed) - kSizeDummyEntry to avoid underflow of size_t when cache_allocated_size_ = 0
-  while (new_mem_used + kSizeDummyEntry <= cache_allocated_size_.load(std::memory_order_relaxed)) {
+  // Decrease to the smallest multiple of kSizeDummyEntry that is greater than
+  // or equal to new_mem_used We do addition instead of new_mem_used <=
+  // cache_allocated_size_.load(std::memory_order_relaxed) - kSizeDummyEntry to
+  // avoid underflow of size_t when cache_allocated_size_ = 0
+  while (new_mem_used + kSizeDummyEntry <=
+         cache_allocated_size_.load(std::memory_order_relaxed)) {
     assert(!dummy_handles_.empty());
     auto* handle = dummy_handles_.back();
     // If insert failed, handle is null so we should not release.
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -77,6 +77,9 @@ Status CacheReservationManager::UpdateCacheReservation(
 // This makes it possible to keep the template definitions in the .cc file.
 template Status CacheReservationManager::UpdateCacheReservation<
     CacheEntryRole::kWriteBuffer>(std::size_t new_mem_used);
+// For cache reservation manager unit tests
+template Status CacheReservationManager::UpdateCacheReservation<
+    CacheEntryRole::kMisc>(std::size_t new_mem_used);
 
 template <CacheEntryRole R>
 Status CacheReservationManager::IncreaseCacheReservation(
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -103,7 +103,7 @@ Status CacheReservationManager::IncreaseCacheReservation(
       return_status = s;
     }
 
-    // We absorb the error here since s is conditionally checked 
+    // We absorb the error here since s is conditionally checked
     // when return_status == Status::OK()
     s.PermitUncheckedError();
 
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -32,9 +32,7 @@ CacheReservationManager::CacheReservationManager(std::shared_ptr<Cache> cache,
 
 CacheReservationManager::~CacheReservationManager() {
   for (auto* handle : dummy_handles_) {
-    if (handle != nullptr) {
-      cache_->Release(handle, true);
-    }
+    cache_->Release(handle, true);
   }
 }
 
@@ -81,26 +79,12 @@ Status CacheReservationManager::IncreaseCacheReservation(
   Status return_status = Status::OK();
   while (new_mem_used > cache_allocated_size_.load(std::memory_order_relaxed)) {
     Cache::Handle* handle = nullptr;
-    Status s = cache_->Insert(GetNextCacheKey(), nullptr, kSizeDummyEntry,
-                              GetNoopDeleterForRole<R>(), &handle);
+    return_status = cache_->Insert(GetNextCacheKey(), nullptr, kSizeDummyEntry,
+                               GetNoopDeleterForRole<R>(), &handle);
 
-    // We do not return immeditately when encountering dummy entry insertion
-    // error to keep cache_allocated_size_ easy to reasonate. In other words, we
-    // still account those dummy entries that fail insertion in
-    // cache_allocation_size_. As a consequence, there will exist null dummy
-    // handles in dummy_handles_ if insertion error occurs.
-    //
-    // Also, We do not overwrite any non-OK status from dummy entry insertion.
-    // This is to prevent possible subsequent successful dummy entry insertion
-    // (due to newly available cache space resulting from manipulation in other
-    // threads) from overriding the non-OK status
-    if (return_status == Status::OK()) {
-      return_status = s;
+    if (return_status != Status::OK()){
+      return return_status;
     }
-
-    // We absorb the error here since s is conditionally checked
-    // when return_status == Status::OK()
-    s.PermitUncheckedError();
 
     dummy_handles_.push_back(handle);
     cache_allocated_size_ += kSizeDummyEntry;
@@ -120,10 +104,7 @@ Status CacheReservationManager::DecreaseCacheReservation(
          cache_allocated_size_.load(std::memory_order_relaxed)) {
     assert(!dummy_handles_.empty());
     auto* handle = dummy_handles_.back();
-    // If insert failed, handle is null so we should not release.
-    if (handle != nullptr) {
-      cache_->Release(handle, true);
-    }
+    cache_->Release(handle, true);
     dummy_handles_.pop_back();
     cache_allocated_size_ -= kSizeDummyEntry;
   }
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -94,20 +94,18 @@ Status CacheReservationManager::IncreaseCacheReservation(
     // still account those dummy entries that fail insertion in
     // cache_allocation_size_. As a consequence, there will exist null dummy
     // handles in dummy_handles_ if insertion error occurs.
-
-    // Ideallly we should prevent this allocation from happening if
-    // this insertion fails. However, the callers to this code path
-    // might not be able to handle failures properly. [TODO] We'll need to
-    // improve it in the future and figure out what to do on error
-    s.PermitUncheckedError();
-
-    // We do not overwrite any non-OK status from dummy entry insertion.
+    //
+    // Also, We do not overwrite any non-OK status from dummy entry insertion.
     // This is to prevent possible subsequent successful dummy entry insertion
     // (due to newly available cache space resulting from manipulation in other
     // threads) from overriding the non-OK status
     if (return_status == Status::OK()) {
       return_status = s;
     }
+
+    // We absorb the error here since s is conditionally checked 
+    // when return_status == Status::OK()
+    s.PermitUncheckedError();
 
     dummy_handles_.push_back(handle);
     cache_allocated_size_ += kSizeDummyEntry;
--- a/cache/cache_reservation_manager.cc
+++ b/cache/cache_reservation_manager.cc
@@ -20,8 +20,6 @@
 #include "table/block_based/block_based_table_reader.h"
 #include "util/coding.h"
 
-
-
 namespace ROCKSDB_NAMESPACE {
 CacheReservationManager::CacheReservationManager(std::shared_ptr<Cache> cache,
                                                  bool delayed_decrease)
@@ -46,7 +44,8 @@ Status CacheReservationManager::UpdateCacheReservation(
   std::size_t cur_cache_allocated_size =
       cache_allocated_size_.load(std::memory_order_relaxed);
   if (new_mem_used == cur_cache_allocated_size ||
-      (cur_cache_allocated_size > 0 && new_mem_used > cur_cache_allocated_size &&
+      (cur_cache_allocated_size > 0 &&
+       new_mem_used > cur_cache_allocated_size &&
        (new_mem_used - cur_cache_allocated_size) < kSizeDummyEntry) ||
       (new_mem_used < cur_cache_allocated_size &&
        (cur_cache_allocated_size - new_mem_used) < kSizeDummyEntry)) {
--- a/cache/cache_reservation_manager.h
+++ b/cache/cache_reservation_manager.h
@@ -7,8 +7,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 //
-// CacheReservationManager is for reserving cache space through
-// inserting/removing dummy entries based on the memory used.
+// CacheReservationManager is for reserving cache space for the memory used 
+// through inserting/releasing dummy entries in the cache. 
 // This class is not thread-safe.
 #pragma once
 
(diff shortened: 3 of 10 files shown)


cache/cache_rep.h
```diff
@@ -0,0 +1,53 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <vector>
+
+#include "rocksdb/cache.h"
+#include "rocksdb/slice.h"
+
+// [DO-NOT-COMMIT] [A header inclusion issue][A temporary hack]
```
hx235, line 22: [NEED_ATTENTION]
Please line 22 - 27 and line 35 - 40 for a brief description of the issue. We can talk offline for this.  The current hack passes the newly added unit test and the existing write_buffer_manager_test.cc

cache/cache_rep.h
```diff
@@ -0,0 +1,53 @@
...
+// Including the following two dependencies in cache_rep.h leads to a wall of compiler errors when including cache_rep.h in write_buffer_manager.h
+// It is suspected to be circular dependency issue
+// Therefore I commented these two dependencies out and redefine the const from these two dependencies in this file [DO-NOT-COMMIT] 
+// #include "table/block_based/block_based_table_reader.h"
+// #include "util/coding.h"
+
+namespace ROCKSDB_NAMESPACE {
+class CacheRep {
```
hx235, line 30: [non-blocking]
question 1: what is the role of such a CacheRep class when being used by other dummy entries clients?

my thought so far: my current implementation basically follows the previous design of the CacheRep as an inner private class in WBM. But now since I have more c++ class knowledge, I start to question if this is what we want.

Do we want this CacheRep to be (1)just a C-style struct that gathers different values without any encapsulation (everything being public to the CacheRep user) OR (2)a class that provides encapsulation that could take care (e.g, encapsulate) some dummy entry insertion/release stuff? 

It seems to me that the previous CacheRep design lean towards (1) But now since other dummy entry clients are going to reuse the class, I wonder if (2) is what we want. If (2) is what we want, a much more careful design is needed for the interface and a bit more work in the WBM side than now.

cache/cache_rep.cc
```diff
@@ -0,0 +1,35 @@
...
+  cache_ = cache;
+  std::memset(cache_key_, 0, kCacheKeyPrefixSize + kMaxVarint64Length);
+  EncodeVarint64(cache_key_, cache->NewId());
+}
+
+Slice CacheRep::GetNextCacheKey() {
+  std::memset(cache_key_ + kCacheKeyPrefixSize, 0, kMaxVarint64Length);
+  char* end =
+      EncodeVarint64(cache_key_ + kCacheKeyPrefixSize, next_cache_key_id_++);
+  return Slice(cache_key_, static_cast<std::size_t>(end - cache_key_));
```
hx235, line 33: [non-blocking]
question 2: just to double check my understanding, are we returning a pointer to the underlying char array of cache_key_, instead of a new char array, FOR efficiency? 

Although this actually gave a chance to GetNextCacheKey()'s caller (or even CacheRep class user) to modify the underlying char array the way they want. If the returned Slice is used after such extra modification, the underlying cache_key_ might be inconsistent.

memtable/write_buffer_manager.cc
```diff
@@ -17,37 +17,7 @@ namespace ROCKSDB_NAMESPACE {
 #ifndef ROCKSDB_LITE
 namespace {
 const size_t kSizeDummyEntry = 256 * 1024;
-// The key will be longer than keys for blocks in SST files so they won't
-// conflict.
-const size_t kCacheKeyPrefix = kMaxVarint64Length * 4 + 1;
 }  // namespace
-
-struct WriteBufferManager::CacheRep {
```
hx235, line 25: [non-blocking]
Question 3: just a general question on WBM to understand the code better: why do we need things like "std::lock_guard<std::mutex> lock(cache_rep_->cache_mutex_);" at line 100? What concurrency issue we are trying to prevent here?

cache/cache_rep.h
```diff
@@ -0,0 +1,53 @@
...
+  static const std::size_t kMaxVarint64Length = 10;
+  static const std::size_t kCacheKeyPrefixSize = kMaxVarint64Length * 4 + 1;
+
+  explicit CacheRep(std::shared_ptr<Cache> cache);
+  Slice GetNextCacheKey();
+
+  std::shared_ptr<Cache> cache_;
+  std::mutex cache_mutex_;
+  std::atomic<std::size_t> cache_allocated_size_;
+  std::uint64_t next_cache_key_id_ = 0;
```
hx235, line 48: [minor_improvement]
I declared "std::uint64_t next_cache_key_id_ = 0;" before "cache_key_" for better readability

cache/cache_rep.cc
```diff
@@ -0,0 +1,35 @@
...
+#include "rocksdb/slice.h"
+#include "table/block_based/block_based_table_reader.h"
+#include "util/coding.h"
+
+namespace ROCKSDB_NAMESPACE {
+CacheRep::CacheRep(std::shared_ptr<Cache> cache)
+    : cache_allocated_size_(0) {
+  assert(cache != nullptr);
+  cache_ = cache;
+  std::memset(cache_key_, 0, kCacheKeyPrefixSize + kMaxVarint64Length);
```
hx235, line 25: [minor_improvement]
I initialized the full char array "cache_key_" in constructor so that all parts of the array is well-defined after object construction.

cache/cache_rep.cc
```diff
@@ -0,0 +1,35 @@
...
+#include "cache/cache_rep.h"
+#include "rocksdb/cache.h"
+#include "rocksdb/slice.h"
+#include "table/block_based/block_based_table_reader.h"
+#include "util/coding.h"
+
+namespace ROCKSDB_NAMESPACE {
+CacheRep::CacheRep(std::shared_ptr<Cache> cache)
+    : cache_allocated_size_(0) {
+  assert(cache != nullptr);
```
hx235, line 23: [minor_improvement]
I think it is safe to use "assert" here for an extra check because the CacheRep class users should pass in a valid Cache pointer if they want a CacheRep on it. I also double checked that the only user of this CacheRep class - WBM - did null check against the cache before passing it into the constructor.

memtable/write_buffer_manager.cc
```diff
@@ -17,37 +17,7 @@ namespace ROCKSDB_NAMESPACE {
...
-
-  Slice GetNextCacheKey() {
-    memset(cache_key_ + kCacheKeyPrefix, 0, kMaxVarint64Length);
-    char* end =
-        EncodeVarint64(cache_key_ + kCacheKeyPrefix, next_cache_key_id_++);
-    return Slice(cache_key_, static_cast<size_t>(end - cache_key_));
-  }
-};
-#else
-struct WriteBufferManager::CacheRep {};
```
hx235, line 50: [minor_improvement]
I think it is safe to completely remove the "else" branch here since (1) there is no longer such an inner CacheRep class declaration that needs a definition and (2)when "ROCKSDB_LITE" is defined, there isn't any code path that uses CacheRep class

cache/cache_rep.cc
```diff
@@ -0,0 +1,35 @@
...
+  cache_ = cache;
+  std::memset(cache_key_, 0, kCacheKeyPrefixSize + kMaxVarint64Length);
+  EncodeVarint64(cache_key_, cache->NewId());
+}
+
+Slice CacheRep::GetNextCacheKey() {
+  std::memset(cache_key_ + kCacheKeyPrefixSize, 0, kMaxVarint64Length);
```
zhichao-cao, line 30: I don't think we need to initialize it here. EncodeVarin64 will do the job later.

cache/cache_rep.cc
zhichao-cao, line 25: Also may not be necessary.

cache/cache_rep.h
```diff
@@ -0,0 +1,53 @@
...
+class CacheRep {
+ public:
+  // The key will be longer than keys for blocks in SST files so they won't
+  // conflict.
+
+  // [DO-NOT-COMMIT][A header inclusion issue]
+  // The following four lines (including the two commented lines) are for adapting to the [A temporary hack] mentioned above
+  // static const std::size_t kCacheKeyPrefixSize =
+  //     BlockBasedTable::kMaxCacheKeyPrefixSize + kMaxVarint64Length;
+  static const std::size_t kMaxVarint64Length = 10;
```
zhichao-cao, line 39: why we need to redefine them here? In my impression, they are already defined in coding.h

hx235: Great question - it is actually due to the reason in the note I left in line 22 in this PR's comment section. This is a temporary hack that is not intended to be committed (therefore this PR is marked "early review"). The reason why this hack exists is that - when including the header where this constant is defined (i.e #include "table/block_based/block_based_table_reader.h") in this file (i.e, cache_rep.h) and including this file (i.e, cache_rep.h) in write_buffer_manager.h, which is what I have to do, there is a wall of compiler errors, indicating a circular dependency issue that I am still troubleshooting for.  Therefore I added the hack here just to bypass this particular issue for reviewing purpose.

I left a note in line 22 and some code comments (e.g, line 22 - 27, line 35 - 38) for this issue. But reading your question makes me realize the way I flag out this hack to potential reviewers is not enough :) Let me flag this issue out more explicitly in the PR comment section. Thank you for you question!

cache/cache_rep.cc
hx235: Good question! I was wondering the same thing when developing :)

However, I think L30 is still needed mainly for the following reason:
L32 in the function "GetNextCacheKey()" actually stores the encoded value of "next_cache_key_id_" in-place into "cache_key_" as the non-prefix part of the cache key. In another word, we generate a new non-prefix part of the cache key every time but keep storing it in the same place (the place is "cache_key_ + kCacheKeyPrefixSize").

Therefore, we would like to reset the place "cache_key_ + kCacheKeyPrefixSize" every time when we want to store a new non-prefix part of the cache key.  

The reason for why we would like to store cache key in place, I guess, is for memory and speed efficiency. 

But now reading your question make me consider it might be worthwhile to add a comment for line 30 ... let me see ...

cache/cache_rep.h
```diff
@@ -20,9 +20,11 @@
 #include "rocksdb/slice.h"
 
 // [DO-NOT-COMMIT] [A header inclusion issue][A temporary hack]
```
hx235, line 22: [BLOCKER][NEED_ATTENTION]
As noted in the code line 22-27, there exists a "suspected-to-be" circular dependency issue when trying to include "table/block_based/block_based_table_reader.h" and "util/coding.h" in the file "cache_rep.h" AND include "cache_rep.h" in "write_buffer_manager.h". We can talk more about this offline so that I can demonstrate the compiler warning. 

Therefore a temporary hack (in line 37-42) of redefining the constants (i.e, BlockBasedTable::kMaxCacheKeyPrefixSize and kMaxVarint64Length) needed from these two headers (i.e,  "table/block_based/block_based_table_reader.h" and "util/coding.h") is included in the current implementation. The hack is here just to bypass this issue that I am still troubleshooting for with my mentors and ideally not to be committed.

If you as reviewers have run into a similar issue before or have any idea about efficiently troubleshooting a circular-dependency issue, I am happy to learn about it :)

cache/cache_rep.h
pdillinger: My early impression is that kCacheKeyPrefixSize can live in cache_rep.h. I don't think we need to spend too much time polishing these details, as I'm working on a replacement schema for cache keys, that would make them all 16 bytes. (Details hopefully in next day or so.) We probably wouldn't switch over to the new schema all at once, so the current schema might need to stick around for a while.

But work toward unifying all the code generating cache keys under a shared API is moving in the right direction, and I'm not really to a point of getting merge/rebase conflicts with you.

Makefile
```diff
@@ -1898,6 +1898,9 @@ clipping_iterator_test: $(OBJ_DIR)/db/compaction/clipping_iterator_test.o $(TEST
 ribbon_bench: $(OBJ_DIR)/microbench/ribbon_bench.o $(LIBRARY)
 	$(AM_LINK)
 
+cache_rep_test: $(OBJ_DIR)/cache/cache_rep_test.o $(TEST_LIBRARY) $(LIBRARY)
```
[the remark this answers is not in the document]
hx235: Interesting! I never thought of that aspect at all! Does "new test binary" refer to the "cache_rep_test" executable produced after the command "make"?

My sole reason of adding the line 1901 here is to create a rule in Makefile so that I can build and run this specific test by doing "make -jN cache_rep_test" and "./cache_rep_test". This also allow others to do the same thing. The ultimate goal here is to facilitate a smaller hence faster testing, for people just interested in testing my cache_rep.

However, there could be other ways to achieve a similar purpose as above that don't need to "add new test binaries" that I don't know of. But I am more than happy to learn about them :)

cache/cache_rep.h
hx235: Thank you for providing the context on cache key unification! I am aware of this unification before starting this task and very excited about it :)

In terms of kCacheKeyPrefixSize, I will still refer to the coding.h since the "suspected-to-be" circular dependency issue mentioned above is made irrelevant to this PR in a non-hack way after a minor fix I did just now. But thank you for the input!

@zhichao-cao tagged you here just for FYI for the update on the "suspected-to-be" circular dependency issue mentioned in my reply to your comment above. Now this issue is made irrelevant to my PR in a non-hack way, see my latest commit https://github.com/facebook/rocksdb/pull/8506/commits/e5ddc41799a616e3d35946b8df9e70f9e0b83e3a) for more information if you are interested.

Makefile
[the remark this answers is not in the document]
hx235: Interesting! This aspect didn't even occur to my mind when I was adding the line, indicating that there is something more in the RocksDB's building process I need to learn :)

So here is a learning question: I was under the impression that the new test binaries (in my case, cache_rep_test executable) are created only under after running: 
(1)"make check" or "make all". The executable will be created regardless of whether line 1901 exists since "cache_rep_test.cc" is in "TEST_MAIN_SOURCE" in "src.mk". 
Or (2)"make cache_rep_test". The rule "cache_rep_test" is valid and the executable will be created after running the rule because I added line 1901. 

Both (1), (2) are intentional commands that are run by people who probably want to run the "cache_rep_test" executable for testing. If people don't run (1) nor (2), no "cache_rep_test test" executable (binary) will be created/built. (This is the part I am not sure about - is there any other automatic mechanism going on that will automatically create new "cache_rep_test" executable/binary even when people don't want to run "make check/make all/make cache_rep_test", just because I added "cache_rep_test.cc" is in "TEST_MAIN_SOURCE" in "src.mk" and added line 1901?)

Also just for me to make sure I understand your comment - are you suggesting to (1)take out "cache_rep_test.cc" from src.mk OR/AND (2)take out the line 1901 in the MakeFile?

Thanks!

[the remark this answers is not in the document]
pdillinger: I'm suggesting adding new tests to an existing test binary and existing test source file unless you have a compelling reason to create a new one. Such a reason would be a necessarily large amount of new testing.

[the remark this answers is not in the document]
pdillinger: If each header file had its own test binary, we would have 500+ test binaries, which would build 100+GB of test binaries. We could potentially switch to LIB_MODE=shared by default, but that's only a partial mitigation.

cache/cache_rep.cc
hx235: Resolved: Slice takes const char* as argument.

Makefile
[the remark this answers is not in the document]
hx235: [non-blocking] [NEED_ATTENTION]
@ajkr - I agree with what @pdillinger mentioned above and tried to look for an existing test file, where I can place my unit test for the new class CacheRep but not awkwardly.

cache_test.cc and lru_cache_test.cc don't seem to be good ones, as the unit tests in those files have coherent themes (i.e, testing the cache or testing the lru_cache). It seems awkward to put my cache_rep_test into any of those files.

Any other test files I should look into?

cache/cache_rep.h
hx235: Resolved: have some thoughts

cache/cache_rep.h
```diff
@@ -0,0 +1,45 @@
...
+#include <mutex>
+#include <vector>
+
+#include "rocksdb/cache.h"
+#include "rocksdb/slice.h"
+#include "table/block_based/block_based_table_reader.h"
+#include "util/coding.h"
+
+namespace ROCKSDB_NAMESPACE {
+class CacheRep {
```
hx235, line 25: [non-blocking][NEED_ATTENTION]
Question 1: what is the role of such a CacheRep class for class users other than WriteBufferManager(e.g, other dummy entries clients like BlockBasedTableBuilder)?

This question needs attention since (1) there might be some follow-up work on top of this working solution, depending on the answer (e.g, a trivial one would be renaming the class) (2)I need to write a class-level description for my new file.

I don't have a very clear answer to the question above due to that:
(1)The existing design of CacheRep did more than just a simple C-style struct (an aggregate class) since we have GetNextCacheKey() and my improved design naturally encapsulated some of its impl details by using private data members.
(2)The current CacheRep does less than fully encapsulating the impl details of dummy entry manipulation (e.g, inserting/releasing/boo-keeping) since dummy entry client (e.g, WriteBufferManager) still needs to interact with the underlying "cache_" or "cache_allocated_size_" for dummy entry manipulation. 

It is also a bit hard for such encapsulation since different dummy entry clients have slightly different needs for such a CacheRep. For example, BlockBasedTableBuilder might not need any bookkeeping on "cache_allocated_size_" at all. This also reflects the issue of not all the public data members in the interface are useful for all class users.

Therefore (1) and (2) lead to some sort of description of CacheRep being "an aggregate class for dummy entry manipulation but with encapsulation on dummy entry cache key generation and with part of the public interface not being applicable to some class users" .

I found the description above for CacheRep class a bit awkward - but may be just me feeling this way...

cache/cache_rep.h
ajkr: Sorry for unresolving/resolving; GitHub has a confusing UI - in some places text box for reply is there, in other places it's mysteriously missing. So I clicked around randomly to try to get it back.

> Do we want this CacheRep to be (1)just a C-style struct that gathers different values without any encapsulation (i.e, everything being public to the CacheRep user; CacheRep gets deeply coupled into the user logic) OR (2)a class that provides encapsulation for dummy entry insertion/release/lock/bookkeeping/key generation stuff?

I agree (2) sounds better. My prediction is other clients use it in the same way so logic like `ReserveMemWithCache()` would be duplicated. Although there are differences like you pointed out, where `WriteBufferManager` does that fancy technique to reduce future insertions by delaying current removals, while that isn't needed for our simple dictionary buffer that does all the insertions followed by all the removal followed by nothing. I expect it does not hurt to have that logic present though. If a future client doesn't like it we can add a flag to disable it.

> (a)Rename general terms e.g, CacheRep to specific terms e.g, CacheRepForUsingDummyEntry
-or maybe we need to do this even in this current implementation.

Haven't read the code yet, but what about something like `CacheReservationManager`? I never knew what Rep meant here anyways.

> -InsertDummyEntry(..)/ReleaseDummyEntry(..) need more thinking to make sure something like these can be helpful for both WBM (gradual insertion/release of some fixed-sized dummy entries) and BlockBasedTableBuilder (insertion of variable-sized dummy entries/all-at-once release of dummy entries accumulated so far) and future dummy entry clients (e.g, Bloom/Ribbon filter building buffer).
-maybe an extra ReleaseAllDummyEntry() for BlockBasedTableBuilder.

Sounds good, seems we are aligned. Or add a flag to `FreeMemWithCache()`  (like `Cache::Release()`'s force flag) that forces immediate deletion, and invoke it with the full dictionary buffer size.

> -this needs some thinking into how CacheRep user interacts with CacheRep for dummy entry stuff. If I understand correctly, one WBM instance owns one CacheRep instance till the end of that WBM instance and that CacheRep instance is not owned by other WBM instances at the same time (through unique_ptr in WBM). Same for BlockBasedTableBuilder. Then when a CacheRep instance gets destructed, it is indicated that the CacheRep user (e.g, WBM or BlockBasedTableBuilder) no longer needs to refer to the remaining unreleased dummy entries (e.g, when WBM or BlockBasedTableBuilder themselves get destructed). Therefore the CacheRep can signal to release those remaining dummy entries.

I think it's correct. I wonder whether it's a valid case that there are remaining dummy entries at destruction time or if that cleanup is just defensive programming. Anyways it's fine to keep.

memtable/write_buffer_manager.cc
ajkr: > above in here #2350 (comment). But I still can't get any answer from the discussion there.

That tends to be the case with discussions involving me. Trying again. I think functions like ReserveMemWithCache cannot run in parallel because of updates to unprotected structure like vector. There is also some complexity with making sure updates across variables are atomic. It could probably be done with more fine-grained locking if needed.

The linked comment I suggested an (important) edge case might work already with the locking removed, not the general case.

Unrelated: does FreeMemWithCache release at most one dummy entry? I feel there should be a loop.

cache/cache_rep.cc
ajkr: The client's cache key still becomes invalid after another client calls GetNextCacheKey() though. Seems fine, probably worth documenting.

cache/cache_rep.cc
ajkr: `[cache_key_, cache_key_ + kCacheKeyPrefixSize)` doesn't change, and `EncodeVarint64()` overwrites `[cache_key_ + kCacheKeyPrefixSize, end)`. Then the `Slice` covers `[cache_key_, end)` which feels safe, though the bytes beyond `end` could be nonzero and stale.

ajkr (commented):
Went through discussion comments. Haven't gone through code yet.

memtable/write_buffer_manager.cc
ajkr: > Unrelated: does FreeMemWithCache release at most one dummy entry? I feel there should be a loop.

Maybe that's why we have to free dummy entries in the destructor, because we're skipping most of them when they're actually freed.

Also this supports your design (2) idea so we do not have to duplicate these logic as it's evidently error-prone.

hx235:
> Went through discussion comments. Haven't gone through code yet.

@ajkr I went through your comment and anticipate another wave of new changes on CacheRep and WBM - you can go through code later with all these new changes in place. 

Below is my todo that I will follow up:
(a) Turn CacheRep into CacheReservationManager based on the discussion above
(b) Re-read and understand your comment on the concurrency/destructor issue related to ReserveMemWithCache (sorry - couldn't quite get it 100% at my first glance)
(c) Document the pitfall where GetNextCacheKey has the side effect of changing the underlying cache_key of other CacheKey
(d) Reach out to Peter or you to understand the static linking/test binary size issue

Makefile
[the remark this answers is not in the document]
pdillinger: > I thought static linking would drop unreferenced symbols

Nope. Need `strip` for that. 

> Any other test files I should look into?

Maybe db_block_cache_test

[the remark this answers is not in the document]
ajkr: I don't know, it's 18MB for me. Maybe it's too easy to accidentally pull in everything because even simple tests like event_logger_test are 200MB+.

```
$ git apply 8506.diff && make -j48 cache_rep_test && ls -l cache_rep_test
...
-rwxr-xr-x 1 andrewkr users 18320624 Jul 15 15:43 cache_rep_test
```

Anyways move to separate file sounds fine. But tests named "db_*" are supposed to be higher-level tests using the `DB` API. How about "cache/cache_test.cc"?

hx235:
The failed checks indicate some affected test cases related to WriteBufferManager in other unit tests - need to investigate later.

cache/cache_reservation_manager.cc
```diff
@@ -0,0 +1,148 @@
...
+Status CacheReservationManager::UpdateCacheReservation(
+    std::size_t new_mem_used) {
+  std::size_t cur_cache_allocated_size =
+      cache_allocated_size_.load(std::memory_order_relaxed);
+  if (new_mem_used == cur_cache_allocated_size ||
+      (cur_cache_allocated_size > 0 &&
+       new_mem_used > cur_cache_allocated_size &&
+       (new_mem_used - cur_cache_allocated_size) < kSizeDummyEntry) ||
+      (new_mem_used < cur_cache_allocated_size &&
+       (cur_cache_allocated_size - new_mem_used) < kSizeDummyEntry)) {
```
hx235, line 51: [Note]
While change in memory release behavior is expected since we are revising the memory releasing algorithm, I need to point out that the memory reserving behavior is also changed a bit. 

The previous memory reserving behavior can be described as the following, where "+(new_mem_used)" stands for what happened in [here](https://github.com/facebook/rocksdb/pull/8506/files#diff-72f8c69a5d31088bb7cbda00065f74f10f931a209fe3a5b342326f115738d9c4L107-L125), that is, looping until "new_mem_used > cache_allocated_size_" is no longer true.

Assume we starts with new_mem_used = cache_allocated_size = 0

op no.| new_mem_used|    operation   |  cache_allocated_size after operation
1.         |         128KB         |   +(128KB)   |                  256KB
2.        |          129KB         |   +(129KB)   |                  256KB  (since new_mem_used < cache_allocated_size)
3.        |          257KB         |   +(257KB)  |                   512KB


However, the under the new code, the reserving behavior will become:

Assume we starts with new_mem_used = cache_allocated_size = 0

op no.| new_mem_used|    operation             |  cache_allocated_size after operation
1.         |         128KB         |   update(128KB)   |                  256KB
2.        |          129KB         |   update(129KB)   |                  256KB  (since new_mem_used < cache_allocated_size && (cur_cache_allocated_size - new_mem_used) < kSizeDummyEntry)
3.        |          257KB         |   update(257KB)  |                   256KB (since new_mem_used > cur_cache_allocated_size && (new_mem_used - cur_cache_allocated_size) < kSizeDummyEntry)
       
The reason for such change is that, when we consolidate reserve/free memory reservation into one update memory reservation API, we currently rely on the relative size order of new_mem_used and cache_allocated_size to decide whether to do an increase or a decrease. If we use a strict comparison such as "if new_mem_used > cache_allocated_size, then increase; if new_mem_used < cache_allocated_size, then decrease", we end up doing a decrease for op 2 above in our new behavior. Therefore, I decided to allow a "fuzzy equal condition", that is, abs(new_mem_used - cache_allocated_size) < kSizeDummyEntry can be seen as new_mem_used = cache_allocated_size and we do not need to update any cache reservation under such case.

One exception to the above "fuzzy equal condition" in my own impl is the "cur_cache_allocated_size > 0" in [here](https://github.com/facebook/rocksdb/pull/8506/files#diff-89e4f3dd4e967498e08e77b49f8d290d242f9f1b1fb30f1434480711545917b7R47). The reason for this exception is to adapt to existing test cases related to WBM that assumes trivial memory increase less than 1 dummy entry size (e.g, increase from 0 to 128 KB) always lead to 1 dummy entry insertion. Such test cases includes "TestWriteBufferNoLimitWithCache - DBTest2", "SharedWriteBufferAcrossCFs/2 - DBTestSharedWriteBufferAcrossCFs/DBTestSharedWriteBufferAcrossCFs", "CacheEntryRoleStats - DBBlockCacheTest".

I think such a behavior change in reserving should be fine as CacheRep/WBM:: ReserveMemWithCache/ WriteBufferManager:: FreeMemWithCache are all private implementation details. And the implementation details described in the public document https://github.com/facebook/rocksdb/wiki/Write-Buffer-Manager#implementation is still compatible will the new changes (However, we might need to change the release part). I hope there isn't any users using the public API  WriteBufferManager::dummy_entries_in_cache_usage() to infer back into how the previous cache reservation algorithms work and rely on the previous one.

Do let me know  though if you see any risk in such changes in reservation behavior to our external users that I am not aware of. I can try to investigate (again) to see how to maintain the previous reservation behavior as much as possible while trying to consolidate reserve/free mem into update...

cache/cache_reservation_manager.cc
```diff
@@ -0,0 +1,148 @@
...
+
+Status CacheReservationManager::DecreaseCacheReservation(
+    std::size_t new_mem_used) {
+  Status return_status = Status::OK();
+  while (new_mem_used < cache_allocated_size_.load(std::memory_order_relaxed)) {
+    assert(!dummy_handles_.empty());
+    auto* handle = dummy_handles_.back();
+    // If insert failed, handle is null so we should not release.
+    if (handle != nullptr) {
+      cache_->Release(handle, true);
```
hx235, line 126: [Note]
I couldn't find an easy way to see whether an error occurred in Release() so it always returns Status::OK() now. 

I noticed that in the document [here](https://github.com/facebook/rocksdb/blob/master/include/rocksdb/cache.h#L413-L416) , it indicated that we should call "deleter"explicitly when there is an error in releasing. However, I honestly don't know how to know whether there is an error ....

memtable/write_buffer_manager.cc
```diff
@@ -10,81 +10,46 @@
...
-#ifndef ROCKSDB_LITE
-  if (cache_rep_) {
-    for (auto* handle : cache_rep_->dummy_handles_) {
-      if (handle != nullptr) {
-        cache_rep_->cache_->Release(handle, true);
-      }
-    }
+WriteBufferManager::~WriteBufferManager() {}
+
+std::size_t WriteBufferManager::dummy_entries_in_cache_usage() const {
```
hx235, line 43: [Note]
I replaced the private dummy_size_ in WBM with cache_allocated_size_ since they look exactly the same to me and this helps decouple the cache reservation logic out of WBM.

cache/cache_reservation_manager_test.cc
```diff
@@ -0,0 +1,345 @@
...
+  lo.num_shard_bits = 0;  // 2^0 shard
+  lo.strict_capacity_limit = true;
+  std::shared_ptr<Cache> cache = NewLRUCache(lo);
+  std::unique_ptr<CacheReservationManager> test_cache_rev_mng(
+      new CacheReservationManager(cache));
+
+  std::size_t new_mem_used = kOneGigabyte + 1;
+  Status s =
+      test_cache_rev_mng
+          ->UpdateCacheReservation<ROCKSDB_NAMESPACE::CacheEntryRole::kMisc>(
```
hx235, line 182: [Note]
For CacheEntryRole::kMisc, it mentioned "Do not use for entries that could potentially add up to large usage" in [here](https://github.com/facebook/rocksdb/blob/master/cache/cache_entry_roles.h#L32-L33). I assume (1GB + 1) can be seen as not-that-large usage considering it is only get called once in this test case and will be soon released after the test is torn down through CacheReservationManager's destructor.

cache/cache_reservation_manager.h
```diff
@@ -0,0 +1,81 @@
...
+#include "cache/cache_entry_roles.h"
+#include "rocksdb/cache.h"
+#include "rocksdb/slice.h"
+#include "rocksdb/status.h"
+#include "table/block_based/block_based_table_reader.h"
+#include "util/coding.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+class CacheReservationManager {
```
ajkr, line 30: Do we need to say this class's functions (besides `GetTotalReservedCacheSize()`) are not thread-safe so require external synchronization?

cache/cache_reservation_manager_test.cc
ajkr: I think that's intended to improve memory attribution. Memory attribution would be difficult if large uses were categorized as "miscellaneous". For tests that wouldn't matter.

memtable/write_buffer_manager.cc
ajkr: > dummy_entries_in_cache_usage() called from thread A missing any changes to the same dummy_size_ from thread B

Should be fine for our purpose (exposing a metric).

ajkr (changes requested):
Great progress on generalizing how we charge block cache for outside memory usage! The main thing I wonder about is if we can simplify how we handle reserving one more or one less dummy entry when the target memory falls in the middle.

cache/cache_reservation_manager.h
hx235: Yes - and it seems like the current way of saying "This class is not thread-safe." in https://github.com/facebook/rocksdb/pull/8506/files#diff-b95042a5d09f06814f920b8c334a4f0aff0574c41981d1fc82a93d1c87e774b5R12 does not grasp as much attention as I expected.

Should I try marking non-thread safe/external synchronization needed in front of every function that needs it?

cache/cache_reservation_manager_test.cc
```diff
@@ -0,0 +1,345 @@
...
+  Slice expected_cache_key_slice(
+      expected_cache_key, static_cast<std::size_t>(end - expected_cache_key));
+
+  std::size_t new_mem_used = 1 * kSizeDummyEntry;
+  Status s =
+      test_cache_rev_mng
+          ->UpdateCacheReservation<ROCKSDB_NAMESPACE::CacheEntryRole::kMisc>(
+              new_mem_used);
+  ASSERT_EQ(s, Status::OK());
+  ASSERT_GE(cache->GetPinnedUsage(), 1 * kSizeDummyEntry);
```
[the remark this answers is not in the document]
hx235: Yes - this is what I missed. Will try kMetaDataChargeOverhead

cache/cache_reservation_manager.cc
hx235: > I think it would be simpler to update the reservation bytes to the new_mem_used rounded up to the nearest multiple of kSizeDummyEntry.
Just to make sure I understand, are you suggesting something like the following:
```
rounded_new_mem_used = round_to_nearest(new_mem_used)

if (rounded_new_mem_used == cache_allocated_size_) {
    do nothing 
} else if (rounded_new_mem_used > cache_allocated_size_) {
    increase_to(rounded_new_mem_used) 
} else {       
    decrease_to(rounded_new_mem_used) 
}       

increase_to(rounded_new_mem_used){
      while (cache_allocated_size_ < rounded_new_mem_used && rounded_new_mem_used - cache_allocated_size_ >  kSizeDummyEntry ) {
           insert dummy entry
       }
}
decrease_to(rounded_new_mem_used){
       while (cache_allocated_size_ > rounded_new_mem_used && cache_allocated_size_ - rounded_new_mem_used > kSizeDummyEntry) {
           release dummy entry
       }
}
```

>The main thing I wonder about is if we can simplify how we handle reserving one more or one less dummy entry when the target memory falls in the middle
If I understand your "rounded up to the nearest multiple of kSizeDummyEntry" correctly, the target memory falling in the middle can be rounded up just as 5 rounded up to 10. And we go from there.

cache/cache_reservation_manager.h
hx235: > Great progress on generalizing how we charge block cache for outside memory usage! The main thing I wonder about is if we can simplify how we handle reserving one more or one less dummy entry when the target memory falls in the middle.

Thanks! Replied to your thought in https://github.com/facebook/rocksdb/pull/8506/files#r692622024 and will reach out offline tmr to clear up some of my confusion. 

And I will also follow up on the to-do suggestions on the test in [here](https://github.com/facebook/rocksdb/pull/8506/files#r692473085) and function comment in [here](https://github.com/facebook/rocksdb/pull/8506/files#r692600511)

ajkr: Oh sorry... I might've overlooked it because it's grouped with the  copyright notices. That is fine already. Maybe above the class definition is slightly more conventional; I think you don't need to put it above every function.

cache/cache_reservation_manager.cc
```diff
@@ -0,0 +1,146 @@
...
+    // This is to prevent possible subsequent successful dummy entry insertion
+    // (due to newly available cache space resulting from manipulation in other
+    // threads) from overriding the non-OK status
+    if (return_status == Status::OK()) {
+      return_status = s;
+    }
+
+    // We absorb the error here since s is conditionally checked
+    // when return_status == Status::OK()
+    s.PermitUncheckedError();
```
hx235, line 108: [Note]
I feel that L98 - 101 describes a valid concern and I haven't thought of another way to handle that. One drawback to this approach of preventing status from being overwritten is the need of s.PermitUncheckedError();. Let me know if L98 - 101 is not a valid concern and I can return s directly without the need for s.PermitUncheckedError();.

Makefile
```diff
@@ -1897,7 +1897,9 @@ clipping_iterator_test: $(OBJ_DIR)/db/compaction/clipping_iterator_test.o $(TEST
 
 ribbon_bench: $(OBJ_DIR)/microbench/ribbon_bench.o $(LIBRARY)
 	$(AM_LINK)
-
+	
+cache_reservation_manager_test: $(OBJ_DIR)/cache/cache_reservation_manager_test.o $(TEST_LIBRARY) $(LIBRARY)
```
hx235, line 1901: @pdillinger I know we have talked about whether to keep a separate testing binary in https://github.com/facebook/rocksdb/pull/8506#discussion_r668095996. For this new test, we have 9 test cases (which is more than the previous test with only 1). Does it make it worth itself a separate testing binary?

cache/cache_reservation_manager.cc
ajkr: Sorry for the confusion. I meant to always round up when there is any fraction, like a ceiling function. So 257 rounds up to 512.

cache/cache_reservation_manager.cc
```diff
@@ -0,0 +1,146 @@
...
+#include "util/coding.h"
+
+namespace ROCKSDB_NAMESPACE {
+CacheReservationManager::CacheReservationManager(std::shared_ptr<Cache> cache,
+                                                 bool delayed_decrease)
+    : delayed_decrease_(delayed_decrease), cache_allocated_size_(0) {
+  assert(cache != nullptr);
+  cache_ = cache;
+  std::memset(cache_key_, 0, kCacheKeyPrefixSize + kMaxVarint64Length);
+  EncodeVarint64(cache_key_, cache_->NewId());
```
hx235, line 30: @pdillinger quoted you here just to double check if any of your changes in unifying all the code generating cache keys mentioned in https://github.com/facebook/rocksdb/pull/8506#discussion_r667265255 needed to be integrated here :)

cache/cache_reservation_manager.h
hx235: Sure - I can move the description and non thread-safe right above the class definition.

hx235: > will reach out offline tmr to clear up some of my confusion.

Discussed offline. We now both agree on the approach of updating cache reservation to the ceil(new_mem_used) with respect to the multiples of dummy entry sizes. It eliminates the unnecessary increase/decrease I described in  https://github.com/facebook/rocksdb/pull/8506#discussion_r692511783 and makes it easier to resonate the amount of dummy entries in the cache after update(new_mem_used) since we can now always expect for ceil(new_mem_used).

hx235:
Update:
1. Made the requested changes to simplify the algorithm as discussed in https://github.com/facebook/rocksdb/pull/8506#discussion_r693291963
2. Added upper bound for cache usage to test for over-charging as suggested in https://github.com/facebook/rocksdb/pull/8506#discussion_r692473085
3. Clarifies class-level/function-level comment for CacheReservationManager as discussed in https://github.com/facebook/rocksdb/pull/8506#discussion_r692628095
4. Reached out to Peter to follow up discussion related to this PR on cache key ([here](https://github.com/facebook/rocksdb/pull/8506#discussion_r692627318)) and test binary ([here]. Will reach out offline next Monday
(https://github.com/facebook/rocksdb/pull/8506#discussion_r692626519))
5. Left note ([here](https://github.com/facebook/rocksdb/pull/8506/files#r692625312)) about the current error handling in acheReservationManager::IncreaseCacheReservation

cache/cache_reservation_manager.cc
ajkr: I think return it directly is good. The old best-effort behavior of adding nullptr handles for failed insertions seems not generally useful. It seems questionable whether it's even useful for `WriteBufferManager` today -- and definitely not useful once we make `WriteBufferManager` able to handle reservation failures.

It would be a behavior change for `WriteBufferManager` in today's form, though. Failure to insert enough dummy entries in one `UpdateCacheReservation()` would start causing extra insertions (or fewer releases) in the next `UpdateCacheReservation()`. I think that's fine since we want to continue targeting an accurate reservation even after some failure.

The downside of this change is an individual `UpdateCacheReservation()` might reserve less because we fail early rather than continue trying more cache keys. It could be mitigated by retrying a few times in `WriteBufferManager` when a failure is returned. But I doubt it's needed.

ajkr (approved):
LGTM!

hx235:
Update:
- Made CacheReservationManager::IncreaseCacheReservation() returns early from dummy entry insertion error
- Revised unit tests of CacheReservationManager and WriteBufferManager to adapt to the following changes 
- Corrected a previous typo in WriteBufferManager
- Reached out to Peter for cache key, test binary questions

Will merge the PR after signals are cleared

hx235:
Update:
Removed const keyword in move constructor/assignment parameter

PR #8510, opened 2021-07-09
The formal parameter types of CompressionOptions constructor should b…

zaorangyang:
…e consistent with the member variables's


mrambacher (approved):
LGTM.

PR #8518, opened 2021-07-13
Avoid updating DBOptions if there's no value updated

jay-zhuang:
Summary: Try to avoid expensive updating DBOptions if
`SetDBOptions()` does not change any option value.

Test Plan: added unittest

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -1134,8 +1134,16 @@ Status DBImpl::SetDBOptions(
   WriteContext write_context;
   {
     InstrumentedMutexLock l(&mutex_);
+    bool is_changed;
     s = GetMutableDBOptionsFromStrings(mutable_db_options_, options_map,
-                                       &new_options);
+                                       &new_options, &is_changed);
+    if (!is_changed) {
+      ROCKS_LOG_INFO(immutable_db_options_.info_log,
+                     "SetDBOptions(), input option value is not changed, "
+                     "skipping updating.");
+      persist_options_status.PermitUncheckedError();
+      return s;
+    }
     if (new_options.bytes_per_sync == 0) {
       new_options.bytes_per_sync = 1024 * 1024;
     }
@@ -4194,6 +4202,8 @@ Status DBImpl::WriteOptionsFile(bool need_mutex_lock,
 
   TEST_SYNC_POINT("DBImpl::WriteOptionsFile:1");
   TEST_SYNC_POINT("DBImpl::WriteOptionsFile:2");
+  TEST_SYNC_POINT_CALLBACK("DBImpl::WriteOptionsFile:PersistOptions",
+                           &db_options);
 
   std::string file_name =
       TempOptionsFileName(GetName(), versions_->NewFileNumber());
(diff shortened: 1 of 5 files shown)


options/db_options.cc
```diff
@@ -869,6 +869,13 @@ Status GetMutableDBOptionsFromStrings(
   if (!s.ok()) {
     *new_options = base_options;
   }
+  // check if new_options is changed by comparing to base_options
+  if (is_changed) {
+    std::string mismatch;
+    *is_changed = !OptionTypeInfo::StructsAreEqual(
+        config_options, "MutableDBOptions", &db_mutable_options_type_info,
+        "MutableDBOptions", &base_options, new_options, &mismatch);
+  }
```
mrambacher, line 878: I think this would be better as a new API (AreMutableDBOptionsEqual ?).  This way, the DB code could do whatever sanitization is necessary before the comparison takes place.  Effectively the same code, just in a new routine.

db/db_options_test.cc
```diff
@@ -98,6 +98,62 @@ TEST_F(DBOptionsTest, ImmutableTrackAndVerifyWalsInManifest) {
...
+  ASSERT_FALSE(is_changed());
+
+  // changing the value
+  ASSERT_OK(dbfull()->SetDBOptions({{"bytes_per_sync", "123"}}));
+  ASSERT_TRUE(is_changed());
+
+  // update again
+  ASSERT_OK(dbfull()->SetDBOptions({{"bytes_per_sync", "123"}}));
+  ASSERT_FALSE(is_changed());
+
```
mrambacher, line 137: What about some tests that set bytes_per_sync=0?

db/db_impl/db_impl.cc
```diff
@@ -1119,8 +1119,16 @@ Status DBImpl::SetDBOptions(
   WriteContext write_context;
   {
     InstrumentedMutexLock l(&mutex_);
+    bool is_changed;
     s = GetMutableDBOptionsFromStrings(mutable_db_options_, options_map,
-                                       &new_options);
+                                       &new_options, &is_changed);
+    if (!is_changed) {
```
mrambacher, line 1125: See the note below about making this a new routine.  Would allow for the sanitization in the following lines to take place before the comparison.

mrambacher (commented):
Perhaps it belongs in another PR, but should a similar effort be done for DB::SetOptions?  If so (and it is a separate PR), please at least update the title of this PR to indicate it applies to DBOptions and not CFOptions.

options/db_options.cc
jay-zhuang: make sense.

db/db_options_test.cc
jay-zhuang: Added. With the checking after sanitization, it can avoid updating the value for setting it to 0. But the default value is not sanitized, but `SetDBOptions()` does. Not sure why, seems it's an inconsistency between DB open and `SetDBOptions()`.

riversand963 (approved):

db/db_impl/db_impl.cc
jay-zhuang: removed.

PR #8520, opened 2021-07-13
Fix stats_history_test failure on Windows

sherriiiliu:
Fixed a stats_history_test failure on Windows
* In StatsHistoryTest.InMemoryStatsHistoryPurging test, the capping memory cost of stats_history_size on Windows increases to 15000 bytes with latest changes

### DIFF
--- a/monitoring/stats_history_test.cc
+++ b/monitoring/stats_history_test.cc
@@ -245,10 +245,10 @@ TEST_F(StatsHistoryTest, InMemoryStatsHistoryPurging) {
   }
   size_t stats_history_size = dbfull()->TEST_EstimateInMemoryStatsHistorySize();
   ASSERT_GE(slice_count, kIterations - 1);
-  ASSERT_GE(stats_history_size, 14000);
-  // capping memory cost at 14000 bytes since one slice is around 10000~14000
-  ASSERT_OK(dbfull()->SetDBOptions({{"stats_history_buffer_size", "14000"}}));
-  ASSERT_EQ(14000, dbfull()->GetDBOptions().stats_history_buffer_size);
+  ASSERT_GE(stats_history_size, 15000);
+  // capping memory cost at 15000 bytes since one slice is around 10000~15000
+  ASSERT_OK(dbfull()->SetDBOptions({{"stats_history_buffer_size", "15000"}}));
+  ASSERT_EQ(15000, dbfull()->GetDBOptions().stats_history_buffer_size);
 
   // Wait for stats persist to finish
   for (int i = 0; i < kIterations; ++i) {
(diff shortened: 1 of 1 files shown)


monitoring/stats_history_test.cc
```diff
@@ -245,10 +245,10 @@ TEST_F(StatsHistoryTest, InMemoryStatsHistoryPurging) {
   }
   size_t stats_history_size = dbfull()->TEST_EstimateInMemoryStatsHistorySize();
   ASSERT_GE(slice_count, kIterations - 1);
-  ASSERT_GE(stats_history_size, 14000);
-  // capping memory cost at 14000 bytes since one slice is around 10000~14000
-  ASSERT_OK(dbfull()->SetDBOptions({{"stats_history_buffer_size", "14000"}}));
-  ASSERT_EQ(14000, dbfull()->GetDBOptions().stats_history_buffer_size);
+  ASSERT_GE(stats_history_size, 15000);
```
mrambacher, line 248: Should the 15000 be a constant?  And should it be 15000U?

mrambacher (commented):
Do you know when this test started failing (PR or version)?  It would be nice to know that the change to make the test pass is correct, rather than the code itself doing something screwy.

monitoring/stats_history_test.cc
sherriiiliu: > Do you know when this test started failing (PR or version)? It would be nice to know that the change to make the test pass is correct, rather than the code itself doing something screwy.

Last time I submitted a PR #7992 to increase the cap to 14000 and this time I am running the test with v6.20.3(8608d75). The one slice memory cost comes from this estimation function:
```
// esitmate the total size of stats_history_
size_t DBImpl::EstimateInMemoryStatsHistorySize() const {
  size_t size_total =
      sizeof(std::map<uint64_t, std::map<std::string, uint64_t>>);
  if (stats_history_.size() == 0) return size_total;
  size_t size_per_slice =
      sizeof(uint64_t) + sizeof(std::map<std::string, uint64_t>);
  // non-empty map, stats_history_.begin() guaranteed to exist
  std::map<std::string, uint64_t> sample_slice(stats_history_.begin()->second);
  for (const auto& pairs : sample_slice) {
    size_per_slice +=
        pairs.first.capacity() + sizeof(pairs.first) + sizeof(pairs.second);
  }
  size_total = size_per_slice * stats_history_.size();
  return size_total;
}
```
I think the difference might come from `sample_slice` since it includes more entries. With MSVC 14.0, the `size_per_slice` calculated is 14331, so I increased the cap to 15000. Do you have better idea to avoid such failure next time? Like you said, the number is not fixed as it will change together with code change

sherriiiliu: @mrambacher gentle ping

mrambacher (approved):
Seems reasonable to me.

PR #8523, opened 2021-07-13
Minor Makefile update to exclude microbench as dependency

jay-zhuang:
Summary: Otherwise the build may report warning about missing
`benchmark.h` for some targets, the error won't break the build.

Test Plan: `make blackbox_ubsan_crash_test` on a machine without
benchmark lib installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -506,7 +506,7 @@ STRESS_OBJECTS =  $(patsubst %.cc, $(OBJ_DIR)/%.o, $(STRESS_LIB_SOURCES))
 # Exclude build_version.cc -- a generated source file -- from all sources.  Not needed for dependencies
 ALL_SOURCES  = $(filter-out util/build_version.cc, $(LIB_SOURCES)) $(TEST_LIB_SOURCES) $(MOCK_LIB_SOURCES) $(GTEST_DIR)/gtest/gtest-all.cc
 ALL_SOURCES += $(TOOL_LIB_SOURCES) $(BENCH_LIB_SOURCES) $(CACHE_BENCH_LIB_SOURCES) $(ANALYZER_LIB_SOURCES) $(STRESS_LIB_SOURCES)
-ALL_SOURCES += $(TEST_MAIN_SOURCES) $(TOOL_MAIN_SOURCES) $(BENCH_MAIN_SOURCES) $(MICROBENCH_SOURCES)
+ALL_SOURCES += $(TEST_MAIN_SOURCES) $(TOOL_MAIN_SOURCES) $(BENCH_MAIN_SOURCES)
 
 TESTS = $(patsubst %.cc, %, $(notdir $(TEST_MAIN_SOURCES)))
 TESTS += $(patsubst %.c, %, $(notdir $(TEST_MAIN_SOURCES_C)))


mrambacher (commented):
This will work to prevent the error, but is not ideal.  This would mean that the microbenchmarks will not be rebuilt if a header file changes for example.  

I also am not sure if a similar problem exists for cmake, and if it does, this would not fix that issue.


Why not introduce a ROCKSDB_BENCHMARKS define (similar to what is done with GFLAGS) to test if the benchmarks are present?  Similar code also exists for all of the compression library tests.

jay-zhuang:
> This will work to prevent the error, but is not ideal. This would mean that the microbenchmarks will not be rebuilt if a header file changes for example.

microbenchmark has it's own target `$ make microbench`, which rebuilds if any header file it depends on changed. I think this change just exclude the microbenchmark tests from the default targets.

> I also am not sure if a similar problem exists for cmake, and if it does, this would not fix that issue.

cmake is similar, it has an option `WITH_BENCHMARK` to enable microbenchmark build. It has the same customized target `$ make microbench` to build and run microbench.
 
> Why not introduce a ROCKSDB_BENCHMARKS define (similar to what is done with GFLAGS) to test if the benchmarks are present? Similar code also exists for all of the compression library tests.

I guess some tests still make sense without GFLAGS, but all microbench tests are going to need `benchmark.h`.

mrambacher:
> > Why not introduce a ROCKSDB_BENCHMARKS define (similar to what is done with GFLAGS) to test if the benchmarks are present? Similar code also exists for all of the compression library tests.
> 
> I guess some tests still make sense without GFLAGS, but all microbench tests are going to need `benchmark.h`.

Things that require GFLAGS but are not compiled (tests, db_bench, etc) with it give an error/message about needing to use GFLAGS.  Something similar could be done here.

jay-zhuang:
> > > Why not introduce a ROCKSDB_BENCHMARKS define (similar to what is done with GFLAGS) to test if the benchmarks are present? Similar code also exists for all of the compression library tests.
> > 
> > 
> > I guess some tests still make sense without GFLAGS, but all microbench tests are going to need `benchmark.h`.
> 
> Things that require GFLAGS but are not compiled (tests, db_bench, etc) with it give an error/message about needing to use GFLAGS. Something similar could be done here.

sorry, I still don't see the benefit of introducing a ROCKSDB_BENCHMARKS which will be added to all microbench tests under `microbench/` folder. Plus cmake is also conditionally build that base on if `with_benchmark` is enabled or not.

mrambacher:
> sorry, I still don't see the benefit of introducing a ROCKSDB_BENCHMARKS which will be added to all microbench tests under `microbench/` folder. Plus cmake is also conditionally build that base on if `with_benchmark` is enabled or not.

1.  Without including the files in ALL_SOURCES, the tools will not get rebuilt properly if header files or other dependencies change;
2. Without being in ALL_SOURCES, the benchmarks become different than every other source.  Why should this be special and an exception?
3.  The build failing if the tool is not installed is also different than the other optional tools, which will build successfully but fail to run (prints an error when you execute it).

jay-zhuang:
> > sorry, I still don't see the benefit of introducing a ROCKSDB_BENCHMARKS which will be added to all microbench tests under `microbench/` folder. Plus cmake is also conditionally build that base on if `with_benchmark` is enabled or not.
> 
> 1. Without including the files in ALL_SOURCES, the tools will not get rebuilt properly if header files or other dependencies change;
> 2. Without being in ALL_SOURCES, the benchmarks become different than every other source.  Why should this be special and an exception?
> 3. The build failing if the tool is not installed is also different than the other optional tools, which will build successfully but fail to run (prints an error when you execute it).

To build microbench, you have to explicitly do `$ make microbench` or `$ make ribbon_bench`, with or without this change. And it will rebuild everything needed for the microbench tests. Which is also matching the same behavior of cmake.
It's also by design when I talk to a few other team members that it should be separated to avoid any impact to existing developers, just like our fuzz tests.

jay-zhuang:
> > > sorry, I still don't see the benefit of introducing a ROCKSDB_BENCHMARKS which will be added to all microbench tests under `microbench/` folder. Plus cmake is also conditionally build that base on if `with_benchmark` is enabled or not.
> > 
> > 
> > 
> > 1. Without including the files in ALL_SOURCES, the tools will not get rebuilt properly if header files or other dependencies change;
> > 2. Without being in ALL_SOURCES, the benchmarks become different than every other source.  Why should this be special and an exception?
> > 3. The build failing if the tool is not installed is also different than the other optional tools, which will build successfully but fail to run (prints an error when you execute it).
> 
> To build microbench, you have to explicitly do `$ make microbench` or `$ make ribbon_bench`, with or without this change. And it will rebuild everything needed for the microbench tests. Which is also matching the same behavior of cmake.
> It's also by design when I talk to a few other team members that it should be separated to avoid any impact to existing developers, just like our fuzz tests.

If you're asking to build (rebuild) microbench tests by default (with `$ make`), I think it's a separate discussion which requires other changes. And for a new experimental test tool like this, just like fuzz test, we would like to have it disabled by default (the same for cmake).

pdillinger (approved):
Seems fine, except I would add this in order to solve the issue Mark points out with code not recompiling on modification to dependency:

    -DEPFILES = $(patsubst %.cc, $(OBJ_DIR)/%.cc.d, $(ALL_SOURCES))
    +DEPFILES = $(patsubst %.cc, $(OBJ_DIR)/%.cc.d, $(ALL_SOURCES) $(MICROBENCH_SOURCES))

jay-zhuang:
> Seems fine, except I would add this in order to solve the issue Mark points out with code not recompiling on modification to dependency:
> 
> ```
> -DEPFILES = $(patsubst %.cc, $(OBJ_DIR)/%.cc.d, $(ALL_SOURCES))
> +DEPFILES = $(patsubst %.cc, $(OBJ_DIR)/%.cc.d, $(ALL_SOURCES) $(MICROBENCH_SOURCES))
> ```

That would cause the same problem that it try to process the MICROBENCH_SOURCES by default.

PR #8528, opened 2021-07-14
Mempurge support for wal

bjlemaire:
In this PR, `mempurge` is made compatible with the Write Ahead Log: in case of recovery, the DB is now capable of recovering the data that was "mempurged" and kept in the `imm()` list of immutable memtables. 
The twist was to add a uint64_t to the `memtable` struct to store the number of the earliest log file containing entries from the `memtable`. When a `Flush` operation is replaced with a `MemPurge`, the `VersionEdit` (which usually contains the new min log file number to pick up for recovery and the level 0 file path of the newly created SST file) is no longer appended to the manifest log, and every time the `deleteWal` method is called, a check is made on the list of immutable memtables. 
This PR also includes a unit test that verifies that no data is lost upon Reopening of the database when the mempurge feature is activated. This extensive unit test includes two column families, with valid data contained in the imm() at time of "crash"/reopening (recovery).

### DIFF
--- a/db/db_flush_test.cc
+++ b/db/db_flush_test.cc
@@ -1106,6 +1106,130 @@ TEST_F(DBFlushTest, MemPurgeAndCompactionFilter) {
   ASSERT_EQ(Get(KEY5), p_v5);
 }
 
+TEST_F(DBFlushTest, MemPurgeWALSupport) {
+  Options options = CurrentOptions();
+
+  options.statistics = CreateDBStatistics();
+  options.statistics->set_stats_level(StatsLevel::kAll);
+  options.create_if_missing = true;
+  options.compression = kNoCompression;
+  options.inplace_update_support = false;
+  options.allow_concurrent_memtable_write = true;
+
+  // Enforce size of a single MemTable to 64MB (64MB = 67108864 bytes).
+  options.write_buffer_size = 1 << 20;
+  // Activate the MemPurge prototype.
+  options.experimental_allow_mempurge = true;
+  ASSERT_OK(TryReopen(options));
+
+  do {
+    CreateAndReopenWithCF({"pikachu"}, options);
+    ASSERT_OK(Put(1, "foo", "v1"));
+    ASSERT_OK(Put(1, "baz", "v5"));
+
+    ReopenWithColumnFamilies({"default", "pikachu"}, options);
+    ASSERT_EQ("v1", Get(1, "foo"));
+
+    ASSERT_EQ("v1", Get(1, "foo"));
+    ASSERT_EQ("v5", Get(1, "baz"));
+    ASSERT_OK(Put(1, "bar", "v2"));
+    ASSERT_OK(Put(1, "foo", "v3"));
+    uint32_t mempurge_count = 0;
+    uint32_t sst_count = 0;
+    ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->SetCallBack(
+        "DBImpl::FlushJob:MemPurgeSuccessful",
+        [&](void* /*arg*/) { mempurge_count++; });
+    ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->SetCallBack(
+        "DBImpl::FlushJob:SSTFileCreated", [&](void* /*arg*/) { sst_count++; });
+    ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->EnableProcessing();
+
+    std::string KEY1 = "IamKey1";
+    std::string KEY2 = "IamKey2";
+    std::string KEY3 = "IamKey3";
+    std::string KEY4 = "IamKey4";
+    std::string KEY5 = "IamKey5";
+    std::string KEY6 = "IamKey6";
+    std::string KEY7 = "IamKey7";
+    std::string KEY8 = "IamKey8";
+    std::string KEY9 = "IamKey9";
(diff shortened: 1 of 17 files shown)


db/memtable_list.cc
```diff
@@ -502,13 +505,24 @@ Status MemTableList::TryInstallMemtableFlushResults(
...
+        *io_s = vset->io_status();
+      } else {
+        // If write_edit is false (e.g: successful mempurge),
+        // then remove old memtables, wake up manifest write queue threads,
+        // and don't commit anything to the manifest file.
+        // Notify new head of manifest write queue.
+        // wake up all the waiting writers
+        RemoveMemTablesOrRestoreFlags(s, cfd, batch_count, log_buffer,
+                                      to_delete, mu);
+        vset->WakeUpWaitingManifestWriters();
```
pdillinger, line 523: If we don't fully understand why this is needed, consider a `TODO: explain full reason needed or investigate more`

db/db_flush_test.cc
```diff
@@ -1106,6 +1106,193 @@ TEST_F(DBFlushTest, MemPurgeAndCompactionFilter) {
...
+
+  options.statistics = CreateDBStatistics();
+  options.statistics->set_stats_level(StatsLevel::kAll);
+  options.create_if_missing = true;
+  options.compression = kNoCompression;
+  options.inplace_update_support = false;
+  options.allow_concurrent_memtable_write = true;
+
+  // Enforce size of a single MemTable to 64MB (64MB = 67108864 bytes).
+  options.write_buffer_size = 1 << 20;
```
pdillinger, line 1120: It would be nice not to use such large sizes on things, for test efficiency.

Also there's a lot of repetitive / copy-paste in this test that would be nice to trim down if reasonably possible.

pdillinger (approved):
LGTM, though testing is pretty basic for now. Would be nice to see a flush in there, though that could be upgraded in a later PR.

I will also assume that recovery (from WAL) testing can come in a later PR. (Still experimental feature)

db/memtable_list.cc
bjlemaire: Good comment - I'm adding a TODO to the code.

db/memtable_list.cc
```diff
@@ -476,8 +476,11 @@ Status MemTableList::TryInstallMemtableFlushResults(
       uint64_t min_wal_number_to_keep = 0;
       if (vset->db_options()->allow_2pc) {
         assert(edit_list.size() > 0);
+        // Note that if mempurge is successful, the edit_list will
+        // not be written to the manifest file.
```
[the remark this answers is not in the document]
bjlemaire: Done!

db/memtable_list.h
```diff
@@ -312,7 +312,18 @@ class MemTableList {
...
+  void FlushRequested() {
+    flush_requested_ = true;
+    // If there are some memtables stored in imm() that dont trigger
+    // flush (eg: mempurge output memtable), then update imm_flush_needed.
+    // Note: if race condition and imm_flush_needed is set to true
+    // when there is num_flush_not_started_==0, then there is no
+    // impact whatsoever. Imm_flush_needed is only used in an assert
+    // in IsFlushPending().
+    if (num_flush_not_started_ > 0) {
+      imm_flush_needed.store(true, std::memory_order_release);
```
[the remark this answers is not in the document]
bjlemaire: Useful to know - thanks. Note that `imm_flush_needed` isn't an `atomic<bool>` variable introduced in this PR - it has been part of the`MemTableList` struct for a long time, but surprisingly is only used for the `assert` mentioned in the comments here. I thought it would be a useful comment to add in the code because I actually asked myself the same question ("what was this `bool imm_flush_needed` used for").

db/db_flush_test.cc
bjlemaire: +1 on the repetitive/CTRL-C - CTRL-V code, I will try and make this unit test more human readable.

db/version_set.h
```diff
@@ -1160,6 +1161,15 @@ class VersionSet {
       if (min_log_num > num && !cfd->IsDropped()) {
         min_log_num = num;
       }
+      // If mempurge is activated, there may be an immutable memtable
+      // that has data not flushed to any SST file.
+      if (db_options_->experimental_allow_mempurge && !(cfd->IsEmpty()) &&
+          !(cfd->IsDropped())) {
+        num = cfd->imm()->EarliestLogContainingData();
```
anand1976, line 1168: Its not entirely clear why we need to do this. Shouldn't ```cfd->GetLogNumber()``` be sufficient? The log number in cfd is updated when appending a VersionEdit containing a log number to the manifest, which would only happen on a flush. So the log number in cfd would always be the earliest log with unflushed data I think.

anand1976 (commented):
@bjlemaire Sorry for the late review. The PR looks good. I had a clarifying question inline.

db/version_set.h
bjlemaire: You're 100% correct, thanks for pointing that out - I added this code snippet at a time when I didn't think I had to stop the VersionEdit from being appended to the manifest. I will submit an update right now.

bjlemaire: You're 100% correct, thanks for pointing that out - I added this code snippet at a time when I didn't think I had to stop the VersionEdit from being appended to the manifest. I will submit an update right now.

PR #4806, opened 2018-12-20
Extend StringAppendTESTOperator to use string delimiter of variable length

azagrebin:
This PR extend StringAppendTESTOperator class to use `std::string` delimiter instead of `char`. `char` becomes then a custom case of `std::string`. This allows for delimiter of variable length, including zero length if delimiter is not needed at all and space can be saved. The PR also extends `StringAppendOperator` in Java API.

PR #8536, opened 2021-07-15
Allow to use a string as a delimiter in StringAppendOperator

east825:
An arbitrary string can be used as a delimiter in StringAppend merge operator
flavor. In particular, it allows using an empty string, combining binary values for
the same key byte-to-byte one next to another.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -9,8 +9,8 @@
 ### Performance Improvements
 * Try to avoid updating DBOptions if `SetDBOptions()` does not change any option value.
 
-### Public API change
-* The constructor of `StringAppendOperator` and the factory method `MergeOperators::CreateStringAppendOperator` additionally accept a string as the delimiter.
+### Behavior Changes
+* `StringAppendOperator` additionally accepts a string as the delimiter.
 
 ## 6.23.0 (2021-07-16)
 ### Behavior Changes
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -9,6 +9,9 @@
 ### Performance Improvements
 * Try to avoid updating DBOptions if `SetDBOptions()` does not change any option value.
 
+### Public API change
+* The constructor of `StringAppendOperator` and the factory method `MergeOperators::CreateStringAppendOperator` additionally accept a string as the delimiter.
+
 ## 6.23.0 (2021-07-16)
 ### Behavior Changes
 * Obsolete keys in the bottommost level that were preserved for a snapshot will now be cleaned upon snapshot release in all cases. This form of compaction (snapshot release triggered compaction) previously had an artificial limitation that multiple tombstones needed to be present.
--- a/utilities/merge_operators.h
+++ b/utilities/merge_operators.h
@@ -20,6 +20,8 @@ class MergeOperators {
   static std::shared_ptr<MergeOperator> CreateUInt64AddOperator();
   static std::shared_ptr<MergeOperator> CreateStringAppendOperator();
   static std::shared_ptr<MergeOperator> CreateStringAppendOperator(char delim_char);
+  static std::shared_ptr<MergeOperator> CreateStringAppendOperator(
+      const std::string& delim);
   static std::shared_ptr<MergeOperator> CreateStringAppendTESTOperator();
   static std::shared_ptr<MergeOperator> CreateMaxOperator();
   static std::shared_ptr<MergeOperator> CreateBytesXOROperator();
--- a/utilities/merge_operators/string_append/stringappend.h
+++ b/utilities/merge_operators/string_append/stringappend.h
@@ -14,6 +14,7 @@ class StringAppendOperator : public AssociativeMergeOperator {
  public:
   // Constructor: specify delimiter
   explicit StringAppendOperator(char delim_char);
+  explicit StringAppendOperator(const std::string& delim);
 
   virtual bool Merge(const Slice& key,
                      const Slice* existing_value,
@@ -24,8 +25,7 @@ class StringAppendOperator : public AssociativeMergeOperator {
   virtual const char* Name() const override;
 
  private:
-  char delim_;         // The delimiter is inserted between elements
-
+  const std::string delim_;  // The delimiter is inserted between elements
 };
 
 }  // namespace ROCKSDB_NAMESPACE
--- a/utilities/merge_operators/string_append/stringappend.h
+++ b/utilities/merge_operators/string_append/stringappend.h
@@ -25,7 +25,7 @@ class StringAppendOperator : public AssociativeMergeOperator {
   virtual const char* Name() const override;
 
  private:
-  const std::string delim_;  // The delimiter is inserted between elements
+  std::string delim_;  // The delimiter is inserted between elements
 };
 
 }  // namespace ROCKSDB_NAMESPACE
--- a/utilities/merge_operators/string_append/stringappend2.cc
+++ b/utilities/merge_operators/string_append/stringappend2.cc
@@ -17,8 +17,10 @@ namespace ROCKSDB_NAMESPACE {
 
 // Constructor: also specify the delimiter character.
 StringAppendTESTOperator::StringAppendTESTOperator(char delim_char)
-    : delim_(delim_char) {
-}
+    : delim_(1, delim_char) {}
+
+StringAppendTESTOperator::StringAppendTESTOperator(const std::string& delim)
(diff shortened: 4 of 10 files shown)


mrambacher (commented):
I generally agree with these changes (and made many of the same ones in #8481).

My one concern is in the stringappend_test.  The way it is set up now it does not look as if the old 'char' interface is ever invoked, which would reduce the test coverage.  Is it possible to somehow restructure the tests so that both the new and old constructors can be exercised?

east825:
Oh, wow, I should have submitted my PR earlier :) 

I agree, not invoking the old constructor in tests is my oversight. I will try to get covered again.

east825:
For the history, I [brought up this idea](https://groups.google.com/u/1/g/rocksdb/c/IXNJNn6a4O8/m/I9kj7U7RAQAJ) in the Google group explaining why we need this at JetBrains specifically (currently we maintain our own fork only to be able to use an empty delimiter on merge). Also, there is an older [PR](https://github.com/facebook/rocksdb/pull/4806/commits) addressing the same limitation, but it seems abandoned for a while.

mrambacher:
@east825 If you wish for your PR to land before mine, I will work on getting it merged once you add the missing tests.  I can update my change to support this PR.  

Alternatively, your PR will have to merge with mine to add the Java APIs/tests.

east825:
@mrambacher Thanks a lot. I'll add these tests today. It indeed might be a bit cumbersome for me to merge after you since I'm not that experienced with the project's internals, but on the whole, I'll be happy with whoever changes adding this functionality to the API.

zhichao-cao:
Thanks for the contribution! It is a very useful extension to the operator. Please follow Mark's comment. I'm reviewing his 8481 PR and it would better this PR can be landed before his. Thanks!

zhichao-cao:
@adamretter Can you help to take a look at the Java changes of this PR?

utilities/merge_operators/string_append/stringappend2.cc
```diff
@@ -51,14 +53,14 @@ bool StringAppendTESTOperator::FullMergeV2(
     printDelim = true;
   } else if (numBytes) {
     merge_out->new_value.reserve(
-        numBytes - 1);  // Minus 1 since we have one less delimiter
+        numBytes - delim_.size());  // Adjust for one less delimiter
```
zhichao-cao, line 56: Need a simple explanation for why we minus delim_.size()

utilities/merge_operators.h
```diff
@@ -20,6 +20,8 @@ class MergeOperators {
   static std::shared_ptr<MergeOperator> CreateUInt64AddOperator();
   static std::shared_ptr<MergeOperator> CreateStringAppendOperator();
   static std::shared_ptr<MergeOperator> CreateStringAppendOperator(char delim_char);
+  static std::shared_ptr<MergeOperator> CreateStringAppendOperator(
```
zhichao-cao, line 23: Since this is a behavior change of StringAppend merge operator, also mention this change in HISTORY.md.

utilities/merge_operators/string_append/stringappend2.cc
east825: We want to reserve the space for the original value and N following operands with preceding delimiters. The size of "N following operands with preceding delimiters" is stored as `numBytes`. However, if there is no original value, the very first delimiter becomes redundant, therefore we reduce the amount of space needed by the size of one delimiter. 

For instance, if the original value was 'foo' and operands 'bar', 'baz' with delimiter ',', we would need `3 + numBytes = 3 + 2 * (3 + 1) = 11 bytes`, however if there is no original value, we need only `0 + numBytes - 1 = 7 bytes`.
This adjustment just takes into account that a delimiter can now be of an arbitrary length, not only one byte.

utilities/merge_operators.h
mrambacher: @zhichao-cao This header is not in the public API.  Should it still be mentioned in HISTORY or not?

utilities/merge_operators/string_append/stringappend.h
```diff
@@ -24,8 +25,7 @@ class StringAppendOperator : public AssociativeMergeOperator {
   virtual const char* Name() const override;
 
  private:
-  char delim_;         // The delimiter is inserted between elements
-
+  const std::string delim_;  // The delimiter is inserted between elements
```
mrambacher, line 28: Please remove the "const" here.  Private should be sufficient and it will not work with the Configurable stuff later.

utilities/merge_operators.h
east825: @mrambacher 
> I generally agree with these changes (and made many of the same ones in #8481).
> 
> My one concern is in the stringappend_test. The way it is set up now it does not look as if the old 'char' interface is ever invoked, which would reduce the test coverage. Is it possible to somehow restructure the tests so that both the new and old constructors can be exercised?

I went for the simplest solution switching between the two constructors depending on the length of a delimiter specified in a test. I don't want to add another layer of indirection on top of `StringAppendOperatorTest::OpenFuncPtr` to accommodate two possible signatures.

east825:
@mrambacher 
> I generally agree with these changes (and made many of the same ones in #8481).
> 
> My one concern is in the stringappend_test. The way it is set up now it does not look as if the old 'char' interface is ever invoked, which would reduce the test coverage. Is it possible to somehow restructure the tests so that both the new and old constructors can be exercised?

I went for the simplest solution, switching between the two constructors depending on the length of a delimiter specified in a test. I don't want to add another layer of indirection on top of `StringAppendOperatorTest::OpenFuncPtr` to accommodate two possible signatures. It will probably be difficult to wrap one's head around.

mrambacher (approved):
LGTM.  Thanks for the fix!  Can you rebase/update so the fix can be accepted?

east825:
I can squash everything into one changeset on my side if you want to. In some projects, it's done automatically, not sure which workflow is preferred here.

mrambacher:
> I can squash everything into one changeset on my side if you want to. In some projects, it's done automatically, not sure which workflow is preferred here.

No need to squash.  As long as you are up-to-date with master and there are no conflicts for the merge, this should be good to go.

utilities/merge_operators.h
zhichao-cao: I suggest to mention it in the HISTORY.md, since this merge operator is used by many users and  they should be area of this behavior change. So mention it in the HISTORY.md would be better.

HISTORY.md
```diff
@@ -9,6 +9,9 @@
 ### Performance Improvements
 * Try to avoid updating DBOptions if `SetDBOptions()` does not change any option value.
 
+### Public API change
```
zhichao-cao, line 12: You can add  ### Behavior Changes and move the comment below it.

utilities/merge_operators/string_append/stringappend2.cc
zhichao-cao: Thanks! Can you add these as the comment to the code?

utilities/merge_operators/string_append/stringappend_test.cc
```diff
@@ -259,7 +271,7 @@ TEST_P(StringAppendOperatorTest, SimpleTest) {
 }
 
 TEST_P(StringAppendOperatorTest, SimpleDelimiterTest) {
-  auto db = OpenDb('|');
```
zhichao-cao, line 262: For the test, can you keep all the original tests and add the new test that uses the string? The old interface is still in use and we do not want to remove the tests until the interface is deprecated. Also, keep the std::shared_ptr<DB> OpenTtlDb(char delim_char).

HISTORY.md
east825: Oh, sorry, I mixed up the two sections.

utilities/merge_operators/string_append/stringappend_test.cc
east825: But the old interface is still used if the specified string is of length 1, addressing the problem @mrambacher pointed at. 

It's a bit tricky to write completely independent tests using the new interface there since they are all parameterized and use two different database opening routines depending on the parameter in their `SetUp`. I believe it will complicate the setup step too much.

utilities/merge_operators.h
east825: I removed the mention of this specific header from HISTORY.md, leaving just `StringAppendOperator` itself.

PR #8537, opened 2021-07-15
Fix a issue with initializing blob header buffer

zaorangyang:


zhichao-cao (approved):

db/blob/blob_log_sequential_reader.h
```diff
@@ -10,6 +10,8 @@
 #include "db/blob/blob_log_format.h"
 #include "rocksdb/slice.h"
 
+#define MAX_HEADER_SIZE(a, b, c) (a > b ? (a > c ? a : c) : (b > c ? b : c))
```
ltamasi, line 13: Would be nice to use `std::max` here but it is unfortunately `constexpr` only since C++14.

ltamasi (approved):

db/blob/blob_log_sequential_reader.h
ltamasi: Could you `#undef` the macro near the bottom of the file just in case?

zaorangyang: Done  :)

PR #8539, opened 2021-07-15
Work around falsely reported data race on LRUHandle::flags

pdillinger:
Summary: Some bits are mutated and read while holding a lock, other
immutable bits (esp. secondary cache compatibility) can be read by
arbitrary threads without holding a lock. AFAIK, this doesn't cause an
issue on any architecture we care about, because you will get some
legitimate version of the value that includes the initialization, as
long as synchronization guarantees the initialization happens before the
read.

I've only seen this in #8538 so far, but it should be fixed regardless.
Otherwise, we'll surely get these false reports again some time.

Test Plan: some local TSAN test runs and in CircleCI

### DIFF
--- a/cache/lru_cache.cc
+++ b/cache/lru_cache.cc
@@ -557,6 +557,9 @@ Status LRUCacheShard::Insert(const Slice& key, uint32_t hash, void* value,
     e->SetSecondaryCacheCompatible(true);
     e->info_.helper = helper;
   } else {
+#ifdef __SANITIZE_THREAD__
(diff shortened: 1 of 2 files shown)


pdillinger:
Example failure: https://circleci.com/api/v1.1/project/github/facebook/rocksdb/163013/output/107/0?file=true&allocation-id=60f067cb701b4336a03d3508-0-build%2F6A7FB463

zhichao-cao (approved):
LGTM, thanks for the work around.

PR #8541, opened 2021-07-16
Fix db_stress SecondaryCache creation using URI

anand1976:
Revert to the original way of creating a SecondaryCache object in db_stress. The FB implementation of SecondaryCache is compatible with the URI format ('cachelibwrapper://.*'), which does not work with ```SecondaryCache::CreateFromString```.  The latter expects an option string of the format 'id=xxx;opt1=yyy;" and requires the implementation to support ```ConfigureFromMap()```.


zhichao-cao (approved):
LGTM.

anand1976:
Fixed by #8547

PR #8543, opened 2021-07-16
Standardize on GCC for TSAN conditional compilation

pdillinger:
Summary: In #8539 I accidentally only checked for GCC TSAN, which is
what I tested locally, while CircleCI and FB CI use clang TSAN. Related:
other existing code like in stack_trace.cc only check for clang TSAN.

I've now standardized these to the GCC convention in port/lang.h, so now

    #ifdef __SANITIZE_THREAD__

can check for any TSAN (assuming lang.h include)

Test Plan: Put an assert(false) in slice_test and look for the NOTE
about "signal-unsafe call", both GCC and clang. Eventually, CircleCI
TSAN in #8538

### DIFF
--- a/trace_replay/io_tracer.h
+++ b/trace_replay/io_tracer.h
@@ -9,6 +9,7 @@
 #include <fstream>
 
 #include "monitoring/instrumented_mutex.h"
+#include "port/lang.h"
 #include "rocksdb/file_system.h"
 #include "rocksdb/options.h"
 #include "trace_replay/trace_replay.h"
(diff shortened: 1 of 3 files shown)


trace_replay/io_tracer.h
```diff
@@ -157,20 +158,6 @@ class IOTracer {
   // mutex and ignore the operation if writer_is null. So its ok if
   // tracing_enabled shows non updated value.
 
-#if defined(__clang__)
```
zhichao-cao, line 160: I did not realize that we put them in the io_tracer.h interesting...

zhichao-cao (approved):
LGTM

PR #8544, opened 2021-07-16
Some fixes and enhancements to `ldb repair`

pdillinger:
Summary:
* Basic handling of SST file with just range tombstones rather than
failing assertion about smallest_seqno <= largest_seqno
* Adds --verbose option so that there exists a way to see the INFO
output from Repairer.

Test Plan: unit test added, manual testing for --verbose


ajkr (approved):
IIRC the range tombstones can stick out arbitrarily past the file's MANIFEST-specified bounds. This doc (https://github.com/cockroachdb/pebble/blob/master/docs/range_deletions.md) explains why the final state we are in is necessary (although tbh I still feel we arrived there by chance).

So including range tombstones can extend a file's bounds beyond what it is supposed to cover. Meanwhile (before this PR) excluding range tombstones could make a file's bounds too small. Maybe we can just add a comment that the current solution is known to be problematic.

ajkr:
I wonder if we should add table properties for smallest/largest key stored to MANIFEST, at least when they don't match the smallest/largest key in the file.

pdillinger:
I see, it should ideally have the extra logic that is in compact_job.cc but not in builder.cc (re: UpdateBoundariesForRange)?

ajkr:
> I see, it should ideally have the extra logic that is in compaction_job.cc but not in builder.cc (re: UpdateBoundariesForRange)?

Right, CompactionJob uses its knowledge about the neighboring files to decide.

pdillinger:
Unfortunately, there is some other issue (alleged memory leak) being reported by ASAN with this change, even with fix #8571 applied.

pdillinger:
@ajkr After more consideration (but not willing to invest a lot of time into this flawed repair tool), do you think it's better only to run this logic if `counter==0`, i.e. the SST file is only range tombstones?

ajkr:
> @ajkr After more consideration (but not willing to invest a lot of time into this flawed repair tool), do you think it's better only to run this logic if `counter==0`, i.e. the SST file is only range tombstones?

Hm, not particular. A couple options that feel consistent to me are:

- Always consider range tombstone boundaries that stick out past the point keys (i.e., this PR).
- Never consider range tombstone boundaries sticking out past the point keys. In case of `counter == 0` this would mean dropping the file.

PR #8545, opened 2021-07-16
Crashtest mempurge

bjlemaire:
Add `experiemental_allow_mempurge` flag support for `db_stress` and `db_crashtest.py`, with a `false` default value.
I succesfully tested locally both `whitebox` and `blackbox` crash tests with `experiemental_allow_mempurge` flag set as true.


akankshamahajan15 (approved):
LGTM.

PR #8546, opened 2021-07-16
Add experimental_allow_mempurge flag to benchmark.

bjlemaire:
Tiny PR to add the `experimental_allow_mempurge` to the `db_bench` tool (`Mempurge` is the current prototype for memtable garbage collection).
This is useful to benchmark the prototype of this new feature, stress test it and help find new meaningful heuristics for GC.
By default, the flag to allow `mempurge` is set to `false`.

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -1229,6 +1229,9 @@ DEFINE_bool(
 DEFINE_bool(allow_concurrent_memtable_write, true,
             "Allow multi-writers to update mem tables in parallel.");
 
+DEFINE_bool(experimental_allow_mempurge, false,
+            "Allow memtable garbage collection.");
+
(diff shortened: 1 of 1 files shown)


tools/db_bench_tool.cc
```diff
@@ -1229,6 +1229,9 @@ DEFINE_bool(
 DEFINE_bool(allow_concurrent_memtable_write, true,
             "Allow multi-writers to update mem tables in parallel.");
 
+DEFINE_bool(experimental_allow_mempurge, false,
```
anand1976, line 1232: Where is this flag used in db_bench? I don't see any reference to it.

bjlemaire: My bad, I never pushed the last commit - sorry about that

anand1976 (approved):

PR #8547, opened 2021-07-16
Allow CreateFromString to work on complex URIs

mrambacher:
Some URIs for creating instances (ala SecondaryCache) use complex URIs like (cache://name;prop=value).  These URIs were treated as name-value properties.  With this change, if the URI does not contain an "id=XX" setting, it will be treated as a single string value (and not an ID and map of name-value properties).


anand1976 (approved):
LGTM. Thanks for the quick turnaround.

PR #8555, opened 2021-07-19
Fix a minor issue with initializing the test path

zaorangyang:
Summary:
The PerThreadDBPath has already specified a slash. It does not need to be specified when initializing the test path.


zhichao-cao (approved):
LGTM

ajkr (changes requested):
Can you revert the file permission changes so we can keep it consistent? (See the "100644 → 100755" on all the changed files.)

jay-zhuang:
> Can you revert the file permission changes so we can keep it consistent? (See the "100644 → 100755" on all the changed files.)

Just updated it for the contributor.

zaorangyang:
Oops, I used a new development environment that runs on Samba. I made a mistake in Samba's configuration, which changed the permissions on the files. Thanks jay-zhuang for helping to solve the problem. But I find the test case of Facebook Internal - Linter is still failing. Could you please tell me what the problem is?

ajkr:
> > Can you revert the file permission changes so we can keep it consistent? (See the "100644 → 100755" on all the changed files.)
> 
> > > Can you revert the file permission changes so we can keep it consistent? (See the "100644 → 100755" on all the changed files.)
> > 
> > 
> > Just updated it for the contributor.
> 
> Oops, I used a new development environment that runs on Samba. I made a mistake in Samba's configuration, which changed the permissions on the files. Thanks jay-zhuang for helping to solve the problem. But I find the test case of Facebook Internal - Linter is still failing. Could you please tell me what the problem is?

It's complaining about some existing code in one of the files you changed (it doesn't want us to use `std::regex`). It's unrelated so we can land this without addressing it.

zaorangyang:
> > > Can you revert the file permission changes so we can keep it consistent? (See the "100644 → 100755" on all the changed files.)
> > 
> > 
> > > > Can you revert the file permission changes so we can keep it consistent? (See the "100644 → 100755" on all the changed files.)
> > > 
> > > 
> > > Just updated it for the contributor.
> > 
> > 
> > Oops, I used a new development environment that runs on Samba. I made a mistake in Samba's configuration, which changed the permissions on the files. Thanks jay-zhuang for helping to solve the problem. But I find the test case of Facebook Internal - Linter is still failing. Could you please tell me what the problem is?
> 
> It's complaining about some existing code in one of the files you changed (it doesn't want us to use `std::regex`). It's unrelated so we can land this without addressing it.

I see, thank you :)

PR #6716, opened 2020-04-16
print single sst meta when dump manifest

Connor1996:
Add `--sst_file_number` field for manifest_dump command to print the specified sst file meta. By the way, add key order check for `sst_dump`.

Usage:
$ ./ldb --db=/data3/zbk/kv4/deploy/data/db --hex manifest_dump  --sst_file_number=63
```
--------------- Column family "write"  (ID 2) --------------
63:132906243[3555338 .. 3555338]['7A311B40EFCC2CB4C5911ECF3937D728DED26AE53FA5E61BE04F23F2BE54EACC73' seq:3555
338, type:1 .. '7A313030302E25CD5F57252E' seq:3555338, type:1] oldest_ancester_time:0 at level 0
```

PR #8557, opened 2021-07-19
reimplement print single sst meta when dump manifest

hunterlxt:
original PR https://github.com/facebook/rocksdb/pull/6716

Add --sst_file_number field for manifest_dump command to print the specified sst file meta. By the way, add key order check for sst_dump.

Usage:
$ ./ldb --db=/data3/zbk/kv4/deploy/data/db --hex manifest_dump --sst_file_number=63
```
--------------- Column family "write"  (ID 2) --------------
63:132906243[3555338 .. 3555338]['7A311B40EFCC2CB4C5911ECF3937D728DED26AE53FA5E61BE04F23F2BE54EACC73' seq:3555
338, type:1 .. '7A313030302E25CD5F57252E' seq:3555338, type:1] oldest_ancester_time:0 at level 0
```

Signed-off-by: Xintao <hunterlxt@live.com>

### DIFF
--- a/db/version_edit_handler.cc
+++ b/db/version_edit_handler.cc
@@ -18,7 +18,8 @@
 namespace ROCKSDB_NAMESPACE {
 
 void VersionEditHandlerBase::Iterate(log::Reader& reader,
-                                     Status* log_read_status) {
+                                     Status* log_read_status,
+                                     uint64_t sst_file_number) {
   Slice record;
   std::string scratch;
   assert(log_read_status);
@@ -64,7 +65,7 @@ void VersionEditHandlerBase::Iterate(log::Reader& reader,
     s = *log_read_status;
   }
 
-  CheckIterationResult(reader, &s);
+  CheckIterationResult(reader, &s, nullptr, sst_file_number);
 
   if (!s.ok()) {
     status_ = s;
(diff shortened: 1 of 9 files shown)


hunterlxt:
@pdillinger PTAL

db/version_edit_handler.cc
```diff
@@ -892,42 +907,73 @@ Status ManifestTailer::VerifyFile(const std::string& fpath,
 }
 
 void DumpManifestHandler::CheckIterationResult(const log::Reader& reader,
-                                               Status* s) {
-  VersionEditHandler::CheckIterationResult(reader, s);
+                                               Status* s,
+                                               std::vector<FileMetaData*>*,
+                                               uint64_t sst_file_number) {
+  std::vector<FileMetaData*>* files = new std::vector<FileMetaData*>;
```
riversand963, line 913: Where is `files` free'ed?

db/version_edit_handler.cc
```diff
@@ -892,42 +907,73 @@ Status ManifestTailer::VerifyFile(const std::string& fpath,
...
+  bool found = false;
+  if (sst_file_number != 0) {
+    VersionEditHandler::CheckIterationResult(reader, s, files);
```
riversand963, line 916: Still trying to understand more about this PR, but at the moment, I am not fully convinced of the necessity of a separate `CheckIterationResult`.

riversand963 (changes requested):
Thanks @hunterlxt for bringing up the issue.
Can the same functionality be achieved without adding more options to `ldb` which already has a lot of options? Will using `grep` after the original result of `ldb` work for your use case?

riversand963:
I believe the same functionality can be achieved in another, simpler way. Closing.

cscetbon:
@riversand963 was it implemented using a simpler way like you said ? I can't find it

PR #8561, opened 2021-07-19
Cache warming blocks during flush

akankshamahajan15:
Summary: Insert warm blocks (data, uncompressed dict, index and filter blocks) during flush in Block cache which is enabled under option BlockBasedTableOptions.prepopulate_block_cache.

Test Plan: Added unit test

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/table/block_based/block_based_table_builder.cc
+++ b/table/block_based/block_based_table_builder.cc
@@ -36,6 +36,7 @@
 #include "table/block_based/block_based_table_factory.h"
 #include "table/block_based/block_based_table_reader.h"
 #include "table/block_based/block_builder.h"
+#include "table/block_based/block_like_traits.h"
 #include "table/block_based/filter_block.h"
 #include "table/block_based/filter_policy_internal.h"
 #include "table/block_based/full_filter_block.h"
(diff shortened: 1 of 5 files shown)


table/block_based/block_based_table_builder.cc
```diff
@@ -994,33 +995,39 @@ void BlockBasedTableBuilder::Flush() {
                                              r->get_offset());
     r->pc_rep->EmitBlock(block_rep);
   } else {
-    WriteBlock(&r->data_block, &r->pending_handle, true /* is_data_block */);
+    CachableEntry<Block> entry;
```
anand1976, line 998: We don't necessarily need to propagate this all the way down to ```InsertBlockInCache```, since the ```CachableEntry``` will not be used in this path.

table/block_based/block_based_table_builder.cc
```diff
@@ -1479,15 +1490,35 @@ Status BlockBasedTableBuilder::InsertBlockInCache(const Slice& block_contents,
...
+            rep_->table_options.filter_policy.get()));
+
+    if (block_holder->own_bytes()) {
+      size_t charge = block_holder->ApproximateMemoryUsage();
+      Cache::Handle* cache_handle = nullptr;
+
+      s = block_cache->Insert(
+          key, block_holder.get(),
+          BlocklikeTraits<TBlocklike>::GetCacheItemHelper(block_type), charge,
+          &cache_handle);
```
anand1976, line 1508: We could pass ```nullptr``` here instead of ```cache_handle```, and the block cache will take care of deleting the object if it could not be inserted due to the cache being full. It'll also be slightly more efficient since we won't have to call release for every block that's inserted here.

table/block_based/block_based_table_builder.cc
```diff
@@ -1247,13 +1255,14 @@ void BlockBasedTableBuilder::WriteRawBlock(const Slice& block_contents,
...
     if (io_s.ok()) {
       assert(s.ok());
-      if (is_data_block &&
-          r->table_options.prepopulate_block_cache ==
-              BlockBasedTableOptions::PrepopulateBlockCache::kFlushOnly) {
+      if (r->table_options.prepopulate_block_cache ==
+          BlockBasedTableOptions::PrepopulateBlockCache::kFlushOnly) {
         if (type == kNoCompression) {
-          s = InsertBlockInCache(block_contents, handle);
+          s = InsertBlockInCache(block_contents, handle, block_entry,
```
anand1976, line 1261: Do we insert all block types or just index/filter/data?

akankshamahajan15: @anand1976 It currently inserts all block types. I wasn't sure about other blocks  and their impact in terms of caching them so I inserted them also. Do we need them or they can be ignored? ```BlockBasedTableReader``` doesn't cache them and those blocks are used only once during opening the table so they can be ignored.

anand1976: @akankshamahajan15 They can be ignored, except for uncompression dict. I don't think ```BlockBasedTableReader``` even looks in the cache for other block types.

anand1976 (approved):

PR #8562, opened 2021-07-19
Add overwrite ratio to db_bench --benchmark=updaterandom.

bjlemaire:
Add a `num_overwrites` and a `overwrite_sample_size` flag to `db_bench`.
Add the possibility of performing an `updaterandom` benchmark with a fixed overwrite ratio _p_ (p=`num_overwrites`/`num`).
For each write operation, there is a probability _p_ that the write is an overwrite. 
When an overwrite is decided, a previously inserted key is chosen from a key sample (akin to a reservoir of existing keys). The size of the sample is set to be `overwrite_sample_size`.
When a pure write is decided, the key is inserted into the DB and randomly inserted into the key sample as well (the sample has a fixed size `overwrite_sample_size`).
The key sample/reservoir is used so that the user can decide if the overwrite are mostly targeting recently inserted keys (when `overwrite_sample_size` is small compared to num), or can target keys inserted "a long time ago" (when `overwrite_sample_size` is comparable to num).
No unit test specifically added.

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -6793,6 +6793,8 @@ class Benchmark {
     std::vector<std::string> inserted_keys_sample;
     std::default_random_engine bergen;
     double p = 0.0;
+    // Create Bernoulli distribution parameter for number
+    // of overwrites if user specified num_overwrites.
     if (FLAGS_num_overwrites >= 0) {
       p = FLAGS_num_overwrites * 1.0 / FLAGS_num;
       p = p > 1.0 ? 1.0 : p;
@@ -6808,21 +6810,32 @@ class Benchmark {
     // the number of iterations is the larger of read_ or write_
     while (!duration.Done(1)) {
       DB* db = SelectDB(thread);
+      // If user specified overwrite ratio:
       if (FLAGS_num_overwrites >= 0) {
+        // Decide if we do an overwrite (probability p).
+        // Note: an overwrite is possible only once at least
+        // one key has been inserted.
         if (inserted_keys_sample.size() > 0 && distribution(bergen)) {
+          // If overwrite, randomly select key from sample (reservoir).
           std::string newkey =
               inserted_keys_sample[thread->rand.Next() %
                                    inserted_keys_sample.size()];
+          // Copy string chars to key.
           for (size_t i = 0; i < key.size(); i++) {
             pos[i] = newkey[i];
           }
         } else {
+          // We perform a pure write.
           Status sget = Status::OK();
+          // Keep generating new key until we obtain a key
+          // not already present in the DB.
           do {
             GenerateKeyFromInt(thread->rand.Next() % FLAGS_num, FLAGS_num,
                                &key);
             sget = db->Get(options, key, &value);
           } while (sget.ok());
+
+          // Update the "inserted_keys_sample" reservoir.
           if (static_cast<uint64_t>(inserted_keys_sample.size()) <
               FLAGS_overwrite_sample_size) {
             inserted_keys_sample.push_back(key.ToString());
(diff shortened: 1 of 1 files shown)


tools/db_bench_tool.cc
```diff
@@ -896,6 +896,20 @@ DEFINE_int64(max_num_range_tombstones, 0,
 DEFINE_bool(expand_range_tombstones, false,
             "Expand range tombstone into sequential regular tombstones.");
 
+DEFINE_int32(num_overwrites, -1,
+             "Number of overwrites in UpdateRandom benchmark. This "
+             "number is used in combination with num to create a ratio r. "
```
pdillinger, line 901: Is `r` = `num_o` / `num` or `r` = `num_o` / `(num + num_o)`? That's not clear at all.

Also, I commonly omit --num and use --duration. It's not clear what this parameter means in that case without looking up the default for num. Can we just make it a double for overwrite_probability?

tools/db_bench_tool.cc
```diff
@@ -6776,15 +6790,63 @@ class Benchmark {
...
+    std::vector<std::string> inserted_keys_sample;
+    std::default_random_engine bergen;
+    double p = 0.0;
+    // Create Bernoulli distribution parameter for number
+    // of overwrites if user specified num_overwrites.
+    if (FLAGS_num_overwrites >= 0) {
+      p = FLAGS_num_overwrites * 1.0 / FLAGS_num;
+      p = p > 1.0 ? 1.0 : p;
+    }
+    std::bernoulli_distribution distribution(p);
```
pdillinger, line 6802: I might call it `overwrite_decider`

tools/db_bench_tool.cc
```diff
@@ -6776,15 +6790,63 @@ class Benchmark {
...
       DB* db = SelectDB(thread);
-      GenerateKeyFromInt(thread->rand.Next() % FLAGS_num, FLAGS_num, &key);
+      // If user specified overwrite ratio:
+      if (FLAGS_num_overwrites >= 0) {
+        // Decide if we do an overwrite (probability p).
+        // Note: an overwrite is possible only once at least
+        // one key has been inserted.
+        if (inserted_keys_sample.size() > 0 && distribution(bergen)) {
+          // If overwrite, randomly select key from sample (reservoir).
+          std::string newkey =
```
pdillinger, line 6820: There's a lot of allocation (std::string) and pseudorandom value generation in here. Have you measured the overhead of that, e.g. vs. num_overwrites=0?

tools/db_bench_tool.cc
```diff
@@ -6776,15 +6790,63 @@ class Benchmark {
...
+          // Copy string chars to key.
+          for (size_t i = 0; i < key.size(); i++) {
+            pos[i] = newkey[i];
+          }
+        } else {
+          // We perform a pure write.
+          Status sget = Status::OK();
+          // Keep generating new key until we obtain a key
+          // not already present in the DB.
+          do {
```
pdillinger, line 6832: It seems like in some cases this could spin and quietly dominate the cost of the test, leading to misleading results. Thoughts?

tools/db_bench_tool.cc
bjlemaire: Fixed

tools/db_bench_tool.cc
bjlemaire: Good call, just updated it

tools/db_bench_tool.cc
bjlemaire: After a bit of testing , it looks like there is a substantial difference in throughput for different `overwrite_probabilities`. The main driver of the difference is actually the `db->Get` operation that checks if the newly generated key is existing or not (even when the do-while loop is replaced by a single iteration).
In my opinion, the only type of benchmark that wouldn't be affected is the `fillSeq` benchmarl - because then we know exactly which keys have been inserted before or not (so no need for `db->Get`), and there is no need for random key generation either (except maybe a single call for selecting a random overwrite).

bjlemaire:
Closing this PR - the `overwrite_probability` knob will be added to another existing benchmark where we have better guarantees of consistent throughput regardless of the `overwrite_probability` value.

PR #8563, opened 2021-07-20
CompactionJob::Install(): fix log truncation

rockeet:
event log info may be truncated, the default buffer size is 512, this PR changes buffer size to 8192.


jay-zhuang (approved):
LGTM

PR #8564, opened 2021-07-20
[db_stress] Cleanup cf handlers before deleting db

jay-zhuang:
Delete column family handlers before deleting db to avoid `last_ref`
assert.

Test Plan: Inject compaction test in db_stress test

### DIFF
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -2568,6 +2568,7 @@ void StressTest::Open() {
               }
               column_families_.clear();
               delete db_;
+              db_ = nullptr;
             }
           }
           if (!s.ok()) {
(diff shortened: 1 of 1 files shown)


db_stress_tool/db_stress_test_base.cc
```diff
@@ -2563,6 +2563,10 @@ void StressTest::Open() {
             s = static_cast_with_check<DBImpl>(db_->GetRootDB())
                     ->TEST_WaitForCompact(true);
             if (!s.ok()) {
+              for (auto cf : column_families_) {
+                delete cf;
+              }
+              column_families_.clear();
               delete db_;
```
pdillinger, line 2570: Unless we're in a destructor or otherwise the "last place" a `delete db_` might happen, I prefer to set `db_ = nullptr` to make diagnosis of double-delete easy.

pdillinger (approved):
Thanks!

siying (approved):
Thank you so much for fixing it.

PR #8565, opened 2021-07-21
Several simple local code clean-ups

riversand963:
This PR tries to remove some unnecessary checks as well as unreachable code blocks to
improve readability. An obvious non-public API method naming typo is also corrected.

Test plan:
make check

### DIFF
--- a/db/db_impl/db_impl_write.cc
+++ b/db/db_impl/db_impl_write.cc
@@ -220,7 +220,8 @@ Status DBImpl::WriteImpl(const WriteOptions& write_options,
 
   bool need_log_sync = write_options.sync;
   bool need_log_dir_sync = need_log_sync && !log_dir_synced_;
-  if (!two_write_queues_ || !disable_memtable) {
+  assert(!two_write_queues_ || !disable_memtable);
+  {
     // With concurrent writes we do preprocess only in the write thread that
     // also does write to memtable to avoid sync issue on shared data structure
     // with the other thread
--- a/utilities/transactions/write_prepared_txn.cc
+++ b/utilities/transactions/write_prepared_txn.cc
@@ -158,7 +158,7 @@ Status WritePreparedTxn::CommitInternal() {
     // When not writing to memtable, we can still cache the latest write batch.
     // The cached batch will be written to memtable in WriteRecoverableState
     // during FlushMemTable
-    WriteBatchInternal::SetAsLastestPersistentState(working_batch);
+    WriteBatchInternal::SetAsLatestPersistentState(working_batch);
   }
 
   auto prepare_seq = GetId();
--- a/utilities/transactions/write_prepared_txn.cc
+++ b/utilities/transactions/write_prepared_txn.cc
@@ -236,14 +236,6 @@ Status WritePreparedTxn::CommitInternal() {
                           NO_REF_LOG, DISABLE_MEMTABLE, &seq_used, ONE_BATCH,
                           &update_commit_map_with_aux_batch);
   assert(!s.ok() || seq_used != kMaxSequenceNumber);
-  if (UNLIKELY(!db_impl_->immutable_db_options().two_write_queues)) {
-    if (s.ok()) {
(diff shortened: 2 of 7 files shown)


db/db_impl/db_impl_write.cc
```diff
@@ -220,7 +220,8 @@ Status DBImpl::WriteImpl(const WriteOptions& write_options,
 
   bool need_log_sync = write_options.sync;
   bool need_log_dir_sync = need_log_sync && !log_dir_synced_;
-  if (!two_write_queues_ || !disable_memtable) {
+  assert(!two_write_queues_ || !disable_memtable);
```
riversand963, line 223: Should `two_write_queues_ && disable_memtable` is true, we would have returned from line 112:
```
if (two_write_queues_ && disable_memtable) {
  return WriteImplWALOnly(...);
}
```

utilities/transactions/write_prepared_txn.cc
```diff
@@ -236,14 +236,6 @@ Status WritePreparedTxn::CommitInternal() {
                           NO_REF_LOG, DISABLE_MEMTABLE, &seq_used, ONE_BATCH,
                           &update_commit_map_with_aux_batch);
   assert(!s.ok() || seq_used != kMaxSequenceNumber);
-  if (UNLIKELY(!db_impl_->immutable_db_options().two_write_queues)) {
```
riversand963, line 239: This cannot be true.
If `!two_write_queues_` is true, then in line 177, we would have `do_one_write` being true. Therefore, the `if` block starting from line 203 will be entered, causing the function to return from there.

utilities/transactions/write_prepared_txn_db.cc
```diff
@@ -508,24 +508,22 @@ void WritePreparedTxnDB::AddCommitted(uint64_t prepare_seq, uint64_t commit_seq,
...
-          // This is a rare case that txn is committed but prepared_txns_ is not
-          // cleaned up yet. Refer to delayed_prepared_commits_ definition for
-          // why it should be kept updated.
-          delayed_prepared_commits_[evicted.prep_seq] = evicted.commit_seq;
-          ROCKS_LOG_DEBUG(info_log_,
-                          "delayed_prepared_commits_[%" PRIu64 "]=%" PRIu64,
-                          evicted.prep_seq, evicted.commit_seq);
-          break;
-        }
+      auto dp_iter = delayed_prepared_.find(evicted.prep_seq);
```
riversand963, line 513: Instead of iterating the map and check for equality, use `map::find()`.

utilities/transactions/write_prepared_txn_db.cc
```diff
@@ -508,24 +508,22 @@ void WritePreparedTxnDB::AddCommitted(uint64_t prepare_seq, uint64_t commit_seq,
...
+        // why it should be kept updated.
+        delayed_prepared_commits_[evicted.prep_seq] = evicted.commit_seq;
+        ROCKS_LOG_DEBUG(info_log_,
+                        "delayed_prepared_commits_[%" PRIu64 "]=%" PRIu64,
+                        evicted.prep_seq, evicted.commit_seq);
       }
     }
+    // After each eviction from commit cache, check if the commit entry should
+    // be kept around because it overlaps with a live snapshot.
+    CheckAgainstSnapshots(evicted);
```
lth, line 526: I don't think the ordering matters, but what was the reasoning behind moving this later?

lth (approved):

utilities/transactions/write_prepared_txn_db.cc
riversand963: Yeah, I don't think the ordering matters here, too.
I switched the position of `CheckAginastSnapshots()` and the code block updating `delayed_prepared_commits_` because we are evicting from the data structures related to prepared transactions in `AdvanceMaxEvictedSeq()`, while this code block is finishing up the eviction process. This change moves the code block closer to `AdvanceMaxEvictedSeq()`, hoping it can be easier to follow the logic.

PR #8567, opened 2021-07-21
Remove TaskLimiterToken::ReleaseOnce for fix

pdillinger:
Summary: Rare TSAN and valgrind failures are caused by unnecessary
reading of a field on the TaskLimiterToken::limiter_ for an assertion
after the token has been released and the limiter destroyed. To simplify
we can simply destroy the token before triggering DB shutdown
(potentially destroying the limiter). This makes the ReleaseOnce logic
unnecessary.

Test Plan: watch for more failures in CI


ajkr (approved):
LGTM. The point of ReleaseOnce() was to avoid deallocation under DB mutex though I have no evidence it makes a meaningful difference.

PR #8568, opened 2021-07-21
Make backup restore atomic, with sync option

### LINKED ISSUE #8500: Corruption upon power loss after restoring a backup
If power loss occurs shortly after restoring a backup, database files may be left in an incomplete state. A workaround is to manually sync the filesystem after the restore.

Side note, corruption also happens if power cuts off during a restore, but this seems to be a deliberate choice of the current implementation (`BackupEngineImpl::RestoreDBFromBackup` effectively starts off with an `rm -r`). Is there any interest in making restores atomic/would that be feasible?

### Expected behavior

Power loss can cause loss of data (last N commits), but will not corrupt database.

### Actual behavior

Errors upon database re-open, e.g. on btrfs:

```
Corruption: file is too short (2487 bytes) to be an sstable
```

Or, sometimes:

```
CURRENT file does not end with newline
```

1. Restore a backup a few times in a row
2. Cut power
3. Goto 1

My testbed for this includes a remote-controlled power switch, but it reliably happens within a few cycles. Tested on 6.20.3.

pdillinger:
Summary: Guarantees that if a restore is interrupted, DB::Open will fail. This works by
restoring CURRENT first to CURRENT.tmp then as a final step renaming to CURRENT.

Also makes restore respect BackupEngineOptions::sync (default true). When set,
the restore is guaranteed persisted by the time it returns OK. Also makes the above
atomicity guarantee work in case the interruption is power loss or OS crash (not just
process interruption or crash).

Fixes #8500

Test Plan: added to backup mini-stress unit test. Passes with
gtest_repeat=100 (whereas fails 7 times without the CURRENT.tmp)


utilities/backupable/backupable_db_test.cc
```diff
@@ -3122,82 +3122,108 @@ TEST_F(BackupEngineTest, Concurrency) {
...
-      BackupEngine* my_be;
-      if (reopen) {
-        ASSERT_OK(BackupEngine::Open(test_db_env_.get(), be_opts, &my_be));
-      } else {
-        my_be = backup_engine_.get();
-      }
+          // Open restored DB to verify its contents, but test atomic restore
+          // by doing it async and ensuring we either get OK or InvalidArgument
+          restore_verify_threads[i] =
+              std::thread([this, &db_opts, restore_db_dir, to_restore] {
```
pdillinger, line 3182: This is essentially the new part. Most of the rest is whitespace change from formatter

akankshamahajan15 (approved):
LGTM! Thanks for the fix.

ajkr:
I realized one more thing for #8500. Besides asking for atomicity they were also asking why a completed restore could be corrupted by a power reset. Actually the bug is straightforward -- we are passing `sync=false` when copying files for restore: https://github.com/facebook/rocksdb/blob/ab7f7c9e497a18c48837556bf2e68b4cc5d2a498/utilities/backupable/backupable_db.cc#L1731

pdillinger:
> I realized one more thing for #8500. Besides asking for atomicity they were also asking why a completed restore could be corrupted by a power reset. Actually the bug is straightforward -- we are passing `sync=false` when copying files for restore:
> 
> https://github.com/facebook/rocksdb/blob/ab7f7c9e497a18c48837556bf2e68b4cc5d2a498/utilities/backupable/backupable_db.cc#L1731

Good find! I'll re-work this slightly so that the existing BackupEngineOptions::sync option controls whether we sync and fsync in Restore (in addition to Backup). Moving the CURRENT file at the end will be unconditional, because that's useful protection even without power interruption or OS crash (e.g. against process crash or interruption) and is essentially free.

utilities/backupable/backupable_db.cc
```diff
@@ -1722,13 +1727,36 @@ Status BackupEngineImpl::RestoreDBFromBackup(const RestoreOptions& options,
     }
     // 3. Construct the final path
     // kWalFile lives in wal_dir and all the rest live in db_dir
-    dst = ((type == kWalFile) ? wal_dir : db_dir) + "/" + dst;
+    if (type == kWalFile) {
+      dst = wal_dir + "/" + dst;
+      if (options_.sync && !wal_dir_for_fsync) {
+        s = db_env_->NewDirectory(wal_dir, &db_dir_for_fsync);
```
ajkr, line 1733: s/db_dir_for_fsync/wal_dir_for_fsync/

ajkr (approved):
LGTM!

utilities/backupable/backupable_db.cc
pdillinger: Nice catch! If only there were a tool for that. https://scan.coverity.com/o/oss_success_stories/45

PR #8569, opened 2021-07-21
Add overwrite_probability for filluniquerandom benchmark in db_bench

bjlemaire:
Add flags `overwrite_probability` and `overwrite_window_size` flag to `db_bench`.
Add the possibility of performing a `filluniquerandom` benchmark with an overwrite probability.
For each write operation, there is a probability _p_ that the write is an overwrite (_p_=`overwrite_probability`).
When an overwrite is decided, the key is randomly chosen from the last _N_ keys previously inserted into the DB (with _N_=`overwrite_window_size`).
When a pure write is decided, the key inserted into the DB is unique and therefore will not be an overwrite.
The `overwrite_window_size` is used so that the user can decide if the overwrite are mostly targeting recently inserted keys (when `overwrite_window_size` is small compared to the total number of writes), or can also target keys inserted "a long time ago" (when `overwrite_window_size` is comparable to total number of writes).
Note that total number of writes = # of unique insertions + # of overwrites.
No unit test specifically added. 
Local testing show the following **throughputs** for `filluniquerandom` with 1M total writes:
- bypass the code inserts (no `overwrite_probability` flag specified): ~14.0MB/s
- `overwrite_probability=0.99`, `overwrite_window_size=10`: ~17.0MB/s
- `overwrite_probability=0.10`, `overwrite_window_size=10`: ~14.0MB/s
- `overwrite_probability=0.99`, `overwrite_window_size=1M`: ~14.5MB/s
- `overwrite_probability=0.10`, `overwrite_window_size=1M`: ~14.0MB/s

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -448,24 +448,23 @@ DEFINE_double(compression_ratio, 0.5, "Arrange to generate values that shrink"
 
 DEFINE_double(
     overwrite_probability, -1.0,
-    "Used in 'filluniquerandom' and "
-    "'filluniquerandomdeterministic' benchmarks: for each write operation, "
+    "Used in 'filluniquerandom' benchmark: for each write operation, "
     "we give a probability to perform an overwrite instead. The key used for "
-    "the overwrite is randomly chosen from the last 'overwrite_reservoir_size' "
+    "the overwrite is randomly chosen from the last 'overwrite_window_size' "
     "keys "
     "previously inserted into the DB. "
     "Valid overwrite_probability values: [0.0, 1.0].");
 
-DEFINE_uint32(overwrite_reservoir_size, 1,
-              "Used in 'filluniquerandom' and "
-              "'filluniquerandomdeterministic' benchmarks. For each write "
+DEFINE_uint32(overwrite_window_size, 1,
+              "Used in 'filluniquerandom' benchmark. For each write "
               "operation, when "
               "the overwrite_probability flag is set by the user, the key used "
               "to perform "
               "an overwrite is randomly chosen from the last "
-              "'overwrite_reservoir_size' keys "
+              "'overwrite_window_size' keys "
               "previously inserted into the DB. "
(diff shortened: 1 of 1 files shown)


tools/db_bench_tool.cc
```diff
@@ -4807,6 +4827,31 @@ class Benchmark {
...
+      // If overwrite set by user, and UNIQUE_RANDOM mode on,
+      // the overwrite_window_size must be > 0.
+      if (write_mode == UNIQUE_RANDOM && FLAGS_overwrite_window_size == 0) {
+        fprintf(stderr,
+                "Overwrite_window_size must be  strictly greater than 0.\n");
+        ErrorExit();
+      }
+    }
+    std::default_random_engine overwrite_gen;
+    std::bernoulli_distribution overwrite_decider(p);
```
pdillinger, line 4845: Minor: Seeing `FLAGS_seed` below reminds me that we should use that for these random choices also.

tools/db_bench_tool.cc
```diff
@@ -446,6 +446,26 @@ DEFINE_int32(num_multi_db, 0,
 DEFINE_double(compression_ratio, 0.5, "Arrange to generate values that shrink"
               " to this fraction of their original size after compression");
 
+DEFINE_double(
+    overwrite_probability, -1.0,
```
pdillinger, line 450: Or should we let 0.0 mean "disable"?

pdillinger (approved):
Overall LGTM, thanks!

tools/db_bench_tool.cc
bjlemaire: Oh right! Here skipping the code snippet (disable) and p=0.0 is strictly equivalent (unlike in other benchmarks) - good catch.

tools/db_bench_tool.cc
bjlemaire: That's a good point, I should use the FLAGS_seed for all random choices. I'm editing it right now.

PR #8571, opened 2021-07-21
Fix use-after-free on implicit temporary FileOptions

pdillinger:
Summary: FileOptions has an implicit conversion from EnvOptions and some
internal APIs take `const FileOptions&` and save the reference, which is
counter to Google C++ guidelines,

> Avoid defining functions that require a const reference parameter to outlive the call, because const reference parameters bind to temporaries. Instead, find a way to eliminate the lifetime requirement (for example, by copying the parameter), or pass it by const pointer and document the lifetime and non-null requirements.

This is at least a problem for repair.cc, which passes an EnvOptions to
TableCache(), which would save a reference to the temporary copy as
FileOptions. This was unfortunately only caught as a side effect of
changes in #8544.

This change fixes the repair.cc case and updates the involved internal
APIs that save a reference to use `const FileOptions*` instead.

Unfortunately, I don't know how to get any of our sanitizers to reliably
report bugs like this, so I can't rule out more existing in our
codebase.

Test Plan: Test that issues seen with #8544 are fixed (can reproduce on
AWS EC2)


mrambacher (commented):
This change looks more or less fine to me, but wonder how much overkill it is.  Would you have essentially the same outcome by changing the env_options_ in Repairer to file_options_, without other changes?

pdillinger:
> This change looks more or less fine to me, but wonder how much overkill it is. Would you have essentially the same outcome by changing the env_options_ in Repairer to file_options_, without other changes?

It leaves another accident waiting to happen. And it wouldn't be so concerning if we had a predictable way to catch this undefined behavior but it appears we have no means of automatic detection for this. Do you have an alternative idea for avoiding this type of bug?

pdillinger:
In general, changing a couple dozen LoC to follow our prescribed style guide does not seem like overkill.

mrambacher:
> > This change looks more or less fine to me, but wonder how much overkill it is. Would you have essentially the same outcome by changing the env_options_ in Repairer to file_options_, without other changes?
> 
> It leaves another accident waiting to happen. And it wouldn't be so concerning if we had a predictable way to catch this undefined behavior but it appears we have no means of automatic detection for this. Do you have an alternative idea for avoiding this type of bug?

I don't think there is a good solution to necessarily avoid this type of bug in general terms.  For this specific case, you could eliminate the constructor FileOptions(const EnvOptions&) and see what blows up.  Eliminating that constructor would prevent the "hidden conversion" of one type to another.  If the conversion is needed, you could make an explicit static method -- e.g. static FileOptions FileOptions::FromEnvOptions(const EnvOptions&) -- to do the conversion.

pdillinger:
We shouldn't just remove public APIs for internal clean-up. I highly suspect (but have not confirmed, @anand1976 #5761) that the implicit conversion was added to ease some API compatibility issues in code using RocksDB. Because we never save `*Options` objects from public APIs without copying, we do not have to worry about the same kind of bug from unexpected temporary copy when entering via public APIs.

And I think it's a great convention not to save references to const& parameters. (Make them const* if they're going to be saved for use after call.)

mrambacher:
> We shouldn't just remove public APIs for internal clean-up. I highly suspect (but have not confirmed, @anand1976 #5761) that the implicit conversion was added to ease some API compatibility issues in code using RocksDB. Because we never save `*Options` objects from public APIs without copying, we do not have to worry about the same kind of bug from unexpected temporary copy when entering via public APIs.
> 
> And I think it's a great convention not to save references to const& parameters. (Make them const* if they're going to be saved for use after call.)

I agree with both of those statements.  Why not change TableCache to store a FileOptions rather than FileOptions& then and avoid most of the rest of these changes?

ajkr (approved):
LGTM, I hadn't thought of using pointer argument to make contracts about lifetime safer (from implicit conversions and accidental misuse in general).

> I agree with both of those statements. Why not change TableCache to store a FileOptions rather than FileOptions& then and avoid most of the rest of these changes?

The current PR keeps the intended existing behavior of not copying FileOptions into TableCache. I also don't think that behavior needs to be kept, but it's not the purpose of this PR, so am ok with leaving it.

pdillinger:
> FIXME: some other memory leaks still reported somehow with #8544 changes

Fixed in #8589

PR #8573, opened 2021-07-22
Fix GetSortedWalFiles() IO error during file deletion

jay-zhuang:
Summary: `GetSortedWalFiles()` may return IO error if WAL files are
deleting. Error message:
`IO error: No such file or directory: while stat a file for size: ...`
It is caused by race condition between files purging and WalManager
getting wals.

Also correct HISTORY for #8518.

Test Plan: Added an unittest to reproduce

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,9 +2,15 @@
 ## Unreleased
 ### Bug Fixes
 * If the primary's CURRENT file is missing or inaccessible, the secondary instance should not hang repeatedly trying to switch to a new MANIFEST. It should instead return the error code encountered while accessing the file.
+* Fix `GetSortedWalFiles()` may return IO error during WAL file deletion.
+
 ### New Features
 * Made the EventListener extend the Customizable class.
 * EventListeners that have a non-empty Name() and that are registered with the ObjectRegistry can now be serialized to/from the OPTIONS file.
+
+### Performance Improvements
+* Try to avoid updating DBOptions if `SetDBOptions()` does not change any option value.
+
 ## 6.23.0 (2021-07-16)
 ### Behavior Changes
 * Obsolete keys in the bottommost level that were preserved for a snapshot will now be cleaned upon snapshot release in all cases. This form of compaction (snapshot release triggered compaction) previously had an artificial limitation that multiple tombstones needed to be present.
@@ -27,9 +33,6 @@
 ### Public API change
 * Added APIs to the Customizable class to allow developers to create their own Customizable classes.  Created the utilities/customizable_util.h file to contain helper methods for developing new Customizable classes.
 * Change signature of SecondaryCache::Name().  Make SecondaryCache customizable and add SecondaryCache::CreateFromString method.
-
-### Performance Improvements
-* Try to avoid updating DBOptions if `SetDBOptions()` does not change any option value.
 
 ## 6.22.0 (2021-06-18)
 ### Behavior Changes
--- a/db/wal_manager.cc
+++ b/db/wal_manager.cc
@@ -314,18 +314,12 @@ Status WalManager::GetSortedWalsOfType(const std::string& path,
 
       uint64_t size_bytes;
       s = env_->GetFileSize(LogFileName(path, number), &size_bytes);
-      // re-try in case the alive log file has been moved to archive.
+      // re-try in case the alive WAL file has been moved to archive.
       if (!s.ok() && log_type == kAliveLogFile) {
         std::string archived_file = ArchivedLogFileName(path, number);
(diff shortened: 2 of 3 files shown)


db/wal_manager.cc
```diff
@@ -321,9 +321,11 @@ Status WalManager::GetSortedWalsOfType(const std::string& path,
           s = env_->GetFileSize(archived_file, &size_bytes);
           if (!s.ok() && env_->FileExists(archived_file).IsNotFound()) {
             // oops, the file just got deleted from archived dir! move on
-            s = Status::OK();
             continue;
           }
+        } else {
+          // the file just got deleted from WAL folder, skip it
```
mrambacher, line 327: I would feel more comfortable if you stored the original FileExists status and asserted here that the status IsNotFound.  Otherwise, you are assuming the status is either OK or NotFound and there is not some other issue.

db/wal_manager.cc
```diff
@@ -321,9 +321,11 @@ Status WalManager::GetSortedWalsOfType(const std::string& path,
           s = env_->GetFileSize(archived_file, &size_bytes);
           if (!s.ok() && env_->FileExists(archived_file).IsNotFound()) {
```
mrambacher, line 322: I am not sure I understand why we need to do an Exists, FileSize, Exists rather than just the FileSize.  The problem seems like there is a potential race between these three operations.  What is wrong with just FileSize and checking there if the file is found or not?

db/wal_manager.cc
```diff
@@ -321,9 +321,11 @@ Status WalManager::GetSortedWalsOfType(const std::string& path,
...
             // oops, the file just got deleted from archived dir! move on
-            s = Status::OK();
```
mrambacher, line 324: Why is this case no longer needed?  It seems like we need to guard against three possible async/deletion points, and this is one of them.

db/wal_manager.cc
jay-zhuang: I don't know. Base on the current logic, seems it could be file exist but FileSize unable to get, in such case it will return error. If we only check FileSize then it has to `continue` to skip that WAL file if FileSize failed.

db/wal_manager.cc
jay-zhuang: sure, make sense.

db/wal_manager.cc
riversand963: I am also curious to know if we can just use `GetFileSize()`, since we need to get its size anyway.

mrambacher: There are two cases -- one for the wal_dir and one for the archive_dir.  The wal_dir uses GetFileSize directly.  It seems like the archive one could as well.

jay-zhuang: make sense, switched to just use `GetFileSize()`.

db/wal_manager.cc
pdillinger: This doesn't seem to distinguish FileSystem error from file not found, so I don't think this simultaneously solves:
* Tolerates background deletion with OK status but potentially incomplete/inconsistent results
* Ensures a FileSystem error is reported as non-OK status (rather than interpreted as missing WAL file, which seems dangerous if caller has ensured there is no background deletion thus results should be pristine on OK)

HISTORY.md
```diff
@@ -2,9 +2,15 @@
...
 ### Bug Fixes
 * If the primary's CURRENT file is missing or inaccessible, the secondary instance should not hang repeatedly trying to switch to a new MANIFEST. It should instead return the error code encountered while accessing the file.
+* Fix `GetSortedWalFiles()` may return IO error during WAL file deletion.
+
 ### New Features
 * Made the EventListener extend the Customizable class.
 * EventListeners that have a non-empty Name() and that are registered with the ObjectRegistry can now be serialized to/from the OPTIONS file.
+
+### Performance Improvements
+* Try to avoid updating DBOptions if `SetDBOptions()` does not change any option value.
```
pdillinger, line 12: Does this need to be moved or is this an accidental merge issue?

jay-zhuang: sorry for the confusion, it needs to move to correct my previous merge issue. I will create a separate minor PR for that.

db/wal_manager.cc
jay-zhuang: Agree, I think we could add `FileExists().IsNotFound()` to explicitly check if the file exists or not.

jay-zhuang:
@siying fixed the problem in #8591 by disabling file deletion during the API call. We believe that would be better, as this patch doesn't fully address the problem that underline file system could be inconsistent. Abandoning the patch.

PR #8574, opened 2021-07-22
Fix an race condition during multiple DB opening

jay-zhuang:
Summary: ObjectLibrary is shared between multiple DB instances, the
Register() could have race condition.

Test Plan: pass the failed test


ajkr (approved):
LGTM

mrambacher (approved):
LGTM.  Thanks!

PR #8575, opened 2021-07-22
Analyze MultiGet in trace_analyzer

zhichao-cao:
Now we can analyze the MultiGet queries in the trace file and generate a set of the statistic and analysis files. Note that, when one MultiGet access N keys, we count each sub-get-query individually. But the over all query number is still the MultiGet not the sub-get-query.

test plan: added new unit test and make check


anand1976 (approved):
LGTM. Thanks

PR #8576, opened 2021-07-22
Correct javadoc for Env#setBackgroundThreads(int)

bremac:
By default, the low priority pool is not the flush pool, so calling `Env#setBackgroundThreads` without providing a priority will not do what the caller expected.


adamretter (approved):
LGTM

adamretter:
@mrambacher Can you merge?

PR #8577, opened 2021-07-22
Add MultiGet to replay

zhichao-cao:
When the trace contains the MultiGet record, with this PR, it can replay the MultiGet.

test plan: make check and replay the real trace.


anand1976 (approved):
LGTM. Thanks

PR #8582, opened 2021-07-23
Allow WAL dir to change with db dir

mrambacher:
Prior to this change, the "wal_dir"  DBOption would always be set (defaults to dbname) when the DBOptions were sanitized.  Because of this setitng in the options file, it was not possible to rename/relocate a database directory after it had been created and use the existing options file.

After this change, the "wal_dir" option is only set under specific circumstances.  Methods were added to the ImmutableDBOptions class to see if it is set and if it is set to something other than the dbname.  Additionally, a method was added to retrieve the effective value of the WAL dir (either the option or the dbname/path).

Tests were added to the core and ldb to test that a database could be created and renamed without issue.  Additional tests for various permutations of wal_dir were also added.


autopear:
Looks good to me. But `OPTIONS-000007` should not be there.

HISTORY.md
```diff
@@ -2,7 +2,7 @@
 ## 6.23.0 (2021-07-16)
 ### Behavior Changes
 * Obsolete keys in the bottommost level that were preserved for a snapshot will now be cleaned upon snapshot release in all cases. This form of compaction (snapshot release triggered compaction) previously had an artificial limitation that multiple tombstones needed to be present.
-
+* Allow the options.wal_filter to be blank.  If the wal_filter is not specified, the system will store the WALs in the database directory.  This change allows databases to be relocated and opened without updating the corresponding Options files.
```
pdillinger, line 5: s/wal_filter/wal_dir/ but I don't think there is such a behavior change. wal_dir could previously be empty (default) and WALs stored in DB directory.

This is only a change (bug fix I would argue) to APIs like `LoadLatestOptions` or `ldb --try_load_options`.

Or I guess if the client is reading wal_dir from `GetOptions()`, it might now be empty instead of sanitized to be set to db directory. Again, that is a change to `GetOptions()`, not to `DB::Open()`. For this case we should explicitly recommend change to `.GetWalDir()` from `.wal_dir`.

db/db_impl/db_impl_files.cc
```diff
@@ -548,7 +549,7 @@ void DBImpl::PurgeObsoleteFiles(JobContext& state, bool schedule_only) {
       dir_to_sync = candidate_file.file_path;
     } else {
       dir_to_sync =
-          (type == kWalFile) ? immutable_db_options_.wal_dir : dbname_;
+          (type == kWalFile) ? immutable_db_options_.GetWalDir() : dbname_;
```
pdillinger, line 552: Nit: is `auto wal_dir` still in scope?

db/db_impl/db_impl_open.cc
```diff
@@ -111,16 +111,22 @@ DBOptions SanitizeOptions(const std::string& dbname, const DBOptions& src,
...
     // Use dbname as default
     result.wal_dir = dbname;
   }
-  if (result.wal_dir.back() == '/') {
-    result.wal_dir = result.wal_dir.substr(0, result.wal_dir.size() - 1);
+  if (!result.wal_dir.empty()) {
+    auto npath = NormalizePath(dbname + "/");
+    if (npath == NormalizePath(result.wal_dir + "/") &&
+        npath == NormalizePath(result.db_paths[0].path + "/")) {
+      result.wal_dir.clear();
```
pdillinger, line 124: Please drop a *why* comment in here.

pdillinger (commented):
Is this forward-compatibility safe? If RocksDB version before this change reads a new option file with wal_dir="", will it be sanitized? (I don't see a call to SanitizeOptions in options_parser.cc nor options_util.cc)

mrambacher:
> Is this forward-compatibility safe? If RocksDB version before this change reads a new option file with wal_dir="", will it be sanitized? (I don't see a call to SanitizeOptions in options_parser.cc nor options_util.cc)

I am not sure I understand what you mean by "forward compatibility safe".  I tried to make this work for older/newer releases by:
- If the WAL_DIR is not set (which is the default when an option is created) "most of the time".  The wal_dir option is not read directly by accessed via the GetWalDir() API to get the true path.  
- If the WAL_DIR matches the dbname, we unset the WAL_DIR.  This would be the typical pattern when reading an old options file (setting wal_dir=dbname was the old behavior of SanitizeOptions).

Since the wal_dir property is changed in the DB::Open code (via SanitizeOptions), it should not need to be updated/sanitized when it is read/written to the Options file.

HISTORY.md
mrambacher: From the client perspective, the only thing that changes is GetOptions().wal_dir will not be sanitized/updated and will (more often than not) be left as the value passed in originally.  Internally, this change allows the DB directory to be relocated under certain circumstances.

The .GetWalDir method is not available as part of the public API.  When reading WAL files (GetCurrent/SortedWalFile, the path is part of the LogFile information returned.  If requested (and if not already present), I can add tests calling these methods with different settings of wal_dir.

db/db_impl/db_impl_files.cc
mrambacher: Fixed.

db/db_impl/db_impl_open.cc
mrambacher: Fixed.

HISTORY.md
mrambacher: Where would you like this to appear in the HISTORY?  The change should allow a database to be moved and still opened via LoadLatestOptions (well, if db_log_dir is not set).  Or it can simply be limited to fixing ldb --try_load_options from the HISTORY perspective.

pdillinger:
> I am not sure I understand what you mean by "forward compatibility safe".

If someone is using `LoadLatestOptions` to open their DB, can they safely downgrade (and upgrade) with/without this change? Because previously we would not write empty wal_dir to the options file. But I guess the answer is that it's taken care of in the sanitization on DB::Open, even if the DBOptions object looks slightly different going in to DB::Open.

So LGTM once HISTORY.md is updated

HISTORY.md
pdillinger: We have evidence of users considering this an unexpected, bad behavior, so I would call it a bug fix. Suggestion:

> Fixed handling of `DBOptions::wal_dir` with `LoadLatestOptions()` or `ldb --try_load_options` on a copied or moved DB. Previously, when the WAL directory is same as DB directory (default), a copied or moved DB would reference the old path of the DB as the WAL directory, potentially corrupting both copies. Under this change, the `wal_dir` from `DB::GetOptions()` or `LoadLatestOptions()` can now be empty, which indicates that the *current* DB directory is used for WALs. This is also a subtle API change.

pdillinger (approved):
Thanks! Don't forget to re-import

autopear:
LGTM

autopear (approved):
LGTM

utilities/options/options_util_test.cc
```diff
@@ -634,6 +634,130 @@ TEST_F(OptionsUtilTest, BadLatestOptions) {
...
+  // should be empty
+  options.wal_dir = dbname_ + "/";
+  options.db_paths.emplace_back(dbname_, std::numeric_limits<uint64_t>::max());
+  ASSERT_OK(DB::Open(options, dbname_, &db));
+  delete db;
+  ASSERT_OK(LoadLatestOptions(dbname_, options.env, &db_opts, &cf_descs));
+  ASSERT_EQ(db_opts.wal_dir, "");
+  ASSERT_OK(DestroyDB(dbname_, options));
+
+  // Open a DB with no wal_dir but db_path != db_name.  The wal_dir == dbname_
```
zaorangyang, line 713: I have some questions about this case. I think wal_dir should still be empty, even if db_path != db_name. After you rename/move db_name, you will lose old WAL file when you restart from the options file, because the directory wal_dir points to no longer exists.

mrambacher: @zaorangyang This case was left for backward compatibility.  Prior to this fix, the wal_dir would always be set -- typically to dbname.  In the case were dbPath was used and it was not the same as the dbname, we would still use the wal_dir as dbname, meaning the wal dir was not the same as the db path.  

Obviously, this is not ideal but changing the code to not set the wal_dir may break existing applications.  One solution to address this may be in improving the checkpoint system and have the checkpoint do a better/complete job of copying all the files into the appropriate directory hierarchy.

If you feel strongly about this case, I suggest opening a new issue/PR and proposing a fix.  I would be more than happy to review it and help it through the process!

zaorangyang: Thanks for your reply. After I finish the https://github.com/facebook/rocksdb/pull/8603, I would try to fix it.

PR #8583, opened 2021-07-23
Add simple heuristics for experimental mempurge.

bjlemaire:
Add `experimental_mempurge_policy` option flag and introduce two new `MemPurge` (Memtable Garbage Collection) policies: 'ALWAYS' and 'ALTERNATE'. Default value: ALTERNATE.
`ALWAYS`: every flush will first go through a `MemPurge` process. If the output is too big to fit into a single memtable, then the mempurge is aborted and a regular flush process carries on. `ALWAYS` is designed for user that need to reduce the number of L0 SST file created to a strict minimum, and can afford a small dent in performance (possibly hits to CPU usage, read efficiency, and maximum burst write throughput). 
`ALTERNATE`: a flush is transformed into a `MemPurge` except if one of the memtables being flushed is the product of a previous `MemPurge`. `ALTERNATE` is a good tradeoff between reduction in number of L0 SST files created and performance. `ALTERNATE` perform particularly well for completely random garbage ratios, or garbage ratios anywhere in (0%,50%], and even higher when there is a wild variability in garbage ratios.
This PR also includes support for `experimental_mempurge_policy` in `db_bench`. 
Testing was done locally by replacing all the `MemPurge` policies of the unit tests with `ALTERNATE`, as well as local testing with `db_crashtest.py` `whitebox` and `blackbox`. Overall, if an `ALWAYS` mempurge policy passes the tests, there is no reasons why an `ALTERNATE` policy would fail, and therefore the mempurge policy was set to `ALWAYS` for all mempurge unit tests.

### DIFF
--- a/db/c.cc
+++ b/db/c.cc
@@ -3034,22 +3034,6 @@ void rocksdb_options_set_experimental_allow_mempurge(rocksdb_options_t* opt,
   opt->rep.experimental_allow_mempurge = v;
 }
 
-void rocksdb_options_set_experimental_mempurge_policy(rocksdb_options_t* opt,
-                                                      int v) {
-  switch (v) {
-    case 0:
-      opt->rep.experimental_mempurge_policy =
-          ROCKSDB_NAMESPACE::MemPurgePolicy::kAlternate;
-      break;
-    case 1:
-      opt->rep.experimental_mempurge_policy =
-          ROCKSDB_NAMESPACE::MemPurgePolicy::kAlways;
-      break;
-    default:
-      assert(0);
-  }
-}
-
 void rocksdb_options_set_access_hint_on_compaction_start(
     rocksdb_options_t* opt, int v) {
   switch(v) {
--- a/db/c.cc
+++ b/db/c.cc
@@ -3034,6 +3034,22 @@ void rocksdb_options_set_experimental_allow_mempurge(rocksdb_options_t* opt,
   opt->rep.experimental_allow_mempurge = v;
 }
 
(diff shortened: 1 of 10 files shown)


db/c.cc
```diff
@@ -3034,6 +3034,22 @@ void rocksdb_options_set_experimental_allow_mempurge(rocksdb_options_t* opt,
   opt->rep.experimental_allow_mempurge = v;
 }
 
+void rocksdb_options_set_experimental_mempurge_policy(rocksdb_options_t* opt,
```
pdillinger, line 3037: I would not bother exporting experimental API to alternate language interfaces.

db/flush_job.cc
```diff
@@ -572,10 +611,32 @@ Status FlushJob::MemPurge() {
   } else {
     TEST_SYNC_POINT("DBImpl::FlushJob:MemPurgeUnsuccessful");
   }
+  const uint64_t micros = clock_->NowMicros() - start_micros;
+  const uint64_t cpu_micros = clock_->CPUNanos() / 1000 - start_cpu_micros;
+  ROCKS_LOG_INFO(db_options_.info_log,
+                 "[%s] [JOB %d] Mempurge lasted %" PRIu64
+                 " microseconds %" PRIu64
```
pdillinger, line 618: Recommend comma after "microseconds"

pdillinger (approved):
I might call these "trivial" or "simple" heuristics but LGTM

pdillinger:
> can afford a small dent in performance.

Specifically, likely hits to CPU usage, read efficiency, and maximum burst write throughput, though not always so.

PR #8584, opened 2021-07-23
Disable DistributedMutex test by default

jay-zhuang:
Summary: DistributedMutex hasn't been used in the code base and enabling
`USE_FOLLY_DISTRIBUTED_MUTEX` only runs the mutex tests from third-party
lib. So disabling it for now.
The implementation may also out of date, should re-sync with folly before
using.

Test Plan: CI


ajkr (approved):
LGTM

PR #8585, opened 2021-07-24
Fix NotifyOnFlushCompleted() for atomic flush

riversand963:
PR #5908 added `flush_jobs_info_` to `FlushJob` to make sure
`OnFlushCompleted()` is called after committing flush results to
MANIFEST. However, `flush_jobs_info_` is not updated in atomic
flush, causing `NotifyOnFlushCompleted()` to skip `OnFlushCompleted()`.

This PR fixes this, in a similar way to #5908 that handles regular flush.

Test plan:
make check


jay-zhuang (approved):
LGTM

riversand963:

PR #8589, opened 2021-07-27
Fix missing Handle release in TableCache::GetRangeTombstoneIterator

pdillinger:
Summary: This appears to be little used code so not a major bug, but is
blocking #8544

Test Plan: Added regression test to the end of
DBRangeDelTest::TableEvictedDuringScan. Without this fix, ASAN reports
memory leak.


ajkr (approved):
LGTM

PR #8590, opened 2021-07-27
Fix insecure internal API for GetImpl

pdillinger:
Summary: Calling the GetImpl function could leave reference to a local
callback function in a field of a parameter struct. As this is
performance-critical code, I'm not going to attempt to sanitize this
code too much, but make the existing hack a bit cleaner by reverting
what it overwrites in the input struct.

Added SaveAndRestore utility class to make that easier.

Test Plan: added unit test for SaveAndRestore; existing tests for
GetImpl

### DIFF
--- a/util/defer.h
+++ b/util/defer.h
@@ -55,6 +55,7 @@ class Defer final {
 template <typename T>
 class SaveAndRestore {
  public:
+  // obj is non-null pointer to value to be saved and later restored.
   explicit SaveAndRestore(T* obj) : obj_(obj), saved_(*obj) {}
   ~SaveAndRestore() { *obj_ = std::move(saved_); }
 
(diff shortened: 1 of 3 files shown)


util/defer.h
```diff
@@ -49,4 +49,22 @@ class Defer final {
   std::function<void()> fn_;
 };
 
+// An RAII utility object that saves the current value of an object so that
+// it can be overwritten, and restores it to the saved value when the
+// SaveAndRestore object goes out of scope.
+template <typename T>
+class SaveAndRestore {
+ public:
+  explicit SaveAndRestore(T* obj) : obj_(obj), saved_(*obj) {}
```
riversand963, line 58: Being paranoid, but I think it is safe to make sure `obj` is not nullptr.

riversand963 (approved):
LGTM with minor comments.

util/defer.h
pdillinger: By the time we can write `assert(obj)` we have already used it. I can put it in a comment though

PR #8592, opened 2021-07-28
Replace macros in compaction_iterator.cc with inline functions

jimmycFB:
Internal task T96186510.

Created new inline member functions in `CompactionIterator`,
`DefinitelyInSnapshot`, `DefinitelyNotInSnapshot`, and
`InEarliestSnapshot` to replace the macros at the top of
`compaction_iterator.cc`.

Placed the definitions in `compaction_iterator.h` in accordance with
Google's style guide for inline functions. Separated the declarations
and definitions, and only placed the `inline` keyword on the
definitions, in line with ISO CPP recommendations.

### DIFF
--- a/db/compaction/compaction_iterator.cc
+++ b/db/compaction/compaction_iterator.cc
@@ -16,22 +16,6 @@
 #include "table/internal_iterator.h"
 #include "test_util/sync_point.h"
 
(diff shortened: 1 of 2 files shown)


riversand963 (approved):
Thanks @jimmycFB for the PR.

PR #8593, opened 2021-07-28
Create fillanddeleteuniquerandom benchmark (db_bench), with new option flags.

bjlemaire:
Introduction of a new `fillanddeleteuniquerandom` benchmark (`db_bench`) with 5 new option flags to simulate a benchmark where the following sequence is repeated multiple times:
"A set of keys S1 is inserted ('`disposable entries`'), then after some delay another set of keys S2 is inserted ('`persistent entries`') and the first set of keys S1 is deleted. S2 artificially represents the insertion of hypothetical results from some undefined computation done on the first set of keys S1. The next sequence can start as soon as the last disposable entry in the set S1 of this sequence is inserted, if the `delay` is non negligible." 
New flags:
- `disposable_entries_delete_delay`: minimum delay in microseconds between insertion of the last `disposable` entry, and the start of the insertion of the first `persistent` entry.
- `disposable_entries_batch_size`: number of `disposable` entries inserted at the beginning of each sequence.
- `disposable_entries_value_size`: size of the random `value` string for the `disposable` entries.
- `persistent_entries_batch_size`: number of `persistent` entries inserted at the end of each sequence, right before the deletion of the `disposable` entries starts.
- `persistent_entries_value_size`: size of the random value string for the `persistent` entries.

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -351,17 +351,35 @@ DEFINE_uint32(overwrite_window_size, 1,
               "Warning: large values can affect throughput. "
               "Valid overwrite_window_size values: [1, kMaxUint32].");
 
-DEFINE_int32(delete_delay, 0,
-             "Minimum delay in microseconds for the delete range "
-             "to be issued. When 0 the insertion of the last targeted entry is "
-             "immediately "
-             "followed by the issuance of the DeleteRange.");
+DEFINE_uint64(
+    selective_deletes_delay, 0,
+    "Minimum delay in microseconds for the series of Deletes "
+    "to be issued. When 0 the insertion of the last targeted entry is "
+    "immediately "
+    "followed by the issuance of the Deletes.");
 
-DEFINE_int32(delete_range_size, 0,
-             "Size of the key range to delete after 'delete_delay' "
-             "microseconds. When 0 no deleterange is issued. A DeleteRange is "
-             "always issued "
-             "once all the keys it covers are inserted into the DB. ");
+DEFINE_uint64(selective_deletes_batch_size, 0,
+              "Number of consecutively inserted KV entries to delete after "
+              "'delete_delay' "
+              "microseconds. When 0 no selective Deletes is issued. A series "
+              "of Deletes is "
+              "always issued "
+              "once all the KV entries it covers are inserted into the DB. ");
+
+DEFINE_uint64(selective_deletes_value_size, 0,
+              "Size of the values (in bytes) of the entries targeted by "
+              "selective deletes. ");
+
+DEFINE_uint64(
+    selective_no_deletes_batch_size, 0,
+    "Number of KV entries being inserted right before the selective deletes "
+    "happen. "
+    "These keys are not targeted by the selective deletes, and will always "
+    "stay valid in the DB. ");
+
+DEFINE_uint64(selective_no_deletes_value_size, 0,
+              "Size of the values (in bytes) of the entries not targeted by "
+              "selective deletes. ");
 
 DEFINE_bool(
     key_before_delete_range, true,
@@ -4682,6 +4700,15 @@ class Benchmark {
       return std::numeric_limits<uint64_t>::max();
     }
 
+    uint64_t Fetch(uint64_t index) {
+      if (mode_ == UNIQUE_RANDOM) {
+        assert(index < values_.size());
+        return values_[index];
+      }
+      assert(false);
+      return std::numeric_limits<uint64_t>::max();
+    }
+
    private:
     Random64* rand_;
     WriteMode mode_;
@@ -4709,6 +4736,8 @@ class Benchmark {
   double SineRate(double x) {
     return FLAGS_sine_a*sin((FLAGS_sine_b*x) + FLAGS_sine_c) + FLAGS_sine_d;
   }
+
+  void WriteOrSelectiveDelete();
 
   void DoWrite(ThreadState* thread, WriteMode write_mode) {
     const int test_duration = write_mode == RANDOM ? FLAGS_duration : 0;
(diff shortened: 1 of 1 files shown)


tools/db_bench_tool.cc
```diff
@@ -4763,6 +4815,48 @@ class Benchmark {
...
+    // some delay another set of keys S2 is inserted ('persistent entries')
+    // and the first set of keys S1 is deleted. S2 artificially represents
+    // the insertion of hypothetical results from some undefined computation
+    // done on the first set of keys S1. The next sequence can start as soon
+    // as the last disposable entry in the set S1 of this sequence is
+    // inserted, if the delay is non negligible"
+    bool skip_for_loop = false, is_disposable_entry = true;
+    std::vector<uint64_t> disposable_entries_index(num_key_gens, 0);
+    std::vector<uint64_t> persistent_ent_and_del_index(num_key_gens, 0);
+    const uint64_t NUM_DISP_AND_PERS_ENTRIES =
```
pdillinger, line 4832: Google C++ style says k prefix for constants, as in kNumDispAndPersEntries

tools/db_bench_tool.cc
```diff
@@ -4814,11 +4908,104 @@ class Benchmark {
...
+          }
         } else {
           rand_num = key_gens[id]->Next();
         }
         GenerateKeyFromInt(rand_num, FLAGS_num, &key);
-        Slice val = gen.Generate();
+        Slice val;
+        if (NUM_DISP_AND_PERS_ENTRIES > 0) {
+          if (is_disposable_entry) {
+            val = Slice(rnd_disposable_entry.RandomString(
```
pdillinger, line 4999: Looks like use-after-free on temporary string. You might want to compile+run your benchmark once with COMPILE_WITH_ASAN=1 USE_CLANG=1 and once with COMPILE_WITH_UBSAN=1 USE_CLANG=1. Be sure to let the process finish (as Ctrl+C would skip errors to report).

tools/db_bench_tool.cc
```diff
@@ -351,6 +352,46 @@ DEFINE_uint32(overwrite_window_size, 1,
...
+    "--benchmarks='filluniquerandom' "
+    "and used when--disposable_entries_batch_size is > 0).");
+
+DEFINE_int32(persistent_entries_value_size, 64,
+             "Size of the values (in bytes) of the entries not targeted by "
+             "deletes. (only compatible with --benchmarks='filluniquerandom' "
+             "and used when--disposable_entries_batch_size is > 0).");
+
+DEFINE_bool(
+    key_before_delete_range, true,
```
pdillinger, line 389: Unused?

tools/db_bench_tool.cc
```diff
@@ -351,6 +352,46 @@ DEFINE_uint32(overwrite_window_size, 1,
               "Warning: large values can affect throughput. "
               "Valid overwrite_window_size values: [1, kMaxUint32].");
 
+DEFINE_uint64(
+    disposable_entries_delete_delay, 0,
```
pdillinger, line 356: I am slightly concerned about "polluting" the global option space with benchmark-specific parameters, but I don't have a good suggested alternative. But please clearly mark these as only (currently) applying to filluniquerandom. Or would it make sense to make this a new benchmark like fillanddeleteuniquerandom? Not sure, legit asking

tools/db_bench_tool.cc
```diff
@@ -4663,6 +4704,15 @@ class Benchmark {
       return std::numeric_limits<uint64_t>::max();
     }
 
+    uint64_t Fetch(uint64_t index) {
+      if (mode_ == UNIQUE_RANDOM) {
+        assert(index < values_.size());
+        return values_[index];
+      }
+      assert(false);
```
pdillinger, line 4712: I think it would be simpler / clearer to use `assert(mode_ == UNIQUE_RANDOM);` above and skip this section. (This isn't production code where we should reasonably try to avoid undefined behavior in case of broken invariants.)

tools/db_bench_tool.cc
```diff
@@ -4691,6 +4741,8 @@ class Benchmark {
     return FLAGS_sine_a*sin((FLAGS_sine_b*x) + FLAGS_sine_c) + FLAGS_sine_d;
   }
 
+  void WriteOrSelectiveDelete();
```
pdillinger, line 4744: Undefined / unused?

pdillinger (changes requested):

tools/db_bench_tool.cc
bjlemaire: This is also something I've been struggling with. 
I can totally create a new benchmark "fillanddeleteuniquerandom" [but this is very superficial since we still have to use the DoWrite function under the hood (used by all the fill... benchmarks)], but in all the scenarios I've thought of, we still end up having these 5 new option flags hanging around...
For now I can create this new benchmark and clearly mark that these flags only apply to this new benchmark (that will make explanations clearer), and I will explicitly mention it for each flag.

tools/db_bench_tool.cc
bjlemaire: I am cleaning that up right now - I cant believe the armada of warnings/compilers didn't catch this.

tools/db_bench_tool.cc
bjlemaire: Forgot to remove it  - thanks

tools/db_bench_tool.cc
bjlemaire: Fixed - thanks for the catch, I'm used to writing const var in all caps

tools/db_bench_tool.cc
bjlemaire: Spot on (impressive). Just fixed it.

pdillinger (approved):
👍

PR #8595, opened 2021-07-29
[RateLimiter] Implement superior user & mid IO priority level in GenericRateLimiter

hx235:
Context:
An extra IO_USER priority in rate limiter allows users to optionally charge WAL writes / SST reads to rate limiter at this priority level, which then has higher priority than IO_HIGH and IO_LOW. With an extra IO_USER priority, it allows users to better specify the relative urgency/importance among different requests in rate limiter. As a consequence, IO resource management can better prioritize and limit resource based on user's need. 

The IO_USER is implemented as superior priority in GenericRateLimiter, in the sense that its request queue will always be iterated first without being constrained to fairness. The reason is that the notion of fairness is only meaningful in helping lower priorities in background IO (i.e, IO_HIGH/MID/LOW) to gain some fair chance to run so that it does not block foreground IO (i.e, the ones that are charged at the level of IO_USER). As we can see, the ultimate goal here is to not blocking foreground IO at IO_USER level, which justifies the superiority of IO_USER.

Similar benefits exist for IO_MID priority.

Summary:
- Rewrote the logic of deciding the order of iterating request queues of high/low priorities to include the extra user/mid priority w/o affecting the existing behavior (see PR's [comment](https://github.com/facebook/rocksdb/pull/8595/files#r678749331))
- Included the request queue of user-pri/mid-pri in the code path of next-leader-candidate signaling and GenericRateLimiter's destructor
- Included the extra user/mid-pri in bookkeeping data structures: total_bytes_through_ and total_requests_
- Re-written the previous impl of explicitly iterating priorities with a loop from Env::IO_LOW to Env::IO_TOTAL 

Test:
- passed existing rate_limiter_test.cc
- passed added unit tests in rate_limiter_test.cc
- run performance test to verify performance with only high/low requests is not affected by this change 
   - Set-up command:
   `TEST_TMPDIR=/dev/shm ./db_bench --benchmarks=fillrandom --duration=5 --compression_type=none --num=100000000 --disable_auto_compactions=true --write_buffer_size=1048576 --writable_file_max_buffer_size=65536 --target_file_size_base=1048576 --max_bytes_for_level_base=4194304 --level0_slowdown_writes_trigger=$(((1 << 31) - 1)) --level0_stop_writes_trigger=$(((1 << 31) - 1))`

    - Test command:
   `TEST_TMPDIR=/dev/shm ./db_bench --benchmarks=overwrite --use_existing_db=true --disable_wal=true --duration=30 --compression_type=none --num=100000000 --write_buffer_size=1048576 --writable_file_max_buffer_size=65536 --target_file_size_base=1048576 --max_bytes_for_level_base=4194304 --level0_slowdown_writes_trigger=$(((1 << 31) - 1)) --level0_stop_writes_trigger=$(((1 << 31) - 1)) --statistics=true --rate_limiter_bytes_per_sec=1048576 --rate_limiter_refill_period_us=1000  --threads=32 |& grep -E '(flush|compact)\.write\.bytes'`

   - Before (on branch upstream/master):
   `rocksdb.compact.write.bytes COUNT : 4014162`
   `rocksdb.flush.write.bytes COUNT : 26715832`
    rocksdb.flush.write.bytes/rocksdb.compact.write.bytes ~= 6.66
   
   - After (on branch rate_limiter_user_pri):
  `rocksdb.compact.write.bytes COUNT : 3807822`
  `rocksdb.flush.write.bytes COUNT : 26098659`
   rocksdb.flush.write.bytes/rocksdb.compact.write.bytes ~= 6.85

### DIFF
--- a/util/rate_limiter.cc
+++ b/util/rate_limiter.cc
@@ -83,7 +83,7 @@ GenericRateLimiter::~GenericRateLimiter() {
   }
   requests_to_wait_ = static_cast<int32_t>(queues_size_sum);
 
-  for (int i = Env::IO_LOW; i < Env::IO_TOTAL; ++i) {
+  for (int i = Env::IO_TOTAL - 1; i >= Env::IO_LOW; --i) {
     std::deque<Req*> queue = queue_[i];
     for (auto& r : queue) {
       r->cv.Signal();
--- a/util/rate_limiter.cc
+++ b/util/rate_limiter.cc
@@ -218,7 +218,10 @@ GenericRateLimiter::GeneratePriorityIterationOrder() {
   pri_iteration_order[0] = Env::IO_USER;
 
   bool high_pri_iterated_after_mid_low_pri = rnd_.OneIn(fairness_);
+  TEST_SYNC_POINT_CALLBACK("GenericRateLimiter::GeneratePriorityIterationOrder::PostRandomOneInFairnessForHighPri", &high_pri_iterated_after_mid_low_pri);
   bool mid_pri_itereated_after_low_pri = rnd_.OneIn(fairness_);
+  TEST_SYNC_POINT_CALLBACK("GenericRateLimiter::GeneratePriorityIterationOrder::PostRandomOneInFairnessForMidPri", &mid_pri_itereated_after_low_pri);
+  
 
   if (high_pri_iterated_after_mid_low_pri) {
     pri_iteration_order[3] = Env::IO_HIGH;
@@ -234,6 +237,7 @@ GenericRateLimiter::GeneratePriorityIterationOrder() {
         (pri_iteration_order[3] == Env::IO_MID) ? Env::IO_LOW : Env::IO_MID;
   }
 
+  TEST_SYNC_POINT_CALLBACK("GenericRateLimiter::GeneratePriorityIterationOrder::PreReturnPriIterationOrder", &pri_iteration_order);
   return pri_iteration_order;
 }
 
--- a/util/rate_limiter.cc
+++ b/util/rate_limiter.cc
@@ -215,14 +215,20 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
 std::vector<Env::IOPriority>
 GenericRateLimiter::GeneratePriorityIterationOrder() {
   std::vector<Env::IOPriority> pri_iteration_order(Env::IO_TOTAL /* 4 */);
-  // We make Env::IO_USER a superior priority by always iterating its queue first
+  // We make Env::IO_USER a superior priority by always iterating its queue
+  // first
   pri_iteration_order[0] = Env::IO_USER;
 
   bool high_pri_iterated_after_mid_low_pri = rnd_.OneIn(fairness_);
-  TEST_SYNC_POINT_CALLBACK("GenericRateLimiter::GeneratePriorityIterationOrder::PostRandomOneInFairnessForHighPri", &high_pri_iterated_after_mid_low_pri);
+  TEST_SYNC_POINT_CALLBACK(
+      "GenericRateLimiter::GeneratePriorityIterationOrder::"
+      "PostRandomOneInFairnessForHighPri",
+      &high_pri_iterated_after_mid_low_pri);
   bool mid_pri_itereated_after_low_pri = rnd_.OneIn(fairness_);
-  TEST_SYNC_POINT_CALLBACK("GenericRateLimiter::GeneratePriorityIterationOrder::PostRandomOneInFairnessForMidPri", &mid_pri_itereated_after_low_pri);
-  
+  TEST_SYNC_POINT_CALLBACK(
+      "GenericRateLimiter::GeneratePriorityIterationOrder::"
+      "PostRandomOneInFairnessForMidPri",
+      &mid_pri_itereated_after_low_pri);
 
   if (high_pri_iterated_after_mid_low_pri) {
     pri_iteration_order[3] = Env::IO_HIGH;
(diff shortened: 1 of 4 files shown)


util/rate_limiter.cc
```diff
@@ -228,6 +240,30 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
   } while (!r.granted);
 }
 
+std::vector<Env::IOPriority>
+GenericRateLimiter::GeneratePriorityIterationOrder() {
```
hx235, line 244: The logic of deciding the order of iterating request queues of different priorities (i.e, the order of satisfying requests of different priorities) is as the following:
(1)There is 1/fairness chance that user_pri requests will be satisfied AFTER BOTH high_pri and low_pri requests. Otherwise, user_pri requests will be satisfied BEFORE BOTH high_pri and low_pri requests.
(2) Regardless of the order of satisfying user_pri requests relative to BOTH high_pri and low_pri, there is 1/fairness chance that high_pri requests will be satisfied AFTER low_pri requests. Otherwise, high_pri requests will be satisfied BEFORE low_pri requests.

Therefore the only 4 possible iteration orders as following:
[user_pri, high_pri, low_pri], [user_pri, low_pri, high_pri], [high_pri, low_pri, user_pri], [low_pri, high_pri, user_pri],

hx235: [TODO]The current impl verbosely manipulating pri in generating the order for the benefit of readability. Should we have more pri levels, this verbose impl becomes inconvenient to change. It might be worthwhile to think about a less verbose way and impl it now.

util/rate_limiter.cc
```diff
@@ -68,17 +68,22 @@ GenericRateLimiter::GenericRateLimiter(
...
+    total_bytes_through_[i] = 0;
+  }
 }
 
 GenericRateLimiter::~GenericRateLimiter() {
   MutexLock g(&request_mutex_);
   stop_ = true;
   requests_to_wait_ = static_cast<int32_t>(queue_[Env::IO_LOW].size() +
-                                           queue_[Env::IO_HIGH].size());
+                                           queue_[Env::IO_HIGH].size() +
```
hx235, line 81: [TODO]
Explicitly iterating container using the key (i.e, pri) like this (and some below) makes it less convenient for future changes of adding new pri.  It might be worthwhile to use lambda function.

util/rate_limiter.cc
```diff
@@ -68,17 +68,22 @@ GenericRateLimiter::GenericRateLimiter(
       prev_num_drains_(0),
       max_bytes_per_sec_(rate_bytes_per_sec),
       tuned_time_(NowMicrosMonotonic()) {
-  total_requests_[0] = 0;
-  total_requests_[1] = 0;
-  total_bytes_through_[0] = 0;
-  total_bytes_through_[1] = 0;
+  for (int i = 0; i < Env::IO_TOTAL; ++i) {
```
hx235, line 71: [TODO]
Iterate from Env::IO_LOW instead of 0

util/rate_limiter.cc
ajkr: So once we have user-pri and mid-pri, are these the new iteration orders?

{user_pri, high_pri, mid_pri, low_pri} with probability (1 - 1/fairness) * (1 - 1/fairness)
{user_pri, high_pri, low_pri, mid_pri} with probability (1 - 1/fairness) * 1/fairness
{user_pri, mid_pri, low_pri, high_pri} with probability 1/fairness * (1 - 1/fairness)
{user_pri, low_pri, mid_pri, high_pri} with probability 1/fairness^2

If user_pri and mid_pri are empty we get high_pri above low_pri at (1 - 1/fairness) * (1 - 1/fairness) + (1 - 1/fairness) * 1/fairness = 1 - 1/fairness. And low_pri above high_pri at 1/fairness * (1 - 1/fairness) + 1/fairness^2 = 1/fairness. So it is compatible.

hx235:
Just a rebase

util/rate_limiter.cc
hx235: >So once we have user-pri and mid-pri, are these the new iteration orders?
Yes - the iteration order and the probability is correct :)

>.. we get high_pri above low_pri at (1 - 1/fairness) * (1 - 1/fairness) + (1 - 1/fairness) * 1/fairness = 1 - 1/fairness. And low_pri above high_pri at 1/fairness * (1 - 1/fairness) + 1/fairness^2 = 1/fairness. So it is compatible.

I agree with the math and the conclusion that it is compatible - just one more note: even when user_pri or mid_pri is not empty, the probability of high_pri above low and low_pri above high stay the same as your calculation. It is because whether user_pri/mid_pri is empty or not is irrelevant to my order generation logic. We generate the order first (as if we are going to have all these pri) and then we iterate the queues according to this order. When a certain pri queue is empty, we just iterate the empty queue as if it is skipped.

(Just in case you wonder, I'd like to think that the reason why we get such a compatible probability result is because we maintain the definition of fairness for a certain priority level P being "one of # times that P will be placed after ALL P's lower pri levels", both in the case where we have only high/low pri or where we have more.)

hx235:
Update:
- Re-proposed for superior user-pri and mid pri
- Improved the previous impl of explicitly iterating priority levels 

Todo:
- Add unit test for changes on user-pri/mid-pri
- Run performance test to verify performance with only high/low requests is not affected by this PR change

util/rate_limiter.cc
hx235: >[TODO]Regardless of which way we are considering the relative order of user_pri to lower pri requests, I need to update the public document and explain how fairness works for user_pri/high_pri/low_pri.

Discussed offline that we are going to go with the current algorithm of deciding order.

>[TODO]The current impl verbosely manipulating pri in generating the order for the benefit of readability. Should we have more pri levels, this verbose impl becomes inconvenient to change. Will consider thinking about a less verbose impl.

Discussed offline that no need for this optimization now since we expect the current amount of priority levels are enough and current verbose impl is good enough for now. If optimization is needed for the future, we can consider rewrite this as recursion.

util/rate_limiter.cc
hx235: Fixed with a loop

util/rate_limiter.cc
hx235: Fixed

hx235:
Update:
local rebase since the Sandcastle auto rebase can't pick up others' changes on Env.h on master

PR #8596, opened 2021-07-29
[better_eng][rate_limiter] Improve rate limiter implementation's readability

hx235:
Context:
As need for new feature of resource management using RocksDB's rate limiter like [#8595](https://github.com/facebook/rocksdb/pull/8595) arises, it is about time to re-learn our rate limiter and make this learning process easier for others by improving its readability. The comment/assertion/one extra else-branch are added based on my best understanding toward the rate_limiter.cc and rate_limiter_test.cc up to date after giving it a hard read. 

Summary:
- Add code comments/assertion/one extra else-branch (that is not affecting existing behavior, see PR comment) to describe how leader-election works under multi-thread settings in GenericRateLimiter::Request()
- Add code comments to describe a non-obvious trick during clean-up of rate limiter destructor
- Add code comments to explain more about the starvation being fixed in GenericRateLimiter::Refill() through partial byte-granting
- Add code comments to the rate limiter's setup in a complicated unit test in rate_limiter_test

Test:
- passed existing rate_limiter_test.cc

### DIFF
--- a/util/rate_limiter.cc
+++ b/util/rate_limiter.cc
@@ -278,7 +278,7 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
       //     request thread does after it was signaled. The new request might
       //     sit at front of a queue and hence become the next-round leader
       //     instead.
-      assert(!timedout);
+      assert(&r != leader_);
     }
   } while (!r.granted);
 }
--- a/util/rate_limiter.cc
+++ b/util/rate_limiter.cc
@@ -118,6 +118,9 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
   }
 
   if (stop_) {
+    // It is now in the clean-up of ~GenericRateLimiter().
+    // Therefore any new incoming request will exit from here
+    // and not get satiesfied.
     return;
   }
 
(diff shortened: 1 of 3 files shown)


util/rate_limiter.cc
```diff
@@ -138,91 +141,150 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
-    // Make sure the waken up request is always the header of its queue
+    // Assertion: request thread running through this point is one of the
+    // following in terms of the request type and quota granting situation:
+    // (1) a leader request that is not fully granted with quota and about
+    //     to carry out its leader's work;
+    // (2) a non-leader request that got fully granted with quota and is
+    //     running to exit;
+    // (3) a non-leader request that is not fully granted with quota and
+    //     is running to particpate in next-round leader election;
+    assert((&r == leader_ && !r.granted) || (&r != leader_ && r.granted) ||
```
hx235, line 208: Note:
This is one of the two code changes in this PR. It should not affect production since code in assert() statements will be excluded from compilation in production builds.

util/rate_limiter.cc
```diff
@@ -138,91 +141,150 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
           if (!queue_[Env::IO_HIGH].empty()) {
             queue_[Env::IO_HIGH].front()->cv.Signal();
           } else if (!queue_[Env::IO_LOW].empty()) {
             queue_[Env::IO_LOW].front()->cv.Signal();
           }
-          // Done
+
+          // The leader request (that was just retired) exits
           break;
+        } else {
```
hx235, line 265: Note:
This extra else-branch is the second of the two code changes in this PR. It is the same as existing behavior since the existing behavior that is evaluated to be false in [L243](https://github.com/facebook/rocksdb/pull/8596/files#diff-64f39d34591bb466f92056637ccb1b88009806eabf366a59eca5b92d27fea2c8R243) will fall out of L267 then L270, which is the same as the new behavior of falling out of L267 then L270

util/rate_limiter.cc
```diff
@@ -138,91 +141,151 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
+          // The leader request (that was just retired) is not fully granted
+          // with quota. It will particpate in leader election and claim back
+          // the leader position immediately.
+          assert(!r.granted);
```
hx235, line 269: Note:
This assertion is the 3rd/(3 code changes) in this PR. It should not affect production since code in assert() statements will be excluded from compilation in production builds.

util/rate_limiter.cc
```diff
@@ -245,7 +308,10 @@ void GenericRateLimiter::Refill() {
     while (!queue->empty()) {
       auto* next_req = queue->front();
       if (available_bytes_ < next_req->request_bytes) {
-        // avoid starvation
+        // Grant partial request_bytes to avoid starvation of requests
```
hx235, line 311: see for info

util/rate_limiter.cc
```diff
@@ -245,7 +307,10 @@ void GenericRateLimiter::Refill() {
...
+        // that become asking for more bytes than available_bytes_
+        // due to dynamically reduced rate limiter's bytes_per_second that
+        // leads to reduced refill_bytes_per_period hence available_bytes_
         next_req->request_bytes -= available_bytes_;
```
ajkr, line 314: It feels kind of odd that we drain all available_bytes_ in Refill() but not when inserting an initial item onto the queue. It seems changing that would make it fairer in terms of giving bytes to earlier requests first. I think it would also allow us to set `available_bytes_ = refill_bytes_per_period` at the start of `Refill()` rather than the quite confusing `available_bytes_ += refill_bytes_per_period`.

ajkr (approved):
LGTM, I might try some follow ups too since now we studied this code in great detail.

util/rate_limiter.cc
ajkr: > It seems changing that would make it fairer in terms of giving bytes to earlier requests first.

It would also allow us to answer this question positively: https://github.com/facebook/rocksdb/blob/ab7f7c9e497a18c48837556bf2e68b4cc5d2a498/util/rate_limiter.cc#L128-L129. I think right now a big request (bytes > available_bytes) can be enqueued, then a small request (bytes <= available_bytes) can be permitted even though somebody's waiting.

util/rate_limiter.cc
ajkr: No need for reply

util/rate_limiter.cc
```diff
@@ -138,91 +141,150 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
+    // (2) a non-leader request that got fully granted with quota and is
+    //     running to exit;
+    // (3) a non-leader request that is not fully granted with quota and
+    //     is running to particpate in next-round leader election;
+    assert((&r == leader_ && !r.granted) || (&r != leader_ && r.granted) ||
+           (&r != leader_ && !r.granted));
+
+    // Assertion: request thread running through this point is one of the
+    // following in terms of its position in queue:
+    // (1) a request got popped off the queue because it is fully granted
```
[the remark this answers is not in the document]
ajkr: No need for reply

util/rate_limiter.cc
```diff
@@ -138,91 +141,150 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
     assert(r.granted ||
            (!queue_[Env::IO_HIGH].empty() &&
             &r == queue_[Env::IO_HIGH].front()) ||
            (!queue_[Env::IO_LOW].empty() &&
             &r == queue_[Env::IO_LOW].front()));
+
+    // Assertion: request thread running through this point is one of the
+    // following in terms of the current leader situation:
+    // (1) there is no leader;
+    // (2) there is a leader and leader sits at the front of its queue;
```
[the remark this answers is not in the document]
ajkr: This isn't relevant if we do #8602, so no need for reply

util/rate_limiter.cc
```diff
@@ -138,91 +141,150 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
            (!queue_[Env::IO_LOW].empty() &&
             leader_ == queue_[Env::IO_LOW].front()));
 
     if (leader_ == &r) {
-      // Waken up from TimedWait()
+      // The leader request thread is now running.
+      // It might or might not has been TimedWait().
       if (timedout) {
-        // Time to do refill!
+        // Time for the leader to do refill and grant bytes to requests
```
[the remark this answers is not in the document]
ajkr: No need for reply

util/rate_limiter.cc
ajkr: This is a suggestion for a follow-up.

ajkr (commented):
btw feel free to proceed, none of my comments here really need attention

util/rate_limiter.cc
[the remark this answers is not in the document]
hx235: Done: renamed this private function and its reference in this class and test since the function name is confusing anyway

util/rate_limiter.cc
hx235: Done: marked it as part of my learning task

hx235:
Update:
Apologize for not getting to these comments as early as I wish - added a couple minor follow-up based on your comments (see commit titles for more info - thank you for sharing your reading experience!!!!). The relatively notable change is renaming the private function Refill() to RefillBytesAndGrantRequests() - reference of this function within/outside the class has been updated. 

Will merge this as soon as all the testing signals are clear.

util/rate_limiter.cc
```diff
@@ -138,98 +141,150 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
-    // (1) a new incoming request,
-    // (2) a previous leader, whose quota has not been not assigned yet due
-    //     to lower priority
-    // (3) a previous waiter at the front of queue, who got notified by
-    //     previous leader
+
+    // Leader election:
+    //  Leader request's duty:
+    //  (1) Waiting for the next refill time;
+    //  (2) Refilling the bytes and granting requests.
```
hx235, line 148: These two lines of comments were originally created by @ajkr in https://github.com/facebook/rocksdb/pull/8602/files#r682014946. They are merged into this PR only because they align with this PR's goal of improving rate limiter's readability with minor code changes and can serve as a good follow-up improvement of this PR.

PR #8602, opened 2021-08-02
Simplify GenericRateLimiter algorithm

ajkr:
`GenericRateLimiter` slow path handles requests that cannot be satisfied
immediately.  Such requests enter a queue, and their thread stays in `Request()`
until they are granted or the rate limiter is stopped.  These threads are
responsible for unblocking themselves.  The work to do so is split into two main
duties.

(1) Waiting for the next refill time.
(2) Refilling the bytes and granting requests.

Prior to this PR, the slow path logic involved a leader election algorithm to
pick one thread to perform (1) followed by (2).  It elected the thread whose
request was at the front of the highest priority non-empty queue since that
request was most likely to be granted.  This algorithm was efficient in terms of
reducing intermediate wakeups, which is a thread waking up only to resume
waiting after finding its request is not granted.  However, the conceptual
complexity of this algorithm was too high.  It took me a long time to draw a
timeline to understand how it works for just one edge case yet there were so
many.

This PR drops the leader election to reduce conceptual complexity.  Now, the two
duties can be performed by whichever thread acquires the lock first.  The risk
of this change is increasing the number of intermediate wakeups, however, we
took steps to mitigate that.

- `wait_until_refill_pending_` flag ensures only one thread performs (1). This\
prevents the thundering herd problem at the next refill time. The remaining\
threads wait on their condition variable with an unbounded duration -- thus we\
must remember to notify them to ensure forward progress.
- (1) is typically done by a thread at the front of a queue. This is trivial\
when the queues are initially empty as the first choice that arrives must be\
the only entry in its queue. When queues are initially non-empty, we achieve\
this by having (2) notify a thread at the front of a queue (preferring higher\
priority) to perform the next duty.
- We do not require any additional wakeup for (2). Typically it will just be\
done by the thread that finished (1).

Combined, the second and third bullet points above suggest the refill/granting
will typically be done by a request at the front of its queue.  This is
important because one wakeup is saved when a granted request happens to be in an
already running thread.

Note there are a few cases that still lead to intermediate wakeup, however.  The
first two are existing issues that also apply to the old algorithm, however, the
third (including both subpoints) is new.

- No request may be granted (only possible when rate limit dynamically\
decreases).
- Requests from a different queue may be granted.
- (2) may be run by a non-front request thread causing it to not be granted even\
if some requests in that same queue are granted. It can happen for a couple\
(unlikely) reasons.
  - A new request may sneak in and grab the lock at the refill time, before the\
thread finishing (1) can wake up and grab it.
  - A new request may sneak in and grab the lock and execute (1) before (2)'s\
chosen candidate can wake up and grab the lock. Then that non-front request\
thread performing (1) can carry over to perform (2).

Test Plan:

- Use existing tests. The edge cases listed in the comment are all performance\
related; I could not really think of any related to correctness. The logic\
looks the same whether a thread wakes up/finishes its work early/on-time/late,\
or whether the thread is chosen vs. "steals" the work.
- Verified write throughput and CPU overhead are basically the same with and\
  without this change, even in a rate limiter heavy workload:

Test command:
```
$ rm -rf /dev/shm/dbbench/ && TEST_TMPDIR=/dev/shm /usr/bin/time ./db_bench -benchmarks=fillrandom -num_multi_db=64 -num_low_pri_threads=64 -num_high_pri_threads=64 -write_buffer_size=262144 -target_file_size_base=262144 -max_bytes_for_level_base=1048576 -rate_limiter_bytes_per_sec=16777216 -key_size=24 -value_size=1000 -num=10000 -compression_type=none -rate_limiter_refill_period_us=1000
```

Results before this PR:

```
fillrandom   :     108.463 micros/op 9219 ops/sec;    9.0 MB/s
7.40user 8.84system 1:26.20elapsed 18%CPU (0avgtext+0avgdata 256140maxresident)k
```

Results after this PR:

```
fillrandom   :     108.108 micros/op 9250 ops/sec;    9.0 MB/s
7.45user 8.23system 1:26.68elapsed 18%CPU (0avgtext+0avgdata 255688maxresident)k
```

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -1215,6 +1215,10 @@ DEFINE_int32(rate_limit_delay_max_milliseconds, 1000,
 
 DEFINE_uint64(rate_limiter_bytes_per_sec, 0, "Set options.rate_limiter value.");
 
+DEFINE_int64(rate_limiter_refill_period_us, 100 * 1000,
+             "Set refill period on "
+             "rate limiter.");
+
 DEFINE_bool(rate_limiter_auto_tuned, false,
             "Enable dynamic adjustment of rate limit according to demand for "
             "background I/O");
@@ -4440,7 +4444,7 @@ class Benchmark {
         exit(1);
       }
       options.rate_limiter.reset(NewGenericRateLimiter(
-          FLAGS_rate_limiter_bytes_per_sec, 100 * 1000 /* refill_period_us */,
+          FLAGS_rate_limiter_bytes_per_sec, FLAGS_rate_limiter_refill_period_us,
           10 /* fairness */,
           FLAGS_rate_limit_bg_reads ? RateLimiter::Mode::kReadsOnly
                                     : RateLimiter::Mode::kWritesOnly,
--- a/util/rate_limiter.cc
+++ b/util/rate_limiter.cc
@@ -196,7 +196,8 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
         ++num_drains_;
         wait_until_refill_pending_ = true;
         r.cv.TimedWait(wait_until);
-        TEST_SYNC_POINT("GenericRateLimiter::Request:PostTimedWait");
+        TEST_SYNC_POINT_CALLBACK("GenericRateLimiter::Request:PostTimedWait",
+                                 &time_until_refill_us);
         wait_until_refill_pending_ = false;
       }
     } else {
--- a/util/rate_limiter.cc
+++ b/util/rate_limiter.cc
@@ -62,7 +62,7 @@ GenericRateLimiter::GenericRateLimiter(
       next_refill_us_(NowMicrosMonotonic()),
       fairness_(fairness > 100 ? 100 : fairness),
       rnd_((uint32_t)time(nullptr)),
-      leader_(nullptr),
+      waiting_(false),
       auto_tuned_(auto_tuned),
       num_drains_(0),
       prev_num_drains_(0),
@@ -139,148 +139,107 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
   Req r(bytes, &request_mutex_);
   queue_[pri].push_back(&r);
 
+  // A thread representing a queued request coordinates with other such threads.
+  // There are two main duties.
+  //
+  // (1) Waiting for the next refill time.
+  // (2) Refilling the bytes and granting requests.
+  //
+  // To simplify the implementation, the duties can be performed by any thread.
+  // (1) could in theory be performed by all threads, whereas (2) can only be
+  // performed by one thread per refill interval while holding the mutex. (2)
+  // may even be performed by a thread not involved in (1).
+  //
+  // We restrict the flexibility a bit to reduce unnecessary wakeups:
+  //
+  // - `waiting_` flag ensures only one thread performs (1). This prevents the
+  //   thundering herd problem at the next refill time. The remaining threads
+  //   wait on their condition variable with an unbounded duration -- thus we
+  //   must remember to notify them to ensure forward progress.
+  // - (1) is typically done by a thread at the front of a queue. This is
+  //   trivial when the queues are initially empty as the first choice that
+  //   arrives must be the only entry in its queue. When queues are initially
+  //   non-empty, we achieve this by having (2) notify a thread at the front
+  //   of a queue (preferring higher priority) to perform the next duty.
+  // - We do not require any additional wakeup for (2). Typically it will just
+  //   be done by the thread that finished (1).
+  //
+  // Combined, the second and third bullet points above suggest the refill/
+  // granting will typically be done by a request at the front of its queue.
+  // This is important because one wakeup is saved when a granted request
+  // happens to be in an already running thread.
+  //
+  // Note this nice property is not guaranteed in a few cases, however.
+  //
+  // - No request may be granted.
+  // - Requests from a different queue may be granted.
+  // - (2) may be run by a non-front request thread causing it to not be granted
+  //   even if some requests in that same queue are granted. It can happen for a
+  //   couple (unlikely) reasons.
+  //    - A new request may sneak in and grab the lock at the refill time,
+  //      before the thread finishing (1) can wake up and grab it.
+  //    - A new request may sneak in and grab the lock and execute (1) before
+  //      (2)'s chosen candidate can wake up and grab the lock. Then that non-
+  //      front request thread performing (1) can carry over to perform (2).
   do {
-    bool timedout = false;
-
-    // Leader election:
-    //  Leader request's duty:
-    //  (1) Waiting for the next refill time;
-    //  (2) Refilling the bytes and granting requests.
-    //
-    //  If the following three conditions are all true for a request,
-    //  then the request is selected as a leader:
-    //  (1) The request thread acquired the request_mutex_ and is running;
-    //  (2) There is currently no leader;
-    //  (3) The request sits at the front of a queue.
-    //
-    //  If not selected as a leader, the request thread will wait
-    //  for one of the following signals to wake up and
-    //  compete for the request_mutex_:
-    //  (1) Signal from the previous leader to exit since its requested bytes
-    //      are fully granted;
-    //  (2) Signal from the previous leader to particpate in next-round
-    //      leader election;
-    //  (3) Signal from rate limiter's destructor as part of the clean-up.
-    //
-    //  Therefore, a leader request can only be one of the following types:
-    //  (1) a new incoming request placed at the front of a queue;
-    //  (2) a previous leader request whose quota has not been not fully
-    //      granted yet due to its lower priority, hence still at
-    //      the front of a queue;
-    //  (3) a waiting request at the front of a queue, which got
-    //      signaled by the previous leader to participate in leader election.
-    if (leader_ == nullptr &&
-        ((!queue_[Env::IO_HIGH].empty() &&
-            &r == queue_[Env::IO_HIGH].front()) ||
-         (!queue_[Env::IO_LOW].empty() &&
-            &r == queue_[Env::IO_LOW].front()))) {
-      leader_ = &r;
-
-      int64_t delta = next_refill_us_ - NowMicrosMonotonic();
-      delta = delta > 0 ? delta : 0;
-      if (delta == 0) {
-        timedout = true;
+    int64_t time_until_refill_us = next_refill_us_ - NowMicrosMonotonic();
+    if (time_until_refill_us > 0) {
+      if (waiting_) {
+        // Somebody is performing (1). Trust we'll be woken up when our request
+        // is granted or we are needed for future duties.
+        r.cv.Wait();
       } else {
-        // The leader request thread waits till next_refill_us_
-        int64_t wait_until = clock_->NowMicros() + delta;
+        // Whichever thread reaches here first performs duty (1) as described
+        // above.
         RecordTick(stats, NUMBER_RATE_LIMITER_DRAINS);
         ++num_drains_;
-        timedout = r.cv.TimedWait(wait_until);
+        waiting_ = true;
+        r.cv.TimedWait(time_until_refill_us);
+        waiting_ = false;
       }
     } else {
-      r.cv.Wait();
-    }
-
-    if (stop_) {
-      // It is now in the clean-up of ~GenericRateLimiter().
-      // Therefore any woken-up request will exit here,
-      // might or might not has been satiesfied.
-      --requests_to_wait_;
-      exit_cv_.Signal();
-      return;
-    }
-
-    // Assertion: request thread running through this point is one of the
-    // following in terms of the request type and quota granting situation:
-    // (1) a leader request that is not fully granted with quota and about
-    //     to carry out its leader's work;
-    // (2) a non-leader request that got fully granted with quota and is
-    //     running to exit;
-    // (3) a non-leader request that is not fully granted with quota and
-    //     is running to particpate in next-round leader election.
-    assert((&r == leader_ && !r.granted) || (&r != leader_ && r.granted) ||
-           (&r != leader_ && !r.granted));
-
-    // Assertion: request thread running through this point is one of the
-    // following in terms of its position in queue:
-    // (1) a request got popped off the queue because it is fully granted
-    //     with bytes;
-    // (2) a request sits at the front of its queue.
-    assert(r.granted ||
-           (!queue_[Env::IO_HIGH].empty() &&
-            &r == queue_[Env::IO_HIGH].front()) ||
-           (!queue_[Env::IO_LOW].empty() &&
-            &r == queue_[Env::IO_LOW].front()));
-
-    if (leader_ == &r) {
-      // The leader request thread is now running.
-      // It might or might not has been TimedWait().
-      if (timedout) {
-        // Time for the leader to do refill and grant bytes to requests
-        RefillBytesAndGrantRequests();
-
-        // The leader request retires after refilling and granting bytes
-        // regardless. This is to simplify the election handling.
-        leader_ = nullptr;
-
-        if (r.granted) {
-          // The leader request (that was just retired)
-          // already got fully granted with quota and will soon exit
-
-          // Assertion: the fully granted leader request is popped off its queue
-          assert((queue_[Env::IO_HIGH].empty() ||
-                    &r != queue_[Env::IO_HIGH].front()) &&
-                 (queue_[Env::IO_LOW].empty() ||
-                    &r != queue_[Env::IO_LOW].front()));
-
-          // If there is any remaining requests, the leader request (that was
-          // just retired) makes sure there exists at least one leader candidate
-          // by signaling a front request of a queue to particpate in
-          // next-round leader election
-          if (!queue_[Env::IO_HIGH].empty()) {
-            queue_[Env::IO_HIGH].front()->cv.Signal();
-          } else if (!queue_[Env::IO_LOW].empty()) {
-            queue_[Env::IO_LOW].front()->cv.Signal();
-          }
-
-          // The leader request (that was just retired) exits
-          break;
-        } else {
-          // The leader request (that was just retired) is not fully granted
-          // with quota. It will particpate in leader election and claim back
-          // the leader position immediately.
-          assert(!r.granted);
+      // Whichever thread reaches here first performs duty (2) as described
+      // above.
+      RefillBytesAndGrantRequests();
+      if (r.granted) {
+        // If there is any remaining requests, make sure there exists at least
+        // one candidate is awake for future duties by signaling a front request
+        // of a queue.
+        // TODO(ajkr): we may wish to re-select no matter what to prevent the
+        // case we get unlucky with a race condition and then stuck working in
+        // the same non-grantable request thread for a while.
+        if (!queue_[Env::IO_HIGH].empty()) {
+          queue_[Env::IO_HIGH].front()->cv.Signal();
+        } else if (!queue_[Env::IO_LOW].empty()) {
+          queue_[Env::IO_LOW].front()->cv.Signal();
         }
-      } else {
-        // Spontaneous wake up, need to continue to wait
-        assert(!r.granted);
-        leader_ = nullptr;
       }
-    } else {
-      // The non-leader request thread is running.
-      // It is one of the following request types:
-      // (1) The request got fully granted with quota and signaled to run to
-      //     exit by the previous leader;
-      // (2) The request is not fully granted with quota and signaled to run to
-      //     particpate in next-round leader election by the previous leader.
-      //     It might or might not become the next-round leader because a new
-      //     request may come in and acquire the request_mutex_ before this
-      //     request thread does after it was signaled. The new request might
-      //     sit at front of a queue and hence become the next-round leader
-      //     instead.
-      assert(&r != leader_);
     }
-  } while (!r.granted);
+    // Invariant: non-granted request is always in one queue, and granted
+    // request is always in zero queues.
+#ifndef NDEBUG
+    int num_found = 0;
+    for (int i = Env::IO_LOW; i < Env::IO_TOTAL; ++i) {
+      if (std::find(queue_[i].begin(), queue_[i].end(), &r) !=
+          queue_[i].end()) {
+        ++num_found;
+      }
+    }
+    if (r.granted) {
+      assert(num_found == 0);
+    } else {
+      assert(num_found == 1);
+    }
+#endif  // NDEBUG
+  } while (!stop_ && !r.granted);
+
+  if (stop_) {
+    // It is now in the clean-up of ~GenericRateLimiter().
+    // Therefore any woken-up request will have come out of the loop and then
+    // exit here. It might or might not have been satisfied.
+    --requests_to_wait_;
+    exit_cv_.Signal();
+  }
 }
 
 void GenericRateLimiter::RefillBytesAndGrantRequests() {
@@ -314,10 +273,8 @@ void GenericRateLimiter::RefillBytesAndGrantRequests() {
       queue->pop_front();
 
       next_req->granted = true;
-      if (next_req != leader_) {
-        // Quota granted, signal the thread to exit
-        next_req->cv.Signal();
-      }
+      // Quota granted, signal the thread to exit
+      next_req->cv.Signal();
     }
   }
 }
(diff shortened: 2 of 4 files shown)


util/rate_limiter.cc
```diff
@@ -136,96 +136,85 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
   Req r(bytes, &request_mutex_);
   queue_[pri].push_back(&r);
 
+  // A thread representing a queued request coordinates with other such threads.
+  // There are two main duties.
+  //
+  // (1) Waiting for the next refill time.
+  // (2) Refilling the bytes and granting requests.
```
hx235, line 143: Very useful comment about the heavy-working duty, regardless of whether we want to keep the concept of leader or not in our impl. I integrated these two comment lines in the readability PR because it helps improve the readability without any big code changes, aligning with the readability PR []. I credited you in here [], but do let me know this credit does not change the fact that I was a thief of good comments!

ajkr: Yes please take whatever's useful

util/rate_limiter.cc
```diff
@@ -136,96 +136,85 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
+  //
+  // To simplify the implementation, the duties can be performed by any thread.
```
hx235, line 145: Just to understand the "simplification" effects in "coding", does the "simplification" result in the following major coding changes:
(1) Remove the non-leader thread else-branch on [L218](https://github.com/facebook/rocksdb/pull/8602/files#diff-64f39d34591bb466f92056637ccb1b88009806eabf366a59eca5b92d27fea2c8L218) 
(2) Cut down the "Spontaneous wake up" else-branch on [L213](https://github.com/facebook/rocksdb/pull/8602/files#diff-64f39d34591bb466f92056637ccb1b88009806eabf366a59eca5b92d27fea2c8L213)
(3) Facilitate the destructor clean up (i.e, if (stop_) {...} outside of the while loop on [L139](https://github.com/facebook/rocksdb/pull/8602/files#diff-64f39d34591bb466f92056637ccb1b88009806eabf366a59eca5b92d27fea2c8L139)

Did I miss any other major coding changes (i.e, non-assertion change) the simplification results in?

hx235:
Update: will resume the review in a couple hours.

util/rate_limiter.cc
ajkr: I can change the claim to say "reduce conceptual complexity" to make it easier to reason about edge cases, change/extend, etc. There isn't any particular line in the old code that's exciting to remove; it's more about how long it takes to convince myself it works in all cases. In the old code it took me a really long time to draw a timeline to understand how it works for just one edge case yet there were so many.

hx235: I agree that conceptual complexity in reasoning about the timeline of different threads and edge cases is the pain point of the previous impl, which is also why we need this PR.

Suggestion: let's add the "why" we need to simplify the previous algorithm in the PR summary too.

hx235: > imo the main simplification to the logic is there is no choice of leader anymore. 
I think the key reason why we can do this is - the constraint of a leader (i.e, the heavy-lifting thread) being the front request in the queue, which is the only leader constraint being relaxed in this PR, is unnecessary for correctness. The reason is that the correctness in our Refill() means - if there is any ungranted request, we make sure there exists one and only one thread to refill the bytes and grant requests when time is right. None of these requires the front-queued constraint.

In addition to that simplification, I will also argue the current coding style of separating the code paths needed to be read by readers in order to trace different threads is also an improvement to simplicity and readability.
 
Suggestion: 
Let's add a PR summary of "what" the simplification are and "why" we can simplify in this way without violate the correctness. 
We can also mention losing the potential benefit of saving one wake-up through leader being the front request, as you have described, is acceptable.

util/rate_limiter_test.cc
```diff
@@ -213,7 +213,7 @@ TEST_F(RateLimiterTest, AutoTuneIncreaseWhenFull) {
   // computes the next refill time (ensuring refill time in the future allows
   // the next request to drain the rate limiter).
   ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->SetCallBack(
-      "GenericRateLimiter::Refill", [&](void* /*arg*/) {
+      "GenericRateLimiter::Request:PostTimedWait", [&](void* /*arg*/) {
```
hx235, line 216: I haven't gotten quite deep into how auto-tune works exactly so I need a quick context "refilling" in order to understand this change. 

Why we decide to change the call back position while still achieving the goals of "to advance (1) after Request()
has determined the bytes are not available; and (2) before Refill() computes the next refill time" mentioned in the comment above?

util/rate_limiter.cc
```diff
@@ -136,96 +136,85 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
+        r.cv.Wait();
       } else {
-        int64_t wait_until = clock_->NowMicros() + delta;
+        // Whichever thread reaches here first performs duty (1) as described
+        // above.
         RecordTick(stats, NUMBER_RATE_LIMITER_DRAINS);
         ++num_drains_;
-        timedout = r.cv.TimedWait(wait_until);
+        waiting_ = true;
+        r.cv.TimedWait(clock_->NowMicros() + time_until_refill_us);
```
hx235, line 194: Maybe just me not seeing it - why the previous impl is a bug?

hx235:
Update
Done:
- understood the what/why/why-we-can of the simplification proposed by this PR (left some thoughts in the comments)
- left suggestion about PR summary
- left clarification questions for non-algorithmic changes

TODO:
- take another look for any concurrency issue (e.g, acceptable data race and non-acceptable data race)
- see if additional testing is needed
- take a closer read into the comment about the current implementation to see whether some of them (1) can be moved to PR summary (2)need further clarification

ajkr (commented):
Thanks for the careful review so far!!

ajkr:
The latest update is just for rebase - I will address the comments in a later update.

hx235:
> The latest update is just for rebase - I will address the comments in a later update.

Thanks! I should be able to get the rest of my comment in by EOD today.

util/rate_limiter.cc
```diff
@@ -139,148 +139,105 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
+  //
+  // (1) Waiting for the next refill time.
+  // (2) Refilling the bytes and granting requests.
+  //
+  // To simplify the implementation, the duties can be performed by any thread.
+  // (1) could in theory be performed by all threads, whereas (2) can only be
+  // performed by one thread per refill interval while holding the mutex. (2)
+  // may even be performed by a thread not involved in (1).
+  //
+  // We restrict the flexibility a bit to reduce unnecessary wakeups:
```
hx235, line 153: I am guessing that the paragraph [L153-165] starting from this line is trying to argue that this implementation will reduce "wasted wake-up" by the 3 bullet-points that followed, since we know from (2) that only 1 thread out of that many woken up threads can proceed and all the other threads have to go back to sleep (aka, wasted wake-up).

Question 1: am I getting the intention of this paragraph right?

Suggestion (ignore the following if a big "NO" to Question 1):
If "YES" to Question 1, I feel like this paragraph should be moved out of the code comment to "design document" (e.g, PR summary or PR comment section).

The reason is that, solely from my own experience as a code reader, I anticipate people who read the code comment in this file, most likely, is trying to get extra help from comment to follow the code, instead of pondering why the code is designed in this way. When they do have questions on why the code is designed in this way (e.g, wondering why we only have one thread doing the timedwait, or deeper subtle question on whether the heavy-lifting thread is always at the front of the queue), they will probably look after PR summary, design document or ask POC. 

My intention for this suggestion is to make the comment section in the file to focus on helping people to follow the code, instead of potentially overwhelming them too much with things they can't really think about when they are still in the process of trying to follow the code.

But I am also aware of the suggestion and reasoning above are solely based on my own experience of reading code comment for the purpose of better following the code. If it does not align with most of RocksDB code reader's reading habit that you know of and keeping the paragraph in the code is better according to that habit, let me know :)



Question: What exactly does the "unnecessary wakeups" refer to? Does it refelyer to the following cases?
(1)Wakes up of more than 1 threads from timedwait during the same refill interval, if we allowed multiple threads doing that timedwait w/o using 'waiting_'flag
(2)Wake up of a thread without its request being granted

ajkr, line 153: > I am guessing that the paragraph [L153-165] starting from this line is trying to argue that this implementation will reduce "wasted wake-up" by the 3 bullet-points that followed, since we know from (2) that only 1 thread out of that many woken up threads can proceed and all the other threads have to go back to sleep (aka, wasted wake-up).
>
> Question 1: am I getting the intention of this paragraph right?

Yes, right.

> Suggestion (ignore the following if a big "NO" to Question 1):
> If "YES" to Question 1, I feel like this paragraph should be moved out of the code comment to "design document" (e.g, PR summary or PR comment section).
>
> The reason is that, solely from my own experience as a code reader, I anticipate people who read the code comment in this file, most likely, is trying to get extra help from comment to follow the code, instead of pondering why the code is designed in this way. When they do have questions on why the code is designed in this way (e.g, wondering why we only have one thread doing the timedwait, or deeper subtle question on whether the heavy-lifting thread is always at the front of the queue), they will probably look after PR summary, design document or ask POC.
>
> My intention for this suggestion is to make the comment section in the file to focus on helping people to follow the code, instead of potentially overwhelming them too much with things they can't really think about when they are still in the process of trying to follow the code.

It's a good point, except the part about people asking POC (code author?) because I'll never remember :). Let me try moving comments about how each technique reduces wakeups to be in-line with the implementation of those techniques and see how it reads.

> But I am also aware of the suggestion and reasoning above are solely based on my own experience of reading code comment for the purpose of better following the code. If it does not align with most of RocksDB code reader's reading habit that you know of and keeping the paragraph in the code is better according to that habit, let me know :)
>
> Question: What exactly does the "unnecessary wakeups" refer to? Does it refelyer to the following cases?
> (1)Wakes up of more than 1 threads from timedwait during the same refill interval, if we allowed multiple threads doing that timedwait w/o using 'waiting_'flag
> (2)Wake up of a thread without its request being granted

Hm I didn't have an exact definition before. Let's say the optimal total number of wakeups is just the number of threads that ever went to sleep. Then (2) is an unnecessary wakeup. (1) (waking up for the purpose of doing refill only to find someone else already did it) would usually lead to (2), though not always (the request coincidentally have been satisfied by the thread that did the refill).

util/rate_limiter.cc
hx235: Suggestion:
How about we move the paragraph from L148 to 183 out of the code comment section and structure them into the PR summary? 

(1) For example, L148-L152 addresses "why" we can simplify the existing algorithm, echoing back to my previous suggestion
> ... and "why" we can simplify in this way without violating the correctness?

(2) L153-163 addresses some code design details on how we reduce unnecessary wake-ups since we know from L149-150 that there could potentially be many wasted wake-ups due to only one of them will hold the mutex and proceed.

(3) L167-170 addresses some code design details on the "front-queued" property and its benefit. Moving these lines to PR summary also helps describe a change in this PR (where "front-queued" property typically holds but not always) from existing algorithm (where the "front-queued" property always holds). 

It also naturally leads to the argument you have in PR that the change is acceptable based on your performance result.

(4) I am a bit confused on some part from L172-183 - see my question along with the suggestion for these lines below.

The reason why these lines can be moved out of the code comment is that, IMO, L142-146 and your re-written version of the loop starting from L184 is self-explanatory. Even for a cautious reader tempting to hand-trace different threads to prove the correctness of the program by him/herself, the current version of impl doesn't seem to be too hard for them to do so. Even it ends up being hard, what they probably need is an illustration of possible timelines consisting of different threads based on different cases in the loop, instead of L148 to 183.

util/rate_limiter.cc
```diff
@@ -139,148 +139,105 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
+  //   of a queue (preferring higher priority) to perform the next duty.
+  // - We do not require any additional wakeup for (2). Typically it will just
+  //   be done by the thread that finished (1).
+  //
+  // Combined, the second and third bullet points above suggest the refill/
+  // granting will typically be done by a request at the front of its queue.
+  // This is important because one wakeup is saved when a granted request
+  // happens to be in an already running thread.
+  //
+  // Note this nice property is not guaranteed in a few cases, however.
```
hx235, line 172: Question:
Is the "nice property" mentioned here referring to the refill/granting will typically (a)"be done by a request at the front of its queue" OR (b) "one wakeup is saved when a granted request happens to be in an already running thread"?

(a)(b) don't seem to be exactly the same to me since even in the existing impl where (a) is always the case, (b) is not always true for the same reason you mentioned in L175.

And it seems to me that L179-183 describes when (a) is not guaranteed while L175 describes when (b) is not guaranteed.

Did I miss anything here?

hx235: Suggestion: 
I feel that L172-183 can be moved to PR summary as coding design details, instead of in the coding comment. The reason is:
(1) for new people who are not exposed to the existing impl and try to follow the code here, I doubt whether they will wonder too much about whether the heavy-lifting thread is front-queued and whether the performance gain from the "front-queued" property always exists. 
(2) for people who are exposed to the existing impl and wonder about the change on "front-queued" property, I anticipate that they would go after git blame and PR summary instead.

util/rate_limiter.cc
```diff
@@ -139,148 +139,105 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
+  //
+  // - No request may be granted.
```
hx235, line 174: Question:
Given that every request's byte size is smaller than the refilled byte size per interval, is there any case where no request is granted, other than the case that there is no request coming in?

util/rate_limiter.cc
```diff
@@ -139,148 +139,105 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
-        // Spontaneous wake up, need to continue to wait
-        assert(!r.granted);
-        leader_ = nullptr;
       }
+    }
+    // Invariant: non-granted request is always in one queue, and granted
+    // request is always in zero queues.
+#ifndef NDEBUG
+    int num_found = 0;
+    for (int i = Env::IO_LOW; i < Env::IO_TOTAL; ++i) {
```
hx235, line 220: Good coding habit of starting from Env::IO_LOW instead of 0!

util/rate_limiter.cc
```diff
@@ -139,148 +139,105 @@ void GenericRateLimiter::Request(int64_t bytes, const Env::IOPriority pri,
...
-            &r == queue_[Env::IO_LOW].front()))) {
-      leader_ = &r;
-
-      int64_t delta = next_refill_us_ - NowMicrosMonotonic();
-      delta = delta > 0 ? delta : 0;
-      if (delta == 0) {
-        timedout = true;
+    int64_t time_until_refill_us = next_refill_us_ - NowMicrosMonotonic();
+    if (time_until_refill_us > 0) {
+      if (waiting_) {
```
hx235, line 187: Suggestion (with bias): 
I have a bias toward verbose variable naming and I myself might name "waiting_" something like "exists_waiting_thread_for_refill_time_"

util/rate_limiter.cc
hx235: Got you - yeah, I was looking at the commit history for the reason of this code change.  I personally prefer the option of revering a non-essential change like this, just to avoid potential future confusion when looking at this specific code change in this PR.

Question: I wonder - can we do "r.cv.TimedWait(next_refill_us_);" for a better semantics?

tools/db_bench_tool.cc
```diff
@@ -4437,7 +4441,7 @@ class Benchmark {
         exit(1);
       }
       options.rate_limiter.reset(NewGenericRateLimiter(
-          FLAGS_rate_limiter_bytes_per_sec, 100 * 1000 /* refill_period_us */,
+          FLAGS_rate_limiter_bytes_per_sec, FLAGS_rate_limiter_refill_period_us,
```
hx235, line 4444: I didn't even think of we can do this when I left the comment - a new thing learnt! 👍

util/rate_limiter.cc
hx235: Hi @ajkr - I am so sorry that, due to me mistakenly clicked "add single comment" too early, https://github.com/facebook/rocksdb/pull/8602/files#r683913267 should be my suggestion about moving things to PR summary instead. It contains more actionable items than the one you are replying to but the big ideas are the same.

Sorry about having you duplicate your time reading my comments!

hx235: > Let's say the optimal total number of wakeups is just the number of threads that ever went to sleep. Then (2) is an unnecessary wakeup.

Is it because - if (2) actually happens, the thread usually can't proceed and will be put back to sleep again (making it unnecessary and the total number of wakeups 1 greater than the optimal total number of wakes up)? I was just trying to guess the "unnecessary wakeup" means "the wake up that is then put back to sleep again".

I feel that we might need to be clearer on "unnecessary wakeups" if we decide to dedicate the whole paragraph from L155 to 165 to describe our restriction to these "unnecessary wakeups" because we feel that is important. The paragraph might unfortunately end up being not so strong if the "unnecessary wakeups" is left un-clear to readers.

util/rate_limiter_test.cc
hx235: Thank you for explaining what the new condition should be based on the impl change in this PR! 
> Unfortunately TimedWait() doesn't seem to advance the time in this mock time thing, though that may avoid the need for callback here entirely.

Feel free to either include the change to the callback (if the callback can be avoid) or open a task for me as part of the auto-tune learning process, IF some investigation into the issue "TimedWait() doesn't seem to advance the time in this mock time thing" is needed, not urgent and not blocking this PR.

If changing from "GenericRateLimiter::RefillBytesAndGrantRequests" to "GenericRateLimiter::Request:PostTimedWait" is enough, then I will really appreciate an update to the comment, since for a n00b like me who learns about auto-tune through reading this test case, unfortunately, can easily get hung up on the comment. 

Let me know how you would like to proceed on this!

util/rate_limiter.cc
ajkr: > Is it because - if (2) actually happens, the thread usually can't proceed and will be put back to sleep again (making it unnecessary and the total number of wakeups 1 greater than the optimal total number of wakes up)? I was just trying to guess the "unnecessary wakeup" means "the wake up that is then put back to sleep again".

Yes that is right. Now that we analyzed in detail, "unnecessary" was not an accurate enough word choice for referring to threads that may be doing useful work, like granting other threads' requests. Maybe a better term is "intermediate wakeup".

> Thanks!! If these techniques read better in-line, then they don't really need to be moved to PR summary as I suggested in https://github.com/facebook/rocksdb/pull/8602/files#r683913267

I'll move them to the summary and then duplicate anything inline that feels important to readers (hopefully nothing)

util/rate_limiter.cc
ajkr: > Question: I wonder - can we do "r.cv.TimedWait(next_refill_us_);" for a better semantics?

Not sure, `next_refill_us_` gets computed using `NowMicrosMonotonic()` while the existing code was specifying a time calculated from `NowMicros()`. I haven't learned what is the difference so stuck with what was there.

util/rate_limiter.cc
ajkr: Hm let me try something like  `wait_for_refill_pending_`

util/rate_limiter.cc
ajkr: I think no request coming in means nobody will even do refill/grant. The edge case I was thinking about was after rate limit is dynamically adjusted downwards.

util/rate_limiter.cc
ajkr: > Question:
> Is the "nice property" mentioned here referring to the refill/granting will typically (a)"be done by a request at the front of its queue" OR (b) "one wakeup is saved when a granted request happens to be in an already running thread"?

Let's see, I meant to refer to (b) (woken up thread granting itself).

> (a)(b) don't seem to be exactly the same to me since even in the existing impl where (a) is always the case, (b) is not always true for the same reason you mentioned in L175.
> 
> And it seems to me that L179-183 describes when (a) is not guaranteed while L175 describes when (b) is not guaranteed.
>
> Did I miss anything here?

I am sure it is not written in the clearest way, but L176-178 (the point introducing the sub-points in L179-L183) intends to exclude cases where (a) is true but (b) isn't. In particular the condition "some requests in that same queue are granted" intends to exclude the L175 case.

util/rate_limiter.cc
hx235: Got you - then I am good with keeping your changes r.cv.TimedWait(clock_->NowMicros() + time_until_refill_us);.OR reverting this non-essential change. They don't seem much difference to me in correctness and reverting this non-essential is only for future reader of this PR.

util/rate_limiter.cc
hx235: Good with me too!

util/rate_limiter.cc
hx235: > Let's see, I meant to refer to (b) (woken up thread granting itself).
Oh - now I can see how the following paragraph describes cases (in order) where (b) is not guaranteed! I think the paragraph L174-183 is clear after (b) is made clear to be the "nice property" in L172. I was a bit confused only because (a) can also be seen a nice property and I ended up reading (b) as a consequence of the nice property (a).

Suggestion (with bias): 
Would you mind adding a few words like - "Note this nice property of woken up thread granting itself ... " just to specify a little bit more?

hx235:
Review Summary

Done:
- Verified correctness of the new algorithm conceptually with my best efforts
    - By correctness, I mean: for every enqueued un-granted request, there exists an refill_interval where the request will be granted in a way that is subjected to rate limiting constraint (i.e, constraint on refill_byte_per_period and priority)
- Hand-traced as many execution orders as I could based on the structure of the new algorithm and verified that these execution orders do not violate correctness
- Left comments on the in-code comments and responded to reply from last review

TODO:
@ajkr - Other than suggestions related to structuring L148 to 183 into PR summary or in-line wherever you see fit (i.e, those marked with "suggestion") and a minor clarification [here](https://github.com/facebook/rocksdb/pull/8602/files#r684065178), the only coding related comment is [this](https://github.com/facebook/rocksdb/pull/8602/files#r683942787). Let me know if any further clarification/feedback is needed.

All others look good to me! Especially, based on my experience with learning both impl, compared with the existing impl, your simplified algorithm captures more explicitly an essential property to the correctness of our rate limiter algorithm (i.e, refill when time is right) and removes correctness-irrelevant property that was maintained in the leader election process of the existing impl (i.e, the front-queued property). This has made the code significantly shorter and the multi-thread code flow easier to draw. Good work 👍

hx235 (approved):
Review Summary

Done:
- Verified correctness of the new algorithm conceptually with my best efforts
   - By correctness, I mean: for every enqueued un-granted request, there exists an refill_interval where the request will be granted in a way that is subjected to rate limiting constraint (i.e, constraint on refill_byte_per_period and priority)
- Hand-traced as many execution orders as I could based on the structure of the new algorithm and verified that these execution orders do not violate correctness
- Left comments on the in-code comments and responded to reply from last review

TODO:
@ajkr - Other than suggestions related to structuring L148 to 183 into PR summary or in-line wherever you see fit (i.e, those marked with "suggestion") and a minor clarification [here](https://github.com/facebook/rocksdb/pull/8602#discussion_r684065178), the only coding related comment is [this](https://github.com/facebook/rocksdb/pull/8602#discussion_r683942787). Let me know if any further clarification/feedback is needed.

All others look good to me! Especially, based on my experience with learning both impl, compared with the existing impl, your simplified algorithm captures more explicitly an essential property to the correctness of our rate limiter algorithm (i.e, refill when time is right) and removes a correctness-irrelevant property that was maintained in the leader election process of the existing impl (i.e, the front-queued property of heavy-lifting thread). This has made the code significantly shorter and the multi-threaded code flow easier to draw. Good work 👍

ajkr:
The comments should all be addressed. Thanks for the great review!

util/rate_limiter_test.cc
ajkr: I only updated the comment. It looks like `TimedWait()` in tests (where we prefer to simulate wait rather than actually wait) is in a pretty confusing/inconsistent state so would take a long time to fix.

util/rate_limiter.cc
ajkr: Reverted.

util/rate_limiter_test.cc
hx235: Got you! I will keep that in mind when learning auto-tuning and reading the test case here.

Issue #8535, opened 2021-07-15
Blob files stored first cf path could cause disk full


zaorangyang:
Currently, the Blob files path are forced to the first CF path. In general, when users specify multiple CF paths, the first path could have less space. Forcing Blob data to write to the first path could easily cause the path full and then cause IO errors.
https://github.com/facebook/rocksdb/blob/803a40d4123e8447a4b271ff5bb2e582ab45c6ef/db/blob/blob_file_builder.cc#L148-L176

ltamasi:
You are correct that multiple paths are currently unsupported for blob files. (Note that blob files are not associated with levels, so the logic for placing SSTs files does not readily generalize to blobs.) Since in BlobDB use cases we expect most of the data to be in the blob files, one workaround would be to use a single path, namely the one where you would like the blob files to be stored.

zaorangyang:
> You are correct that multiple paths are currently unsupported for blob files. (Note that blob files are not associated with levels, so the logic for placing SSTs files does not readily generalize to blobs.) Since in BlobDB use cases we expect most of the data to be in the blob files, one workaround would be to use a single path, namely the one where you would like the blob files to be stored.

Thanks for your reply. Maybe it would be better if a configuration parameter about blob file path was introduced. I can try to enhance it.

PR #8603, opened 2021-08-02
Support for integrated blobdb specific paths

zaorangyang:
Summary:
This PR is for https://github.com/facebook/rocksdb/issues/8535
A new option called blob_path to set blob file path for integrated blobdb is added. To maintain backward compatibility, if blob_path is empty, cf_paths.front() will be used.

### DIFF
--- a/db/blob/blob_file_builder_test.cc
+++ b/db/blob/blob_file_builder_test.cc
@@ -122,10 +122,8 @@ TEST_F(BlobFileBuilderTest, BuildAndCheckOneFile) {
   constexpr size_t value_offset = 1234;
 
   Options options;
-  options.cf_paths.emplace_back(
-      test::PerThreadDBPath(&mock_env_,
-                            "BlobFileBuilderTest_BuildAndCheckOneFile"),
-      0);
+  options.blob_path = test::PerThreadDBPath(
+      &mock_env_, "BlobFileBuilderTest_BuildAndCheckOneFile");
   options.enable_blob_files = true;
   options.env = &mock_env_;
 
@@ -177,9 +175,8 @@ TEST_F(BlobFileBuilderTest, BuildAndCheckOneFile) {
 
   const std::string& blob_file_path = blob_file_paths[0];
 
-  ASSERT_EQ(
-      blob_file_path,
-      BlobFileName(immutable_options.cf_paths.front().path, blob_file_number));
+  ASSERT_EQ(blob_file_path,
+            BlobFileName(immutable_options.blob_path, blob_file_number));
 
   ASSERT_EQ(blob_file_additions.size(), 1);
 
@@ -205,10 +202,8 @@ TEST_F(BlobFileBuilderTest, BuildAndCheckMultipleFiles) {
   constexpr size_t value_offset = 1234567890;
 
   Options options;
-  options.cf_paths.emplace_back(
-      test::PerThreadDBPath(&mock_env_,
-                            "BlobFileBuilderTest_BuildAndCheckMultipleFiles"),
-      0);
+  options.blob_path = test::PerThreadDBPath(
+      &mock_env_, "BlobFileBuilderTest_BuildAndCheckMultipleFiles");
   options.enable_blob_files = true;
   options.blob_file_size = value_size;
   options.env = &mock_env_;
@@ -262,8 +257,7 @@ TEST_F(BlobFileBuilderTest, BuildAndCheckMultipleFiles) {
     const uint64_t blob_file_number = i + 2;
 
     ASSERT_EQ(blob_file_paths[i],
-              BlobFileName(immutable_options.cf_paths.front().path,
-                           blob_file_number));
+              BlobFileName(immutable_options.blob_path, blob_file_number));
 
     const auto& blob_file_addition = blob_file_additions[i];
 
@@ -292,9 +286,8 @@ TEST_F(BlobFileBuilderTest, InlinedValues) {
   constexpr size_t value_offset = 1234567890;
 
   Options options;
-  options.cf_paths.emplace_back(
-      test::PerThreadDBPath(&mock_env_, "BlobFileBuilderTest_InlinedValues"),
-      0);
+  options.blob_path =
+      test::PerThreadDBPath(&mock_env_, "BlobFileBuilderTest_InlinedValues");
   options.enable_blob_files = true;
   options.min_blob_size = 1024;
   options.env = &mock_env_;
@@ -346,8 +339,8 @@ TEST_F(BlobFileBuilderTest, Compression) {
   constexpr size_t value_size = 100;
 
   Options options;
-  options.cf_paths.emplace_back(
-      test::PerThreadDBPath(&mock_env_, "BlobFileBuilderTest_Compression"), 0);
+  options.blob_path =
+      test::PerThreadDBPath(&mock_env_, "BlobFileBuilderTest_Compression");
   options.enable_blob_files = true;
   options.blob_compression_type = kSnappyCompression;
   options.env = &mock_env_;
@@ -387,9 +380,8 @@ TEST_F(BlobFileBuilderTest, Compression) {
 
   const std::string& blob_file_path = blob_file_paths[0];
 
-  ASSERT_EQ(
-      blob_file_path,
-      BlobFileName(immutable_options.cf_paths.front().path, blob_file_number));
+  ASSERT_EQ(blob_file_path,
+            BlobFileName(immutable_options.blob_path, blob_file_number));
 
   ASSERT_EQ(blob_file_additions.size(), 1);
 
@@ -428,9 +420,8 @@ TEST_F(BlobFileBuilderTest, CompressionError) {
   }
 
   Options options;
-  options.cf_paths.emplace_back(
-      test::PerThreadDBPath(&mock_env_, "BlobFileBuilderTest_CompressionError"),
-      0);
+  options.blob_path =
+      test::PerThreadDBPath(&mock_env_, "BlobFileBuilderTest_CompressionError");
   options.enable_blob_files = true;
   options.blob_compression_type = kSnappyCompression;
   options.env = &mock_env_;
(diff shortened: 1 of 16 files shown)


options/options_settable_test.cc
```diff
@@ -388,6 +388,7 @@ TEST_F(OptionsSettableTest, ColumnFamilyOptionsAllFieldsSettable) {
        sizeof(std::shared_ptr<MemTableRepFactory>)},
       {offset_of(&ColumnFamilyOptions::table_properties_collector_factories),
        sizeof(ColumnFamilyOptions::TablePropertiesCollectorFactories)},
+      {offset_of(&ColumnFamilyOptions::blob_path), sizeof(std::string)},
```
mrambacher, line 391: I do not think this option should be in the skip list but should be explicitly set by this test.

options/cf_options.cc
```diff
@@ -829,7 +829,8 @@ ImmutableCFOptions::ImmutableCFOptions(const ColumnFamilyOptions& cf_options)
           cf_options.memtable_insert_with_hint_prefix_extractor),
       cf_paths(cf_options.cf_paths),
       compaction_thread_limiter(cf_options.compaction_thread_limiter),
-      sst_partitioner_factory(cf_options.sst_partitioner_factory) {}
+      sst_partitioner_factory(cf_options.sst_partitioner_factory),
+      blob_path(cf_options.blob_path) {}
```
mrambacher, line 834: The blob path option should be added to the cf_immutable_options_type_info map

db/column_family.cc
```diff
@@ -359,6 +359,10 @@ ColumnFamilyOptions SanitizeOptions(const ImmutableDBOptions& db_options,
     }
   }
 
+  if (result.blob_path.empty() && !result.cf_paths.empty()) {
+    result.blob_path = result.cf_paths.front().path;
+  }
+
```
mrambacher, line 365: I worry about this approach and relocating/moving directories.  I think the approach taken in #8582 for dealing with the wal_dir would be a better solution.  Leave the value blank if not set or unset it if if matches the cf_path.  Do not force it to be set.

mrambacher (changes requested):
Having just been through this exercise with "wal_dir" and relocating files for checkpoints, I do not think this is entirely the right approach to use.  I agree that the "blob_path" should be added as an option, but think that it should be left as blank unless explicitly set.  There should be a "GetBlobPath off the ImmutableOptions that will either return "blob_path" (if set) or the cf_paths.front (if not set).

Additionally, I think tests with blobs and checkpoints should be written to make sure that the "blob_path" does not cause any issues.

zaorangyang:
> Having just been through this exercise with "wal_dir" and relocating files for checkpoints, I do not think this is entirely the right approach to use. I agree that the "blob_path" should be added as an option, but think that it should be left as blank unless explicitly set. There should be a "GetBlobPath off the ImmutableOptions that will either return "blob_path" (if set) or the cf_paths.front (if not set).
> 
> Additionally, I think tests with blobs and checkpoints should be written to make sure that the "blob_path" does not cause any issues.

Hi mrambacher, Thanks for your review.  I have browsed the https://github.com/facebook/rocksdb/pull/8582. Yes, I agree with you. I will update this PR later.  :)

db/blob/blob_file_builder_test.cc
```diff
@@ -122,10 +122,8 @@ TEST_F(BlobFileBuilderTest, BuildAndCheckOneFile) {
   constexpr size_t value_offset = 1234;
 
   Options options;
-  options.cf_paths.emplace_back(
-      test::PerThreadDBPath(&mock_env_,
-                            "BlobFileBuilderTest_BuildAndCheckOneFile"),
-      0);
+  options.blob_path = test::PerThreadDBPath(
+      &mock_env_, "BlobFileBuilderTest_BuildAndCheckOneFile");
```
mrambacher, line 126: Is this change necessary/expected?  Before this change, both Blob and SST files would be written to this test directory.  After the change, Blob and SST files will be written to separate places.  I agree this change is not likely to cause problems, but am wondering if there might be unintended side effects.


Note that this same change is applied to many tests...

mrambacher (commented):
This looks much better.  Thanks for addressing my earlier comments.  Only a couple of issues:
1. The options_settable_test change at 391 is probably no longer necessary
2. Can you please add a checkpoint test where the blob path is set?  Be nice to validate that:
- Blob files are copied as part of a checkpoint
- The Options file of a checkpoint has the appropriate settings for blob_path
- It is possible to load a DB from a checkpoint and get the appropriate behavior

3.  The same test case also applies to a "backup", though I am not sure how much backups and checkpoints differ.

Thanks!

zaorangyang:
> This looks much better. Thanks for addressing my earlier comments. Only a couple of issues:
> 
> 1. The options_settable_test change at 391 is probably no longer necessary
> 2. Can you please add a checkpoint test where the blob path is set?  Be nice to validate that:
> 
> * Blob files are copied as part of a checkpoint
> * The Options file of a checkpoint has the appropriate settings for blob_path
> * It is possible to load a DB from a checkpoint and get the appropriate behavior
> 
> 1. The same test case also applies to a "backup", though I am not sure how much backups and checkpoints differ.
> 
> Thanks!

Hi, mrambacher, I am working on the checkpoint with specified blob_path. I found some issues in current codebase:
1. If specified db_paths, A IO error ouccurs while executing checkpoint. The reason is that the path of source SST file just is ‘dbname_/sstname’。 Take a look at the following example：
`
TEST_F(CheckpointTest, CheckpointWithDBPath) {

options.db_paths.emplace_back(dbname_ + "_db", std::numeric_limits<uint64_t>::max());
Reopen(options);
constexpr char key[] = "key";

constexpr char value[] = "value";

ASSERT_OK(Put(key, value));

ASSERT_OK(Flush());

// Create a checkpoint
Checkpoint* checkpoint = nullptr;

ASSERT_OK(Checkpoint::Create(db_, &checkpoint));

std::unique_ptr<Checkpoint> checkpoint_guard(checkpoint);

ASSERT_OK(checkpoint->CreateCheckpoint(snapshot_name_));
}
`

This line of ASSERT_OK(checkpoint->CreateCheckpoint(snapshot_name_)) will report an error：

`
IO error: No such file or directory: while link file to snapshot_name_.tmp/000009.sst: dbname_/000009.sst: No such file or directory
`

2. If specified wal_dir,  the wal_dir in the OPTIONS file generated by checkpoint points the old dir. I think  it should be point to the checkpoint dir.

To support specified blob_path, I think the issue1 should be addressed first. The key is GetLiveFiles function。The GetLiveFiles just return the file name currently, I think it should return the dir and name for getting the right path when checkpointing. If you agree with me, I will create a new PR to address issue1 before finishing this PR.

mrambacher:
> > This looks much better. Thanks for addressing my earlier comments. Only a couple of issues:
> > 
> > 1. The options_settable_test change at 391 is probably no longer necessary
> > 2. Can you please add a checkpoint test where the blob path is set?  Be nice to validate that:
> > 
> > 
> > * Blob files are copied as part of a checkpoint
> > * The Options file of a checkpoint has the appropriate settings for blob_path
> > * It is possible to load a DB from a checkpoint and get the appropriate behavior
> > 
> > 
> > 1. The same test case also applies to a "backup", though I am not sure how much backups and checkpoints differ.
> > 
> > Thanks!
> 
> Hi, mrambacher, I am working on the checkpoint with specified blob_path. I found some issues in current codebase:
> 
> Firstly, If specified db_paths, A IO error ouccurs while executing checkpoint. The reason is that the path of source SST file is just ‘dbname_/sstname’。 Take a look at the following example：
> 
> ```
> TEST_F(CheckpointTest, CheckpointWithDBPath) {
>       Options options = CurrentOptions();
> 
>       options.create_if_missing = true;
> 
>       options.db_paths.emplace_back(dbname_ + "_db", std::numeric_limits<uint64_t>::max());
> 
>       Reopen(options);
> 
>       constexpr char key[] = "key";
>       
>       constexpr char value[] = "value";
>       
>       ASSERT_OK(Put(key, value));
>       
>       ASSERT_OK(Flush());
>       
>       // Create a checkpoint
>       Checkpoint* checkpoint = nullptr;
>       
>       ASSERT_OK(Checkpoint::Create(db_, &checkpoint));
>       
>       std::unique_ptr<Checkpoint> checkpoint_guard(checkpoint);
>       
>       ASSERT_OK(checkpoint->CreateCheckpoint(snapshot_name_));
> }
> ```
> 
> This line of ASSERT_OK(checkpoint->CreateCheckpoint(snapshot_name_)) will report an error：
> 
> `IO error: No such file or directory: while link file to snapshot_name_.tmp/000009.sst: dbname_/000009.sst: No such file or directory`
> 
> Secondly, If specified wal_dir, the wal_dir in the OPTIONS file generated by checkpoint points the old dir. I think it should be point to the checkpoint dir.
> 
> To support specified blob_path, I think the issue1 should be addressed first. The key is GetLiveFiles function。The GetLiveFiles just return the file name currently, I think it should return the dir and name for getting the right path when checkpointing. If you agree with me, I will create a new PR to address issue1 before finishing this PR. :)

@zaorangyang  : Please create issues for these two problems.  I do not think you can change "GetLiveFiles" to return full paths (it is documented not to) but might need to use a different API (GetLiveFileMetaData? GetColumnFamilyMetaData?) inside of CreateCheckpoint.  

The second issue (with the bad Options path), please assign to me.  There was an effort recently to fix something like that which I will follow up on and attempt to address.

mrambacher (commented):
Another thing.  One of the arugments to the CompactionJob is the blob output directory.  This appears at the moment to use GetDataDir of the first cf_path (at least one call from db_impl_remote_compaction).  This path also needs to be changed/updated.

zaorangyang:
> Thanks a lot for the patch @zaorangyang ! I see that you've marked is as draft; is it because it depends on #8647 getting fixed one way or the other?

Yes,  it depends on https://github.com/facebook/rocksdb/issues/8647

liam-thunder:
Seems we have a PR https://github.com/facebook/rocksdb/pull/8968 to disable checkpoint + db_path or cf_path as a workaround for https://github.com/facebook/rocksdb/issues/8647, so I suppose this PR could move forward? 
Looking forward to trying this option in some tiered storage scenarios where we could possibly put blob files on a slow device and SST files on a fast device. @zaorangyang @ltamasi @mrambacher

Issue #8440, opened 2021-06-22
Frequent SIGSEGV in DBImpl::GetAggregatedIntProperty() because of a race condition


igorcanadi:
### Expected behavior
DBImpl::GetAggregatedIntProperty() doesn't crash.

### Actual behavior
DBImpl::GetAggregatedIntProperty() crashes. Stack trace:
```
#0  0x00007fc56d4134b2 in rocksdb::ThreadLocalPtr::Swap (this=0x0, ptr=0x7fc56d7da960 <rocksdb::SuperVersion::dummy>) at util/thread_local.cc:539
#1  0x00007fc56d133dda in rocksdb::ColumnFamilyData::GetThreadLocalSuperVersion (this=this@entry=0x7fc398c15800, db=db@entry=0x7fc4310fc800) at db/column_family.cc:1185
#2  0x00007fc56d17417e in rocksdb::DBImpl::GetAndRefSuperVersion (cfd=0x7fc398c15800, this=0x7fc4310fc800) at db/db_impl/db_impl.cc:3260
#3  rocksdb::DBImpl::GetIntPropertyInternal (this=0x7fc4310fc800, cfd=0x7fc398c15800, property_info=..., is_locked=<optimized out>, value=0x7fc55b374180) at db/db_impl/db_impl.cc:3188
#4  0x00007fc56d174283 in rocksdb::DBImpl::GetAggregatedIntProperty (this=0x7fc4310fc800, property=..., aggregated_value=0x7fc55b3741e0) at db/db_impl/db_impl.cc:3242
```

`local_sv_` is `nullptr` here: https://github.com/facebook/rocksdb/blob/065bea158798f5d67eb95ce339f3650db174c109/db/column_family.cc#L1200

I think the root cause is unlocking the DB mutex in `ColumnFamilyData::UnrefAndTryDelete()`, here: https://github.com/facebook/rocksdb/blob/065bea158798f5d67eb95ce339f3650db174c109/db/column_family.cc#L688

This code block executes when remaining ref count is 1, and it is correctly assumed that the ref count comes from the `super_version_` (see here: https://github.com/facebook/rocksdb/blob/065bea158798f5d67eb95ce339f3650db174c109/db/column_family.cc#L683). However, if we release the mutex another piece of code can grab the new ref, as `GetAggregatedIntProperty()` does. As I understand it, there is an implicit assumption that `local_sv_` is non-nullptr if a ref is non-zero.

[...]

Issue #8382, opened 2021-06-10
SegFault during GetAggregateIntProperty call.


matthewvon:
I am logging this issue in case someone else has time to produce a test case before I do.

db->GetAggregatedIntProperty("rocksdb.estimate-table-readers-mem", & table_readers_size);

The above call internally acquires the database mutex before it walks the column families.  In our use case, this call is on an independent thread that has been awoken by an EventListener event, specifically the compaction complete event.  It appears that the compaction completion can release column family references leading to the delete of a column family without acquiring the database mutex.  That means the thread walking the column families for GetAggregatedIntProperty ends up iterating a deleted object.  Segfault results.

matthewvon:
I have known about this issue for a couple of months.  Feel bad for having not even mentioned it previously.  My apologies.

akankshamahajan15:
@matthewvon  Do you mind sharing the test code, if you have, to reproduce it.

matthewvon:
@akankshamahajan15 That is the problem.  Been too busy to create a sample test case.  And not likely to get time for another few weeks.  Felt it important to at least log this now, test case later.

ltamasi:
Hi @matthewvon ! Could you confirm if #8605 fixes the issue you've been seeing?

ltamasi:
Assuming fixed, please feel free to reopen if you still see this after the fix mentioned above.

PR #8605, opened 2021-08-02
Fix a race in ColumnFamilyData::UnrefAndTryDelete

ltamasi:
Summary:
The `ColumnFamilyData::UnrefAndTryDelete` code currently on the trunk
unlocks the DB mutex before destroying the `ThreadLocalPtr` holding
the per-thread `SuperVersion` pointers when the only remaining reference
is the back reference from `super_version_`. The idea behind this was to
break the circular dependency between `ColumnFamilyData` and `SuperVersion`:
when the penultimate reference goes away, `ColumnFamilyData` can clean up
the `SuperVersion`, which can in turn clean up `ColumnFamilyData`. (Assuming there
is a `SuperVersion` and it is not referenced by anything else.) However,
unlocking the mutex throws a wrench in this plan by making it possible for another thread
to jump in and take another reference to the `ColumnFamilyData`, keeping the
object alive in a zombie `ThreadLocalPtr`-less state. This can cause issues like
#8440 ,
#8382 ,
and might also explain the `was_last_ref` assertion failures from the `ColumnFamilySet`
destructor we sometimes observe during close in our stress tests.

Digging through the archives, this unlocking goes way back to 2014 (or earlier). The original
rationale was that `SuperVersionUnrefHandle` used to lock the mutex so it can call
`SuperVersion::Cleanup`; however, this logic turned out to be deadlock-prone.
https://github.com/facebook/rocksdb/pull/3510 fixed the deadlock but left the
unlocking in place. https://github.com/facebook/rocksdb/pull/6147 then introduced
the circular dependency and associated cleanup logic described above (in order
to enable iterators to keep the `ColumnFamilyData` for dropped column families alive),
and moved the unlocking-relocking snippet to its present location in `UnrefAndTryDelete`.
Finally, https://github.com/facebook/rocksdb/pull/7749 fixed a memory leak but
apparently exacerbated the race by (otherwise correctly) switching to `UnrefAndTryDelete`
in `SuperVersion::Cleanup`.

The patch simply eliminates the unlocking and relocking, which has been unnecessary
ever since #3510 made `SuperVersionUnrefHandle` lock-free.
This closes the window during which another thread could increase the reference count,
and hopefully fixes the issues above.

Test Plan:
Ran `make check` and stress tests locally.

### DIFF
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -678,15 +678,12 @@ bool ColumnFamilyData::UnrefAndTryDelete() {
     return true;
   }
 
-  // If called under SuperVersion::Cleanup, we should not re-enter Cleanup on
-  // the same SuperVersion. (But while installing a new SuperVersion, this
-  // cfd could be referenced only by two SuperVersions.)
   if (old_refs == 2 && super_version_ != nullptr) {
     // Only the super_version_ holds me
     SuperVersion* sv = super_version_;
     super_version_ = nullptr;
 
-    // Release SuperVersion reference kept in ThreadLocalPtr.
+    // Release SuperVersion references kept in ThreadLocalPtr.
     local_sv_.reset();
 
     if (sv->Unref()) {
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -457,7 +457,7 @@ void SuperVersion::Cleanup() {
     to_delete.push_back(m);
   }
   current->Unref();
-  cfd->UnrefAndTryDelete(this);
+  cfd->UnrefAndTryDelete();
 }
 
 void SuperVersion::Init(ColumnFamilyData* new_cfd, MemTable* new_mem,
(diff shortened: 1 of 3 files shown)


db/column_family.cc
```diff
@@ -678,22 +678,16 @@ bool ColumnFamilyData::UnrefAndTryDelete(SuperVersion* sv_under_cleanup) {
...
+
+    // Release SuperVersion references kept in ThreadLocalPtr.
     local_sv_.reset();
-    sv->db_mutex->Lock();
 
     if (sv->Unref()) {
-      // May delete this ColumnFamilyData after calling Cleanup()
+      // Note: sv will delete this ColumnFamilyData during Cleanup()
       sv->Cleanup();
       delete sv;
```
pdillinger, line 692: Presumably deleting memtables while holding db mutex here is not a big deal because we are deleting or shutting down a column family, right?

db/column_family.cc
```diff
@@ -678,22 +678,16 @@ bool ColumnFamilyData::UnrefAndTryDelete(SuperVersion* sv_under_cleanup) {
...
-    // This must be done outside of mutex_ since unref handler can lock mutex.
-    sv->db_mutex->Unlock();
+
+    // Release SuperVersion references kept in ThreadLocalPtr.
     local_sv_.reset();
-    sv->db_mutex->Lock();
 
     if (sv->Unref()) {
-      // May delete this ColumnFamilyData after calling Cleanup()
+      // Note: sv will delete this ColumnFamilyData during Cleanup()
```
pdillinger, line 690: Add `assert(sv->cfd == this);`?

pdillinger (approved):
After significant analysis of calling context and invariants, I believe this is right.

db/column_family.cc
ltamasi: Right. Also, note that the earlier code relocked the mutex after `local_sv_.reset()`, so`sv->Cleanup()` has always been done under the mutex (in fact, it *has to be* done under the mutex).

db/column_family.cc
ltamasi: Good call, will add

ltamasi: Done. Also realized we can do away with `SuperVersion::db_mutex` altogether, since this was the only place where it was used.

PR #8608, opened 2021-08-02
[WritePrepared] Fix bug caused by releasing snapshot(s) during compaction

riversand963:
In debug mode, we are seeing assertion failure as follows

```
db/compaction/compaction_iterator.cc:980: void rocksdb::CompactionIterator::PrepareOutput(): \
Assertion `ikey_.type != kTypeDeletion && ikey_.type != kTypeSingleDeletion' failed.
```

It is caused by releasing earliest snapshot during compaction between the execution of
`NextFromInput()` and `PrepareOutput()`.

In one case, as demonstrated in unit test `WritePreparedTransaction.ReleaseEarliestSnapshotDuringCompaction_WithSD2`,
incorrect result may be returned by a following range scan if we disable assertion, as in opt compilation
level: the SingleDelete marker's sequence number is zeroed out, but the preceding PUT is also
outputted to the SST file after compaction. Due to the logic of DBIter, the PUT will not be
skipped and will be returned by iterator in range scan. #8661 illustrates what happened.

Fix by taking a more conservative approach: make compaction zero out sequence number only
if key is in the earliest snapshot when the compaction starts.

Another assertion failure is
```
Assertion `current_user_key_snapshot_ == last_snapshot' failed.
```

It's caused by releasing the snapshot between the PUT and SingleDelete during compaction.

Test plan:
make check

### DIFF
--- a/db/compaction/compaction_iterator.cc
+++ b/db/compaction/compaction_iterator.cc
@@ -567,6 +567,8 @@ void CompactionIterator::NextFromInput() {
         // SingleDelete.
         if (prev_snapshot == 0 ||
             DefinitelyNotInSnapshot(next_ikey.sequence, prev_snapshot)) {
+          TEST_SYNC_POINT_CALLBACK(
+              "CompactionIterator::NextFromInput:SingleDelete:2", nullptr);
           if (next_ikey.type == kTypeSingleDeletion) {
             // We encountered two SingleDeletes in a row.  This could be due to
             // unexpected user input.
--- a/db/compaction/compaction_iterator.cc
+++ b/db/compaction/compaction_iterator.cc
@@ -632,7 +632,8 @@ void CompactionIterator::NextFromInput() {
         // iteration. If the next key is corrupt, we return before the
         // comparison, so the value of has_current_user_key does not matter.
         has_current_user_key_ = false;
-        if (compaction_ != nullptr && InEarliestSnapshot(ikey_.sequence) &&
+        if (compaction_ != nullptr &&
+            DefinitelyInSnapshot(ikey_.sequence, earliest_snapshot_) &&
             compaction_->KeyNotExistsBeyondOutputLevel(ikey_.user_key,
                                                        &level_ptrs_)) {
           // Key doesn't exist outside of this range.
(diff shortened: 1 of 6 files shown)


riversand963:
cc @lth

db/compaction/compaction_iterator.cc
```diff
@@ -663,7 +677,7 @@ void CompactionIterator::NextFromInput() {
                (ikey_.type == kTypeDeletion ||
                 (ikey_.type == kTypeDeletionWithTimestamp &&
                  cmp_with_history_ts_low_ < 0)) &&
-               InEarliestSnapshot(ikey_.sequence) &&
+               DefinitelyInSnapshot(ikey_.sequence, earliest_snapshot_) &&
```
jay-zhuang, line 680: Seems the difference between `InEarliestSnapshot()` and `DefinitelyInSnapshot()` is previous one will add released snapshot and check with the next snapshot:
https://github.com/facebook/rocksdb/blob/a207c278096d2d7a37fa28198e8c3a0f7e60a48c/db/compaction/compaction_iterator.cc#L1052-L1057
I find it's not very intuitive to find out.

jay-zhuang (approved):
LGTM

riversand963:

utilities/transactions/write_prepared_transaction_test.cc
```diff
@@ -2799,11 +2800,200 @@ TEST_P(WritePreparedTransactionTest, ReleaseEarliestSnapshotDuringCompaction) {
...
-                      {"key1", "value1", 0, kTypeValue}});
+                      {"key1", "value1", put_seq, kTypeValue}});
   db->ReleaseSnapshot(snapshot2);
   SyncPoint::GetInstance()->ClearAllCallBacks();
 }
 
+TEST_P(WritePreparedTransactionTest,
+       ReleaseEarliestSnapshotDuringCompaction_WithSD) {
+  constexpr size_t snapshot_cache_bits = 7;  // same as default
+  constexpr size_t commit_cache_bits = 0;    // minimum commit cache
```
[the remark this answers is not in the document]
riversand963: Done.

db/compaction/compaction_iterator.cc
riversand963: I think `InEarliestSnapshot()` accounts for the case in which a snapshot becomes the earliest snapshot during compaction because other earlier have been released.
In contrast, the implementation of write-prepared snapshot checker cannot give a deterministic answer when the (old) snapshot is released. Therefore, the name `DefinitelyInSnapshot()` indicates that:
- Snapshot is still active, and
- key is visible in the given snapshot.

Maybe we can rename `InEarliestSnapshot()` to `InCurrentEarliestSnapshot()`.

PR #8611, opened 2021-08-02
Make TraceRecord and Replayer public

autopear:
New public interfaces:
`TraceRecord` and `TraceRecord::Handler`, available in "rocksdb/trace_record.h".
`Replayer`, available in `rocksdb/utilities/replayer.h`.

User can use `DB::NewDefaultReplayer()` to create a Replayer to auto/manual replay a trace file.

Unit tests:
- `./db_test2 --gtest_filter="DBTest2.TraceAndReplay"`: Updated with the internal API changes.
- `./db_test2 --gtest_filter="DBTest2.TraceAndManualReplay"`: New for manual replay.


db/db_impl/db_impl.cc
```diff
@@ -5125,6 +5126,11 @@ Status DBImpl::EndTrace() {
   return s;
 }
 
+Replayer* DBImpl::NewReplayer(const std::vector<ColumnFamilyHandle*>& handles,
```
zhichao-cao, line 5129: I may prefer NewReplayer->NewDefaultReplayer.

db/db_impl/db_impl.cc
```diff
@@ -5125,6 +5126,11 @@ Status DBImpl::EndTrace() {
...
+                              std::unique_ptr<TraceReader>&& reader) {
+  return new ReplayerImpl(this, handles, std::move(reader));
```
zhichao-cao, line 5131: It might be risky the replayer user forget to deconstruct the instance.

include/rocksdb/trace_reader_writer.h
```diff
@@ -36,6 +36,10 @@ class TraceReader {
 
   virtual Status Read(std::string* data) = 0;
   virtual Status Close() = 0;
+
+  // Allow to restart from the beginning such that Replayer can replay multiple
+  // times without being recreated.
```
zhichao-cao, line 41: Indicate in the comment that the trace reader will seek back to the trace file header.

include/rocksdb/utilities/replayer.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <memory>
+
+#include "rocksdb/rocksdb_namespace.h"
+#include "rocksdb/status.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+struct ReplayOptions {
+  uint32_t num_threads;
```
zhichao-cao, line 17: Add comments for each option.

include/rocksdb/utilities/replayer.h
```diff
@@ -0,0 +1,52 @@
...
+  double fast_forward;
+
+  ReplayOptions() : num_threads(1), fast_forward(1.0) {}
+  ReplayOptions(uint32_t num_of_threads, double fast_forward_ratio)
+      : num_threads(num_of_threads), fast_forward(fast_forward_ratio) {}
+};
+
+// Replayer helps to replay the captured RocksDB operations.
```
zhichao-cao, line 25: "captured RocksDB operations." - >"captured RocksDB query level operations." We may have other replay later (e.g., block level, IO level)

include/rocksdb/utilities/replayer.h
```diff
@@ -0,0 +1,52 @@
...
+  // Return Status::OK() if the execution was successful (Get() and MultiGet()
+  // will still return Status::OK() even if they returned Status::NotFound());
+  // Status::Incomplete() if no more trace available;
+  // Status::NotSupported() if the operation is not supported;
+  // Otherwise, return the corresponding error.
+  virtual Status Step() { return Status::NotSupported(); }
+
+  // Replay all the traces from the provided trace stream, taking the delay
+  // between the traces into consideration.
+  virtual Status Replay(ReplayOptions /*options*/ = ReplayOptions()) {
```
zhichao-cao, line 46: I prefer to set the options at Prepare()

trace_replay/trace_replay.cc
```diff
@@ -508,132 +504,73 @@ Status Replayer::Replay() {
...
-      } else {
-        TracerHelper::DecodeIterPayload(&trace, &iter_payload);
-      }
-      if (iter_payload.cf_id > 0 &&
-          cf_map_.find(iter_payload.cf_id) == cf_map_.end()) {
-        return Status::Corruption("Invalid Column Family ID.");
-      }
+Status ReplayerImpl::Replay(ReplayOptions options) {
+  if (options.fast_forward <= 0.0) {
+    return Status::InvalidArgument("Wrong fast forward speed!");
```
zhichao-cao, line 536: We may check it in Prepare.

trace_replay/trace_replay.cc
```diff
@@ -508,132 +504,73 @@ Status Replayer::Replay() {
...
-        assert(i < multiget_payload.cf_ids.size() &&
-               i < multiget_payload.multiget_keys.size());
-        if (cf_map_.find(multiget_payload.cf_ids[i]) == cf_map_.end()) {
-          return Status::Corruption("Invalid Column Family ID.");
-        }
-        v_cfd.push_back(cf_map_[multiget_payload.cf_ids[i]]);
-        keys.push_back(Slice(multiget_payload.multiget_keys[i]));
+  {
+    std::lock_guard<std::mutex> guard(mutex_);
+    if (!prepared_) {
```
zhichao-cao, line 541: If it is not Prepared, do you think it is better to fail the replay, instead of prepare it internally? We may need an explicit prepare

include/rocksdb/utilities/replayer.h
autopear: I think we should not provide `Prepare()`, but instead, provide `Reset()`. Given the current API, a user may do a few `Step()`s then start `Replay()`. Or he/she can start `Replay()` and interrupt the process, then start `Step()`, and then start `Replay()`. Considering the where Replayer is used currently (db_bench), we can make the prepare/init step on-demand when a replay/step starts, and hide this step from the user. But the user does need a function to restart if a replay is finished, or the user wants to abort and restart.

In this case, it's strange to set ReplayOptions in `Prepare()`, unless you want 2 sets of options, one for controlling playback speed, such as num threads, fast forward rate, etc.  Another set is explicit used in initialization, for example, control how to read/parse the header or each trace.

db/db_impl/db_impl.cc
autopear: Done

db/db_impl/db_impl.cc
autopear: Done

include/rocksdb/utilities/replayer.h
autopear: Done

include/rocksdb/utilities/replayer.h
autopear: Done

include/rocksdb/trace_reader_writer.h
autopear: Done

trace_replay/trace_replay.cc
autopear: I prefer to have on-demand auto prepare, but allow the user to reset.

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,133 @@
...
+  virtual ~TraceRecord() {}
+
+  virtual TraceType GetType() const = 0;
+
+  // Timestamp (in microseconds) of this trace.
+  uint64_t timestamp;
+};
+
+// Base class for all query types of trace records.
+class QueryTraceRecord : public TraceRecord {
```
zhichao-cao, line 52: Is it better to add the trace type as the member?

trace_replay/trace_replay.h
```diff
@@ -222,74 +204,93 @@ class Tracer {
...
-  Status ReadFooter(Trace* footer);
-  Status ReadTrace(Trace* trace);
+  using Replayer::Prepare;
+  Status Prepare() override;
 
-  // The background function for MultiThreadReplay to execute Get query
-  // based on the trace records.
-  static void BGWorkGet(void* arg);
+  using Replayer::NextTraceRecord;
+  Status NextTraceRecord(std::unique_ptr<TraceRecord>* record) override;
```
zhichao-cao, line 217: We should indicate that, the trace header and footer are excluded

include/rocksdb/trace_record.h
autopear: It's inherited from the base class `TraceRecord::GetType()`.

trace_replay/trace_replay.h
autopear: Added

include/rocksdb/db.h
```diff
@@ -1680,6 +1682,14 @@ class DB {
   }
 #endif  // ROCKSDB_LITE
 
+  // Create a default trace replayer.
+  virtual Status NewDefaultReplayer(
```
zhichao-cao, line 1686: We should mention that this Replayer can only replay the trace to this DB.

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,133 @@
...
+#include "rocksdb/write_batch.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// Supported Trace types.
+enum TraceType : char {
+  kTraceNone = 0,
+  kTraceBegin = 1,
+  kTraceEnd = 2,
+  kTraceWrite = 3,
```
zhichao-cao, line 21: Between L20 and L21, add comment "Query level tracing related trace types" and replace L33. L25 changes to "Block cache tracing related trace types" and L31 changes to "IO tracing related trace types"

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,133 @@
...
+
+// To do: ROCKSDB_LITE ?
+
+#include "rocksdb/db.h"
+#include "rocksdb/slice.h"
+#include "rocksdb/write_batch.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// Supported Trace types.
```
zhichao-cao, line 16: "Supported trace record types"

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,133 @@
...
+  ColumnFamilyHandle* handle;
+
+  // Key to get.
+  Slice key;
+};
+
+// Base class for all Iterator related operations.
+class IteratorQueryTraceRecord : public QueryTraceRecord {
+ public:
+  explicit IteratorQueryTraceRecord(uint64_t ts = 0) : QueryTraceRecord(ts) {}
```
zhichao-cao, line 88: What's the benefit of introducing the IteratorQueryTraceRecord?

include/rocksdb/db.h
autopear: Do you mean the replayer can only replay trace file generated from the same db, or the replay will only be executed on this db? It is possible to create the trace file from a different db and replay on another db.

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,133 @@
...
+class GetQueryTraceRecord : public QueryTraceRecord {
+ public:
+  explicit GetQueryTraceRecord(uint64_t ts = 0)
+      : QueryTraceRecord(ts), handle(nullptr) {}
+  virtual ~GetQueryTraceRecord() override {}
+
+  TraceType GetType() const override { return kTraceGet; };
+
+  // Column family to search.
+  ColumnFamilyHandle* handle;
```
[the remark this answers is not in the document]
autopear: I'm not sure about this. If we just provide cf_id, how can a user translate it to a handle? A vector of handles is needed to create the replayer, but given a cf_id, the user may need to use linear search to map the cf_id to the handle from the vector, or have his/her own mapping for this purpose.

include/rocksdb/trace_record.h
autopear: If later we have more trace operations related to iterator, such as seek to first/last, next, perhaps it's better to groups them together into a virtual parent class.

trace_replay/trace_replay.cc
```diff
@@ -471,269 +470,167 @@ Status Tracer::WriteTrace(const Trace& trace) {
...
   if (!s.ok()) {
     return s;
   }
   s = TracerHelper::ParseTraceHeader(header, &trace_file_version_, &db_version);
   if (!s.ok()) {
     return s;
   }
+  header_ts_ = header.ts;
+  prepared_ = true;
+  trace_end_ = false;
```
mrambacher, line 504: Won't these values be set incorrectly if Prepare succeeded the first time and failed the second?

trace_replay/trace_replay.cc
```diff
@@ -471,269 +470,167 @@ Status Tracer::WriteTrace(const Trace& trace) {
...
-        return Status::Corruption("Invalid Column Family ID.");
-      }
+  Trace trace;
+  Status s = ReadTrace(&trace);  // ReadTrace is atomic
+  // Reached the trace end.
+  if (s.ok() && trace.type == kTraceEnd) {
+    trace_end_ = true;
+    return Status::Incomplete("Trace end.");
+  }
+  if (!s.ok() || record == nullptr) {
```
mrambacher, line 523: Why is the input record being null a valid case and not an assert?

trace_replay/trace_replay.cc
```diff
@@ -471,269 +470,167 @@ Status Tracer::WriteTrace(const Trace& trace) {
...
   }
-  ThreadPoolImpl thread_pool;
-  thread_pool.SetHostEnv(env_);
 
-  if (threads_num > 1) {
-    thread_pool.SetBackgroundThreads(static_cast<int>(threads_num));
-  } else {
-    thread_pool.SetBackgroundThreads(1);
+  Status s = Status::OK();
+  ThreadPoolImpl* thread_pool = nullptr;
```
mrambacher, line 547: Seems like this would be better as a std::unique_ptr

trace_replay/trace_replay.cc
```diff
@@ -471,269 +470,167 @@ Status Tracer::WriteTrace(const Trace& trace) {
...
+        replay_epoch +
+        std::chrono::microseconds(static_cast<uint64_t>(
+            llround(1.0 * (trace.ts - header_ts_) / options.fast_forward))));
+
+    if (thread_pool) {
+      std::unique_ptr<ReplayerWorkerArg> ra(new ReplayerWorkerArg);
+      ra->db = db_;
+      ra->trace_entry = std::move(trace);
+      ra->cf_map = &cf_map_;
+      ra->trace_file_version = trace_file_version_;
```
mrambacher, line 585: Why not pass these arguments to the ReplayWorkerArg constructor?

trace_replay/trace_replay.cc
```diff
@@ -764,152 +661,260 @@ Status Replayer::ReadFooter(Trace* footer) {
...
       reinterpret_cast<ReplayerWorkerArg*>(arg));
   assert(ra != nullptr);
-  auto cf_map = static_cast<std::unordered_map<uint32_t, ColumnFamilyHandle*>*>(
-      ra->cf_map);
+
+  std::unique_ptr<TraceRecord> record;
+  Status s = ToTraceRecord(&(ra->trace_entry), ra->cf_map,
+                           ra->trace_file_version, &record);
+  if (s.ok()) {
+    ExecuteTrace(ra->db, std::move(record)).PermitUncheckedError();
```
mrambacher, line 728: Why is it okay if a thread ignores the ExecuteTrace status but the main thread does not?

trace_replay/trace_replay.cc
```diff
@@ -764,152 +661,260 @@ Status Replayer::ReadFooter(Trace* footer) {
...
+    Trace* trace, std::unordered_map<uint32_t, ColumnFamilyHandle*>* /*cf_map*/,
+    int trace_file_version, std::unique_ptr<TraceRecord>* record) {
+  assert(trace != nullptr);
+  assert(trace->type == kTraceWrite);
+  if (record == nullptr) {
+    return Status::OK();
+  }
+
+  // To do: is there any validation to convert a payload string to a WriteBatch?
+  if (record != nullptr) {
```
mrambacher, line 742: Doesn't this have to be true based on line 737?

trace_replay/trace_replay.cc
```diff
@@ -764,152 +661,260 @@ Status Replayer::ReadFooter(Trace* footer) {
...
+    WriteQueryTraceRecord* r = new WriteQueryTraceRecord(trace->ts);
+
+    if (trace_file_version < 2) {
+      r->batch = WriteBatch(trace->payload);
+    } else {
+      WritePayload w_payload;
+      TracerHelper::DecodeWritePayload(trace, &w_payload);
+      r->batch = WriteBatch(w_payload.write_batch_data.ToString());
+    }
+    record->reset(r);
```
mrambacher, line 752: I think most of this should be done inside of the WriteQueryTraceRecord constructor and not have the fields set externally.  That seems like a cleaner solution and would allow the values to be const.

include/rocksdb/utilities/replayer.h
```diff
@@ -0,0 +1,68 @@
...
+
+  ReplayOptions() : num_threads(1), fast_forward(1.0) {}
+  ReplayOptions(uint32_t num_of_threads, double fast_forward_ratio)
+      : num_threads(num_of_threads), fast_forward(fast_forward_ratio) {}
+};
+
+// Replayer helps to replay the captured RocksDB query level operations.
+// The Replayer can either be created from DB::NewReplayer method, or be
+// instantiated via db_bench today, on using "replay" benchmark.
+class Replayer {
```
mrambacher, line 36: I am not sure I understand this interface class.  I can see why/where Prepare and Replay are invoked publicly but I do not see any cases where Next and Execute are used outside of the Replay implementation.

Is there a need for those two methods publicly?  What is the use case?  Without them in the public API this PR could be greatly simplified as there would be many fewer public moving parts (TraceRecord and TraceType would no longer need to be public).

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,133 @@
...
+  // For IOTracing.
+  kIOTracer = 12,
+  // For query tracing
+  kTraceMultiGet = 13,
+  // All trace types should be added before kTraceMax
+  kTraceMax,
+};
+
+// Base class for all types of trace records.
+class TraceRecord {
```
mrambacher, line 40: This class appears to be a more complex version of Trace (from trace_replay.h).  Is there a reason not to simply move Trace into the public API?  There is already a "TODO" comment there stating that should be done (probably renamed to "class TraceRecord").  You could create accessor functions for the various payload types in the TraceRecord and do away with the many types of classes (e.g. 

Status TraceRecord::GetWriteBatch(WriteBatch *wb) const { 
  if (payload_map & kWriteBatchData) { /
    / fill and return ok;
  } else {
   return Status::NotFound("No WriteBatch in this record");
 } 
}

mrambacher (commented):
I am concerned with the complexity of this change (there are a lot of classes in the public API) and what is actually gained in this implementation.  It seems like it solves part of a problem (allow replays against a database) but does not necessarily lend itself to solving other problems (implementing custom TraceReader/Writer classes).

include/rocksdb/trace_record.h
autopear: I think we are not going to make Trace public, so payload content will not be visible to the external. That is to say, TraceRecord should not contain anything from payload. There won't be public APIs that convert a Trace to a TraceRecord, or from payload to key(s)/cf_id(s), etc. That also the reason why TraceRecord is a class type, not struct type, since you cannot have a struct that has all information for different operations.

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,133 @@
...
+ public:
+  explicit TraceRecord(uint64_t ts = 0) : timestamp(ts) {}
+  virtual ~TraceRecord() {}
+
+  virtual TraceType GetType() const = 0;
+
+  // Timestamp (in microseconds) of this trace.
+  uint64_t timestamp;
```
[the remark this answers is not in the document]
autopear: It's intended. User can modify the timestamp or any record content, or even artificially create a trace record to replay.

include/rocksdb/trace_record.h
autopear: Trace type of a trace record should be read only, thus not class member.

include/rocksdb/trace_record.h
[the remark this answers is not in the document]
autopear: I'm ok with either. If the user has no problem retrieving the handle pointer from its cf_id to execute the operation, then it's fine.

include/rocksdb/utilities/replayer.h
```diff
@@ -0,0 +1,68 @@
...
+
+  // Make some preparation before replaying the trace. This will also reset the
+  // replayer in order to restart replaying.
+  virtual Status Prepare() = 0;
+
+  // Return the timestamp when the trace recording was started.
+  virtual uint64_t GetHeaderTimestamp() const = 0;
+
+  // Atomically read one trace into a TraceRecord (excluding the header and
+  // footer traces). Return Status::OK() on succes; Status::Incomplete() if
```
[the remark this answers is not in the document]
autopear: ack

include/rocksdb/utilities/replayer.h
```diff
@@ -0,0 +1,68 @@
...
+  // from DB::Get() or DB::MultiGet();
+  // Status::Incomplete() if Prepare() was not called or no more available
+  // trace;
+  // Status::NotSupported() if the operation is not supported;
+  // Otherwise, return the corresponding error status.
+  virtual Status Execute(std::unique_ptr<TraceRecord>&& record) = 0;
+
+  // Replay all the traces from the provided trace stream, taking the delay
+  // between the traces into consideration.
+  virtual Status Replay(const ReplayOptions& options = ReplayOptions()) = 0;
```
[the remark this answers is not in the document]
autopear: ack

include/rocksdb/trace_record.h
autopear: ack

include/rocksdb/trace_record.h
autopear: ack

include/rocksdb/utilities/replayer.h
autopear: We need 2 sets of APIs for replay. One is auto-replay, following timestamps thus controlled by number of threads and fast_forward speed. The other set is for manual replay, which basically does 2 things, one is decoding some string (acquired from TraceReader) to a TraceRecord (via Trace internally), the other is executing it (actually this is not necessary since user know the contents of the operation, and he/she can just call the corresponding DB APIs to execute them).

include/rocksdb/trace_record.h
mrambacher: I would argue that you are effectively making Trace public via the TraceRecord, but in a way that is more complex than the original Trace (lots of classes and potential to go astray).  I would suggest making Trace public and have a way of getting the fields out of it that you want/need.  With some thought you may not need to make the Payload types public and can get/set them via a public API (or via Trace constructors).

There is quite a bit to this TraceRecord class hierarchy and I think most of it duplicates something that already exists in the Trace/Payload methods.  Why re-invent the wheel?

db/db_test2.cc
```diff
@@ -4271,6 +4280,62 @@ TEST_F(DBTest2, TraceAndReplay) {
...
+  ASSERT_OK(replayer->Replay(ReplayOptions(2, 2.0)));
+
+  // Re-replay using 2 threads, 1/2 speed.
+  ASSERT_OK(replayer->Prepare());
+  ASSERT_OK(replayer->Replay(ReplayOptions(2, 0.5)));
+
+  // Manual replay
+  // Iteration 1: Manul replay should succeed;
+  // Iteration 2: Manul replay should fail with no Prepare();
+  // Iteration 3: Manul replay again should succeed with Prepare();
```
mrambacher, line 4298: "Manual"

db/db_test2.cc
```diff
@@ -4271,6 +4280,62 @@ TEST_F(DBTest2, TraceAndReplay) {
...
+    while (s.ok()) {
+      std::unique_ptr<TraceRecord> record;
+      s = replayer->NextTraceRecord(&record);
+      // Skip unsupported operations.
+      if (s.IsNotSupported()) {
+        continue;
+      }
+      if (s.ok()) {
+        // A TraceRecord's timestamp must be no smaller than the header
+        // timestamp and the timestamp of its previous trace record.
```
mrambacher, line 4317: Since we are reading these from the file, won't this always be true?  

Shouldn't this condition be checked inside of Execute?

And shouldn't there be a test where this condition is not true to validate the check works?

include/rocksdb/utilities/replayer.h
```diff
@@ -0,0 +1,68 @@
...
+  // Prepare() was not called or no more available trace; Status::NotSupported()
+  // if the read trace type is not supported.
+  virtual Status NextTraceRecord(std::unique_ptr<TraceRecord>* record) = 0;
+
+  // Execute one TraceRecord.
+  // Return Status::OK() if the execution was successful. Get/MultiGet traces
+  // will still return Status::OK() even if they got Status::NotFound()
+  // from DB::Get() or DB::MultiGet();
+  // Status::Incomplete() if Prepare() was not called or no more available
+  // trace;
```
mrambacher, line 58: How would a caller differentiate between the Replayer not being prepared versus no more records?

include/rocksdb/trace_record.h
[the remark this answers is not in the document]
mrambacher: Please create some tests where the records are being created and inserted into the Replay stream

include/rocksdb/utilities/replayer.h
autopear: Currently no. We should either introduce a new status indicating end of file/stream, or have a bool function that indicates the end of trace replay in Replayer.

include/rocksdb/trace_record.h
autopear: We can make Trace public, but then we also have to make `TracePayloadType`, `WritePayload`, `GetPayload`, `IterPayload`, `MultiGetPayload`, `TraceHelper`, and other types of payloads used in block cache tracing and IO tracing public. These payload types are specific to the trace type thus are not generic. This incurs more complexity than having a single `TraceRecord` class.

db/db_test2.cc
autopear: ack

include/rocksdb/trace_record.h
mrambacher: I do not think you need to make TracePayloadType, WritePayload, etc public.  

If you wanted to only read (and not create) new Trace records, you need to know the TraceType and have accessor methods to read the payload fields (GetWriteBatch, GetColumnFamilyID, etc).  

If you want to be able to create new Trace records, you could either do that via various constructors (mimicking the Payload arguments) or have setter functions (potentially more complex).

include/rocksdb/trace_record.h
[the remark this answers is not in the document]
autopear: ack

trace_replay/trace_replay.cc
autopear: I don't think there is a validation function for constructing a WriteBatch via the string only constructor. I guess the only time we know if a write batch is valid is when it's executed. For this function, since we cannot do any validation before creating the batch, if record is null, it probably means the caller doesn't care about this record and just want to skip it. (For example, quickly scan a trace file and see what are the operations).

trace_replay/trace_replay.cc
autopear: It's intended to leave the values non-const.

db/db_test2.cc
autopear: `s.ok()` is false when reaching the trace end, so the execution should only be called if `NextTraceRecord` returns OK.

The timestamp checking is trivial in this test. But I wonder if it's possible to have incorrect/out-of-order timestamps in a trace file?

I can add a test for manual replay, and move the current manual replay part into a separate test.

include/rocksdb/trace_record.h
autopear: Yes, if we don't make payload types public, we need to make TraceHelper public, or provide another class with lots of static functions to GetGetTraceKey(), GetGetTraceCFID(), GetWriteTraceRep(), ... etc.

Later whenever we add a new trace type, we need add one or multiple public functions to this class. Isn't it more complex than using `TraceRecord`?

mrambacher: I do not think it is more complex and TraceHelper does not need to be public.  GetTraceKey would internally use TraceHelper but TraceHelper does not need to be public.

You only need different methods for different types of data.  If multiple types return a ColumnFamilyId, they all can use the same method.

The proposed approach requires a new class for every type.  The approach I am suggesting requires a new method for every unique field in payload (from the payload mask).   If a new Trace type is added and we are allowing users to construct them, then a new constructor would also be required.

include/rocksdb/db.h
zhichao-cao: > Do you mean the replayer can only replay trace file generated from the same db, or the replay will only be executed on this db? It is possible to create the trace file from a different db and replay on another db.

No I mean, all the traces can only be replayed on a certain DB which create this replayer. Initially, replayer is independent to the DB. But now, the replayer created from this interface is fixed to the DB.

include/rocksdb/trace_record.h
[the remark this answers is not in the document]
zhichao-cao: @autopear , you can see, the user pass in the CF handle pointer, they already have all the pointers. Just ID here make more scence.

zhichao-cao:
> I am concerned with the complexity of this change (there are a lot of classes in the public API) and what is actually gained in this implementation. It seems like it solves part of a problem (allow replays against a database) but does not necessarily lend itself to solving other problems (implementing custom TraceReader/Writer classes).

So, currently, the logic is, RocksDB will do the tracing, encode the trace record to a string, and replay. User only has the flexibility to decide how to write the trace record string (trace writer). Also, to decode the trace, user can feed in the string and RocksDB will do the decoding, provide the decoded content to the user or do the replay. So it is enough to make the TraceRecord public.

include/rocksdb/utilities/replayer.h
zhichao-cao: @mrambacher I think it is ok for now end of trace and no prepare behave the same. In general, it means, no more replay avaialble.

include/rocksdb/trace_record.h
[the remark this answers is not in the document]
autopear: Yes, changed. The decoding process should not use any column family handle from the db instance. Handles should be validated during execution.

db/db_test2.cc
autopear: Added `DBTest2.TraceAndManualReplay`.

include/rocksdb/trace_record.h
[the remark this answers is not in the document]
autopear: Added `DBTest2.TraceAndManualReplay`.

trace_replay/trace_replay.cc
autopear: The thread function requires some callback function to update the execution status and break the loop. But there is a synchronization issue between the execution starts and finishes. I'm still thinking the best way to handle this issue. Feel free to leave any suggestions..

ajkr (commented):
> > I am concerned with the complexity of this change (there are a lot of classes in the public API) and what is actually gained in this implementation. It seems like it solves part of a problem (allow replays against a database) but does not necessarily lend itself to solving other problems (implementing custom TraceReader/Writer classes).
> 
> So, currently, the logic is, RocksDB will do the tracing, encode the trace record to a string, and replay. User only has the flexibility to decide how to write the trace record string (trace writer). Also, to decode the trace, user can feed in the string and RocksDB will do the decoding, provide the decoded content to the user or do the replay. So it is enough to make the TraceRecord public.

"provide the decoded content to the user" - Is this needed right now? If it is not needed does that mean we don't need to make `TraceRecord` public yet? It seems kind of tricky to design the `TraceRecord` interface (for example right now it's using `std::string`s for kvs but what if we want `Slice`?)

ajkr:
> > > I am concerned with the complexity of this change (there are a lot of classes in the public API) and what is actually gained in this implementation. It seems like it solves part of a problem (allow replays against a database) but does not necessarily lend itself to solving other problems (implementing custom TraceReader/Writer classes).
> > 
> > 
> > So, currently, the logic is, RocksDB will do the tracing, encode the trace record to a string, and replay. User only has the flexibility to decide how to write the trace record string (trace writer). Also, to decode the trace, user can feed in the string and RocksDB will do the decoding, provide the decoded content to the user or do the replay. So it is enough to make the TraceRecord public.
> 
> "provide the decoded content to the user" - Is this needed right now? If it is not needed does that mean we don't need to make `TraceRecord` public yet? It seems kind of tricky to design the `TraceRecord` interface (for example right now it's using `std::string`s for kvs but what if we want `Slice`?)

On second thought I am fine with it. It is needed to implement the natural API to separate decode and execution steps without any restrictions. It also allows synthesizing execution step only.

Overall I read the whole code and it's excellent. There is one suggestion I have to avoid downcasting the `TraceRecord` using more object-oriented techniques that I will write up today.

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,188 @@
...
+  // Serialized string object to construct a WriteBatch.
+  std::string rep;
+};
+
+// Trace record for DB::Get() operation
+class GetQueryTraceRecord : public QueryTraceRecord {
+ public:
+  GetQueryTraceRecord();
+  GetQueryTraceRecord(uint32_t cf_id, const std::string& key, uint64_t ts);
+  GetQueryTraceRecord(uint32_t cf_id, std::string&& key, uint64_t ts);
```
ajkr, line 103: This pattern is fine to save a string move construction. But wouldn't it be more beneficial to avoid per-field buffers entirely? We should at least try to avoid exposing per-field buffers in the public interface because then we're basically stuck supporting them.

I can see it's non-trivial because buffers are copied from `Trace::payload` which goes out of scope before `ReplayerImpl::Next()` returns. Could `TraceRecord`s take ownership of the underlying storage buffer (`Trace::payload`) and then only expose `Slice`s in its public interface?

include/rocksdb/utilities/replayer.h
```diff
@@ -0,0 +1,72 @@
...
+  // Return the timestamp when the trace recording was started.
+  virtual uint64_t GetHeaderTimestamp() const = 0;
+
+  // Atomically read one trace into a TraceRecord (excluding the header and
+  // footer traces).
+  // Return Status::OK() on success;
+  // Status::Incomplete() if Prepare() was not called or no more available
+  // trace;
+  // Status::NotSupported() if the read trace type is not supported.
+  virtual Status Next(std::unique_ptr<TraceRecord>* record) = 0;
```
ajkr, line 53: `TraceRecord` is not very user friendly right now because it requires downcasting to a concrete type based on an enum value (`TraceType`). I think a more usual way to do this is with a handler class accepted by our data type (`TraceRecord`). Client will invoke a virtual function on `TraceRecord` that accepts a `Handler` base class.  That function will be overridden to delegate to a `Handler` function corresponding to the `TraceRecord` subclass. (See "visitor design pattern", "double dynamic dispatch".)

Some pseudocode:

```
class TraceRecord {
 public:
  class Handler {
   public:
    // Another way is to name the functions differently rather than relying on overloading
    virtual Status Handle(const WriteQueryTraceRecord& record) = 0;
    virtual Status Handle(const GetQueryTraceRecord& record) = 0;
    ...
  };
  virtual Status Accept(Handler* handler) = 0;
  ...
};

class WriteQueryTraceRecord : public QueryTraceRecord {
 public:
  virtual Status Accept(Handler* handler) override {
    return handler->Handle(*this);
  }
  ...
};
...
```

Then we need to migrate `TraceAnalyzer`, etc., to use it. They need to implement a `TraceRecord::Handler` and pass it to `Accept()`.

ajkr (changes requested):

include/rocksdb/trace_record.h
zhichao-cao: That's the good point. I had several discussion with Merlin in terms of using Slice or using string. If we consider to provide full flexibility to the user, then user should be able to exam and make the change to the trace record content before apply it to Execute. So, using string which carry the ownership may be a better solution.

But I thought it again today. Actually, if the user want to change any content of the trace record and apply the updated trace record in `Execute`, user can create its own trace record, which reference to the content that user hold the ownership (user create the key or value string and reference them in the new trace record). Therefore, in this case, Slice still works. 

@autopear I agree with Andrew that we should use Slice in the public interface. The ownership of the original trace record should be the storage layer. It is also more efficient.

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,192 @@
...
+class TraceRecord {
+ public:
+  TraceRecord();
+  explicit TraceRecord(uint64_t ts);
+  virtual ~TraceRecord();
+
+  virtual TraceType GetType() const = 0;
+
+  // Execute the trace record on the specified DB and handles.
+  virtual Status Execute(DB* db,
```
zhichao-cao, line 57: It is a little strange to include Execute in the TraceRecord. Basically, TraceRecord is just similar to a data structure which stores the trace information. Execution should be the logic in the place where we use the trace record.

include/rocksdb/trace_record.h
autopear: Done

include/rocksdb/trace_record.h
autopear: I'm changing it to handlers.

autopear: Done

include/rocksdb/utilities/replayer.h
autopear: Done

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,232 @@
...
+
+    virtual Status Handle(const WriteQueryTraceRecord& record) override;
+    virtual Status Handle(const GetQueryTraceRecord& record) override;
+    virtual Status Handle(const IteratorSeekQueryTraceRecord& record) override;
+    virtual Status Handle(const MultiGetQueryTraceRecord& record) override;
+
+   private:
+    DB* db_;
+    std::unordered_map<uint32_t, ColumnFamilyHandle*> cf_map_;
+  };
```
ajkr, line 93: Can we expose `NewExecutionHandler(DB* db, const std::vector<ColumnFamilyHandle*>& handles)` factory function only and keep the class definition internal?

include/rocksdb/trace_record.h
```diff
@@ -0,0 +1,232 @@
...
+ public:
+  WriteQueryTraceRecord(const WriteBatch& batch, uint64_t timestamp);
+
+  WriteQueryTraceRecord(WriteBatch&& batch, uint64_t timestamp);
+
+  WriteQueryTraceRecord(const std::string& rep, uint64_t timestamp);
+
+  WriteQueryTraceRecord(std::string&& rep, uint64_t timestamp);
+
+  WriteQueryTraceRecord(std::shared_ptr<WriteBatch> batch, uint64_t timestamp);
```
ajkr, line 121: I think we should just do `WriteQueryTraceRecord(PinnableSlice&& batch_rep, uint64_t timestamp)` and `Slice batch_rep()` now and worry about how to make a `WriteBatch` without copy later. I'm sure we can figure out a way to make a `WriteBatch` that reads from a `Slice`.

ajkr (changes requested):
Sorry for more comments. I think they will be relatively easy for you to address though.

ajkr (approved):
Great PR, thanks! I guess the CI will point you to some small things to fixup before landing.

include/rocksdb/trace_record.h
autopear: Done

include/rocksdb/trace_record.h
autopear: Done

PR #8613, opened 2021-08-03
Support dynamic sector size in alignment validation for Windows.

burtonli:
- Use dynamic section size when calling IsSectorAligned()
- Support relative path for GetSectorSize().
- Move buffer and sector alignment check to assert for better retail performance.
- Typo fixes.


port/win/io_win.cc
```diff
@@ -680,12 +680,9 @@ inline IOStatus WinRandomAccessImpl::ReadImpl(uint64_t offset, size_t n,
...
   if (file_base_->use_direct_io()) {
-    if (!IsSectorAligned(static_cast<size_t>(offset)) ||
-        !IsAligned(alignment_, scratch)) {
-      return IOStatus::InvalidArgument(
-          "WinRandomAccessImpl::ReadImpl: offset or scratch is not properly "
-          "aligned");
-    }
+    assert(IsSectorAligned(file_base_->GetSectorSize(),
+                           static_cast<size_t>(offset)));
+    assert(IsAligned(alignment_, scratch));
```
mrambacher, line 685: Why are the asserts here better than the old method of returning a bad status?  The assert will not be triggered in non-debug mode, whereas the old code would return bad status regardless of the compilation mode.

port/win/io_win.cc
```diff
@@ -774,14 +771,10 @@ inline IOStatus WinWritableImpl::AppendImpl(const Slice& data) {
...
-    if (!IsSectorAligned(data.size()) ||
-        !IsAligned(static_cast<size_t>(GetAlignement()), data.data())) {
-      s = IOStatus::InvalidArgument(
-          "WriteData must be page aligned, size must be sector aligned");
-    } else {
-      s = pwrite(file_data_, data, next_write_offset_, bytes_written);
-    }
+    assert(IsSectorAligned(file_data_->GetSectorSize(), next_write_offset_));
+    assert(IsSectorAligned(file_data_->GetSectorSize(), data.size()));
+    assert(IsAligned(static_cast<size_t>(GetAlignment()), data.data()));
```
mrambacher, line 776: Same comment as above re: returning a Status versus assert

port/win/io_win.h
```diff
@@ -67,12 +67,17 @@ class WinFileData {
...
   const bool use_direct_io_;
+  const size_t sector_size_;
 
  public:
   // We want this class be usable both for inheritance (prive
   // or protected) and for containment so __ctor and __dtor public
-  WinFileData(const std::string& filename, HANDLE hFile, bool direct_io)
-      : filename_(filename), hFile_(hFile), use_direct_io_(direct_io) {}
+  WinFileData(const std::string& filename, HANDLE hFile, bool direct_io,
+              size_t sector_size)
```
mrambacher, line 76: Rather than passing in the sector_size in the constructor, why not determine it inside the constructor?  Seems simpler and less error-prone.

port/win/io_win.cc
```diff
@@ -624,10 +626,8 @@ IOStatus WinSequentialFile::PositionedRead(uint64_t offset, size_t n,
     return IOStatus::NotSupported("This function is only used for direct_io");
   }
 
-  if (!IsSectorAligned(static_cast<size_t>(offset)) || !IsSectorAligned(n)) {
-    return IOStatus::InvalidArgument(
-        "WinSequentialFile::PositionedRead: offset is not properly aligned");
-  }
+  assert(IsSectorAligned(GetSectorSize(), static_cast<size_t>(offset)));
+  assert(IsSectorAligned(GetSectorSize(), static_cast<size_t>(n)));
```
mrambacher, line 630: Should IsSectorAligned be a member function of WinFileData?

port/win/env_win.cc
```diff
@@ -433,9 +435,10 @@ IOStatus WinFileSystem::OpenWritableFile(
   } else {
     // Here we want the buffer allocation to be aligned by the SSD page size
     // and to be a multiple of it
-    result->reset(new WinWritableFile(
-        fname, hFile, std::max(GetSectorSize(fname), GetPageSize()),
-        c_BufferCapacity, local_options));
+    size_t sector_size = GetSectorSize(fname);
+    result->reset(
+        new WinWritableFile(fname, hFile, std::max(sector_size, GetPageSize()),
+                            c_BufferCapacity, sector_size, local_options));
```
mrambacher, line 441: I might not be following this well, but won't this potentially align based on sector rather than page size?  This might be correct, but is not what the comment indicates.

burtonli: The buffer alignment should go with either sector size or page size, depends on which one is larger. I think the comment was added originally because the it's really rare that sector size is greater.

port/win/io_win.cc
burtonli: similarly like io_poix.cc, using assert provides a slight better performance under retail build. If the size is not aligned, it fails call the system api anyway, so adding an extra checking is not really necessary.

port/win/io_win.cc
burtonli: yes, that would be better.

mrambacher (approved):
LGTM.  Thanks for the fix and addressing my comments!

PR #8616, opened 2021-08-03
Remove corruption error injection in FaultInjectionTestFS

anand1976:
```FaultInjectionTestFS``` injects various types of read errors in ```FileSystem``` APIs. One type of error is corruption errors, where data is intentionally corrupted or truncated. There is corresponding validation in db_stress to verify that an injected error results in a user visible Get/MultiGet error. However, for corruption errors, its hard to know when a corruption is supposed to be detected by the user request, due to prefetching and, in case of direct IO, padding. This results in false positives. So remove that functionality.

Block checksum validation for Get/MultiGet is confined to ```BlockFetcher```, so we don't lose a lot by disabling this since its a small surface area to test.

### DIFF
--- a/utilities/fault_injection_fs.cc
+++ b/utilities/fault_injection_fs.cc
@@ -730,10 +730,9 @@ void FaultInjectionTestFS::UntrackFile(const std::string& f) {
   open_files_.erase(f);
 }
 
-IOStatus FaultInjectionTestFS::InjectThreadSpecificReadError(ErrorOperation /*op*/,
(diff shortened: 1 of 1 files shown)


zhichao-cao:
So for now, we do not corrupt the read content, just return IOError?

zhichao-cao (approved):
LGTM

anand1976:
> So for now, we do not corrupt the read content, just return IOError?

Yes.

PR #8620, opened 2021-08-04
Dynamically configure BlockBasedTableOptions.prepopulate_block_cache

akankshamahajan15:
Summary: Dynamically configure BlockBasedTableOptions.prepopulate_block_cache using DB::SetOptions.

Test Plan: Added new unit test

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/include/rocksdb/utilities/options_type.h
+++ b/include/rocksdb/utilities/options_type.h
@@ -262,54 +262,8 @@ class OptionTypeInfo {
   // @param map The string to enum mapping for this enum
   template <typename T>
   static OptionTypeInfo Enum(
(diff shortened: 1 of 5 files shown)


include/rocksdb/utilities/options_type.h
```diff
@@ -306,6 +306,52 @@ class OptionTypeInfo {
         });
   }  // End OptionTypeInfo::Enum
 
+  template <typename T>
+  static OptionTypeInfo Enum(
```
anand1976, line 310: Would be good to share code with the previous definition of ```Enum()```, since the only addition here is the ```flags``` argument.

anand1976 (approved):
LGTM

PR #8621, opened 2021-08-04
Fix clang failure

akankshamahajan15:
Summary: Fixed clang failure because of memory leak

Test Plan: CircleCI clang job

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/table/block_based/block_based_table_builder.cc
+++ b/table/block_based/block_based_table_builder.cc
@@ -1508,18 +1508,17 @@ Status BlockBasedTableBuilder::InsertBlockInCache(const Slice& block_contents,
             false /*rep_->blocks_definitely_zstd_compressed*/,
             rep_->table_options.filter_policy.get()));
 
-    if (block_holder->own_bytes()) {
-      size_t charge = block_holder->ApproximateMemoryUsage();
-      s = block_cache->Insert(key, block_holder.release(), charge,
-                              &DeleteEntryCached<TBlocklike>);
+    assert(block_holder->own_bytes());
+    size_t charge = block_holder->ApproximateMemoryUsage();
+    s = block_cache->Insert(key, block_holder.release(), charge,
+                            &DeleteEntryCached<TBlocklike>);
 
-      if (s.ok()) {
-        BlockBasedTable::UpdateCacheInsertionMetrics(
-            block_type, nullptr /*get_context*/, charge, s.IsOkOverwritten(),
-            rep_->ioptions.stats);
-      } else {
-        RecordTick(rep_->ioptions.stats, BLOCK_CACHE_ADD_FAILURES);
-      }
+    if (s.ok()) {
+      BlockBasedTable::UpdateCacheInsertionMetrics(
+          block_type, nullptr /*get_context*/, charge, s.IsOkOverwritten(),
+          rep_->ioptions.stats);
+    } else {
+      RecordTick(rep_->ioptions.stats, BLOCK_CACHE_ADD_FAILURES);
     }
   }
   return s;
--- a/table/block_based/block_based_table_builder.cc
+++ b/table/block_based/block_based_table_builder.cc
@@ -1501,15 +1501,16 @@ Status BlockBasedTableBuilder::InsertBlockInCache(const Slice& block_contents,
     const size_t read_amp_bytes_per_bit =
         rep_->table_options.read_amp_bytes_per_bit;
 
-    TBlocklike* block_holder = BlocklikeTraits<TBlocklike>::Create(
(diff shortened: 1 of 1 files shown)


table/block_based/block_based_table_builder.cc
```diff
@@ -1501,15 +1501,16 @@ Status BlockBasedTableBuilder::InsertBlockInCache(const Slice& block_contents,
...
-        false /*rep_->blocks_definitely_zstd_compressed*/,
-        rep_->table_options.filter_policy.get());
+    std::unique_ptr<TBlocklike> block_holder(
+        BlocklikeTraits<TBlocklike>::Create(
+            std::move(results), read_amp_bytes_per_bit,
+            rep_->ioptions.statistics.get(),
+            false /*rep_->blocks_definitely_zstd_compressed*/,
+            rep_->table_options.filter_policy.get()));
 
     if (block_holder->own_bytes()) {
```
pdillinger, line 1511: Seems like this should be an `assert` instead of `if`

table/block_based/block_based_table_builder.cc
```diff
@@ -1501,15 +1501,16 @@ Status BlockBasedTableBuilder::InsertBlockInCache(const Slice& block_contents,
...
       size_t charge = block_holder->ApproximateMemoryUsage();
-      s = block_cache->Insert(key, block_holder, charge,
+      s = block_cache->Insert(key, block_holder.release(), charge,
```
pdillinger, line 1513: It is trickier than this because the block cache only takes ownership on OK return. See calls to `block_cache->Insert()` in block_based_table_reader.cc

table/block_based/block_based_table_builder.cc
```diff
@@ -1501,15 +1501,16 @@ Status BlockBasedTableBuilder::InsertBlockInCache(const Slice& block_contents,
...
                               &DeleteEntryCached<TBlocklike>);
```
pdillinger, line 1514: Also, it looks like this feature was developed alongside block cache entry statistics and secondary cache, so it ended up compatible with neither. (`DeleteEntryCached` should not be used here. See how `block_cache->Insert()` is called in block_based_table_reader.cc)

This is an example of drift that can happen with code duplication, in this case from BlockBasedTable::PutDataBlockToCache, rather than trying to share common code. (`cached_block` could be made optional)

pdillinger (changes requested):

pdillinger:
But OK for this PR to be minimum fix for memory leak

table/block_based/block_based_table_builder.cc
akankshamahajan15: Thanks for the input. I will create the separate PR for BlockBasedTableBuilder to reuse the code of BlockBasedTableReader.

table/block_based/block_based_table_builder.cc
akankshamahajan15: In block_based_table_reader.cc, if ```block_cache_->Insert() ``` returns Not Ok status (failure), it doesn't release the pointer. Does ```insert``` takes care of that?

pdillinger: Here we release unique_ptr on OK (skip automatic delete): https://github.com/facebook/rocksdb/blob/dbe3810c74a0103919ccbcccee52feba7e652866/table/block_based/block_based_table_reader.cc#L1202

And here we delete raw ptr on non-OK:
https://github.com/facebook/rocksdb/blob/dbe3810c74a0103919ccbcccee52feba7e652866/table/block_based/block_based_table_reader.cc#L1295

Either works but unique_ptr is obviously preferred.

table/block_based/block_based_table_builder.cc
```diff
@@ -1501,24 +1501,28 @@ Status BlockBasedTableBuilder::InsertBlockInCache(const Slice& block_contents,
...
+        BlocklikeTraits<TBlocklike>::Create(
+            std::move(results), read_amp_bytes_per_bit,
+            rep_->ioptions.statistics.get(),
+            false /*rep_->blocks_definitely_zstd_compressed*/,
+            rep_->table_options.filter_policy.get()));
+
+    assert(block_holder->own_bytes());
+    size_t charge = block_holder->ApproximateMemoryUsage();
+    s = block_cache->Insert(
+        key, block_holder.get(),
```
pdillinger, line 1516: I don't think it's fixed yet. (Now use-after-free or double-delete on Insert OK, but Cache::DisownData() means it won't be caught unless running under ASAN or Valgrind.)

pdillinger (approved):
👍

PR #8622, opened 2021-08-04
Do not attempt to rename non-existent info log

ajkr:
Previously we attempted to rename "LOG" to "LOG.old.*" without checking
its existence first. "LOG" had no reason to exist in a new DB.

Errors in renaming a non-existent "LOG" were swallowed via
`PermitUncheckedError()` so things worked. However the storage service's
error monitoring was detecting all these benign rename failures. So it
is better to fix it. Also with this PR we can now distinguish rename failure
for other reasons and return them.

Test Plan: new unit test


akankshamahajan15 (approved):
LGTM!. Thanks for the fix.

ajkr:
Thanks for the review!

clang-analyze appears broken on master ( https://app.circleci.com/pipelines/github/facebook/rocksdb/8218/workflows/be7d6ea9-2870-485e-b11e-b8d7e7b9705f/jobs/175012).

akankshamahajan15:
> Thanks for the review!
> 
> clang-analyze appears broken on master ( https://app.circleci.com/pipelines/github/facebook/rocksdb/8218/workflows/be7d6ea9-2870-485e-b11e-b8d7e7b9705f/jobs/175012).

Yeah. I have the PR ready for that https://github.com/facebook/rocksdb/pull/8621. It should be fixed by today.

PR #8624, opened 2021-08-04
Attempt to deflake ObsoleteFilesTest.DeleteObsoleteOptionsFile

ltamasi:
Summary:
We've been seeing occasional crashes on CI while inserting into the
vectors in `ObsoleteFilesTest.DeleteObsoleteOptionsFile`. The crashes
don't reproduce locally (could be either a race or an object lifecycle
issue) but the good news is that the vectors in question are not really
used for anything meaningful by the test. (The assertion about the sizes
of the two vectors being equal is guaranteed to hold, since the two sync
points where they are populated are right after each other.) The patch
simply removes the vectors from the test, alongside the associated
callbacks and sync points.

Test Plan:
`make check`


akankshamahajan15 (approved):
LGTM!

PR #8627, opened 2021-08-05
Attempt to deflake DBTestXactLogIterator.TransactionLogIteratorCorruptedLog

ltamasi:
Summary:
The patch attempts to deflake `DBTestXactLogIterator.TransactionLogIteratorCorruptedLog`
by disabling file deletions while retrieving the list of WAL files and truncating the first WAL file.
This is to prevent the `PurgeObsoleteFiles` call triggered by `GetSortedWalFiles` from
invalidating the result of `GetSortedWalFiles`. The patch also cleans up the test case a bit
and changes it to using `test::TruncateFile` instead of calling the `truncate` syscall directly.

Test Plan:
`make check`

### DIFF
--- a/db/db_log_iter_test.cc
+++ b/db/db_log_iter_test.cc
@@ -191,15 +191,12 @@ TEST_F(DBTestXactLogIterator, TransactionLogIteratorCorruptedLog) {
     ASSERT_OK(dbfull()->Flush(FlushOptions()));
     ASSERT_OK(dbfull()->FlushWAL(false));
     // Corrupt this log to create a gap
-    ROCKSDB_NAMESPACE::VectorLogPtr wal_files;
+    VectorLogPtr wal_files;
     ASSERT_OK(dbfull()->GetSortedWalFiles(wal_files));
+    ASSERT_FALSE(wal_files.empty());
     const auto logfile_path = dbname_ + "/" + wal_files.front()->PathName();
-    if (mem_env_) {
-      mem_env_->Truncate(logfile_path, wal_files.front()->SizeFileBytes() / 2);
-    } else {
-      ASSERT_EQ(0, truncate(logfile_path.c_str(),
-                   wal_files.front()->SizeFileBytes() / 2));
-    }
+    test::TruncateFile(env_, logfile_path,
+                       wal_files.front()->SizeFileBytes() / 2);
 
     // Insert a new entry to a new log file
     ASSERT_OK(Put("key1025", DummyString(10)));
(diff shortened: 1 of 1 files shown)


akankshamahajan15 (approved):
LGTM. Thanks for fixing it.

db/db_log_iter_test.cc
```diff
@@ -191,15 +191,12 @@ TEST_F(DBTestXactLogIterator, TransactionLogIteratorCorruptedLog) {
...
     ASSERT_OK(dbfull()->GetSortedWalFiles(wal_files));
+    ASSERT_FALSE(wal_files.empty());
     const auto logfile_path = dbname_ + "/" + wal_files.front()->PathName();
-    if (mem_env_) {
-      mem_env_->Truncate(logfile_path, wal_files.front()->SizeFileBytes() / 2);
-    } else {
-      ASSERT_EQ(0, truncate(logfile_path.c_str(),
-                   wal_files.front()->SizeFileBytes() / 2));
-    }
+    test::TruncateFile(env_, logfile_path,
```
riversand963, line 198: Should we use `mem_env_` when necessary?
I guess this is currently ok because `SpecialEnv` forwards calls to the underlying mem env. In the future, if `SpecialEnv` impl changes, will this break?

riversand963 (approved):

db/db_log_iter_test.cc
ltamasi: I believe the only reason `mem_env_` had a special branch there because `truncate` only works with actual on-disk files. I think `test::TruncateFile` would cover both cases.

ltamasi:
I believe I've managed to trace the flakiness back to https://github.com/facebook/rocksdb/pull/8591. Based on our discussion with @siying on that PR, I've updated the code to disable file deletions before the `GetSortedWalFiles` call.

@akankshamahajan15 @riversand963 Since there have been some changes since you've approved, could you take another look at the patch?

jay-zhuang:
> I believe I've managed to trace the flakiness back to #8591. Based on our discussion with @siying on that PR, I've updated the code to disable file deletions before the `GetSortedWalFiles` call.
> 
> @akankshamahajan15 @riversand963 Since there have been some changes since you've approved, could you take another look at the patch?

Is not `GetSortedWalFiles()` disable the file deletion internally? :
https://github.com/facebook/rocksdb/blob/e8f218cb6863949a05bc38c8399433d8a0ff3368/db/db_filesnapshot.cc#L133

ltamasi:
> 
> Is not `GetSortedWalFiles()` disable the file deletion internally? :
> https://github.com/facebook/rocksdb/blob/e8f218cb6863949a05bc38c8399433d8a0ff3368/db/db_filesnapshot.cc#L133

Yes but then it re-enables it, which results in an immediate purge, potentially invalidating the results of the call before the caller gets to use them. Disabling/enabling also at the call site prevents this from happening, since `disable_delete_obsolete_files_` is really a counter; i.e. you have to re-enable purge as many times as you have disabled it.

jay-zhuang (approved):

PR #8628, opened 2021-08-06
Memtable sampling for mempurge heuristic.

bjlemaire:
Changes the API of the MemPurge process: the `bool experimental_allow_mempurge` and `experimental_mempurge_policy` flags have been replaced by a `double experimental_mempurge_threshold` option. 
This change of API reflects another major change introduced in this PR: the MemPurgeDecider() function now works by sampling the memtables being flushed to estimate the overall amount of useful payload (payload minus the garbage), and then compare this useful payload estimate with the `double experimental_mempurge_threshold` value.
Therefore, when the value of this flag is `0.0` (default value), mempurge is simply deactivated. On the other hand, a value of `DBL_MAX` would be equivalent to always going through a mempurge regardless of the garbage ratio estimate. 
At the moment, a `double experimental_mempurge_threshold` value else than 0.0 or `DBL_MAX` is opnly supported`with the `SkipList` memtable representation. 
Regarding the sampling, this PR includes the introduction of a `MemTable::UniqueRandomSample` function that collects (approximately) random entries from the memtable by using the new `SkipList::Iterator::RandomSeek()` under the hood, or by iterating through each memtable entry, depending on the target sample size and the total number of entries. 
The unit tests have been readapted to support this new API.

### DIFF
--- a/db/flush_job.cc
+++ b/db/flush_job.cc
@@ -576,12 +576,12 @@ Status FlushJob::MemPurge() {
           cfd_->imm()->RemoveMemPurgeOutputID(mt->GetID());
         }
         new_mem->SetID(new_mem_id);
-        cfd_->imm()->AddMemPurgeOutputID(new_mem_id);
+        new_mem_capacity = (new_mem->ApproximateMemoryUsage()) * 1.0 /
+                           mutable_cf_options_.write_buffer_size;
+        cfd_->imm()->AddMemPurgeOutputID(new_mem_id, new_mem_capacity);
         // This addition will not trigger another flush, because
         // we do not call SchedulePendingFlush().
         cfd_->imm()->Add(new_mem, &job_context_->memtables_to_free);
-        new_mem_capacity = (new_mem->ApproximateMemoryUsage()) * 1.0 /
-                           mutable_cf_options_.write_buffer_size;
         new_mem->Ref();
         db_mutex_->Unlock();
       } else {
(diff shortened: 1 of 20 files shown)


memtable/skiplistrep.cc
```diff
@@ -95,6 +95,37 @@ class SkipListRep : public MemTableRep {
...
+                          std::unordered_set<const char*>* entries) override {
+    entries->clear();
+    SkipListRep::Iterator iter(&skip_list_);
+    // If the sample size is at least 25% of the total number of elements
+    // in the skip list, then simply iterate linearly over the list.
+    // (25% is arbitrary, could sample_size>N/log(N) if we assume iterating
+    // over each sample is O(N) and RandomSeek calls would be
+    // O(sample_size*log(N))).
+    if (sample_size >= (num_entries / 4)) {
+      Random rnd(1234);
```
pdillinger, line 109: Nit: Use GetTLSInstance()

memtable/skiplistrep.cc
```diff
@@ -95,6 +95,37 @@ class SkipListRep : public MemTableRep {
...
+      int n = (sample_size > num_entries)
```
pdillinger, line 110: The best pattern for this problem uses a dynamic threshold on a ranged random value. See https://cs.stackexchange.com/a/129757

memtable/skiplistrep.cc
```diff
@@ -95,6 +95,37 @@ class SkipListRep : public MemTableRep {
...
+      for (iter.SeekToFirst(); iter.Valid(); iter.Next()) {
+        if (rnd.OneIn(n)) {
+          entries->insert(iter.key());
+        }
+      }
+    } else {
+      for (uint64_t i = 0; i < sample_size; i++) {
+        do {
+          iter.RandomSeek();
+        } while (!iter.Valid() ||
```
pdillinger, line 122: What is the !Valid case? Infinite loop bugs are typically worse than crash bugs, so I am immediately skeptical of code that looks susceptible to infinite loop.

memtable/skiplistrep.cc
```diff
@@ -95,6 +95,37 @@ class SkipListRep : public MemTableRep {
...
+                 (entries->find(iter.key()) != entries->end()));
+        entries->insert(iter.key());
```
pdillinger, line 124: I assumed the unordered_set was so that we didn't put the same (by pointer) key in more than once if we happen to sample it more than once. But the loop bound is currently number of items inserted, not number of items ending up in `entries`. (I might be suggesting violating my own rule about risking infinite loops. I'll see what you come up with.)

memtable/skiplistrep.cc
```diff
@@ -95,6 +95,37 @@ class SkipListRep : public MemTableRep {
     return (end_count >= start_count) ? (end_count - start_count) : 0;
   }
 
+  void UniqueRandomSample(const uint64_t& num_entries,
+                          const uint64_t& sample_size,
+                          std::unordered_set<const char*>* entries) override {
+    entries->clear();
+    SkipListRep::Iterator iter(&skip_list_);
+    // If the sample size is at least 25% of the total number of elements
```
pdillinger, line 103: (My threshold guesstimate would be more like ~10%.)

db/flush_job.cc
```diff
@@ -630,6 +630,117 @@ bool FlushJob::MemPurgeDecider() {
...
+                "unrelated to whether it has been deleted or not.");
+          }
+          payload += entry_slice.size();
+
+          // TODO(bjlemaire): evaluate typeMerge.
+          // This is where the sampled entry is estimated to be
+          // garbage or not. Note that this is a garbage *estimation*
+          // because we do not include certain items such as
+          // CompactionFitlers triggered at flush, or if the same delete
+          // has been inserted twice or more in the memtable.
```
pdillinger, line 718: 👍

db/memtable_list.h
```diff
@@ -457,7 +462,7 @@ class MemTableList {
 
   // Store the IDs of the memtables installed in this
   // list that result from a mempurge operation.
```
pdillinger, line 464: What is the map to? (useful_payload_ratio)

db/flush_job.cc
```diff
@@ -630,6 +630,117 @@ bool FlushJob::MemPurgeDecider() {
...
+          ROCKS_LOG_WARN(
+              db_options_.info_log,
+              "Mempurge kSampling policy: null payload measured, and collected "
+              "sample size is %zu\n.",
+              sentries.size());
+        }
+        ROCKS_LOG_INFO(
+            db_options_.info_log,
+            "Mempurge kSampling policy: garbage ratio from sampling: %f.\n",
+            (payload - useful_payload) * 1.0 / payload);
```
pdillinger, line 740: Potential divide by 0, which could be UB https://stackoverflow.com/questions/42926763/the-behaviour-of-floating-point-division-by-zero

include/rocksdb/memtablerep.h
```diff
@@ -194,6 +196,15 @@ class MemTableRep {
     return 0;
   }
 
+  // Returns a vector of unique random memtable entries of size 'sample_size'.
+  virtual void UniqueRandomSample(const uint64_t& num_entries,
+                                  const uint64_t& sample_size,
+                                  std::unordered_set<const char*>* entries) {
+    (void)num_entries;
```
pdillinger, line 203: Seems like there should be an assert(false) or have the caller explicitly check for no entries. The declaration in MemTable should document what an implementation that doesn't support the operation should do.

db/memtable.h
```diff
@@ -145,6 +146,17 @@ class MemTable {
     return approximate_memory_usage_.load(std::memory_order_relaxed);
   }
 
+  // Returns a vector of unique random memtable entries of size 'sample_size'.
+  //
+  // REQUIRES: external synchronization to prevent simultaneous
+  // operations on the same MemTable (unless this Memtable is immutable).
+  void UniqueRandomSample(const uint64_t& sample_size,
+                          std::unordered_set<const char*>* entries) {
```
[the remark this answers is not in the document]
bjlemaire: This is a good remark, I am adding the comment.

memtable/skiplistrep.cc
bjlemaire: Perfect, this is exactly what was needed

memtable/skiplistrep.cc
bjlemaire: Here is my suggestion. I actually plugged some numbers using the dynamic thresholding, and when we choose this sampling method the likelihood of inserting twice the same element are completely negligible. (I added more comments in there). 
In brief, the probability of having at least one duplicate is P = 1-(N!/((N-m)! * N^m)) (N # of elements in table, and m the `target_sample_size`). For N=1000, since the max `m` is going to be m=1 ( because sqrt(N)/10 = 1), we get P=0.1% . This value goes up to 0.5% for N=10**10 entries, which wont ever happen.
So I am not checking anything anymore, and the unsorted_set will take care of potential duplicates. I am also renaming the variable 'target_sample_size' instead of 'sample_size' because it is not strictly enforced (note that it isn't strictly enforced in the other sampling method either). I am also making this clear in the comments.

memtable/skiplistrep.cc
bjlemaire: The !valid test actually should never happen, because as long as the `memtable` has at least one element (I added an assert for that), `iter` will point to it.

memtable/skiplistrep.cc
bjlemaire: I am now giving it 5 attempts to find a non-duplicate entry. 
At worst, for the final pick , when m=N/10 there is 10% chances to find a duplicate. So chances to fail finding a non-duplicate after 5 attempts is 0.001%.

memtable/skiplistrep.cc
bjlemaire: I am going for 10% - the `sample_size`<<sqrt(`num_entries`) from the reference https://cs.stackexchange.com/a/129757 is not satifying for our situation (if N=1000, then `sample_size` of 2 would redirect to the first sampling algorithm and not the random picks...)

memtable/skiplistrep.cc
bjlemaire: Actually, I am going for 10% of num_entries- the sample_size<<sqrt(num_entries) from the reference https://cs.stackexchange.com/a/129757 is not satifying for our situation (if N=1000, then sample_size of 2 would redirect to the first sampling algorithm and not the random picks...)

PR #8632, opened 2021-08-06
Fix db_stress failure

akankshamahajan15:
Summary: FaultInjectionTestFS injects error in Rename operation. Because
of injected error, info.log fails to be created if rename  returns error and info_log is set to nullptr which leads to this assertion

Test Plan: run the db_stress job locally

Reviewers:

Subscribers:

Tasks:

Tags:


ajkr (approved):
LGTM, thanks!

PR #8634, opened 2021-08-06
Upgrade xxhash, add Hash128

pdillinger:
Summary: With expected use for a 128-bit hash, xxhash library is
upgraded to current dev (2c611a76f914828bed675f0f342d6c4199ffee1e)
as of Aug 6 so that we can use production version of XXH3_128bits
as new Hash128 function (added in hash128.h).

To make this work, however, we have to carve out the "preview" version
of XXH3 that is used in new SST Bloom and Ribbon filters, since that
will not get maintenance in xxhash releases. I have consolidated all the
relevant code into xxph3.h and made it "inline only" (no .cc file). The
working name for this hash function is changed from XXH3p to XXPH3
(XX Preview Hash) because the latter is easier to get working with no
symbol name conflicts between the headers.

Test Plan: no expected change in existing functionality. For Hash128,
added some unit tests based on those for Hash64 to ensure some basic
properties and that the values do not change accidentally.


pdillinger:
Argh, forgot about unity build

pdillinger:
(format check expected to fail because of outside source code)

zhichao-cao (approved):
LGTM, should we mention it in the HISTORY.md?

pdillinger:
> LGTM, should we mention it in the HISTORY.md?

There should be no user-visible change. :)

PR #8635, opened 2021-08-07
Prevent joining detached thread in ThreadPoolImpl

ajkr:
This draining mechanism should not be run during `JoinThreads()` because it can detach threads that will be joined. Joining detached threads would throw an exception.

With this PR, we skip draining when `JoinThreads()` has already decided what threads to `join()`, so the threads will exit naturally once the work queue empties.

Test Plan: verified it unblocked using `WaitForJobsAndJoinAllThreads()` in #8611.

### DIFF
--- a/util/threadpool_imp.cc
+++ b/util/threadpool_imp.cc
@@ -200,9 +200,7 @@ void ThreadPoolImpl::Impl::BGThread(size_t thread_id) {
           queue_.empty()) {
         break;
        }
-    }
-
-    if (!exit_all_threads_ && IsLastExcessiveThread(thread_id)) {
+    } else if (IsLastExcessiveThread(thread_id)) {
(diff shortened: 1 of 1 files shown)


util/threadpool_imp.cc
```diff
@@ -202,10 +202,12 @@ void ThreadPoolImpl::Impl::BGThread(size_t thread_id) {
        }
     }
 
-    if (IsLastExcessiveThread(thread_id)) {
+    if (!exit_all_threads_ && IsLastExcessiveThread(thread_id)) {
```
riversand963, line 205: Can this be an `else if (IsLastExcessiveThread())` branch together with the preceding `if (exit_all_threads_)`?

ajkr: Yes, done.

riversand963 (approved):

PR #8636, opened 2021-08-09
Make SystemClock into a Customizable Class

mrambacher:
Made SystemClock into a Customizable class, complete with CreateFromString.

Cleaned up some of the existing SystemClock implementations that were redundant (NoSleep was the same as the internal one for MockEnv).

Changed MockEnv construction to allow Clock to be passed to the Memory/MockFileSystem.


include/rocksdb/system_clock.h
```diff
@@ -23,15 +25,19 @@ struct ConfigOptions;
...
 
   static const char* Type() { return "SystemClock"; }
-
+  static Status CreateFromString(const ConfigOptions& options,
+                                 const std::string& value,
+                                 std::shared_ptr<SystemClock>* result);
   // The name of this system clock
   virtual const char* Name() const = 0;
 
+  static const char* kDefaultName() { return "DefaultClock"; }
```
zhichao-cao, line 39: Why we need the kDefaultName() here?

mrambacher: There are two classes (Posix and Windows) which are both default, so both of these classes return true for IsInstanceOf(kDefaultName).

Additionally, you can get an instance of the default system clock via SystemClock::CreateFromString(kDefaultName, ...)

db/blob/blob_file_builder_test.cc
```diff
@@ -108,7 +109,7 @@ class BlobFileBuilderTest : public testing::Test {
     ASSERT_EQ(footer.expiration_range, ExpirationRange());
   }
 
-  MockEnv mock_env_;
```
zhichao-cao, line 111: Any essential need to change to std::unique_ptr<Env>?

env/mock_env.h
```diff
@@ -16,20 +16,51 @@
...
+#include "rocksdb/system_clock.h"
 
 namespace ROCKSDB_NAMESPACE {
+// A SystemClock that fakes sleeps.
+class FakeSleepSystemClock : public SystemClockWrapper {
+ public:
+  explicit FakeSleepSystemClock(const std::shared_ptr<SystemClock>& c)
+      : SystemClockWrapper(c), fake_sleep_micros_(0) {}
+
+  void SleepForMicroseconds(int micros) override {
```
zhichao-cao, line 28: I may use "FakeSleepForMicroseconds" here and also rename the "MockSleepForMicroseconds" to "FakeSleepForMicroseconds", what do you think. Otherwise, the name itself does not provide the information that we "Fake" the sleep action.

mrambacher: The class is FakeSleepSystemClock.  It is meant to override a "Sleep" with a fake.  I am not sure how the renaming does what you want?

db/blob/blob_file_builder_test.cc
mrambacher: It allows me to make the MockEnv constructor private.  The MockFileSystem requires a Clock at the time it is being created, which requires two steps.

env/mock_env.h
zhichao-cao: Right, but after user has something like std<FakeSleepSystemClock> clock, the will just will clock->SleepForMicroseconds, which can be misleading in the code. I may prefer we keep the "Fake" in the function name.

mrambacher: This is meant to be a test class that simulates a sleep by just keeping track of the delay and not actually sleeping.  If the internal code calls SleepForMicroseconds and this method is called "FakeSleepForMicroseconds", this method/class will have no effect.  

The point of this class is to NOT do a real Sleep when the code (internal or test) calls Sleep.

The SpecialEnv class does something similar to sometimes (but not always) override the true sleep.

zhichao-cao (approved):
LGTM, thanks!

env/mock_env.cc
```diff
@@ -1034,57 +1100,30 @@ Status MockFileSystem::CorruptBuffer(const std::string& fname) {
...
- public:
-  explicit MockSystemClock(const std::shared_ptr<SystemClock>& c)
-      : SystemClockWrapper(c), fake_sleep_micros_(0) {}
-
-  void FakeSleepForMicroseconds(int64_t micros) {
-    fake_sleep_micros_.fetch_add(micros);
-  }
 
-  const char* Name() const override { return "MockSystemClock"; }
+MockEnv::MockEnv(Env* env, const std::shared_ptr<FileSystem>& fs,
```
zhichao-cao, line 1104: The new TimeElapesSystemClock class looks good to me in general (the name may not be that appropriate......). Can you describe what's the testing failure when using MockSystemClock and why it is resolved in current implementation?

mrambacher: The TimeElapseSystemClock was extracted from SpecialEnv.  Feel free to suggest a better name.  SpecialClock sounded nondescript and there is another class already called MockSystemClock.

There were some failures with the "FakeSystemClock" when running under MEM_ENV.  The problem is that the SpecialEnv would do a "Sleep" and defer to the wrapper clock, which in this case was a FakeSystemClock that did a fake sleep.  Tests that were depending on a real sleep would therefore fail.  I believe the failures were typically around the rate limiter tests but my recollection may be off.

env/time_elapse_clock.h
```diff
@@ -0,0 +1,114 @@
...
+
+#include <atomic>
+#include <string>
+
+#include "rocksdb/status.h"
+#include "rocksdb/system_clock.h"
+
+namespace ROCKSDB_NAMESPACE {
+// A SystemClock that can "mock" sleep and counts its operations.
+class TimeElapseSystemClock : public SystemClockWrapper {
```
zhichao-cao, line 19: how about "TimeEmulatedSystemClock"?

env/time_elapse_clock.h
```diff
@@ -0,0 +1,114 @@
...
+    // We cannot set these before destroying the last DB because they might
+    // cause a deadlock or similar without the appropriate options set in
+    // the DB.
+    time_elapse_only_sleep_ = enabled;
+    no_slowdown_ = enabled;
+  }
+
+  bool IsTimeElapseOnlySleep() const { return time_elapse_only_sleep_.load(); }
+  void SetMockSleep(bool enabled = true) { no_slowdown_ = enabled; }
+  bool UseMockSleep() const { return no_slowdown_; }
```
zhichao-cao, line 69: "UseMockSleep()" -> "IsMockSleepUsed()"

zhichao-cao (approved):
LGTM, do remember to mention in HISTORY.md that customizable class is extended to SystemClock. And also, even EmulatedSystemClock is not a public API, we may need to inform user this new class.

PR #8638, opened 2021-08-09
Make WalFilter, SstPartitionerFactory, FileChecksumGenFactory, and TableProperties Customizable

mrambacher:


include/rocksdb/utilities/customizable_util.h
```diff
@@ -68,6 +68,9 @@ static Status NewSharedObject(
     if (config_options.ignore_unsupported_options && status.IsNotSupported()) {
       return Status::OK();
     }
+  } else if (opt_map.empty()) {
```
zhichao-cao, line 71: Add the logic change to the comment. Previously, if id is empty, we will return not supported. But now, we will reset the `result` to nullptr. Also, can you explain a little bit why we need this change?

utilities/wal_filter.cc
```diff
@@ -0,0 +1,23 @@
...
+#include "rocksdb/wal_filter.h"
+
+#include <memory>
+
+#include "rocksdb/convenience.h"
+#include "rocksdb/options.h"
+#include "rocksdb/utilities/customizable_util.h"
+
+namespace ROCKSDB_NAMESPACE {
+Status WalFilter::CreateFromString(const ConfigOptions& config_options,
```
zhichao-cao, line 15: If it is just the CreateFromString, can we move it to wal_filter.h? Declaration+definition.

test_util/testutil.cc
```diff
@@ -601,5 +602,41 @@ Status CreateEnvFromSystem(const ConfigOptions& config_options, Env** result,
   }
 }
 
+#ifndef ROCKSDB_LITE
+// This method loads existing test classes into the ObjectRegistry
+int RegisterTestObjects(ObjectLibrary& library, const std::string& /*arg*/) {
```
zhichao-cao, line 607: Is this going to be used outside customizable_test.cc?

util/file_checksum_helper.cc
```diff
@@ -133,4 +134,16 @@ Status GetFileChecksumsFromManifest(Env* src_env, const std::string& abs_path,
   return retriever.status();
 }
 
+Status FileChecksumGenFactory::CreateFromString(
```
zhichao-cao, line 137: Why don't we add the RegisterFileChecksumFactories here?

utilities/table_properties_collectors/compact_on_deletion_collector.cc
```diff
@@ -118,5 +128,34 @@ NewCompactOnDeletionCollectorFactory(size_t sliding_window_size,
...
-}  // namespace ROCKSDB_NAMESPACE
+namespace {
+static int RegisterTablePropertiesCollectorFactories(
+    ObjectLibrary& library, const std::string& /*arg*/) {
+  library.Register<TablePropertiesCollectorFactory>(
+      CompactOnDeletionCollectorFactory::kClassName(),
+      [](const std::string& /*uri*/,
+         std::unique_ptr<TablePropertiesCollectorFactory>* guard,
+         std::string* /* errmsg */) {
+        guard->reset(new CompactOnDeletionCollectorFactory(0, 0, 0));
```
zhichao-cao, line 139: Is "0, 0, 0" a correct default initialization argument?

mrambacher: > Is "0, 0, 0" a correct default initialization argument?

I missed putting in OptionType parameters for the CompactOnDeletionCollectorFactory earlier -- since corrected.

As far as defaults, the first and last 0 (window size, deletion ratio) should effectively disable this collector.  The middle argument (deletion trigger) really be set to a meaningful value for the user -- we could make a guess as to a different default but I do not know what that should reasonably be.

utilities/wal_filter.cc
mrambacher: > If it is just the CreateFromString, can we move it to wal_filter.h? Declaration+definition.

It should really be here.  If there is a default or "shipped" implementation later, it should be registered here.

utilities/table_properties_collectors/compact_on_deletion_collector.cc
zhichao-cao: Can we avoid this default construction here? I'm not comfortable with the "0,0,0" arguments. Or maybe, at least, you should add the comments say  by default, the collector is disabled since we give 0 window size.

mrambacher: Added comment to the code.

zhichao-cao (approved):
LGTM, also update the history.md. Thanks for addressing the review comments.

PR #7737, opened 2020-12-02
Add more tests to ASSERT_STATUS_CHECKED (5) 

adamretter:
Fifth batch of adding more tests to ASSERT_STATUS_CHECKED.

* db_test
* db_test2

Requires https://github.com/facebook/rocksdb/pull/7819 for all tests to pass.

PR #8640, opened 2021-08-09
Add db_test2 to to ASSERT_STATUS_CHECKED

adamretter:
This is the `db_test2` parts of https://github.com/facebook/rocksdb/pull/7737 reworked on the latest HEAD.


trace_replay/trace_replay.cc
```diff
@@ -599,13 +606,17 @@ Status Replayer::Replay() {
...
       } else {
         single_iter = db_->NewIterator(roptions, cf_map_[iter_payload.cf_id]);
       }
       single_iter->SeekForPrev(iter_payload.iter_key);
-      ops++;
+      s = single_iter->status();
+      if (!s.ok()) {
+        break;
+      }
       delete single_iter;
```
mrambacher, line 620: Doesn't the single_iter delete need to come before this line?

trace_replay/trace_replay.cc
```diff
@@ -575,13 +578,17 @@ Status Replayer::Replay() {
...
       } else {
         single_iter = db_->NewIterator(roptions, cf_map_[iter_payload.cf_id]);
       }
       single_iter->Seek(iter_payload.iter_key);
-      ops++;
+      s = single_iter->status();
+      if (!s.ok()) {
+        break;
+      }
       delete single_iter;
```
mrambacher, line 592: Doesn't the delete need to come before the break?

file/filename.cc
```diff
@@ -395,7 +395,7 @@ IOStatus SetCurrentFile(FileSystem* fs, const std::string& dbname,
       s = directory_to_fsync->Fsync(IOOptions(), nullptr);
     }
   } else {
-    fs->DeleteFile(tmp, IOOptions(), nullptr);
+    fs->DeleteFile(tmp, IOOptions(), nullptr).PermitUncheckedError();
```
mrambacher, line 398: Please add a comment as to why this is okay (because we are handling an error case already).

db/db_test_util.h
```diff
@@ -403,7 +403,7 @@ class SpecialEnv : public EnvWrapper {
       Status Sync() override {
         ++env_->sync_counter_;
         if (env_->corrupt_in_sync_) {
-          Append(std::string(33000, ' '));
+          Append(std::string(33000, ' ')).PermitUncheckedError();
```
mrambacher, line 406: EXPECT_OK rather than PermitUncheckedError?

trace_replay/trace_replay.cc
```diff
@@ -529,14 +528,19 @@ Status Replayer::Replay() {
     if (trace.type == kTraceWrite) {
       if (trace_file_version_ < 2) {
         WriteBatch batch(trace.payload);
-        db_->Write(woptions, &batch);
+        s = db_->Write(woptions, &batch);
+        if (!s.ok()) {
+          break;
+        }
```
mrambacher, line 534: Are all these break statements really necessary?  Isn't this all inside a "while s.ok" loop?

adamretter: As the body of the loop is quite large - I do think it actually makes the intention much clearer

db/db_test_util.h
adamretter: Okay, I have changed that now.

file/filename.cc
adamretter: Thanks, have done.

trace_replay/trace_replay.cc
adamretter: It needs to also be present in the body of the if expression. I have now fixed this

trace_replay/trace_replay.cc
adamretter: Ah good catch! It needs to also be present in the body of the if expression. I have now fixed this

trace_replay/trace_replay.cc
```diff
@@ -575,13 +578,18 @@ Status Replayer::Replay() {
...
         single_iter = db_->NewIterator(roptions, cf_map_[iter_payload.cf_id]);
       }
       single_iter->Seek(iter_payload.iter_key);
-      ops++;
+      s = single_iter->status();
+      if (!s.ok()) {
+        delete single_iter;
+        break;
+      }
       delete single_iter;
```
mrambacher, line 593: You have a delete in the if and the else (outside the break) statement.  Might as well move the delete before the if.

trace_replay/trace_replay.cc
```diff
@@ -599,13 +607,18 @@ Status Replayer::Replay() {
...
         single_iter = db_->NewIterator(roptions, cf_map_[iter_payload.cf_id]);
       }
       single_iter->SeekForPrev(iter_payload.iter_key);
-      ops++;
+      s = single_iter->status();
+      if (!s.ok()) {
+        delete single_iter;
+        break;
+      }
       delete single_iter;
```
mrambacher, line 622: Same comment as above.

db/db_test2.cc
```diff
@@ -891,8 +894,12 @@ TEST_F(DBTest2, WalFilterTestWithChangeBatch) {
                                   bool* batch_changed) const override {
       if (current_record_index_ >= change_records_from_index_) {
         ChangeBatchHandler handler(new_batch, num_keys_to_add_in_new_batch_);
-        batch.Iterate(&handler);
-        *batch_changed = true;
+        Status s = batch.Iterate(&handler);
+        if (s.ok()) {
+          *batch_changed = true;
+        } else {
+          assert(false);
```
mrambacher, line 901: Should this be assert(false) or FAIL()?

db/db_test2.cc
```diff
@@ -985,8 +992,12 @@ TEST_F(DBTest2, WalFilterTestWithChangeBatchExtraKeys) {
    WalProcessingOption LogRecord(const WriteBatch& batch, WriteBatch* new_batch,
                                  bool* batch_changed) const override {
      *new_batch = batch;
-     new_batch->Put("key_extra", "value_extra");
-     *batch_changed = true;
+     Status s = new_batch->Put("key_extra", "value_extra");
+     if (s.ok()) {
+       *batch_changed = true;
+     } else {
+       assert(false);
```
mrambacher, line 999: or FAIL?

db/db_test2.cc
```diff
@@ -4205,6 +4255,7 @@ TEST_F(DBTest2, TraceAndReplay) {
   ASSERT_OK(db_->Write(wo, &batch));
 
   single_iter = db_->NewIterator(ro);
+  ASSERT_OK(single_iter->status());
```
mrambacher, line 4258: Should this line be immediately before he delete rather than here?

db/db_test2.cc
```diff
@@ -5581,6 +5650,7 @@ TEST_F(DBTest2, ChangePrefixExtractor) {
     // Re-execute similar queries after a full compaction
     {
       std::unique_ptr<Iterator> iterator(db_->NewIterator(ReadOptions()));
+      ASSERT_OK(iterator->status());
```
mrambacher, line 5653: Please move this either after the Valid() or before the iterator is deleted.

db/db_test2.cc
```diff
@@ -5605,6 +5675,7 @@ TEST_F(DBTest2, ChangePrefixExtractor) {
     }
     {
       std::unique_ptr<Iterator> iterator(db_->NewIterator(ro));
+      ASSERT_OK(iterator->status());
```
mrambacher, line 5678: Please move to right before/after Valid check.

db/db_test2.cc
```diff
@@ -5625,6 +5696,7 @@ TEST_F(DBTest2, ChangePrefixExtractor) {
     ub = Slice(ub_str);
     {
       std::unique_ptr<Iterator> iterator(db_->NewIterator(ro));
+      ASSERT_OK(iterator->status());
```
mrambacher, line 5699: Please move to right before/after Valid check.

mrambacher (approved):
This looks pretty good (and a lot of work).  Thanks!

Some minor comments mainly.

db/db_test2.cc
adamretter: `FAIL` can't be used here which is why I went with `assert(false)`:

```
db/db_test2.cc:999:8: error: cannot initialize return object of type 'rocksdb::WalFilter::WalProcessingOption' with an rvalue of type 'void'
       FAIL();
       ^~~~~~
```

db/db_test2.cc
adamretter: Done.

db/db_test2.cc
adamretter: Done.

db/db_test2.cc
adamretter: Done.

db/db_test2.cc
adamretter: `FAIL` can't be used here which is why I went with `assert(false)`:
```
db/db_test2.cc:901:8: error: cannot initialize return object of type 'rocksdb::WalFilter::WalProcessingOption' with an rvalue of type 'void'
       FAIL();
       ^~~~~~
```

trace_replay/trace_replay.cc
adamretter: Done.

trace_replay/trace_replay.cc
adamretter: Done.

akankshamahajan15 (approved):
LGTM. Thanks for working on this.

PR #8641, opened 2021-08-10
Make SliceTransform into a Customizable class

mrambacher:
Made SliceTransform into a Customizable class.

Would be nice to write a test that stored and used a custom transform  in an SST table.

There are a set of tests (DBBlockFliterTest.PrefixExtractor*, SamePrefixTest.InDomainTest, PrefixTest.PrefixAndWholeKeyTest that run the same with or without a SliceTransform/PrefixFilter.  Is this expected?


util/slice.cc
```diff
@@ -60,19 +75,19 @@ class FixedPrefixTransform : public SliceTransform {
...
-        // changes on options_helper in order to make RocksDBOptionsParser work
-        // for the new change.
-        // TODO(yhchiang): move serialization / deserializaion code inside
-        // the class implementation itself.
-        name_("rocksdb.CappedPrefix." + ToString(cap_len_)) {}
+  explicit CappedPrefixTransform(size_t cap_len) : cap_len_(cap_len) {
+    RegisterOptions(Name(), &cap_len_, &slice_transform_length_info);
+  }
 
-  const char* Name() const override { return name_.c_str(); }
```
zhichao-cao, line 75: Is there any backward compatibility issue when Name() is used here. In the legacy code, we will get "rocksdb.CappedPrefix.(cap_len)" But now, we only have `rocksdb.CappedPrefix`. I'm not sure if user will change them to `GetId()`

zhichao-cao (commented):
In general, it looks good

util/slice.cc
mrambacher: I do not think this will be an issue.  First, I believe the Name() method was primarily used for Dump and related functions.  I did miss one place in Dump that I updated and will update HISTORY with this change.

zhichao-cao (approved):
LGTM, do remember to mention the change in HISTORY.md

PR #8650, opened 2021-08-11
Re-add retired mempurge flag definitions for legacy-options-file temporary support.

bjlemaire:
Current internal regression tests pass in an old option flag `experimental_allow_mempurge` to a more recently built db.
This flag was retired and removed in a recent PR (#8628), and therefore, the following error comes up : `Failed: Invalid argument: Could not find option: : experimental_allow_mempurge`. 
In this PR, I reintroduce the two flags retired in #8628, `experimental_allow_mempurge` and `experimental_mempurge_policy` in `db_options.cc` and mark them both as `kDeprecated`.
This is a temporary fix to save us time to find a long term solution, which hopefully will consist in ignoring options prefixed with `experimental_` that are no longer recognized.


pdillinger:
The problem:

    Failed: Invalid argument: Could not find option: : experimental_allow_mempurge

I don't think this fixes the problem:

    $ ./db_bench --experimental_allow_mempurge=true
    ERROR: unknown command line flag 'experimental_allow_mempurge'

bjlemaire:
> The problem:
> 
> ```
> Failed: Invalid argument: Could not find option: : experimental_allow_mempurge
> ```
> 
> I don't think this fixes the problem:
> 
> ```
> $ ./db_bench --experimental_allow_mempurge=true
> ERROR: unknown command line flag 'experimental_allow_mempurge'
> ```



I believe what you want to try is to put the flag in an option file and load the option flag.
For example, with the options file `myopt`:
```
[Version]
  rocksdb_version=4.3.0
  options_file_version=1.1
[DBOptions]
  experimental_allow_mempurge=true
```

and the folowing command line:
```
$ ./db_bench --options_file=myopt
```
then you obtain the following output:
```
Unable to load options file myopt --- Invalid argument: Could not find option: : experimental_allow_mempurge
```

I assumed that some piping happens and somehow only the `-- Invalid argument: Could not find option: : experimental_allow_mempurge` is reported by the internal test log.

pdillinger:
Yes, it's looking for these: https://github.com/bjlemaire/rocksdb/blob/ab7f7c9e497a18c48837556bf2e68b4cc5d2a498/options/db_options.cc#L200-L207

You can see in there there's a way to mark deprecated

pdillinger (approved):
Please fix the PR summary (& re-import or also fix in Phabricator)

This should have been caught by check_format_compatible.sh, but it hasn't been updated for a few releases (!! - I'm working on a PR): https://app.circleci.com/pipelines/github/facebook/rocksdb/8359/workflows/909f1ac6-2fc6-4202-b553-43f1b42c20b9/jobs/180140
And that should suffice as a regression test.

bjlemaire:
I'm also looking at ignoring options prefixed with `experimental_` that are no longer recognized.

zhichao-cao (approved):

PR #8656, opened 2021-08-12
Improve MemPurge sampling

bjlemaire:
Previously, the `MemPurge` sampling function was assessing whether a random entry from a memtable was garbage or not by simply querying the given memtable (see #8628 for more details). 
In this diff, I am updating the sampling function by querying not only the memtable the entry was drawn from, but also all subsequent memtables that have a greater memtable ID. 
I also added the size of the value for KV entries in the payload/useful payload estimates (which was also one of the reasons why sampling was not as good as mempurging all the time in terms of L0 SST files reduction).
Once these changes were made, I was able to clean obsolete objects and functions from the `MemtableList` struct, and did a bit of cleanup everywhere.

### DIFF
--- a/db/memtable_list.cc
+++ b/db/memtable_list.cc
@@ -5,10 +5,12 @@
 //
 #include "db/memtable_list.h"
 
+#include <algorithm>
 #include <cinttypes>
 #include <limits>
 #include <queue>
 #include <string>
+
 #include "db/db_impl/db_impl.h"
 #include "db/memtable.h"
 #include "db/range_tombstone_fragmenter.h"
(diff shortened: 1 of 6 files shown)


db/memtable_list.cc
```diff
@@ -361,6 +371,15 @@ void MemTableList::PickMemtablesToFlush(uint64_t max_memtable_id,
   if (!atomic_flush || num_flush_not_started_ == 0) {
     flush_requested_ = false;  // start-flush request is complete
   }
+
+  // Sort the list of memtables by increasing memtable ID.
+  // This is useful when the mempurge feature is activated
+  // and the memtables are not guaranteed to be sorted in
+  // the memlist vector.
+  std::sort(ret->begin(), ret->end(),
```
pdillinger, line 379: Based on caller, I believe this happens under DB mutex. Probably not a big deal, but noting that in case it can be delayed or something. Or should we keep the source list sorted?

pdillinger (approved):
LGTM. Simplifying some pieces == good.

It would be nice to eventually see more unit tests for heuristic behavior (explicitly seed randomness if necessary), but it's not yet high priority as it doesn't affect data correctness.

db/memtable_list.cc
bjlemaire: I'm torn on this because keeping the source list sorted would also imply iterating through the source list with the DB mutex. Overall this `ret` array is extremely small in practice (no way there is more than 10 pointers in it at a time... more than 2 would honestly be surprising), so std::sort would be lightning fast.
Another option would be to use another data structure to store the immutable memtables (to use an ordered set for example), but I feel like this is a much more invasive change and that change may bring its own set of challenges.

PR #8657, opened 2021-08-12
Allow Replayer to report the results of TraceRecords.

autopear:
`Replayer::Execute()` can directly returns the result (e.g, request latency, DB::Get() return code, returned value, etc.)
`Replayer::Replay()` reports the results via a callback function.

New interface:
`TraceRecordResult` in "rocksdb/trace_record_result.h".

`DBTest2.TraceAndReplay` and `DBTest2.TraceAndManualReplay` are updated accordingly.


include/rocksdb/trace_record_result.h
```diff
@@ -0,0 +1,145 @@
...
+  virtual std::vector<Status> GetMultiStatus() const;
+
+  // Values for the searched keys.
+  virtual std::vector<std::string> GetValues() const;
+
+  virtual Status Accept(Handler* handler) override;
+
+ private:
+  std::vector<Status> multi_status_;
+  std::vector<std::string> values_;
```
ajkr, line 142: I guess you want this to be able to present bytes/sec. I think we are OK starting with QPS, though.

include/rocksdb/trace_record_result.h
```diff
@@ -0,0 +1,145 @@
...
+ public:
+  TraceExecutionResult(uint64_t latency, TraceType trace_type);
+
+  virtual ~TraceExecutionResult() override;
+
+  // Request latency in nanoseconds.
+  virtual uint64_t GetLatency() const;
+
+ private:
+  uint64_t latency_;
```
ajkr, line 71: Sorry I forget, did we discuss what is the benefit of moving latency measurement to RocksDB? In any case it'd be good to write that in the description.

include/rocksdb/trace_record_result.h
autopear: This might be used for bytes/sec, but it's not the major goal. Currently you can get the TraceRecord execution status from `Replayer::Next` and `Replayer::Execute`, but you have no way to obtain the actual DB execution status from these 2 APIs. Also you cannot get any execution status from `Replayer::Replay`. 

The `TraceRecordResult` API set provides an interface to query the execution result from `TraceRecord::Handler`. If it is the execution handler, then you can get the DB calls returned status and/or values.

If we use Handlers for TraceAnalyzer as well, `TraceRecordResult` may be used for parsing or outputting to stdout or file.

include/rocksdb/trace_record_result.h
autopear: Actually I feel it's better to report the start/end timestamp of the execution. Then the user can use these values to order the execution result for analysis, or compute latency, etc for other types of analysis.

include/rocksdb/trace_record_result.h
```diff
@@ -0,0 +1,145 @@
...
+ private:
+  TraceType trace_type_;
+};
+
+// Base class for the execution results types of trace records.
+class TraceExecutionResult : public TraceRecordResult {
+ public:
+  TraceExecutionResult(uint64_t latency, TraceType trace_type);
+
+  virtual ~TraceExecutionResult() override;
```
jay-zhuang, line 65: nit: `virtual` is not needed here as it already has override.
https://google.github.io/styleguide/cppguide.html#Inheritance
> Do not use virtual when declaring an override.

include/rocksdb/utilities/replayer.h
```diff
@@ -61,13 +65,40 @@ class Replayer {
...
+    return Execute(std::move(record), nullptr);
+  }
 
   // Replay all the traces from the provided trace stream, taking the delay
   // between the traces into consideration.
-  virtual Status Replay(const ReplayOptions& options) = 0;
-  virtual Status Replay() { return Replay(ReplayOptions()); }
+  //
+  // result_callback reports the status of executing a trace record, and the
+  // actual operation execution result (See the desciption for Execute()).
```
jay-zhuang, line 88: typo: description

trace_replay/trace_record_result.cc
```diff
@@ -0,0 +1,98 @@
...
+
+#include "rocksdb/trace_record_result.h"
+
+namespace ROCKSDB_NAMESPACE {
+
+// TraceRecordResult
+TraceRecordResult::TraceRecordResult(TraceType trace_type)
+    : trace_type_(trace_type) {}
+
+TraceRecordResult::~TraceRecordResult() {}
```
jay-zhuang, line 14: nit: use `=default`?

trace_replay/trace_record_result.cc
```diff
@@ -0,0 +1,98 @@
...
+Status StatusOnlyTraceExecutionResult::GetStatus() const { return status_; }
+
+Status StatusOnlyTraceExecutionResult::Accept(Handler* handler) {
+  assert(handler != nullptr);
+  return handler->Handle(*this);
+}
+
+// SingleValueTraceExecutionResult
+SingleValueTraceExecutionResult::SingleValueTraceExecutionResult(
+    const Status& status, const std::string& value, uint64_t latency,
```
jay-zhuang, line 43: nit: pass by value then move:
```
SingleValueTraceExecutionResult::SingleValueTraceExecutionResult(
    Status status, std::string value, uint64_t latency,
    TraceType trace_type)
    : TraceExecutionResult(latency, trace_type),
      status_(std::move(status)),
      value_(std::move(value)) {}
```

db/db_test2.cc
```diff
@@ -4458,48 +4615,84 @@ TEST_F(DBTest2, TraceAndManualReplay) {
...
   // Get mixed of existing and non-existing keys, should still return
   // Status::OK().
   record.reset(new MultiGetQueryTraceRecord(
       std::vector<uint32_t>({handles[0]->GetID(), handles[1]->GetID()}),
       std::vector<std::string>({"a", "no2"}), fake_ts++));
-  ASSERT_OK(replayer->Execute(std::move(record)));
+  ASSERT_OK(replayer->Execute(std::move(record), &result));
+  ASSERT_TRUE(result != nullptr);
+  MultiValuesTraceExecutionResult* mvr =
+      reinterpret_cast<MultiValuesTraceExecutionResult*>(result.get());
```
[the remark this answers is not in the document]
autopear: Done

include/rocksdb/trace_record.h
```diff
@@ -41,40 +42,77 @@ enum TraceType : char {
...
 class GetQueryTraceRecord;
 class IteratorSeekQueryTraceRecord;
 class MultiGetQueryTraceRecord;
+class TraceRecordResult;
+class WriteQueryTraceRecord;
 
 // Base class for all types of trace records.
 class TraceRecord {
  public:
   TraceRecord();
```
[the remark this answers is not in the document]
autopear: Done

include/rocksdb/trace_record_result.h
```diff
@@ -0,0 +1,145 @@
...
+  explicit TraceRecordResult(TraceType trace_type);
+
+  virtual ~TraceRecordResult();
+
+  // Trace type of the corresponding TraceRecord.
+  virtual TraceType GetTraceType() const;
+
+  class Handler {
+   public:
+    virtual ~Handler() {}
```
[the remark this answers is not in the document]
autopear: Done

include/rocksdb/trace_record_result.h
autopear: Done

trace_replay/trace_record_result.cc
autopear: Done

include/rocksdb/utilities/replayer.h
autopear: Done

trace_replay/trace_record_result.cc
jay-zhuang: sorry, I didn't see there's an rvalue overload for `value`. Then just change that for `status`:

```
SingleValueTraceExecutionResult::SingleValueTraceExecutionResult(
    Status  status, const std::string& value, uint64_t latency,
    TraceType trace_type)
    : TraceExecutionResult(latency, trace_type),
      status_(std::move(status)),
      value_(value) {}
```

include/rocksdb/trace_record_result.h
jay-zhuang: `NowNanos()` uses `CLOCK_MONOTONIC`, which is not real time:
https://linux.die.net/man/3/clock_gettime
And it may not be accurate between threads (cores): https://stackoverflow.com/questions/46893072/clock-gettimeclock-monotonic-monotonicity-across-cores-threads

Typically we cannot get current absolute time at nanos level, only to the Micros level.
So is it okay to use Micros? or switch back to latency?

trace_replay/trace_record_result.cc
autopear: Done

include/rocksdb/trace_record.h
```diff
@@ -97,9 +132,11 @@ class WriteQueryTraceRecord : public QueryTraceRecord {
 
   TraceType GetTraceType() const override { return kTraceWrite; }
 
+  // rep string for the WriteBatch.
   virtual Slice GetWriteBatchRep() const;
 
-  virtual Status Accept(Handler* handler) override;
+  virtual Status Accept(Handler* handler,
```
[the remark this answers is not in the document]
autopear: Done

include/rocksdb/trace_record.h
```diff
@@ -118,25 +155,24 @@ class GetQueryTraceRecord : public QueryTraceRecord {
...
   TraceType GetTraceType() const override { return kTraceGet; }
 
+  // Column family ID.
   virtual uint32_t GetColumnFamilyID() const;
 
+  // Key to get.
   virtual Slice GetKey() const;
 
-  virtual Status Accept(Handler* handler) override;
+  virtual Status Accept(Handler* handler,
```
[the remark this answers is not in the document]
autopear: Done

trace_replay/trace_record_result.cc
```diff
@@ -0,0 +1,108 @@
...
+uint64_t TraceExecutionResult::GetEndTimestamp() const { return ts_end_; }
+
+// StatusOnlyTraceExecutionResult
+StatusOnlyTraceExecutionResult::StatusOnlyTraceExecutionResult(
+    const Status& status, uint64_t start_timestamp, uint64_t end_timestamp,
+    TraceType trace_type)
+    : TraceExecutionResult(start_timestamp, end_timestamp, trace_type),
+      status_(status) {}
+
+StatusOnlyTraceExecutionResult::~StatusOnlyTraceExecutionResult() {}
```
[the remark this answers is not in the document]
autopear: Done

include/rocksdb/trace_record_result.h
autopear: > `NowNanos()` uses `CLOCK_MONOTONIC`, which is not real time:
> https://linux.die.net/man/3/clock_gettime
> And it may not be accurate between threads (cores): https://stackoverflow.com/questions/46893072/clock-gettimeclock-monotonic-monotonicity-across-cores-threads
> 
> Typically we cannot get current absolute time at nanos level, only to the Micros level.
> So is it okay to use Micros? or switch back to latency?

Changed to `NowMicros()`. Is this OK?

trace_replay/trace_record_handler.cc
```diff
@@ -28,33 +30,72 @@ TraceExecutionHandler::TraceExecutionHandler(
...
+  }
   auto it = cf_map_.find(record.GetColumnFamilyID());
   if (it == cf_map_.end()) {
     return Status::Corruption("Invalid Column Family ID.");
   }
 
-  Iterator* single_iter = db_->NewIterator(read_opts_, it->second);
+  uint64_t start = clock_->NowMicros();
 
+  Iterator* single_iter = db_->NewIterator(read_opts_, it->second);
```
ajkr, line 98: I think one of the benefits of moving timing down to the TraceRecord handler is we can exclude setup/teardown like iterator creation/destruction. Should we move start and end accordingly so it only measures `Seek()` / `SeekForPrev()` since that's what the optype indicates? (Maybe in the future we have separate iterator creation/destruction record types and then this operation would just reference an already existing iterator.)

include/rocksdb/utilities/replayer.h
```diff
@@ -61,13 +65,40 @@ class Replayer {
...
+                         std::unique_ptr<TraceRecordResult>* result) = 0;
+  inline Status Execute(const std::unique_ptr<TraceRecord>& record) {
+    return Execute(record, nullptr);
+  }
+
+  virtual Status Execute(std::unique_ptr<TraceRecord>&& record,
+                         std::unique_ptr<TraceRecordResult>* result) = 0;
+  inline Status Execute(std::unique_ptr<TraceRecord>&& record) {
+    return Execute(std::move(record), nullptr);
+  }
```
ajkr, line 82: I feel like these API overloads are proliferating unnecessarily:

- This PR changes which overload is pure virtual, so it's not compatible (which is fine - we haven't released this code yet so can feel free to make incompatible changes). But if we're not aiming for compatibility, why keep the single argument `Execute()`s?
- I can't think of what scenario we'd need overloads for taking unique_ptr by const lvalue reference, and by rvalue reference:
  - If we take ownership of record we only need the rvalue reference interface (std::unique_ptr is not copyable).
  - If we don't take ownership of record we only need the lvalue reference interface.

My guess is `virtual Status Execute(const std::unique_ptr<TraceRecord>& record, std::unique_ptr<TraceRecordResult>* result) = 0;` is good enough.

ajkr (changes requested):
Looks great overall, mostly comments on the API

include/rocksdb/trace_record_result.h
```diff
@@ -0,0 +1,153 @@
...
+                                  uint64_t start_timestamp,
+                                  uint64_t end_timestamp, TraceType trace_type);
+
+  virtual ~MultiValuesTraceExecutionResult() override;
+
+  // Return status of DB::MultiGet(), etc.
+  virtual std::vector<Status> GetMultiStatus() const;
+
+  // Values for the searched keys.
+  virtual std::vector<std::string> GetValues() const;
```
ajkr, line 144: Sorry had some questions on these interfaces too.

- I find it hard to imagine the typical use case requires the vector returned by `GetValues()` outlives the MultiValuesTraceExecutionResult. Can we return a const reference to the member variable instead?
- I also find it hard to imagine the `std::vector` move constructor consumes meaningful time that makes two `MultiValuesTraceExecutionResult`s worthwhile. `std::string` was a bit controversial because of the short string optimization that inlines a small buffer which had to be copied during a move. But `std::vector` just has size/capacity/pointer to buffer, it should be very very cheap to move. So I think just one constructor that takes vector by value (and callers can invoke it with a moved-from vector if they wish) is sufficient.

include/rocksdb/trace_record_result.h
```diff
@@ -0,0 +1,153 @@
...
+    virtual Status Handle(const SingleValueTraceExecutionResult& /*result*/) {
+      return Status::NotSupported(
+          "SingleValueTraceExecutionResult is not supported.");
+    }
+
+    // Handle MultiValuesTraceExecutionResult
+    virtual Status Handle(const MultiValuesTraceExecutionResult& /*result*/) {
+      return Status::NotSupported(
+          "MultiValuesTraceExecutionResult is not supported.");
+    }
```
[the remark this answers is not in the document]
autopear: Done

utilities/trace/replayer_impl.cc
```diff
@@ -124,19 +125,34 @@ Status ReplayerImpl::Replay(const ReplayOptions& options) {
...
           std::chrono::microseconds(static_cast<uint64_t>(std::llround(
-              1.0 * (trace.ts - header_ts_) / options.fast_forward))));
+              1.0 * (trace.ts - header_ts_) / options.fast_forward)));
+      if (sleep_to > std::chrono::system_clock::now()) {
+        std::this_thread::sleep_until(sleep_to);
+      }
+
+      // Skip unsupported traces, stop for other errors.
+      if (s.IsNotSupported()) {
+        if (result_callback == nullptr) {
```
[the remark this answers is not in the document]
autopear: Done

trace_replay/trace_record_handler.cc
autopear: Done

include/rocksdb/utilities/replayer.h
autopear: Done

include/rocksdb/trace_record_result.h
autopear: Done

include/rocksdb/trace_record.h
```diff
@@ -41,49 +42,83 @@ enum TraceType : char {
...
+    virtual ~Handler() = default;
+
+    // Handle WriteQueryTraceRecord
+    virtual Status Handle(const WriteQueryTraceRecord& /*record*/,
+                          std::unique_ptr<TraceRecordResult>* /*result*/) {
+      return Status::NotSupported("WriteQueryTraceRecord is not supported.");
+    }
+    inline Status Handle(const WriteQueryTraceRecord& record) {
+      return Handle(record, nullptr);
+    }
```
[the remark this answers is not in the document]
autopear: Done

ajkr (approved):
Nice work!

include/rocksdb/trace_record_result.h
```diff
@@ -0,0 +1,139 @@
...
+
+#include <string>
+#include <vector>
+
+#include "rocksdb/rocksdb_namespace.h"
+#include "rocksdb/status.h"
+#include "rocksdb/trace_record.h"
+
+namespace ROCKSDB_NAMESPACE {
+
```
zhichao-cao, line 16: I may prefer a general comment for the trace record result to explain how it works, how to use it, and the example use cases.

PR #8658, opened 2021-08-13
Add support to the ObjectRegistry for ManagedObjects

mrambacher:
ManagedObjects are  shared pointer objects where RocksDB wants to share a single object between multiple configurations.  For example, the Cache may be shared between multiple column families/tables or the Statistics may be shared between multiple databases.

ManagedObjects are stored in the ObjectRegistry by Type (e.g. Cache) and ID.  For a given type/ID name, a single object is stored.

APIs were added to get/set/create these objects.


include/rocksdb/utilities/object_registry.h
```diff
@@ -265,15 +265,9 @@ class ObjectRegistry {
   // If the object was found in the registry, the result is updated and OK is
```
pdillinger, line 265: Update comment

include/rocksdb/customizable.h
```diff
@@ -190,6 +190,20 @@ class Customizable : public Configurable {
...
+  // - addr is the memory address of this object;
+  // - pid is the process ID of this process ID for this process.
+  // Note that if obj1 and obj2 have the same unique IDs, they must be the
+  // same.  However, if an object is deleted and recreated, it may have the
+  // same unique ID as a predecessor
+  //
+  // This method is useful for objects (especially ManagedObjects) that
+  // wish to generate an ID that is unique for this instance and wish to
+  // override the GetId() method.
+  std::string GenerateUniqueId() const;
```
pdillinger, line 205: I think calling this a "unique ID" is confusing in comparison to other APIs (unique_id.h, Env::GenerateUniqueId()). And Generate suggests it's new each time.

How about `GetInstanceId()`?

pdillinger (changes requested):

pdillinger (approved):
👍

include/rocksdb/utilities/object_registry.h
```diff
@@ -249,13 +255,136 @@ class ObjectRegistry {
...
+  // Creates a new ManagedObject in the registry for the id if one does not
+  // currently exist.  If an object with that ID already exists, the current
+  // object is returned.
+  //
+  // The ID is the identifier of the object to be returned/created and returned
+  // in result
+  // If a new object is created (using the object factories), the cfunc
+  // parameter will be invoked to configure the new object.
+  template <typename T>
+  Status GetOrCreateManagedObject(const std::string& id,
```
zhichao-cao, line 317: One general question, the managed object can only be shared within the same process right? For example, a process create N DB instance and one Cache instance, so the cache is shard by N DB. So once the ID is the same, we always get the ptr of the same object instance. LoadSharedObject ensure that, each time we call it, it will return different instance even we give the same ID?

mrambacher: @zhichao-cao There is only sharing of objects within a process.  There is nothing (using shared memory, etc) to allow multiple processes on a single host.  Is there a requirement/request to do so?  I do not know of any objects that are currently shared that way now.   It would be theoretically possible for a MemoryAllocator to be "shared" between processes, where each process had their own Allocator referencing a common/shared pool.  This would be a feature of how the allocator was set up and I do not know how locks/concurrency would be addressed between the processes

zhichao-cao: For now, without the ManagedObject related interface, the way we do object sharing across DB is, the main thread create this object and pass the shared_ptr to different DB instances. For now, with this change, the main thread only need to remember the ID? The object is always alive until the process is ended since it is managed by the objectregestry?

mrambacher: > For now, without the ManagedObject related interface, the way we do object sharing across DB is, the main thread create this object and pass the shared_ptr to different DB instances. For now, with this change, the main thread only need to remember the ID? The object is always alive until the process is ended since it is managed by the objectregestry?

@zhichao-cao If someone does not maintain a reference to the underlying object, then it will be removed from the ObjectRegistry (it is stored as a weak pointer).  For purposes here, I will assume the object being shared is Cache, but it applies to any managed object.  There are a couple of ways this can be handled.
1.  The existing way.  The main thread creates the cache and then passes the ID (or shared ptr) to the other DBs.  Each DB can get the cache from its ID (or shared ptr)
2. The OPTIONS file.  If the Cache is stored in the Options file and all DBs in the same process use the same cache, then all of them will have the same ID.  When the options are hydrated from disk, the ID will be read and all of the DBs in the same process will have stored the same ID in their options file, causing all of them to use the same managed object.

Note that the two methods cannot be combined.  If a cache is create from scratch in the main thread, it will have a different ID than the one in the options file and therefore will use a different cache instance.  If the cache is re-hydrated from the options file, then a new cache will be created and then destroyed/replaced by replacing the one passed in with the one re-created.

PR #8659, opened 2021-08-13
Stable cache keys using DB session ids in SSTs

pdillinger:
Summary: Use DB session ids in SST table properties to make cache keys
stable across DB re-open and copy / move / restore / etc.

These new cache keys are currently only enabled when FileSystem does not
provide GetUniqueId. For now, they are typically larger, so slightly
less efficient.

Relevant to #7405

This change has a minor regression in PersistentCache functionality:
metaindex blocks are no longer cached in PersistentCache. Table properties
blocks already were not but ideally should be. I didn't spent effort to
fix & test these issues because we don't believe PersistentCache is used much
if at all and expect SecondaryCache to replace it. (Though PRs are welcome.)

FIXME: there is more to be fixed for stable cache keys on external SST files

Test Plan: new unit test added, which fails when disabling new
functionality

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -56,6 +56,7 @@
 #include "table/meta_blocks.h"
 #include "table/multiget_context.h"
 #include "table/persistent_cache_helper.h"
+#include "table/persistent_cache_options.h"
 #include "table/sst_file_writer_collectors.h"
 #include "table/two_level_iterator.h"
 #include "test_util/sync_point.h"
@@ -370,7 +371,7 @@ Cache::Handle* BlockBasedTable::GetEntryFromCache(
 // Helper function to setup the cache key's prefix for the Table.
 void BlockBasedTable::SetupCacheKeyPrefix(Rep* rep,
                                           const std::string& db_session_id,
-                                          uint64_t cur_file_num) {
+                                          uint64_t file_num) {
   assert(kMaxCacheKeyPrefixSize >= 10);
   rep->cache_key_prefix_size = 0;
   rep->compressed_cache_key_prefix_size = 0;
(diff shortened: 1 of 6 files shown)


table/block_based/block_based_table_reader.cc
```diff
@@ -370,27 +371,36 @@ Cache::Handle* BlockBasedTable::GetEntryFromCache(
...
+        file_num);
   }
   if (rep->table_options.block_cache_compressed != nullptr) {
     GenerateCachePrefix<Cache, FSRandomAccessFile>(
         rep->table_options.block_cache_compressed.get(), rep->file->file(),
         &rep->compressed_cache_key_prefix[0],
-        &rep->compressed_cache_key_prefix_size, "", cur_file_num);
+        &rep->compressed_cache_key_prefix_size, db_session_id, file_num);
+  }
+  if (rep->table_options.persistent_cache != nullptr) {
```
zhichao-cao, line 390: I guess, no one is using the persistent cache. Maybe we can deperate it in the future when secondary cache is ready

pdillinger: Microsoft might be using it. See #6932 @zhli1142015

table/block_based/block_based_table_reader.cc
```diff
@@ -619,6 +626,33 @@ Status BlockBasedTable::Open(
...
+      if (file_num > 0) {
+        db_session_id = rep->table_properties->db_session_id;
+      }
+    } else {
+      // External (ingested) SST file - should not use current file number
+      // (changed from original), so that same file ingested into different DBs
+      // can share block cache entries. Although they can modify the embedded
+      // global_seqno, that information is not currently cached under these
+      // portable/stable keys.
+      // Note: For now, each external SST file gets its own unique session id.
```
zhichao-cao, line 649: Maybe also mention that cc->NewId() will be used instead file_number as the suffix.

zhichao-cao (approved):
LGTM, thanks for the PR, it will make the db_session_id stable when db is reopened.

table/block_based/block_based_table_reader.cc
pdillinger: Oops! My intention was to use a fixed file number as suffix but 0 means "unknown."

I will fix this and add to unit test.

pdillinger: Actually, fixing it requires a bit more intervention than I want to throw into this PR after approval.

zhichao-cao: I will take a look after you push the updates.

PR #8661, opened 2021-08-13
[NO REVIEW] Add a unit test for another case in which SD marker's seq may be zeroed

riversand963:
Add a unit test for another case in which SD marker's seq may be zeroed. Incorrect
result may be returned by iterator afterwards.

Also remove some assertion to emulate what happens if RocksDB is built
in opt mode.

Test plan:
```
 ./write_prepared_transaction_test --gtest_filter=*/WritePreparedTransactionTest.ReleaseEarliestSnapshotDuringCompaction_WithSD2/*
```


riversand963:
Unit test failure is intended.

PR #8663, opened 2021-08-16
HistogramStat::Add(): fix race condition

rockeet:
In `HistogramStat::Add()`, atomic is not properly used and there are race conditions, this PR fixed this issue.


mrambacher (approved):
LGTM!

ajkr (changes requested):
It's intentional (#2720).

Issue #8647, opened 2021-08-11
A IO error occurs when doing Checkpoint


zaorangyang:
### Expected behavior
Checkpoint should be executed correctly when db_paths specified
### Actual behavior
It throws a IO error -- "IO error: No such file or directory: while link file to checkpont_dir.tmp/000009.sst: main/000009.sst: No such file or directory"
### Steps to reproduce the behavior
Take a look at the following example：
```
#include "rocksdb/db.h"
#include "rocksdb/utilities/checkpoint.h"
#include <iostream>

int main() {
    rocksdb::DB* db;
    rocksdb::Options options;
    options.create_if_missing = true;
    options.enable_blob_files = true;
    options.min_blob_size = 0;
    std::string dbname = "main";
    options.wal_dir = dbname + "_wal";
    options.db_paths.emplace_back(dbname + "_db", std::numeric_limits<uint64_t>::max());
    rocksdb::Status status = rocksdb::DB::Open(options, dbname, &db);
    assert(status.ok());
    std::string key = "key1";
    std::string value = "value1";
    status = db->Put(rocksdb::WriteOptions(), key, value);
    assert(status.ok());
    rocksdb::FlushOptions flush_opts;
    status = db->Flush(flush_opts);
    assert(status.ok());
    status = db->Get(rocksdb::ReadOptions(), key, &value);
    assert(status.ok());
    
    std::string checkpoint_dir = "checkpont_dir";
    rocksdb::Checkpoint* checkpoint;
    std::cout << "Start Checkpoint" << std::endl;
    rocksdb::Checkpoint::Create(db, &checkpoint);
    status = checkpoint->CreateCheckpoint(checkpoint_dir);
    std::cout << status.ToString() << std::endl;
    delete db;
    delete checkpoint;
}
```

zaorangyang:
The root cause is that the path of source SST file is just ‘dbname_/sstname’: 
https://github.com/facebook/rocksdb/blob/e3a96c48230658270c95812aea9ab9878d9cb61f/utilities/checkpoint/checkpoint_impl.cc#L341
This way can not get the right path of SST/blob file when db_paths specified.

zaorangyang:
Hi, @ajkr. I'm trying to fix it, please assign it to me. Thank you :)

autopear:
The same issue may also apply to `db_paths`, `cf_paths`, `wal_dir` and `db_log_dir`, or any potential path options.

A few things to note in checkpoint_impl:
[...]

PR #8664, opened 2021-08-16
Make checkpoint support for non-default path of live files

zaorangyang:
Make checkpoint support for non-default path of live files

Summary:
This PR is for https://github.com/facebook/rocksdb/issues/8647. A new API based on GetLiveFiles to db.h is added. This API can retrieves the all files with path and name in the database. For a live file with non-default path, the correct path can be located using this API when doing a checkpoint.

### DIFF
--- a/db/db_filesnapshot.cc
+++ b/db/db_filesnapshot.cc
@@ -73,7 +73,7 @@ Status DBImpl::GetLiveFiles(std::vector<std::string>& ret,
   }
 
   // Make a set of all of the live table and blob files
-  std::vector<uint64_t> live_table_files;
+  std::vector<FileDescriptor> live_table_files;
   std::vector<uint64_t> live_blob_files;
   for (auto cfd : *versions_->GetColumnFamilySet()) {
     if (cfd->IsDropped()) {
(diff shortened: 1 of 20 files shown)


zaorangyang:
Could you please take a look at this PR.  And let me know that if any additional changes are required. Thank you.  :)

pdillinger:
Perhaps this is more motivation for my proposed unified version of GetLiveFiles, GetSortedWalFiles, GetLiveFilesChecksumInfo, and now this.

Internal ref: T89016156

pdillinger:
This solution doesn't work anyway because the Checkpoint is not readable as a self-contained DB, because the manifest still expects the pieces in alternate dirs to be in alternate dirs, not in the main DB dir.

The test doesn't check for this. If I modify it to delete the DB and open the checkpoint it fails.

```
diff --git a/utilities/checkpoint/checkpoint_test.cc b/utilities/checkpoint/checkpoint_test.cc
index 3ad84e5ef..998c143b5 100644
--- a/utilities/checkpoint/checkpoint_test.cc
+++ b/utilities/checkpoint/checkpoint_test.cc
@@ -923,6 +923,10 @@ TEST_F(CheckpointTest, CheckpointWithNoDefaultDbpath) {
 
   ASSERT_OK(checkpoint->CreateCheckpoint(snapshot_name_));
 
+  Close();
+  DestroyDB(dbname_, options);
+  DestroyDB(dbname_ + "_db", options);
+
   // Make sure the checkpoint can be opened and the value read
   options.create_if_missing = false;
   DB* checkpoint_db = nullptr;
```

If I add `options.db_paths.clear()` before opening the checkpoint DB, some things work but others (including compaction) don't work.

PR #8666, opened 2021-08-16
Add a stat to count secondary cache hits

anand1976:
Add a stat for secondary cache hits. The ```Cache::Lookup``` API had an unused ```stats``` parameter. This PR uses that to pass the pointer to a ```Statistics``` object that ```LRUCache``` uses to record the stat.

Test plan:
Update a unit test in lru_cache_test


zhichao-cao (approved):
LGTM, should we mention it in HISTORY.md?

PR #8668, opened 2021-08-17
Exclude property kLiveSstFilesSizeAtTemperature from stress_test

jay-zhuang:
Summary: Just like other per_level properties.

Test Plan: stress_test


zhichao-cao (approved):
LGTM, thanks for the fix!

PR #8669, opened 2021-08-17
Stable cache keys on ingested SST files

pdillinger:
Summary: Extends #8659 to work for ingested external SST files, even
the same file ingested into different DBs sharing a block cache.

Note: These new cache keys are currently only enabled when FileSystem
does not provide GetUniqueId. For now, they are typically larger,
so slightly less efficient.

Test Plan: Extended unit test

### DIFF
--- a/db/repair.cc
+++ b/db/repair.cc
@@ -93,6 +93,7 @@ class Repairer {
            const ColumnFamilyOptions& default_cf_opts,
            const ColumnFamilyOptions& unknown_cf_opts, bool create_unknown_cfs)
       : dbname_(dbname),
+        db_session_id_(DBImpl::GenerateDbSessionId(db_options.env)),
         env_(db_options.env),
         file_options_(),
         db_options_(SanitizeOptions(dbname_, db_options)),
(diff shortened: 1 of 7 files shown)


db/repair.cc
```diff
@@ -109,21 +110,16 @@ class Repairer {
...
+                                    /*io_tracer=*/nullptr, db_session_id_)),
         wb_(db_options_.db_write_buffer_size),
         wc_(db_options_.delayed_write_rate),
-        // TODO: db_session_id for VersionSet should be initialized after
-        // db_session_id_ is set and use it for initialization.
         vset_(dbname_, &immutable_db_options_, file_options_,
               raw_table_cache_.get(), &wb_, &wc_,
               /*block_cache_tracer=*/nullptr, /*io_tracer=*/nullptr,
-              /*db_session_id*/ ""),
+              db_session_id_),
```
zhichao-cao, line 122: One question, for the ingested SST file, we will use the newly generated db_session_id+ 1 + offset as the cache key, do we have the collision issues? The file number and the db_session_id for all the injected SST files are the same?

pdillinger: External SST files, which are the only ones allowed to be ingested, are assigned unique session ids in SstFileWriter::Open. If Repairer generates SST files, they will have assigned non-colliding file numbers.

zhichao-cao (approved):
LGTM, thanks!

PR #8671, opened 2021-08-17
Self tuning memtable bloom filter

bjlemaire:
Create a new DBOption flag, `options.memtable_self_tuning_bloom`, that updates the size of the memtable bloom filter every time a new mutable memtable is installed in `SwitchMemtable`. The new size is calculated by approximating the number of unique entries in the old memtable bloom filter, and targeting the FPR to be 1% (number of probes/hash functions then stays k=-log_2(FPR)=6).
A unit test is added that verifies that the BF size is updated on the fly, and that nothing explicitly crashes.


db/memtable.cc
```diff
@@ -117,10 +118,12 @@ MemTable::MemTable(const InternalKeyComparator& cmp,
   // use bloom_filter_ for both whole key and prefix bloom filter
   if ((prefix_extractor_ || moptions_.memtable_whole_key_filtering) &&
       moptions_.memtable_prefix_bloom_bits > 0) {
+    uint32_t bf_bits =
+        filter_bits ? *filter_bits : moptions_.memtable_prefix_bloom_bits;
```
mrambacher, line 122: Is it possible for *filter_bits = 0?  Should that be guarded/tested against?

db/memtable.h
```diff
@@ -107,7 +107,8 @@ class MemTable {
                     const ImmutableOptions& ioptions,
                     const MutableCFOptions& mutable_cf_options,
                     WriteBufferManager* write_buffer_manager,
-                    SequenceNumber earliest_seq, uint32_t column_family_id);
+                    SequenceNumber earliest_seq, uint32_t column_family_id,
+                    const uint32_t* filter_bits = nullptr);
```
mrambacher, line 111: Is there an advantage to making this be a pointer defaulting to nullptr vs a value defaulting to 0?

db/memtable.h
```diff
@@ -495,6 +496,10 @@ class MemTable {
   // Returns a heuristic flush decision
   bool ShouldFlushNow();
 
+  uint64_t BFUniqueEntryEstimate() const {
+    return bloom_filter_->UniqueEntryEstimate();
```
mrambacher, line 500: Shouldn't this check that bloom_filter_ is not nullptr, and return 0 if it is?

mrambacher (commented):
A few general comments:
- You need to add the value to options_settable_test to the parameter list
- From "CreateNewMemtable, please pass in a filter_bits value and add a comment why it is okay to be nullptr/0
- This appears to only work on an existing "run", meaning we adapt the bloom filter size as data is added to memtables during a process execution but if the process restarts, any adaptation is lost.  Is that correct?  Is there anything that can be done about that case?
- This new option only applies if some other options are also set and bloom filtering is enabled.  Should the comment be updated to reflect that interdependency?  Should there be validation/sanitization in place so that this option should only be true if the other values are set accordingly?

db/db_impl/db_impl_write.cc
```diff
@@ -1801,7 +1804,24 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
   }
   if (s.ok()) {
     SequenceNumber seq = versions_->LastSequence();
-    new_mem = cfd->ConstructNewMemtable(mutable_cf_options, seq);
+    std::unique_ptr<uint32_t> new_filterbits;
```
anand1976, line 1807: This seems a little unusual. Why not just use an automatic uint32_t with a value of 0 indicating no auto-tuning?

db/memtable.h
anand1976: The Google C++ style guide recommends not using default values for parameters.

db/db_impl/db_impl_write.cc
bjlemaire: That sounds like a much better idea - I'm making the change right now

db/db_impl/db_impl_write.cc
```diff
@@ -1801,7 +1804,24 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
...
+
+    if (mutable_cf_options.memtable_self_tuning_bloom) {
+      const uint64_t pastentries = cfd->mem()->BFUniqueEntryEstimate();
+      new_filterbits.reset(new uint32_t);
+      // Calculate new BF size by assuming the number of unique keys
+      // will be similar to the mutable memtable being switched,
+      // (for 6 probes (hash functions) and a target FPR of 1%)
+      // m = - n_approx * ln(0.01)/[ln(2)**2]
```
pdillinger, line 1815: I don't think we need to be fancy here. The cache locality makes FP rates slightly off anyway. We can just say for 1% FP rate, 10 bits per key and 6 probes is known to work well. You can wiki link to https://github.com/facebook/rocksdb/wiki/RocksDB-Bloom-Filter#the-math

pdillinger (changes requested):

db/db_impl/db_impl_write.cc
```diff
@@ -6,7 +6,10 @@
 // Copyright (c) 2011 The LevelDB Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
+// Used to import M_LN2 from <cmath>
+#define _USE_MATH_DEFINES
```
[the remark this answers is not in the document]
bjlemaire: Yep following https://github.com/facebook/rocksdb/pull/8671#discussion_r691696982 , I can simply remove the define_include<cmath>.

db/db_impl/db_impl_write.cc
```diff
@@ -1801,7 +1804,24 @@ Status DBImpl::SwitchMemtable(ColumnFamilyData* cfd, WriteContext* context) {
...
+      // Calculate new BF size by assuming the number of unique keys
+      // will be similar to the mutable memtable being switched,
+      // (for 6 probes (hash functions) and a target FPR of 1%)
+      // m = - n_approx * ln(0.01)/[ln(2)**2]
+      *new_filterbits = static_cast<uint32_t>(
+          (pastentries * (-1.0) * (-2.0 * M_LN10 /* ln(0.01) */)) /
+          (M_LN2 * M_LN2));
+      TEST_SYNC_POINT_CALLBACK("DBImpl::SelfTuningBloom:NewBFSize",
+                               new_filterbits.get());
+    }
```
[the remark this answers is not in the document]
bjlemaire: I rememebr seeing an upper limit of 25% memtable bloom ratio here: https://github.com/facebook/rocksdb/blob/c625b8d017d43d0bceffbd7c0cf039708578fd1f/db/column_family.cc#L260 
Shoudl we also maybe update that value? Unless the user might want to reduce the FP rate down from 1%, I think based on https://github.com/facebook/rocksdb/pull/8671#discussion_r691696982 we may want to update this threshold.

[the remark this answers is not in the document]
bjlemaire: Actually the rose case scenario you be a key of size 2 bytes and a value fo size 1 byte, because then the encoded form  stored in the memtable is 1 + (2+8) + 1 + 1 = 13 bytes, details here: 
https://github.com/facebook/rocksdb/blob/c625b8d017d43d0bceffbd7c0cf039708578fd1f/db/db_flush_test.cc#L354
So maybe the upper hard limit should be 10% memtable Bloom Ratio. I dont see why a ratio > 10% would be necessary beside lowering the FPR

db/db_bloom_filter_test.cc
```diff
@@ -1066,6 +1066,82 @@ TEST_F(DBBloomFilterTest, MemtableWholeKeyBloomFilter) {
   ASSERT_EQ(1, get_perf_context()->bloom_memtable_hit_count);
 }
 
+TEST_F(DBBloomFilterTest, MemtableSelfTuningBloomFilter) {
+  const int kMemtableSize = 64 << 20;             // 1MB
```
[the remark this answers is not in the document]
bjlemaire: Good catch - I just corrected it and put it back to 1MB.

db/db_impl/db_impl_write.cc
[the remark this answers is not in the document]
pdillinger: I was including skiplist overhead of 1.33 * 8 bytes per entry. I believe only the vector memtable would have significantly less space overhead, and that should be rarely used (though bloom filter would pay off a lot for misses!).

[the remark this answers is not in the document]
pdillinger: But 10% seems reasonable for self-tuning firewall limit if you want to use that.

db/memtable.cc
bjlemaire: I changed the logic, now we should be protected against this situation of 0 filter bits.

db/memtable.h
bjlemaire: Following @anand1976 's recommendation,  I changed the way the filter bits are passed to the memtable, I am now updating the MutableCFOptions of the cfd directly.

pdillinger (commented):
I would like to see unit test for DynamicBloom::UniqueEntryEstimate but also a couple of db_bench runs with some debugging output showing how it is configuring itself. And preferably that there is no fixed cost to write throughput by enabling self-tuning, by e.g. comparing a fixed memtable bloom configuration with a self-tuning that consistently ends up using essentially that same configuration.

I also think that especially in combination with mempurge, we might need a longer memory than the 1 previous flush (attempt?). If you want to call this out as future work, that would be OK.

And update HISTORY.md

PR #8672, opened 2021-08-17
Add condition on NotifyOnFlushComplete that FlushJob was not mempurge. Add event listeners to mempurge tests.

bjlemaire:
Previously, when a `FlushJob` was redirected to a MemPurge, the function `DBImpl::NotifyOnFlushComplete` was called, which created a series of issues because the JobInfo was not correctly collected from the memtables.
This diff aims at correcting these two issues (`FlushJobInfo` collection in `FlushJob::MemPurge` , no call to `DBImpl::NotifyOnFlushComplete` after successful mempurge).
Event listeners were added to the unit tests to handle these situations.
Surprisingly none of the crashtests caught this issue, I will try to add event listeners to crash tests in the future.

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -480,7 +480,7 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
     assert(exec_status.size() > 0);
     assert(!file_meta.empty());
     exec_status[0].second = jobs[0]->Run(
-        &logs_with_prep_tracker_, &file_meta[0], &(switched_to_mempurge[0]));
+        &logs_with_prep_tracker_, &file_meta[0], switched_to_mempurge.data());
     exec_status[0].first = true;
     io_status[0] = jobs[0]->io_status();
 
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -415,6 +415,8 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
   }
 
   std::vector<FileMetaData> file_meta(num_cfs);
+  // Use of deque<bool> because vector<bool>
+  // is specific and doesn't allow &v[i].
   std::deque<bool> switched_to_mempurge(num_cfs, false);
   Status s;
   IOStatus log_io_s = IOStatus::OK();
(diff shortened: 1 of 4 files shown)


bjlemaire:
I use `uint8_t` instead of `bool` solely because `vector<bool>` is a special variable that created a lot of issues when I tried to make calls with `&switched_to_mempurge[i] `(or `switched_to_mempurge.data()+i`) in DBimpl::AtomicMemFlush[...].
Definitely interested in hearing what people say about that, if using `uint8_t` is too confusing or not.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -411,6 +414,7 @@ Status DBImpl::AtomicFlushMemTablesToOutputFiles(
   }
 
   std::vector<FileMetaData> file_meta(num_cfs);
+  std::vector<uint8_t> switched_to_mempurge(num_cfs, false);
```
akankshamahajan15, line 417: Shouldn't it be 0 instead of false? I don't think it will make any difference apart from readability.

db/flush_job.cc
```diff
@@ -228,6 +228,9 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker,
     prev_cpu_read_nanos = IOSTATS(cpu_read_nanos);
   }
   Status mempurge_s = Status::NotFound("No MemPurge.");
+  if (switched_to_mempurge) {
+    *switched_to_mempurge = 0;
```
akankshamahajan15, line 232: Do we need to set it or it will be 0 by default?

db/flush_job.cc
```diff
@@ -244,6 +247,16 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker,
         ROCKS_LOG_WARN(db_options_.info_log, "Mempurge process failed: %s\n",
                        mempurge_s.ToString().c_str());
       }
+    } else {
+      if (switched_to_mempurge) {
+        *switched_to_mempurge = 1;
+      } else {
+        // The mempurge process was successful, but no switch_to_mempurge
+        // pointer provided so no way to propagate the state of flush job.
+        ROCKS_LOG_WARN(db_options_.info_log,
```
akankshamahajan15, line 256: Just curious, what are the reasons that ```switched_to_mempurge``` is not provided

db/db_impl/db_impl_compaction_flush.cc
bjlemaire: My bad, thanks for catching that, I was trying to use `bool`s the whole time so that's just a remnant of what I used before switching to `uint8_t`.

db/flush_job.cc
bjlemaire: It is `0` by default, I added this out of paranoia. It's probably cleaner to remove this you are right, thanks for your input.

db/flush_job.cc
bjlemaire: This is a good question. The first two arguments of `FlushJob::Run`, `prep_tracker` and `file_meta`, are already optional arguments, so adding a third argument required a default value as well.
In practice, the only moment sin the current code base when `Run()` (aka no arguments provided, including no `switched_to_mempurge`) is called is in `flush_job_test.cc`, so I am (arguably)  hesitant to make it a required argument for this reason.

akankshamahajan15: Make sense. Thanks for clarifying it.

akankshamahajan15 (approved):
LGTM!

PR #8675, opened 2021-08-18
Expose blob file information through the EventListener interface

akankshamahajan15:
Summary: 1. Extend FlushJobInfo and CompactionJobInfo with information about the blob files generated by flush/compaction jobs. This PR add two structures BlobFileInfo and BlobFileGarbageInfo that contains the required information of blob files.
 2. Notify the creation and deletion of blob files through OnBlobFileCreationStarted, OnBlobFileCreated, and OnBlobFileDeleted.
 3. Test OnFile*Finish operations notifications with Blob Files.
 4. Log the blob file creation/deletion events through EventLogger in Log file.
 
Test Plan: Add new unit tests in listener_test

Reviewers:

Subscribers:

Tasks:

Tags:


db/blob/blob_file_builder_test.cc
```diff
@@ -145,7 +145,7 @@ TEST_F(BlobFileBuilderTest, BuildAndCheckOneFile) {
       TestFileNumberGenerator(), fs_, &immutable_options, &mutable_cf_options,
       &file_options_, job_id, column_family_id, column_family_name, io_priority,
       write_hint, nullptr /*IOTracer*/, nullptr /*BlobFileCompletionCallback*/,
-      &blob_file_paths, &blob_file_additions);
+      options.cf_paths.front().path, &blob_file_paths, &blob_file_additions);
```
mrambacher, line 148: There is another PR in the works to introduce a "blob_db_path" option.  To make things simpler in the future, can we introduce a method now to ImmutableCFOptions "GetBlobDBPath() which returns cf_paths.front()?  

Tracking down all of the places that use cf_paths.front to mean blob_db_path is not simple.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -1342,6 +1342,12 @@ Status DBImpl::CompactFilesImpl(
                                    newf.second.fd.GetNumber(),
                                    newf.second.fd.GetPathId()));
     }
+
+    for (const auto& blob_file : c->edit()->GetBlobFileAdditions()) {
+      (*output_file_names)
+          .push_back(BlobFileName(c->immutable_options()->cf_paths.front().path,
```
mrambacher, line 1348: Why not output_file_names->push_back?  More common syntax...

include/rocksdb/listener.h
```diff
@@ -206,6 +230,22 @@ struct FileOperationInfo {
...
+  CompressionType compression_type;
+};
+
+struct BlobFileGarbageInfo {
+  std::string blob_file_path;
+  uint64_t blob_file_number;
+  uint64_t garbage_blob_count;
+  uint64_t garbage_blob_bytes;
+  CompressionType compression_type;
+};
```
mrambacher, line 247: Why two structs with identical content?

db/db_impl/db_impl_files.cc
```diff
@@ -360,6 +360,13 @@ void DBImpl::DeleteObsoleteFileImpl(int job_id, const std::string& fname,
         &event_logger_, job_id, number, fname, file_deletion_status, GetName(),
         immutable_db_options_.listeners);
   }
+#ifndef ROCKSDB_LITE
+  if (type == kBlobFile) {
+    EventHelpers::NotifyBlobFileDeletion(immutable_db_options_.listeners,
+                                         job_id, number, fname,
+                                         file_deletion_status, GetName());
+  }
```
mrambacher, line 368: For SST files, the Notify also logs something to the Logger (independent or LITE mode). 

 Is that not a requirement/useful for Blob files?

include/rocksdb/listener.h
ltamasi: They're not identical: one has `total_blob_{count,bytes}` and the other has `garbage_blob_{count,bytes}`.

db/blob/blob_file_builder_test.cc
ltamasi: Seems to me this is actually the new `dbname` argument? If so, we could probably pass a hardcoded value in these tests.

db/db_impl/db_impl_compaction_flush.cc
akankshamahajan15: Yeah I will update it. Thanks!

db/db_impl/db_impl_files.cc
akankshamahajan15: I wasn't sure about it so I didn't add it at first. But @ltamasi  let me know if we want to log in  the Logger or not. I will make the changes accordingly.

include/rocksdb/listener.h
mrambacher: The contents are exactly the same but the variable name is different.  And you are putting the values in two separate lists that are already differentiated.  This means that to traverse the two lists I need to write twice as much code as I cannot process them using a common subroutine.

ltamasi: To me, this is a feature not a bug. The two structures (and the blob count/bytes variables) have different semantics; using dedicated structures improves type safety and is more future-proof.

ltamasi (commented):
Thanks for the PR @akankshamahajan15 ! Please some (mostly minor) comments below

include/rocksdb/listener.h
```diff
@@ -206,6 +230,22 @@ struct FileOperationInfo {
   }
 };
 
+struct BlobFileInfo {
+  std::string blob_file_path;
+  uint64_t blob_file_number;
+  uint64_t total_blob_count;
+  uint64_t total_blob_bytes;
+  CompressionType compression_type;
```
ltamasi, line 238: We could simply store `blob_compression_type` once as part of `FlushJobInfo` / `CompactionJobInfo`; we don't have to expose it on a per-file basis because it's guaranteed to be the same for all bllob files.

include/rocksdb/listener.h
```diff
@@ -59,6 +62,23 @@ struct TableFileCreationInfo : public TableFileCreationBriefInfo {
   std::string file_checksum_func_name;
 };
 
+struct BlobFileCreationBriefInfo : public FileCreationBriefInfo {};
+
+struct BlobFileCreationInfo : public BlobFileCreationBriefInfo {
+  BlobFileCreationInfo() = default;
```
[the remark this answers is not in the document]
akankshamahajan15: @ltamasi Should I do the same for Table also or leave that as it is since its in public API.

ltamasi (commented):
Thanks for all the updates @akankshamahajan15 ! The patch looks great overall, just some minor comments

include/rocksdb/listener.h
```diff
@@ -59,6 +70,44 @@ struct TableFileCreationInfo : public TableFileCreationBriefInfo {
...
+                       uint64_t _total_blob_count, uint64_t _total_blob_bytes,
+                       Status _status, std::string _file_checksum,
+                       std::string _file_checksum_func_name)
+      : BlobFileCreationBriefInfo(_db_name, _cf_name, _file_path, _job_id,
+                                  _reason),
+        total_blob_count(_total_blob_count),
+        total_blob_bytes(_total_blob_bytes),
+        status(_status),
+        file_checksum(_file_checksum),
+        file_checksum_func_name(_file_checksum_func_name) {}
```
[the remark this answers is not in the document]
akankshamahajan15: @ltamasi which constructor arguments are you referring to use `std::move`.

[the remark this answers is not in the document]
ltamasi: I meant doing

```
...
file_checksum(std::move(_file_checksum)),
file_checksum_func_name(std::move(_file_checksum_func_name)) {}
```

HISTORY.md
```diff
@@ -21,10 +21,13 @@
...
 * Allow a single write batch to include keys from multiple column families whose timestamps' formats can differ. For example, some column families may disable timestamp, while others enable timestamp.
+* Added new callback APIs `OnBlobFileCreationStarted`,`OnBlobFileCreated`and `OnBlobFileDeleted` in `EventListener` class of listener.h. It notifies listeners during creation/deletion of individual blob files in Integrated BlobDB. It also log blob file creation finished event and deletion event in LOG file.
 
 ### Public API change
 * Remove obsolete implementation details FullKey and ParseFullKey from public API
 * Add a public API RateLimiter::GetTotalPendingRequests() for the total number of requests that are pending for bytes in the rate limiter.
+* Extended `FlushJobInfo` and `CompactionJobInfo` in listener.h to provide information about the blob files generated by a flush/compaction and deleted during compaction in Integrated BlobDB. Added struct members `blob_file_addition_infos` and `blob_file_garbage_infos` that contain this information.
```
ltamasi, line 29: I feel "garbage collected" instead of "deleted" would be more accurate.

db/blob/blob_file_builder.cc
```diff
@@ -35,14 +36,15 @@ BlobFileBuilder::BlobFileBuilder(
     const std::string& column_family_name, Env::IOPriority io_priority,
     Env::WriteLifeTimeHint write_hint,
     const std::shared_ptr<IOTracer>& io_tracer,
-    BlobFileCompletionCallback* blob_callback,
+    BlobFileCompletionCallback* blob_callback, const std::string& dbname,
```
ltamasi, line 39: Sorry, what I meant earlier is that DB name could also be a member of `BlobFileCompletionCallback` (in addition to the listeners). We don't really need it in `BlobFileBuilder`

ltamasi (approved):
Thanks again @akankshamahajan15, this looks awesome!

PR #8676, opened 2021-08-19
Update the block_read_count/block_read_byte counters in MultiGet

anand1976:
MultiGet in block based table reader doesn't use BlockFetcher. As a result, the block_read_count and block_read_byte PerfContext counters were not being updated. This fixes that by updating them in MultiRead.

### DIFF
--- a/table/block_based/block_based_table_reader.cc
+++ b/table/block_based/block_based_table_reader.cc
@@ -1715,6 +1715,9 @@ void BlockBasedTable::RetrieveMultipleBlocks(
       req_offset_for_block.emplace_back(0);
     }
     req_idx_for_block.emplace_back(read_reqs.size());
+
+    PERF_COUNTER_ADD(block_read_count, 1);
+    PERF_COUNTER_ADD(block_read_byte, block_size(handle));
   }
   // Handle the last block and process the pending last request
   if (prev_len != 0) {
--- a/table/block_based/block_based_table_reader_test.cc
+++ b/table/block_based/block_based_table_reader_test.cc
@@ -228,7 +228,15 @@ TEST_P(BlockBasedTableReaderTest, MultiGet) {
 
   // Execute MultiGet.
   MultiGetContext::Range range = ctx.GetMultiGetRange();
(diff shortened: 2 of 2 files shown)


table/block_based/block_based_table_reader.cc
```diff
@@ -1715,6 +1715,9 @@ void BlockBasedTable::RetrieveMultipleBlocks(
       req_offset_for_block.emplace_back(0);
     }
     req_idx_for_block.emplace_back(read_reqs.size());
+
+    PERF_COUNTER_ADD(block_read_count, 1);
+    PERF_COUNTER_ADD(block_read_byte, block_size(handle));
```
mrambacher, line 1720: Putting the code here (as opposed to later) is the number of reads we attempted.  Is that what is expected or should it be put later and be the number of reads that were successful?

table/block_based/block_based_table_reader.cc
```diff
@@ -1715,6 +1715,9 @@ void BlockBasedTable::RetrieveMultipleBlocks(
       req_offset_for_block.emplace_back(0);
     }
     req_idx_for_block.emplace_back(read_reqs.size());
+
+    PERF_COUNTER_ADD(block_read_count, 1);
```
mrambacher, line 1719: Is there any advantage to using temporary/stack variables for this counters and then updating the PERF_COUNTER variables only once (rather than on every pass through the loop)?

table/block_based/block_based_table_reader.cc
anand1976: Its consistent with how the counters are updated in block_fetcher.cc. Its not explicitly documented in perf_context.h, so just retain existing behavior.

table/block_based/block_based_table_reader.cc
anand1976: I don't think so. The PerfContext is thread local, so directly updating it shouldn't add much overhead.

zhichao-cao (approved):
LGTM, thanks!

PR #8677, opened 2021-08-19
Add iterator's lower and upper bounds to `TraceRecord`

autopear:
Trace file V2 added lower/upper bounds to `Iterator::Seek()` and `Iterator::SeekForPrev()`. They were not used anywhere during the execution of a `TraceRecord`. Now they are added to be used by `ReadOptions` during `Iterator::Seek()` and `Iterator::SeekForPrev()` if they are set.

Added test cases in `DBTest2.TraceAndManualReplay`.


zhichao-cao (approved):
LGTM, thanks for adding the lower and upper bounds to trace&replay new interface.

PR #8678, opened 2021-08-19
Refactor TraceAnalyzer to use `TraceRecord::Handler` to avoid casting.

autopear:
`TraceAnalyzer` privately inherits `TraceRecord::Handler` and `WriteBatch::Handler`.

`trace_analyzer_test` can pass with this change.


tools/trace_analyzer_tool.cc
```diff
@@ -469,96 +472,29 @@ Status TraceAnalyzer::StartProcessing() {
       break;
     }
 
+    // To do: why increment total_requests_ for trace end?
```
zhichao-cao, line 475: Sometimes, the trace might be incomplete. We may not be able to the type::end.

tools/trace_analyzer_tool.h
```diff
@@ -182,24 +183,64 @@ class TraceAnalyzer {
...
+                std::unique_ptr<TraceRecordResult>* result) override;
+  Status Handle(const GetQueryTraceRecord& record,
+                std::unique_ptr<TraceRecordResult>* result) override;
+  Status Handle(const IteratorSeekQueryTraceRecord& record,
+                std::unique_ptr<TraceRecordResult>* result) override;
+  Status Handle(const MultiGetQueryTraceRecord& record,
+                std::unique_ptr<TraceRecordResult>* result) override;
+
+  using WriteBatch::Handler::PutCF;
+  Status PutCF(uint32_t column_family_id, const Slice& key,
```
zhichao-cao, line 200: What does "CF" means here? CallFunction? or ColumnFamily?

autopear: It's ColumnFamily, from
https://github.com/facebook/rocksdb/blob/ff8953380f356768d5ea7bf26b2734d86484d228/include/rocksdb/write_batch.h#L182-L193

tools/trace_analyzer_tool.cc
autopear: Yes, my point is, you should not increment total_requests if you read a `TraceEnd`, so it should be 
```
    end_time_ = trace.ts;
    if (trace.type == kTraceEnd) {
      break;
    }
    total_requests_++;
```
Then `end_time_` is the timetamp of the last trace (or trace end), and `total_requests_` will be the actual number of requests.

tools/trace_analyzer_tool.h
```diff
@@ -213,6 +254,7 @@ class TraceAnalyzer {
   uint64_t total_access_keys_;
   uint64_t total_gets_;
   uint64_t total_writes_;
+  // To do: why no counters for seeks and multigets?
```
zhichao-cao, line 257: Good catch! I forgot to add them........

trace_replay/trace_replay.cc
```diff
@@ -126,222 +126,218 @@ bool TracerHelper::SetPayloadMap(uint64_t& payload_map,
...
+    case kTraceIteratorSeekForPrev: {
+      uint32_t cf_id = 0;
+      Slice iter_key;
+      Slice lower_bound;
+      Slice upper_bound;
+
+      if (trace_file_version < 2) {
+        DecodeCFAndKey(trace->payload, &cf_id, &iter_key);
+      } else {
+        Slice buf(trace->payload);
```
zhichao-cao, line 225: It is not that elegant to put the payload decoding in the trace analyzer.......

tools/trace_analyzer_tool.cc
zhichao-cao: > Yes, my point is, you should not increment total_requests if you read a `TraceEnd`, so it should be
> 
> ```
>     end_time_ = trace.ts;
>     if (trace.type == kTraceEnd) {
>       break;
>     }
>     total_requests_++;
> ```
> 
> Then `end_time_` is the timetamp of the last trace (or trace end), and `total_requests_` will be the actual number of requests.

Sure, just follow what you proposed above.

trace_replay/trace_replay.cc
autopear: This is in TracerHelper, not in TraceAnalyzer. I merged DecodeWriteTraceRecord / DecodeGetTraceRecord, ... into a single `DecodeTraceRecord`, then you don't need to call each individual decoding functions in `TraceAnalyzer` and `ReplayerImpl`.

tools/trace_analyzer_tool.cc
autopear: Done

tools/trace_analyzer_tool.h
autopear: Added

trace_replay/trace_replay.cc
zhichao-cao: Sorry, I did not notice that....

zhichao-cao (approved):
LGTM, thank you so much for the refactor work, excellent job!

PR #8679, opened 2021-08-20
Add Bloom/Ribbon hybrid API support

pdillinger:
Summary: This is essentially resurrection and fixing of level support
in #8198 and reverted in #8212. TODO: finish summary

Test Plan: new + updated tests
TODO: run crash test for a while


include/rocksdb/filter_policy.h
```diff
@@ -266,12 +280,12 @@ extern const FilterPolicy* NewBloomFilterPolicy(
 // Also consider using optimize_filters_for_memory to save filter
 // memory.
 extern const FilterPolicy* NewRibbonFilterPolicy(
-    double bloom_equivalent_bits_per_key);
+    double bloom_equivalent_bits_per_key, int bloom_before_level = 0);
```
jay-zhuang, line 283: I'm wondering if bloom_before_level could be 1 by default. especially for leveled compaction.
And is it okay to change the default behavior?

db/c.cc
```diff
@@ -3840,7 +3840,8 @@ void rocksdb_filterpolicy_destroy(rocksdb_filterpolicy_t* filter) {
   delete filter;
 }
 
-rocksdb_filterpolicy_t* rocksdb_filterpolicy_create_bloom_format(int bits_per_key, bool original_format) {
+rocksdb_filterpolicy_t* rocksdb_filterpolicy_create_bloom_format(
+    double bits_per_key, bool original_format) {
```
jay-zhuang, line 3844: I think it worth to mention the C API update in HISTORY.

jay-zhuang (approved):
LGTM

include/rocksdb/filter_policy.h
pdillinger: I don't think it's easy to change default only for leveled compaction, and I don't want to mess up single-level universal with the default behavior of NewRibbonFilterPolicy never creating ribbon filters.

PR #8680, opened 2021-08-20
Add extra information to RemoteCompaction APIs

jay-zhuang:
Summary: Currently, we only provide job_id in RemoteCompaction APIs, the
main problem of `job_id` is it cannot uniquely identify a compaction job
between DB instances or between sessions.
Providing DB and session id to the user, which will make building cross
DB compaction service easier.

Test Plan: unittest

### DIFF
--- a/db/compaction/compaction_service_test.cc
+++ b/db/compaction/compaction_service_test.cc
@@ -128,7 +128,7 @@ class MyTestCompactionService : public CompactionService,
 
   const char* Name() const override { return kClassName(); }
 
-  CompactionServiceJobStatus Start(
+  CompactionServiceJobStatus StartV2(
       const CompactionServiceJobInfo& info,
       const std::string& compaction_service_input) override {
     InstrumentedMutexLock l(&mutex_);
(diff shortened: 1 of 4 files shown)


db/compaction/compaction_service_test.cc
```diff
@@ -119,59 +252,70 @@ class CompactionServiceTest : public DBTestBase {
...
-TEST_F(CompactionServiceTest, BasicCompactions) {
-  Options options = CurrentOptions();
-  options.env = env_;
-  options.statistics = CreateDBStatistics();
-  std::shared_ptr<Statistics> compactor_statistics = CreateDBStatistics();
-  options.compaction_service = std::make_shared<MyTestCompactionService>(
-      dbname_, options, compactor_statistics);
+typedef testing::Types<MyTestCompactionService, MyTestCompactionServiceLegacy>
+    MyTestCompactionServiceTypes;
+TYPED_TEST_CASE(CompactionServiceTest, MyTestCompactionServiceTypes);
```
ajkr, line 264: I feel like `TEST_P` is more understandable, even considering it'd need some redundancy in specifying the classes to test (maybe an enum for specifying the classes in `INSTANTIATE_TEST_SUITE_P` and then select the class to test accordingly in `CompactionServiceTest`). But this is fine too since it minimizes change to run on a new `CompactionService` subclass...

ajkr (approved):
LGTM!

db/compaction/compaction_service_test.cc
jay-zhuang: Make sense, another benefit is we could remove all `CompactionServiceTest<TypeParam>` with TEST_P.

PR #8681, opened 2021-08-20
Fix blob callback in compaction and atomic flush

akankshamahajan15:
Summary: Pass BlobFileCompletionCallback  in case of atomic flush and
compaction job which is currently nullptr(default parameter).
BlobFileCompletionCallback is used in case of IntegratedBlobDB to report new blob files to
SstFileManager.

Test Plan: CircleCI jobs

Reviewers:

Subscribers:

Tasks:

Tags:


ltamasi (approved):
Thanks for the fix @akankshamahajan15 !

ltamasi:
Would be great to add a unit test for this (I think it could be done as a separate follow-up).

ltamasi:
@akankshamahajan15 Could you also mention the fix in the changelog?

ltamasi (approved):

Issue #8670, opened 2021-08-17
Avoid exception if fail to insert block to block-cache


Myasuka:
### Expected behavior
If fail to insert block to block-cache, just abot the insert but return the result as normal.

### Actual behavior
If we use default `ReadOptions` with `strict_capacity_limit` block cache, RocksDB would throw exception if `Insert failed due to LRU cache being full`.

### Steps to reproduce the behavior
This is because current `RetrieveBlock` would [check the status of `MaybeReadBlockAndLoadToCache`](https://github.com/facebook/rocksdb/blob/add68bd28a512da751e2bdc612685fdeb7e6dde4/table/block_based/block_based_table_reader.cc#L1909-L1916):

~~~c
    s = MaybeReadBlockAndLoadToCache(
        prefetch_buffer, ro, handle, uncompression_dict, wait_for_cache,
        block_entry, block_type, get_context, lookup_context,
        /*contents=*/nullptr);

    if (!s.ok()) {
      return s;
    }
~~~

And if fail to insert, the block cache would report [incomplete status](https://github.com/facebook/rocksdb/blob/add68bd28a512da751e2bdc612685fdeb7e6dde4/cache/lru_cache.cc#L326-L339):

~~~c
    if ((usage_ + total_charge) > capacity_ &&
        (strict_capacity_limit_ || handle == nullptr)) {
      e->SetInCache(false);
      if (handle == nullptr) {
        // Don't insert the entry but still return ok, as if the entry inserted
        // into cache and get evicted immediately.
        last_reference_list.push_back(e);
      } else {
        if (free_handle_on_fail) {
          delete[] reinterpret_cast<char*>(e);
          *handle = nullptr;
        }
        s = Status::Incomplete("Insert failed due to LRU cache being full.");
      }
~~~

As more and more applications move to cloud, and fine-granularity memory control is required in many use cases. I think enabling the strict capacity limit of block cache in production environment should be something valueable.
How about this solution:
1. Introduce a new status as `FailInsertCache`.
1. Introduce a new filed named `fill_cache_If_possible` in ReadOptions, which means if cannot fill block to cache, we would not treat `FailInsertCache` status as a problem, and continue the read process as normal.

What do you think of this problem and the propsed solution?

Myasuka:
cc @zhichao-cao , @ltamasi @akankshamahajan15 who might be interested in this topic.

zhichao-cao:
[...]

PR #8683, opened 2021-08-20
Return ok status if fail to insert to block cache as full

Myasuka:
A simple solution which try to fix https://github.com/facebook/rocksdb/issues/8670.


Myasuka:
This PR hopes to bring the discussion of https://github.com/facebook/rocksdb/issues/8670 to see whether this idea works.

zhichao-cao (commented):
We may not prefer to add a very specific API for cache only usage. What do you think @anand1976

PR #8686, opened 2021-08-20
Embed original file number in SST table properties

pdillinger:
Summary: I very recently realized that with #8669 we cannot later add
file numbers to external SST files (so that more can share db session
ids for better uniqueness properties), because of forward compatibility.
We would have a version of RocksDB that assumes session IDs are unique
on external SST files and therefore can't really break that invariant in
future files.

This change adds a table property for "orig_file_number" which is
populated by normal SST files and also external SST files generated by
SstFileWriter. SstFileWriter now keeps a db_session_id for life of the
object and increments its own file numbers for embedding in table
properties. (They are arguably "fake" file numbers because these numbers
and not embedded in the file name.)

While updating block_based_table_builder, I removed several unnecessary
fields from Rep, because following the pattern would have created
another unnecessary field.

This change also updates block_based_table_reader to use this new
property when available, which means that for newer SST files, we can
determine the stable/original <db_session_id,file_number> unique
identifier using just the file contents, not the file name. (It's a bit
complicated; detailed comments in block_based_table_reader.)

Also added DB host id to properties listing by sst_dump, which could be
useful in debugging.

Test Plan: majorly overhauled StableCacheKeys test for this change


table/sst_file_writer.cc
```diff
@@ -170,7 +174,14 @@ SstFileWriter::SstFileWriter(const EnvOptions& env_options,
                              bool invalidate_page_cache,
                              Env::IOPriority io_priority, bool skip_filters)
     : rep_(new Rep(env_options, options, io_priority, user_comparator,
-                   column_family, invalidate_page_cache, skip_filters)) {
+                   column_family, invalidate_page_cache, skip_filters,
```
zhichao-cao, line 177: So it is possible that the SstFileWriter instance is not changed during the period when db_seesion_id is created for several times due to DB reopen or other actions?

table/block_based/block_based_table_reader.cc
```diff
@@ -637,20 +637,34 @@ Status BlockBasedTable::Open(
     auto version_iter = uprops.find(ExternalSstFilePropertyNames::kVersion);
     if (version_iter == uprops.end()) {
       // Normal (non-external) SST file - can only use embedded db_session_id
-      // with current file number (which should be original file number)
-      if (file_num > 0) {
+      // if original or current file number are known (which should be same)
+      if (rep->table_properties->orig_file_number > 0) {
```
zhichao-cao, line 641: Is there any case that the original file number of an SST file is different from the current_file_number?

table/sst_file_writer.cc
```diff
@@ -241,22 +252,19 @@ Status SstFileWriter::Open(const std::string& file_path) {
...
       TableFileCreationReason::kMisc, 0 /* creation_time */,
       0 /* oldest_key_time */, 0 /* file_creation_time */,
-      "SST Writer" /* db_id */, db_session_id, 0 /* target_file_size */, 0);
+      "SST Writer" /* db_id */, r->db_session_id, 0 /* target_file_size */,
+      r->next_file_number);
+  // External SST files used to each get a unique session id. Now for
+  // slightly better uniqueness probability in constructing cache keys, we
+  // assign fake file numbers to each file (into table properties) and keep
+  // the same session id for the life of the SstFileWriter.
+  r->next_file_number++;
```
zhichao-cao, line 267: So it is like an exchange. In the previous code, db_seesion_id is unique for each file and the file number is always 1. For now, the SST files generated by the same SstFileWriter instance will have the same db_seesion_id but different file number.

table/block_based/block_based_table_reader.cc
pdillinger: For SST files not generated by SstFileWriter, I'm not aware of any RocksDB infrastructure that changes the file number or file name (after last '/') after it is initially written. I'm pushing a change with an assertion here to check for this.

But we have to deal with cases when orig_file_number is not known (old SST file) or current file number is not known (depending on how the file reader is initialized).

table/sst_file_writer.cc
pdillinger: Exactly. :)

table/block_based/block_based_table_reader.cc
pdillinger: It looks like import_column_family_test shows that my assumption was wrong, and we cannot rely on current file number for stable cache key in any case. Regardless of normal vs. external SST file, we either have to use <orig_db_session_id,orig_file_number> entirely from table properties if available, or fall back on <cur_db_session_id,cur_file_number>. At least that simplifies the logic.

zhichao-cao (approved):
Thanks for the PR, it is an essential step for further cache optimizations.

Issue #8673, opened 2021-08-18
HashIndexReader does not work with Prefix Extractor, due to properties retrieval error


mrambacher:
The HashIndexReader index does not properly read the properties from the SST metadata to restore the hash index prefixes.  This leads to HashIndexReader that does not have a BlockPrefixIndex.

The underlying problem is that Metadata blocks contain property names -- which are simple strings -- but are treated as if they contain Internal keys (by NewDataIterator).  Because they are treated as Internal keys, the last 8 bytes are truncated from the property names.  When the code goes to search for the names, it fails to find the proper name and things blow up.

In this particular case, one issue is that two keys are identical ("rocksdb.hashindex.prefixes" and "rocksdb.hashindex.metadata") in all but the last 8 characters.

If this problem is fixed, two tests will fail:
./db_with_timestamp_basic_test --gtest_filter=Timestamp/DBBasicTestWithTimestampTableOptions.SeekWithPrefixLongerThanKey/1
 ./db_with_timestamp_basic_test --gtest_filter=Timestamp/DBBasicTestWithTimestampTableOptions.MultiGetPrefixFilter/1

mrambacher:
Here is a test that shows the wrong key being returned for the "rocksdb.hashindex.prefixes" key, which is causing the other issues:

+  
TEST_P(BlockBasedTableTest, SeekMetaBlocks) {
  // The properties meta-block should come at the end since we always need to
  // read it when opening a file, unlike index/filter/other meta-blocks, which
  // are sometimes read depending on the user's configuration. This ordering
  // allows us to do a small readahead on the end of the file to read properties
  // and meta-index blocks with one I/O.
  TableConstructor c(BytewiseComparator(), true /* convert_to_internal_key_ */);
  c.Add("foo_a1", "val1");
  c.Add("foo_b2", "val2");
  c.Add("foo_c3", "val3");
  c.Add("foo_d4", "val4");
  c.Add("foo_e5", "val5");
  c.Add("foo_f6", "val6");
  c.Add("foo_g7", "val7");
  c.Add("foo_h8", "val8");
  c.Add("foo_j9", "val9");

  // write an SST file
  Options options;
  BlockBasedTableOptions table_options = GetBlockBasedTableOptions();
  table_options.index_type = BlockBasedTableOptions::kHashSearch;
  table_options.filter_policy.reset(NewBloomFilterPolicy(
      8 /* bits_per_key */, false /* use_block_based_filter */));
[...]

PR #8692, opened 2021-08-23
Add NewMetaDataIterator method

mrambacher:
Fixes a problem where the iterator for metadata was being treated as a non-user key when in fact it was a user key.  This led to a problem where the property keys could not be searched for correctly.

The main exposure of this problem was that the HashIndexReader could not get the "prefixes" property correctly, resulting in the failure of retrieval/creation of the BlockPrefixIndex.

Added BlockBasedTableTest.SeekMetaBlocks test to validate this condition.

Fixing this condition exposed two other tests (SeekWithPrefixLongerThanKey, MultiGetPrefixFilter) that passed incorrectly previously and now failed.  Updated those two tests to pass.  Not sure if the tests are functionally correct/still appropriate, but made them pass...


mrambacher:
Fixes #8673

table/block_based/block.cc
```diff
@@ -968,6 +968,13 @@ Block::Block(BlockContents&& contents, size_t read_amp_bytes_per_bit,
   }
 }
 
+DataBlockIter* Block::NewMetaDataIterator() {
+  DataBlockIter* ret_iter =
+      NewDataIterator(BytewiseComparator(), kDisableGlobalSequenceNumber);
+  ret_iter->raw_key_.SetIsUserKey(true);
```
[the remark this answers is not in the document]
mrambacher: There are other differences between a DataBlockIter and a MetaBlockIter:
- MetaBlockIter always use BytewiseComparator; DatBlockIter uses one associated with the table;
- MetaBlockIter uses non-assert functions for parsing keys (CheckAndDecodeEntry) whereas DataBlockIter asserts (DecodeEntry)
- MetaBlockIter will not change read amplification and is not reflected in the Statistics

The naming of "MetaBlockIter" was chosen as it is used to read from a variable called "metablock_index" in a method called "FindMetaBlock".  

If you prefer a different name, feel free to suggest one.

table/block_based/block.cc
```diff
@@ -607,12 +675,12 @@ bool IndexBlockIter::ParseNextIndexKey() {
 // shared_size is 0, which included the first entry in each restart point.
 // Otherwise the format is delta-size = block handle size - size of last block
 // handle.
-void IndexBlockIter::DecodeCurrentValue(uint32_t shared) {
+void IndexBlockIter::DecodeCurrentValue(bool is_shared) {
   Slice v(value_.data(), data_ + restarts_ - value_.data());
   // Delta encoding is used if `shared` != 0.
```
pdillinger, line 680: Update comment

pdillinger (commented):
This needs performance testing. Changing templates (static polymorphism) to virtual functions (dynamic polymorphism) is probably a regression in this hot code, possibly also heap allocating objects that can go on the stack.

(Doesn't have to be here and now...) I think we're more likely to see net improvement if we change from an "iterator" interface to a "visitor" interface for meta blocks. Metaindex and properties should be parsed once on read, never on the fly. I don't know how much code that would involve, but naively it would seem less entangled with data and index block iteration.

mrambacher:
> This needs performance testing. Changing templates (static polymorphism) to virtual functions (dynamic polymorphism) is probably a regression in this hot code, possibly also heap allocating objects that can go on the stack.
> 
> (Doesn't have to be here and now...) I think we're more likely to see net improvement if we change from an "iterator" interface to a "visitor" interface for meta blocks. Metaindex and properties should be parsed once on read, never on the fly. I don't know how much code that would involve, but naively it would seem less entangled with data and index block iteration.

@pdillinger I will run some performance tests but the Iterator class already has virtual functions.  This PR is not introducing any polymorphism.

table/block_based/block.cc
```diff
@@ -709,9 +776,9 @@ bool BlockIter<TValue>::BinarySeek(const Slice& target, uint32_t* index,
     // The `mid` is computed by rounding up so it lands in (`left`, `right`].
     int64_t mid = left + (right - left + 1) / 2;
     uint32_t region_offset = GetRestartPoint(static_cast<uint32_t>(mid));
-    uint32_t shared, non_shared;
-    const char* key_ptr = DecodeKeyFunc()(
-        data_ + region_offset, data_ + restarts_, &shared, &non_shared);
+    uint32_t shared, non_shared, dummy;
+    const char* key_ptr = DecodeKV(data_ + region_offset, data_ + restarts_,
```
pdillinger, line 780: DecodeKeyFunc -> DecodeKV (static -> dynamic polymorphism)

mrambacher:
> This needs performance testing. Changing templates (static polymorphism) to virtual functions (dynamic polymorphism) is probably a regression in this hot code, possibly also heap allocating objects that can go on the stack.
> 
> (Doesn't have to be here and now...) I think we're more likely to see net improvement if we change from an "iterator" interface to a "visitor" interface for meta blocks. Metaindex and properties should be parsed once on read, never on the fly. I don't know how much code that would involve, but naively it would seem less entangled with data and index block iteration.

I ran the following command on both main and on this PR:
TEST_TMPDIR=/dev/shm ./db_bench -benchmarks=filluniquerandom,compact -target_file_size_base=1048576 -compression_type=none
TEST_TMPDIR=/dev/shm ./db_bench -readonly=1 -use_existing_db=1 -benchmarks=seekrandom -seek_nexts=1000 -target_file_size_base=1048576 -cache_size=0 -duration=30

Results for main:
seekrandom   :     164.420 micros/op 6081 ops/sec;  672.5 MB/s (182999 of 182999 found)
seekrandom   :     165.281 micros/op 6050 ops/sec;  669.0 MB/s (181999 of 181999 found)
seekrandom   :     168.022 micros/op 5951 ops/sec;  658.1 MB/s (178999 of 178999 found)

Results for this PR:
seekrandom   :     165.602 micros/op 6038 ops/sec;  667.7 MB/s (181999 of 181999 found)
seekrandom   :     165.443 micros/op 6044 ops/sec;  668.3 MB/s (181999 of 181999 found)
seekrandom   :     169.142 micros/op 5912 ops/sec;  653.7 MB/s (177999 of 177999 found)

pdillinger (approved):
LGTM. I ran my own perf test as in #9126 and if anything, this is slightly faster than baseline, 122474 vs. 121924 ops/sec.

pdillinger:
Note: FB test coverage indicates MetaBlockIter::SeekForPrevImpl is not covered.

mrambacher:
> Note: FB test coverage indicates MetaBlockIter::SeekForPrevImpl is not covered.

I am not surprised this is not covered.  The method is required by the signature but everything that uses MetaBlocks either does a seek or scan.  I can delete the contents if you wish and have it assert/fail.

pdillinger:
> I am not surprised this is not covered. The method is required by the signature but everything that uses MetaBlocks either does a seek or scan. I can delete the contents if you wish and have it assert/fail.

I don't have a specific preference

table/block_based/block.h
```diff
@@ -582,12 +588,30 @@ class DataBlockIter final : public BlockIter<Slice> {
...
+    return value_;
+  }
+
+ protected:
+  virtual void SeekToFirstImpl() override;
+  virtual void SeekToLastImpl() override;
+  virtual void SeekImpl(const Slice& target) override;
+  virtual void SeekForPrevImpl(const Slice& target) override;
+  virtual void NextImpl() override;
+  virtual void PrevImpl() override;
```
riversand963, line 614: https://stackoverflow.com/questions/39932391/should-i-use-virtual-override-or-both-keywords
Please remove 'virtual'.

table/block_based/block.cc
```diff
@@ -309,23 +359,21 @@ bool DataBlockIter::SeekForGetImpl(const Slice& target) {
...
-  } else {
-    limit = data_ + restarts_;
+  uint32_t limit = restarts_;
+  if (restart_index + 1 < num_restarts_) {
+    limit = GetRestartPoint(restart_index + 1);
   }
-
-  while (true) {
+  while (current_ < limit) {
+    bool shared;
```
riversand963, line 369: Init

table/block_based/block.h
```diff
@@ -582,12 +588,30 @@ class DataBlockIter final : public BlockIter<Slice> {
...
 
-  template <typename DecodeEntryFunc>
-  inline bool ParseNextDataKey(const char* limit = nullptr);
-
   bool SeekForGetImpl(const Slice& target);
-  void NextOrReportImpl();
-  void SeekToFirstOrReportImpl();
+};
+
+class MetaBlockIter final : public BlockIter<Slice> {
```
riversand963, line 594: Please add some comment describing why we need this class.

db/db_with_timestamp_basic_test.cc
```diff
@@ -772,11 +772,13 @@ TEST_P(DBBasicTestWithTimestampTableOptions, SeekWithPrefixLessThanKey) {
...
 }
 
-TEST_P(DBBasicTestWithTimestampTableOptions, SeekWithPrefixLongerThanKey) {
+TEST_P(DBBasicTestWithTimestampTableOptions, SeekWithCappedPrefix) {
   Options options = CurrentOptions();
   options.env = env_;
   options.create_if_missing = true;
-  options.prefix_extractor.reset(NewFixedPrefixTransform(20));
+  // All of the keys or this test must be longer than 3 characters
+  static int kMinKeyLen = 3;
```
riversand963, line 780: Nit: can be constexpr

table/block_based/block.cc
```diff
@@ -462,7 +536,20 @@ void DataBlockIter::SeekToLastImpl() {
     return;
   }
   SeekToRestartPoint(num_restarts_ - 1);
-  while (ParseNextDataKey<DecodeEntry>() && NextEntryOffset() < restarts_) {
+  bool is_shared;
```
riversand963, line 539: Nit: better to init

table/block_based/block.cc
```diff
@@ -432,20 +480,46 @@ void DataBlockIter::SeekForPrevImpl(const Slice& target) {
...
 }
 
-void DataBlockIter::SeekToFirstOrReportImpl() {
+void MetaBlockIter::SeekToFirstImpl() {
   if (data_ == nullptr) {  // Not init yet
     return;
   }
   SeekToRestartPoint(0);
-  ParseNextDataKey<CheckAndDecodeEntry>();
+  bool is_shared;
```
riversand963, line 521: better to init

table/block_based/block.cc
```diff
@@ -432,20 +480,46 @@ void DataBlockIter::SeekForPrevImpl(const Slice& target) {
...
+  }
+}
+
 void DataBlockIter::SeekToFirstImpl() {
   if (data_ == nullptr) {  // Not init yet
     return;
   }
   SeekToRestartPoint(0);
-  ParseNextDataKey<DecodeEntry>();
+  bool is_shared;
```
riversand963, line 512: better to init

table/block_based/block.cc
```diff
@@ -153,6 +165,27 @@ void IndexBlockIter::PrevImpl() {
...
+    if (restart_index_ == 0) {
+      // No more entries
+      current_ = restarts_;
+      restart_index_ = num_restarts_;
+      return;
+    }
+    restart_index_--;
+  }
+  SeekToRestartPoint(restart_index_);
+  bool is_shared;
```
riversand963, line 182: Nit: init

table/block_based/block.cc
```diff
@@ -462,7 +536,20 @@ void DataBlockIter::SeekToLastImpl() {
...
+    // Keep skipping
+  }
+}
+
+void MetaBlockIter::SeekToLastImpl() {
+  if (data_ == nullptr) {  // Not init yet
+    return;
+  }
+  SeekToRestartPoint(num_restarts_ - 1);
+  bool is_shared;
```
riversand963, line 550: Nit: init

table/block_based/block.cc
```diff
@@ -126,10 +126,22 @@ struct DecodeKeyV4 {
   }
 };
 
-void DataBlockIter::NextImpl() { ParseNextDataKey<DecodeEntry>(); }
+struct DecodeEntryV4 {
+  inline const char* operator()(const char* p, const char* limit,
+                                uint32_t* shared, uint32_t* non_shared,
+                                uint32_t* value_length) {
+    *value_length = 0;
```
riversand963, line 133: Nit: our internal linter may complain about nullptr ref. So assert(value_length).

table/block_based/block.h
```diff
@@ -582,12 +588,30 @@ class DataBlockIter final : public BlockIter<Slice> {
...
+ public:
+  void Initialize(const char* data, uint32_t restarts, uint32_t num_restarts,
+                  bool block_contents_pinned) {
+    InitializeBase(BytewiseComparator(), data, restarts, num_restarts,
+                   kDisableGlobalSequenceNumber, block_contents_pinned);
+    raw_key_.SetIsUserKey(true);
```
riversand963, line 600: Given this is the core fix, maybe a line or two of explanation helps emphasizing the significance.

riversand963 (changes requested):
I am not expert at block-related code, and rely on the author's benchmarking to demonstrate no regression on the hot path.

We follow https://google.github.io/styleguide/cppguide.html#Local_Variables, and as a good practice, we should always initialize variables. Relying on a subsequent function call to always set the value sounds risky.

riversand963 (approved):
LGTM. Thanks @mrambacher for the fix.

PR #8693, opened 2021-08-23
Add port::GetProcessID()

pdillinger:
Summary: Useful in some places for object uniqueness across processes.
Currently used for generating a host-wide identifier of Cache objects
but expected to be used soon in some unique id generation code.

`int64_t` is chosen for return type because POSIX uses signed integer type,
usually `int`, for `pid_t` and Windows uses `DWORD`, which is `uint32_t`.

Future work: avoid copy-pasted declarations in port_*.h, perhaps with
port_common.h always included from port.h

Test Plan: manual for now

### DIFF
--- a/db/internal_stats.cc
+++ b/db/internal_stats.cc
@@ -23,6 +23,7 @@
 #include "cache/cache_entry_stats.h"
 #include "db/column_family.h"
 #include "db/db_impl/db_impl.h"
+#include "port/port.h"
 #include "rocksdb/system_clock.h"
 #include "rocksdb/table.h"
 #include "table/block_based/cachable_entry.h"
@@ -602,7 +603,8 @@ void InternalStats::CacheEntryRoleStats::BeginCollection(
   ++collection_count;
   role_map_ = CopyCacheDeleterRoleMap();
   std::ostringstream str;
-  str << cache->Name() << "@" << static_cast<void*>(cache);
+  str << cache->Name() << "@" << static_cast<void*>(cache) << "#"
+      << port::GetProcessId();
   cache_id = str.str();
   cache_capacity = cache->GetCapacity();
 }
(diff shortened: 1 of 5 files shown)


ajkr (approved):
LGTM

include/rocksdb/env.h
```diff
@@ -619,6 +619,9 @@ class Env {
   // Returns the ID of the current thread.
   virtual uint64_t GetThreadID() const;
 
+  // Returns the current process ID.
+  virtual int64_t GetProcessID() const = 0;
```
anand1976, line 623: Declaring it as pure virtual might break some code compilation. Can we provide a default implementation that calls a corresponding function in ```port/```?

pdillinger: I'm just trying to follow the pattern, but unlike GetThreadID, GetProcessID does not have a platform-independent implementation via C++ APIs. I don't really understand the point of having functions like this in Env anyway. I've been criticized for not putting things with platform-specific implementations in Env that didn't have a specific need to be in the public API (criticized by @mrambacher re: GenerateRfcUuidFromPlatform in #8362).

anand1976: Agree about not having it in Env. I don't think anyone would have a reason to override it, which seems to be the main advantage of having it in Env. How about defining it in ```port_posix.cc``` (```getpid``` is Posix anyway) and ```port_win.cc```?

mrambacher: @pdillinger I (mostly) agree about it not being in Env.  At the moment, the Env is the only container for port-specific values.  Since this method (along with ThreadID) seems unlikely to be overridden by anything, the only advantage to putting it in Env is if the function was going to be publicly exposed (since port/* is not in the pubic API).  In reality, most of the methods that are left as truly "Env" methods (and not in FileSystem or SystemClock) fall into that same pattern (where they will not be overridden by implementation).

A slightly different approach would be to make a "static Env::GetProcessID()" method and have that method call the appropriate port/* method.  This allows exposure without the concerns of virtual/override and implementations.

mrambacher (approved):
LGTM.

anand1976 (approved):
LGTM

PR #8697, opened 2021-08-24
Simplify `TraceAnalyzer`

autopear:
Handler functions now use a common output function to output to stdout/files.

Test:
`trace_analyzer_test` can pass.

### DIFF
--- a/tools/trace_analyzer_tool.cc
+++ b/tools/trace_analyzer_tool.cc
@@ -132,7 +132,7 @@ DEFINE_bool(analyze_single_delete, false, "Analyze the SingleDelete query.");
 DEFINE_bool(analyze_range_delete, false, "Analyze the DeleteRange query.");
 DEFINE_bool(analyze_merge, false, "Analyze the Merge query.");
 DEFINE_bool(analyze_iterator, false,
-            " Analyze the iterate query like seek() and seekForPrev().");
+            " Analyze the iterate query like Seek() and SeekForPrev().");
 DEFINE_bool(analyze_multiget, false,
             " Analyze the MultiGet query. NOTE: for"
             " MultiGet, we analyze each KV-pair read in one MultiGet query. "
@@ -280,13 +280,14 @@ TraceAnalyzer::TraceAnalyzer(std::string& trace_path, std::string& output_path,
   total_access_keys_ = 0;
   total_gets_ = 0;
   total_writes_ = 0;
+  total_seeks_ = 0;
+  total_seek_prevs_ = 0;
+  total_multigets_ = 0;
   trace_create_time_ = 0;
   begin_time_ = 0;
   end_time_ = 0;
   time_series_start_ = 0;
   cur_time_sec_ = 0;
-  // Set the default trace file version as version 0.2
-  trace_file_version_ = 2;
   if (FLAGS_sample_ratio > 1.0 || FLAGS_sample_ratio <= 0) {
     sample_max_ = 1;
   } else {
@@ -360,7 +361,11 @@ TraceAnalyzer::~TraceAnalyzer() {}
 Status TraceAnalyzer::PrepareProcessing() {
   Status s;
   // Prepare the trace reader
-  s = NewFileTraceReader(env_, env_options_, trace_name_, &trace_reader_);
+  if (trace_reader_ == nullptr) {
+    s = NewFileTraceReader(env_, env_options_, trace_name_, &trace_reader_);
+  } else {
+    s = trace_reader_->Reset();
+  }
   if (!s.ok()) {
     return s;
   }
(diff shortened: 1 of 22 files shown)


tools/trace_analyzer_tool.cc
```diff
@@ -1504,322 +1582,176 @@ Status TraceAnalyzer::Handle(const WriteQueryTraceRecord& record,
...
-  if (ta_[TraceOperationType::kGet].sample_count > 0) {
-    ta_[TraceOperationType::kGet].sample_count++;
-    return Status::OK();
-  }
-  ta_[TraceOperationType::kGet].sample_count++;
-
-  if (!ta_[TraceOperationType::kGet].enabled) {
-    return Status::OK();
-  }
-  value_size = 10;
```
autopear, line 1549: @zhichao-cao  Why change value size to 10?

zhichao-cao: In the Get trace record, we do not collect the value. So, we do not have the "value" part. In order to simplify the later on calculation, we just give a fixed value. Maybe it would be better to add a comments here to make it more clear

tools/trace_analyzer_tool.h
```diff
@@ -164,8 +165,89 @@ struct CfUnit {
   std::map<uint32_t, uint32_t> cf_qps;
 };
 
+class TraceAnalysisResult : public TraceRecordResult {
```
zhichao-cao, line 168: So, each query type will have a TraceAnalysisResult to hold all the trace results? What if the trace size is TB level, we cannot hold them in memory?

zhichao-cao:
In general, trace file can be TB level size. Trace analyzer will consume a lot of memory space to hold the information (e.g., key, value and others). Therefore, we should be careful when adding the data structure. Usually, user needs to first estimate the memory consumption, then, using the analyzing options to decide what should be kept in memory in this round of analyzing. I'm not sure, if we introduce the TraceAnalysisReults is memory efficient.

tools/trace_analyzer_tool.h
```diff
@@ -164,8 +165,89 @@ struct CfUnit {
...
+ public:
+  // All operations must have the same timestamp and trace type, e.g, MultiGet.
+  TraceAnalysisResult(TraceOperationType op_type, std::vector<uint32_t> cf_ids,
```
zhichao-cao, line 171: What if in the future, we wan to get some special results from different types? E.g., for iterator, we my need to get the lower and upper bound? So, is a general TraceAnalysisResult enough?

tools/trace_analyzer_tool.cc
```diff
@@ -1504,322 +1614,187 @@ Status TraceAnalyzer::Handle(const WriteQueryTraceRecord& record,
...
 // Handle the Merge request in the write batch of the trace
 Status TraceAnalyzer::MergeCF(uint32_t column_family_id, const Slice& key,
                               const Slice& value) {
+  assert(write_result_ != nullptr);
+  write_result_->Reset(TraceOperationType::kMerge, column_family_id, key,
+                       value.size());
+  return write_result_->Accept(this);
+}
+
+Status TraceAnalyzer::Handle(const TraceAnalysisResult& result) {
```
zhichao-cao, line 1754: Need some comments here to explain what is the job of this "Handle".

tools/trace_analyzer_tool.h
autopear: A general `TraceAnalysisResult` may not be enough for all potential special requirements, but at least now it's better to have individual functions with mostly the same code for each operation type.

As for iterator's lower/upper bound, they are used when creating an iterator. I guess later we may need to have some iterator creation trace type, then, the 2 bounds can be handled there.

tools/trace_analyzer_tool.cc
autopear: ACK

tools/trace_analyzer_tool.cc
autopear: ACK.

It is used to handle the generic analysis result for outputting purpose. If we don't use `TraceAnalysisResult`, we can probably replace it with a single function that does the same job.

tools/trace_analyzer_tool.cc
autopear: One more question, why we don't do the same thing for other key only operations (seeks, deletes, etc.)?

zhichao-cao: obviously, we forgot to do the same thing for other non-value recorded operations........

tools/trace_analyzer_tool.cc
```diff
@@ -169,6 +168,8 @@ DEFINE_double(sample_ratio, 1.0,
 
 namespace ROCKSDB_NAMESPACE {
 
+const size_t kShadowValueSize = 10;
```
zhichao-cao, line 171: Maybe add comment to explain why we need it.

zhichao-cao (approved):
LGTM, thanks for the code change, which makes it more clear and simplier.

tools/trace_analyzer_tool.cc
autopear: There's a brief description in the header file.

PR #8699, opened 2021-08-24
Batch blob read IO for MultiGet

riversand963:
In batched `MultiGet()`, RocksDB batches blob read IO and uses `RandomAccessFileReader::MultiRead()`
to read the blobs instead of issuing multiple `Read()`.

Test plan:
```
make check
```

### DIFF
--- a/db/blob/blob_file_reader.cc
+++ b/db/blob/blob_file_reader.cc
@@ -426,6 +426,9 @@ void BlobFileReader::MultiGetBlob(
   }
 
   for (size_t i = 0; i < num_blobs; ++i) {
+    if (!read_reqs[i].status.ok() || !statuses[i]->ok()) {
+      continue;
+    }
     const Slice& record_slice = read_reqs[i].result;
     const Slice value_slice(record_slice.data() + adjustments[i],
                             value_sizes[i]);
--- a/db/blob/blob_file_reader.cc
+++ b/db/blob/blob_file_reader.cc
@@ -351,15 +351,16 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
   return Status::OK();
 }
 
-Status BlobFileReader::MultiGetBlob(
+void BlobFileReader::MultiGetBlob(
     const ReadOptions& read_options,
     const autovector<std::reference_wrapper<Slice>>& user_keys,
     const autovector<uint64_t>& offsets,
-    const autovector<uint64_t>& value_sizes, autovector<PinnableSlice*>& values,
-    uint64_t* bytes_read) const {
+    const autovector<uint64_t>& value_sizes, autovector<Status*>& statuses,
+    autovector<PinnableSlice*>& values, uint64_t* bytes_read) const {
   const size_t num_blobs = user_keys.size();
   assert(num_blobs == offsets.size());
   assert(num_blobs == value_sizes.size());
+  assert(num_blobs == statuses.size());
   assert(num_blobs == values.size());
 
   std::vector<FSReadRequest> read_reqs(num_blobs);
@@ -400,15 +401,26 @@ Status BlobFileReader::MultiGetBlob(
   s = file_reader_->MultiRead(IOOptions(), read_reqs.data(), read_reqs.size(),
                               direct_io ? &aligned_buf : nullptr);
   if (!s.ok()) {
-    return s;
+    for (auto& req : read_reqs) {
+      req.status.PermitUncheckedError();
+    }
+    for (size_t i = 0; i < num_blobs; ++i) {
+      assert(statuses[i]);
+      *statuses[i] = s;
+    }
+    return;
   }
 
   if (read_options.verify_checksums) {
     for (size_t i = 0; i < num_blobs; ++i) {
+      if (!read_reqs[i].status.ok()) {
+        continue;
+      }
       const Slice& record_slice = read_reqs[i].result;
       s = VerifyBlob(record_slice, user_keys[i], value_sizes[i]);
       if (!s.ok()) {
-        return s;
+        assert(statuses[i]);
+        *statuses[i] = s;
       }
     }
   }
@@ -420,7 +432,8 @@ Status BlobFileReader::MultiGetBlob(
     s = UncompressBlobIfNeeded(value_slice, compression_type_, clock_,
                                statistics_, values[i]);
     if (!s.ok()) {
-      return s;
+      assert(statuses[i]);
+      *statuses[i] = s;
     }
   }
 
@@ -431,7 +444,6 @@ Status BlobFileReader::MultiGetBlob(
     }
     *bytes_read = total_bytes;
   }
-  return Status::OK();
 }
 
 Status BlobFileReader::VerifyBlob(const Slice& record_slice,
--- a/db/blob/blob_file_reader.cc
+++ b/db/blob/blob_file_reader.cc
@@ -380,6 +380,8 @@ void BlobFileReader::MultiGetBlob(
     total_len += read_reqs[i].len;
   }
 
+  RecordTick(statistics_, BLOB_DB_BLOB_FILE_BYTES_READ, total_len);
+
   Buffer buf;
   AlignedBuf aligned_buf;
 
@@ -411,9 +413,20 @@ void BlobFileReader::MultiGetBlob(
     return;
   }
 
+  assert(s.ok());
+  for (size_t i = 0; i < num_blobs; ++i) {
+    auto& req = read_reqs[i];
+    assert(statuses[i]);
+    if (req.status.ok() && req.result.size() != req.len) {
+      req.status = IOStatus::Corruption("Failed to read data from blob file");
+    }
+    *statuses[i] = req.status;
+  }
+
   if (read_options.verify_checksums) {
     for (size_t i = 0; i < num_blobs; ++i) {
-      if (!read_reqs[i].status.ok()) {
+      assert(statuses[i]);
+      if (!statuses[i]->ok()) {
         continue;
       }
       const Slice& record_slice = read_reqs[i].result;
@@ -426,7 +439,8 @@ void BlobFileReader::MultiGetBlob(
   }
 
   for (size_t i = 0; i < num_blobs; ++i) {
-    if (!read_reqs[i].status.ok() || !statuses[i]->ok()) {
+    assert(statuses[i]);
+    if (!statuses[i]->ok()) {
       continue;
     }
     const Slice& record_slice = read_reqs[i].result;
@@ -435,7 +449,6 @@ void BlobFileReader::MultiGetBlob(
     s = UncompressBlobIfNeeded(value_slice, compression_type_, clock_,
                                statistics_, values[i]);
     if (!s.ok()) {
-      assert(statuses[i]);
       *statuses[i] = s;
     }
   }
--- a/db/blob/blob_file_reader.cc
+++ b/db/blob/blob_file_reader.cc
@@ -353,7 +353,7 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
 
 void BlobFileReader::MultiGetBlob(
     const ReadOptions& read_options,
-    const autovector<std::reference_wrapper<Slice>>& user_keys,
+    const autovector<std::reference_wrapper<const Slice>>& user_keys,
     const autovector<uint64_t>& offsets,
     const autovector<uint64_t>& value_sizes, autovector<Status*>& statuses,
     autovector<PinnableSlice*>& values, uint64_t* bytes_read) const {
@@ -391,9 +391,9 @@ void BlobFileReader::MultiGetBlob(
     }
   } else {
     buf.reset(new char[total_len]);
-    uint64_t pos = 0;
+    std::ptrdiff_t pos = 0;
     for (size_t i = 0; i < read_reqs.size(); ++i) {
-      read_reqs[i].scratch = pos + buf.get();
+      read_reqs[i].scratch = buf.get() + pos;
       pos += read_reqs[i].len;
     }
   }
--- a/db/blob/blob_file_reader.cc
+++ b/db/blob/blob_file_reader.cc
@@ -264,6 +264,17 @@ Status BlobFileReader::ReadFromFile(const RandomAccessFileReader* file_reader,
   return Status::OK();
 }
 
+Status BlobFileReader::MultiReadFromFile(
+    const RandomAccessFileReader* file_reader,
+    const autovector<uint64_t>& read_offsets,
+    const autovector<size_t>& read_sizes, Statistics* statistics) {
+  assert(file_reader);
+  (void)read_offsets;
+  (void)read_sizes;
+  (void)statistics;
+  return Status::OK();
+}
+
 BlobFileReader::BlobFileReader(
     std::unique_ptr<RandomAccessFileReader>&& file_reader, uint64_t file_size,
     CompressionType compression_type, SystemClock* clock,
@@ -348,6 +359,23 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
     *bytes_read = record_size;
   }
 
+  return Status::OK();
+}
+
+Status BlobFileReader::MultiGetBlob(
+    const ReadOptions& read_options,
+    const autovector<std::reference_wrapper<Slice>>& user_keys,
+    const autovector<uint64_t>& offsets,
+    const autovector<uint64_t>& value_sizes,
+    const autovector<CompressionType>& compression_types,
+    autovector<PinnableSlice*>& values, uint64_t* bytes_read) const {
+  (void)read_options;
+  (void)user_keys;
+  (void)offsets;
+  (void)value_sizes;
+  (void)compression_types;
+  (void)values;
+  (void)bytes_read;
   return Status::OK();
 }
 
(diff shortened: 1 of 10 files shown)


db/blob/blob_file_reader.cc
```diff
@@ -351,6 +351,101 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
...
+  assert(num_blobs == value_sizes.size());
+  assert(num_blobs == statuses.size());
+  assert(num_blobs == values.size());
+
+  std::vector<FSReadRequest> read_reqs(num_blobs);
+  autovector<uint64_t> adjustments;
+  uint64_t total_len = 0;
+  for (size_t i = 0; i < num_blobs; ++i) {
+    const size_t key_size = user_keys[i].get().size();
+    assert(IsValidBlobOffset(offsets[i], key_size, value_sizes[i], file_size_));
```
ltamasi, line 371: Is this currently checked by the caller (i.e. `Version`)? I think we might prefer to encapsulate/hide low-level blob file format related stuff like this in `BlobFileReader`

db/blob/blob_file_reader.cc
```diff
@@ -351,6 +351,101 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
...
+    const uint64_t adjustment =
+        read_options.verify_checksums
+            ? BlobLogRecord::CalculateAdjustmentForRecordHeader(key_size)
+            : 0;
+    assert(offsets[i] >= adjustment);
+    adjustments.push_back(adjustment);
+    read_reqs[i].offset = offsets[i] - adjustment;
+    read_reqs[i].len = value_sizes[i] + adjustment;
```
ltamasi, line 379: We could theoretically merge requests for adjacent bytes ranges here. (I think we could do this as a separate optimization though.)

db/blob/blob_file_reader.cc
```diff
@@ -351,6 +351,101 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
...
+    for (size_t i = 0; i < read_reqs.size(); ++i) {
+      read_reqs[i].scratch = nullptr;
+    }
+  } else {
+    buf.reset(new char[total_len]);
+    uint64_t pos = 0;
+    for (size_t i = 0; i < read_reqs.size(); ++i) {
+      read_reqs[i].scratch = pos + buf.get();
+      pos += read_reqs[i].len;
+    }
```
ltamasi, line 398: How about `read_reqs[i].scratch = buf.get() + pos;` ? Also, `pos` could be `ptrdiff_t`

db/db_impl/db_impl.cc
```diff
@@ -2389,17 +2390,9 @@ void DBImpl::PrepareMultiGetKeys(
...
-
-    for (size_t index = 1; index < sorted_keys->size(); ++index) {
-      const KeyContext* const lhs = (*sorted_keys)[index - 1];
-      const KeyContext* const rhs = (*sorted_keys)[index];
-
-      // lhs should be <= rhs, or in other words, rhs should NOT be < lhs
-      assert(!key_context_less(rhs, lhs));
-    }
+    assert(std::is_sorted(sorted_keys->begin(), sorted_keys->end(),
+                          CompareKeyContext()));
```
ltamasi, line 2394: 👍👍

db/version_set.cc
```diff
@@ -1863,6 +1864,118 @@ Status Version::GetBlob(const ReadOptions& read_options, const Slice& user_key,
...
+        *(blob_it->s) = Status::Corruption("Invalid blob offset");
+        continue;
+      }
+      if (blob_index.compression() != compression) {
+        *(blob_it->s) =
+            Status::Corruption("Compression type mismatch when reading a blob");
+        continue;
+      }
+      blob_read_its.emplace_back(blob_it);
+      user_keys.emplace_back(std::ref(blob_it->ukey_with_ts));
```
ltamasi, line 1946: `std::cref` ?

db/version_set.cc
```diff
@@ -1863,6 +1864,118 @@ Status Version::GetBlob(const ReadOptions& read_options, const Slice& user_key,
...
+    for (auto it = range.blob_begin(); it != range.blob_end(); ++it) {
+      assert(it->s->ok());
+      *(it->s) = s;
+    }
+    return;
+  }
+
+  using BlobIterator = MultiGetRange::BlobIterator;
+  std::unordered_map<uint64_t, std::vector<std::pair<BlobIndex, BlobIterator>>>
+      blob_rqs;
```
ltamasi, line 1883: I'm wondering if we built this map in `Version::MultiGet` as we're encountering the blob references, would that eliminate the need for the new mask and related logic in `MultiGetContext`?

ltamasi (commented):
Thanks so much for the patch @riversand963 !

db/blob/blob_file_reader.cc
riversand963: Yeah, that's what I thought, and I have a TODO in `Version::MultiGetBlob()`.

db/version_set.cc
riversand963: Done.
The benefit of an extra mask `blob_mask_` is that we can conveniently iterate over the multiget requests and locate individual requests that needs blob reads. It can be removed, though.

db/blob/blob_file_reader.cc
riversand963: @ltamasi  Yes.
Encapsulating low-level stuff in `BlobFileReader` is desirable. However, `RandomAccessFileReader::MultiRead()` requires a contiguous buffer for `FSReadRequests`. If we do not do the check in `Version`, then we need to pass `BlobIndex` info to `BlobFileReader` where it will be checked so that `BlobFileReader` will set up the buffer properly. The current implementation of `BlobFileReader` is `BlobIndex`-decoding agnostic. Consequently, I have to do the check in the `Version` layer. Let me know if you have an alternative, without having to construct intermediate vectors for offsets, sizes, and statuses.

ltamasi: Yeah, that makes sense. Thanks for the explanation @riversand963 !

riversand963: I plan to add a few additional unit tests before publishing this diff.

db/version_set.cc
```diff
@@ -2170,24 +2271,24 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
...
-                }
-
-                continue;
+              const Slice& blob_index_slice = *(iter->value);
+              BlobIndex blob_index;
+              Status tmp_s = blob_index.DecodeFrom(blob_index_slice);
+              if (tmp_s.ok()) {
+                const uint64_t blob_file_num = blob_index.file_number();
+                blob_rqs[blob_file_num].emplace_back(
+                    std::make_pair(blob_index, iter));
```
ltamasi, line 2280: I think we might have potential lifecycle issues with storing `iter` in the map since it holds a pointer to `file_range`, which lives on the stack. If so, we could probably store a pointer to the `KeyContext` instead to work around this

db/blob/blob_file_reader.cc
```diff
@@ -351,6 +351,104 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
...
+  if (!s.ok()) {
+    for (auto& req : read_reqs) {
+      req.status.PermitUncheckedError();
+    }
+    for (size_t i = 0; i < num_blobs; ++i) {
+      assert(statuses[i]);
+      *statuses[i] = s;
+    }
+    return;
+  }
```
ltamasi, line 412: I'm wondering if we would want to propagate `read_reqs[i].status` to `*statuses[i]` assuming `s.ok()` (e.g. in an `else` branch). That way, we could make sure per-request failures are not lost and we would not need to care about `read_reqs[i].status` for the rest of the method.

db/blob/blob_file_reader.cc
```diff
@@ -351,6 +351,104 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
...
+    if (!s.ok()) {
+      assert(statuses[i]);
+      *statuses[i] = s;
+    }
+  }
+
+  if (bytes_read) {
+    uint64_t total_bytes = 0;
+    for (const auto& req : read_reqs) {
+      total_bytes += req.result.size();
```
ltamasi, line 446: Question: would it make sense to check that we got as many bytes as we wanted for each request and report `Corruption` if not, like we do in `ReadFromFile` ? Or does `MultiRead` guarantee that?

db/blob/blob_file_reader.cc
```diff
@@ -351,6 +351,104 @@ Status BlobFileReader::GetBlob(const ReadOptions& read_options,
...
+    assert(IsValidBlobOffset(offsets[i], key_size, value_sizes[i], file_size_));
+    const uint64_t adjustment =
+        read_options.verify_checksums
+            ? BlobLogRecord::CalculateAdjustmentForRecordHeader(key_size)
+            : 0;
+    assert(offsets[i] >= adjustment);
+    adjustments.push_back(adjustment);
+    read_reqs[i].offset = offsets[i] - adjustment;
+    read_reqs[i].len = value_sizes[i] + adjustment;
+    total_len += read_reqs[i].len;
```
ltamasi, line 380: Can we record a tick for `BLOB_DB_BLOB_FILE_BYTES_READ` with a value of `total_len` after the loop?

db/blob/blob_file_reader_test.cc
```diff
@@ -181,8 +239,9 @@ TEST_F(BlobFileReaderTest, CreateReaderAndGetBlob) {
     uint64_t bytes_read = 0;
 
     ASSERT_TRUE(reader
-                    ->GetBlob(read_options, key, blob_offset - 1, blob_size,
-                              kNoCompression, &value, &bytes_read)
+                    ->GetBlob(read_options, keys[2], blob_offsets[2] - 1,
```
ltamasi, line 242: Very minor but shouldn't this be the other way round? I.e. `blob_offsets[0] - 1` would be too close to the start and `blob_offsets[2] + 1` too close to the end?

db/blob/blob_file_reader.cc
riversand963: SG. Done

db/blob/blob_file_reader.cc
riversand963: Done.

db/blob/blob_file_reader.cc
riversand963: Done.

db/blob/blob_file_reader_test.cc
```diff
@@ -217,12 +277,42 @@ TEST_F(BlobFileReaderTest, CreateReaderAndGetBlob) {
...
 
-    ASSERT_TRUE(reader
-                    ->GetBlob(read_options, shorter_key,
-                              blob_offset - (sizeof(key) - sizeof(shorter_key)),
-                              blob_size, kNoCompression, &value, &bytes_read)
-                    .IsCorruption());
+    ASSERT_TRUE(
+        reader
+            ->GetBlob(read_options, shorter_key,
+                      blob_offsets[0] - (keys[0].size() - sizeof(shorter_key)),
```
ltamasi, line 283: I think technically there's an off-by-one issue here: I think the correct adjustment would be `keys[0].size() - (sizeof(shorter_key) - 1)`. (Earlier, when both keys were stored in `char` arrays, the two -1's canceled each other out.)

Not sure if it makes a difference though

db/blob/blob_file_reader_test.cc
riversand963: Done

db/blob/blob_file_reader_test.cc
```diff
@@ -232,10 +322,38 @@ TEST_F(BlobFileReaderTest, CreateReaderAndGetBlob) {
     uint64_t bytes_read = 0;
 
     ASSERT_TRUE(reader
-                    ->GetBlob(read_options, incorrect_key, blob_offset,
-                              blob_size, kNoCompression, &value, &bytes_read)
+                    ->GetBlob(read_options, incorrect_key, blob_offsets[0],
```
ltamasi, line 325: I think we might want to change the incorrect key to `foo1` so the size matches

riversand963: Done

db/blob/blob_file_reader_test.cc
riversand963: Done

riversand963:

db/version_set.cc
```diff
@@ -1863,6 +1864,102 @@ Status Version::GetBlob(const ReadOptions& read_options, const Slice& user_key,
...
+
+    // TODO: sort blobs_in_file by file offset.
+    autovector<std::reference_wrapper<const KeyContext>> blob_read_key_contexts;
+    autovector<std::reference_wrapper<const Slice>> user_keys;
+    autovector<uint64_t> offsets;
+    autovector<uint64_t> value_sizes;
+    autovector<Status*> statuses;
+    autovector<PinnableSlice*> values;
+    for (const auto& blob : blobs_in_file) {
+      const auto& blob_index = blob.first;
```
ltamasi, line 1915: We could report `Corruption` here if the blob index is of the TTL or inlined variety (shouldn't happen with the integrated implementation)

db/version_set.cc
```diff
@@ -1863,6 +1864,102 @@ Status Version::GetBlob(const ReadOptions& read_options, const Slice& user_key,
...
+        *(key_context.s) = s;
+      }
+    }
+    return;
+  }
+
+  assert(!blob_rqs.empty());
+  Status status;
+  for (auto& elem : blob_rqs) {
+    uint64_t blob_file_number = elem.first;
```
ltamasi, line 1886: We could check to make sure the blob file is part of the `Version` here for consistency, and report `Corruption` if it's not

db/version_set.cc
```diff
@@ -1863,6 +1864,102 @@ Status Version::GetBlob(const ReadOptions& read_options, const Slice& user_key,
...
+    assert(num == user_keys.size());
+    assert(num == offsets.size());
+    assert(num == value_sizes.size());
+    assert(num == statuses.size());
+    assert(num == values.size());
+    for (size_t i = 0; i < num; ++i) {
+      if (!statuses[i]->ok()) {
+        if (statuses[i]->IsIncomplete()) {
+          auto& get_context = *(blob_read_key_contexts[i].get().get_context);
+          get_context.MarkKeyMayExist();
```
ltamasi, line 1949: I *think* this might belong to the "no I/O" block at the top of the method; I don't think it can be hit in this position

db/version_set.cc
```diff
@@ -1863,6 +1864,102 @@ Status Version::GetBlob(const ReadOptions& read_options, const Slice& user_key,
...
+        }
+      } else {
+        range.AddValueSize(blob_read_key_contexts[i].get().value->size());
+        if (range.GetValueSize() > read_options.value_size_soft_limit) {
+          *(blob_read_key_contexts[i].get().s) = Status::Aborted();
+        } else {
+          *(blob_read_key_contexts[i].get().s) = Status::OK();
```
ltamasi, line 1956: I think it's already set to OK in this case, no? I mean, `blob_read_key_contexts[i].get().s` points to the same `Status` object as `statuses[i]`

db/version_set.cc
```diff
@@ -2085,6 +2182,10 @@ void Version::MultiGet(const ReadOptions& read_options, MultiGetRange* range,
   uint64_t num_data_read = 0;
   uint64_t num_sst_read = 0;
 
+  MultiGetRange keys_with_blobs_range(*range, range->begin(), range->end());
```
ltamasi, line 2185: Do we use this `MultiGetRange` somewhere?

riversand963: It's used in `Version::MultiGetBlob()` for `range.AddValueSize()`.

db/blob/blob_file_reader_test.cc
```diff
@@ -277,12 +277,13 @@ TEST_F(BlobFileReaderTest, CreateReaderAndGetBlob) {
...
-    ASSERT_TRUE(
-        reader
-            ->GetBlob(read_options, shorter_key,
-                      blob_offsets[0] - (keys[0].size() - sizeof(shorter_key)),
-                      blob_sizes[0], kNoCompression, &value, &bytes_read)
-            .IsCorruption());
+    ASSERT_TRUE(reader
+                    ->GetBlob(read_options, shorter_key,
+                              blob_offsets[0] -
+                                  (keys[0].size() - sizeof(shorter_key) - 1),
```
ltamasi, line 283: Sorry, I meant `keys[0].size() - (sizeof(shorter_key) - 1)`  (which is equivalent to `keys[0].size() - sizeof(shorter_key) + 1`) here

db/version_set.cc
ltamasi: Right, thanks :)

db/version_set.cc
riversand963: Done

db/version_set.cc
riversand963: Done

db/version_set.cc
riversand963: Right.

db/blob/blob_file_reader_test.cc
riversand963: Done

db/version_set.cc
riversand963: Make sense

ltamasi (approved):
Looks great!! Thanks so much for the patch @riversand963 !

PR #8701, opened 2021-08-24
Fix legocastle Python commands for CentOS 8

ajkr:
There is no longer an unversioned `python` command that refers to Python
2; the recommended alternative is `/usr/bin/env python2`.


riversand963:
LGTM, but I wonder whether we can just use python3 here for CentOS8

pdillinger:
python2 is not installed by default, and our tools/check_all_python.py suggests it's at least syntax compatible with python3

riversand963 (approved):
LGTM. Thanks @ajkr for the fix

ajkr:
Updated to python3. Jobs still running.

- `/usr/bin/env python2` - https://www.internalfb.com/intern/sandcastle/group/nonce/5099999999999999/
- `/usr/bin/env python3` - https://www.internalfb.com/intern/sandcastle/group/nonce/5100000000000000/

ajkr:
> Updated to python3. Jobs still running.
> 
> * `/usr/bin/env python2` - https://www.internalfb.com/intern/sandcastle/group/nonce/5099999999999999/
> * `/usr/bin/env python3` - https://www.internalfb.com/intern/sandcastle/group/nonce/5100000000000000/

Hm the second one seems to have run on centos 8 and uncovered that /usr/facebook/ops/scripts/asan_symbolize.py uses `!/usr/bin/env python`.

ajkr:
Start a new one that is working - https://www.internalfb.com/intern/sandcastle/group/nonce/5100000000000001/, probably due to CentOS 7. I guess we can land this as long as it works and fix the rest separately.

PR #8703, opened 2021-08-24
Temporarily disable block-based filter when stress testing timestamp

riversand963:
Current implementation does not support user-defined timestamp when
block-based filter is used. Will implement the support in the future, or
wait to see if block-based filter can be deprecated and removed.

Test plan:
make whitebox_crash_test_with_ts


pdillinger (approved):
I was going to suggest disallowing the option combination in DB but it's not so easy to check before already operating, due to techdebt in FilterPolicy.

